Issue #6, opened 2021-09-07
Handle cert trusting in pods / nodes for registry & repos


RothAndrew:
Migrated from https://repo1.dso.mil/platform-one/big-bang/apps/product-tools/zarf/-/issues/13

RothAndrew:
Old, stale. Closing. Can reopen if needed.

PR #38, opened 2021-09-14
Add repo mirroring

RothAndrew:
Add file /etc/rancher/k3s/registries.yaml to init package which will tell k3s to automatically use `zarf.localhost` as a registry mirror for various remote registries

### DIFF
--- a/assets/misc/registries.yaml
+++ b/assets/misc/registries.yaml
@@ -1,0 +1,10 @@
+mirrors:
+  registry1.dso.mil:
+    endpoint:
+      - "http://zarf.localhost"
+  docker.io:
+    endpoint:
+      - "http://zarf.localhost"
+  registry-1.docker.io:
+    endpoint:
+      - "http://zarf.localhost"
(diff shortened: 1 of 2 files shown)


jeff-mccoy:
Thanks, still just had a question on http vs https for the mirror, did you test https and run into cert trust issues?  I think it’s fine if that’s the case, can just roll that into #6.

RothAndrew:
> Thanks, still just had a question on http vs https for the mirror, did you test https and run into cert trust issues? I think it’s fine if that’s the case, can just roll that into #6.

I'll check, I think I just assumed it wouldn't work with https.

RothAndrew:
Ran out of time today, but will definitely look at this tomorrow.

jeff-mccoy:
Confirmed https works.  Thinking about this a little more this actually makes sense because Zarf handles host-level trust of the Zarf ephemeral CA already, so containerd shouldn't get sad either.  Going to merge this now, thanks!

Screen Shot 2021-09-15 at 4 27 34 AM
Screen Shot 2021-09-15 at 4 27 00 AM

jeff-mccoy (approved):

PR #45, opened 2021-09-15
Fixes #41, ensure standard naming for download helm charts

### LINKED ISSUE #41: Tiny-Kafka example doesn't work
The Tiny-Kafka example doesn't work. `zarf package create` works fine, but when running `zarf package deploy` I get the error message:

```
INFO[0008] Copying file                                  Destination=/var/lib/rancher/k3s/server/static/charts/strimzi-kafka-operator-0.24.0.tgz Source=/tmp/zarf-019814346/charts/strimzi-kafka-operator-0.24.0.tgz
FATA[0008] Unable to copy the contents of the asset      Destination=/var/lib/rancher/k3s/server/static/charts/strimzi-kafka-operator-0.24.0.tgz Source=/tmp/zarf-019814346/charts/strimzi-kafka-operator-0.24.0.tgz
```

Digging deeper, it looks to be because the .tgz file that Strimzi publishes is actually named `strimzi-kafka-operator-helm-3-chart-0.24.0.tgz`, and that's what gets saved in the tarball, but Zarf expects the chart files to be named `<ChartName>-<Version>.tgz`

image

I'm submitting a PR to fix the example by pulling it down under `files:` instead of `charts:` but it's just a bandaid fix and deserves to be looked at more closely.

jeff-mccoy:


RothAndrew:
LGTM so far. Pulling down and testing with tiny-kafka example now.

RothAndrew:
Error on creating the init package

image

Still investigating

jeff-mccoy:
boo okay I may have to play with that, I thought that worked for me, but must have been mistaken

RothAndrew:
I wrote a unit test for this. Okay if I push it?

RothAndrew:
The issue lies in using `downloader.VerifyIfPossible`. One isn't being specified when setting up the ChartDownloader.

jeff-mccoy:
Yeah I forgot to go back and check that when I was porting that from helms code, we can just turn off for now

PR #59, opened 2021-09-24
Add resource requests and limits to Utility Cluster deployments

RothAndrew:
## What/Why

- Add resource requests and limits to Gitea and the private docker registry because not having them is a finding in `kubescape` and other k8s security tools

### DIFF
--- a/assets/manifests/utility/gitea.yaml
+++ b/assets/manifests/utility/gitea.yaml
@@ -23,7 +23,6 @@ spec:
                 name: stuart-gitea-http
                 port:
                   number: 3000
-
 ---
 apiVersion: helm.cattle.io/v1
 kind: HelmChart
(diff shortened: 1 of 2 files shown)


jeff-mccoy (commented):
quick question--how did you determine the resource values?  They seem reasonable, just curious

RothAndrew:
> quick question--how did you determine the resource values? They seem reasonable, just curious

I watched their max usage using `k9s`, which was in both cases early on in startup and loading of the package, and made those the requests, then added limits as some multiple of the request that felt right.

jeff-mccoy (approved):
Okay seems reasonable, we probably need to do something to formalize how we define resources for edge use cases where this could really impact where the utility cluster can run

RothAndrew:
Maybe we could do something like specify the minimal amount that it is reasonably expected to work (slowly) and offer the ability to customize it?

PR #60, opened 2021-09-24
Add `kubescape` to Big Bang Core example package

RothAndrew:
## What/Why

- Add `kubescape` to the Big Bang Core example so that it can be used in airgapped environments


jeff-mccoy (commented):
Have you played with the yaml-only scanning in kubescape?  I would love to see this as a first wag at pre-package scanning we've discussed before.  Something that makes the report about the package before you ever run.  

As far as being in the init package, I'd rather keep it under "examples" for now as an add-on capability until we can discuss further.  Super valuable add either way though.

RothAndrew:
> Have you played with the yaml-only scanning in kubescape? I would love to see this as a first wag at pre-package scanning we've discussed before. Something that makes the report about the package before you ever run.
> 
> As far as being in the init package, I'd rather keep it under "examples" for now as an add-on capability until we can discuss further. Super valuable add either way though.

I have not. That sounds awesome, but definitely bigger scope and different use case.

Moving it to examples works for me. I'll add it to the Big Bang Core example.

jeff-mccoy:
Yeah thanks, examples for now is probably good, and then we can look again at it in the future, just trying to be careful of initial binary size, it's already fatter than I would like.  Thanks, this is a really useful addition.

jeff-mccoy (approved):

PR #61, opened 2021-09-24
Misc refactoring and cleanup

RothAndrew:
## What/Why

- `examples/Makefile`: Remove `clean` target as dependency from `fetch-release` and `build-release` and add it to `all` and `all-dev` so that the subtargets can be run without blowing away the whole `sync` folder
- `examples/Vagrantfile`: Increase the `boot_timeout` so that it doesn't sometimes timeout when it really only needs a bit longer to start up
- `values.yaml`: Delete accidentally added file


jeff-mccoy (approved):
LGTM

PR #62, opened 2021-09-24
Update Big Bang Core example with `kubescape` remediations

RothAndrew:
## What/Why

- `examples/big-bang/template/bigbang/values.yaml`: Remediate kubescape findings and add a TODO to fix the control plane CIDR
- `examples/big-bang/zarf.yaml`: Add missing image


jeff-mccoy (approved):
LGTM besides what you already marked as todo.

PR #67, opened 2021-09-27
Upgrade Big Bang Core example to BB v1.17.0

RothAndrew:
## What/Why

- Update Big Bang example to BB v1.17.0
- Misc adjustments
    - Change port forwarding in Vagrantfile to directly forward the 2 new ports rather than expose different ports on the host than are exposed on the guest. From the user's perspective this makes no difference whatsoever but it feels cleaner to me
    - Update examples/big-bang/README.md: Add a table with all the service endpoints, usernames, and passwords, and add a note about needing to delete the `EnvoyFilter` from [this bug](https://repo1.dso.mil/platform-one/big-bang/bigbang/-/issues/802)
    - Update the NetworkPolicies with real a real CIDR
    - Update to the new `bigbang.dev` cert
    - Tweak resources for a better fit on a developer laptop
    - Reorder the YAML blocks in examples/big-bang/template/bigbang/values.yaml so they are in the same order as upstream's values.yaml (makes it easier to compare them)
    - Add a bunch of exclusions to the OPA Gatekeeper policies
    - Update the `kubescape` binary version

### DIFF
--- a/examples/Vagrantfile
+++ b/examples/Vagrantfile
@@ -16,8 +16,8 @@ Vagrant.configure("2") do |config|
 
   config.vm.network "forwarded_port", guest: 80, host: 8080
   config.vm.network "forwarded_port", guest: 443, host: 8443
-  config.vm.network "forwarded_port", guest: 8080, host: 9080
-  config.vm.network "forwarded_port", guest: 8443, host: 9443
+  config.vm.network "forwarded_port", guest: 9080, host: 9080
+  config.vm.network "forwarded_port", guest: 9443, host: 9443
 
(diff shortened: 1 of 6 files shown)


jeff-mccoy:
would you mind squashing these commits before merge?

RothAndrew:
It's set to squash on merge, is there a particular reason you want them squashed before the merge? There's a merge commit in between, so given my level of Git skill the best I'm comfortable with is

* squashed commit
* merge commit
* squashed commit

jeff-mccoy (approved):

Issue #85, opened 2021-10-04
Simplify the Demo/Example Vagrant situation


RothAndrew:
## What/Why
- We should not have arbitrary "vagrant entrypoints" that do similar things (root and examples)
    - It's confusing
    - We should have meaningful and well defined boundaries (local dev vs demos, different kinds of demos) but there should be a unified method of getting into those environments
- ~~We should not specify other OSes if we aren't investing in making sure they are supported. Let's slim down to just Ubuntu for now, then add new OSes one at a time as we want to support them~~
    - ~~Existence implies support, which isn't true right now for several of the OSes in the root Vagrantfile~~
- Since existence implies support, we should make sure all OSes that are specified as available options work with the examples and demos without additional manual configuration or documentation
- When doing demos/examples we want to be as DRY as we can. If different examples want different vagrant configurations we should utilize a "master" vagrantfile and make small overrides vs having fully separate vagrantfiles for each example
- We want to be able to run multiple examples at a time during a demo.
- We should still be able to sync files in mounted folders back and forth for local development

Co-authored with @btlghrants

jeff-mccoy:
The multi os statement is incorrect.  I regularly test the other OS’s and our prime customer isn’t using ubuntu, it’s just handy for dev.  We should just add some e2e to cover other OS’s.  In fact the e2e runs on amazon linux 2 right now.

I have no problem consolidating vagrant files if we need to but am opposed to vagrant for dev.  It’s needless complexity to save installing golang.  Debug would be harder and it would slow performance.

RothAndrew:
Copy on the OSes. Will update.

RothAndrew:
@btlghrants thoughts on the other comment about vagrant for dev? I know that's how you really like to operate, and I'm definitely in agreement that it would be pretty valuable for getting new devs spun up quickly. If we said "here's a way you can use Vagrant to do dev of this thing if you want to, but nobody's requiring any particular method" would that make everyone happy?

btlghrants:
Seems like its a matter of what we're _willing to support_ really, so if we don't want to take that on as part of this project then so be it.  As you said, though, it's how I plan to operate personally (because I like the isolation) but we don't have to "officially" adopt it if I'm the only one... but it'd be a shame to "waste" the code I'm going to write to make it possible for myself, wouldn't it?

[...]

PR #91, opened 2021-10-06
feat: examples vagrant auto 'sudo su'

btlghrants:
Reduces an example-user's workflow by two whole shell commands.

1. Removes the need for the user to "sudo su" before starting work inside the examples VM.
2. Moves mounted-in example/demo content to an "/examples" folder (since there's no reason to work out of "/home/vagrant" anymore, right?) AND makes that the default directory the user is dropped into after a "vagrant up".

#85


btlghrants:
K.  Looks like I've got it straight now; shows verified.

RothAndrew:
LGTM, quickly testing it locally then I'll put through the approval

RothAndrew:
It's beautiful! It'll save me 2 seconds multiplied by probably a couple thousand times? Love it.

Issue #47, opened 2021-09-16
Update the main README


jeff-mccoy:
Need to cleanup the README and asciinema to use the latest version of zarf.

PR #98, opened 2021-10-07
README revamp

btlghrants:
Not done or anything, but I've got (what I think is) a not-so-clunky layout (mostly) in-place for the top-level README.

I expect to continue working my way down the page & moving content over to subREADMEs as it makes sense.  Just wanted to get something up for others to peep to make sure I wasn't taking us off in a direction we didn't want to go.

#47

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,151 +1,337 @@
-# Meet Zarf, K8s Airgap Buddy
+# Zarf - Kubernetes Airgap Buddy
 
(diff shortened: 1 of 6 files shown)


RothAndrew:
Nevermind actually. I'm gonna use admin privs to merge anyway, but do a squash so it results in a signed commit on master. I've looked through all the changes and it all looks good.

PR #104, opened 2021-10-11
Fix Big Bang example NetworkPolicy CIDRs

RothAndrew:
## What/Why

* Change the NetworkPolicy CIDRs back to `0.0.0.0/0` so that the example can work on different Linux distros and other user environments.


jeff-mccoy:
I guess we don’t block without tests yet?

PR #112, opened 2021-10-18
Pre-commit and linting

### LINKED ISSUE #74: Pre-Commit hooks
Integrating pre-commit hooks would let us validate that linting has been done, along with a host of other good checks, like automated protection from accidentally committing some secrets, automated style guide, automated security scanning of golang and terraform code, etc

RothAndrew:
## What

- Add pre-commit hooks including golang linting
- Update Contributor documentation

## Why

By using pre-commit hooks we can easily and automatically catch simple issues. They run automatically on every `git commit` and we can validate that they ran by running them in the CI pipeline and failing if they report any anomalies.

Pay particular attention to the changes in CONTRIBUTING.md so the new prerequisites are well understood.

This first PR adds the `.pre-commit-config` file including the following hooks:

- `check-added-large-files`: Fails if any files over 500kb are added. Is great for keeping repos small. If a bigger file needs to be added you can add it by passing the filename to an exceptions list.
- `check-merge-conflict`: Ensures that there are no orphaned Git merge conflict strings in the codebase. These can appear when a git merge is done incorrectly.
- `detect-aws-credentials`: Detects accidentally added AWS credentials before they make it into the Git history. Huzzah!
- `detect-private-key`: Detects private SSH keys and private certificates before they get accidentally added to the git history
- `end-of-file-fixer`: Ensures that every file in the codebase ends with a new line, which is a POSIX standard.
- `fix-byte-order-marker`: Strips BOMs which are not recommended when using UTF-8
- `trailing-whitespace`: Strips trailing whitespace (except for markdown since it has special meaning)
- `fix-smartquotes`: Automatically changes smartquotes to dumb quotes. Robots hate smartquotes.
- `go-fmt`: Automatically runs `go fmt` on the codebase
- `golangci-lint`: Automatically runs a linter on the golang codebase. This one is commented out for now but it works and can be turned on very easily once we fix the linting errors (there aren't that many of them)

In the documentation I have said that the CI pipeline requires these checks, but that won't be true at the time of merging this PR. That will come later.

Fixes #74 
Related to #75


RothAndrew:
`golangci-lint` results:

image

YrrepNoj:
This doesn't add any checks in the CI pipeline right? Someone could try to be sneaky (or accidentally) commit with the `--no-verify` flag and push non-conforming code.

PR #116, opened 2021-10-20
Add Zarf Version to CLI options

YrrepNoj:
Since the zarf binary can be passed around easily, it would be useful for people who know what version of the binary they have. `zarf version` will now display the latest tag and the specific commit sha it was built with.

This also updates the Makefile to store the version as a string within a variable.


YrrepNoj:
> > Should we update the General CLI E2E test to test this new command?
> 
> Yes I think we should add a quick test for this to the E2E. Also curious if we want to use this version info at all, like embedded into the packaging on package create. This would make it easy later on to notify on version mismatches or even auto-embedding the zarf binary into a package for updates.

Just to make sure I'm reading this correct. I don't think you're saying "Is this really the version number we want to use" but I'm not entirely sure what you're saying.

Are you suggesting we also add the zarf version into a yaml that gets packaged into the `SOME_PACKAGE.tar.zst`?

PR #139, opened 2021-10-28
Add flag for setting log levels and log errors for debugging.

YrrepNoj:
Added a global flag (works for any subcommand) for setting the log level of logrus. This gets applied to a local utility logger that all the other packages will use.  Having the local utility logger buys us the ability to set it once and forget about it instead of having to set the log level for every package. This also means we can have universal styling/formatting across all of our packages if that's something we care about in the future.

This PR also logs all errors to `Debug` to make it easy to figure out what's happening when something happens to go wrong. I chose `Debug` mode since we probably don't want that filling the console all the time, but at the same time errors should be rare and maybe we want to know exactly what happened whenever it happens so I can also see an argument for using `Error` mode instead..

### DIFF
--- a/cli/cmd/prepare.go
+++ b/cli/cmd/prepare.go
@@ -44,6 +44,7 @@ var prepareTransformGitLinks = &cobra.Command{
 			// Overwrite the file
 			err = ioutil.WriteFile(fileName, []byte(processedText), 0640)
 			if err != nil {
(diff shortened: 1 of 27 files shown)


jeff-mccoy:
random ?, does logrus.Debug give us any sort of stack trace as well?

jeff-mccoy (approved):

PR #143, opened 2021-10-29
Add ability to deploy remote packages

YrrepNoj:
Adding the ability for Zarf to deploy remote packages. This checks if the deployPath given is a URL with a file extension we recognize. If the path is a valid package it gets downloaded to /temp and processed.

This also renames `packageName` to `packagePath` to try to show that the variable is more than just a name and could either be a full path to a file or even a URL to some online resource.

**Potential Issues To Add After Merge:**
- [ ] Add more verification checks on the file being downloaded (in addition to the file extension and sha256 check we're currently doing?)
- [ ] Add credentials to access the url provided, currently we are only able to download packages that are public.
- [ ] Check remote filetype instead of simply checking the file extension.

### DIFF
--- a/cli/cmd/package.go
+++ b/cli/cmd/package.go
@@ -28,7 +28,7 @@ var packageCreateCmd = &cobra.Command{
 
 var packageDeployCmd = &cobra.Command{
 	Use:   "deploy PACKAGE",
-	Short: "deploys an update package file (runs offline)",
+	Short: "Deploys an update package from a local file or URL (runs offline)",
 	Args:  cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		packageName := choosePackage(args)
--- a/cli/cmd/package.go
+++ b/cli/cmd/package.go
@@ -12,6 +12,8 @@ import (
 var confirmCreate bool
 var confirmDeploy bool
 var deployComponents string
+var insecureDeploy bool
+var shasum string
 
 var packageCmd = &cobra.Command{
 	Use:   "package",
@@ -32,7 +34,8 @@ var packageDeployCmd = &cobra.Command{
 	Args:  cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		packageName := choosePackage(args)
-		packager.Deploy(packageName, confirmDeploy, deployComponents)
+		localPackagePath := packager.HandleIfURL(packageName, shasum, insecureDeploy)
+		packager.Deploy(localPackagePath, confirmDeploy, deployComponents)
 	},
 }
 
@@ -71,4 +74,6 @@ func init() {
 	packageCreateCmd.Flags().BoolVar(&confirmCreate, "confirm", false, "Confirm package creation without prompting")
 	packageDeployCmd.Flags().BoolVar(&confirmDeploy, "confirm", false, "Confirm package deployment without prompting")
 	packageDeployCmd.Flags().StringVar(&deployComponents, "components", "", "Comma-separated list of components to install.  Adding this flag will skip the init prompts for which components to install")
+	packageDeployCmd.Flags().BoolVar(&insecureDeploy, "insecure", false, "Skip shasum validation of remote package. Required if deploying a remote package and `--shasum` is not provided")
+	packageDeployCmd.Flags().StringVar(&shasum, "shasum", "", "Shasum of the package to deploy. Required if deployinga remote package and `--insecure` is not provided")
 }
--- a/cli/cmd/package.go
+++ b/cli/cmd/package.go
@@ -41,7 +41,7 @@ var packageDeployCmd = &cobra.Command{
 
 var packageInspectCmd = &cobra.Command{
 	Use:   "inspect PACKAGE",
-	Short: "lists the paylod of an update package file (runs offline)",
+	Short: "lists the payload of an update package file (runs offline)",
 	Args:  cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		packageName := choosePackage(args)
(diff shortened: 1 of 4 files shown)


cli/internal/packager/deploy.go
```diff
@@ -176,3 +210,17 @@ func deployComponents(tempPath componentPaths, assets config.ZarfComponent) {
 		git.PushAllDirectories(tempPath.repos)
 	}
 }
+
+func isValidFileExtension(filename string) bool {
+	// TODO @JPERRY What other extensions do we want to allow?
+	validExtensions := []string{".tar.zst", ".tar", "zip"}
```
YrrepNoj, line 216: `zip` -> `.zip`

jeff-mccoy:
The PR looks good, though we might want to breakup that Deploy() function a little, it's getting heft.  However I do have concerns with this idea overall until we have more SBOM components.  If we aren't validating a signature or even SHA hash then we have no way to vet the quality of this content or that it should be trusted.  I'm not certain what the right answer is, but I think we should discuss that concern a little before merging.  CC @RothAndrew @mikhailswift

RothAndrew:
> The PR looks good, though we might want to breakup that Deploy() function a little, it's getting heft. However I do have concerns with this idea overall until we have more SBOM components. If we aren't validating a signature or even SHA hash then we have no way to vet the quality of this content or that it should be trusted. I'm not certain what the right answer is, but I think we should discuss that concern a little before merging. CC @RothAndrew @mikhailswift

I do find value in this from an ease-of-use perspective, though I definitely do see the security implications.

We could potentially require the user to add a `--shasum` flag, or add an `--insecure` flag if they don't want to. That would at least make them internalize that this is not a secure thing to do.

Example:

```bash
# The right way
zarf package deploy https://example.com/some-package.tar.zst --shasum aec070645fe53ee3b3763059376134f058cc337247c978add178b6ccdfb0019f

# The insecure way
zarf package deploy https://example.com/some-package.tar.zst --insecure
```

We could also require `--insecure` if they don't specify `https://`

jeff-mccoy:
I like that format, think it should be part of this PR before merge @YrrepNoj

cli/cmd/package.go
```diff
@@ -71,4 +74,6 @@ func init() {
 	packageCreateCmd.Flags().BoolVar(&confirmCreate, "confirm", false, "Confirm package creation without prompting")
 	packageDeployCmd.Flags().BoolVar(&confirmDeploy, "confirm", false, "Confirm package deployment without prompting")
 	packageDeployCmd.Flags().StringVar(&deployComponents, "components", "", "Comma-separated list of components to install.  Adding this flag will skip the init prompts for which components to install")
+	packageDeployCmd.Flags().BoolVar(&insecureDeploy, "insecure", false, "Skip shasum validation of remote package. Required if deploying a remote package and `--shasum` is not provided")
+	packageDeployCmd.Flags().StringVar(&shasum, "shasum", "", "Shasum of the package to deploy. Required if deployinga remote package and `--insecure` is not provided")
```
jeff-mccoy, line 78: small typo "deployinga"

jeff-mccoy, line 78: small typo "deployinga"

cli/internal/packager/deploy.go
```diff
@@ -17,19 +22,21 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func Deploy(packageName string, confirm bool, componentRequest string) {
+var validExtensions = []string{".tar.zst", ".tar", ".zip"}
```
jeff-mccoy, line 25: things like this I prefer to makes constants at the top of the file or move to the config package.

cli/internal/packager/deploy.go
```diff
@@ -178,3 +185,69 @@ func deployComponents(tempPath componentPaths, assets config.ZarfComponent) {
 		git.PushAllDirectories(tempPath.repos)
 	}
 }
+
+// If privded package is a URL download it to a temp directory
```
jeff-mccoy, line 189: "privded" :-)

cli/internal/packager/deploy.go
```diff
@@ -178,3 +185,69 @@ func deployComponents(tempPath componentPaths, assets config.ZarfComponent) {
...
+		return packagePath
+	}
+
+	if !insecureDeploy && shasum == "" {
+		logrus.Fatal("When deploying a remote package you must provide either a `--shasum` or the `--insecure` flag. Neither were provided.")
+	}
+
+	// Check the extension on the package is what we expect
+	if !isValidFileExtension(providedURL.Path) {
+		logrus.Fatalf("The URL provided (%s) had an unrecognized file extension.\n", providedURL.String())
```
jeff-mccoy, line 204: maybe instead of "unrecognized extension" we could say something like, "only xx extensions are permitted".  Later on we may want to consider not using file extensions and instead inspect file type as that's more reliable.  Something like https://github.com/h2non/filetype might work.

test/e2e/e2e_general_cli_test.go
```diff
@@ -91,4 +92,7 @@ func testGeneralCliStuff(t *testing.T, terraformOptions *terraform.Options, keyP
 	require.Error(t, err, output)
 	output, err = ssh.CheckSshCommandE(t, publicHost, fmt.Sprintf("cd /home/%s/build && ./zarf pki regenerate --host some_unique_server", username))
 	require.Error(t, err, output)
+
+	// Test that `zarf package deploy` doesn't die when given a URL
+	output, err = ssh.CheckSshCommandE(t, publicHost, fmt.Sprintf("sudo bash -c 'cd /home/%s/build && ./zarf package deploy https://zarf-examples.s3.amazonaws.com/zarf-package-appliance-demo-doom.tar.zst --confirm'", username))
```
jeff-mccoy, line 97: we'll probably need a couple tests around the "--insecure" and ""shasum" flags

jeff-mccoy (changes requested):

cli/internal/packager/deploy.go
YrrepNoj: Spelling is one of my biggest weaknesses.. I just downloaded a spell checking extension for VSCode :P

cli/internal/packager/deploy.go
YrrepNoj: I'll move it to the config package, you're right that it makes more sense there.

Golang doesn't allow making slices/arrays constants because arrays are essentially pointers to their elements, the best we can do is make the array a fixed size. 
https://stackoverflow.com/a/13137568

 Given your other comment on checking filetype in the future

test/e2e/e2e_general_cli_test.go
YrrepNoj: It would also be useful if I checked the output and err results.. whoops

cli/internal/packager/deploy.go
YrrepNoj: 👍 I'll add that to the list of potential new tasks. I really like that idea.

jeff-mccoy (approved):
lgtm.  Only comment on the slice constant problem.  We _could_ make it private and use a getter with a by value return, I _think_ that would get after the immutability quality of a constant.

cli/internal/packager/deploy.go
YrrepNoj: I added a initialization function to give us the immutability similarity as described here https://qvault.io/golang/golang-constant-maps-slices/

jeff-mccoy (approved):

Issue #132, opened 2021-10-28
Main README: Coming Soon - Roll Your Own


btlghrants:
As a learner of Zarf
I need to get comfortable with making my own Zarf packages
Because it's how I'll use Zarf "in the field"

---

1. Write new docs/your-own-package.md page walking through creation & deployment of a custom package
2. Update "Roll Your Own - Coming Soon" link in main README to point to new page

---

Spun off from #47

Daemonslayer2048:
I am new to zarf and have deployed the demo, I am now moving on to trying to roll my own package. As this issue is still open is it possible someone may be willing to share notes? 
Would the documentation also include how to handle secrets and parameters?

RothAndrew:
The best way right now IMO is to use the examples that are in the examples folder. They mostly cover the various features and are constantly updated. Since Zarf is changing so fast right now it has been hard to keep up with the docs.

RothAndrew:
Feel free to ask if you have questions, either here in the issues or in the discussions section.

RothAndrew:
Blocked by #237

jeff-mccoy:
Going to re-evaluate these stale doc issues, closing for now.Going to re-evaluate these stale doc issues, closing for now.

PR #144, opened 2021-10-29
Add authentication to utility registry

### LINKED ISSUE #5: Make utility registry require credentials
At the very least, the utility registry should require credentials to push images.  Ideally these creds would be dynamically handled similarly to [#3](https://repo1.dso.mil/platform-one/big-bang/apps/product-tools/zarf/-/issues/3).

---

Migrated from https://repo1.dso.mil/platform-one/big-bang/apps/product-tools/zarf/-/issues/9

RothAndrew:
Fixes #5


RothAndrew:
@jeff-mccoy what's the best way to get Crane to log in now that the registry requires auth?

I've been trying a few things out but nothing feels right.

Would love to do some pairing on this if you have the time.

jeff-mccoy:
Can you clarify the question?  Under the hood crane uses the docker login info, with zarf you could do a login and pull without docker installed and it would work.

RothAndrew:
I looked at running Crane's `login()` function [HERE](https://github.com/google/go-containerregistry/blob/main/cmd/crane/cmd/auth.go), but it isn't exported.

Is there a way to internally call a Cobra command from within the golang codebase?

RothAndrew:
@jeff-mccoy The podinfo example's helm chart does not support `imagePullSecrets`. I'm proposing we just delete the podinfo example, since it doesn't do anything new that the doom game example doesn't already do.

Alternatively, we'll need to either find a different helm chart or submit a PR to add `imagePullSecrets` support to Podinfo's helm chart and fork the chart in the meantime.

Thoughts?

jeff-mccoy:
agreed, let's just drop it

RothAndrew:
@jeff-mccoy ready for review on this. Went through and tested all of the examples, they all look good.

jeff-mccoy:
sorry still looking at this, just really hate how much K8s sucks for image things, including private registries.  The mirror config with alternate domains will suck to maintain as well as all the image pull secrets will be a pain.  The more I think about this, the more it might make sense to leverage the specified hostname/ip for the registry or use internal routes for appliance mode.  This would allow us to not use mirror config at all (very important for when we move out of just k3s) and only have one imagepullsecret config that would could potentially just inject as well as a part of the deployment process without forcing the users to maintain that too.  Thoughts?

RothAndrew:
The more intuitive (and less "magical") we can make it the better. My worry would be that we make it harder to understand as we make it a bit less verbose. It made total sense to me that I'd have to add new secrets to each namespace that got created for each example, and I was totally fine with it.

If we got rid of image mirroring we'd want to add a `zarf prepare patch-k8s` just like we have a `patch-git`, but that's definitely not the end of the world, and it's something we'd need to handle anyway if we were pointing at a separate Zarf utility cluster.

RothAndrew:
We're feeling pain because we have a bunch of these examples to deal with, but an end user won't experience as much churn as we are. IMO this way is fine, as long as we are fine with telling the user they have to do ImagePullSecrets when creating their zarf packages because the registry requires auth.

RothAndrew:
Which I guess brings up a deficiency with this PR, it doesn't update the docs at all. I'd better fix that before @btlghrants notices and give me a whoopin

jeff-mccoy:
With OCI Distribution you shouldn't need to change anything but the hostname of the image, there wouldn't be any funny transformation stuff happening as the registry would just respect the path anyway. We leverage that concept already when push to the registry.  My concern is around how this takes some of the simplicity away just because K8s is dumb about private registries.  We don't have to go along with it, just like for TLS we don't have people run a bunch of openssl commands to get a valid cert, we do the work for them.

jeff-mccoy (changes requested):
Update docs for this change and then g2g.

jeff-mccoy:
Gone back and forth on this (and went on a random code tangent to investigate), I do think we can do more later on down the road.  But the most important thing right now is that we're not adding direct complexity to the deployer of a zarf package--so I think besides docs this all looks good and satisfies an important missing security piece.  Thanks for finally getting this done, I put it off _way_ too long.

cli/internal/utils/auth.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/docker/cli/cli/config"
+	"github.com/docker/cli/cli/config/types"
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"log"
+	"os"
+)
+// Login adds the given creds to the user's Docker config, usually located at $HOME/.docker/config.yaml. It does not try
+// to connect to the given registry, it just simply adds another entry to the config file.
+// This function was mostly adapted from https://github.com/google/go-containerregistry/blob/main/cmd/crane/cmd/auth.go
```
YrrepNoj, line 13: If we link to an external repo it might be helpful to put a release/commit SHA. That project seems to get updated regularly and might be useful in the future to know exactly what the inspiration was.

RothAndrew:
Actually I don't think there is anything to add yet documentation-wise. The changes would happen under "Roll Your Own" #132 

@btlghrants do you agree?

jeff-mccoy (dismissed):

RothAndrew:
@jeff-mccoy one more approval please so it's ready to merge if @btlghrants agrees with my assessment that there aren't any docs to update right now.

btlghrants:
Yeah, as far as I know, the game example is the only one we've done any sort of explicit user documentation for and _that_ has the imagePullSecrets added as part of this PR.

Seems fine to merge to me & I think Andy's right--it'll change how we write "Roll Your Own" #132 but we haven't gotten there yet.

jeff-mccoy (approved):

PR #155, opened 2021-11-04
Git Repository Overhaul

### LINKED ISSUE #154: Make Zarf Deployment Push All Git Repository Branches
Currently, when using Zarf to deploy a git repository only the master branch and tags are pushed to the Gitea mirror. It would be ideal if this mirror also contained the other branches that the original git repository contained.

My approach to this would be as follows:
When provided with a Git repository on package deployment, utilize the git refs in `.git/refs/remotes` to populate the refs in Gitea. This directory, unlike `.git/refs/heads` is updated on `git fetch` rather than a `git checkout`, which Zarf runs after cloning the repository, implying that the git refs are already there and just need to be referenced.

JustinFirsching:
Update the pull function so that tag-provided mirrors do not fetch all
tags, instead later fetching (externally) only the tag they need.

Implement tag fetching function to retrieve _only_ the desired tag when
creating a tag-provided repo mirror.

Checkout the tag into a detatched HEAD state at the end of the create
stage of tag-provided repository mirrors.

Implement ref removal and addition functions used during package
creation and mirror push to ensure that refs that aren't wanted on the
mirror won't be pushed.

Update the refspecs used in the push to push detatched HEAD, branches,
online remote, and tags to the offline mirror. Note that if we later
checkout a branch from the remote and do not clean up the remote ref it
will lead to a duplicate ref name and the push will fail on one of the
refs (likely the online one since it is later in the refspec slice).

Fixes #154 

feat: Allow for repos to be provided without a tag to mirror all branches/tags
feat: Make tag-provided repository mirrors use the tag as master
fix: Prevent tag-provided repo mirrors from storing extra refs

### DIFF
--- a/cli/internal/git/pull.go
+++ b/cli/internal/git/pull.go
@@ -5,20 +5,27 @@ import (
 
 	"github.com/defenseunicorns/zarf/cli/internal/utils"
 	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/sirupsen/logrus"
+
+	"strings"
 )
 
 const onlineRemoteName = "online-upstream"
 
 func DownloadRepoToTemp(gitUrl string) string {
 	path := utils.MakeTempDir()
+	// If downloading to temp, grab all tags since the repo isn't being
+	// packaged anyways and it saves us from having to fetch the tags
+	// later if we need them
 	pull(gitUrl, path)
 	return path
 }
 
-func Pull(gitUrl string, targetFolder string) {
+func Pull(gitUrl string, targetFolder string) string {
 	path := targetFolder + "/" + transformURLtoRepoName(gitUrl)
 	pull(gitUrl, path)
+	return path
 }
 
 func pull(gitUrl string, targetFolder string) {
@@ -29,10 +36,16 @@ func pull(gitUrl string, targetFolder string) {
 
 	gitCred := FindAuthForHost(gitUrl)
 
+	matches := strings.Split(gitUrl, "@")
+	fetchAllTags := len(matches) == 1
 	cloneOptions := &git.CloneOptions{
-		URL:        gitUrl,
+		URL:        matches[0],
 		Progress:   os.Stdout,
 		RemoteName: onlineRemoteName,
+	}
+
+	if !fetchAllTags {
+		cloneOptions.Tags = git.NoTags
 	}
 
 	// Gracefully handle no git creds on the system (like our CI/CD)
@@ -41,13 +54,39 @@ func pull(gitUrl string, targetFolder string) {
 	}
 
 	// Clone the given repo
-	_, err := git.PlainClone(targetFolder, false, cloneOptions)
+	repo, err := git.PlainClone(targetFolder, false, cloneOptions)
 
 	if err == git.ErrRepositoryAlreadyExists {
 		logContext.Info("Repo already cloned")
 	} else if err != nil {
 		logContext.Debug(err)
 		logContext.Fatal("Not a valid git repo or unable to clone")
+	}
+
+	if !fetchAllTags {
+		tag := matches[1]
+
+		// Identify the remote trunk branch name
+		trunkBranchName := plumbing.NewBranchReferenceName("master")
+		head, err := repo.Head()
+
+		if err != nil {
+			// No repo head available
+			logContext.Debug(err)
+			logContext.Warn("Failed to identify repo head. Tag will be pushed to 'master'.")
+		} else if head.Name().IsBranch() {
+			// Valid repo head and it is a branch
+			trunkBranchName = head.Name()
+		} else {
+			// Valid repo head but not a branch
+			logContext.Warn("No branch found for this repo head. Tag will be pushed to 'master'.")
+		}
+
+		RemoveLocalBranchRefs(targetFolder)
+		RemoveOnlineRemoteRefs(targetFolder)
+
+		FetchTag(targetFolder, tag)
+		CheckoutTagAsBranch(targetFolder, tag, trunkBranchName)
 	}
 
 	logContext.Info("Git repo synced")
--- a/cli/internal/git/utils.go
+++ b/cli/internal/git/utils.go
@@ -4,11 +4,14 @@ import (
 	"bufio"
 	"net/url"
 	"os"
+	"path"
 	"regexp"
 	"strings"
 
 	"github.com/defenseunicorns/zarf/cli/config"
 	"github.com/defenseunicorns/zarf/cli/internal/utils"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/transport/http"
 	"github.com/sirupsen/logrus"
 )
@@ -43,6 +46,11 @@ func transformURL(baseUrl string, url string) string {
 		"New": output,
 	}).Info("Transformed Git URL")
 	return output
+}
+
+func transformRepoDirToURL(baseUrl string, repoDir string) string {
+	baseDir := path.Base(repoDir)
+	return baseUrl + "/zarf-git-user/" + baseDir
 }
 
 func credentialFilePath() string {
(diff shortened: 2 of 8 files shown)


RothAndrew:
I'll let @jeff-mccoy be the final say here but I don't think this is something we want to do. We want the declarative state to be a specific git reference. Its an outstanding bug that `zarf package create` actually clones more than is intended.

JustinFirsching:
Responded in the issue as well, but to make sure it is seen here too:

Would it be fair to say that if there is no tag provided everything is brought over, but if a tag is provided only that tag is brought over? If so, I could likely get the rest of this implemented this week.

jeff-mccoy:
I commented on the issue, open to discussing this as an opt-in, but the reason for tag-based tracking is we will be using that to do diffs in the future without having to fetch each remote asset and see if it changed.  In git tags can’t be moved, they can only be removed/ added (ie forced) but that’s not the standard convention so it’s a good way to track changes quickly.  Long term idea would be a new command to compare a current zarf.yaml to an older one to see what resources have changed and only include those.

jeff-mccoy:
@YrrepNoj please review for code flow.  @RothAndrew please review for overall user flow and how this impacts current examples such as the Big Bang example.

jeff-mccoy:
I’ll do a final review after both of you.

RothAndrew:
@JustinFirsching pls pull in latest master changes

JustinFirsching:
@RothAndrew Sorry about that. Rebased and pushed.

RothAndrew:
No worries, just wanted to make sure the stuff added in 0.13 got included, namely the change to the private registry to require auth.

cli/internal/packager/create.go
```diff
@@ -119,10 +119,16 @@ func addLocalAssets(tempPath componentPaths, assets config.ZarfComponent) {
 		// Load all specified git repos
 		for _, url := range assets.Repos {
 			matches := strings.Split(url, "@")
-			if len(matches) < 2 {
-				logrus.WithField("remote", url).Fatal("Unable to parse git url. Ensure you use the format url.git@tag")
+			// Pull all of the references if there is no `@` in the string
+			pullAll := len(matches) == 1
+			cloneDir := git.Pull(url, tempPath.repos, pullAll)
+			if !pullAll {
```
YrrepNoj, line 125: I'm fine with it either way but I think this might make a lot of sense to put at the end of the `git.Pull()` call.

YrrepNoj (approved):

RothAndrew (changes requested):
- When deploying the `gitops-service-data` example based on the zarf.yaml I expect to see 3 repos, but I only see 1.
image

- Recommend having the `gitops-service-data` example pull in a helm chart that gets deployed. I see the Twistlock chart getting pulled in in the zarf.yaml but I don't really have any good way to verify that it is there
- examples/gitops-data/README.md should be updated to talk about the different behaviors being demoed (single tag & full repo)
- IMO we should add an E2E test for the `gitops-service-data` example but I won't hold up the PR for it if everyone else just wants to keep pushing forward

cli/internal/packager/create.go
YrrepNoj, line 125: I'm with with it either way but I think this logic would make sense to put at the end of the `git.Pull( ... )` function.

JustinFirsching: Actually, the first implementation I had for this did look like that haha. I ended up switching to this because I wasn't sure of the scope of git pull usage for Zarf. I figured that it would be easier to implement it with no limitations (such as deleting the refs) so that if a pull were ever needed for a different reason, that didn't fetch all tags, a refactor wasn't needed. Hopefully that at least clears up my thought process for not including it directly in the pull. If `git.Pull` will only ever be needed for charts pulls and repo mirrors, I see no reason not include the tag checkout/reference deletion logic at the end, and if you'd like, I can make that happen while handling the README update.

JustinFirsching:
@RothAndrew 

- Is it possible you are using an old copy of the Zarf package? I am getting all three repositories as shown in the screenshot attached. Running `tar -I unzstd -tf zarf-package-gitops-service-data.tar.zst | grep 'mirror__[^/]*/.git/$'` should list all of the repos in the package, and I expect the tar you have will only have the one repository.
image
- Would I just need to deploy the chart using a Kubernetes manifest? If so, I think that would look like joining the current (master) version of `examples/gitops-data` with `examples/single-big-bang-package`, which I can do, but want to verify before creating a duplicate code chunk.
- As for the README.md, I won't have a chance to get around to that tonight, but can definitely get that updated to explain the new features. 
- Assuming all that needs to be done for the E2E test is updating the `e2e.sh` script to include a few calls to the Gitea API confirming that a gitops-data deployment pushed the repositories as expected, I can toss that in with the README.md updates.

RothAndrew:
Its possible. Ill check again im the morning

RothAndrew:
e2e.sh is the old way that we are moving away from. The new way is golang tests with the Terratest library, in the /test directory. If you feel comfortable taking a stab at it thats great. If not that's okay too, we can make an issue to follow up with a new test

RothAndrew:
> Is it possible you are using an old copy of the Zarf package? I am getting all three repositories as shown in the screenshot attached. Running tar -I unzstd -tf zarf-package-gitops-service-data.tar.zst | grep 'mirror__[^/]*/.git/$' should list all of the repos in the package, and I expect the tar you have will only have the one repository.

Yep, I had `master` checked out rather than the branch you're working in. In your branch I see all 3 repos, and everything looks good inside them. The Zarf repo only has master (at the point in time when 0.12.0 was cut) and the 0.12.0 tag, and podinfo and twistlock have everything 👍 

> Would I just need to deploy the chart using a Kubernetes manifest? If so, I think that would look like joining the current (master) version of examples/gitops-data with examples/single-big-bang-package, which I can do, but want to verify before creating a duplicate code chunk.

IMO I'd actually just take out the chart from the `gitops-data` example. That functionality is already demoed in a couple of other places, including the `single-big-bang-package` example.

> As for the README.md, I won't have a chance to get around to that tonight, but can definitely get that updated to explain the new features.

👍 

> Assuming all that needs to be done for the E2E test is updating the e2e.sh script to include a few calls to the Gitea API confirming that a gitops-data deployment pushed the repositories as expected, I can toss that in with the README.md updates.

See above comment from last night

JustinFirsching:
Updated the README as a new commit to make it clear that nothing else was changed. Let me know what still needs to be changed and I can rebase from there. 

I haven't looked into the E2E tests much, but I don't have an AWS account, so if I understand correctly that may rule me out entirely.

RothAndrew:
> Updated the README as a new commit to make it clear that nothing else was changed. Let me know what still needs to be changed and I can rebase from there.
> 
> I haven't looked into the E2E tests much, but I don't have an AWS account, so if I understand correctly that may rule me out entirely.

The tests would run in our account, they'd just need to be written and wired up, but I don't necessarily expect everyone to do that, at least until we write a guide on how to do it.

This can be merged without changes to the E2E test suite with an issue to have one of the core devs to do it.

examples/gitops-data/README.md
```diff
@@ -1,8 +1,83 @@
...
+
+### Tag-Provided Git Repository Clone
+
+Tag-provided git repository cloning is the recommended way of cloning a git
+repository for air-gapped deployment. Tag-provided clones are provided using
+the format `url.git@tag` as seen in the example with the `defenseunicorns/zarf`
+repository (`https://github.com/defenseunicorns/zarf.git@v0.12.0`).
+
+A tag-provided clone only mirrors the provided tag of the repository,
+displaying on the Gitea mirror as both the `master` branch and the tag itself.
```
RothAndrew, line 25: If the trunk branch of my repo is named `main` would it still show up as `master` in Gitea?

JustinFirsching: It would have, recent changes correct this behavior.

cli/internal/git/pull.go
```diff
@@ -29,19 +36,26 @@ func pull(gitUrl string, targetFolder string) {
 
 	gitCred := FindAuthForHost(gitUrl)
 
+	// This is here for backwards compatibility
```
jeff-mccoy, line 39: Can you clarify this comment?  I think I understand what you are saying in that we now support with and with the tag, but to be clear the default and normal behavior we will document/encourage will use the tag as that will be very important for upcoming diff engine work

cli/internal/git/pull.go
```diff
@@ -50,5 +64,31 @@ func pull(gitUrl string, targetFolder string) {
 		logContext.Fatal("Not a valid git repo or unable to clone")
 	}
 
+	if !fetchAllTags {
+		tag := matches[1]
+
+		// Identify the remote trunk branch name
+		trunkBranchName := plumbing.NewBranchReferenceName("master")
```
jeff-mccoy, line 71: I redid this a little, I've had issue in golang before doing `err == nil` and try to always use `err != nil` because of past issues, https://yourbasic.org/golang/gotcha-why-nil-error-not-equal-nil/.  The nested conditionals were a little hard to follow.  Most of this is subjective, but I always strive to reduce cognitive load when possible.  Feel free to change some more if I messed this up.

cli/internal/git/pull.go
JustinFirsching: Good catch! That comment was placed when I did the first commit to switch to full mirrors and although I have changed the variable below it since, I missed deleting the comment. This comment needs to be deleted. I can push a deletion commit now, or include one when I force push a squash. I've just been keeping everything as separate commits lately so reviewing changes such as tag branch name targets could be done in steps, which I assume is how it has been missed until now.

cli/internal/git/pull.go
JustinFirsching: That looks good to me. I wasn't a huge fan of the nested conditional either, was just trying to avoid this:
```go
if err != nil {
    logContext.Debug(err)
    logContext.Warn(message)
} else if !head.Name().IsBranch() {
    logContext.Warn(message)
} else {
    trunkBranchName = head.Name()
}
```

since the `if` and `else if` warning log are the same message. With how you have the messages setup I like this much better.

JustinFirsching: That's very interesting with the whole `nil != nil` thing. Thanks for the share!

cli/internal/git/utils.go
```diff
@@ -151,3 +159,155 @@ func CredentialsGenerator() string {
...
+
+	head, err := repo.Head()
+	if err != nil {
+		logContext.Debug(err)
+		logContext.Fatal("Failed to identify head")
+	}
+
+	removedRefs := []*plumbing.Reference{}
+	err = references.ForEach(func(ref *plumbing.Reference) error {
+		passHead := !(ref.Name() == plumbing.HEAD || ref.Name() == head.Name())
```
jeff-mccoy, line 245: Not super clear on this, the negative logic is a little awkward, can we do `ref.Name() != plumbing.HEAD && ref.Name() != head.Name()` instead?  The name passHead isn't very clear either.  Maybe a little more comments on what is happening here?

jeff-mccoy (commented):
Overall this looks good, some minor comments on some conditionals that are a little unclear (I patched one of them already) and look to @RothAndrew's final review as well.  Think we're very close.  Thanks for the work on this, it will make git operations much more robust.

cli/internal/git/pull.go
jeff-mccoy: either way we'll just need all the commits in before final test/review, up to you, thanks.

cli/internal/git/pull.go
jeff-mccoy: Yeah I had to play with the logic a little to get it how I thought it might read better.  I also did different warns since technically they are different conditions occurring between the err and not being a branch.

cli/internal/git/utils.go
JustinFirsching: Sure! 

To explain, I called it `passHead` because it determines if we pass the HEAD check **for deletion**, which is `and`ed with if we pass the user check for deletion. I can find a better name for it though since `passHead` certainly isn't very descriptive on its own. As for the awkward logic, I wrote it that way to reflect "what would a HEAD check look like", and then "now invert that since we want to pass the HEAD check when the ref is _not_ HEAD or what HEAD points to". Refactoring to "what implies that the ref isn't HEAD or what HEAD points to" makes sense to me though. I'll add comments when I do so as well.

JustinFirsching:
Thanks! I will get the requested changes up as soon as possible.

Looking forward to the final review. I assume it would be easiest to review as separate commits since @RothAndrew has been tracking along the way, but if that is not the case, let me know and I can force push with the commits squashed.

This was a much more interesting and complex issue than I expected it to be when I first filed the issue and I'm glad to have been able to contribute so much of the solution.

RothAndrew:
You can go ahead and squash if you want. Im off tomorrow but I'll take a look if I have some down time. I just had a couple of nitpicks on the documentation that may have been resolved now. @jeff-mccoy if you feel good about it thats good enough for me

jeff-mccoy:
I’d say press with the squash and run a “/test all” @JustinFirsching

jeff-mccoy:
@JustinFirsching once the review comments I made a couple days ago are addressed I think we can proceed to squash/merge this PR.

JustinFirsching:
Sorry for the delay on this! Made the changes and got sidetracked before I had a chance to rebase and push. I don't think I have permissions to run tests, so someone else may need to kick those off for me, but I think this is what we are looking for. Happy to resolve any remaining issues with the changes and/or documentation associated with it!

examples/gitops-data/README.md
```diff
@@ -1,8 +1,83 @@
...
+
+```sh
+zarf package deploy zarf-package-gitops-service-data.tar.zst
+```
+
+> _**Important**_
+>
+> It's possible to try a package deploy _before the Zarf cluster is ready to receive it_. If you see an error like `"https://<cluster ip>/v2/": dial tcp ,<cluster ip>:443: connect: connection refused;` then it's very likely that you've beat the Zarf startup routines.
+>
+> The fix is simple: just wait for the cluster to finish starting & try again.
```
jeff-mccoy, line 74: good news with the native apply changes this wont be an issue anymore

jeff-mccoy (approved):
everything looks good, thanks for all the work on this!

JustinFirsching:
Glad to contribute! Thank you all for the feedback and being so responsive on the PR and issue thread!

PR #156, opened 2021-11-04
Remove unnecessary logging package and set log level in root

YrrepNoj:
This removes the logging package that I meant to remove when we realized we didn't need to wrap the Logrus object. 

@RothAndrew do you have any ideas on how we can create a test to verify the CLI was able to interpret and set the correct logging level?


jeff-mccoy:
We could just add a message acknowledging when the user changes the log level and test for that.  It would be better UX anyway.

jeff-mccoy (dismissed):
LGTM, would just suggest consider adding some acknowledgment to the user that they changed the log level, would make testing cleaner too.

jeff-mccoy (dismissed):

jeff-mccoy:
@YrrepNoj note the merge conflict currently

jeff-mccoy (approved):

PR #172, opened 2021-11-10
docs: what is zarf?

btlghrants:
Added a new file: docs/what-is-zarf.md.

Makes an attempt to:

1) distill all the "What does Zarf really do?" conversations I've been having (with new devs & outside users) into something we can have others read,

2) visualize Zarf **in a software delivery context** to show just how it helps move "software across the air gap",

3) give us an artifact to talk to & update as we have discussions about what is/isn't, should/shouldn't be in-scope for the project.

Some of this will need expansion / fix-up but I thought it was in a good enough shape to fill some documentation holes we've had for a while.

Happy to have further conversations about what should / shouldn't be here, if we want to chunk some of this out into separate pages, etc.


jeff-mccoy:
haven't got to go through all this yet, but it's **_really_** good so far

btlghrants:
All right @YrrepNoj , give it another look!

In particular, these sections:
- https://github.com/defenseunicorns/zarf/blob/docs/what-is-zarf/docs/what-is-zarf.md#how-it-helps
- https://github.com/defenseunicorns/zarf/blob/docs/what-is-zarf/docs/what-is-zarf.md#3---zarf-publishes-your-package

Do you think they make it any clearer to a new Zarfer that, 1) the project comes in 2 parts, and 2) that one of them is long-running (the Zarf cluster) while the other isn't (the binary)..?

docs/what-is-zarf.md
```diff
@@ -0,0 +1,225 @@
...
+
+## How it helps
+
+Zarf supercharges the way you deliver complex, k8s-native applications to remote systems by reducing update preparation & installation to a few simple terminal commands. This support comes in two parts:
+
+A precompiled `zarf` binary, which:
+
+- rolls numerous software types / updates into a single distributable package (while on an internet-accessible network), and
+
+- unrolls package contents "into place" for use on production systems (while on an internet-isolated network).
```
YrrepNoj, line 21: If we really want to sell how much Zarf `supercharges` deployments I think we could add a bit of context of how much easier the deployment of packages onto production systems is compared to the other ways people typically deploy.

docs/what-is-zarf.md
```diff
@@ -0,0 +1,225 @@
...
+
+In the first scenario, your package consists of a single application (plus dependencies) and you configure the Zarf cluster to _serve your application directly to end users_. This mode of operation is called "Appliance Mode"&mdash; because it's small & self-contained like a kitchen appliance&mdash;and it is intended for use in environments where you want to run k8s-native tooling but need to keep a small footprint (i.e. single-purpose / constrained / "edge" environments).
+
+Find out more about what direct user service looks like in the "[Appliance Mode](#appliance-mode)" section.
+
+In the second scenario, your package consists of updates for many apps / systems and you configure the Zarf cluster to _propagate updates to downstream systems_ rather than to serve users directly. This mode of operation is called "Utility Cluster Mode"&mdash;as it's main job is to _add utility_ to other clusters&mdash;and it is intended for use in places where you want to run independent, full-service production environments (your own [Big Bang](https://github.com/DoD-Platform-One/big-bang) cluster, perhaps?) but you need help tracking, caching & disseminating system / dependency updates.
```
YrrepNoj, line 64: Having the words `configure the Zarf cluster` and `is called "Utility Cluster Mode"` somewhat clash for me since it wouldn't be unreasonable for a new Zarfer to think there are now two clusters that just got spun up. 

What do you think of making it read more like `In the second scenario, called "Utility Cluster Mode", ........... you configure the Utility cluster to ....`?

YrrepNoj: Is `Zarf cluster` a term we want to be a thing?

docs/what-is-zarf.md
btlghrants: > Zarf supercharges the way you deliver complex, k8s-native applications to remote systems by reducing update preparation & installation to a few simple terminal commands.

...becomes...

> Zarf supercharges the way you deliver complex, k8s-native applications to remote systems by reducing _tens/hundreds of individual software updates, movements, and manual installations to a few simple terminal commands._

Like that?  Problem is that it's really difficult to talk much about "other ways people typically deploy" without specifying / assuming how they work.  Best I could think to do was to compare the stuff **Zarf does** against what you would have to do if you wanted to do **Zarf's work manually**.  :thinking:

docs/what-is-zarf.md
btlghrants: Mmm, I think it's useful to have "Zarf cluster" be a thing, yes.  Because, in either scenario, there _is_ always only one cluster that Zarf touches--the "Zarf(-controlled) cluster".

It's a matter of how we stuck another label on top of it depending on how it's being used: in "applicance" vs "utility cluster" mode.  I was trying to mirror the terminology I've heard in conversation but perhaps that's not helping..?

What about if:

> Once your package has landed, depending on what you've built into it, there are a couple of ways for Zarf to expose package contents.

became,

> Once your package has landed, depending on what you've built into it, there are a couple of ways for **the Zarf cluster** to expose package contents.

Does that make it any clearer that there is _a single Zarf cluster_ (independent of usage) and that the "appliance" vs "utility" terms are monikers for the usage scenario rather than the cluster itself?

btlghrants: I'd be happy to come up with some totally separate terms for the "modes" if that helps, too.  Or, maybe just dropping the word "cluster" from the "modes" altogether--as in: "Appliance Mode" and "Utility Mode".  That might make it more apparent that we expect there to always be "the Zarf cluster" and it can be used in a couple of different ways.

I've seen quite a bit of confusion from the people we've been demo'ing to thinking that Zarf was going to (somehow) stand up their entire production cluster _for them_--driven by how we're doing "appliance mode" demos, probably--whereas Zarf was always conceived of (I think?) as a "cluster to sit next to" the production cluster.

btlghrants: > it wouldn't be unreasonable for a new Zarfer to think there are now two clusters that just got spun up

Interesting. The "Zarf publishes your package" block reads like something you'd do _all_ of at the same time?  I tried to use the "scenarios" to make it clear that this was an "either-or" thing (for the purposes of this write-up, anyway) but that didn't come across, eh?

What if we make the lead-ins more like:
> In the first, simple scenario, ...

and...
> In a totally separate, more intensive scenario, ...

Does that make it clearer that these should be considered exclusively?

jeff-mccoy:
I'd like to nail down these definitions a little better and then be consistent across the entire project (there are plenty of places in the codebase that aren't consistent right now).

btlghrants:
> I'd like to nail down these definitions a little better and then be consistent across the entire project (there are plenty of places in the codebase that aren't consistent right now).

I'm game.  For now I've resolved all the open comments with the substitutions recommended in the comments.  I think it's made _this file_ self-consistent, at least.

Do we want to wait on normalizing terminology in _all the docs_ before accepting this PR?  Or is this good enough to get into master so that external peeps can start finding it?

docs/what-is-zarf.md
```diff
@@ -0,0 +1,225 @@
...
+
+- a "packaging" workstation &mdash; which must have the `zarf` [CLI tool installed](./workstation.md#just-gimmie-zarf), and
+
+- a `zarf.yaml` file &mdash; which tells the Zarf what you need packaged.
+
+Find some detailed uses of the `zarf.yaml` file in [our examples](../examples/).
+
+#### (2) - Package
+
+Actually making a Zarf package out of a `zarf.yaml` file is a matter of calling a single, simple command: `zarf package`.  You'll see a `zarf-package-*.tar.zst` file pop into existence afterward&mdash;that's your package.
```
jeff-mccoy, line 99: technically this is `zarf package deploy` to see the menu or you can just do `zarf zarf.yaml` to skip the menu

docs/what-is-zarf.md
```diff
@@ -0,0 +1,225 @@
...
+
+Move a Zarf release + your desired packages to your Zarf cluster machine.
+
+#### (1) - Zarf cluster
+
+Make your machine into a **single node** Zarf cluster with the command: `zarf init`.
+
+Recommended Zarf components: `management`.
+
+Deploy your package into the Zarf cluster with the command: `zarf deploy`.
```
jeff-mccoy, line 141: same here it's actually `zarf *zarffilename*` or `zarf package deploy` to get the menu

docs/what-is-zarf.md
```diff
@@ -0,0 +1,225 @@
...
+
+Here are the things we think Zarf should get really good at, listed top-down in order of importance:
+
+- Make movement of software "across the air gap" an open source "solved problem".
+
+- Make it trivial to deploy & run Kubernetes apps "at the Edge".
+
+- Make it easy to support GitOps-based k8s cluster updates in isolated environments.
+
+- Make it possible to support GitOps-based k8s cluster updates in internet-connected-but-independent environments (think: dependency caching per availability zone, etc).
```
jeff-mccoy, line 191: I'd add comments about saving bandwidth / increase fault tolerance / etc here.

docs/what-is-zarf.md
```diff
@@ -0,0 +1,225 @@
...
+
+## What can be packaged?
+
+To reiterate: Zarf's possibilities are endless&mdash;Zarf developers' time is not. Thus, scope definition is again in order.
+
+Given Zarf's being a "k8s cluster to serve _other_ k8s clusters", the following types of software can be rolled into a Zarf package:
+
+- container images &mdash; to serve images for the Zarf & downstream clusters to run containers from.
+
+- git repositories &mdash; to serve as the git-based "source of truth" for downstream "GitOps"ed k8s clusters to watch.
```
jeff-mccoy, line 204: now also any generic git repos needed from the online to offline side since the last big git PR merge.  Think "I'd love to have the docs for this cool tool in my airgap, can zarf bring those from github to here for me?"

docs/what-is-zarf.md
```diff
@@ -0,0 +1,225 @@
...
+
+- pre-compiled binaries &mdash; to provide the software necessary to start & support the Zarf cluster.
+
```
jeff-mccoy, line 207: Not just pre-compiled binaries, that's a particular file type zarf is aware of, but could also be infra code, ie. terraform modules, scripts, config files, etc.  Also big one that was left off is the data injection one

jeff-mccoy (commented):
Overall this is an absolutely fantastic document.  I would like to try and settle a definition for the "zarf cluster" "utility cluster" "gitops service" "appliance mode" "standard mode" as they are all partially overlapping ideas/terms I think we can really get settled now.  I'll open a new issue for that.

docs/what-is-zarf.md
btlghrants: Wait, but this para here is about _making_ the package rather than deploying it, right?  So I guess the command should be `zarf package create`.

So you can run it both ways, right?  Either directly with `zarf package create` or with `zarf zarf.yaml` will both build the package?

I kinda prefer the one with `package` in it cuz that makes the CLI API a bit more explicit, which seems fitting for a "first touch"-ish file like this.

You think it'd read more clearly with just `zarf zarf.yaml` though?

docs/what-is-zarf.md
btlghrants: Definitely forgot the "package" part in there too--good catch.  Should read `zarf package deploy` for sure.

docs/what-is-zarf.md
jeff-mccoy: yeah updated me comment.  I agree `zarf package create` is better introduction, power users will like move to `zarf zarf.yaml`

jeff-mccoy:
Okay created the other issue with comments:  https://github.com/defenseunicorns/zarf/issues/174

docs/what-is-zarf.md
btlghrants: Mmm... lemme make sure I'm clear on this one:

You're suggesting we add another "type of software" line in here to cover stuff that, 1) gets installed on the Zarf cluster machine, but 2) is _not_ a binary?  Stuff for Zarf to use to bootstrap other stuff?

btlghrants: Also, where would the "data injection" part fit?  I sort of imagined that was covered by the _container images_ line item but you're thinking it should be it's own thing?

btlghrants: Oh, unless you're thinking that we should call out that you can use Zarf to package stuff that has _nothing_ to do with the Zarf cluster at all.  So, like, to add an additional block that talks about if you just want to use the packaging facilities to seed your own non-Zarf-cluster stuff (i.e. TF modules for your own infra, etc.).

jeff-mccoy: Bingo

jeff-mccoy (approved):
I don't want this to get stale and think it's good enough in its current state.  We can continue to address the final definitions over the coming weeks.

PR #175, opened 2021-11-16
#115 Feature/enable linting

### LINKED ISSUE #115: Uncomment `golangci-lint` pre-commit hook
Uncomment the hook after fixing the golang linting errors

`golangci-lint` results:

image

_Originally posted by @RothAndrew in https://github.com/defenseunicorns/zarf/issues/112#issuecomment-946268723_

matt-strong:
Still need to fix regexp lint issue

Fixes #115


cli/cmd/root.go
```diff
@@ -64,6 +64,6 @@ func setLogLevel(logLevel string) {
 	case "panic":
 		logrus.SetLevel(logrus.PanicLevel)
 	default:
-		logrus.Fatal("Unrecognized log level entry: %s", logLevel)
```
jeff-mccoy, line 67: This looks like a typo, should be FatalF() I believe.

cli/internal/packager/inspect.go
```diff
@@ -32,7 +32,7 @@ func Inspect(packageName string) {
 	utils.ColorPrintYAML(text)
 
 	// Load the config to get the build version
-	config.LoadConfig(tempPath.base + "/zarf.yaml")
+	_ = config.LoadConfig(tempPath.base + "/zarf.yaml")
```
jeff-mccoy, line 35: We should handle an error here or they could get a crash

jeff-mccoy (changes requested):
Looks good after addressing the comments, thanks

cli/cmd/root.go
YrrepNoj: logrus.Fatalf( .... )

matt-strong: thanka

cli/internal/packager/inspect.go
matt-strong: copied what inspect.go does on an error loading config

cli/cmd/root.go
```diff
@@ -64,6 +64,6 @@ func setLogLevel(logLevel string) {
...
+		logrus.Fatalf("Unrecognized log level entry:", logLevel)
```
jeff-mccoy, line 67: `logrus.Fatalf("Unrecognized log level entry: %s", logLevel)` and it should be g2g, would have just fixed online but think that would bork our gpg signature requirement.

jeff-mccoy (changes requested):

jeff-mccoy (approved):

PR #180, opened 2021-11-29
Change terraform version

YrrepNoj:
We are currently using the latest docker image of CloudPosse which updated the version of Terraform which was breaking our e2e tests.

This is a major update of terraform though so hopefully nothing breaks.. Lets find out..


RothAndrew (approved):

RothAndrew:
@YrrepNoj let's merge

PR #185, opened 2021-12-02
Fix Big Bang example to do GitOps correctly

RothAndrew:
Fix the Big Bang Core example to do GitOps properly, by applying a `GitRepository` and `Kustomization` manifest, rather than applying the output of `kustomize build`

Because Flux can't pull from private bases without making modifications to the Flux deployment we decided to vendor the base rather than modify the deployment of flux. The `package-example-big-bang` Make target automatically pulls the base from the upstream Big Bang repo. It has a hard coded tag version right now which we should fix in a future issue. Because of this, reviewers should disregard any files that are in the `vendor` folder. They need to be included in the repo since the repo is what is pulled into the Zarf package that Flux uses inside the airgap.


examples/big-bang/manifests/big-bang/manifests.yaml
```diff
@@ -0,0 +1,59 @@
...
+    # exclude file extensions
+    /**/*.md
+    /**/*.txt
+    /**/*.sh
+  interval: 5m
+  url: http://stuart-gitea-http.git.svc.cluster.local:3000/zarf-git-user/mirror__github.com__defenseunicorns__zarf.git
+  secretRef:
+    name: zarf-git-secret
+  ref:
+    branch: feature/145-big-bang-core-example-correct-gitops
```
RothAndrew, line 34: This should change to `master` right before we merge

examples/big-bang/manifests/big-bang/manifests.yaml
```diff
@@ -0,0 +1,59 @@
...
+  name: bigbang
+  namespace: bigbang
+spec:
+  interval: 5m
+  path: "./examples/big-bang/template/bigbang"
+  prune: true
+  sourceRef:
+    kind: GitRepository
+    name: zarf
+#  healthChecks:
```
RothAndrew, line 49: Still need to do healthchecks here, need to figure out what the best thing to put here is

RothAndrew:
Proof this works:

image

image

RothAndrew:
Needs more documentation to be added before we merge also

runyontr:
Is there a reason any of the BB repo need to be private? Since upstream is publicly readable, I would expect no one to have an issue with a publicly readable BB on this repo either.

RothAndrew:
@runyontr Zarf automatically makes all repos that it loads into Gitea private. If you think it would be valuable we can create a feature request to be able to configure whether repos are public or private.

runyontr:
There's obviously a tradeoff on complexity between making zarf more complex/flexible (public/private repos) vs making the repos needed to build our packages more complex.  Just wondering if there were other considerations on that.

RothAndrew:
I'll write up an issue for the feature request.

RothAndrew:
@YrrepNoj this is ready for review and merge, other than the one thing to change the branch ref back to `master` right before we merge.

YrrepNoj (approved):
Small nit in the README but looks good overall

RothAndrew:
@YrrepNoj Where's the nit? I don't see a comment from you.

Changing the branch ref and I'll fix your nit, then I'll need one last approve then we can merge

examples/big-bang/README.md
```diff
@@ -8,19 +8,21 @@ Because the same cluster will be running both Traefik and Istio, Istio's Virtual
...
+   - `make all-dev` - Build Zarf locally, build the deploy package, and start a VM with Vagrant
+1. Run: `./zarf init --confirm --components management,gitops-service --host localhost` - Initialize Zarf, telling it to install the management component and gitops service and skip logging component (since BB has logging already) and tells Zarf to use `localhost` as the domain. If you want to use interactive mode instead just run `./zarf init`.
+1. Wait a bit, run `k9s` to see pods come up. Don't move on until everything is running
+1. Run: `./zarf package deploy zarf-package-big-bang-core-demo.tar.zst --confirm` - Deploy Big Bang Core. If you want interactive mode instead just run `./zarf package deploy`, it will give you a picker to choose the package.
+1. Wait several minutes. Run `k9s` to watch progress
+1. :warning: `kubectl delete -n istio-system envoyfilter/misdirected-request` (due to [this bug](https://repo1.dso.mil/platform-one/big-bang/bigbang/-/issues/802))
+1. Use a browser to visit the various services, available at https://*.bigbang.dev:9443
+1. When you're done, run `make vm-destroy` to bring everything down
```
YrrepNoj, line 25: Nit: Following the steps as they're listed, the user will be in the vagrant box. They'll either need to exit that or bring up a new terminal first.

RothAndrew: Updated

YrrepNoj (approved):

PR #187, opened 2021-12-06
Add BigBang Umbrella Package Example

YrrepNoj:
This example package extends upon the already existing Big Bang Core example and adds a handful of BigBang addons as
well as some additional helm charts for deploying Jira and Confluence.


YrrepNoj:
@RothAndrew I believe this is in a good enough state to merge in as it is. I'll start working on getting this to deploy in a GitOps way but that can be in a different PR.

examples/big-bang-umbrella/template/bigbang/values.yaml
```diff
@@ -472,10 +472,10 @@ twistlock:
     resources:
       requests:
         cpu: "100m"
-        memory: "256Mi"
+        memory: "1Gi"
       limits:
         cpu: "500m"
-        memory: "256Mi"
+        memory: "2Gii"
```
david--wright, line 478: Looks like this might be a typo? Should be 2Gi.

RothAndrew:
@YrrepNoj integrate master pls

examples/big-bang-umbrella/template/bigbang/values.yaml
```diff
@@ -0,0 +1,579 @@
...
+      resources:
+        requests:
+          cpu: "100m"
+          memory: "256Mi"
+        limits:
+          cpu: "500m"
+          memory: "256Mi"
+
+jaeger:
+  enabled: true
```
RothAndrew, line 218: What do you think about disabling this stuff that (a) is already demonstrated in the BB Core demo and (b) makes the package bigger and need more compute/memory without really showing the cool stuff (gitlab, mattermost, jira, confluence, etc)? I'm thinking if we just changed `enabled:` to false but left the rest, then commented out the lines in `zarf.yaml` it would slim down the package and still let people see an easy path to re-enabling if they want to.

Thoughts?

RothAndrew: Services I would disable are:

* jaeger
* kiali
* clusterAuditor
* gatekeeper
* logging
* eckoperator
* fluentbit
* monitoring
* twistlock

examples/big-bang-umbrella/zarf.yaml
```diff
@@ -0,0 +1,182 @@
...
+        shasum: 6645659d0737367a9465f626883e7266eeefe5668e3aca788cbab9d7ce7e10e1
+        target: "/usr/local/bin/kubescape"
+        executable: true
+      - source: https://github.com/armosec/regolibrary/releases/download/v1.0.62/nsa
+        shasum: bf66e316904b2adb3bc69529ea3fb6bd81229e3e951b6a02157cd31e42efbe58
+        target: "/usr/local/bin/kubescape-framework-nsa.json"
+      - source: "files/kubescape-exceptions.json"
+        target: "/usr/local/bin/kubescape-exceptions.json"
+
+  - name: bb-core
```
RothAndrew, line 30: rename, not just core anymore

RothAndrew (changes requested):

examples/big-bang-umbrella/template/bigbang/values.yaml
YrrepNoj: I agree and think this is a good idea. One of the thoughts that crossed my mind is if we should just remove them all together since we will still be pulling the images for all those disabled things so the package build time will be longer and the package itself will be larger than necessary.

I don't think that's an issue we need to solve and something we can adjust if we feel the need in the future.

RothAndrew (approved):

PR #192, opened 2021-12-07
add zarf-registry PullSecret to minio-instance for Postgres example

YrrepNoj:
@jeff-mccoy I know you said we could push to your branch if we wanted to work on things but that sounds like a cardinal sin of git so I couldn't bring myself to do it..

After this change the postgres example comes up fully!


jeff-mccoy (approved):
ugh im dumb, thx!

PR #196, opened 2021-12-09
Extract unrelated code fixes/features out of k8s native apply work

### LINKED ISSUE #18: Remove Containerd direct-dependency
We only use containerd for parsing the docker URI and bring in the containerd library for that.  Considering we also go-containertools and other tools in the ecosystem, we should evaluate alternatives for https://repo1.dso.mil/platform-one/big-bang/apps/product-tools/zarf/-/blob/master/cli/internal/images/push.go#L6.

---

Migrated from https://repo1.dso.mil/platform-one/big-bang/apps/product-tools/zarf/-/issues/32

### LINKED ISSUE #44: Unable to use "short" version of docker repo in "remote:" section of zarf.yaml
If I put a "short" version of a docker image in the `remote:` section, it throws an error when running ~~`zarf package create`~~ `zarf package deploy`:

image

Changing it to fully qualified e.g. `docker.io/rancher/coredns-coredns:1.8.3` works fine.

### LINKED ISSUE #193: Bug: (or docs issue): Games Example doesn't work on zarf 0.13.3 when using cert imported from Lets Encrypt Free CA
**Edit -- TLDR summary of findings:**         
Zarf fails to populate it's container registry, when imported HTTPS certs are used that don't allow for 127.0.0.1 (so public internet CA signed ones won't work), but zarf's self generated HTTPS certs will work.

**Summary:** 
So about the docs... I've messed with zarf several times, but have never been able to get anything beyond zarf init to work correctly. (which offers an empty git repo + empty registry)
(This is why I didn't realize that zarf can populates it's registry + git repo, when I talked over zoom with @jeff-mccoy)
I tried the game example doc (with and without slight modifications) and I get image pull backoff, which shows the registry never got populated.

Before I go into steps to reproduce the bug I think you'll find my desired use case valuable. 

<br><br>

**Background context info about my ultimate use case:**
1. I have a script to imperatively generate a declarative images.txt list of 20-100 images (depending on input parameters)
1. I plan to write a script to convert images.txt to a config like that can declaratively populate a registry. Basically script + images.txt ---> imperatively generated declarative zarf.yaml
1. Said script will look similar to this (was investigating rancher's hauler but it doesn't work with authenticated registries, but this will give you a concrete idea of where my thinking is at / how I plan to imperatively generate a declarative zarf.yaml)
```bash
export TEMPLATIZED_IMAGES_LIST=$(cat imperatively-generated-bb-images.txt | sed 's/^/    - ref: /' )
echo "$TEMPLATIZED_IMAGES_LIST"

cat > declarative-hauler-config.yaml <<EOF
apiVersion: content.hauler.cattle.io/v1alpha1
kind: Images
metadata:
  name: bigbang-images
spec:
  images:
$TEMPLATIZED_IMAGES_LIST
EOF

hauler store sync -f declarative-hauler-config.yaml
```
* Side Note: Can you bump registry:2's default PVC size from 10GB to 100GB? (It'll help my use case + won't hurt anything since local-path storage is thin provisioned)

<br><br>

**Copy Paste-able Reproducibility commands:**
* I use vagrant to provision a fresh centos7 VM
```bash
# provisioned fresh rhel7 for this and ssh'd in
# install git
sudo yum install git -y

# become root / per zarf docs
# I'm assuming all commands need to be run as root
sudo su - 

# install zarf 0.13.3 and zarf's dependencies artifact tar
cd ~
curl -L https://zarf-public.s3-us-gov-west-1.amazonaws.com/release/v0.13.3/zarf > zarf-cli
chmod +x zarf-cli
sudo mv zarf-cli /usr/bin/zarf
curl -L https://zarf-public.s3-us-gov-west-1.amazonaws.com/release/v0.13.3/zarf-init.tar.zst > zarf-init.tar.zst #538mb

# Install yq
export VERSION=v4.14.1
export BINARY=yq_linux_amd64
sudo wget https://github.com/mikefarah/yq/releases/download/${VERSION}/${BINARY} -O /usr/bin/yq && sudo chmod +x /usr/bin/yq

# install jq
sudo yum install jq -y

# verify yq / jq installed correctly
yq --version
jq --version

# pull dev cert/key pair for *.bigbang.dev
curl -L https://repo1.dso.mil/platform-one/big-bang/bigbang/-/raw/master/chart/ingress-certs.yaml | yq eval '.istio.gateways.public.tls.cert' - > bigbang.dev.crt
curl -L https://repo1.dso.mil/platform-one/big-bang/bigbang/-/raw/master/chart/ingress-certs.yaml | yq eval '.istio.gateways.public.tls.key' - > bigbang.dev.key

# Install zarf 0.13.3 & grab the zarf-init bootstrap artifact
cd ~
curl -L https://zarf-public.s3-us-gov-west-1.amazonaws.com/release/v0.13.3/zarf > zarf-cli
chmod +x zarf-cli
sudo mv zarf-cli /usr/bin/zarf

curl -L https://zarf-public.s3-us-gov-west-1.amazonaws.com/release/v0.13.3/zarf-init.tar.zst > zarf-init.tar.zst #538mb, takes about 1m20sec

# Pre-Login to registry1.dso.mil per zarf docs
export REGISTRY1_USERNAME=REPLACE_ME
export REGISTRY1_PASSWORD=REPLACE_ME
zarf tools registry login registry1.dso.mil -u $REGISTRY1_USERNAME -p $REGISTRY1_PASSWORD
# Lets you login w/o dependency on docker being installed, nice

# init zarf
cd ~
zarf init --server-key=$HOME/bigbang.dev.key --server-crt=$HOME/bigbang.dev.crt --host=bigbang.dev --components=management,gitops-service --confirm 
# ^-- takes about 20 secs

kubectl get pod -A  # zarf installs kubectl, just do this until it looks running

# Let's verify zarfs registry is empty b4 proceeding
# / Test Driven Development / helps to see what before and after look like
echo "127.0.0.1 registry.bigbang.dev" | tee -a /etc/hosts
export DOCKER_USER=$(sudo cat /root/.docker/config.json | jq '.auths."127.0.0.1".auth' | tr -d '"' | base64 -d | cut -d ':' -f 1) 
export DOCKER_PASS=$(sudo cat /root/.docker/config.json | jq '.auths."127.0.0.1".auth' | tr -d '"' | base64 -d | cut -d ':' -f 2)
zarf tools registry login registry.bigbang.dev -u $DOCKER_USER -p $DOCKER_PASS
zarf tools registry catalog registry.bigbang.dev
# (shows zarf cli splash, but under zarf is blank --implies--> empty registry)


# Zarf deploy the Game Example
cd ~
git clone https://github.com/defenseunicorns/zarf.git
cd ~/zarf/examples/game
zarf package create --confirm
# ^-- 8 seconds
zarf package deploy zarf-package-appliance-demo-doom.tar.zst --confirm
# INFO[0000] Deploy Zarf package confirmed
# INFO[0000] Loading dynamic config                        path=/tmp/zarf-768846375/zarf.yaml
# INFO[0000] Deploying Zarf component                      name=baseline
# INFO[0000] Loading images for local install
# INFO[0000] Loading images for gitops service transfer
# INFO[0000] Loading images
# INFO[0000] Updating image                                image="registry.dso.mil/platform-one/big-bang/apps/product-tools/zarf/game:doom"
# INFO[0000] 127.0.0.1/platform-one/big-bang/apps/product-tools/zarf/game:doom
# WARN[0000] Unable to push the image to the registry      image="registry.dso.mil/platform-one/big-bang/apps/product-tools/zarf/game:doom"
# INFO[0000] Loading manifests for local install, this may take a minute or so to reflect in k3s
# INFO[0000] Processing manifest file                      path=/tmp/zarf-768846375/components/baseline/manifests/game.yaml
# INFO[0000] Processing manifest file                      path=/tmp/zarf-768846375/components/baseline/manifests/image-pull-secret.yaml
# INFO[0000] Copying file                                  Destination=/var/lib/rancher/k3s/server/manifests Source=/tmp/zarf-768846375/components/baseline/manifests
# INFO[0000] Cleaning up temp files
########################################

# The WARN mentions an Error
# The Error is that it tried to push to registry.dso.mil  (repo1's registry built into gitlab)
# Why is zarf trying to do that? Bug? or config error?


# zarf's registry is still blank
zarf tools registry catalog registry.bigbang.dev
# (shows zarf cli splash, but under zarf is blank --implies--> empty registry)


kubectl get pod -A
# NAMESPACE  NAME                    READY    STATUS             RESTARTS   AGE
# default    game-69f5486bff-xjl7x   0/1      ImagePullBackOff   0          4m31s```

### LINKED ISSUE #194: Feature/Doc Request: zarf.yaml currently requires docker.io/ prefix to be specified for images from docker hub
**Short Description of Minor Issue:** 

I'm investigating tooling / workflow for an Ephemeral Transport Registry (for air gap registry imports with a good UX) and Zarf looks like a great candidate, I discovered If I have this:

```
kind: ZarfPackageConfig
metadata:
  name: ephemeral-transport-registry
components:
- name: images
  required: true
  images:
  - docker.io/busybox
  - docker.io/rancher/backup-restore-operator:v1.2.0
  - rancher/cis-operator:v1.0.6
  - rancher/configmap-reload:v0.3.0-rancher4
  - rancher/coredns-coredns:1.6.2
```
Then only the docker.io/... will get uploaded 
and the imperative url will be ignored
(so only 2/5 images will get loaded into zarf's registry)

<br>

**So my feature request / ask is for either:**
1. zarf cli to get smart enough to add docker.io/ prefix to images from docker hub
2. or document the limitation / requirement that images use full domain names.

<br>

**The following is additional details of an interesting use case / example usage:**

What I'm trying to do is as follows:
1. generate an arbitrary images.txt list of container images
2. imperatively generate a declarative zarf.yaml config from images.txt
3. use zarf to declaratively populate an ephemeral registry

```bash
# provision fresh centos7 and ssh in

# become root
sudo su - 

# install zarf 0.13.3 and zarf's dependencies artifact tar
cd ~
curl -L https://zarf-public.s3-us-gov-west-1.amazonaws.com/release/v0.13.3/zarf > zarf-cli
chmod +x zarf-cli
sudo mv zarf-cli /usr/bin/zarf
curl -L https://zarf-public.s3-us-gov-west-1.amazonaws.com/release/v0.13.3/zarf-init.tar.zst > zarf-init.tar.zst #538mb, takes about 1m20sec

# init zarf
cd ~
zarf init --components=management --host=127.0.0.1 --confirm

# wait for zarf registry to become healthy
time while [[ $(kubectl get pods -l app=docker-registry -n=registry -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do echo "waiting for registry to be ready (about 2 mins)..." && sleep 5; done; echo "Registry is ready"



# Create test data (that won't require registry auth for the sake of convenience)
cat > images.txt <<EOF
docker.io/busybox
docker.io/rancher/backup-restore-operator:v1.2.0
rancher/cis-operator:v1.0.6
rancher/configmap-reload:v0.3.0-rancher4
rancher/coredns-coredns:1.6.2
EOF

# If images in images.txt required auth could use this commented out stuff
# Pre-login to Authenticated Registry
# export REGISTRY1_USERNAME=REPLACE_ME
# export REGISTRY1_PASSWORD=REPLACE_ME
# zarf tools registry login registry1.dso.mil -u $REGISTRY1_USERNAME -p $REGISTRY1_PASSWORD

# Imperatively generate declarative zarf.yaml config from images.txt
TEMPLATIZED_IMAGES_LIST=$(cat images.txt | sed 's/^/  - /' )
cat > zarf.yaml <<EOF
kind: ZarfPackageConfig
metadata:
  name: ephemeral-transport-registry
components:
- name: images
  required: true
  images:
$TEMPLATIZED_IMAGES_LIST
EOF

# Eyeball validation of zarf config
head zarf.yaml

time zarf package create --confirm
time zarf package deploy zarf-package-ephemeral-transport-registry.tar.zst --confirm
# gives info about the 2 that worked, warn about the 3 the failed

# verify contents of registry
zarf tools registry catalog 127.0.0.1
```

jeff-mccoy:
### Breaking Changes:
* `localhost` is no longer a valid option for cluster ingress when initializing a zarf cluster. Instead you have to use a `127.0.0.1` or some other local ip found via `ifconfig`

### Fixes:
* No longer depends on 127.0.0.1 local bindings for the registry / gitops service
    * should fix #193 
* Resolve outstanding issues with image hostname swapping and
    * fixes #18
    * fixes #44
    * fixes #194

### Features:
* Adds `before` and `after` script options when defining a `zarf.yaml` with an optional retry flag
* Add symlink to ZarfFile for creating links to places files
* Add template boolean to ZarfFile to allow injection of zarf variables into text files
* Adds a new `zarf tool` command to print out config schema and commit the output to the repo (will need to make a git hook or something later on)
* Changes `zarf destroy` command to run any script that starts with `zarf-clean` instead of only running the k3s-remove script
* Add new ZarfState and `.zarf-state.yaml` for persisting host information from `zarf init` to `zarf package deploy`
* Remove all hard-coded logic for k3s install, now uses only standard zarf component features like everything else

#### Misc:
* Upgrades k3s from v1.21.2 to v1.21.6
* Adds optional regex filter for when performing RecursiveFileList()
* Adds more description to the components in zarf.yaml
* Renames type ZarfConfig to ZarfPackage in the config pkg
* Handful of general code organizing changes (moving yaml related functions to the `...../utils/yaml.go`, etc.)
* Expose execCommand() with stdout control
* Move traefik to standalone component and drop the internal k3s install of traefik
* Use the airgap tarball of K3s instead of manually listing images


cli/internal/images/push.go
```diff
@@ -24,19 +24,18 @@ func PushAll(imageTarballPath string, buildImageList []string) {
...
 		err = crane.Push(img, offlineName, cranePlatformOptions)
 		if err != nil {
 			logContext.Debug(err)
 			logContext.Warn("Unable to push the image to the registry")
 		}
 	}
 }
+
+func SwapHost(src string, targetHost string) string {
+	var parser = regexp.MustCompile(`(?im)^([a-z0-9\-.]+\.[a-z0-9\-]+:?[0-9]*)?/?(.+)$`)
```
YrrepNoj, line 38: Can we add a human readable docstring that describes what this regex is doing?

RothAndrew:
I think we do need to add in blocking the user from using `localhost` as the host. When I try it I get this error:

image

`zarf tools registry catalog localhost` gives me this:

image

If we do need to block the user from using `localhost`, we should also add a "tooltip" or some other kind of help text. Something like:

```
Enter a host DNS entry or IP Address for the cluster ingress.
Note: Due to <InsertReasonHere> you can't use "localhost" or "127.0.0.1.
Here are some suggestions that I've determined might work for you: [10.0.2.15]
```

YrrepNoj:
> I think we do need to add in blocking the user from using `localhost` as the host. When I try it I get this error:
> 
> image
> 
> `zarf tools registry catalog localhost` gives me this:
> 
> image
> 
> If we do need to block the user from using `localhost`, we should also add a "tooltip" or some other kind of help text. Something like:
> 
> ```
> Enter a host DNS entry or IP Address for the cluster ingress.
> Note: Due to <InsertReasonHere> you can't use "localhost" or "127.0.0.1".
> Here are some suggestions that I've determined might work for you: [10.0.2.15]
> ```

There's a chance you might have been using a different zarf version. I believe this PR prevents you from using localhost as the ingress.

cli/internal/images/push.go
jeff-mccoy: that's very human-readable sir!!!  j/k sorry regex is one of things, it reads pretty simple once you know it will, I'll do you one better I link to an example you can test with that explains each element.

jeff-mccoy:
So on the `localhost` thing, nothing should be changing behavior now that would break that.  But it's very important we understand you can't use `localhosts` and `127.0.0.1` interchangeably due to the certs not auto-magically adding those SAN values.  Can you show me your `.zarf-state.yaml` when you're having this issue?  Also I do think there might be some issues using `localhost` vs `127.0.0.1` that we've encountered before--we probably should block the actual DNS `localhost`.  Thoughts?

RothAndrew:
@jeff-mccoy You're right that `127.0.0.1` works, but `localhost` doesn't.

When I use `--host localhost` here's my .zarf-state.yaml, with the errors already mentioned above

```
kind: ZarfState
tls:
  certPublicPath: zarf-pki/zarf-server.crt
  certPrivatePath: zarf-pki/zarf-server.key
  host: localhost
```

jeff-mccoy:
Yeah this was an issue we ran into before if I recall, why we switched everything to hard-coded `127.0.0.1` vs `localhost`.  How do you feel about just blocking `localhost` for now so we don't totally change how you use this version of zarf?

cli/internal/images/push.go
jeff-mccoy: Please check it now and tell me what you think.

jeff-mccoy:
@RothAndrew made that change in the last commit

cli/internal/images/push.go
YrrepNoj: That's awesome!

cli/cmd/initialize.go
```diff
@@ -41,39 +48,51 @@ func handleTLSOptions() {
...
+		prompt.Message = "Enter a file path to the ingress private key"
+		_ = survey.AskOne(prompt, &state.TLS.CertPrivatePath, survey.WithValidator(survey.Required))
 	}
-	if !utils.CheckHostName(initOptions.PKI.Host) {
-		logrus.Fatalf("The hostname provided (%v) was not a valid hostname. The hostname can only contain: 'a-z', 'A-Z', '0-9', '-', and '.' characters.\n", initOptions.PKI.Host)
+
+	if !utils.CheckHostName(state.TLS.Host) {
+		// On error warn user and cycle the function
+		logrus.Warnf("The hostname provided (%v) was not a valid hostname. The hostname can only contain: 'a-z', 'A-Z', '0-9', '-', and '.' characters as defined by RFC-1035.  If using localhost, you must use the 127.0.0.1.\n", state.TLS.Host)
+		handleTLSOptions()
```
YrrepNoj, line 80: I think `survey.AskOne()` interprets `ctrl-C` as a character input (or it just stops the AskOne process instead of killing the parent command?). Because we have this cursive call a user can get stuck trying to exit this command leading to spamming of `ctrl-C` without actually killing the command.

We should (maybe not in this PR but make a ticket) check what output `survey.AskOne()` generated when a user uses `ctrl-C` and make sure we kill the process if we receive that.

RothAndrew:
For some reason, in the Big Bang example, the `flux_generated.yaml` manifest is not being deployed. I can see it there in the `/var/lib/rancher/k3s/server/manifests` folder, but it doesn't deploy.

Deploying the file with `kubectl apply -f flux_generated.yaml` works fine, and that unblocks the rest of the process from continuing.

cli/cmd/initialize.go
jeff-mccoy: yeah I noticed that too--super annoying

RothAndrew:
I changed the name of the file from `flux_generated.yaml` to `abc.yaml` and it deployed fine, so it must be something about the name of the file. I'll push a commit that changes the name of the generated file to something else that works.

jeff-mccoy:
that's super odd....

RothAndrew:
`flux_generated.yaml`: doesn't work

`flux-generated.yaml`: works

🤯 🤯 🤯

Makefile
```diff
@@ -51,7 +51,7 @@ build-cli: clean build-cli-linux build-cli-mac ## Build the CLI
 init-package: ## Create the zarf init package
 	$(ZARF_BIN) package create --confirm
 	mv zarf-init.tar.zst build
-	cd build && sha256sum -b zarf* > zarf.sha256
+	cd build && shasum -a 256 -b zarf* > zarf.sha256
```
RothAndrew, line 54: @jeff-mccoy reason behind this change? The test harness that CI uses doesn't have `shasum` installed, but it does have `sha256sum`. Is there any harm in changing it back to using `sha256sum`?

jeff-mccoy: thanks I changed that back and left a note, it was for macos, but there's an easy fix using brew.

jeff-mccoy:
hmm I thought we resolved these terraform versions issues already...also why do they keep popping up?  I've never seen so many issues with terraform versions before.  😭

RothAndrew:
We can loosen up the version constraint. It is very restrictive right now.

jeff-mccoy:
Set it to constrain per the TF docs rec, but we really can probably just open it wider to say `> 1.0.0` since the TF is so simple.  Punt that for a later discussion though.

YrrepNoj:
The e2e tests are failing because we're still using `localhost` in the game example and we rearranged the order of checks for the general cli. 

I'll work on fixing our e2e tests now.

cli/cmd/initialize.go
```diff
@@ -41,39 +48,51 @@ func handleTLSOptions() {
...
-			prompt := &survey.Input{
-				Message: "Enter a file path to the ingress public key",
-				Suggest: func(toComplete string) []string {
-					// Give some suggestions to users
-					files, _ := filepath.Glob(toComplete + "*")
-					return files
-				},
-			}
-			_ = survey.AskOne(prompt, &initOptions.PKI.CertPublicPath, survey.WithValidator(survey.Required))
+	// Always ask for a host entry to avoid having to guess which entry in a cert if provided
```
YrrepNoj, line 55: This update to always ask is now failing our e2e tests because it times out waiting for an input. This also means zarf can't be fully scripted and I feel like that's a bad thing overall.

@jeff-mccoy I'm not super familiar with networking/certs/anything.. Could you help me understand why this was changed?

YrrepNoj: This also currently make the `--host` flag really clunky if not irrelevant.

jeff-mccoy: Nope that's a bug 😱 we need to fix it

cli/config/config.go
```diff
@@ -17,12 +16,18 @@ const K3sManifestPath = "/var/lib/rancher/k3s/server/manifests"
...
-const ZarfLocalIP = "127.0.0.1"
 const ZarfGitUser = "zarf-git-user"
+const ZarfStatePath = ".zarf-state.yaml"
 
 var CLIVersion = "unset"
+var config ZarfPackage
+var state ZarfState
 
-var config ZarfConfig
+func init() {
```
YrrepNoj, line 26: Should this be an exported function? It's not being used in this package and I think it would be useful the in cmd package for initializing?

cli/cmd/initialize.go
```diff
@@ -41,39 +48,51 @@ func handleTLSOptions() {
...
+	} else {
+		if err := config.WriteState(state); err != nil {
+			logrus.Debug(err)
+			logrus.Fatal("Unable to save the zarf state file.")
+		}
 	}
 }
 
 func init() {
+	state := config.GetState()
```
YrrepNoj, line 90: I'm going to remove this line. We defined `state` up above around line 18. This line overrides that variable but only exists in the local scope so we don't have the provided host/tls-paths users provided.

cli/config/config.go
jeff-mccoy: init()?

https://stackoverflow.com/questions/24790175/when-is-the-init-function-run

cli/cmd/initialize.go
YrrepNoj: TIL. Cool!

cli/config/config.go
YrrepNoj: TIL... Cool!

cli/cmd/initialize.go
RothAndrew: I assume this will get fixed already as well, just documenting here in case it really is some different issue:

If I say "Generate TLS chain with ephemeral CA", then type "localhost" for the host, it is rejected as expected. If I then type `127.0.0.1` as the host, it tells me "Enter a file path to the ingress public key" as if I had chosen to import the cert rather than generate it.

image

test/e2e/e2e_general_cli_test.go
```diff
@@ -83,12 +83,13 @@ func testGeneralCliStuff(t *testing.T, terraformOptions *terraform.Options, keyP
...
 
 	// Test for expected failure when given a bad component input
-	output, err = ssh.CheckSshCommandE(t, publicHost, fmt.Sprintf("cd /home/%s/build && ./zarf init --components management,foo,logging", username))
+	output, err = ssh.CheckSshCommandE(t, publicHost, fmt.Sprintf("cd /home/%s/build && ./zarf init --host 127.0.0.1 --components management,foo,logging", username))
 	require.Error(t, err, output)
 
 	// Test for expected failure when given invalid hostnames
-	output, err = ssh.CheckSshCommandE(t, publicHost, fmt.Sprintf("cd /home/%s/build && ./zarf init --host bad!hostname", username))
-	require.Error(t, err, output)
+	// output, err = ssh.CheckSshCommandE(t, publicHost, fmt.Sprintf("cd /home/%s/build && ./zarf init --host bad!hostname", username))
```
YrrepNoj, line 90: This PR recursively asks the user for the hostname if an invalid option is provided which would hang this test. Probably should have removed instead of just commenting it out but I wanted to see if anyone had opinions/thoughts on how we could still test this.

cli/cmd/initialize.go
YrrepNoj: Similar but that was a slightly different issue. That should be fixed now.

test/e2e/e2e_general_cli_test.go
jeff-mccoy: I think it's better to prompt again rather than just abort if it's an active user prompt, but if they pass `--host` with an invalid entry it should abort.  For now if it's easier we can just maintain the existing abort behavior.

cli/cmd/initialize.go
```diff
@@ -75,8 +82,9 @@ func handleTLSOptions() {
 	}
 
 	if !utils.CheckHostName(state.TLS.Host) {
-		// On error warn user and cycle the function
+		// On error warn user, reset the field, and cycle the function
 		logrus.Warnf("The hostname provided (%v) was not a valid hostname. The hostname can only contain: 'a-z', 'A-Z', '0-9', '-', and '.' characters as defined by RFC-1035.  If using localhost, you must use the 127.0.0.1.\n", state.TLS.Host)
+		state.TLS.Host = ""
```
jeff-mccoy, line 87: great catch t

test/e2e/e2e_general_cli_test.go
RothAndrew: I just ran across this. Running `./zarf init --confirm --components management,gitops-service --host localhost` prompts me for a different host rather than throwing a fatal

RothAndrew: `./zarf init --confirm --components management,gitops-service --host 127.0.0.1` also get stuck waiting for a user prompt.

(I'm fairly sure you guys already know about this and are already working on it, just making sure I'm documenting the things I do as I test)

YrrepNoj: For the second comment, I added a `--generate` flag so we wouldn't get stuck waiting for a user prompt. 

I haven't addressed the first comment though, Jeff messaged me saying he was going to rewrite the cert/hostname logic so I was going to wait for him to push thinking he was going to take care of it.

RothAndrew: Got it, so we are now expecting either `--generate` or `--server-crt` and `--server-key`? That works for now. Later we should add a check that fails if `--confirm` is set but `--generate` or `--server-crt --server-key` isn't.

RothAndrew: Should we name it `--generate-ca` or something? `--generate` doesn't seem quite descriptive enough (generate what?)

RothAndrew:
Running `./zarf package deploy zarf-package-big-bang-core-demo.tar.zst --confirm` prompts the user to choose whether they want kubescape or not. Honestly I'm not sure if this was the case before these changes or not.

test/e2e/e2e_general_cli_test.go
jeff-mccoy: hey sorry I'm about done rewriting all this, messaged @YrrepNoj should have put it in here.

jeff-mccoy: I actually threw away the generate flag as it's duplicate of providing the cert paths

jeff-mccoy: sorry @RothAndrew

YrrepNoj:
> 

That was the case before but I agree that it shouldn't need to be.

test/e2e/e2e_general_cli_test.go
RothAndrew: Not a problem. I'll go through and undo adding the `--generate` flag to the docs and stuff, should only take a couple of minutes

jeff-mccoy:
can we gather?

RothAndrew:
yep

RothAndrew:
As soon a Jeff pushes up the fix to remove `--generate` and not need it I think I'll be ready to approve and merge

RothAndrew:
I'm not sure how it's possible that the E2E tests are passing if we aren't using `--generate` but Jeff's fix hasn't been pushed up yet. They should be stuck waiting for a nonexistent user to hit the enter key

test/e2e/e2e_general_cli_test.go
jeff-mccoy: these issues are addressed in the last commit

cli/cmd/initialize.go
jeff-mccoy: fixed in the last push

RothAndrew:
I'm still being prompted to choose whether to generate the CA or to import one when running `./zarf init --confirm --components management,gitops-service --host 127.0.0.1`. I'm completely at a loss why I'm seeing it on my machine but the E2E test isn't.

jeff-mccoy:
> I'm still being prompted to choose whether to generate the CA or to import one when running `./zarf init --confirm --components management,gitops-service --host 127.0.0.1`. I'm completely at a loss why I'm seeing it on my machine but the E2E test isn't.

Okay let me look at this

RothAndrew:
image

I thought maybe I was still on an older commit somehow but there's the "press tab to see your valid values" stuff...

image

jeff-mccoy:
oh weird its a race condition

jeff-mccoy:
If you let it sit there it will keep running

jeff-mccoy:
crap I remember now, I was supposed to put a top-level confirm box but forgot to.... ugh

jeff-mccoy:
........and now registry1 auth is down lol

RothAndrew:
Yep, that was my signal to quit for the night :D

jeff-mccoy:
Okay I'm done with the PR now, all tests pass and I set it to squash/automerge whenever you guys get back to work.   I'm going to pick back up my other work later tonight and reconcile against this branch even if it doesn't get merged until next week.  Thanks for all the coordination, #teamworkMakesTheDreamWork.

YrrepNoj (approved):

PR #207, opened 2021-12-14
Bug Fix: Fix regexp creation to use a proper string

YrrepNoj:
Without this, the regex will not catch charts provided by git and all charts downloaded as a published chart.

I was trying to build the postgres example package and was getting an error because it was trying to download helm charts as a published helm chart instead of as a chart hosted on a git repo.


YrrepNoj:
@jeff-mccoy I realize that all other places that use the `regexp.MustCompile` are using the ` instead of a string quote "".  Me switching to the string quotes let me actually create the package though..

jeff-mccoy (approved):

PR #209, opened 2021-12-15
Add example for istio with preloaded cert

### LINKED ISSUE #120: Lightweight Big Bang Core example that expects Istio's cert to already be present
Quick persona:
I'm Joe Schmoe. I'm a user of Zarf who is using it to deploy Big Bang, but I don't want my TLS certificate to be in the main package, since I want to make the zarf package freely available, but need to protect the cert as a secret.

Story:
As Joe Schmoe, I want an example Zarf package that I can use as a reference to create a Big Bang deployment that expects the ingressgateway's certificate to already exist in the correct kubernetes secret in the correct namespace, so that I can provide my private secret separately from the zarf package.

RothAndrew:
Fixes #120


examples/istio-with-separate-cert/manifests/big-bang/manifests.yaml
```diff
@@ -0,0 +1,65 @@
...
+    # exclude file extensions
+    /**/*.md
+    /**/*.txt
+    /**/*.sh
+  interval: 5m
+  url: http://stuart-gitea-http.git.svc.cluster.local:3000/zarf-git-user/mirror__github.com__defenseunicorns__zarf.git
+  secretRef:
+    name: zarf-git-secret
+  ref:
+    branch: feature/120-preloaded-istio-cert
```
RothAndrew, line 34: This will need to change back to master before we merge

PR #211, opened 2021-12-15
Fix error handling when building a zarf package with gitops repos that have `@tag` format

### LINKED ISSUE #210: `zarf package deploy` broken when using gitops with repos that use `@tag` notation
When in when using gitops-service, repos with `@tag` notation fail to resolve when running `zarf package deploy`. This fixes it. I'm guessing what is going on is that the error that is being returned doesn't affect the functionality of the checkout, so it should just be ignored instead of acted on, but you guys are closer to this than I am so feel free to poke holes in it.

image

RothAndrew:
Fix #210

### DIFF
--- a/cli/internal/git/utils.go
+++ b/cli/internal/git/utils.go
@@ -195,7 +195,7 @@ func RemoveHeadCopies(gitDirectory string) []*plumbing.Reference {
 	head, err := repo.Head()
 	if err != nil {
 		logContext.Debug(err)
-		logContext.Fatal("Failed to identify references")
+		logContext.Fatal("Failed to identify references when running `repo.Head()`")
 	}
 
 	headHash := head.Hash().String()
@@ -225,7 +225,7 @@ func removeReferences(
 	references, err := repo.References()
 	if err != nil {
 		logContext.Debug(err)
-		logContext.Fatal("Failed to identify references")
+		logContext.Fatal("Failed to identify references when running `repo.References()`")
 	}
 
 	head, err := repo.Head()
--- a/cli/internal/git/utils.go
+++ b/cli/internal/git/utils.go
@@ -195,7 +195,7 @@ func RemoveHeadCopies(gitDirectory string) []*plumbing.Reference {
 	head, err := repo.Head()
 	if err != nil {
 		logContext.Debug(err)
-		logContext.Fatal("Failed to identify references when running `repo.Head()`")
+		logContext.Fatal("Failed to identify references when getting the repo's head")
 	}
 
 	headHash := head.Hash().String()
@@ -225,7 +225,7 @@ func removeReferences(
 	references, err := repo.References()
 	if err != nil {
 		logContext.Debug(err)
-		logContext.Fatal("Failed to identify references when running `repo.References()`")
+		logContext.Fatal("Failed to identify references when getting the repo's references")
 	}
 
 	head, err := repo.Head()
--- a/cli/internal/git/checkout.go
+++ b/cli/internal/git/checkout.go
@@ -57,7 +57,11 @@ func checkoutHashAsBranch(path string, hash plumbing.Hash, branch plumbing.Refer
 		logContext.Fatal("Not a valid git repo or unable to open")
 	}
 
-	objRef, _ := repo.Object(plumbing.AnyObject, hash)
+	objRef, err := repo.Object(plumbing.AnyObject, hash)
+	if err != nil {
+		logContext.Debug(err)
+		logContext.Fatal("An error occurred when getting the repo's object reference")
+	}
 
 	var commitHash plumbing.Hash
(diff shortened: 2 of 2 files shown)


jeff-mccoy:
I'm going to need some more help understanding this issue / reproducing it.  I have a concern with not handling the error, most functions can have different errors, some that matter, some that don't.  I also cannot reproduce the issue with the `gitops-data` example, do you have an example I can you to reproduce?

YrrepNoj:
I think I might have been seeing this on my umbrella example in my PR branch.

RothAndrew:
It’s happening consistently on my branch for doing Istio with a preloaded cert

jeff-mccoy:
Can you reproduce on a clean branch (master or branch off of master) so we have a good baseline for isolation?

RothAndrew:
Yep

RothAndrew:
It's happening with the `gitops-data` example without making any changes to it.

* `cd examples`
* `make clean build-release package-example-gitops-data vm-init`
* (inside VM) `./zarf init`
  * host? `127.0.0.1`
  * Generate CA? `yes`
  * management? `yes`
  * logging? `no`
  * gitops-service? `yes`
* `./zarf package deploy <TheGitOpsServiceTarball>`

image

RothAndrew:
I'd like to merge this asap since all it does is revert a previous change. We can then add an issue for handling this error properly. Does that work for everyone?

jeff-mccoy:
Still concerning since it's not happening to me on master with that example, but we can proceed with the merge for now I guess.  Will look at this more later on.

YrrepNoj:
That works for me as long as we update the revert _slightly_ so that it doesn't just drops the err but logs it as a warning.

YrrepNoj:
@jeff-mccoy that's weird that it's not happening for you on master, it is for me. Did you build the package from the same cli version? This change is when the package gets created, it just becomes obvious that something is weird when you deploy it.

jeff-mccoy:
Yeah I built from master, but let's just proceed because there other other strange things I need to look at more later on here anyway, like the fact the tag is captured in the url replace name.

RothAndrew:
I'm making the change Jon requested. I'll have it pushed up in a few minutes.

RothAndrew:
whelp, now it's not happening. I wonder if the earlier change from ``` to `"` fixed it and I didn't have it pulled down? If so that's gonna make me feel real dumb

jeff-mccoy:
I live in a state perpetually dumbness, it's not so bad really

YrrepNoj:
It's still happening for me.. I double checked to make sure I had the most recent version of master installed and did a full clean.. Fun!

RothAndrew:
It's definitely still happening to me on master if I try to deploy the `gitops-data` example. What's weird is I'm not seeing the warning messages that I'm expecting when I switch over to this branch and try

RothAndrew:
@YrrepNoj do the warnings that I added look okay? I also updated a couple of error messages in another file to help them be distinguished from eachother

YrrepNoj:
Yeah those logs make sense.

I think I realized what the issue was before hand though. The original line 62 should have been `if err == nil`. There is no error happening (which is why you weren't seeing your logs). But because we had the error check backwards we weren't actually doing any of the work.

RothAndrew:
Ah got it. So instead of doing that let's do a standard fatal in a regular error checking block. I'll make that update and get it pushed out shortly.

YrrepNoj (approved):

RothAndrew:
@YrrepNoj one more review please then I think we'll be ready to merge

cli/internal/git/utils.go
```diff
@@ -195,7 +195,7 @@ func RemoveHeadCopies(gitDirectory string) []*plumbing.Reference {
 	head, err := repo.Head()
 	if err != nil {
 		logContext.Debug(err)
-		logContext.Fatal("Failed to identify references")
+		logContext.Fatal("Failed to identify references when running `repo.Head()`")
```
YrrepNoj, line 198: Oh I didn't see this, I won't block this but I wouldn't typically put the function called in a log like this, at least not in a fatal log. It might make more sense in a debugging log since it can help a developer figure out whats wrong but telling an end user what function was called isn't useful information to them.

@jeff-mccoy opinions?

RothAndrew: I'll reword it a bit, the intent was to make two different blocks have slightly different error messages. They both used to say the same thing.

YrrepNoj (dismissed):

cli/internal/git/utils.go
RothAndrew: updated

RothAndrew:
@YrrepNoj one more review pls, sorry :)

YrrepNoj (approved):

PR #233, opened 2022-01-19
reintroduce postgres operator example database pods

### LINKED ISSUE #201: The `postgres-operator` example doesn't quite work
When deploying the `postgres-operator` example, it looks like the operators come up fine, but a postgres database isn't being created like before. It doesn't look like any of the helm chart or container version changed, but I guess we can't really know whether they mutated the tag that we are using for the postgres operator.

image

wadedesir:
Should fix #201 It appears the postgres-cluster yaml was not present in the newer versions of this example and thus the database pods were not being created accordingly.

### DIFF
--- a/examples/postgres-operator/README.md
+++ b/examples/postgres-operator/README.md
@@ -2,8 +2,6 @@
 
 This example demonstrates deploying a performant and highly available PostgreSQL database to a Zarf airgap cluster. It uses Zalando's [postgres-operator](https://github.com/zalando/postgres-operator) and provides the Postgres Operator UI and a deployment of PGAdmin for demo purposes.
 
-:warning: NOTE: It looks like this example doesn't currently quite work. The operators come up but it doesn't deploy a postgres database like it used to. We are working on a fix.
-
 ## Tool Choice
 
 After looking at several alternatives, Zalando's postgres operator felt like the best choice. Other tools that were close runners-up were the postgres-operator by [CrunchyData](https://github.com/CrunchyData/postgres-operator) and [KubeDB](https://github.com/kubedb/operator).
--- a/examples/postgres-operator/manifests/image-pull-secret.yaml
+++ b/examples/postgres-operator/manifests/image-pull-secret.yaml
@@ -3,6 +3,37 @@ kind: Secret
 type: kubernetes.io/dockerconfigjson
 metadata:
   name: private-registry
+  namespace: default
+stringData:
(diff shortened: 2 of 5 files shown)


YrrepNoj:
Do you happen to know around which commit this yaml was removed in? I'm curious if it was an accident or if we were trying to do something different/specific that didn't work the way we expected.

RothAndrew (changes requested):
* Clean up the warning in the readme saying this example doesn't work

RothAndrew (changes requested):
Getting an ImagePullBackoff

image

image

wadedesir:
> Do you happen to know around which commit this yaml was removed in? I'm curious if it was an accident or if we were trying to do something different/specific that didn't work the way we expected.

@YrrepNoj looks like they were removed with commit f96ae20 / #144

> * Clean up the warning in the readme saying this example doesn't work

Will do

> Getting an ImagePullBackoff. Make sure the pod spec has the imagePullSecret in its config

@RothAndrew That's weird. They were coming up fine without it on my end. Let's take a look at it together when you have some time but try the new changes in the meantime.

RothAndrew:
@wadedesir still getting ImagePullBackOff. See [this github issue](https://github.com/zalando/postgres-operator/issues/546).

Not sure why you aren't getting it. Are you using the vagrantfile that modifies the hosts file of the VM? If you aren't then it will just pull directly from the upstream registry and won't need authorization for the internal zarf docker registry.

wadedesir:
@RothAndrew @YrrepNoj Everything should be working as expected now.

RothAndrew:
Screenshot coming in from @wadedesir. Through debugging it sounds like you can't configure imagePullSecrets or a different service account for the pooler pod, but we found this [github issue](https://github.com/zalando/postgres-operator/issues/1644) that talks about patching the default serviceaccount in that namespace. I recommended that we do that same thing but as a `scripts.after:` script in the component.

wadedesir:
> Screenshot coming in from @wadedesir. Through debugging it sounds like you can't configure imagePullSecrets or a different service account for the pooler pod, but we found this [github issue](https://github.com/zalando/postgres-operator/issues/1644) that talks about patching the default serviceaccount in that namespace. I recommended that we do that same thing but as a `scripts.after:` script in the component.
Screen Shot 2022-01-24 at 3 43 29 PM

RothAndrew (approved):

PR #234, opened 2022-01-20
Add skeleton of new E2E test that tests Gitea and Grafana

RothAndrew:
This PR will add the baseline for a new test, a new PR will add the actual test. It has to be done this way because GitHub Actions only pulls its job definitions from master, not branches.


YrrepNoj (approved):
I am not super knowledgeable about github pipelines yet but this looks good to me.

PR #237, opened 2022-01-20
Multi-distro support (Zarf Native Apply Work)

### LINKED ISSUE #2: ARM64 Support
**Will need to break this out into an epic**

Although the Iron Bank lacks significant ARM support at the moment, we only need two images for shift pack's cluster to function (registry and gitea) that could be targeted for Iron Bank ARM support as a priority.  

Reference on where to begin:
https://repo1.dso.mil/platform-one/big-bang/apps/product-tools/shift/pack/-/blob/master/Earthfile#L112

---

Migrated from https://repo1.dso.mil/platform-one/big-bang/apps/product-tools/zarf/-/issues/2

### LINKED ISSUE #16: Natively handle manifest application
For greater flexibility, change from K3s manifest applier to zarf applying manifests.  Because CRDs will also require some concept of ordering, we should evaluate how Helm does this and use that if possible.  The client-go APIs are very low-level for the complexity of this operation.

---

Migrated from https://repo1.dso.mil/platform-one/big-bang/apps/product-tools/zarf/-/issues/30

### LINKED ISSUE #17: Make k3s installer stages completely generic to the packaging system
Removing all hard-coded elements of K3s will enable transparent support for alternative cluster technologies including K3d to make development and testing simpler.  This will likely require #23 to be implemented.

---

Migrated from https://repo1.dso.mil/platform-one/big-bang/apps/product-tools/zarf/-/issues/31

### LINKED ISSUE #35: Support Kustomizations
As a user of Zarf, I want the ability to provide Kustomizations to be applied, so that I don't have to build the manifests myself using `kustomize build`

### LINKED ISSUE #36: Migrate example containers to github (or dockerhub) from gitlab
The [game example](https://github.com/defenseunicorns/zarf/tree/master/examples/game) has an image it needs to run the demo that we need to move off of Repo1 container registry to turn off that feature in the project mirror.  

https://repo1.dso.mil/platform-one/big-bang/apps/product-tools/zarf/container_registry

### LINKED ISSUE #109: Add ability to make Zarf quieter
I'd like the ability to do something like `zarf --quiet` that will suppress the big Zarf ascii image, so that if I'm logging output from Zarf it doesn't clutter my logs with extra fluff.

### LINKED ISSUE #140: Evaluate options to replace containerd mirroring config
K3s containerd mirror config works really well, but makes it harder to easily support other K8s distros without knowing the underlying container runtime details.  Deploying a mutating webhook would enable a k8s-native approach to this problem.  The Google surfaced a couple possibilities to consider adopting (both have very small contributor lists):

https://github.com/jainishshah17/tugger
https://github.com/estahn/k8s-image-swapper?ref=golangrepo.com

jeff-mccoy:
# Changelog

- add arm64/generic arch support
- adds in-memory embedded registry to serve registry image from "seed" config spec
- add local-distro image injection support
- handles helm charts install/upgrade/rollback/uninstall via helm directly
- manages raw manifests as zarf-generated helm charts for helm-native management
- add tunnel capability to access zarf endpoints without ingress config
- auto-inject imagePullSecret into any service account zarf touches and any namespace default service account
- auto-inject registry pull credentials into all namespaces for use by service accounts
- allow multiple helm charts to be installed in the same namespace without conflict
- flattens image collection to single tarball
- remove root / linux requirements for zarf deployments
- remove k3s install requirement
- remove traefik component requirement
- add K8s distro-detection and distro-specific seed image behaviors
- document local registry per https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/generic/1755-communicating-a-local-registry
- reorganize logging, remove logrus, replace with pterm, add progress cli feedback
- add support for stacking helm values files
- normalize log levels: info, warn, debug, trace 
- update go mod deps
- move zarf state to a k8s secret in the zarf namespace
- support zarf destroy for 3rd-party clusters
- move crane image cache to sub folder under home
- add support for multiple users with htpasswd
- add path key for git-based helm charts
- remove need for PKI prompts on default init
- add storage class tracking in zarf state for multi-distro compatibility 
- reduce code for init/deploy to the same function call
- add command line aliases for the cli
- vendor k9s into zarf to support multi-arch/multi-os 
- make more games for the game example to demonstrate multiple `zarf connect ` in the same component
- add `zarf prepare find-images` to lookup images for a zarf config to help with package creation
- add `zarf destroy --remove-components` for removing zarf-installed charts outside of the zarf ns    
- add `zarf connect 🦄` combined with a new svc label, `zarf.dev/connect-name=🦄` to make dynamic app tunnels

Fixes #140 
Fixes #2 
Fixes #16 
Fixes #109 
Fixes #17
Fixes #35 
Fixes #36 

## Manual Validation of Examples:

Some of the key examples should be confirmed to work both in the existing "appliance mode" and using this new external Kubernetes installation mode.  Note that the external Kubernetes cluster currently has to be a *local* cluster, so validation of the new capabilities should be done via Kind.  The manual tests should include the following steps:

### External Cluster

1. Show output of `kubectl get all -A`
2. Create and deploy the Zarf package for the example using the documentation
3. Verify all components up and healthy
4. Verify ingresses/UIs up and healthy
5. Teardown/Destroy installation
6. Ensure output of `kubectl get all -A` matches output from 1.

### Appliance Mode

1. Create and deploy the Zarf package for the example using the documentation
2. Verify all components up and healthy
3. Verify ingresses/UIs up and healthy
4. Teardown/Destroy installation

##

- [x] Big Bang
  - [x] Update Docs (@RothAndrew)
  - [x] ~~KIND installation successful (@RothAndrew)~~
  - [x] Appliance installation successful (zarf init w/k3s install) (@RothAndrew)
  - [x] Revert back to "correct" way to do GitOps (potentially future backlog item) (#268)
- [x] Data Injection
  - [x] Update Docs
  - [x] KIND installation successful
- [X] Game (@RothAndrew) (#289)
  - [X] Update Docs (@RothAndrew)
  - [x] Add asciinema graphics back in once they get updated (future backlog item) (@RothAndrew) (#286)
  - [x] KIND installation successful
- [x] Gitops-data (@RothAndrew)
  - [x] Update Docs (@RothAndrew)
  - [x] KIND installation successful
- [x] Postgres-operator (@RothAndrew)
  - [x] Update Docs (@RothAndrew)
  - [x] KIND installation successful
- [x] single-big-bang-package
  - [x] Update Docs (@jeff-mccoy)
  - [x] KIND installation successful
- [x] tiny-kafka
  - [x] Update Docs (@jeff-mccoy)
  - [x] KIND installation successful


zarf.yaml
```diff
@@ -1,14 +1,25 @@
...
       K3s provides the cluster need for Zarf running in Appliance MOde as well as can
       host a low-resource Gitops Service if not using an existing Kubernetes platform.
-    required: true
+    default: true
     scripts:
       retry: true
+      before:
+        # If running RHEL variant, disable firewalld
+        # https://rancher.com/docs/k3s/latest/en/advanced/#additional-preparation-for-red-hat-centos-enterprise-linux
+        - "[ -e /etc/redhat-release ] && systemctl disable firewalld --now"
```
YrrepNoj, line 22: If not running a RHEL variant this hangs and eventually makes the init fail because of a timeout.

YrrepNoj: Ooooh It's because of the retries. It's taking the `1` response when checking if the redhat-release file exists as a bash error

cli/internal/helm/chart.go
```diff
@@ -0,0 +1,354 @@
...
+			} else {
+				spinner.Updatef("Performing chart uninstall")
+				_, _ = uninstallChart(actionConfig, options.ReleaseName)
+			}
+			spinner.Errorf(nil, "Unable to complete helm chart install/upgrade")
+			break
+		}
+
+		spinner.Updatef("Checking for existing helm deployment")
+		if _, histErr := histClient.Run(options.ReleaseName); histErr == driver.ErrReleaseNotFound {
```
YrrepNoj, line 74: I think this conditional block has a couple bugs in it. Does this look more right to you?


```
if _, histErr := ............ {
    ........
} else if histErr == driver.ErrReleaseExists {.         // check histErr instead of err and check for upgrade before just assuming failure
    upgradeChart( ... )
} else {
    fatal("unable to verify chart"
}

**Comment by YrrepNoj** [comment]
Almost all of the `zarf tools registry <.....>` sub-commands are giving me a panic because of a segfault/nil pointer.

File: cli/internal/helm/chart.go (hunk 2 above)
Comment (YrrepNoj) [reply cli/internal/helm/chart.go:74]: Just kidding, this change breaks the `zarf init` with a nil pointer. Either way there's a bug here we should work out.

File: zarf.yaml (hunk 1 above)
Comment (YrrepNoj) [reply zarf.yaml]: I resolved this. @jeff-mccoy let me know if you don't like this approach.

**Comment by matt-strong** [comment]
https://user-images.githubusercontent.com/90784516/151041048-c5d70538-6bd3-4b30-a814-8562a8a93a0a.mov

Not a huge deal, but on zarf package create for big-bang the pulling images step always goes up to 63% then immediately completes.

**Comment by jeff-mccoy** [comment]
> pulling-images.mov 
> Not a huge deal, but on zarf package create for big-bang the pulling images step always goes up to 63% then immediately completes.

yeah I've struggled with how to handle this--that's actually the caching system, it can't really do progress.  If you have time could you delete `~/.zarf-image-cache/` and try again?

File: cli/internal/packager/common.go [hunk 3]
```diff
@@ -187,7 +204,7 @@ func HandleIfURL(packagePath string, shasum string, insecureDeploy bool) string
 func isValidFileExtension(filename string) bool {
 	for _, extension := range config.GetValidPackageExtensions() {
 		if strings.HasSuffix(filename, extension) {
-			logrus.WithField("packagePath", filename).Warn("Package extension is valid.")
+			message.Warnf("Extension for %s is invalid", filename)
```
YrrepNoj, line 207: I'm not sure why this was a warning before but this looks like we have our log flipped.

jeff-mccoy: hmm that's super odd......yeah that whole thing doesn't even warrant a log at that level I think

cli/internal/packager/common.go
```diff
@@ -134,50 +152,49 @@ func HandleIfURL(packagePath string, shasum string, insecureDeploy bool) string
...
 	resp, err := http.Get(packagePath)
 	if err != nil {
-		logrus.Fatal("Unable to download the package: ", err)
+		message.Fatal(err, "Unable to download the package")
 	}
 	defer resp.Body.Close()
 
 	// Write the package to a local file
 	tempPath := createPaths()
+	defer tempPath.clean()
```
YrrepNoj, line 176: I think we need to think of a better way to do this, either pass off the responsibility of clearing to somewhere else or return the package body instead of a path to the local filesystem.

This defer wipes away the downloaded package before it's able to actually be used during the `deploy` stage.

YrrepNoj: Or even simpler.. We can create the tmp directory in the `Run` section of the cobra DeployCmd and leave the defer in there since the first thing `packager.Deploy()` does is create the /tmp/ directory anyways.

jeff-mccoy: the way we do it everywhere in the codebase is to pass the parent temp path and just let it be handled by the parent deploy function

jeff-mccoy:
thanks for the very detailed PR updates @runyontr, looks way better with all that data

examples/tiny-kafka/manifests/kafka.yaml
```diff
@@ -1,8 +1,3 @@
-apiVersion: v1
```
YrrepNoj, line 1: Was this purposefully removed? There doesn't seem to be anything elsewhere that would create this namespace (the charts definition in the zarf.yaml for this example creates the kafka-operator namespace).

I'll be adding it back in a bit but wanted to make sure I wasn't missing anything.

YrrepNoj: After looking at this more I think there's an issue here that needs a little tinkering still.  The expected `kafka-demo` namespace isn't being created and the helm install/upgrade is erroring out because of that. Seems like other manifests aren't auto-created by helm and we'll need to do that if we encounter a manifested not already defined.

YrrepNoj: @jeff-mccoy

jeff-mccoy: please let me know how this goes tomorrow, that was actually a really nasty problem in the end, 04a801d4160ceb0ae153586cc6db73e0379455f6 should correct this behavior

YrrepNoj: Seems good to me now, thanks.

RothAndrew: This sounds worthy of a robust couple of tests.

examples/big-bang/template/bigbang/vendor/bigbang/base/flux/kustomization.yaml
```diff
@@ -33,7 +33,7 @@ patches:
               cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
           spec:
             imagePullSecrets:
-            - name: private-registry
+            - name: zarf-registry
```
YrrepNoj, line 36: This change gets blown away with our current Make recopies. `make package-example-big-bang` calls the `vendor-big-bang-base` recipe which pulls down the remote tagged version of BB which uses `private-registry` as the pull secret. 

My questions are:
  1. Should we be checking in vendored code to this repo?
  2. If we're checking in vendored code so that we can modified values like this do we need that make recipie that re-pulls with every build?

RothAndrew: we can't modify anything in the `.../vendor` folder since it will get blown away by the build process.

RothAndrew: Given that Flux is not able to pull private bases we need to vendor the code, unless some other solution is come up with

jeff-mccoy: I may need some help understanding this one.  Big bang has a value that will let you define what the image pull cred is last time I checked, am I missing something?  I'm also super confused by how complex our flux manifest section is in general for the BB examples.

examples/tiny-kafka/manifests/kafka.yaml
jeff-mccoy: Yeah I think so, we're already testing no namespace defined in basically every example.  A test with a package that creates and upgrades a package with a namespace defined in the manifests would probably be good

examples/big-bang/template/bigbang/vendor/bigbang/base/flux/kustomization.yaml
jeff-mccoy: Okay I have a solution for this and will be working tonight / tomorrow to wrap it up for both bb example packages.  Will update when it's done.

jeff-mccoy: this is resolved now with the addition of kustomization sources + serviceaccount updates to use the zarf-registry on zarf+helm deploy

jeff-mccoy:
Updated the PR reqs to remove manual tests for appliance mode for examples already covered by E2E now.

jeff-mccoy:
Deployed on Mac M1 with KIND (game example)

Landing page
Screen Shot 2022-02-01 at 3 09 40 AM
    
In game
Screen Shot 2022-02-01 at 3 10 25 AM

jeff-mccoy:
Deployed on Mac M1 with KIND (data injection example)

Screen Shot 2022-02-01 at 3 26 09 AM

jeff-mccoy:
Deployed on Mac M1 with KIND (gitops data example)

YrrepNoj:
tiny-kafka demo deployed on Intel Mac with KIND 
Screen Shot 2022-02-01 at 10 11 08 AM

YrrepNoj:
Postgres demo deployed on Intel Mac with KIND


Screen Shot 2022-02-01 at 10 32 27 AM

YrrepNoj:
BigBang Single package (twistlock) demo deployed on Intel Mac with KIND

Screen Shot 2022-02-01 at 10 40 42 AM

assets/charts/registry-values.yaml
```diff
@@ -13,8 +13,8 @@ service:
   nodePort: "###ZARF_REGISTRY_NODEPORT###"
 resources:
   requests:
-    cpu: "100m"
-    memory: "512Mi"
+    cpu: "500m"
+    memory: "256Mi"
   limits:
-    cpu: "1"
+    cpu: "3"
```
YrrepNoj, line 19: That's so much cpu!! 😮

RothAndrew:
Working on updating the readme for the Game example. Gonna just pull out the asciinema graphics out for now and add a new backlog item to add them back in once we can get them updated.

jeff-mccoy:
@YrrepNoj you'll probably want to sanity check my rebase/merge of the 7bc2e323754a1af8417616207407f92f1b26161b.  Also @RothAndrew 7bc2e323754a1af8417616207407f92f1b26161b is a good example of how the e2e tests have changed in this PR.

RothAndrew:
BB example working in Vagrant+K3s. Still troubleshooting issues in KinD. 

image

image

RothAndrew:
@jeff-mccoy I noticed K3s still has Traefik. Can we turn that off so it consistently relies on the user to provide an ingress controller? Otherwise I can't go in and change Istio back to 443 in the BB example

jeff-mccoy:
> @jeff-mccoy I noticed K3s still has Traefik. Can we turn that off so it consistently relies on the user to provide an ingress controller? Otherwise I can't go in and change Istio back to 443 in the BB example

does traefik bind to 443 by default?  I have no problem removing from appliance mode except that it comes out of the box on other k3s/k3d installs so we can't assume it won't be in the way.

RothAndrew:
I think the difference is we are in control of the K3s install. We have to just document the need for (or absence of) an ingress controller if Zarf is being used against a cluster that it didn’t make.

jeff-mccoy:
Copy, have you verified that the default traefik install is actually conflicting with istio and not just how we were using it before?

RothAndrew:
No, I assumed it would. Haven’t confirmed.

RothAndrew:
ImagePullBackOff when attempting to use the data-injection example on KinD
image

jeff-mccoy:
> ImagePullBackOff when attempting to use the data-injection example on KinD. Looks like it is missing the pull secret in the pod spec image

that seems like a bug, it should be injected via the serviceaccount mutation

YrrepNoj:
> > ImagePullBackOff when attempting to use the data-injection example on KinD. Looks like it is missing the pull secret in the pod spec image
> 
> that seems like a bug, it should be injected via the serviceaccount mutation

Looking into this now. It seems like it's working perfectly fine on Linux KIND but I get the same issue on MacOS KIND. Rebuilding everything now to make sure that's actually what's happening.

cli/internal/helm/post-render.go
```diff
@@ -0,0 +1,235 @@
...
+			}
+		}
+
+		// Keep the default SAs up-to-date so they will use the zarf-registry pull secret for the namespace
+		if defaultSvcAccount, err := k8s.GetServiceAccount(name, corev1.NamespaceDefault); err != nil {
+			return nil, fmt.Errorf("unable to get service accounts for namespace %s", name)
+		} else {
+			// Look to see if the service account is already patched
+			if defaultSvcAccount.Labels[managedByLabel] != "zarf" {
+				updateSvcAccount(defaultSvcAccount)
```
YrrepNoj, line 203: We should be catching and checking the error return of the `updateSvcAccount`. The reason why our data-injection example was sometimes getting a ImagePullBackoff is because the ServiceAccount wasn't able to update because of a race condition where the SA gets updated before we're able to add our special sauce. Giving the error of: ` DEBUG   unable to update the default service account for the demo namespace: Operation cannot be fulfilled on serviceaccounts "defa
ult": the object has been modified; please apply your changes to the latest version and try again`

I'm not sure why this was only happening for KIND clusters but I'm happy enough just waving that off to the implementation differences between the distros and what they do behind the scenes when a namespace is just created.

I propose breaking this logic out to a function so we can through a simple retry-loop around it. I'll make a separate PR into this branch to do that so @jeff-mccoy can either merge it in if he agrees with the solution or come up with his own fix.

RothAndrew:
Whelp, that didn't do what I expected it to do... I clicked the "Update Branch" button here in the PR on GitHub, and chose the "Rebase" option. It looks like it added me as an author to all the commits, which invalidated their signatures.

@jeff-mccoy opinions on how to fix? When we merge if we squash it will still result in a signed commit to master. Or we could squash and force-push a signed commit to the branch, or... something else? Idk.

cli/internal/packager/seed.go
```diff
@@ -0,0 +1,247 @@
...
+		inject.command = "docker"
+		inject.args = []string{"load", "-i", tempPath.seedImages}
+
+	case k8s.DistroIsMicroK8s:
+		state.Registry.SeedType = config.ZarfSeedTypeCLIInject
+		inject.command = "microk8s"
+		inject.args = []string{"ctr", "images", "import", tempPath.seedImages}
+
+	default:
+		state.Registry.SeedType = config.ZarfSeedTypeRuntimeRegistry
```
YrrepNoj, line 164: I could use a big more context about when/what this default state means. When we set this as the `state.Registry.SeedType` I think we are going to be running into one of our own Fatals downstream and I want to try to understand the intended logic a bit more.

@jeff-mccoy

jeff-mccoy:
I need to rebase and force push to clean this up I think.  I don't want to squash this one anymore than we have to, I've already squashed 200 or so commits on this branch.

cli/internal/packager/seed.go
YrrepNoj: I'm not too worried about it (for this PR), but the currently flow will lead to a Fatal in `cli/internal/message/tls/prompts.PromptAndValidateHost()` since the `config.TLS.Host` will be an empty string but the `confirmed` flag will ALWAYS true because the user already went through the interactive confirm step.

examples/tiny-kafka/README.md
```diff
@@ -1,16 +1,157 @@
...
+
+&nbsp;
+
+### Credits
+
+✨ Special thanks to these fine references! ✨
+
+- https://www.reddit.com/r/programming/comments/nap4pt/dos_gaming_in_docker/
+
+- https://earthly.dev/blog/dos-gaming-in-docker/
```
RothAndrew: don't need this stuff, these refs are for the game example

RothAndrew (changes requested):

examples/tiny-kafka/README.md
jeff-mccoy: haha yeah I had already removed it, just pushed it up.

jeff-mccoy: too much copy/paste fun

jeff-mccoy:
Data Injection example

Screen Shot 2022-02-07 at 10 17 28 PM

jeff-mccoy:
omg...I think we're there.......

RothAndrew (approved):

PR #238, opened 2022-01-21
Write E2E test for Gitea and Grafana

RothAndrew:
## What/Why
* Write E2E test that ensures Gitea and Grafana come up cleanly
* Fix a lint error that pre-commit was picking up

### DIFF
--- a/cli/internal/packager/create.go
+++ b/cli/internal/packager/create.go
@@ -75,7 +75,7 @@ func addLocalAssets(tempPath componentPaths, assets config.ZarfComponent) {
 	if len(assets.Charts) > 0 {
 		logrus.Info("Loading static helm charts")
 		_ = utils.CreateDirectory(tempPath.charts, 0700)
-		re := regexp.MustCompile("\\.git$")
+		re := regexp.MustCompile(`\.git$`)
 		for _, chart := range assets.Charts {
 			matched := re.MatchString(chart.Url)
 			if matched {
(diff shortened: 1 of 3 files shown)


RothAndrew:
Ready for review & merge

cli/internal/packager/create.go
```diff
@@ -75,7 +75,7 @@ func addLocalAssets(tempPath componentPaths, assets config.ZarfComponent) {
 	if len(assets.Charts) > 0 {
 		logrus.Info("Loading static helm charts")
 		_ = utils.CreateDirectory(tempPath.charts, 0700)
-		re := regexp.MustCompile("\\.git$")
+		re := regexp.MustCompile(`\.git$`)
```
YrrepNoj, line 78: Was this breaking before? I almost expect to see some other changes to how `.git` files are being written when looking at this change.

RothAndrew: `pre-commit` picked it up as a lint error. It worked fine before, and should still work fine, just without throwing a lint error.

YrrepNoj:
Is there a particular reason why we're commenting out old sections of the e2e.sh script instead of just remove sections that have been successfully migrated?

RothAndrew:
> Is there a particular reason why we're commenting out old sections of the e2e.sh script instead of just remove sections that have been successfully migrated? Not totally against it, just curious if we're expecting to do something with it later.

I could remove them entirely if you feel strongly about it. Since the whole file is destined to be deleted I didn't see it as a big deal, and having the commented out code helps me keep track of what we already did and what we still have to do.

YrrepNoj:
> 

I don't feel that strongly about it. I would argue it's easier to see what we still need to do if we remove it and just look at what's left but at the end of the day, you're right that the file is getting deleted and it doesn't matter.

YrrepNoj (approved):

PR #252, opened 2022-01-31
Fix issue where user will still be prompted for components if the --confirm flag was provided

### LINKED ISSUE #200: `--confirm` not respected when deploying zarf package
**What I saw:**

If I deploy the Big Bang example, which has an optional component called `kubescape`, it prompts me to choose whether I want Kubescape or not

**What I expected:**

If I run `zarf package deploy foobar.tar.gz --confirm` I expect that it would never stop and prompt me to interact, because I used the `--confirm` flag. It would either make a choice itself or fatal out if it cant. If the `--components` flag was not used I'm thinking it should use the defaults that are set for each component.

YrrepNoj:
- [x] Write explicit test to verify we can deploy a package without being prompted when the confirm flag is provided


RothAndrew:
@YrrepNoj if I'm understanding correctly just the change in c3e138d37141941a213787ebb377816f9d9b43c7 is needed? Can we open an new PR against Jeff's branch and get that one line changed?

RothAndrew:
I believe this PR will become OBE if #271 gets merged. If so this can be closed.

PR #269, opened 2022-02-04
update post-render to retry updating service account on error

YrrepNoj:
We experienced a race condition when updating the ServiceAccount for a newly created namespace.
Sometimes KIND distro would  error because it was still updating the SA at the same time we were.

Without this change we were sometimes seeing the following error on KIND clusters:
` DEBUG   unable to update the default service account for the demo namespace: Operation cannot be fulfilled on serviceaccounts "default": the object has been modified; please apply your changes to the latest version and try again`

I've tested this several times (more than 5..) locally (macOS and Linux) and have not run into the ImagePullBackoff because of an un-updated ServiceAccount.

### DIFF
--- a/cli/internal/helm/post-render.go
+++ b/cli/internal/helm/post-render.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/defenseunicorns/zarf/cli/config"
 	"github.com/defenseunicorns/zarf/cli/internal/k8s"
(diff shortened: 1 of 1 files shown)


YrrepNoj:
I'm not a fan of naive retry loops but I couldn't find any way to check if KIND was done updating the SA before we try to add our keys and label.

Any thoughts? @RothAndrew @jeff-mccoy ?

RothAndrew:
It's definitely a bandaid fix, but I don't have any objections at this point in time.

jeff-mccoy:
I'll take a look in a couple hours if that's okay--I might have an idea.

jeff-mccoy:
@YrrepNoj I am unable to reproduce the original issue so having a hard time validating this.  When you get a chance could you try this instead?  This is how the K8s E2E tests do it and I think is a bit cleaner than a retry loop.  Thanks!

YrrepNoj:
@jeff-mccoy It doesn't seem to be happening nearly as often but I got the error message after about 10-15 attempts
`DEBUG   unable to update the default service account for the demo namespace: Operation cannot be fulfilled on serviceaccounts "default": the object has been modified; please apply your changes to the latest version and try again
 WARNING  Unable to update the default service account for the demo namespace`

My assumption is that KIND will create a blank/minimalistic ServiceAccount and then create a background job to update it with some metadata or something. Since this package is a lot more lightweight than any of the other packages I guess it just happens to run into this issue.

I got this result on a Ubuntu 18.04 machine with an Intel CPU. Zarf binary was built with version tag ` v0.14.0-99-g58c7321` and `zarf package inspect ` on the data-injection package shows it was built from the same version.

jeff-mccoy:
Sorry which package is doing this?

YrrepNoj:
data-injection

YrrepNoj:
@jeff-mccoy all cleaned up ;)

PR #271, opened 2022-02-04
Fix issue where `--confirm` flag isn't always respected.

### LINKED ISSUE #200: `--confirm` not respected when deploying zarf package

RothAndrew:
I *think* this can take the place of #252 and will fix #200 once #237 is merged

### DIFF
--- a/cli/internal/packager/common.go
+++ b/cli/internal/packager/common.go
@@ -4,7 +4,6 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
-	"github.com/defenseunicorns/zarf/cli/types"
 	"io"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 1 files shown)


RothAndrew:
@YrrepNoj can you confirm this is all you think is needed to fix #200?

YrrepNoj:
@RothAndrew Sorry what was the point in making this instead of #252? 

And I believe this will fix the issue but I would like to write an e2e test just to have some form of proof for the future. I can do that today.

PR #297, opened 2022-02-08
200 confirm prompting components

YrrepNoj:
- [x] Check if the confirm flag has been set before prompting for components
- [x] Update interactive confirm to not write response to config *
- [x] Write explicit test to verify we can deploy a package without being prompted when the confirm flag is provided

* This is so we can read the config.confirm value later to determine if the command is being run in 'script' mode.
This gives us the ability to be confident if/when we can prompt the user with a question.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,10 @@ package-example-single-big-bang-package: ## Create the Zarf package for single-b
 package-example-gitops-data:
 	cd examples/gitops-data && ../../$(ZARF_BIN) package create --confirm && mv zarf-package-* ../../build/
 
+.PHONY: package-example-tiny-kafka
+package-example-tiny-kafka:
+	cd examples/tiny-kafka && ../../$(ZARF_BIN) package create --confirm && mv zarf-package-* ../../build/
+
 .PHONY: test-cloud-e2e-example-game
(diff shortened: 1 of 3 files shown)


.github/workflows/test-command.yml
```diff
@@ -507,6 +507,7 @@ jobs:
           export GOPATH="$HOME/go"
           export PATH="$PATH:$GOPATH/bin"
           chmod +x build/zarf
+          ./build/zarf tools registry login registry1.dso.mil --username "${{ secrets.REGISTRY1_USERNAME_ZARF_ROBOT }}" --password "${{ secrets.REGISTRY1_PASSWORD_ZARF_ROBOT }}"
```
YrrepNoj, line 510: The e2e general-cli step is failing because it's unable to pull down a Docker image from registry1 when building the tiny-kafka package but I thought this would should have resolved that 😕

RothAndrew: Did you just add this line?

YrrepNoj: Oh is this one of those cases where it has to be in master first.. Womp womp.

RothAndrew: GitHub Actions only look at master. If you've changed one of them you have to get that change into master before it will happen. It's definitely not ideal and something we should look more closely at, but that's the way it works right now.

RothAndrew: Submit a 1-line PR with just this change. We'll get it approved and merged asap

jeff-mccoy (approved):
lgtm

Issue #174, opened 2021-11-15
Let's get our definitions straight :-)


jeff-mccoy:
As the project has matured a lot of words and definitions have surfaced that need to be sorted out so the CLI and docs and clearly communicate to users what exactly is happening and how to use Zarf.  

**Background:**
- Zarf deploys a [K3s](https://k3s.io/) cluster with some pre-configured technologies in a simple, low-resource, declarative way we have been calling **Appliance Mode**.  
  - As the #153 work finishes, Zarf will also support deploying to non-k3s clusters, but I believe we will always lean on K3s as the defacto cluster tech used for this mode.
- The **Utility Cluster** is actually an **Appliance Mode** deployment done by the `zarf init` command today.  This is where it gets interesting--the native apply work will allow for "HA Utility Clusters" so it might not be an actual **Appliance Mode** deployment for the **Utility Cluster**, but some HA system (pick your K8s distro).
- We've determined we're going to not advertise/broadly support a 3rd use-case for K8s here, where there is no **Appliance Mode** or **Utility Cluster** (not possible anyway until the native apply work is done).  The Zarf solutions will always be either 
  - Moving simple files for things like IaC (terraform and friends)
  - **Appliance Mode**
  - **Utility Cluster**
- I started using the term **Gitops Service** as we moved the registry into every zarf appliance,  that's actually going to change a little with native apply.  There will always be an _embedded registry_ for appliance mode (system service outside K8s) and a second registry for **Utility Cluster** / **Gitops Service**, this will run on K8s and also support HA as an option.

**Other Notes:**
- I find the word embedded / edge to be far too abused to use for naming anything in zarf.  Not everyone loves **Appliance Mode**, looking at you @unicornbunny182.  But we need a word / phrase to communicate something that is long-lived, lightweight, single-purpose, easy to recreate / redeploy and fully declarative for the airgap environment 

**Maybe??:**
- One phrase we could use instead of **Appliance Mode** could be **Zarf Agent** instead.  Thinking here would be it's a special thing that Zarf deploys, and that it continuously runs (i.e an agent).  We could describe variations, **Utility Cluster**, as "agent features".  So the **Zarf Agent** could have custom apps/features/thingys (e.g. Doom) or otherwise **Gitops Service** or similar.

unicornbunny182:
I just hate "special words". If we can avoid making up our own words, would be easier to explain.  I know there isn't a "standard", but clearly other's have looked at "tiny edge" before and come up with some words to describe.

[...]

PR #298, opened 2022-02-08
New Example - Istio with separately loaded TLS cert

### LINKED ISSUE #120: Lightweight Big Bang Core example that expects Istio's cert to already be present

RothAndrew:
Fixes #120


RothAndrew:
Ready for review

.pre-commit-config.yaml
```diff
@@ -28,13 +29,13 @@ repos:
     rev: v0.4.0
     hooks:
       - id: go-fmt
-## Normally we wouldn't need to do a local hook but we need to modify the shell script that gets run to first change directories into the `cli` folder
+  ## Normally we wouldn't need to do a local hook but we need to modify the shell script that gets run to first change directories into the `cli` folder
```
jeff-mccoy, line 32: can you clarify why we need to cd into cli when the go.mod is in the root folder now?

examples/istio-with-separate-cert/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+> NOTE:
+>
+> All subsequent commands should be happening INSIDE the Vagrant VM
+
+### Initialize Zarf
+
+```shell
+# Initialize Zarf
+./zarf init --confirm --components k3s,gitops-service
```
jeff-mccoy, line 57: This assumes appliance mode, is that the only support mode for this example?  I'd much rather not lean on makefiles, vagrant, appliance mode when doing any of the examples as I think it mixes concerns.

examples/istio-with-separate-cert/files/bigbangdev.cert
```diff
@@ -0,0 +1,91 @@
+-----BEGIN CERTIFICATE-----
```
jeff-mccoy, line 1: Why are we committing these to git if they are also loaded via a maefile?  I think it should be one or the other.

examples/istio-with-separate-cert/zarf.yaml
```diff
@@ -0,0 +1,67 @@
+kind: ZarfPackageConfig
```
jeff-mccoy, line 1: I do really like that this example shows as little as possible to highlight the istio/secret stuff.

.pre-commit-config.yaml
RothAndrew: IDK. It may not be necessary any more. This change was an automated format of the file, I didn't actually change any of this behavior. It would take more more time to go in and figure out whether this is necessary or not. Recommend putting this off for now and adding it to #309

examples/istio-with-separate-cert/README.md
RothAndrew: It's possible that this example would work in KinD, but most Big Bang examples do not due to issues with Elasticsearch. I would rather leave this using the same pattern for now rather than trying to do something different.

Side-note: What is your definition of Appliance Mode? I thought it was "appliance mode" when you don't use the gitops-service.

examples/istio-with-separate-cert/files/bigbangdev.cert
RothAndrew: It would require adding a dependency on [yq](https://github.com/mikefarah/yq) for the user to be able to perform this example. Doing it this way keeps the dependency list the same as it was. I added a helper make target for maintainers to quickly and easily update the cert.

FWIW, the Big Bang example also has the cert in the codebase, by having it in the `values.yaml` file.

RothAndrew:
@jeff-mccoy ready for re-review

examples/istio-with-separate-cert/README.md
jeff-mccoy: okay thanks and yes we need to sort out #174 pretty soon too I think.  I've used "appliance mode" in multiple ways and generally it was something without the gitops service, but have recently been using it for the version of zarf that also deploys k3s since we control all of it--but that's probably not totally the right term.

jeff-mccoy (approved):

PR #299, opened 2022-02-08
sign into registry1 for general cli e2e test

YrrepNoj:


RothAndrew (approved):

RothAndrew:
Using admin override to merge without tests, since this change is to the test pipeline itself, which only uses the master branch to get its config

PR #300, opened 2022-02-08
296: Using /usr/sbin instead of /usr/local/bin to accommodate RHEL

### LINKED ISSUE #296: Unable to inject seed image on RHEL host.
This is likely because RHEL $PATH doesn't include /usr/local/bin by default and that's where we currently put the k3s executable. 

```
WARNING  Unable to inject the seed image from the /tmp/zarf-1741909834/seed-images.tar archive
  DEBUG   exec: "k3s": executable file not found in $PATH
└ (/Users/jon/projects/du/zarf/cli/internal/packager/seed.go:179)
```

This can be resolved by moving the binary to `/sbin` or `/usr/bin`. Additionally, moving where the binary gets placed would mean updating the cleanup script, service, and zarf.yaml to reference the new path.

YrrepNoj:
Closes #296

### DIFF
--- a/assets/scripts/k3s.service
+++ b/assets/scripts/k3s.service
@@ -24,4 +24,4 @@ RestartSec=5s
 ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service'
 ExecStartPre=-/sbin/modprobe br_netfilter
 ExecStartPre=-/sbin/modprobe overlay
-ExecStart=/usr/local/bin/k3s server --write-kubeconfig-mode=700 --disable traefik
+ExecStart=/usr/sbin/k3s server --write-kubeconfig-mode=700 --disable traefik
--- a/assets/scripts/zarf-clean-k3s.sh
+++ b/assets/scripts/zarf-clean-k3s.sh
@@ -78,8 +78,8 @@ fi
 rm -f /etc/systemd/system/k3s.service
 
 for cmd in kubectl crictl ctr; do
-	if [ -L /usr/local/bin/$cmd ]; then
-		rm -f /usr/local/bin/$cmd
+	if [ -L /usr/sbin/$cmd ]; then
+		rm -f /usr/sbin/$cmd
 	fi
 done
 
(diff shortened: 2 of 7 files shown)


RothAndrew:
According to [this](https://askubuntu.com/a/308048) I'm concerned about putting stuff in `/sbin`. Is there a better directory we could pick? From reading that post, it sounds like `/usr/sbin` would be a better choice.

jeff-mccoy:
I don't think it matters much if you remember this is for appliance mode only, k3s will be a boot item as well.  If /usr/sbin is always in the rhel path (note that link was for Ubuntu, which doesn't have this issue), then that's probably fine.  I'm not certain that's the case but am fairly certain that /sbin or /bin always is regardless of the os. We'd probably want to verify running as root vs sudo too since this is really an env variable issue. 

Alternatively, we could just add /usr/local/bin to the path as an env variable on exec instead.

RothAndrew:
> I don't think it matters much if you remember this is for appliance mode only, k3s will be a boot item as well. If /usr/sbin is always in the rhel path (note that link was for Ubuntu, which doesn't have this issue), then that's probably fine. I'm not certain that's the case but am fairly certain that /sbin or /bin always is regardless of the os. We'd probably want to verify running as root vs sudo too since this is really an env variable issue.
> 
> Alternatively, we could just add /usr/local/bin to the path as an env variable on exec instead.

My assumption is that the `rhel8` vagrant box that we have set up is representative of RHEL in general. Given that assumption, RHEL does include `/usr/sbin` by default in the PATH. Here's a screenshot:

image

Given this. I propose we go with `/usr/sbin` instead of `/sbin`. I do think it would work either way, but `/usr/sbin` feels more aligned with the stated intention of these different directories in Linux.

jeff-mccoy:
Works for me

YrrepNoj:
Funnily enough (at least on the Vagrant rhel8) `/sbin` simply symbolically links to `usr/sbin`. Changed  to `usr/sbin`.

RothAndrew (approved):

jeff-mccoy (approved):

PR #301, opened 2022-02-08
208 basic composability

mike-winberry:


RothAndrew:
@mike-winberry Given the decision to move the Software Factory package to its own repo, I believe this should get the same treatment. I propose we make a repo https://github.com/defenseunicorns/zarf-package-flux and add this code to that repo.

Thoughts?

/cc @jeff-mccoy @YrrepNoj

jeff-mccoy:
I think it's find to move it to it's own repo once this PR is merged--but I do have a concern with a whole bunch of packages sitting at the top-level of the Defense Unicorn group.  Maybe we need to discuss a new org something for all these repos?

RothAndrew:
If we're just gonna move it I'm not sure why we would merge this PR, just to pull it back out.

Do we want to wait on a decision here until we are able to make a decision on what org these go in?

mike-winberry:
That works for me. I’m gonna play with a repo tomorrow and see if we can push to GitHub packages with a sha using GitHub ci to make it easy to pull with validation. I’ll also push what I have to a new repo on DU as it really doesn’t need to be in zarf.

packages/Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+PACKAGE_PATH :=  $(shell cd $(name) && pwd)
+
+.PHONY: default 
+default: help
+
+.PHONY: package
+package:
+ifdef name
+	@cd $(name) && $(ZARF_BIN) package create --confirm
```
YrrepNoj, line 25: This only current works if $(ZARF_BIN) is somewhere in $PATH. I don't think we make that assumption in our other Makefiles.

mike-winberry: No and that can change now that the pr is done so I’ll update that as well.

mike-winberry: As soon as we update the zarf version it can be done the way the others are.

mike-winberry: One issue I have is in the examples (which this wouldn't be) we assume they have cloned the repo. If we are building composable packages we should be assuming they are using a release rather than using the github repo which uses another make command to build and place zarf into a created folder called ../sync

mike-winberry: but it only does that if you use the make command within the examples folder otherwise it places it in build which is relative to the path that exists for the repo and not anything related to the os. I think its a way more resonable assumption that if a user wants to use zarf to build composed packages then they would have zarf. I am open to other suggestions though.

mike-winberry: Similar to how when running helm it is expected that it exists within the os path otherwise it will not work.

RothAndrew: Take a look at the way I set it up in the zarf-package-software-factory repo. Not saying that's the way we always have to do it but it should give some ideas. In there I just added documentation that having `zarf` installed and present in PATH a prerequisite.

mike-winberry: @jeff-mccoy which if fair, but zarf isn't native to itself unless itself exists. 

I did add it to the pre-requisites in the readme. 

I also don't think that this will be the way we build them long term. This was just a way to provide the initial behavior without relying on a registry like artifact hub at this point. I think we could eliminate most of this once that is the case. Though that will still have the pre-requisite that zarf exists and you are able to use it regardless of where it exists in its environment.

RothAndrew: I can definitely see us leveraging OCI Artifact spec for this in the future.

mike-winberry: Yeah, I think if you run it how it is it definitely shows that idea of replacing locally existing packages with a sha and url. Which I’m hoping will be possible as a more useful mvp than what I currently have. Because then those actions could always change to point to any package registry.

jeff-mccoy:
@mike-winberry and I are going to sync tomorrow on this PR so I can take review responsibility for that.

mike-winberry:
Closing to move to new repo. https://github.com/defenseunicorns/zarf-package-flux

PR #305, opened 2022-02-09
Normalize trace references for debug/trace output

jeff-mccoy:
Depending on what functions are called for debug/trace the wrong callstack offsets could occur creating a problem where the wrong debug point is listed (sometimes within a library making it difficult to understand the issue)


YrrepNoj (approved):
lgtm

PR #306, opened 2022-02-09
Zarf agent mutating webhook

jeff-mccoy:
Initial mutating webhook work.


RothAndrew:
Is this the highest priority right now? I thought we were going to be spending time making the testing and documentation more robust and paying down tech debt now that the big PR is merged. This feels like jumping right back to feature development.

RothAndrew:
Before we get too far we also need to get an issue written up in the backlog for this work so we can have a clear idea on when the Definition of Done is met and it can be merged.

jeff-mccoy:
I'm not working this week due to Covid, so this is how I am spending my time.  And yes it is important to resolve unfinished work around the new postRender behavior and operators / Big Bang. We discussed before the bandaid of manually patching image repos in Big Bang and fixing it right after #237 was merged, this is that fix.

RothAndrew:
> I'm not working this week due to Covid, so this is how I am spending my time. And yes it is important to resolve unfinished work around the new postRender behavior and operators / Big Bang. We discussed before the bandaid of manually patching image repos in Big Bang and fixing it right after #237 was merged, this is that fix.

Copy. That makes sense. Can we get together and write out a new GitHub issue for this? It will help us keep the scope small.

jeff-mccoy:
Yeah for sure, I'll ping you later today if you're free. I was going through old issues last night but still need to make new issues for things we didn't get into #237 that will be needed or desired to fully realize everything that #237 provided.

.github/workflows/build-zarf-agent.yml
```diff
@@ -12,29 +12,72 @@ on:
         default: "master"
 
 jobs:
+<<<<<<< HEAD
```
YrrepNoj, line 15: @jeff-mccoy looks like GitKraken has failed you again :(

Makefile
```diff
@@ -58,6 +58,11 @@ init-package: ## Create the zarf init package, macos "brew install coreutils" fi
 
 build-test: build-cli init-package ## Build the CLI and create the init package
 
+publish-image:
+	docker buildx build --push --platform linux/arm64/v8,linux/amd64 --tag defenseunicorns/zarf-agent:${TAG} .
```
YrrepNoj, line 62: This `${TAG}` is from an env right? When it fails right now there's no clear message that you need to export some. value to `TAG` unless you go look at the Makefile. Is an error message worth it to you?

jeff-mccoy: It's a makefile arg and already passed in the manual trigger?

.github/workflows/build-zarf-agent.yml
jeff-mccoy: Ughhhh it totally did...

RothAndrew (commented):
Nothing jumped out at me when I looked through everything. Seeing so many deleted lines in the Big Bang example's values.yaml made me ridiculously happy.

I'm assuming you did manual testing of this with the Big Bang example? since we don't have that one running in the testing pipeline. I'm happy to test with it as well in the morning if you want, though I'm also just fine with taking your word for it if you've already been testing that 😄 

Did you by any chance try DI2-ME out with this build? It would be a pain, but it is as far as I know the only example of a package that (a) uses Big Bang and (b) tells Flux to watch a git repo for changes.

examples/big-bang/zarf.yaml
```diff
@@ -13,88 +13,68 @@ components:
...
-      - https://repo1.dso.mil/platform-one/big-bang/apps/core/monitoring.git@30.0.1-bb.8
-      - https://repo1.dso.mil/platform-one/big-bang/apps/security-tools/twistlock.git@0.4.0-bb.0
+      - https://repo1.dso.mil/platform-one/big-bang/bigbang.git@1.33.0
+      - https://repo1.dso.mil/platform-one/big-bang/apps/core/istio-controlplane.git@1.13.2-bb.1
+      - https://repo1.dso.mil/platform-one/big-bang/apps/core/istio-operator.git@1.13.2-bb.1
+      - https://repo1.dso.mil/platform-one/big-bang/apps/core/jaeger.git@2.29.0-bb.1
+      - https://repo1.dso.mil/platform-one/big-bang/apps/core/kiali.git@1.49.0-bb.0
+      - https://repo1.dso.mil/platform-one/big-bang/apps/core/cluster-auditor.git@1.4.0-bb.2
+      - https://repo1.dso.mil/platform-one/big-bang/apps/core/policy.git@3.7.1-bb.0
+      - https://repo1.dso.mil/platform-one/big-bang/apps/sandbox/loki@3.0.1-bb.1
```
YrrepNoj, line 28: @jeff-mccoy The read-only user is getting errors for loki and promtail because they're the only two repos listed that don't have `.git` at the end of the link/before the `@{VERSION_TAG}`.

The code for the read-only user assumes .git is in the repo string provided and since it wasn't there the strings.split() call gave the wrong repo name to add the read-only user to.

When I originally wrote this, I was assuming `.git` in the repo name was a 'required' to the schema since that's how we have it everywhere else but obviously it doesn't look like that's that a checked requirement. Should it be?

jeff-mccoy: Interesting I think we already prescribe a unique format anyway so yes it should be, we should open an issue for a new validator for git urls, excellent find!

src/Makefile
```diff
@@ -11,6 +11,7 @@ CLI_VERSION := $(if $(shell git describe --tags), $(shell git describe --tags),
 build:
 	mkdir -p ../build
 	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="-s -w -X 'github.com/defenseunicorns/zarf/src/config.CLIVersion=$(CLI_VERSION)'" -o ../build/zarf main.go
+	CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -ldflags="-s -w -X 'github.com/defenseunicorns/zarf/src/config.CLIVersion=$(CLI_VERSION)'" -o ../build/zarf-arm main.go
```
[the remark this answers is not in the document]
jeff-mccoy: fine

packages/zarf-agent/zarf.yaml
```diff
@@ -0,0 +1,20 @@
+kind: ZarfPackageConfig
+metadata:
+  name: "init-package-zarf-agent"
+  description: "Install the zarf agent mutating webhook on a new cluster"
+
+components:
+  - name: zarf-agent
+    required: true
+    images:
+      - "defenseunicorns/zarf-agent:v0.18.3-8"
```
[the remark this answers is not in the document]
jeff-mccoy: templates don't run against image paths last I checked?

src/internal/agent/operations/patch.go
```diff
@@ -0,0 +1,61 @@
+package operations
+
+const (
+	addOperation     = "add"
```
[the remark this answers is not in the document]
jeff-mccoy: We need to decide in a styleguide because we're inconsistent, but uppercase or all caps would be exported, which we don't want here.  Uber has some thoughts on this too:  https://github.com/uber-go/guide/blob/master/style.md#prefix-unexported-globals-with-_

packages/zarf-agent/zarf.yaml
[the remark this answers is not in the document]
YrrepNoj: You are correct, I think at the moment it's only run at deploy and post-render operations. This comment was more broaching the question of if templating out things in the zarf.yaml during package creation was worth writing an issue for.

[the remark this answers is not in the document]
jeff-mccoy: Yeah there's already a semi-related task in #386 around templating the package arch.  We could just make that a separate issue as well.  But that's out of scope for this PR me thinks.

jeff-mccoy:
BB Core on 1.33 w/Loki vs EFK is happy w/Zarf agent.

Screen Shot 2022-05-17 at 6 47 21 PM

Screen Shot 2022-05-17 at 6 47 34 PM

YrrepNoj (approved):
Changes and updates look good to me 👍

jeff-mccoy:
We have a winner.  Fixed the fun BB issue with the gatekeeper.

Screen Shot 2022-05-19 at 1 11 26 AM

YrrepNoj (approved):
Thanks for checking up and updating on the gatekeeper stuff. LGTM!

PR #316, opened 2022-02-11
Check for a $KUBECONFIG env-variable

### LINKED ISSUE #315: Zarf doesn't consider $KUBECONFIG env when running commands
## Summary
Now that Zarf can support deploying onto existing clusters, we should be considering that users local $KUBECONFIG env when determining if there is an existing cluster to deploy onto. Currently, Zarf is hard coded to look for an existing kubeconfig at `~/.kube/config`. This was perfectly fine when Zarf also was responsible for standing up its on k3s cluster but isn't as clear now.

## Steps to reproduce 
`export KUBECONFIG=~/.kube/otherconfig`
`kind create cluster`
`kubectl get pods -A`        -> Returns the pods on the newly created cluster
`zarf init --confirm`           -> Returns an error because it can't find a k8s even though the above kubectl command could

## Expected Behavior
The Zarf CLI should check if the $KUBECONFIG env is set and prioritize using it whenever possible.

YrrepNoj:
This PR adds a check for the $KUBECONFIG environment variable when Zarf is trying to generate a config to talk to a native kubernetes cluster.

Closes #315

### DIFF
--- a/cli/internal/k8s/common.go
+++ b/cli/internal/k8s/common.go
@@ -34,13 +34,18 @@ func init() {
 }
 
 func getRestConfig() *rest.Config {
-	homePath, err := os.UserHomeDir()
-	if err != nil {
-		message.Fatal(nil, "Unable to load the current user's home directory")
+	// use the KUBECONFIG context if it exists
+	configPath := os.Getenv("KUBECONFIG")
+	if configPath == "" {
+		// use the current context in the default kubeconfig in the home path of the user
+		homePath, err := os.UserHomeDir()
+		if err != nil {
+			message.Fatal(nil, "Unable to load the current user's home directory")
+		}
+		configPath = homePath + "/.kube/config"
 	}
 
-	// use the current context in kubeconfig
-	config, err := clientcmd.BuildConfigFromFlags("", homePath+"/.kube/config")
+	config, err := clientcmd.BuildConfigFromFlags("", configPath)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


jeff-mccoy:
It would be swell to see some type of test for this.  Like maybe move the k3s kube config and set the env variable.

cli/internal/k8s/common.go
```diff
@@ -34,13 +34,18 @@ func init() {
...
-	homePath, err := os.UserHomeDir()
-	if err != nil {
-		message.Fatal(nil, "Unable to load the current user's home directory")
+	// use the KUBECONFIG context if it exists
+	configPath := os.Getenv("KUBECONFIG")
+	if configPath == "" {
+		// use the current context in the default kubeconfig in the home path of the user
+		homePath, err := os.UserHomeDir()
+		if err != nil {
+			message.Fatal(nil, "Unable to load the current user's home directory")
```
RothAndrew, line 43: What's the latest on doing this vs returning `error`? I thought at some point I had heard @jeff-mccoy say we were switching to returning `error` to make testing easier.

Though, I'm seeing right below this it doing the same thing, so maybe the question is whether a refactor of this method is warranted along with this change.

jeff-mccoy: we are moving towards bubbling errors, but that would be a larger refactor than this code right now, so I think this one is okay.

jeff-mccoy (approved):
lgtm

PR #317, opened 2022-02-11
Bug: Fetch release now depends on clean. It fails when the sync folde…

AustinAbro321:
Was going through the zarf big bang example and noticed this! Here is an example of what happens when you try to run the make  fetch-release while the sync folder isn't created. 
zarf_make_fetch_release_error

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -35,6 +35,7 @@ clean: ## Clean the sync dir
 
 .PHONY: fetch-release
 fetch-release: ## Grab the latest release as an alternative to needing to build the binaries
+	@mkdir -p sync
 	@# This probably isn't the cleanest way to get a release, but since we're moving to github, not worth adding the code until post-migration
 	@curl -fL "https://zarf-public.s3-us-gov-west-1.amazonaws.com/release/$$(git describe --tags --abbrev=0)/{zarf,zarf-mac-intel,zarf-mac-apple,zarf-init.tar.zst}" -o "sync/#1"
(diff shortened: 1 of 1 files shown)


YrrepNoj:
@AustinAbro321, thanks for pushing this. I feel like I remember running into this the first time I tried to build and then just forgot about it since I already had the sync directory afterwards.

What do you think about adding `	@mkdir -p sync` in the steps instead of executing the `clean` recipe? I worry that someone might lose all their packages just from trying to make sure they have the latest binary release. To be honest, I'm perfectly fine keeping it like this, just wanted to get your thoughts.

We also are trying to keep the commits in this repo signed and verified. Here are some of GitHubs docs on doing that. Let me know if you have any questions on this because I know these instructions aren't always the most clear. https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

RothAndrew:
I agree with Jon. I'd rather add `mkdir -p sync` to the target than add `clean` as a dependency.

AustinAbro321:
Yeah that makes sense. I hadn't realized that more gets stored in sync than just the releases. I'll configure my git to sign commits and push that change up

RothAndrew:
You'll need to squash with the last commit and force push (or make a new branch). We have GitHub configured to require ALL commits in a PR to be signed before it is able to merge.

AustinAbro321:
Change is made and commits are squashed and verified. And I learned about git rebase along the way!

YrrepNoj (approved):

Issue #101, opened 2021-10-07
Run E2E tests on multiple distros


RothAndrew:
We should run the E2E tests on more than just Ubuntu. RHEL 8 seems to be a high priority for our Navy customer, so that would be one to consider closely.

PR #319, opened 2022-02-14
E2E test optimization for multi-distro support

YrrepNoj:
This PR updates the GitHub Actions to run e2e pipeline in on the native GitHub runners (instead of needing to use chat-ops and Terratest to stand up AWS EC2 instances to perform the tests on.)

This PR enables the following:
* Modular e2e util code to easily test multiple k8s distros with the same test suite.
* Ability to run e2e test suite locally on a local cluster (will even stand up a cluster for you if one is not currently running)
* Updates the GitHub Actions pipeline to automatically run test for the KinD and K3D distros (without needing chat-ops to kick off the test)


RothAndrew:
> - The `pull_request` event simulates a merge of the pull request, and triggers Actions workflows based on the configuration and code at the merge commit. This is intended for e.g. running tests, and verifying that the code would still work if the pull request was merged. However, since the code in the pull request is potentially malicious, workflows triggered by the pull_request event are run without access to the repository’s secrets.
> - The `pull_request_target` event triggers Actions workflows based on the configuration and code at the base branch of the pull request. Since the base branch is part of the base repository itself and not part of a fork, workflows triggered by pull_request_target are trusted and run with access to secrets. This is intended for e.g. adding comments and labels to new pull requests (which requires a GitHub API token).

Actions triggered from `pull_request` can be run from forks, but they won't have secrets, so we would not be able to run Terratest things in this way, since it requires access to the AWS access and secret keys.

That just leaves annoyances like malicious PRs that mine bitcoin or just sit there spending our GHA minutes that are, while annoying, not security risks. Sounds like GitHub will require approval to run any workflows for first time contributors, but then will let them through after that.

Questions:

- @YrrepNoj how are you planning to do K3s-based testing using the new setup you've been working on?
- It looks like different distros run in parallel, but the test suite runs serially. Do have that right? If so, is that going to be sustainable as our test suite grows? You mentioned that it runs in the same amount of time now as the current suite does, but as we add more tests the test runs will get longer and longer.

RothAndrew:
- How will this handle running in different linux distros?

RothAndrew:
* Using this method, how can we handle doing tests on cloud distros like EKS? Workflows that use the `pull_request` trigger don't have access to repo secrets

RothAndrew:
I suppose we can do a combination, with part of the test suite being runnable using the `pull_request` event, and others that require access to secrets being kicked off manually like we do now.

YrrepNoj:
> * Using this method, how can we handle doing tests on cloud distros like EKS? Workflows that use the `pull_request` trigger don't have access to repo secrets

I'll look through your comments more tomorrow but I don't think this is a fully accurate statement. The test pipeline here still uses the repo REGISTRY credential secrets to sign into repo1 to build packages.

RothAndrew:
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

image

It's working because you aren't submitting the PR from a fork.

YrrepNoj:
@RothAndrew Responding the majority of your questions/observations all at once 😄 

---

> * How will this handle running in different linux distros?

This by itself won't, GitHub runners only come in Windows, Ubuntu, and macOS environments. Doing multi-linux-distro testing will require a bit more overhead to stand up other OS environments. I think one way we could do this elegantly is to use an action like ['AWS CodeBuild Run Build'](https://github.com/marketplace/actions/aws-codebuild-run-build-action-for-github-actions) to easily leverage AWS CodeBuild resources while still being able to use GitHub Actions event-based workflows.

---

> * how are you planning to do K3s-based testing using the new setup you've been working on?

I fully expect to be able to do this in the native GitHub Runner. The runners run with passwordless sudo and I was super close to getting it to work but was having issues with the $GOPATH/$GOBIN installing dependencies in the runner user path but not being able to export those env-variables when switching to root to run zarf with k3s. I think a little of tinkering should have this working though. Worst case scenario is we could use something like the above GitHub Action to run Zarf on a fully controlled AWS resource.

---

> * It looks like different distros run in parallel, but the test suite runs serially. Do have that right? If so, is that going to be sustainable as our test suite grows? You mentioned that it runs in the same amount of time now as the current suite does, but as we add more tests the test runs will get longer and longer.

You do have that all correct; There's a couple ways for use to make sure this is sustainable off the top of my head but there's plenty of conversation to be had to make sure we have a good solution. 
1. It is possible for use to create multiple kind/k3d clusters on the runner to run the tests in parallel (but that would require some overhead to make sure the `zarf connect {registry/logging/git}` commands don't error out because the port is being used by another test case). Managing resources to make sure we don't try to create too many clusters/pods on the runner would also be something we have to implement.
2. We could create a new action for each test-distro combination. This would end up looking a bit like how the terratest config is now which means the amount of yaml we have with each test drastically increases but it ensures our tests complete as quick as possible.

---

> * General comments on secrets for AWS and forked PRs

@jeff-mccoy pointed out that we can make it so outside collaborators (not just first-time contributors) need to get approval from a maintainer for their actions to run. I think that mitigates a lot of the comments and concerns.

jeff-mccoy:
Really excited to get some of these tests in.  Agreed that eventually parallel would be better for these tests but we're already getting much more just by testing a different distro besides just K3s we deployed via zarf.

RothAndrew:
Are you planning on staying with the `pull_request` event trigger? Or something else? My understanding is that even if we give the pipeline run permission to run (because it is from a fork), that it will not have access to the secrets (because it is from a fork).

RothAndrew:
If you'd like I can go create a quick PoC so we can test.

jeff-mccoy:
> If you'd like I can go create a quick PoC so we can test.

I think that would be valuable

YrrepNoj:
> Are you planning on staying with the `pull_request` event trigger? Or something else? My understanding is that even if we give the pipeline run permission to run (because it is from a fork), that it will not have access to the secrets (because it is from a fork).

Nope that's a good point. I think we'll change the trigger to `pull_request_target` 
as specified in the [github docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target).

rothandrewseviltwin:
We should test just in case, but I'm almost positive that, in an example of submitting a pull request to master, `pull_request_target` will use the latest commit on master, rather than the latest commit in the branch/fork. We could potentially use the `checkout` action to clone the branch first, then run our tests on the clone, in the same way that [the existing pipeline works](https://github.com/defenseunicorns/zarf/blob/775d66c9edb411d6f067e28c6141632ebc18e763/.github/workflows/test-command.yml#L97)

rothandrewseviltwin:
From the docs you linked:

> Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. Additionally, any caches share the same scope as the base branch. To help prevent cache poisoning, you should not save the cache if there is a possibility that the cache contents were altered.

RothAndrew:
Jon and I tried a few things.

1. We confirmed that secrets are not available to PRs from forks that use the `pull_request` event, even if they are approved.
2. We discovered that the pipeline run approval only happens once for a given user. If I want to do malicious things to this repo I can submit an innocent-looking PR from a fork and have its pipeline run approved, then push a commit that exfiltrates the repo's secrets. The 2nd pipeline will run automatically without requiring an approval. The mitigation to this is to either ensure pipeline runs do not need secrets to operate or to force a manual review before each pipeline run that needs secrets by doing a chatops command.

We can talk further about it later this afternoon when we meet to discuss the GitLab vs Keycloak thing

RothAndrew:
I reached out to CloudPosse for insight. Their `slash-command-dispatch` action is the one I copied/pasted when I built our CI framework the first time around. @osterman if you have any advice we'd welcome it.

RothAndrew:
^^^ That comment should not work, since `rothandrewseviltwin` is not a developer of this repo

YrrepNoj:
@RothAndrew @jeff-mccoy This is ready for review now.

Also, I figured out what we needed to do to test Zarf with it's self-initialized k3s cluster on a native GitHub Runner so we don't need to rely on Terratest standing up an ec2 instance for that. I currently have the slash-dispatch action still there to test the game_example on an ec2 instance but was going to remove it since we don't need that anymore. Any thoughts on that, I know we'll need it in the future when testing other packages/linux distros but I don't think we need it currently?

jeff-mccoy:
This is super good news.  I didn't realize that native github runners were actually VMs (unlike gitlab land where we were in containers).  I think for now we can just go all in on the native runners and drop the EC2 / terratest requirement and the chat-ops requirement.  Since native runners are only Ubuntu, we will need to standup self-hosted runners later on for other Linux Distros to satisfy #101, but I'm a huge fan of not having to leverage terraform for the majority of the tests.

One thing I do think we are losing that we need to discuss though is the current terratest setup does a better job of breaking the tests down into smaller identifiable chunks.  I know it's possibly more yaml soup--but I think we shouldn't lose that fidelity if can prevent it.  Thoughts?

YrrepNoj:
I'm not opposed to it, but I'm also not convinced we need that level of granularity we had before. Were you referencing the way the current tests have a different response status for each e2e test? That would lead to a large number of workflow yaml files but would solve any future issues with the test suite taking too long to run.


Another option is that we could break that down a little more and have each test run in its own step, at the potential cost of a handful more seconds of overhead per test.  Currently this PR runs all of the tests in the `Run Tests` step, the output can be seen in the [action details](https://github.com/defenseunicorns/zarf/runs/5426612819) which prints the individual test results as well as the stdout/stderr for any failing tests.

RothAndrew:
IMO we should not completely strip out the chatops and terratest framework, for 2 reasons:

1. We need the chatops for any tests that require secrets
2. We will very soon need a way to do a test on an EKS cluster

I won't hold up this PR if you guys feel differently, though I expect if we rip that stuff out we're going to end up putting some/most of it back in later. Which is fine as long as we can go pull it from the git history.

jeff-mccoy:
I personally just like the fidelity in the PR we get for showing test results here vs digging into the logs, but it's super preference.  @RothAndrew, you're probably right, we will likely need to bring back something later on unless we discover a new way to manage the secret concern--but I'd rather pull chatops until we are certain we need it and then go from there.  What are your thoughts on the test breakdown in this view vs in the logs?

jeff-mccoy:
Thinking about this a little more I think if we can get the ADR done we can table breaking out the tests more for another issue @YrrepNoj

jeff-mccoy (approved):
This is a major overhaul of the test, but nothing is too scary in the code.  I think this will end up putting us on a much better local and ci testing path.  Thanks for working this.

jeff-mccoy:
Seems a little strange to do that but keeps in line with our existing process

jeff-mccoy:
actually n/m that won't be possible 😆.  Okay I'm going to merge this.

Issue #215, opened 2021-12-16
Use Iron Bank image for Gitea


RothAndrew:
Currently our Gitea image comes from DockerHub. It needs to come from Iron Bank.

jeff-mccoy:
Closing IB stories for init package until we have a better plan, will consolidate into a new story.

Issue #214, opened 2021-12-16
Use Iron Bank images for the PLG stack


RothAndrew:
Currently we use PLG images from DockerHub. We need to get them from Iron Bank.

* `grafana/loki`
* `grafana/promtail`
* `grafana/grafana`
* `kiwigrid/k8s-sidecar`

matt-strong:
@RothAndrew for this, https://github.com/defenseunicorns/zarf/issues/213 and https://github.com/defenseunicorns/zarf/issues/215 are we looking for getting/using the current versions we are using in DockerHub into IB?

RothAndrew:
If IB has the images we need that will work for us. We should definitely make it a priority to get the latest versions of everything into IB but that doesn’t need to block this. The bigger priority is just that IB is the source for those images.

jeff-mccoy:
We actually need to discuss this some more and make a plan, we have some housekeeping to do post-mega pr on how we manage arm vs amd and IB has no arm support at all because it's super lame. I'm less worried about pgl than gitea or registry2 for init beuxse those would be used with a Big Bang example--the only place someone would care about iron bank images. And between those two--we won't be able to use iron bank at all with the new injector stuff until iron bank uses an appropriate base images.  Iron bank takes an 9 msg image tarball and makes it over 100 mega by using the wrong base image.

jeff-mccoy:
kicking this back to planned so we can sort through how we want to handle IB images for the init package.

jeff-mccoy:
Closing IB stories for init package until we have a better plan, will consolidate into a new story.

PR #329, opened 2022-02-16
Distro-Agnostic Zarf Registry Bootstrap

### LINKED ISSUE #264: Zarf support for EKS
**Story:**
As a user of EKS and Zarf, I want the ability to deploy Zarf packages to EKS, so that I can use Zarf in my environment.

**Acceptance Criteria:**
- The Games example deploys successfully to EKS (Commercial) (definition of "successfully" is that all expected pods are running)

### LINKED ISSUE #335: Distro-Agnostic Zarf Registry Bootstrap
As a user of Zarf, I want the ability to boostrap the Zarf Registry in a distro-agnostic way, so that I can use Zarf against production-level distros instead of just development-centric ones.

Notes:
- Kubernetes has no way to add an image to the cluster without a registry
- If you want to support adding images to the cluster you have to do that somehow
- There's only one image we need to do something special with, the registry image (chicken/egg)
- The way we do it now is to utilize the binary (`kind` for example) to inject the image, but most production-level distros don't have that

AC
- I can `zarf init` without needing the remote cluster to come connect back to me
- The solution does not require me to manage PKI/TLS certs
- This outcome happens in largely the same way across all distributions. Local development-centric distros are not done in a different/special way.

jeff-mccoy:
Fixes #335
Fixes #264

This PR provides a mechanism for injecting a OCI image into a K8s cluster using only the given Kube context.  This prevents the user from having to worry about network routing, TLS, firewall rules, etc to push the initial image into the cluster so Zarf can complete the init process in an air gapped environment. 

Changes:
 - Cleaned up PKI flow / removed user interaction from `zarf init`
 - Add injector system, see [Readme](https://github.com/defenseunicorns/zarf#zarf-nerd-notes)
 - Change zarf seed config from array to single image
 - Add several additional K8s functions to the k8s package
 - Add arch diagram
 - Resolve some issues with system architecture flags and packages


jeff-mccoy:
This is still very brittle around the TCP transport of the tar ball from go -> busybox.  Seems to be a go TCP dialer thing that needs some more testing.  It did work on clean K3D/Kind/EKS clusters--but the EKS cluster failed due to likely race conditions twice before finally bootstrapping properly.  Will continue cleaning up / testing in the AM.

Screen Shot 2022-02-17 at 5 00 38 AM
Screen Shot 2022-02-17 at 5 00 24 AM

YrrepNoj:
As you're working on this can you `go mod tidy` and push up the updated go.sum. Seem to be some new dependencies for some azure stuff.

Makefile
```diff
@@ -46,11 +46,15 @@ build-cli-linux: ## Build the Linux CLI
...
+build-bootstrap-registry:
+	cd cli && $(MAKE) build-bootstrap-registry
+
 build-cli: clean build-cli-linux build-cli-mac ## Build the CLI
 
-init-package: ## Create the zarf init package, macos "brew install coreutils" first
+init-package: build-bootstrap-registry ## Create the zarf init package, macos "brew install coreutils" first
 	$(ZARF_BIN) package create --confirm
 	mv zarf-init.tar.zst build
+	rm -f build/zarf-registry
```
YrrepNoj, line 57: I have mixed feelings about this.

I felt like the fact that you could do a `zarf package create --confirm` at the main directory level to create the init package was a cool 'feature'. It also makes the init zarf.yaml conform with how we want our other example yamls to be (but as writing this I realize we do have some temporary Makefile abstraction for kustomize building resources for the BigBang example- but I think that's different because that's something that's more obvious since we're not removing the kustomzie files afterwards). 

The fact that the logic for handling the zarf-registry binary is in the `packager/create.go` and not referenced in the init zarf.yaml itself now means that someone who is creating package is forced to either use the Make command or to dive into source code to figure out why they're getting an error about some `unable to copy file` error for a file not referenced in the yaml. 

This will also make it harder for people to write their own custom init zarf.yaml in the future since 
1. it's something the CLI is expecting but it is not defined as part of the init schema.   #This is my biggest concern
2. It means that they'll have to clone the repo (or have their own registry.go to compile) instead of writing a yaml pointing to their own resources. (I'm not 100% sure this statement is true but I think it is..)


The creation now also depends on the busybox binary and the init.sh but since those assets aren't being explicitly removed I don't have as many concerns about that.

I'm not sure I worded all my thoughts on this well enough, does this make sense @jeff-mccoy?

jeff-mccoy:
Will update the task list for this PR tomorrow, still a good bit of ground to cover for complete testing/validation, but things did work well several times on EKS and GKE--it's looking promising.  Want to move the 2 new binary assets to dockerhub with cosign/sget (already pushed and signed), but keeping them local until we have time to discuss.

Screen Shot 2022-02-23 at 5 16 34 AM

Screen Shot 2022-02-23 at 3 56 17 AM

zarf.yaml
```diff
@@ -56,7 +56,7 @@ components:
       - source: assets/misc/empty-file
         target: /etc/rancher/k3s/k3s.yaml
         symlinks:
-          - /root/.kube/config
+          - /root/.kube/configkube
```
YrrepNoj, line 59: Is this an intentional change? I don't see anything else in the PR that's changed to look at/for a `configkube` file. With this change, initializing zarf with the k3s component serving as the distro only works if you manually create a symlink `ln -s ~/.kube/configkube ~/.kube/config` before running 'zarf init'.

jeff-mccoy: Nope, no idea how the at changed, feel free to fix, thanks

cli/internal/packager/seed.go
```diff
@@ -124,89 +47,26 @@ func preSeedRegistry(tempPath tempPaths) {
...
-			pki.HandlePKI()
-		}
-
-		// Start the registry again read-only now
-		startSeedRegistry(config.TLS.Host, true)
-
-	default:
-		message.Fatalf(nil, "Unknown seed registry status")
-	}
+	runInjectionMadness(tempPath)
```
YrrepNoj, line 69: This is really sensitive when initializing with the k3s component. It seems like sometimes it is trying to create an injector pod when there aren't an existing images in the cluster yet and is stopping with a generic `Unable to perform the injection` fatal error. When adding a small sleep statement before this function call I haven't had any issues.

cli/internal/k8s/pods.go
```diff
@@ -2,17 +2,80 @@ package k8s
...
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: corev1.SchemeGroupVersion.String(),
+			Kind:       "Pod",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels: map[string]string{
+				// track the creation of this ns by zarf
+				"app.kubernetes.io/managed-by": "zarf",
```
RothAndrew, line 30: I've seen this same string a few times now. Worth making more DRY?

cli/internal/k8s/tunnel.go
```diff
@@ -115,18 +125,20 @@ func (tunnel *Tunnel) Connect(target string, blocking bool) {
 		}
 	}
 
+	// On error abbort
```
RothAndrew, line 128: typo abbort -> abort

cosign.pub
```diff
@@ -0,0 +1,4 @@
+-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEQ8hMf0oDYyKmMQZdtDEZs6iTJWvX
+r4lOvDAH5xnnL2w48oiGmt012qwUfvUIBAMViIjByJ+3g6QzKpvEfZl1vQ==
+-----END PUBLIC KEY-----
```
RothAndrew, line 4: Looks like this is the key that is used to sign a couple of images that are hosted in dockerhub. Do I have that right?

configmap.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
RothAndrew: Suggest adding a comment to this file to explain what it is and why we need it

examples/eks.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+metadata:
+  name: zarf-eks-test
```
RothAndrew, line 4: Suggest finding a way to add randomness to the name, else only one will be able to be active at a time in our AWS account.

Not critical for this PR so we don't creep the scope but should get added to the backlog.

injector/stage1/README.md
```diff
@@ -0,0 +1,5 @@
+## Compile must be statically with musl in order to fit as a configmap
```
RothAndrew, line 1: Not sure what this is trying to say. "statically linked"?

injector/stage1/src/main.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    let mut file_partials = file_partials.unwrap();
+
+    // ensure a default sort-order
+    file_partials.sort();
+
+    // get a buffer of the final merged file contents
+    let contents = collect_binary_data(&file_partials).unwrap();
+
+    // @todo: finish testing/use sha256sum
```
RothAndrew, line 60: is this still todo?

zarf.yaml
```diff
@@ -2,9 +2,9 @@ kind: ZarfInitConfig
 metadata:
   name: Zarf Official Init Package
   description: "Used to establish a new Zarf cluster"
+  architecture: amd64
```
RothAndrew, line 5: Does this mean we need to write a 2nd Zarf init config for ARM?

cli/internal/k8s/configmap.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/defenseunicorns/zarf/cli/internal/message"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func ReplaceConfigmap(namespace string, name string, labels map[string]string, data map[string][]byte) (*corev1.ConfigMap, error) {
+	message.Debugf("k8s.ReplaceConfigmap(%s, %s, data)", namespace, name)
+
+	if err := DeleteConfigmap(namespace, name); err != nil {
```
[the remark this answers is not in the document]
jeff-mccoy: When it's only one line I could see your point, the problem I have is we have dozens of those per file now and it gets very noisy.  I think it's something we should work towards agreement on a pattern for the style guide, but this is a common golang pattern that I prefer.

examples/eks.yaml
jeff-mccoy: yeah that's a great point, I'm not sure what the best way to handle that is, I like providing a sample for users to just run though.  I wonder if name collisions occur per-vpc or per-account?

injector/stage1/src/main.rs
jeff-mccoy: yes this has an issue I believe with the lack of a hex conversion so we aren't using it to validate in the current iteration.

zarf.yaml
jeff-mccoy: Technically no--but if we want to make all the init package images (#214 #215), we would need 2 separate ones.  You can pass the `-a` flag to zarf to override the architecture, that came with 0.15, but this just makes the default more obvious since our injector stuff right now is all amd64 compiled only rn.

jeff-mccoy:
> As you're working on this can you `go mod tidy` and push up the updated go.sum. Seem to be some new dependencies for some azure stuff.

I ran go mod tidy and don't have any new changes right now, are you having issues?  That was actually really for GCP, it needed those imports too so figured it would be cleaner to bring them all in now vs guess what we need later.

YrrepNoj:
> I ran go mod tidy and don't have any new changes right now, are you having issues?  That was actually really for GCP, it needed those imports too so figured it would be cleaner to bring them all in now vs guess what we need later.


Looks like you handled this in your latest commit :)

examples/eks.yaml
RothAndrew: I’m pretty sure it is per region

cli/internal/k8s/pods.go
jeff-mccoy: Yeah I think so

jeff-mccoy:
Validated Gitops service and Tiny Kafka examples deployed onto EKS.

Screen Shot 2022-02-28 at 2 29 26 AM

jeff-mccoy:
Validated gitops service and the game example deployed on Azure Kubernetes Service.

Screen Shot 2022-02-28 at 3 06 55 AM

cli/internal/packager/seed.go
jeff-mccoy: this now waits for images to exist

examples/eks.yaml
jeff-mccoy: resolved by using dynamic names now

cli/internal/k8s/images.go
```diff
@@ -0,0 +1,81 @@
...
+type ImageMap map[string]bool
+
+func GetAllImages() ([]string, error) {
+	var images []string
+	var err error
+	timeout := time.After(5 * time.Minute)
+
+	for {
+		// delay check 3 seconds
+		time.Sleep(2 * time.Second)
```
YrrepNoj, line 21: I like this approach a lot more!

Nit: 2 seconds vs 3 second comment :P

cli/internal/k8s/distro.go
```diff
@@ -72,3 +92,16 @@ func DetectDistro() (string, error) {
...
+	nodes, err := GetNodes()
+	if err != nil {
+		return "", err
+	}
+
+	for _, node := range nodes.Items {
+		return node.Status.NodeInfo.Architecture, nil
+	}
+
+	return "", fmt.Errorf("could not identify node architecture")
```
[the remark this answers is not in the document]
jeff-mccoy: Nah that was just mean mindlessly doing what I had done before.  It's not invalid, but `errors.New()` is more logical so I'll change it.

injector/stage2/Makefile
```diff
@@ -0,0 +1,2 @@
+build-bootstrap-registry:
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="-s -w" -o zarf-registry registry.go
```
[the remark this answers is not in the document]
jeff-mccoy: For the go binary one it's probably find to move it back to compile vs store binary, but the Rust one is extra tooling/work so what I'd really like to do is move it to https://github.com/sigstore/cosign#sget and ideally just add support for that into Zarf later on. But for now, a static binary isn't the end of the world for me.

I already built/signed/pushed both binaries as a test:
`sget --key cosign.pub defenseunicorns/zarf-injector:0.1.0`
`sget --key cosign.pub defenseunicorns/zarf-registry:0.1.0`

YrrepNoj (approved):
Still need to add more explanation around decisions made for the injector work, but the changes themselves itself LGTM.

RothAndrew (changes requested):
ADR still needs to be completed, and latest changes from master merged in/rebased

jeff-mccoy:
yeah I got bored writing last night, want to finish the adr and a little my drawing im working on too somewhere

RothAndrew:
What other docs do we need? There are a few new READMEs, but I don't see any changes to any of the existing documentation. Is this all behind-the-scenes such that we don't really need to fix any docs?

jeff-mccoy:
Think it would help to have something explaining how this works--because some ppl will ask before adopting and it's unusual enough I think it warrants some sort of explanation.  Back to the idea around public website vs github docs.

injector/stage2/Makefile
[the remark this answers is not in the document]
jeff-mccoy: We can move this to another issue down the road if it's valuable I think.

cosign.pub
jeff-mccoy: This is the public key to verify those images

YrrepNoj:
When you've been testing this have you run into an issue where Zarf isn't able to determine the node architecture when deploying with k3s? I'm a little concerned that one of the workflows failed because of that but when re-run with unrelated changes it passed.

(Give GitHub 5 seconds after clicking the link to take you down to the specific line)
https://github.com/defenseunicorns/zarf/runs/5434718689?check_suite_focus=true#step:6:505

jeff-mccoy:
I haven't seen this but thinking about how that works I bet it's a race condition between cluster init and trying to get the distro type.  We don't have any logic right now that blocks if no node is up yet but think we should add that.  I think you might only see it in test but better safe than sorry.

YrrepNoj:
Can you help me understand why that issue would only show up in the tests? If Zarf isn't able to seed the registry because k3s isn't up yet it sounds to me like this could happen anytime we `zarf init --components=k3s --confirm`, test environment or not.

Also, aren't the commands we have in the `after` section for the k3s component supposed to be acting as a blocking mechanism for k3s to come up?

```
after:
  - "/usr/sbin/kubectl get nodes"
  - "/usr/sbin/kubectl wait --for=condition=available deployment/coredns -n kube-system"
  ```
  
I can update the `get nodes` after-operation to make sure at least one node is actually returned with something like `nodecount=$(kubectl get nodes --output json | jq '.items | length') && [[ $nodecount -gt 0 ]]` but I feel like if the `kubectl wait ...` completes, then Zarf should be able to find a node when seeding the registry.

File: cli/internal/packager/injector.go [hunk 16]
```diff
@@ -21,7 +21,7 @@ import (
 // We may need to make this dynamic in the future
 var payloadChunkSize = 1024 * 512
 
-func runInjectionMadness(tempPath tempPaths) {
+func runCrazyConfigmapInjector(tempPath tempPaths) {
 	message.Debugf("packager.runInjectionMadness(%v)", tempPath)
```
YrrepNoj, line 25: nit; the function name has changed

jeff-mccoy: hahe well I thought it was confusing not stating what _type_ of injection it was since I had to add the other one....

injector/stage1/src/main.rs
jeff-mccoy: as discussed, #380 was open to cover this in a future PR

jeff-mccoy:
Ready for final review

docs/adr/0002-image-injection-into-remote-clusters-without-native-support.md
```diff
@@ -14,10 +14,8 @@ In order to create any workloads in K8s, an image has to exist on the node or be
```
YrrepNoj, line 14: rename file to 0003

docs/adr/0002-image-injection-into-remote-clusters-without-native-support.md
```diff
@@ -14,10 +14,8 @@ In order to create any workloads in K8s, an image has to exist on the node or be
...
 
-@todo: explain decision around the injector work
-
+After iterating on a number of solutions including `kubectl cp`, `kubectl exec`, a busybox configmap injection that brought in netcat to stream files and a few variations on those ideas, we found using only configmaps to be the most predictable method for injection.  The challenge wiht this mode is that the ETCD limit is 1 MB total, with binary data base64-encoded leaving roughtly 670 KBs of data per configmap, far too small to bring the registry tarball in a single configmap.  Instead we used a combination of a custom Rust binary compiled with MUSL to perform file conctenation, untarring and verification of a set of split tarball chunks across a variable number of configmaps.  We also found chunking by less than the near-maximum configmap size was more performant on some resource-constrained control planes. The details of this solution and a drawing are located in the main README of this repo.
```
YrrepNoj, line 17: misc typos:

wiht -> with
roughtly -> roughly
conctenation ->

docs/adr/0002-image-injection-into-remote-clusters-without-native-support.md
```diff
@@ -14,10 +14,8 @@ In order to create any workloads in K8s, an image has to exist on the node or be
...
 
 ## Consequences
 
-@todo: replace me
-What becomes easier or more difficult to do and any risks introduced by the change that will need to be mitigated.
+Because this mechanism uses only a standard Podspec and Configmap definition, there should be no problem running this on any generic K8s cluster.  However, the current implementation requires manually compiling the Rust binary and mini go binary for injection, these binaries are also currently committed to the git repo.  A future update should move this two compile actions to pipeline stages to eleminate stale binaries.  The other consideration is this does bring in a new language, even if it's only for a few dozen lines of code, it is a context swtich for the engineering team should an update need to be made in the future.
```
YrrepNoj, line 21: misc typos:

eleminate -> eliminate
swtich -> switch

README.md
```diff
@@ -270,11 +264,18 @@ Zarf is written entirely in [go](https://go.dev/), except for a single 400Kb bin
 - Currently the Registry and Git servers _are not HA_, see [#375](https://github.com/defenseunicorns/zarf/issues/376) and [#376](https://github.com/defenseunicorns/zarf/issues/376) for discussion on this
 - In order to avoid TLS issues, Zarf binds to `127.0.0.1:319999` on each node as a [NodePort](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport) to allow all nodes to access the pod(s) in the cluster
 - Until [#306](https://github.com/defenseunicorns/zarf/pull/306) is merged, during helm install/upgrade a [Helm PostRender](https://helm.sh/docs/topics/advanced/#post-rendering) function is called to mutate images and [ImagePullSecrets](https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod) so the deployed resoures use the NodePort binding
-- Once [#329](https://github.com/defenseunicorns/zarf/pull/329) is merged, Zarf will use a new injector system to bootstrap a new cluster:
+- With [#329](https://github.com/defenseunicorns/zarf/pull/329), Zarf uses a new injector system to bootstrap a new cluster:
```
YrrepNoj, line 267: I don't think the main README is the best place to document stuff like this. General 'nerd notes' is great and I can understand having a few 'coming soon' messages saying what interesting upcoming issues/PRs are tackling. But saying 'With this specific PR you can do this now' seems a lot more like something that belongs in tagged release notes than something that belongs in a README.

README.md
```diff
@@ -270,11 +264,18 @@ Zarf is written entirely in [go](https://go.dev/), except for a single 400Kb bin
...
-  - After this the main docker registry chart is deployed, pulls the image from the ephemeral pod and destroys the created configmaps, pod and service
\ No newline at end of file
+  - After this the main docker registry chart is deployed, pulls the image from the ephemeral pod and destroys the created configmaps, pod and service
+
+&nbsp;
+### Zarf Architecture
+![Architecture Diagram](./.images/architecture.svg)
+
+
+[Full SVG](./.images/architecture.svg) | [Source DrawIO](docs/zarf-injector.drawio)
```
YrrepNoj, line 281: @Madeline-UX should weigh in but would it be better to move the `Zarf-Managed Deployments` underneath the other other boxed diagram? 

Having the arrows leaving Zarf ordered '5,1,2,3,4' doesn't seem as clear at first glance as it could be.

YrrepNoj (changes requested):
Mostly small docs suggestions

README.md
jeff-mccoy: Yeah I had it that way initially but it makes giant loops around zarf or over across several other lines and looked really messy. They're also very clearly numbered and that one spaced outside the list visual cue so I think it's fine for now.

docs/adr/0002-image-injection-into-remote-clusters-without-native-support.md
jeff-mccoy: ooop thats super annoying--i had my adr first but it wasnt merged #lame lol

jeff-mccoy:
changes made, will need @RothAndrew and @YrrepNoj to verify and mark changes approved.

jeff-mccoy:
that ADR rename and image rename + merge conflict made that last set of changes fun.......

RothAndrew (approved):

YrrepNoj (approved):

PR #339, opened 2022-02-17
WIP: Configurable Script Timeout

### LINKED ISSUE #338: Configurable Script Timeout
In the big Software Factory package, the Atlassian suite and Jenkins are installed as separate components after Big Bang deploys, but they have to wait until Istio is running or they won't get sidecar injections. I have this in those components, so it will wait for Istio to be running before trying to deploy:

```yaml
    scripts:
      retry: true
      before:
        # Check to see if istiod is running before trying to deploy. If Istio is running then so is Gatekeeper. This is a poor man's version of "DependsOn" since we aren't doing real gitops yet
        - |
          test $(/usr/sbin/kubectl get pods -n istio-system -l app=istiod --field-selector=status.phase=Running -o json | ./jq -r '.items | length') -gt 0
```

The problem is that it can sometimes take a few minutes for Istio to be running and Zarf times it out at ~~2 minutes~~ 60 retries with a 2 second wait between tries which roughly equates to 2 minutes if you don't count actual processing time.

I'd like the ability to configure the timeout, instead of it being hardcoded to 2 minutes.

I'm not really aware of any that would let me do this all in one package, automatically, without doing something super hacky with initcontainers or something.

RothAndrew:
DO NOT MERGE

NOTE: I'm not really trying to fix this right now, but I want people to see what I'm doing. In the short term I'm just increasing the number of retries from 60 to 300 so I can be unblocked on the work I'm doing with the Software Factory Zarf package.

Might fix #338 depending on what we want to do

### DIFF
--- a/cli/internal/packager/common.go
+++ b/cli/internal/packager/common.go
@@ -225,8 +225,7 @@ func loopScriptUntilSuccess(script string, retry bool) {
 		script = strings.ReplaceAll(script, "./zarf ", binaryPath+" ")
 	}
 
-	// 2 minutes per script (60 * 2 second waits)
-	tries := 60
+	tries := 300
 	for {
 		scriptEnvVars := []string{
 			"ZARF_REGISTRY=" + config.ZarfRegistry,
@@ -243,7 +242,7 @@ func loopScriptUntilSuccess(script string, retry bool) {
 
 				// If there are no more tries left, we have failed
 				if tries < 1 {
-					spinner.Fatalf(nil, "Script timed out after 2 minutes")
+					spinner.Fatalf(nil, "Script timed out after 300 retries")
 				} else {
 					// if retry is enabled, on error wait 2 seconds and try again
 					time.Sleep(time.Second * 2)


RothAndrew:
@jeff-mccoy @YrrepNoj Can I get your feedback on this? See #338 for context. The options in front of me right now are:

- Just push to get this change merged. It would immediately unblock me from getting the Software Factory package working, at the cost of the behavior of the script timeout changing from timeout at ~2 minutes to timeout at ~10 minutes.
- Work on a "real" fix, to make the timeout configurable by adding another parameter to the `scripts:` block in the `zarf.yaml` schema. I'm not very familiar with the codebase to do that so It'll take me a while to do it.
- Try to figure out how else I can have Jira, Confluence, and Jenkins wait until Istio is running before starting their deployments. I could maybe have multiple scripts, each that end successfully other than the last one, that timeout themselves before being killed at 2 minutes by Zarf. It would look something like:

```yaml
scripts:
  retry: true
  before:
    - <Spend 1.5 minutes retrying>
    - <Spend 1.5 minutes retrying>
    - <Spend 1.5 minutes retrying>
    - <Spend 1.5 minutes retrying>
    - <Spend 1.5 minutes retrying>
    - <Spend 1.5 minutes retrying>
    - <Fail if it still isn't done yet>
```

YrrepNoj:
Instead of change the hardcoded retries from 60 to 300 for everything, what do you think of checking for some ZARF_RETRY_OVERRIDE environment variable? That way

1. It's more obvious that it's a band-aid fix and an obvious eyesore so we have more motivation to fix it
2. It doesn't change the default retries for everything else where 60 seems to be more than enough time

jeff-mccoy:
Normally I'd say env bar would be a great band-aid but here, but the timeout I picked was totally arbitrary anyway so I don't think increasing it hurts anything.

jeff-mccoy:
But to clarify it's not 300 retries as much as "10 minutes" for the user feedback.

jeff-mccoy:
Might want to update the message.

RothAndrew:
Is it though? The way I'm reading the code is that it tries 60 times, and it waits 2 seconds between each try. It doesn't look to be including the actual processing time of the script being run.

jeff-mccoy:
Yeah that's fair we would need to switch to a better pattern with goroutine/channels to make it a real timeout.  But 300 times is poor wording compared to 10 minutes--even if it's fuzzy, most scripts zarf was designed to run this with would fail immediately each time until they didn't.  This wasn't totally what this feature was made for.

RothAndrew:
`jq` is included in the Zarf package and `test` is a shell builtin.

RothAndrew:
Though, now that I'm thinking about it more, `/usr/sbin/kubectl` only applies to K3s-mode, which shouldn't be expected long term

Issue #235, opened 2022-01-20
Upgrade Big Bang Example to latest version


runyontr:
As a new user of Zarf, I would like to have a newer version of Big Bang as the default example.

Current version is 1.17.0. 

https://github.com/defenseunicorns/zarf/blob/master/examples/big-bang/zarf.yaml#L38

RothAndrew:
@matt-strong are you still actively working on this?

matt-strong:
@RothAndrew I just pushed my attempt at it to the upgrade_bb branch, but it looked like the mega-PR changed some of those files, so I halted on it.

RothAndrew:
Next time please push up a Draft PR rather than leaving it local-only.

If you want to keep working on it that is fine. If this work gets merged before the mega PR we can merge those changes in.

matt-strong:
Sure thing

RothAndrew:
@matt-strong is this something you're currently working on? If not that's fine, I just want to make sure the kanban board reflects reality. If you aren't, I'll move it back to "Planned".

PR #346, opened 2022-02-23
upgrade big bang version to latest version

matt-strong:
closes https://github.com/defenseunicorns/zarf/issues/235


examples/big-bang/zarf.yaml
```diff
@@ -13,7 +13,7 @@ components:
       - name: flux-installer
         # This will be built on the package create side and deployed as a regular manifest on package deploy
         kustomizations:
-          - https://repo1.dso.mil/platform-one/big-bang/bigbang.git//base/flux?ref=1.17.0
+          - https://repo1.dso.mil/platform-one/big-bang/bigbang.git//base/flux?ref=1.27.1
```
YrrepNoj, line 16: I think with a version change this big there are going to be changes needed in the `kustomizations/bigbang/values.yaml` that are going to need to accompany this. I suspect the helm charts for some of these repos have changed and are going to need their keys and values updated to get the BB package to deploy.

examples/big-bang/zarf.yaml
```diff
@@ -43,9 +43,9 @@ components:
           - "kustomizations/bigbang"
     # 1. helm template bigbang ./chart |  yq e '. | select(.kind == "GitRepository") | "- " + .spec.url + "@" + .spec.ref.tag' -
     # 2. Add the actual bigbang repo as well
-    # https://repo1.dso.mil/platform-one/big-bang/bigbang/-/tags/1.17.0
+    # https://repo1.dso.mil/platform-one/big-bang/bigbang/-/tags/1.27.1
     repos:
-      - https://repo1.dso.mil/platform-one/big-bang/bigbang.git@1.17.0
+      - https://repo1.dso.mil/platform-one/big-bang/bigbang.git@1.27.1
```
YrrepNoj, line 48: Given the change to the bigbang version, I think it might be useful to update the repos/images we pull in to the git versions and image tags that the new bigbang version uses.

YrrepNoj (changes requested):

RothAndrew:
Keep in mind the testing pipeline for these changes is meaningless, since the Big Bang example isn't one of the examples that is tested. We are meaning to add it but it will need to be done strategically since it is very large and would increase the length of time that the testing pipeline takes to finish by a lot.

examples/big-bang/zarf.yaml
jeff-mccoy: See https://kubernetes.io/docs/tasks/access-application-cluster/list-all-running-container-images/ for a list of ways to get all the images in the cluster.

jeff-mccoy (changes requested):
Need to update images in the package too

RothAndrew:
@matt-strong and I talked and decided that it would be best to transition this off of his plate since it would not be able to be his 1st priority. We are going to get started on this work today as part of Dash Days since we believe it is a prerequisite to [this issue](https://repo1.dso.mil/platform-one/big-bang/apps/sandbox/neuvector/-/issues/13), then I will finish it out next week. Given our need to also upgrade the big bang version in "DI2-ME" this work will be very valuable there as well.

"DI2-ME" issue: https://github.com/defenseunicorns/zarf-package-software-factory/issues/38

jeff-mccoy:
makes sense, thanks @RothAndrew

RothAndrew:
image

image

image

image

image

image

image
not sure why there is a system time difference... probably a quirk of vagrant?

image

RothAndrew:
Ready for review and merge

jeff-mccoy:
Isn't pgl available now vs efk so we can properly test this in local distros? Is that a different issue already?

RothAndrew:
I'll write up a separate issue for it. I wanted this one to be just a straight version upgrade of what is already there to keep the scope small

jeff-mccoy:
Makes sense to me

examples/Makefile
```diff
@@ -60,7 +60,7 @@ package-examples: package-example-big-bang package-example-software-factory pack
 
 .PHONY: package-example-big-bang
 package-example-big-bang: ## Create the Big Bang Core example
-	cd big-bang && $(ZARF_BIN) package create --confirm && mv zarf-package-* ../sync/
+	cd big-bang && $(ZARF_BIN) package create -l debug --confirm && mv zarf-package-* ../sync/
```
jeff-mccoy, line 63: Was this debug flag left intentionally?

RothAndrew: No, but now that it is there I guess I kinda like it being there. Gives a lot more context as the thing is being built. 

But I can take it out if you think that would be best

jeff-mccoy: Debug exposed to users is odd to me, but the fact that we still are even using make files exposed for normal users is totally wrong, so I think we can leave it until we finish getting rid of make files for non-dev work.

jeff-mccoy (approved):

YrrepNoj (approved):

PR #348, opened 2022-02-24
make sure we get an image before trying to create injector pod

YrrepNoj:
`zarf init --components=k3s --confirm` goes so quick that it tries to inject the zarf registry before k3s is able to actually populate a cluster. This ensure we wait until a pod exists within the k3s cluster that way we can actually fetch an image to use as part of our injection pod.


jeff-mccoy:
Great catch--I was actually thinking if it's the k3s installed version though we should just use the old injection, this adds needless complexity for that model and we fully control it.  I don't mind dropping cli-based injection for others though.

jeff-mccoy:
thanks I fixed it a little differently in the main PR.

Issue #208, opened 2021-12-15
Basic composability, starting with Flux


RothAndrew:
Add support for "composable zarf packages".  This will require syntax change to the zarf package definitions as well as code changes to support.

Tasks:
- [x] Define zarf.yaml specification change, should be agreed upon by the team overall
- [ ] Require component-to-component mapping to keep things sane
- [x] Syntax on the producer side (package authors) should clearly state this is an import/inclusion of other package component data
- [x] Should be completely transparent on the consumer (package deployer) side.  The component tree should be flattened during `zarf package create` so that no changes are needed for `zarf package deploy`.  
- [x] There should be no makefile or other dependencies for package producers or consumers, it should all be done natively through the zarf cli
- [x] This feature can be limited in scope to local packages/components only for now and a new issue added for remote package/component support
- [x] E2E tests should be created to demonstrate a basic composite package

mike-winberry:
Need to discuss the initial hosting of composable packages such as flux. Right now it is working as an external package and I am going to create some examples and documentation on using it, but without a current way to host the package (external repo, s3 etc...) the package will have to be created locally.

jeff-mccoy:
One thing I've been thinking about for that is we try to require shasums for other remote resources...until we have some type of singling or attestation, we may need to stick with local only.  Realize that's not ideal from UX, but I think we have a responsibility to protect producers/consumers too.

RothAndrew:
Where do we want to put pre-built packages? I see this as the beginnings of the "Zarf Appstore". What do you guys think about either:
- A new repo just for this Flux package
    - Would be less DRY but easier to have separate release lifecycles for different pre-built packages
- A new repo that will have all pre-built packages
    - More DRY, but more "monorepo" which is a pattern I'm not a huge fan of

mike-winberry:
Moved to https://github.com/defenseunicorns/zarf-package-flux

mike-winberry:
TODO: 
- [ ] validate with team the confirm imported package behavior and types changes.
- [ ]  Handle duplicate imports from nested packages. 
- [ ] Integration tests.
- [ ] Clean up

tukaanimhaka:
Maybe an example on the docs would be good to have...

PR #351, opened 2022-02-25
#208. Basic Composability

mike-winberry:


cli/config/config.go
```diff
@@ -95,6 +95,11 @@ func GetSeedImages() []string {
 	}
 }
 
+// SetSeedImages sets the list of image string specified in the package.
```
jeff-mccoy, line 98: Confused on what this is doing here?   Composability shouldn't really touch seed images as that's a feature exclusive to init packages and in the latest multi-distro PR will only be a string instead of slice.  Can you clarify why this was needed?

jeff-mccoy: To further clarify, there will never be more than a single initial seed image in the future and that should be outside the scope of the composability feature.

cli/types/types.go
```diff
@@ -54,8 +54,8 @@ type ZarfComponent struct {
 	// Scripts are custom commands that run before or after package deployment
 	Scripts ZarfComponentScripts `yaml:"scripts,omitempty"`
 
-	// Import refers to another zarf.yml package.
-	Import string `yaml:"import,omitempty"`
+	// Import refers to another zarf.yaml package.
+	Import ZarfImport `yaml:"import,omitempty"`
```
jeff-mccoy, line 58: Curious where you think this is going to need a separate struct.  Total fine if it's needed, just like to understand if you see there being more than just a Path element in the future.

jeff-mccoy (commented):
will need to go back and look more closely at the compose.go, but here are some quick thoughts.

cli/types/types.go
mike-winberry: I was thinking so we don’t have to change everything later for pulling from external, but I really wanted validation on that one cause I wasn’t sure if it would be needed. I figured it would leave room for a sha with out introducing a breaking change later. Could just be getting ahead though and can revert.

cli/config/config.go
mike-winberry: Ok great! I wanted validation on that, I was thinking it may not be needed, but since they were pulled in package create I went ahead and added but was going to ask you about that.

mike-winberry: @jeff-mccoy Removed all references to the seed images in the compose methods in commit [08943fa](https://github.com/defenseunicorns/zarf/pull/351/commits/08943fa1f8b182a27bb98342315cd78f24b4892c)

cli/types/types.go
mike-winberry: Also if we want to make this extensible and map overrides it will be pretty simple to add something like import.Components to override the imported packages component fields where there are updates.

mike-winberry:
Ready for review.

cli/internal/packager/compose.go
```diff
@@ -0,0 +1,108 @@
...
+	return component.Import != types.ZarfImport{}
+}
+
+// Validates and confirms inclusion of imported package.
+func shouldComposePackage(component *types.ZarfComponent) bool {
+	return canComposePackage(component) && (config.DeployOptions.Confirm || component.Required || ConfirmOptionalComponent(*component))
+}
+
+// Validates the sub component, errors out if validation fails.
+func canComposePackage(component *types.ZarfComponent) bool {
```
YrrepNoj, line 40: There's a bit of mixed approaches with how this function is handling outcomes. It returns `true` if nothing goes wrong but instead of returning a `false` it will only ever end the process with a fatal message. There's nothing logically wrong with this but I think it makes it a little harder to understand.

Maybe instead of returning an error, `ValidateImportPackage()` could just immediately go end the process with a fatal message like how `./cli/internal/utils/shasum.ValidateSha256Sum()` does?

test/e2e/e2e_composability_test.go
```diff
@@ -0,0 +1,57 @@
+package test
```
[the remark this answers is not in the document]
mike-winberry: Ok ill take a look at those and set it up.

cli/internal/packager/compose.go
mike-winberry: I would argue that it makes it less explicit and also violates open closed. It is a pattern I am seeing throughout the code that there are little to pure functions or classes and instead of returning values that are handled when called giving the ability to iterate or change behavior, we are hard coding and using direct manipulation and violating singleton patterns. By returning an error and making it explicit where it is used that we are failing out. it leaves the code open for change. By terminating within the function I am simple making void calls that make little to no sense when read in the code. The proposed change: 

```go
func canComposPackage(component *types.ZarfComponent) bool {
  	validate.ValidateImportPackage(component)
	return true
}
```
Is hiding the fatal logic. 

But I am open to changing it.

YrrepNoj: I'm also not super strong in my opinion. But if ValidateImportPackage() were to just fatal then there would be no need for the function canComposePakcage() and the validation function would just be called earlier.

My point was more that having a function that claims to return a bool but really only returns true is just delaying which layer you're making the void call at.

YrrepNoj: Actually I think I see your point now, there may be times where we want to check the validation without stopping the program. I guess the thing that caught my eye was the way we were only really ever returning 'true' which seems to make the return value meaningless to check.

mike-winberry: Yeah it's not my favorite but it avoids if/else hell and makes it composable by wrapping like that. Also it's likely there will be more ways to escape that method when we get to overriding imported components or even pulling from a remote source. 

I see your point with the called earlier, but when composed it makes it extremely useful and still gets called first. 
```
// Validates and confirms inclusion of imported package.
func shouldComposePackage(component *types.ZarfComponent) bool {
	return canComposePackage(component) && (config.DeployOptions.Confirm || component.Required || ConfirmOptionalComponent(*component))
}
```
By providing boolean logic I am actually allowing it to be called earlier otherwise I would have to check that the component is composable then inside the if I would have to run the function next. Because it is composed into a larger check, where the other options don't really fit within the scope of validation and are instead user preference based.

mike-winberry: And likely I should pull out those other 3 boolean checks into a function to clean that up a bit. Im going to do that now.

test/e2e/e2e_composability_test.go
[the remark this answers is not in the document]
jeff-mccoy: would love to get the other PR merged first if we agree on the test breakdown comments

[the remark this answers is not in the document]
mike-winberry: got the tests updated and working with the new test style. Gonna resolve this.

mike-winberry:
Ready for review and merge.

mike-winberry:
I need another review and approval on this. I needed to sign my old commits. Bad news I lost three commits. Good news I learned how to not have this happen again and how to properly sign an old commit if it somehow does happen again. I replaced the lost commits in the most recent single commit and all tests are passing again. Thanks!

YrrepNoj (approved):

PR #362, opened 2022-03-04
Pull Kubescape out of the Big Bang example

### LINKED ISSUE #361: Take Kubescape out of the Big Bang example
Let's pull Kubescape out of the Big Bang example. We originally put it in when we were running scans against Big Bang but it doesn't feel to me like it should permanently belong as part of the package. Removing it will slim down the package size a little and ease the maintenance burden on the package by not having to keep the Kubescape CLI version up to date.

RothAndrew:
## Description

Removes Kubescape from the Big Bang example and documentation.

## Related Issue

Fixes #361

## Type of change

It's technically breaking since it removes existing functionality but since this is an example IMO it doesn't warrant mentioning as a breaking change in release notes or any other special treatment that usually might go with breaking changes.

## Checklist before merging


jeff-mccoy (approved):
yep, makes sense to me, thanks

PR #379, opened 2022-03-10
Add a pre-commit for validating zarf schema

### LINKED ISSUE #216: Pre-commit hook for ensuring the Zarf schema is up to date
As a developer of Zarf I'd like a pre-commit hook to automatically verify the schema file is correct, so that if a change to the `zarf.yaml` is made we always ensure that change is properly reflected in the schema definition.  This will further help increase discussion that should take place around schema changes.

YrrepNoj:
## Description

Add a pre-commit hook that will validate that the ZarfInitConfig `./zarf.yaml` confirms to the schema defined at `./zarf.schema.json`. This will help prevent issues where developers update the init config without updating the schema.

## Related Issue(s)

Fixes #216


.pre-commit-config.yaml
```diff
@@ -39,3 +39,11 @@ repos:
         language: script
         pass_filenames: false
         description: "Runs `golangci-lint`, requires https://github.com/golangci/golangci-lint"
+  - repo: https://github.com/python-jsonschema/check-jsonschema
```
RothAndrew, line 42: This hook will validate zarf.yaml against the given schema, which is not what we are looking for in #216. The hook should automatically modify zarf.schema.json if the codebase has changed the schema. The way hooks like that work is they create the file and diff it against the one in the codebase. If they aren't identical, it applies the changes to the codebase and fails the hook. The `terraform-docs` hook in [this repo](https://github.com/antonbabenko/pre-commit-terraform#terraform_docs) is a good example of what we want. Here's [the code](https://github.com/antonbabenko/pre-commit-terraform/blob/master/hooks/terraform_docs.sh) for that hook.

RothAndrew: I do see this hook as being valuable though, just not as a fix to #216

jeff-mccoy: I updated the the story, I think it's fine to fail on pre-commit and warn them for now.  If we feel it needs to do more later on, we can always expand on it, but the intent is met to make sure we don't push up invalid type changes without also the schema updates.

jeff-mccoy (approved):
lgrm, just tested and it rejected as expected.

YrrepNoj:
This is the result of the pre-commit hook if you change the zarf.yaml without changing the schema definition.

Screen Shot 2022-03-10 at 4 06 17 PM

.pre-commit-config.yaml
RothAndrew: > As a developer of Zarf I'd like a pre-commit hook to automatically verify the schema file is correct

I have no problem with that, though what this does is different from that. If the story was "As a developer of Zarf I'd like a pre-commit hook to automatically verify `zarf.yaml` against the existing schema file" this would be a good resolution to that.

RothAndrew: The codebase is the only source of truth for the schema definition. If the codebase changes, let's say by removing a previously existing field, but the schema isn't changed, `zarf.yaml` could still have that field that now doesn't exist and the check would pass even though the schema isn't correct any more.

This hook is good, but only as a protection against making mistakes in `zarf.yaml`.

jeff-mccoy: Oh I see what you're saying.  Yes that's a problem.  We need to make sure that the schema matches the types and I didn't realize this was just checking the `zarf.yaml`, which won't necessarily have every field anyway.  I think we'll need to change this to PR for that to work @YrrepNoj.

jeff-mccoy (changes requested):
We will need to look into this some more.  This only checks the `zarf.yaml` for init packages and isn't fully verifying that the generated schema actually is correct.  Would it be simpler to just run `zarf tools config-schema > zarf.schema.json` in a pre-commit step?

.pre-commit-config.yaml
RothAndrew: Let's merge this PR, but leave #216 open. The combination of #216 and this will be strong, by first verifying that the schema is up to date, then verifying zarf.yaml against the now-known-good schema

RothAndrew:
> We will need to look into this some more. This only checks the `zarf.yaml` for init packages and isn't fully verifying that the generated schema actually is correct. Would it be simpler to just run `zarf tools config-schema > zarf.schema.json` in a pre-commit step?

It's not quite this simple, but yes, it should be along these lines. `pre-commit` is already smart enough to know that a changed file means a failed hook.

RothAndrew:
I would like to merge this PR, as what it is doing is still valuable, just without closing #216.

RothAndrew:
Though, it would be better to have the hook check all `zarf.yaml` files that get changed in the repo, rather than just the root one. I haven't looked closely at the hook's capabilities, but if it does what I think it should we can change the work here to run the hook an all files named `zarf.yaml`

YrrepNoj:
This hook will run against any file named `zarf.yaml` that has been staged for commit.

RothAndrew:
Cool, then yeah, I'd like to merge this, just without closing #216

YrrepNoj:
> Would it be simpler to just run `zarf tools config-schema > zarf.schema.json` in a pre-commit step?

If we're going to do something like that I would rather do it in an automated pipeline instead of a pre-commit hook. 
1. I don't want the pre-commit to be dependent on having the CLI built and building it in the hook would be too slow which seems like a lot of friction.
2. We could probably do it without the zarf CLI but that would require dependencies to be manually installed which is also too much friction IMO.

jeff-mccoy:
It doesn't need to "be built", it could be `go run cli/main.go tools`.  But 🤷

RothAndrew:
The pre-commit hooks don't have to be run locally, though once we turn on verification of them in the pipeline it will be a lot easier to run them locally then get the feedback loop on what still needs to be fixed from the pipeline.

My vision for the pipeline is that it will run `pre-commit run -a` which will fail if any of the hooks fail, which would easily and automatically get us linting, checking for secrets, code style (after we add a styler similar to how Prettier or Black work), and other great automated checks.

RothAndrew:
Doing so will require certain CLI tools to be installed on the developer's machine, such as `golangci-lint`.

jeff-mccoy (dismissed):

YrrepNoj:
Screen Shot 2022-03-11 at 12 46 05 PM

pre-commit hook now verifies the zarf-schema is up-to-date if the `cli/types/types.go` file is ever updated. It also still verifies any modified `zarf.yaml` conforms to the schema.

RothAndrew (approved):

PR #382, opened 2022-03-11
go-lint pre-commit only applies to staged changes

YrrepNoj:
The current pre-commit hook applies linting to all of the *.go files in the repo. I _think_ this was necessary back when we had multiple go.mod and go.sum files in the repo. Either way, it doesn't seem to be necessary anymore. Now the pre-commit config will only run the linting against files that are staged for commit. This also uses the same repo we use to run our other go pre-commit tasks.

NOTE: It will run against the entire file, not just the parts that have been updated.


.pre-commit-config.yaml
```diff
@@ -29,13 +29,7 @@ repos:
...
+    rev: v0.4.0
     hooks:
-      - id: golangci-lint
-        name: golangci-lint
-        entry: hooks/run-golangci-lint.sh
-        types: [go]
-        language: script
-        pass_filenames: false
-        description: "Runs `golangci-lint`, requires https://github.com/golangci/golangci-lint"
+      - id: go-lint
```
RothAndrew, line 35: I believe the hook we want is actually `id: golangci-lint`. `go-lint` is a different tool that is deprecated

RothAndrew (changes requested):

RothAndrew (changes requested):
Change

```yaml
  - repo: https://github.com/dnephin/pre-commit-golang
    rev: v0.4.0
    hooks:
      - id: go-fmt
  - repo: https://github.com/dnephin/pre-commit-golang
    rev: v0.4.0
    hooks:
      - id: golangci-lint
```
to
```yaml
  - repo: https://github.com/dnephin/pre-commit-golang
    rev: v0.4.0
    hooks:
      - id: go-fmt
      - id: golangci-lint
```

RothAndrew (approved):

YrrepNoj:
@RothAndrew I had to re-push to fix a merge conflict, when you have a moment could you re-review? Thanks.

PR #383, opened 2022-03-11
slightly re-arrange zarf arch diagram layout

jeff-mccoy:
per https://github.com/defenseunicorns/zarf/pull/329#discussion_r824874967


YrrepNoj (approved):
❤️

PR #388, opened 2022-03-12
Composed zarf init

### LINKED ISSUE #387: Organize zarf-init config to support composable components
Now that we have composable components, organizing the zarf-init package into composable components will allow for better organization of the assets as well as make it easier for separately adding components if skipped during init.  For example adding logging for gitea later on without having to re-initialize the cluster.

jeff-mccoy:
Closes #387.  This is a small reorg of the zarf init components to leverage composability and clean up the root project folder.  This change has no impact on how zarf operates other than where files are stored prior to building the init package in the pipeline.  Requires #385 to be merged first.


jeff-mccoy:
Since they are totally identical and may have identical usage (like pgl could be an init package component or a user-installed example).  I'd rather rename it all to packages.  There is nothing that makes these packages unique, except for how they are used by us.  That being said..our current examples mostly need Iron Bank which breaks ARM support and I think will really confuse people.

RothAndrew:
> Since they are totally identical and may have identical usage (like pgl could be an init package component or a user-installed example). I'd rather rename it all to packages. There is nothing that makes these packages unique, except for how they are used by us. That being said..our current examples mostly need Iron Bank which breaks ARM support and I think will really confuse people.

The difference to me is that the “packages” are intended to be used operationally, the “examples” are not.

jeff-mccoy:
That's odd to me since we have Big Bang Core as an example....

jeff-mccoy:
I don't disagree, but think we should decide on that and make it very clear if that's actually the case.  I'd also just like to do away with "examples" except for a few basic ones if that's the case.  Everything else should be actual "packages" in that sense.

RothAndrew:
> That's odd to me since we have Big Bang Core as an example....

The Big Bang Core example is definitely just an example. It is not configured such that it could be used operationally without being insecure. For example, the network CIDRs are set to "0.0.0.0/0", it uses the `bigbang.dev` domain and TLS cert, and it uses resource requests and limits that are far too low for operational use.

RothAndrew:
> I don't disagree, but think we should decide on that and make it very clear if that's actually the case. I'd also just like to do away with "examples" except for a few basic ones if that's the case. Everything else should be actual "packages" in that sense.

Works for me, as long as we clearly designate which ones are examples or demos, and which ones we will be maintaining for operational use. There's a big difference in how we will treat them from a sustainability standpoint between an example/demo and an operationally significant package

examples/README.md
```diff
@@ -1,7 +1,7 @@
 # Zarf Examples
 
-The Zarf examples demonstrate different ways to utilize Zarf in your environment.  All of these examples follow the same general release pattern and assume an offline / air-gapped deployment target.  
+The Zarf examples demonstrate different ways to utilize Zarf in your environment.  All of these examples follow the same general release pattern and assume an offline / air-gapped deployment target. 
 
 To test create a virtual area to test all examples, you can run `make all` or `make vm-init` if you've already run the examples before. Run `make vm-destroy` to clean up.
 
-> **NOTE**: This tool utilizes software pulled from multiple sources and _some_ of them require authenticated access! Check the examples themselves for the specific accounts / logins required!
+> **NOTE**: Examples are not meant for production use, they exist to demonstrate how to use Zarf in various ways.  Also, exmaples utilize software pulled from multiple sources and _some_ of them require authenticated access. Check the examples themselves for the specific accounts / logins required.
```
RothAndrew, line 7:
```suggestion
> **NOTE**: Examples are not meant for production use, they exist to demonstrate how to use Zarf in various ways.  Also, examples utilize software pulled from multiple sources and _some_ of them require authenticated access. Check the examples themselves for the specific accounts / logins required.
```

RothAndrew (commented):
I'm really liking what I'm seeing so far here. Will pull it down and run a couple of exploratory tests tomorrow when I get back to my hotel

YrrepNoj (approved):
Changes look good. Verified the newly composed packages (init and example/big-bang) deploy as expected on mac-kind and ubuntu-k3s.

YrrepNoj (approved):
Changes look good. Verified the newly composed packages (init and example/big-bang) deploy as expected on mac-kind and ubuntu-k3s.

Issue #386, opened 2022-03-12
Complete ARM Support


jeff-mccoy:
Currently we do not publish an arm init package (it was temporarily hard-coded to amd64) and lack several things for robust arm/amd support.  Some issues, such as [Iron Bank](https://p1.dso.mil/products/iron-bank/) lacking arm images we will not be able to overcome, but those packages are already coded for amd only.

- [x] Product an arm-base `zarf-init.tar.zst` along with the amd-based one on each release
- [x] Update package names to include the architecture as a suffix to avoid having to decompress/inspect to identify, e.g. `zarf-init-amd64.tar.zst` or `zarf-package-appliance-demo-multi-games-arm64.tar.zst`
- [x] Provide the package arch as a variable for use with templating components to simplify switching arch, e.g. `https://github.com/k3s-io/k3s/releases/download/v1.21.6+k3s1/k3s-airgap-images-{{architecture}}.tar.zst`
- [ ] Compile arm versions of the injector binaries (rust and go) and include them in the arm-based init package.  Note without this M1 Mac's with Rosetta will actually work since they magically allow amd binaries to run in arm containers, but other platforms would not.

PR #392, opened 2022-03-13
Resolve Apple M1 Arch Support Issues

jeff-mccoy:
## Description
Cleans up amd/arm64 arch behavior for Apple M1 computers.  This PR will not cover all tasks in #386, only those related to Apple M1 usage.

## Related Issue
#386 

## Type of change

## Checklist before merging

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,8 +49,9 @@ build-cli-mac: ## Build the Mac CLI
 build-cli: build-cli-linux build-cli-mac ## Build the CLI
 
 init-package: ## Create the zarf init package, macos "brew install coreutils" first
-	$(ZARF_BIN) package create --confirm
-	mv zarf-init.tar.zst build
+	$(ZARF_BIN) package create --confirm --architecture amd64
+	$(ZARF_BIN) package create --confirm --architecture arm64
+	mv zarf-init-*.tar.zst build
 	cd build && sha256sum -b zarf* > zarf.sha256
 	ls -lh build
 
--- a/cli/config/config.go
+++ b/cli/config/config.go
@@ -19,8 +19,7 @@ import (
 const (
 	IPV4Localhost = "127.0.0.1"
 
-	PackageInitName = "zarf-init.tar.zst"
-	PackagePrefix   = "zarf-package-"
+	PackagePrefix = "zarf-package"
 
 	// ZarfMaxChartNameLength limits helm chart name size to account for K8s/helm limits and zarf prefix
 	ZarfMaxChartNameLength = 40
@@ -102,11 +101,17 @@ func GetSeedImage() string {
 
 func GetPackageName() string {
 	metadata := GetMetaData()
-	if metadata.Uncompressed {
-		return PackagePrefix + metadata.Name + ".tar"
-	} else {
-		return PackagePrefix + metadata.Name + ".tar.zst"
+	prefix := PackagePrefix
+	suffix := "tar.zst"
+
+	if IsZarfInitConfig() {
+		prefix = "zarf"
 	}
+
+	if metadata.Uncompressed {
+		suffix = "tar"
+	}
+	return fmt.Sprintf("%s-%s-%s.%s", prefix, strings.ToLower(metadata.Name), GetArch(), suffix)
 }
 
 func GetDataInjections() []types.ZarfData {
--- a/cli/cmd/initialize.go
+++ b/cli/cmd/initialize.go
@@ -22,7 +22,7 @@ var initCmd = &cobra.Command{
 		_, _ = fmt.Fprintln(os.Stderr, zarfLogo)
 
 		// Continue running package deploy for all components like any other package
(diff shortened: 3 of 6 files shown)


Makefile
```diff
@@ -49,8 +49,9 @@ build-cli-mac: ## Build the Mac CLI
 build-cli: build-cli-linux build-cli-mac ## Build the CLI
 
 init-package: ## Create the zarf init package, macos "brew install coreutils" first
-	$(ZARF_BIN) package create --confirm
-	mv zarf-init.tar.zst build
+	$(ZARF_BIN) package create --confirm --architecture amd64
```
YrrepNoj, line 52: Do we want to try to be smart and only build the arch defined by the UNAME_S and UNAME_P?

cli/config/config.go
```diff
@@ -102,11 +101,17 @@ func GetSeedImage() string {
...
+		prefix = "zarf"
+	}
+
 	if metadata.Uncompressed {
-		return PackagePrefix + metadata.Name + ".tar"
-	} else {
-		return PackagePrefix + metadata.Name + ".tar.zst"
+		suffix = "tar"
 	}
+	return fmt.Sprintf("%s-%s-%s.%s", prefix, strings.ToLower(metadata.Name), GetArch(), suffix)
```
YrrepNoj, line 114: Am I wrong in thinking that some packages will have images that support multiple architectures (or be so simple that they don't actually have an images in the package)?

It doesn't seem like including the current arch into the package name make sense in those situations.

docs/first-time-build.md
```diff
@@ -68,7 +68,7 @@ make init-package
 > ```
 > It is likely that you've forgotten to setup access to Iron Bank _or_ that your credentials have changed. In either case, you should go back through the steps to [Log into Iron Bank](#2-log-into-iron-bank) & try to build again!
 
-Assuming everything works out, you should see a shiny new `zarf-init.tar.zst` in your `./build` directory.
+Assuming everything works out, you should see a shiny new `zarf-init-<arch>.tar.zst` in your `./build` directory.
```
YrrepNoj, line 71: There were a lot of updates to the docs and it seems like this was the only place that did <arch> instead of just writing out the two different init filenames fully.

I'm currently liking this way more but I would just ask we be consistent with however we do it.

Makefile
jeff-mccoy: Not unless we are splitting init-package for dev/CI vs release.  Issue is we need to produce both on release.  I'm fine splitting that, but you actually don't need to specify arch if your building for your own arch, because zarf already figures that out internally

cli/config/config.go
jeff-mccoy: yeah that's an interesting point, but zarf checks arch regardless because it is needed for images and crane (with how we use it) doesn't support multiarch builds rn.  I think it's a worth UX to mixing naming conventions or leave off the arch when it is really important most of the time.  This will especially be true with the mutating webhook work since it will assume all k8s resources use the internal registry by default.

cli/config/config.go
```diff
@@ -102,11 +101,17 @@ func GetSeedImage() string {
...
+		return fmt.Sprintf("zarf-init-%s.tar.zst", GetArch())
+	}
+
 	if metadata.Uncompressed {
-		return PackagePrefix + metadata.Name + ".tar"
-	} else {
-		return PackagePrefix + metadata.Name + ".tar.zst"
+		suffix = "tar"
 	}
+	return fmt.Sprintf("%s-%s-%s.%s", prefix, metadata.Name, GetArch(), suffix)
```
YrrepNoj, line 114: I don't think there's an official check to make enforce `metadata.Name` to be spaceless. What do you think about doing a strings.trim() or a replace and switch the spaces with a - character?

jeff-mccoy: I changed this because I added a validator in the last commit.  It felt cleaner to be explicit at package create time.

YrrepNoj: Would you look at that!

YrrepNoj:
Tests are failing because package names generation has been changed but we're still using the old name in the tests. Once the test get updated I will approve.

docs/first-time-build.md
jeff-mccoy: I'll try to sift back through those in a bit to standardize them

Makefile
```diff
@@ -2,6 +2,8 @@
 ZARF_BIN := ./build/zarf
 UNAME_S := $(shell uname -s)
 UNAME_P := $(shell uname -p)
+# Need a clean way to map this, arch and uname -a return x86_64 for amd64
+ARCH := amd64
```
YrrepNoj, line 6: Should we be checking if the arch should be set to arm somewhere around here?

jeff-mccoy: Yeah we could, but there's no clean way to do it that I've seen.  For ci it doesn't matter since we don't have arm VMs right now, but for local arm testing it would be a problem.  Also I really hate makefile variables in general, so that's a thing.

PR #397, opened 2022-03-14
Defer test cleanup function

### LINKED ISSUE #396: E2E tests don't cleanup properly on failure.
When looking at failed E2E test logs I realized I wasn't seeing any logs for the cluster being torn down after a failed `require...` check. This is because the cleanup function wasn't getting deferred and never being called because the function exited too soon.

### Steps to reproduce
*  Have an e2e test fail while running on a k3s cluster

### Expected result
* e2e cluster gets cleaned up after each test case when `e2e.cleanupAfterTest(...)` is called

### Actual Result
* `require.NoError()` fails the test and stops further progress which leads to the cleanup function never getting called.

YrrepNoj:
## Description

use `defer` to ensure the cleanup function gets run after each test case, even (especially..) if the case failed a require check.

## Related Issue

Fixes #396 

## Type of change


test/e2e/common.go
```diff
@@ -273,7 +274,7 @@ func (e2e *ZarfE2ETest) checkIfClusterRunning() bool {
 
 func (e2e *ZarfE2ETest) waitForHealthyCluster() error {
 	attempt := 0
-	for attempt < 10 {
+	for attempt < 15 {
```
jeff-mccoy, line 277: any details on why this change was needed?  Nbd, just curuious

jeff-mccoy (dismissed):

test/e2e/common.go
YrrepNoj: I noticed a k3d test failed once because the cluster didn't have enough time to be healthy. It's somewhat a shot in the dark but I figured it wouldn't hurt to give the runners a few extra seconds to spinup the cluster since we don't actually know the underlying specs of the hardware.

jeff-mccoy (approved):

Issue #371, opened 2022-03-08
'zarf destroy' unable to do anything if 'zarf init' doesn't complete


YrrepNoj:
### Environment
Device and OS: Ubuntu 18.04
App version: v0.15.0
Kubernetes distro being used: None natively, using k3s that Zarf brings with it

### Steps to reproduce
1. `zarf init --components=k3s --confirm`
    * Process crashes or is manually halted before initialization is completed
2. `zarf destroy --confirm`
3. What ever the issue preventing the initialization is addressed externally 
4. `zarf init --components=k3s --confirm`

### Expected result
Step '2' destroys the started up k3s cluster and step '4' stands up a new k3s cluster.

### Actual Result
If step '1' was stopped before the zarf state was saved into the registry, then nothing actually happens during the `zarf destroy --confirm`. This means a k3s cluster is still running on the host and when you try to `zarf init` again you get an unclear error message about not being able connect to the k8s cluster while still not saving anything to the zarf-state secret so you're not stuck in this loop.

The only way out of this loop is to manually remove the k3s cluster through /usr/sbin/zarf-clean-k3s.sh or some other k3s cleanup script.

PR #404, opened 2022-03-16
destroy command actually checks for empty state

### LINKED ISSUE #403: Destroy doesn't actually check for an empty state
Related to #371 

The way we defined an empty state has changed with a recent PR. We used to automatically store an 'unknown' string value in the `Distro` field in the `ZarfState` struct. Now we no logger automatically populate that field but the destroy command is still considering that necessary to be an 'empty' state.

YrrepNoj:
## Description

If a Zarf install didn't  complete successfully we might not have a `ZarfState` in the cluster to parse while doing a `zarf destroy --confirm`. In cases like this the destroy command checks if there is no `ZarfState` to figure what to do. We recently stopped automatically adding a DistroIsUnknown string to the Distro field in the ZarfState but the destroy logic is still checking and expecting that string value.

## Related Issue
Fixes #403 
Related to #371 

## Type of change

### DIFF
--- a/test/e2e/e2e_composability_test.go
+++ b/test/e2e/e2e_composability_test.go
@@ -29,11 +29,11 @@ func TestE2eExampleComposability(t *testing.T) {
 
 	// Right now we're just checking that `curl` returns 0. It can be enhanced by scraping the HTML that gets returned or something.
 	resp, err := http.Get("http://127.0.0.1:22333?doom")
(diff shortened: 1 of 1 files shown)


YrrepNoj:
Shamelessly added a _mostly_ unrelated commit to jump-start the test pipeline still GitHub went down yesterday and never started them.

RothAndrew (approved):

Issue #373, opened 2022-03-09
Support `KUBECONFIG` env var that references multiple files


RothAndrew:
**Is your feature request related to a problem? Please describe.**
My `KUBECONFIG` env var references multiple files, so that configs for different clusters can reside in their own file instead of being combined into one massive one. Zarf doesn't seem to like that pattern.

![image](https://user-images.githubusercontent.com/16000938/157492207-454b890f-aa33-42e1-b349-65630592c891.png)

Here's how I do it in my `~/.zshrc`:

```sh
# Collect all kubeconfig yaml files in directory $HOME/.kube/config. Advantage here is that each context can be in its own file which makes things easier to find.
function getKubeconfigs {
    local KUBECONFIG="${HOME}/.kube/config"
    for CONFIG in ${HOME}/.kube/configs/*; do
        KUBECONFIG="${KUBECONFIG}:${CONFIG}"
    done
    echo $KUBECONFIG
}
```

which results in:

```sh
echo $KUBECONFIG
/Users/andrewroth/.kube/config:/Users/andrewroth/.kube/configs/<redactedClusterOne>.yaml:/Users/andrewroth/.kube/configs/<redactedClusterTwo>.yaml:/Users/andrewroth/.kube/configs/<redactedClusterThree>:/Users/andrewroth/.kube/configs/<redactedClusterFour>.yaml
```

This is a supported feature of `kubectl`

![image](https://user-images.githubusercontent.com/16000938/157492984-b00ab877-d0f2-4c46-9fcb-ff8871e77a39.png)

**Describe the solution you'd like**
I'd like Zarf to support the same syntax for `$KUBECONFIG` env var that `kubectl` does.

**Describe alternatives you've considered**
For now I've changed my `$KUBECONFIG` env var to only reference one file but that's not a fix, just a workaround.

PR #407, opened 2022-03-16
Simplify cluster instantiation in tests

### LINKED ISSUE #406: Simplify Kubernetes cluster instantiation in tests
Currently, the test suite utilizes the Golang packages for K3d and KinD to create clusters internally. This is causing issues as I attempt to write a new E2E test that tests functionality with non-standard `KUBECONFIG` environment variables, since it does a complicated series of checks, validations, and logic. It currently does this:

1. Determine if a cluster is already running, and if it is, use it instead of creating one
2. If a cluster isn't running, create one, using a hard-coded `KUBECONFIG` file (`~/.kube/config`)

I believe it would be far simpler and more scalable to abstract that out into the environment. I propose:

1. Expect (and validate) a cluster to be running already, unless explicitly told to instantiate the built-in K3s cluster

It would add a bit more code to the GitHub Action yaml scripts, since the Action would now need to handle creating a cluster and destroying it after the tests finish, but would make the Golang code in the `test/e2e` folder much simpler and easier to maintain.

RothAndrew:
## Description

Simplifies the handling of kubernetes clusters from this logic:

1. Determine if a cluster is already running, and if it is, use it instead of creating one
1. If a cluster isn't running, create one using the Golang libraries for Kind and/or K3d, using a hard-coded `KUBECONFIG` file (`~/.kube/config`)
    - If TESTDISTRO is not set, run the tests twice, once on KinD, and again on K3d
    - If TESTDISTRO is set to either `kind` or `k3d` only do that one
1. Run the tests
1. Destroy the cluster if it was created and delete the `~/.kube/config` file 

To this:

1. Expect TESTDISTRO env var to be set, with value one of `provided`, `kind`, `k3d`, or `k3s` and error out if it isn't
1. If TESTDISTRO is 
    - `kind` or `k3d` -- create that cluster using the associated golang library, using a temporary file for the KUBECONFIG.
    - `provided` -- Use the provided cluster, using the existing KUBECONFIG
    - `k3s` -- Use `zarf init --components k3s` to create the cluster and `zarf destroy` to delete it (done as part of each test so that each test gets a fresh cluster)
1. Run the tests
1. If not `provided` or `k3s`, destroy the cluster and temporary KUBECONFIG file

The pipeline will provide the KinD cluster, and will expect the tests to create the K3d and K3s clusters

## Related Issue

Fixes #406
Related to #373 (except for needing a new e2e test)

## Type of change

- [X] Refactor (non-breaking change which preserves functionality and increases maintainability)

## Checklist before merging

### DIFF
--- a/.github/workflows/test-k3s.yml
+++ b/.github/workflows/test-k3s.yml
@@ -17,5 +17,7 @@ jobs:
       - name: Make Packages
         run: make init-package package-example-game package-example-data-injection package-example-gitops-data package-example-compose
       - name: Run Tests
-        # NOTE: "PATH=$PATH" preserves the default user $PATH. This is needed to maintain the version of go installed in a previous step
+        # NOTE: "PATH=$PATH" preserves the default user $PATH. This is needed to maintain the version of go installed
+        #       in a previous step. This test run will use Zarf to create a K3s cluster, and a brand new cluster will be
+        #       used for each test
         run: sudo env "PATH=$PATH" TESTDISTRO=k3s make test-e2e
--- a/.github/workflows/test-k3s.yml
+++ b/.github/workflows/test-k3s.yml
@@ -6,17 +6,17 @@ jobs:
   build-and-test:
     runs-on: ubuntu-latest
     steps:
-      - name: Install GoLang
+      - name: "Install GoLang"
         uses: actions/setup-go@v2
         with:
           go-version: 1.16.x
-      - name: Checkout Repo
+      - name: "Checkout Repo"
         uses: actions/checkout@v2
-      - name: Build CLI
+      - name: "Build CLI"
         run: make build-cli-linux
-      - name: Make Packages
+      - name: "Make Packages"
         run: make init-package package-example-game package-example-data-injection package-example-gitops-data package-example-compose
-      - name: Run Tests
+      - name: "Run Tests"
         # NOTE: "PATH=$PATH" preserves the default user $PATH. This is needed to maintain the version of go installed
         #       in a previous step. This test run will use Zarf to create a K3s cluster, and a brand new cluster will be
         #       used for each test
--- a/.github/workflows/test-kind.yml
+++ b/.github/workflows/test-kind.yml
@@ -6,15 +6,17 @@ jobs:
   build-and-test:
     runs-on: ubuntu-latest
     steps:
-      - name: Install GoLang
+      - name: "Install GoLang"
         uses: actions/setup-go@v2
         with:
           go-version: 1.16.x
-      - name: Checkout Repo
+      - name: "Checkout Repo"
         uses: actions/checkout@v2
-      - name: Build CLI
+      - name: "Create k8s Kind Cluster"
+        uses: helm/kind-action@v1.2.0
+      - name: "Build CLI"
         run: make build-cli-linux
-      - name: Make Packages
+      - name: "Make Packages"
         run: make init-package package-example-game package-example-data-injection package-example-gitops-data package-example-compose
-      - name: Run Tests
-        run: TESTDISTRO=kind make test-e2e
+      - name: "Run Tests"
+        run: make test-e2e
--- a/.github/workflows/test-kind.yml
+++ b/.github/workflows/test-kind.yml
@@ -19,4 +19,7 @@ jobs:
       - name: "Make Packages"
         run: make init-package package-example-game package-example-data-injection package-example-gitops-data package-example-compose
       - name: "Run Tests"
-        run: make test-e2e
+        # NOTE: We want to test providing a cluster to the test framework so this one creates its own KinD cluster
+        #       rather than having the test suite do it. The K3d tests do a self-provisioned cluster and the K3s tests
+        #       use Zarf to create the cluster. In this test a single cluster will be used throughout the test run.
+        run: TESTDISTRO=provided make test-e2e
--- a/Makefile
+++ b/Makefile
@@ -86,9 +86,9 @@ package-example-compose:
 	cd examples/composable-packages && ../../$(ZARF_BIN) package create --confirm && mv zarf-package-* ../../build/
 
 # TODO: This can be cleaned up a little more when `zarf init` is able to provide the path to the `zarf-init-<arch>.tar.zst`
-.PHONY: test-new-e2e
-test-e2e: ## Run e2e tests. All dependencies are assumed to be built and in the ./build directory
-	@ #Check to make sure all the packages we need exist
+.PHONY: test-e2e
+test-e2e: ## Run e2e tests. Will automatically build any required dependencies that aren't present
+	@#Check to make sure all the packages we need exist
 	@if [ ! -f $(ZARF_BIN) ]; then\
 		$(MAKE) build-cli;\
 	fi
@@ -107,4 +107,4 @@ test-e2e: ## Run e2e tests. All dependencies are assumed to be built and in the 
 	@if [ ! -f ./build/zarf-package-compose-example-$(ARCH).tar.zst ]; then\
 		$(MAKE) package-example-compose;\
 	fi
-	cd test/e2e && cp ../../build/zarf-init-$(ARCH).tar.zst . && go test ./... -v -timeout 2400s && rm zarf-init-$(ARCH).tar.zst
+	cd test/e2e && cp ../../build/zarf-init-$(ARCH).tar.zst . && go test ./... -gcflags="all=-N -l" -v -count=1 --run TestE2eExampleGame -timeout 2400s && rm zarf-init-$(ARCH).tar.zst
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ package-example-compose:
 
 # TODO: This can be cleaned up a little more when `zarf init` is able to provide the path to the `zarf-init-<arch>.tar.zst`
 .PHONY: test-new-e2e
-test-e2e: ## Run e2e tests on a KiND cluster. All dependencies are assumed to be built and in the ./build directory
+test-e2e: ## Run e2e tests. All dependencies are assumed to be built and in the ./build directory
 	@ #Check to make sure all the packages we need exist
 	@if [ ! -f $(ZARF_BIN) ]; then\
 		$(MAKE) build-cli;\
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ package-example-compose:
 
 # TODO: This can be cleaned up a little more when `zarf init` is able to provide the path to the `zarf-init-<arch>.tar.zst`
 .PHONY: test-e2e
-test-e2e: ## Run e2e tests. Will automatically build any required dependencies that aren't present
+test-e2e: ## Run e2e tests. Will automatically build any required dependencies that aren't present. Requires env var TESTDISTRO=[provided|kind|k3d|k3s]
 	@#Check to make sure all the packages we need exist
 	@if [ ! -f $(ZARF_BIN) ]; then\
 		$(MAKE) build-cli;\
--- a/Makefile
+++ b/Makefile
@@ -107,4 +107,4 @@ test-e2e: ## Run e2e tests. Will automatically build any required dependencies t
 	@if [ ! -f ./build/zarf-package-compose-example-$(ARCH).tar.zst ]; then\
 		$(MAKE) package-example-compose;\
 	fi
-	cd test/e2e && cp ../../build/zarf-init-$(ARCH).tar.zst . && go test ./... -gcflags="all=-N -l" -v -count=1 --run TestE2eExampleGame -timeout 2400s && rm zarf-init-$(ARCH).tar.zst
+	cd test/e2e && cp ../../build/zarf-init-$(ARCH).tar.zst . && go test ./... -v -count=1 -timeout 2400s && rm zarf-init-$(ARCH).tar.zst
--- a/cli/internal/k8s/common.go
+++ b/cli/internal/k8s/common.go
@@ -154,6 +154,8 @@ func init() {
 func getRestConfig() *rest.Config {
 	message.Debug("k8s.getRestConfig()")
 
+	// Build the config from the currently active kube context in the default way that the k8s client-go gets it, which
+	// is to look at the KUBECONFIG env var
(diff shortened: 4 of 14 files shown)


RothAndrew:
I don't think an ADR is needed here, but happy to write one if the team feels it is needed.

Still need to update docs, will do that next after I ensure this is working how I expect it to.

Since the whole thing is about tests, I'll go ahead and pull that label

RothAndrew:
Updated description based on offline conversation with @YrrepNoj

RothAndrew:
@YrrepNoj ready for you to look at this based on what we talked about. Still need to update docs but I think it's mostly there.

RothAndrew:
Ready for review and merge

YrrepNoj (approved):
Worked together in person to address comments.

jeff-mccoy:
Reading back through all this rn after getting back from spring break, give me a few to verify too before merge please. Also note @RothAndrew you'll probably want to squash your commits as you have an unsigned one blocking merge.

.github/workflows/test-kind.yml
```diff
@@ -6,15 +6,20 @@ jobs:
...
+      - name: "Build CLI"
         run: make build-cli-linux
-      - name: Make Packages
+      - name: "Make Packages"
         run: make init-package package-example-game package-example-data-injection package-example-gitops-data package-example-compose
-      - name: Run Tests
-        run: TESTDISTRO=kind make test-e2e
+      - name: "Run Tests"
+        # NOTE: We want to test providing a cluster to the test framework so this one creates its own KinD cluster
+        #       rather than having the test suite do it. The K3d tests do a self-provisioned cluster and the K3s tests
```
jeff-mccoy, line 23: Is there a particular reason to have the test suite create a k3d cluster vs do what we are doing for kind?  I have no preference, just the inconsistency is confusing.

Makefile
```diff
@@ -86,9 +86,9 @@ package-example-compose:
 	cd examples/composable-packages && ../../$(ZARF_BIN) package create --confirm && mv zarf-package-* ../../build/
 
 # TODO: This can be cleaned up a little more when `zarf init` is able to provide the path to the `zarf-init-<arch>.tar.zst`
-.PHONY: test-new-e2e
-test-e2e: ## Run e2e tests on a KiND cluster. All dependencies are assumed to be built and in the ./build directory
-	@ #Check to make sure all the packages we need exist
+.PHONY: test-e2e
+test-e2e: ## Run e2e tests. Will automatically build any required dependencies that aren't present. Requires env var TESTDISTRO=[provided|kind|k3d|k3s]
```
jeff-mccoy, line 90: Really like this pattern of explicitly defining what/how you are testing.

cli/internal/k8s/common.go
```diff
@@ -153,21 +152,13 @@ func init() {
...
-		// use the current context in the default kubeconfig in the home path of the user
-		homePath, err := os.UserHomeDir()
-		if err != nil {
-			message.Fatal(nil, "Unable to load the current user's home directory")
-		}
-		configPath = homePath + "/.kube/config"
-	}
-
-	config, err := clientcmd.BuildConfigFromFlags("", configPath)
+	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
```
jeff-mccoy, line 155: This is a pretty critical change (I think for the better) to the core of how we talk to the control plane that is well outside the scope of a PR for simplifying tests.  Would like to see more code comments here as to what each step is doing and a comment in the PR on why this change was needed here vs it's own PR as changing core code behavior for a tests is usually and anti-pattern.

.github/workflows/test-k3s.yml
```diff
@@ -17,5 +17,7 @@ jobs:
       - name: Make Packages
```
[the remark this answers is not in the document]
RothAndrew: IDK if it's necessary to create a convention, unless we are able to automate checks for it.

I added the quotes on a whim, since it felt like the name stood out better in the IDE since they turn green when they are wrapped in quotes. I'm perfectly fine with pulling out the quotes, or adding quotes to the other yaml files that i missed.

.github/workflows/test-kind.yml
RothAndrew: We wanted to make sure the test pipeline tested `TESTDISTRO=provided`. That's the only reason one creates the cluster and one expects one to exist.

test/e2e/README.md
```diff
@@ -1,40 +1,50 @@
...
  1. GoLang >= `1.16.x`
  2. Make
- 3. Docker 
-
+ 3. (for the `kind` and `k3d` options only) Docker
+ 4. (for the `k3s` cluster only) Linux with root privileges
 
 ### Existing K8s Cluster
-If you have a cluster already running, great! As long as your kubeconfig is accessible at `~/.kube/.config` or is exposed in your `$KUBECONFIG` environment variable, the e2e tests will be able to use your cluster to run your tests. This means that you can even test remote distros like EKS, AKS, and GKE. When the tests are completed, your cluster will still be accessible but note that since the tests are actively using your cluster it might be in a very different state.
+If you have a cluster already running, and use the env var `TESTDISTRO=provided`, the test suite will use the `KUBECONFIG` env var and the cluster that is currently configured as the active context. To make sure you are running in the right cluster, run something like `kubectl get nodes` with no optional flags and if the nodes that appear are the ones you expect, then the tests will use that cluster as well.
```
jeff-mccoy, line 16: This might be a good new issue to expose the distro detect stuff on the CLI.  It would be really clean for a `zarf tools what-distro` or something command to tell the user about their kubecontext + detected distro.

cli/internal/k8s/common.go
RothAndrew: I can pull it back out into its own PR. It was originally its own PR, but it got incorporated here later.

RothAndrew:
> Also note @RothAndrew you'll probably want to squash your commits as you have an unsigned one blocking merge.

The commit is signed, but for some reason GitHub is not seeing it as verified, even though the same GPG key was used for it as all the other commits. I'm having trouble figuring out how to squash/force-push to fix since there's a Merge Commit in the middle.

test/e2e/clusters/clusters.go
```diff
@@ -0,0 +1,256 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+	kindCluster "sigs.k8s.io/kind/pkg/cluster"
+	kindcmd "sigs.k8s.io/kind/pkg/cmd"
+)
+
+// DistroToUse is an "enum" for helping determine which k8s distro to run the tests on.
+type DistroToUse int
+
+const (
+	// DistroUnknown is the "enum" representation for when we don't know what distro the user wants.
```
jeff-mccoy, line 23: we already have constants used in the codebase to designate distros that look like they are compatible with this code, [distro.go](https://github.com/defenseunicorns/zarf/blob/master/cli/internal/k8s/distro.go#L10).  Seems like it would be cleaner to use that vs buildout an iota enum + map.

test/e2e/clusters/clusters.go
```diff
@@ -0,0 +1,256 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// TryValidateClusterIsRunning establishes a valid connection to the currently configured cluster, then returns,
+// throwing an error if anything went wrong.
+func TryValidateClusterIsRunning() error {
+	clientSet, err := GetClientSet()
```
jeff-mccoy, line 121: Why are we implementing custom logic to do what [common.go](https://github.com/defenseunicorns/zarf/blob/master/cli/internal/k8s/common.go#L104) already does?

test/e2e/clusters/clusters.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	if len(pods.Items) < 1 {
+		return fmt.Errorf("established connection to cluster, but no pods were found")
+	}
+
+	return nil
+}
+
+// GetConfig gets the currently configured Kubeconfig, or throws an error if something went wrong.
+func GetConfig() (*rest.Config, error) {
```
jeff-mccoy, line 137: Concerned about these functions being similar + similar name to the k8s package, is there a reason they had to be custom logic here vs use the core code?  We don't want custom connect logic for tests because then we're cheating the E2E tests.

test/e2e/clusters/clusters.go
YrrepNoj: This was moved from a different an existing function in the test code that existed before the new cli-supported common.go function. You're right that we should be using that instead now that it exists though.

test/e2e/clusters/clusters.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return clientSet, nil
+}
+
+// createKindCluster creates a KinD cluster given the path to use to create a kubeconfig file.
+// It returns an error if something went wrong.
+func createKindCluster(kubeconfigFile string) error {
```
jeff-mccoy, line 165: Little confused, I thought that Kind was now handled outside of tests as the docs stated above?

test/e2e/clusters/clusters.go
```diff
@@ -0,0 +1,256 @@
...
+	err := deleteClusterCommand.ExecuteContext(context.TODO())
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// waitForHealthyCluster can be used to wait until the cluster is healthy if it needs a little extra time to spin up.
+func waitForHealthyCluster() error {
```
jeff-mccoy, line 238: Same comments as before, we already have wait logic to wait for a healthy cluster we should be testing, not implementing custom logic to bypass it for tests.

test/e2e/clusters/clusters.go
RothAndrew: Different need. Here we are getting user input about what cluster they want. There we are determining what cluster Zarf is being run against.

test/e2e/main_test.go
```diff
@@ -4,102 +4,86 @@ import (
...
+	os.Exit(retCode)
+}
+
+// doAllTheThings just wraps what should go in TestMain. It's in its own function so it can
+// [a] Not have a bunch of `os.Exit()` calls in it
+// [b] Do defers properly
+// [c] Handle errors cleanly
+//
+// It returns the return code passed from `m.Run()` and any error thrown.
+func doAllTheThings(m *testing.M) (int, error) {
```
jeff-mccoy, line 36: lol like the function name 🤣

test/e2e/clusters/clusters.go
RothAndrew: The custom logic was already there before this PR. I can look at pulling it out.

jeff-mccoy (changes requested):
Overall I think this will be an improvement to the testing flow, but some custom logic may need to use the main codebase instead and we should be careful to make sure we're not doing something different than how a normal user would use Zarf as that's the point of E2E is to be as real as possible.  Thanks for working this.

test/e2e/clusters/clusters.go
RothAndrew: This doesn't cheat the E2E tests, they still use the golang application code to do its thing, through the Zarf binary that gets used.

This gets used in the TestMain function to set up the cluster.

test/e2e/clusters/clusters.go
RothAndrew: If you use `TESTDISTRO=kind` a kind cluster will be created

cli/internal/k8s/common.go
jeff-mccoy: I think it can be fine here is you can explain in this PR what's going on.  Also some code comments per line would be nice here because it's k8s magic block box right now.

test/e2e/clusters/clusters.go
jeff-mccoy: Different usage of the same data in my mind.  One is saying "this is what this cluster is" the other is saying "this is the cluster I want".  But it's still the same data in the end.  Just don't see the need to add logic with same values elsewhere--but I won't lose sleep over this going into the merge.

test/e2e/clusters/clusters.go
jeff-mccoy: Yeah let's use the k8s package for this, thx

test/e2e/clusters/clusters.go
jeff-mccoy: copy must have misunderstood that part from the docs, makes sense though.

test/e2e/clusters/clusters.go
jeff-mccoy: Copy so I guess a better question is how does this differ from the same functions in the k8s package?

test/e2e/clusters/clusters.go
jeff-mccoy: Okay then put another way, we should use the existing logic in the codebase to wait for a cluster, I think that's already been agreed in a prior comment.

test/e2e/clusters/clusters.go
RothAndrew: I'd like to leave it for now. It's already gone through a couple of iterations. I'll make an issue in the backlog to revisit it.

test/e2e/README.md
RothAndrew: Created #412

test/e2e/clusters/clusters.go
RothAndrew: Created #411

test/e2e/clusters/clusters.go
RothAndrew: Do we want to move `k8s` out of the `internal` folder so we can do this? That's a bigger decision/change than I'm willing to take on in this PR.

jeff-mccoy: It's all one go.mod, would we need to move it out of internal?  I think _we should move it_, but agree that's out of scope for this PR.

RothAndrew: image

Might just be an IDE thing, but still not something I want to do even if it is technically possible.

RothAndrew: I suggest we stick with what's there for now and updated it as part of #411

RothAndrew: @jeff-mccoy are you good with that?

jeff-mccoy (approved):

Issue #370, opened 2022-03-08
Complete Release Pipeline Gitlab -> Github Migration


jeff-mccoy:
Platform One will be reorganizing https://repo1.dso.mil and we will need to remove Zarf build actions/code off of the site.  Zarf needs a release pipeline within Github to create release artifacts and publish them. 

- [x] Replace existing Gitlab release pipeline code with Github actions able to produce release artifacts on S3 Govcloud 
- [x] Generate a Github release through automation as Gitlab did on Repo1

Note: It might be worth evaluating [GoReleaser](https://github.com/goreleaser/goreleaser/) with this transition as well.

PR #408, opened 2022-03-18
Migrate GitLab Release Pipeline to GitHub 

YrrepNoj:
## Description
Moving our release build pipeline off of PlatformOne and hosting here instead. This will create release artifacts, publish them to an S3 bucket, and create a new GitHub release with basic notes on changes since the last release.

## Related Issue
#370 

## Type of change


.github/workflows/release.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Build Things For Release
+        run: make build-cli init-package
+      - name: Set AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_GOV_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_GOV_SECRET_ACCESS_KEY }}
+          aws-region: us-gov-west-1
+      - name: Push Release Artifacts to S3 Bucket
+        run: aws s3 cp build s3://zarf-public/release/${{ github.ref_name }} --region us-gov-west-1 --recursive
```
YrrepNoj, line 29: @jeff-mccoy Do we need to be uploading release artifacts to an S3 bucket if we can just include them as assets in the GitHub release as we're doing below?

jeff-mccoy: I personally see value in having the files in s3 and on GitHub, would give our govt customers confidence that zarf releases were accessible even if there are issues with GitHub.  Also makes it easy to script version updates as the url is predictable--not sure if GitHub release artifact urls are, I know gitlab's we're not.

jeff-mccoy: That being said, if we wanted to just have this cut the release on GitHub and create artifacts without s3 for this issue, we can make a new issue for s3 later on and discuss there.

RothAndrew: Github’s URLs are definitely predictable, and as an added bonus there is even a URL to get artifacts from the “latest” release.

jeff-mccoy (changes requested):
This needs to add the new handling for amd/arm64 package names that was already merged into the old gitlab release pipeline

.github/workflows/release.yml
YrrepNoj: Keeping release artifacts on s3 as a backup incase GitHub goes down at an inopportune time is simple and reasonable. 

I'll just need someone to add the `AWS_GOV_ACCESS_KEY_ID` and `AWS_GOV_SECRET_ACCESS_KEY` that we were using in the GitLab pipeline to the GitHub actions secrets.

jeff-mccoy: I added the new secrets

jeff-mccoy (approved):

PR #409, opened 2022-03-18
Replace assert with require in critical parts

### LINKED ISSUE #405: Use Testify's `require.NoError` instead of `assert.NoError` in tests
Using `require.NoError` in cases where all subsequent assertions would be guaranteed to fail will allow the tests to fail faster. If we limit the change to just areas where all subsequent assertions will be guaranteed to fail we can fail fast when we are able to and get more context on failed tests when that context is available.

---

Original description:

Using `require.NoError` instead of `assert.NoError` will allow the tests to fail faster. `assert.NoError` allows the test to continue, only failing at the very end. `require.NoError` fails immediately.

RobertKwiatkowski:
## Description

Now using require.NoError in test parts which are critical

## Related Issue

Fixes #405

## Type of change


YrrepNoj (approved):
Thanks for updating this!

PR #413, opened 2022-03-21
Add a read-only Gitea user to repos

### LINKED ISSUE #183: Zarf read-only Gitea user
As a user of Zarf's Utility Cluster mode, I want the ability to utilize a read-only Gitea user, so that I'm not required to hand out read/write access to services that don't need it.

Notes:

1. Current discussions right now are centered on modifying Zarf to create a Gitea organization, and creating 2 users (the usual `zarf-git-user` plus a read-only user that are both part of the org). This would lock us into Gitea because it requires using their API to do it, but we're okay with that.

YrrepNoj:
## Description
If Zarf is initialized with the `gitops-service` component we will now be creating a non-admin account that will be added as a read-only user to future repos pushed via Zarf. 
Screen Shot 2022-03-21 at 3 35 10 PM

The credentials are displayed when `zarf init` completes and are stored as a secret in the zarf namespace with the name `git-pull`.  

Screen Shot 2022-03-21 at 3 42 30 PM

## Related Issue
Closes #183

## Type of change

## Checklist before merging


cli/internal/git/utils.go
```diff
@@ -22,7 +30,7 @@ type Credential struct {
 func MutateGitUrlsInText(host string, text string) string {
 	extractPathRegex := regexp.MustCompilePOSIX(`https?://[^/]+/(.*\.git)`)
 	output := extractPathRegex.ReplaceAllStringFunc(text, func(match string) string {
-		if strings.Contains(match, "/zarf-git-user/") {
+		if strings.Contains(match, config.ZarfGitPushUser) {
```
YrrepNoj, line 33: Add the `/` back

cli/internal/packager/deploy.go
```diff
@@ -82,6 +82,16 @@ func Deploy() {
 	}
 
 	if config.IsZarfInitConfig() {
+		// Create a gitea org and a read-only user
+		err := git.CreateZarfOrg()
```
YrrepNoj, line 86: This org isn't being used currently. If it's dead code it should be removed.

I was originally going to push the repos to zarf org instead of pushing to `zarf-git-user` but that would need more Gitea api calls to create an empty repo before we push the first time (totally possible but wanted to try this out to validate first.)

@jeff-mccoy @RothAndrew any opinions?

YrrepNoj, line 86: These two operations should only happen if Gitea was listed as a component. I need to think about how the best way to figure that out would be.

cli/internal/git/utils.go
```diff
@@ -232,3 +240,132 @@ func deleteBranchIfExists(gitDirectory string, branchName plumbing.ReferenceName
...
+	tunnel.Connect(k8s.ZarfGit, false)
+	defer tunnel.Close()
+
+	client := &netHttp.Client{Timeout: time.Second * 10}
+
+	// Create the user
+	createUserBody := map[string]interface{}{
+		"username":             config.ZarfGitReadUser,
+		"password":             config.GetSecret(config.StateGitPull),
+		"email":                "zarf-reader@localhost.local",
```
YrrepNoj, line 293: NOTE: The gitea chart lets us use `zarf@localhost` as the email for the admin user but when creating a user via the API we need to include a `.*` character in the email address

cli/internal/git/utils.go
```diff
@@ -232,3 +240,132 @@ func deleteBranchIfExists(gitDirectory string, branchName plumbing.ReferenceName
...
+	body := map[string]string{
+		"username":   config.ZarfGitOrg,
+		"visibility": "limited",
+	}
+	jsonData, err := json.Marshal(body)
+	if err != nil {
+		return err
+	}
+
+	request, err := netHttp.NewRequest("POST", fmt.Sprintf("http://%s:%d/api/v1/orgs", config.IPV4Localhost, k8s.PortGit), bytes.NewBuffer(jsonData))
```
YrrepNoj, line 259: This file is inconsistent with how the request, response, body variables are named.

- [ ] make consistent

cli/internal/packager/deploy.go
jeff-mccoy: The org was only needed because I assumed we couldn't grant another user access to a private user repo, I saw let's KISS and just drop that requirement.

cli/internal/git/utils.go
jeff-mccoy: Makes sense, email doesn't really matter much for this rn

src/config/config.go
```diff
@@ -24,6 +24,8 @@ const (
 	// ZarfMaxChartNameLength limits helm chart name size to account for K8s/helm limits and zarf prefix
 	ZarfMaxChartNameLength = 40
 	ZarfGitPushUser        = "zarf-git-user"
+	ZarfGitReadUser        = "zarf-git-read-user"
+	ZarfGitOrg             = "zarf-org"
```
jeff-mccoy, line 28: Think this was a hold over that should be removed

src/test/e2e/e2e_gitea_and_grafana_test.go
```diff
@@ -32,4 +43,29 @@ func TestGiteaAndGrafana(t *testing.T) {
...
+	getRepoRequest.Header.Add("accept", "application/json")
+	getRepoRequest.Header.Add("Content-Type", "application/json")
+	getRepoResponse, err := client.Do(getRepoRequest)
+	assert.NoError(t, err)
+	assert.GreaterOrEqual(t, 300, getRepoResponse.StatusCode)
+	assert.LessOrEqual(t, 200, getRepoResponse.StatusCode)
+
+	// Make sure the only permissions are pull (read)
+	getRepoResponseBody, _ := io.ReadAll(getRepoResponse.Body)
+	admin := gjson.Get(string(getRepoResponseBody), "permissions.admin").Raw
```
jeff-mccoy, line 65: Do we need to add a dep / use gjson for this, could `encoding/json` do it?

src/internal/git/utils.go
```diff
@@ -232,3 +240,95 @@ func deleteBranchIfExists(gitDirectory string, branchName plumbing.ReferenceName
 
 	return nil
 }
+
+func CreateReadOnlyUser() error {
```
jeff-mccoy, line 244: nit but this whole function could use some readability work, maybe a few more comments about what the REST calls or doing or even some line breaks added.  Realize go isn't doing any favors here, but it was a lot to read through bunched together--especially on github:-).  It might be more readable (and DRY) to wrap some of the REST call boilerplate.

YrrepNoj: @jeff-mccoy This should be more readable now

src/.scratch/main.go
```diff
@@ -0,0 +1,15 @@
+package main
```
jeff-mccoy, line 1: 👀🤔🫠

jeff-mccoy:
Looks like one of the injector binaries is showing up

src/.scratch/main.go
YrrepNoj: awwwwkward

src/config/config.go
```diff
@@ -34,6 +35,8 @@ const (
 
 	ZarfManagedByLabel     = "app.kubernetes.io/managed-by"
 	ZarfCleanupScriptsPath = "/opt/zarf"
+
+	ZarfGiteaComponentName = "gitops-service"
```
jeff-mccoy, line 39: Is this still used?

YrrepNoj: no, removed.

jeff-mccoy (approved):
LGTM

PR #414, opened 2022-03-22
Fix workflow yaml for github release 

jeff-mccoy:
Difficult to test master workflows in PRs--this tiny PR addresses issues found after merge, fix for #370 after #408.  Relevant docs:  https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,7 @@
-Name: Publish Zarf Packages on Tag
+name: Publish Zarf Packages on Tag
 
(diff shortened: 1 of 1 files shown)


YrrepNoj (approved):
Exactly as I envisioned!!

jeff-mccoy:
I don't know if I should laugh or cry.....guess I'll just do both.

PR #416, opened 2022-03-23
Hotfix: Change release names of resources uploaded to S3

YrrepNoj:
## Description

The name of the artifacts uploaded to the S3 bucket should match the name of the resources when they're built locally.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,11 +31,11 @@ jobs:
         with:
           generate_release_notes: true    # Generates basic release notes that we can edit later
           body: |
-            - [zarf-linux-amd64](https://zarf-public.s3-us-gov-west-1.amazonaws.com/release/${{ github.ref_name }}/zarf)
-            - [zarf-macos-m1](https://zarf-public.s3-us-gov-west-1.amazonaws.com/release/${{ github.ref_name }}zarf-mac-apple)
-            - [zarf-macos-intel](https://zarf-public.s3-us-gov-west-1.amazonaws.com/release/${{ github.ref_name }}zarf-mac-intel)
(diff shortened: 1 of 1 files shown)


jeff-mccoy:
Funny you should bring this up.  @Madeline-UX and I chatted about this today and think that it would be cleaner to just use the GitHub release artifacts and only keep the s3 uploads as a backup for customers in case of issues with GitHub. Thoughts @YrrepNoj?

YrrepNoj:
> Funny you should bring this up. @Madeline-UX and I chatted about this today and think that it would be cleaner to just use the GitHub release artifacts and only keep the s3 uploads as a backup for customers in case of issues with GitHub. Thoughts @YrrepNoj?

That sounds good to me. Is that something we should be adding to the README with a description along the lines of 'mirrors of the release artifacts? If so, right now there's no way to explore the S3 bucket so unless you know the exact path you won't be able to actually use the bucket to download releases. I _think_ we could make https://zarf-public.s3-us-gov-west-1.amazonaws.com/release/ public so that we could essentially 'explore' the different releases and get their downloads.

jeff-mccoy:
Tbh I don't think it matters.  If github is down (not super likely) then they wouldn't have the readme unless they cloned the repo already--in which case they could also just build the release themselves.  I think backup to GovCloud will matter to give more confidence to government customers, but a public link isn't really that critical in my mind.  An alternative might be to just cache the release URLs via a CDN or just state they are also backed up to GovCloud.

YrrepNoj:
@jeff-mccoy  That makes sense to me. Updated.

jeff-mccoy (approved):
LGTM, txh.

Issue #418, opened 2022-03-28
credentialParser exists pointer nil if cred file has a typo


jsburckhardt:
### Environment
Device and OS: amd64 ubuntu
App version: latest
Kubernetes distro being used: k3d
Other:

### Steps to reproduce
1. if your `~/.git-credentials` has a typo, the `credentialParser` will identify the error but still creates tries to capture the password

### Expected result
- skip that cred

### Actual Result
```bash
anic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1d74e54]

goroutine 1 [running]:
github.com/defenseunicorns/zarf/cli/internal/git.credentialParser(0x0, 0x0, 0x0)
	/home/runner/work/zarf/zarf/cli/internal/git/utils.go:66 +0x1b4
github.com/defenseunicorns/zarf/cli/internal/git.FindAuthForHost(0xc0004998f0, 0x2c, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/runner/work/zarf/zarf/cli/internal/git/utils.go:85 +0x4f
github.com/defenseunicorns/zarf/cli/internal/git.pull(0xc0004998f0, 0x2c, 0xc000028fc0, 0x58, 0xc0001171b8)
	/home/runner/work/zarf/zarf/cli/internal/git/pull.go:32 +0xf7
github.com/defenseunicorns/zarf/cli/internal/git.Pull(0xc0004998f0, 0x2c, 0xc0007a3f20, 0x29, 0xc0001171b8, 0xc0001171b8, 0x46)
	/home/runner/work/zarf/zarf/cli/internal/git/pull.go:25 +0xc9
github.com/defenseunicorns/zarf/cli/internal/packager.addComponent(0xc000161230, 0x13, 0xc0003f8240, 0x21, 0xc0003f8300, 0x21, 0xc0003f8360, 0x22, 0xc00050df60, 0x1e, ...)
	/home/runner/work/zarf/zarf/cli/internal/packager/create.go:172 +0x2c5
github.com/defenseunicorns/zarf/cli/internal/packager.Create()
	/home/runner/work/zarf/zarf/cli/internal/packager/create.go:59 +0x458
github.com/defenseunicorns/zarf/cli/cmd.glob..func4(0x44da340, 0xc0007936e0, 0x0, 0x3)
[...]

PR #419, opened 2022-03-28
fix(credentialParser): validate error before parsing password to avoid nil pointer. #418

jsburckhardt:
…

## Description

Even though the credentialParser is capturing an error during the parsing, it tries to retrieve the password without being sure the error is nil

## Related Issue

Fixes # (issue)
#418 

## Type of change

## Checklist before merging

### DIFF
--- a/cli/internal/git/utils.go
+++ b/cli/internal/git/utils.go
@@ -63,10 +63,10 @@ func credentialParser() []Credential {
 	scanner := bufio.NewScanner(credentialsFile)
 	for scanner.Scan() {
 		gitUrl, err := url.Parse(scanner.Text())
-		password, _ := gitUrl.User.Password()
 		if err != nil {
 			continue
 		}
+		password, _ := gitUrl.User.Password()
 		credential := Credential{
 			Path: gitUrl.Host,
 			Auth: http.BasicAuth{


jeff-mccoy (approved):
Thanks for making the PR and good catch.

jeff-mccoy:
I have no preference, @YrrepNoj / @RothAndrew would it be cleaner to have @jsburckhardt sign and push or just have one of us do a sign-off on the commits like we did before for old commits?

RothAndrew:
IMO we temporarily let admins override the requirement, ensuring that "Squash and Merge" is chosen, which will result in a commit signed by GitHub on master

cli/internal/git/utils.go
```diff
@@ -63,10 +63,10 @@ func credentialParser() []Credential {
 	scanner := bufio.NewScanner(credentialsFile)
 	for scanner.Scan() {
 		gitUrl, err := url.Parse(scanner.Text())
-		password, _ := gitUrl.User.Password()
 		if err != nil {
```
YrrepNoj, line 66: Because of the way we're using the scanner I understand that logging this error here probably doesn't make sense. 

@jeff-mccoy Would it be reasonable to check that there is >= 1 item in the credentials after we're done parsing through the scanner and log a warning or return an err (which the functions in this file are 'legacy Jeff' and don't really do)? 

If the latter, then that def shouldn't be in this PR. Just putting the question out there since it's getting touched here.

YrrepNoj (approved):
Good catch with this change.

@jeff-mccoy I think it's perfectly reasonable to merge this in and have one of us do the signing later / as necessary.

cli/internal/git/utils.go
jeff-mccoy: Agreed, maybe as a debug, but an empty slice returned I think is still valid.  Also it is totally valid that we might no creds at all.

PR #422, opened 2022-03-29
WIP: Update data-injection demo

RothAndrew:
Update data-injection demo such that it uses an init container, which is closer to what a realistic scenario would use.

Currently, the data-injection demo will fail on this, since the init container waits until the data injection is done, but the data injection won't start until the pod reports that it is healthy, which won't happen until after the init container finishes.


RothAndrew:
@jeff-mccoy ☝️

jeff-mccoy:
merged into #425

PR #425, opened 2022-03-30
Make Data Injections Happen asyc w/ Component Deployment

### LINKED ISSUE #423: Support injecting data to pods with init containers
Currently, if a pod that is to receive a data injection has an init container, Zarf will become stuck in an infinite loop. When Zarf deploys the manifests it waits until the pods that it deploys report that they are healthy, which only happens after the init containers finish. Meanwhile, the init container is waiting until the data injection has happened, which doesn't happen until after deployment of the components, which doesn't happen until all the pods report that they are healthy.

Zarf should inject data asynchronously as the pods that it is injecting into are available, even if it is just an init container

* Change Zarf to not wait -- Breaks the Seed Registry stuff.
* Modifying Zarf to support an extra parameter in zarf.yaml for `wait: false` -- More of a bandaid than a real fix.
* Modify the application container to not require an init container -- Not a realistic request since we don't maintain the applications that users of Zarf use.

YrrepNoj:
## Description
Data injections used to operate after all the components have been deployed. This is an issue when apps are expecting data before they report themselves as healthy. When this happens the pod is never able to report healthy because their init container is waiting for the data which isn't getting injected because the pods are not reporting healthy yet. A un-fun loop.

This PR moves the data-injection operation at the same time as the component deployment and has the injection happen as a go routine that way once the pods come up the data will be injected in the background.

## Related Issue
Fixes #423 

## Type of change

## Checklist before merging


jeff-mccoy (dismissed):

test/e2e/e2e_data_injection_test.go
```diff
@@ -22,12 +22,21 @@ func TestDataInjection(t *testing.T) {
 	output, err = e2e.execZarfCommand("package", "deploy", path, "--confirm", "-l=trace")
 	require.NoError(t, err, output)
 
+	time.Sleep(5 * time.Second)
+
+	// Get the data injection pod
+	namespace := "demo'"
```
RothAndrew, line 28:
```suggestion
	namespace := "demo"
```

RothAndrew:
update the schema file?

YrrepNoj:
@jeff-mccoy I thought this update would mean that we wouldn't need the loop when checking the data-injection results but that doesn't seem to be the case.

jeff-mccoy (dismissed):

cli/internal/packager/deploy.go
```diff
@@ -261,44 +262,73 @@ func deployComponents(tempPath tempPaths, component types.ZarfComponent) {
...
 				// Handle top-level directory targets
 				destination = "/"
 			}
-			cpPodExecArgs := []string{"-n", data.Target.Namespace, "cp", sourceFile, pod + ":" + destination}
 
-			if data.Target.Container != "" {
-				// Append the container args if they are specified
-				cpPodExecArgs = append(cpPodExecArgs, "-c", data.Target.Container)
-			}
+			// Inject into all the pods that
```
RothAndrew, line 302:
```suggestion
			// Inject into all the pods that... what?
```

jeff-mccoy: lol...oops

RothAndrew (approved):

RothAndrew (dismissed):

YrrepNoj:
For some reason, the k8s api must be doing some weird cashing things. The exec is not showing files in the volume that should be there because:
1. we have the init container waiting for the file to exist before the pod starts
2. we have a ReadinessProbe waiting for the files to exist and Zarf is waiting for all pods to start being continuing with the test
3. doing a manual  go exec`kubectl exec POD -- ls /test` actually shows the files there. 

Screen Shot 2022-03-30 at 10 28 52 PM

Because of that I'm going to add the retry loop to the test. 👎

jeff-mccoy (approved):
Freaking finally LGTM

PR #437, opened 2022-04-11
added asset lib for rke2 zarg config: will be prepping for tests next

### LINKED ISSUE #378: K3s Alternative Example
An example package to demonstrate how to standup a K8s cluster that isn't K3s would very helpful in demonstrating how Zarf supports cloud-native technologies below the K8s layer as well.  An easy example would be RKE2 due to it's small footprint, relatively similar technologies and out of the box air gap support.

saintskeeper:
## Description
Added RKE2 1.21.11 as a distribution package for Zarf

## Related Issue
378-ks-alternative-example

Fixes #378

## Type of change

## Checklist before merging


jeff-mccoy:
@saintskeeper what do you think about adding a test job for RKE2 like we do k3s/d, kind?

saintskeeper:
Will be working on that tomorrow  @jeff-mccoy. I might need to touch base with you on your workflow, but I intend to build out an e2e for RKE2.

RothAndrew:
E2Es are set up to accept an env var `TESTDISTRO`, of which one of the valid values is `provided`. Creating a new GitHub Actions run that stands up the RKE2 cluster, then uses `provided`, should be a good way to do it. Here's an [example](https://github.com/defenseunicorns/zarf/blob/master/.github/workflows/test-kind.yml) of one that uses `provided`, though for RKE2 I guess we'd do a `zarf package deploy`, followed by the test run? Not sure I've ever tried that before, doing a `zarf package deploy` without first doing a `zarf init`...

jeff-mccoy:
Thanks @RothAndrew for the added context!

packages/distros/rke2/zarf.yaml
```diff
@@ -0,0 +1,49 @@
+kind: ZarfInitConfig
+metadata:
+  name: "Zarf Init Package Component - RKE2"
```
YrrepNoj, line 3: The validator for the package name is pretty strict.. Right now it only allows package names to be all lowercase and contain no special characters other than '-'.

https://regex101.com/r/vpi8a8/1

@jeff-mccoy Could you help me understand why this is so strict? Right now the validator doesn't let numbers in package names either.

packages/distros/rke2/zarf.yaml
```diff
@@ -0,0 +1,49 @@
...
+      # Transfer the RKE2 image tar
+      - source: https://github.com/rancher/rke2/releases/download/v1.21.11%2Brke2r1/rke2-images.linux-amd64.tar.zst
+        shasum: 3e9ebd2f6eab58a8b791f32ef69d25061e8b850ef36c82d48ab37009b3400048
+        target: /var/lib/rancher/rke2/agent/images/rke2.tar.zst
+      # RKE2 removal script
+      - source: assets/rke2-uninstall.sh
+        target: /opt/zarf/zarf-clean-rke2.sh
+        executable: true
+      # The RKE2 systemd service definition
+      - source: assets/rke2-service.service
```
YrrepNoj, line 41: Typo? `rke2-server.service`?

packages/distros/rke2/zarf.yaml
```diff
@@ -0,0 +1,49 @@
+kind: ZarfInitConfig
```
YrrepNoj, line 1: Since this is an `init` config, the internal logic is looking for an injector at a hardcoded path of `src/injector/zarf-registry` which only works if the zarf definition is at the root level of this project. 

That said, that' a bug on the CLI side and not a bug in this PR. @jeff-mccoy

YrrepNoj (changes requested):

packages/distros/rke2/zarf.yaml
RothAndrew: Do we want this to be an init config? I'm thinking no, that it should be a package config

packages/distros/rke2/zarf.yaml
saintskeeper: good catch! thank you @YrrepNoj

packages/distros/rke2/zarf.yaml
jeff-mccoy: Yeah we do have an example of this now using `ZarfPackageConfig`: https://github.com/defenseunicorns/zarf/blob/master/packages/distros/eks/zarf.yaml.

jeff-mccoy:
Closing this issue as stale, will keep #378 open

PR #443, opened 2022-04-13
Update gitops-service component in init config

RothAndrew:
## Description

Update the name of the `gitops-service` component in the init config to `git-server`, since it now only includes a git server. 

## Related Issue

none

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist before merging

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -15,10 +15,10 @@ Backed by tooling you already know (and love) & structured to fill the gaps you 
 
 Zarf's work necessitates that some components are "always on" (a.k.a. required & cannot be disabled). Those include:
 
-|                   |Description|
-|---                |---|
-|container-seed-registry|Adds a container registry so Zarf can bootstrap itself into the cluster.|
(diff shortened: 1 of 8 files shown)


jeff-mccoy:
So just to understand--you're proposing we rename "gitops-service" to "gitea"?  Is there any particular reason or is this part of a larger set of changes?

RothAndrew:
I thinks that’s about it, yeah. It started with just updating the description since gitops-service doesn’t do a registry anymore, but then I got to thinking there could be a better name for it. Gitops-service is a pain to type and since it is only GitLab it made sense to just call it gitea

RothAndrew:
Or maybe just call it `git`? That would give us the flexibility to possibly swap out the tool in the future if we feel the need to without needing to change the component name again.

YrrepNoj:
My preference is `git-server`

jeff-mccoy:
> My preference is `git-server`

concur.

jeff-mccoy (approved):
LGTM

Issue #402, opened 2022-03-16
Optional Flag For Image Cache


YrrepNoj:
**Is your feature request related to a problem? Please describe.**

Some (potential) users are not able to change the size of their home partition when standing up a virtual instance to work on. When working with them we had to manually edit the file where the `cachePath` was defined so we could point it to a place with actual space.

**Describe the solution you'd like**
A global flag for specifying where the `zarf-image-cache` is. something like `--zarf-cache`

**Describe alternatives you've considered**
Manually editing the file wasn't the best user experience but it worked.

Issue #426, opened 2022-03-30
Make Windows Release for Zarf


jeff-mccoy:
Need to create a release of the zarf binary for Windows users

- [ ] Update release to generate a windows Zarf binary
- [ ] Validate no unsafe file path issues, e.g. "/" (https://stackoverflow.com/questions/9371031/how-do-i-create-crossplatform-file-paths-in-go)

JasonvanBrackel:
I'm speaking to Philly.NET on AirGap deployments in October, I'd love to get this done before then.  I'm bumping up the list.

Noxsios:
Prev stuffs:

https://github.com/defenseunicorns/zarf/pull/833/files

https://github.com/defenseunicorns/zarf/pull/431

PR #446, opened 2022-04-15
adds zarf cache flag

UncleGedd:
## Description

Adds a `zarf-cache` flag to specify where images are stored on a user's machine.

## Related Issue
Fixes issue #402  

## Type of change


UncleGedd:
Currently, the Zarf image cache resides in the home directory and this PR adds the option to put the cache in a subpath within the home directory. Would we like the option to put the image cache _outside_ of the home directory?

RothAndrew:
@YrrepNoj you were working on something kinda-adjacent to this kind of thing right? IIRC they had very little space on their root drive (where the home directory is mounted) but they had a separate filesystem mount that had tons of space.

Given that, perhaps the user should be able to specify anywhere they want, rather than just somewhere inside their home directory.

src/cmd/root.go
```diff
@@ -55,10 +60,10 @@ func init() {
 		// Re-add the original help function
 		originalHelp(c, s)
 	})
-
 	rootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
 	rootCmd.PersistentFlags().StringVarP(&zarfLogLevel, "log-level", "l", "", "Log level when running Zarf. Valid options are: warn, info, debug, trace")
 	rootCmd.PersistentFlags().StringVarP(&arch, "architecture", "a", "", "Architecture for OCI images")
+	rootCmd.PersistentFlags().StringVar(&zarfImageCache, "zarf-cache", config.ZarfDefaultImageCachePath, "Specify the location of the Zarf image cache")
```
YrrepNoj, line 66: @jeff-mccoy I know you're trying to be mindful about how/when we add flags, especially if they're global flags. Do you think this flag is reasonable to be global? I think it should also be possible to only add the flag to the commands that use the .zarf-image-cache to avoid having a global flag at the cost of having more flags to maintain.

I also realized having a flag to dictate where the tmp directory gets created would also be needed to better solve the 'not enough space on root drive so I want to use a mounted drive' problem so I'm asking the above question twice I guess.

jeff-mccoy: @YrrepNoj this flag would only be used for `package create` operations so I think it should be under that probably.  You do lose the ability to do `zarf zarf.yaml` with this flag, but that's actually consistent with the rest of our behaviors already.

jeff-mccoy:
Thanks @UncleGedd, this is a helpful PR.  I think we should discuss moving the flag down to `zarf package create` though as mentioned by @YrrepNoj.

UncleGedd:
> Thanks @UncleGedd, this is a helpful PR. I think we should discuss moving the flag down to `zarf package create` though as mentioned by @YrrepNoj.

Easy fix! Will update and add an e2e test for `zarf package create`

src/cmd/root.go
UncleGedd: Refactored so that the the new `zarf-cache` flag can only be used in `zarf package create` commands, also added an e2e test

src/cmd/package.go
```diff
@@ -68,13 +75,23 @@ func choosePackage(args []string) string {
 	return path
 }
 
+func cachePathClean(cachePath string) bool {
+	var isCleanPath = regexp.MustCompile(`^[a-zA-Z0-9\_\-\/\.\~]+$`).MatchString
```
jeff-mccoy, line 79: I wonder how this might need to be dealt with when we add Windows support from #426?  Seems like https://pkg.go.dev/path/filepath might be needed in that case.  Thoughts?

src/internal/images/common.go
```diff
@@ -1,13 +0,0 @@
-package images
```
jeff-mccoy, line 1: Good refactor, this was all that remained after the multi-arch support was added and doesn't really make sense to stand on it's own now.

jeff-mccoy (commented):
Overall looks good and is a valuable addition, only comment is whether the regex for path detection is the best way to do that vs some path function.

src/test/e2e/e2e_zarf_create_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestE2eZarfCreate(t *testing.T) {
+	defer e2e.cleanupAfterTest(t)
+
+	// run `zarf create` with a specified image cache location
+	imageCachePath := "/tmp/.image_cache-location"
```
YrrepNoj, line 15: At some point in the test we should add this path to the `e2e.filesToRemove` list so it can get cleaned up from the host.

src/test/e2e/e2e_zarf_create_test.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"io/ioutil"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestE2eZarfCreate(t *testing.T) {
+	defer e2e.cleanupAfterTest(t)
```
YrrepNoj, line 12: This is failing on the k3s portion of the e2e test because the cleanup function is trying to do a `zarf destroy ...` when there is no k8s cluster running (since Zarf would spawn that when running the init command) which leads to some exception being thrown. 

I propose adding a check before it runs `zarf destroy`, something like this perhaps:
```
func (e2e *ZarfE2ETest) cleanupAfterTest(t *testing.T) {
	// Check if we have a running cluster to
	err := clusters.TryValidateClusterIsRunning()
	if err == nil {
		// Use Zarf to perform chart uninstallation
		output, err := e2e.execZarfCommand("destroy", "--confirm", "--remove-components", "-l=trace")
		require.NoError(t, err, output)
	}

    ....
```

Thoughts?

UncleGedd: 👍

src/test/e2e/e2e_zarf_create_test.go
UncleGedd: done!

src/test/e2e/e2e_zarf_create_test.go
UncleGedd: Added this with latest commit. Good catch, I had no idea why the k3s test kept failing

src/cmd/package.go
UncleGedd: Okay, looking a bit closer at this. That [filepath](https://pkg.go.dev/path/filepath) lib doesn't seem to do input validation. The goal of the regex was to guard against malicious input/prevent unexpected behavior when a user sets the flag to `**/**` or something weird. In the case of Windows, I think we can just add the backslash char to the regex. Thoughts @jeff-mccoy ?

UncleGedd: (backslash char and a colon that is)

jeff-mccoy: yeah I think that's sufficient, I wasn't certain anything existed in that library and waiting to error out until later on maybe isn't ideal.

src/cmd/package.go
```diff
@@ -68,13 +75,23 @@ func choosePackage(args []string) string {
...
+}
+
 func init() {
 	rootCmd.AddCommand(packageCmd)
 	packageCmd.AddCommand(packageCreateCmd)
 	packageCmd.AddCommand(packageDeployCmd)
 	packageCmd.AddCommand(packageInspectCmd)
 
 	packageCreateCmd.Flags().BoolVar(&config.DeployOptions.Confirm, "confirm", false, "Confirm package creation without prompting")
+	packageCreateCmd.Flags().StringVar(&zarfImageCache, "zarf-cache", config.ZarfDefaultImageCachePath, "Specify the location of the Zarf image cache")
```
jeff-mccoy, line 94: Not necessary for this PR, but this is another example of where [Viper](https://github.com/spf13/viper) might actually be useful as this is flag you'd likely want to be durable and not have to type it each time.

jeff-mccoy (approved):

YrrepNoj (approved):

PR #449, opened 2022-04-18
Rust Binary Checks SHA256

YrrepNoj:
When deploying an init package and performing the registry injection the Go CLI will dynamically calculate the payload sha256sum and provide it to the Rust injector binary to validate before installing the registry onto the cluster.


YrrepNoj:
This PR might look small but I promise I spent HOURS stumbling around learning all the wrong ways to do things in Rust 🤣

jeff-mccoy:
Had to do a little cleanup so we could push through the outstanding rust release issues / test this properly.

Screen Shot 2022-04-20 at 3 01 41 AM

jeff-mccoy (approved):
lgtm

PR #451, opened 2022-04-20
Remove saving of payload archive locally

YrrepNoj:
The archive should only be put in the tmp directory that gets compressed into the zarf package. This was leftover from manual testing that snuck in..


jeff-mccoy (approved):
#4.... you know what this means.

PR #453, opened 2022-04-20
Adding kustomized options consistent with the ones used by flux v2

### LINKED ISSUE #452: Kustomize options does not allow to include patches from a different subdirectory
### Environment
Device and OS: Windows WSL
App version: v0.17.0
Kubernetes distro being used: Client v1.22.5 / Server v1.23.4

### Steps to reproduce
1. Having the following structure:

```
root/
├─ applications/
│  ├─ alpha/
│  ├─ beta/
│  ├─ my-system/ManifestDeployments
│  │  ├─ prometheus/
│  │  │  ├─ patches/
│  │  │  │  ├─ labels.yaml
│  │  │  ├─ kustomization.yaml
│  │  │  ├─ namespace.yaml
│  │  │  ├─ release.yaml
│  │  │  ├─ repository.yaml
├─ clusters/
│  ├─ prod/
│  │  ├─ prometheus/
│  │  │  ├─ kustomization.yaml
│  ├─ edge/
├─ zarf.yaml
```

2. The kustomization.yaml inside clusters have a structure similar to:

```
kind: Kustomization
resources:
  - ../../../../../../applications/my-system/ManifestDeployments/prometheus
patches:
  - ../../../../../../applications/my-system/ManifestDeployments/prometheus/patches/labels.yaml
```

### Expected result

Patches can be applied to different clusters targets/versions, without need to re-create different kustomization files. Success at packaging the build.

### Actual Result

```
...; file '.../applications/my-system/ManifestDeployments/prometheus/patches/labels.yaml' is not in or below '.../clusters/edge/coral-system/ManifestDeployments/prometheus/all'

ERROR:  unable to build the kustomization for clusters/edge/my-system/ManifestDeployments/prometheus/all
```

### Severity/Priority

TBD

### Additional Context

This restriction was added here https://github.com/kubernetes-sigs/kustomize/pull/995 and here https://github.com/kubernetes-sigs/kustomize/pull/700, but later allowed under certain circumstances here https://github.com/kubernetes-sigs/kustomize/issues/865, which is an option that flux v2 uses.

mahomedalid:
## Description

Adding kustomized options consistent with the ones used by flux v2. In particular this changes the restriction to allow including patches under a different subdirectory than the main kustomization file. It also fixes some lint issues.

## Related Issue

Fixes #452 

## Type of change

## Checklist before merging


mahomedalid:
Running local lint, I see several issues, so I am not sure what kind of rules are we using. I can submit the changes fixing local lint issues, please advice.

Details:

A dot at the end of:

`// BuildKustomization reads a kustomization and builds it into a single yaml file.`

Removing the else block, as:
```
	yaml, err := resources.AsYaml();

	if err != nil {
		return fmt.Errorf("problem converting kustomization to yaml: %w", err)
	}
	
	return utils.WriteFile(destination, yaml)
```

RothAndrew:
Feel free to fix whatever lint errors pop up, though it is also fine to ignore them for now until we get a lint check added to the CI pipeline.

mahomedalid:
This requires adding the `needs-tests` label, although I don't have access.

jeff-mccoy:
Thanks @mahomedalid I think it's a great idea to follow Flux's convention here as they've paved much of these roads before and we're not trying to replace other gitops tools, but augment them with air gap thinking.  My only question for you and maybe @runyontr (since he loves all things Flux 🤣) was there any discussion about concerns with setting `kustypes.LoadRestrictionsNone`?  Where did the Flux community land on this issue, overriding by default or making it opt-in, or something else?  

Looking at the admittedly sparse docs it seems like it only affects accessing files outside the current directory, correct?  As I understand it, this would allow someone to _hypothetically_ make a zarf package with the path to some private files that could then be injected into the cluster (that was the reasoning given for the original PR).  I _think_ I'm okay with this but believe we should evaluate before just merging.

mahomedalid:
I completely agree. Before submitting this PR, I tried to look into flux repo history about the thinking behind this change. However, seems that this was at the time where a big refactor was introduced, and the main maintainer did not leave any track of that. This restriction applies to only local files and was introduced in kustomize here: https://github.com/kubernetes-sigs/kustomize/pull/998, so yes, potentially could include files to someone that has access to privileged files (for example if zarf is run as root, then potentially all files on the system could be included).

I'm in the same boat, I think this change is ok, since anyway is consistent with flux, but another path could be to be able to tweak any kustomize params as part of the zarf yaml definition and/or the zarf command line. This of course would require an architectural/design discussion and a proper ADR.

mahomedalid:
Local tests pass, on a second thought I don't think we need new tests, so it's ok to remove the label.

jeff-mccoy:
@mahomedalid please see my comments on #452, happy to discuss in the issue or PR.  Think we can get this merged soon, but with the updates mentioned in the issue.

jeff-mccoy (approved):
Overall LGTM, thanks for the contribution.  @YrrepNoj this doesn't add a new test for the new kustomize behavior, but the current E2E tests cover the primary behavior.  How do you feel about merging without a specific test for that usage?  I'm 🤷.

@mahomedalid External contributors require 2 reviewers and one of us will co-author/sign-off this PR as well to meet our government user requirements.  Thanks again for the contribution.

jeff-mccoy:
Will hold `needs-tests` label pending @YrrepNoj's feedback.

YrrepNoj (approved):
These changes are small and concise and I think make sense to merge in as they are :)

PR #459, opened 2022-04-24
338 configurable script timeout

### LINKED ISSUE #338: Configurable Script Timeout

jeff-mccoy:
Add the ability to define a script timeout length in seconds and show logs inline.  Also added an EKS example use-case for when this would be useful

Example:
```
  - name: example-one-hour-timeout-with-show-logs
    scripts:
      timeout: 3600
      showOutput: true
      before:
        - "some-long-running-command-here"
```

Fixes #338 

## Type of change

### DIFF
--- a/packages/distros/eks/eks.yaml
+++ b/packages/distros/eks/eks.yaml
@@ -2,10 +2,12 @@
 apiVersion: eksctl.io/v1alpha5
 kind: ClusterConfig
 metadata:
-  region: us-east-1
+# name: CHANGE_ME_MUST_BE_UNIQUE  
+  region: us-west-1
   version: "1.21"
 managedNodeGroups:
 - instanceType: t3.small
+# name: CHANGE_ME_MUST_BE_UNIQUE
   minSize: 3
   maxSize: 6
   spot: true
--- a/src/internal/utils/exec.go
+++ b/src/internal/utils/exec.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -17,10 +18,10 @@ const colorCyan = "\x1b[36;1m"
 const colorWhite = "\x1b[37;1m"
 
 //nolint
-func ExecCommand(showLogs bool, envVariables []string, commandName string, args ...string) (string, error) {
+func ExecCommandWithContext(ctx context.Context, showLogs bool, commandName string, args ...string) (string, error) {
 	if showLogs {
 		fmt.Println()
-		fmt.Printf("%s", colorGreen)
+		fmt.Printf("  %s", colorGreen)
 		fmt.Print(commandName + " ")
 		fmt.Printf("%s", colorCyan)
 		fmt.Printf("%v", args)
@@ -29,11 +30,9 @@ func ExecCommand(showLogs bool, envVariables []string, commandName string, args 
 		fmt.Println("")
 	}
 
-	cmd := exec.Command(commandName, args...)
+	cmd := exec.CommandContext(ctx, commandName, args...)
+
 	env := os.Environ()
-	if envVariables != nil {
-		env = append(env, envVariables...)
-	}
 	cmd.Env = env
 
 	var stdoutBuf, stderrBuf bytes.Buffer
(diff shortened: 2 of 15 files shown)


src/internal/utils/exec.go
```diff
@@ -17,10 +18,10 @@ const colorCyan = "\x1b[36;1m"
 const colorWhite = "\x1b[37;1m"
 
 //nolint
-func ExecCommand(showLogs bool, envVariables []string, commandName string, args ...string) (string, error) {
+func ExecCommandWithContext(ctx context.Context, showLogs bool, commandName string, args ...string) (string, error) {
 	if showLogs {
 		fmt.Println()
-		fmt.Printf("%s", colorGreen)
+		fmt.Printf("  %s", colorGreen)
```
YrrepNoj, line 24: Was this space intentional?

zarf.schema.json
```diff
@@ -155,6 +155,12 @@
     },
     "ZarfComponentScripts": {
       "properties": {
+        "showOutput": {
+          "type": "boolean"
+        },
+        "timeout": {
```
YrrepNoj, line 161: The Go variable is timeoutSeconds, which I like since it makes units clear. Since a good amount of apps also use milliseconds for their timeout, what do you think about adding 'seconds' to the key name to avoid confusion?

src/internal/utils/exec.go
jeff-mccoy: yes it was, just aligning that with other logs in the message pkg.

zarf.schema.json
jeff-mccoy: Hmmm I guess we could, I did seconds because this will be to set the timeout for a long period (in most cases) and milliseconds seems a bit crazy for that.  But it doesn't destroy my soul to change it.

src/internal/utils/exec.go
```diff
@@ -17,10 +18,10 @@ const colorCyan = "\x1b[36;1m"
 const colorWhite = "\x1b[37;1m"
 
 //nolint
-func ExecCommand(showLogs bool, envVariables []string, commandName string, args ...string) (string, error) {
+func ExecCommandWithContext(ctx context.Context, showLogs bool, commandName string, args ...string) (string, error) {
```
YrrepNoj, line 21: I know it wasn't added by this PR but am I correct in understanding this `showLogs` toggle is copying logs directly  to stdout/stderr?

YrrepNoj (approved):
LGTM. I left a few optional suggestions for changes/new tasks that you can take as you see fit.

packages/distros/eks/eks.yaml
```diff
@@ -2,10 +2,12 @@
 apiVersion: eksctl.io/v1alpha5
 kind: ClusterConfig
 metadata:
-  region: us-east-1
+# name: CHANGE_ME_MUST_BE_UNIQUE
```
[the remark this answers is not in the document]
jeff-mccoy: that sounds like a really useful new issue.  I imagined down the road it's something we could be smarter in like accept user input / vars.

YrrepNoj (approved):
Thanks for the updates! LGTM.

PR #470, opened 2022-05-02
add flag --skip-sbom to package create to speed up dev packages

jeff-mccoy:


src/internal/sbom/catalog.go
```diff
@@ -24,6 +25,12 @@ var viewerAssets embed.FS
 var tranformRegex = regexp.MustCompile(`(?m)[^a-zA-Z0-9\.\-]`)
 
 func CatalogImages(tagToImage map[name.Tag]v1.Image, sbomDir, tarPath string) {
+	// Ignore SBOM creation if there the flag is set
+	if config.SkipSBOM {
```
YrrepNoj, line 29: I personally prefer the pattern of checking this before the function is called. It's all personal preference though so I'm just talking to talk right now.

YrrepNoj (approved):
Nice QoL change!

jeff-mccoy:
Here for u devz 🦄

src/internal/sbom/catalog.go
jeff-mccoy: I did that at first but it was an extra line of code and I'm evil 😭.  I don't mind if you want to change it though.

Issue #456, opened 2022-04-21
Improve pre-reqs of the e2e tests readme


mahomedalid:
While trying to run the e2e tests locally I am struggling to understand the pre-reqs. After some debugging I have found some preq-reqs that could be useful to add to the [README](https://github.com/defenseunicorns/zarf/tree/master/src/test/e2e#readme). At the moment I have:

- The scripts will try to get authentication from ~/.git-credentials, so store credentials there will be required, specifically for:

https://repo1.dso.mil/platform-one/big-bang/apps/security-tools/twistlock.git@0.0.9-bb.0authentication
https://github.com/defenseunicorns/zarf.git@v0.15.0

It would probably also be a good idea to either add a way to add verbosity to the e2e tests (debug flags) or add a more specific error message to src/internal/git/pull.go including git error and repo url.

I am planning to submit a PR after I got the e2e tests running successfully.

jeff-mccoy:
@YrrepNoj this might be a good thing to start for the doc updates we discussed.

jeff-mccoy:
Could also be related to https://github.com/defenseunicorns/zarf/issues/417 comments, having a way to fallback to host git seems like a nice solution to these situations.

mahomedalid:
@jeff-mccoy yes! that could also alleviate the issues/confusion with running e2e testing locally.

PR #471, opened 2022-05-02
Fallback to host-provided git on clone failure during package create

### LINKED ISSUE #417: Azure DevOps repos support
Our project hosts the target repos in Azure DevOps. I can see the tool is currently using go-git which doesn't support he Azure DevOps git protocol (libgit2 is required)

Should be able to target a `https://dev.azure.com/<sample org>/<sample project>/_git/<sample repo>` link

- Mirroring repository || not secure and a pain to maintain
- Inspected how flux is doing it -> for azdo they use libgit2

**Additional context**
- Flux used to have the same issue -> https://fluxcd.io/docs/components/source/gitrepositories/#git-implementation

jeff-mccoy:
## Description

This PR allows for graceful failure on various go-git cloning issues related to authentication, git protocol, etc during the `zarf package create` execution.  On a clone failure, the git package will attempt to call the host-based `git clone` before completely failing.  Note that due to the format of Azure http git clone urls, this command will not support single-tag/single-depth clones in this PR.  

## Related Issue

Related to #456 
Fixes #417 

## Type of change

## Checklist before merging

- [x] Tests included via the addition of the new test case in the gitops example

### DIFF
--- a/src/internal/git/pull.go
+++ b/src/internal/git/pull.go
@@ -60,7 +60,7 @@ func pull(gitUrl, targetFolder string, spinner *message.Spinner) {
 
 		// If we can't clone with go-git, fallback to the host clone
 		// Only support "all tags" due to the azure clone url format including a username
-		stdOut, stdErr, err := utils.ExecCommandWithContext(context.TODO(), false, "git", "clone", gitUrl, targetFolder)
+		stdOut, stdErr, err := utils.ExecCommandWithContext(context.TODO(), false, "git", "clone", "--origin", onlineRemoteName, gitUrl, targetFolder)
 		spinner.Updatef(stdOut)
 		spinner.Debugf(stdErr)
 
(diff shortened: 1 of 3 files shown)


jeff-mccoy:
Successful clone via fallback:
Screen Shot 2022-05-02 at 12 14 22 AM

Failed clone (invalid git url) via fallback:
Screen Shot 2022-05-02 at 12 14 46 AM

src/internal/git/pull.go
```diff
@@ -54,7 +56,19 @@ func pull(gitUrl string, targetFolder string, spinner *message.Spinner) {
 	if err == git.ErrRepositoryAlreadyExists {
 		spinner.Debugf("Repo already cloned")
 	} else if err != nil {
```
YrrepNoj, line 58: Is there a specific err type that we can check for instead of catching all of them?

Or at least, we should debug log the original err. If there is an err with go-git cloning the repo and the user doesn't have `git` in their path then the messing logged to debug will not be (as?) useful.

YrrepNoj (changes requested):
Just request one extra debug statement.

src/internal/git/pull.go
jeff-mccoy: There are other errors we _could_ catch here, but I think it's cleaner to fall-through on any error for now because the intent is to get the git data one way or another and there will likely be other edge cases we haven't identified yet.

YrrepNoj (approved):

PR #478, opened 2022-05-06
Use registry secrets

### LINKED ISSUE #477: Use Registry Secrets When Deploying Packages
`zarf package deploy` currently relies on the local ~/.docker/config.json for the credentials into the docker registry hosted in the connected k8s cluster. These credentials get populated when the `zarf init` command is run.

This is unintentional as it leads to issues when deploying packages as a different user or on a different host than the machine that ran the `zarf init` command.

### Expected result
`zarf package deploy` should be using the secrets in the zarf namespace to get the credentials to push to the docker registry. Since we use crane, this probably means we will need to fetch the secret and do a 'docker login' (or equivalent) when doing a deploy so that we can deploy packages to an initialized cluster no matter who/where we are.

### Actual Result
`zarf package deploy` uses crane which doesn't have any flags/configs for passing in credentials during a push. Instead, it just reads from the ~/.docker/config.json

vyta:
## Related Issue

Fixes #477
## Type of change

## Checklist before merging

### DIFF
--- a/src/config/config.go
+++ b/src/config/config.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/defenseunicorns/zarf/src/internal/message"
 	"github.com/defenseunicorns/zarf/src/internal/utils"
+	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/crane"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 )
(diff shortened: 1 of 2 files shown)


YrrepNoj (approved):
This looks great! Our internal team rules require two approvals for outside contributors so once someone else on the team gives this a look over we'll merge it in. Thanks again :)

jeff-mccoy (approved):
thanks, lgtm.

jeff-mccoy:
@YrrepNoj / @RothAndrew I did a git sign-off amendment as part of the process we discussed for 3rd-party contributors.  Based on how github treats that I think we need to update the contributor guides to have the first internal team member do a `git commit --amend --signoff` and then require another team member to do the PR approval.  Thoughts?

YrrepNoj (approved):

YrrepNoj:
@jeff-mccoy That makes sense to me, we should definitely update our contributor guide to explain why we're going the extra step to signoff on the commits though

PR #480, opened 2022-05-08
Add helm release name property to zarf chart

### LINKED ISSUE #479: Add Release Name Support for Multiple Deployments of a Helm Chart with Different Names
I have a use case where I need to deploy multiple copies of a Helm chart with different release names.  Currently Zarf uses the chart name only.  If you attempt to create a Zarf package with more than one ZarfChart of the same name, it will install the first and then upgrade it to the second.

Just like the Helm CLI, it would be useful to provide an optional `releaseName` ZarfChart property that would allow users to override the default chart name with something different.

Some charts can be deployed in different namespaces, but this workaround does not work for charts that have global resources (e.g., Service Accounts, etc.) where the rendered chart resource names will conflict.

**Additional context**
N/A

dsmithbauer:
## Description

Added `releaseName` property to ZarfChart and updated the install/upgrade chart function to use `releaseName` if provided or fall back to `name` if not provided to keep backwards compatibility and not require the new property.

Tested this change by deploying multiple copies of cert-manager chart (just as a test) to the same namespace using different Zarf components and the same Zarf component with a list of charts too.

```
kind: ZarfPackageConfig
metadata:
  name: test-helm-releasename
  description: "Deploys same chart with different names"
components:
  - name: cert-manager
    required: false
    charts:
      - name: cert-manager
        # releaseName: cert-manager # use the default name if releaseName not specified
        url: https://charts.jetstack.io
        version: 1.6.3
        namespace: cert-manager
    images:
      - "quay.io/jetstack/cert-manager-cainjector:v1.6.3"
      - "quay.io/jetstack/cert-manager-controller:v1.6.3"
      - "quay.io/jetstack/cert-manager-webhook:v1.6.3"
      - "quay.io/jetstack/cert-manager-ctl:v1.6.3"
  - name: cert-manager-2
    required: false
    charts:
      - name: cert-manager
        releaseName: cert-manager-2
        url: https://charts.jetstack.io
        version: 1.6.3
        namespace: cert-manager
    images:
      - "quay.io/jetstack/cert-manager-cainjector:v1.6.3"
      - "quay.io/jetstack/cert-manager-controller:v1.6.3"
      - "quay.io/jetstack/cert-manager-webhook:v1.6.3"
      - "quay.io/jetstack/cert-manager-ctl:v1.6.3"
  - name: cert-manager-3-4
    required: false
    charts:
      - name: cert-manager
        releaseName: cert-manager-3
        url: https://charts.jetstack.io
        version: 1.6.3
        namespace: cert-manager-three-four
      - name: cert-manager
        releaseName: cert-manager-4
        url: https://charts.jetstack.io
        version: 1.6.3
        namespace: cert-manager-three-four
    images:
      - "quay.io/jetstack/cert-manager-cainjector:v1.6.3"
      - "quay.io/jetstack/cert-manager-controller:v1.6.3"
      - "quay.io/jetstack/cert-manager-webhook:v1.6.3"
      - "quay.io/jetstack/cert-manager-ctl:v1.6.3"
  ```

## Related Issue

Fixes #479 

## Type of change

## Checklist before merging

### DIFF
--- a/src/test/e2e/e2e_helm_test.go
+++ b/src/test/e2e/e2e_helm_test.go
@@ -19,13 +19,12 @@ func TestE2eExampleHelm(t *testing.T) {
 	path := fmt.Sprintf("../../../build/zarf-package-test-helm-releasename-%s.tar.zst", e2e.arch)
 
 	// Deploy the charts
-	output, err = e2e.execZarfCommand("package", "deploy", path, "--confirm", "-l=trace")
+	output, err = e2e.execZarfCommand("package", "deploy", path, "--confirm")
 	require.NoError(t, err, output)
 
-	// Verify the configmap was properly templated
-	kubectlOut, _ := exec.Command("kubectl", "get", "pods", "-A", "--selector=app.kubernetes.io/name=etcd", "--no-headers").Output()
-	assert.Contains(t, string(kubectlOut), "zarf-etcd-0")
-	assert.Contains(t, string(kubectlOut), "zarf-etcd-2-0")
-	assert.Contains(t, string(kubectlOut), "zarf-etcd-3-0")
-	assert.Contains(t, string(kubectlOut), "zarf-etcd-4-0")
+	// Verify multiple helm installs of different release names were deployed
+	kubectlOut, _ := exec.Command("kubectl", "get", "pods", "-A", "--selector=app.kubernetes.io/name=nginx", "--no-headers").Output()
+	assert.Contains(t, string(kubectlOut), "zarf-nginx-2")
+	assert.Contains(t, string(kubectlOut), "zarf-nginx-3")
+	assert.Contains(t, string(kubectlOut), "zarf-nginx-4")
 }
--- a/src/test/e2e/e2e_helm_test.go
+++ b/src/test/e2e/e2e_helm_test.go
@@ -23,7 +23,7 @@ func TestE2eExampleHelm(t *testing.T) {
 	require.NoError(t, err, output)
 
 	// Verify the configmap was properly templated
-	kubectlOut, _ := exec.Command("kubectl", "-A", "get", "pods", "--selector=app.kubernetes.io/name=etcd", "--no-headers").Output()
+	kubectlOut, _ := exec.Command("kubectl", "get", "pods", "-A", "--selector=app.kubernetes.io/name=etcd", "--no-headers").Output()
 	assert.Contains(t, string(kubectlOut), "zarf-etcd-0")
 	assert.Contains(t, string(kubectlOut), "zarf-etcd-2-0")
 	assert.Contains(t, string(kubectlOut), "zarf-etcd-3-0")
--- a/src/test/e2e/e2e_helm_test.go
+++ b/src/test/e2e/e2e_helm_test.go
@@ -1,0 +1,31 @@
+package test
+
+import (
+	"fmt"
+	"os/exec"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestE2eExampleHelm(t *testing.T) {
+	//defer e2e.cleanupAfterTest(t)
+
+	//run `zarf init`
+	output, err := e2e.execZarfCommand("init", "--confirm")
+	require.NoError(t, err, output)
+
+	path := fmt.Sprintf("../../../build/zarf-package-test-helm-releasename-%s.tar.zst", e2e.arch)
+
+	// Deploy the charts
+	output, err = e2e.execZarfCommand("package", "deploy", path, "--confirm")
+	require.NoError(t, err, output)
+
+	// Verify the configmap was properly templated
+	kubectlOut, _ := exec.Command("kubectl", "-A", "get", "pods", "--selector=app.kubernetes.io/name=etcd", "--no-headers").Output()
+	assert.Contains(t, string(kubectlOut), "zarf-etcd-0")
+	assert.Contains(t, string(kubectlOut), "zarf-etcd-2-0")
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by jeff-mccoy** [comment]
Thanks @dsmithbauer, this seems like a pretty reasonable change to match Helm's behavior.  As a part of our 3rd-party contributors process we will run the pipeline initially, then after it passes will have a team member sign-off your PR with their GPG key (you'll be maintained as the author) and a final pipeline run/secondary team member approval will occur.

That being said, I added the `needs-tests` label to this PR as we have introduced a new feature without any test for it.  @YrrepNoj may be able to assist in recommending the right place to test this.

**Comment by dsmithbauer** [comment]
That sounds good to me. Thank you! I'm happy to add a built in test. I'll take a look.

**Comment by dsmithbauer** [comment]
@jeff-mccoy @YrrepNoj I've updated this PR with an example and test. However, I had issues trying to run the e2e tests, so I'm not entirely sure the new test works correctly.  If someone could try it or let me know know what to do to run the tests correctly it would be much appreciated!

Let me know if anything else is needed! Thanks.

**Comment by YrrepNoj** [comment]
@dsmithbauer I'll take a look at this today! We have a manual approval to run workflows for outside contributors so I'll take a look at the PR and get the tests running 😄

File: src/test/e2e/e2e_helm_test.go [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+	"fmt"
+	"os/exec"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestE2eExampleHelm(t *testing.T) {
+	//defer e2e.cleanupAfterTest(t)
```
YrrepNoj, line 13: Can you uncomment this deferred cleanup call :)

All of our test files are run sequentially on the same cluster on whatever host runner GitHub gives us. Since they're all running on the same cluster after tests that modify the cluster (doing an init and deploying a package like we are in this test) we defer this cleanup to get the cluster to a 'fresh' state. 

Without this cleanup call, depending on the order the `go test ./...` command runs the tests, the test that gets run immediately after this one will fail because you cannot `zarf init` on the same cluster twice currently.

src/test/e2e/e2e_helm_test.go
```diff
@@ -0,0 +1,31 @@
...
+	require.NoError(t, err, output)
+
+	path := fmt.Sprintf("../../../build/zarf-package-test-helm-releasename-%s.tar.zst", e2e.arch)
+
+	// Deploy the charts
+	output, err = e2e.execZarfCommand("package", "deploy", path, "--confirm")
+	require.NoError(t, err, output)
+
+	// Verify the configmap was properly templated
+	kubectlOut, _ := exec.Command("kubectl", "-A", "get", "pods", "--selector=app.kubernetes.io/name=etcd", "--no-headers").Output()
```
YrrepNoj, line 26: Probably doesn't have to be this PR (but feel free if you want to 😆) but we should probably update our internal `GetPods()` [function](https://github.com/defenseunicorns/zarf/blob/fdaff86627fa1e2bba5bd5cf612aa6f1d50dce31/src/internal/k8s/pods.go#L80) to take in a `metav1.ListOptions{}` object so we could use that function instead of having to use kubectl on the host.

I think that would be useful for other tests / things in the future, I'll add an issue in the backlog for that.

src/test/e2e/e2e_helm_test.go
dsmithbauer: Yes, oops! I meant to put that back before committing.

src/test/e2e/e2e_helm_test.go
```diff
@@ -0,0 +1,31 @@
...
+	//defer e2e.cleanupAfterTest(t)
+
+	//run `zarf init`
+	output, err := e2e.execZarfCommand("init", "--confirm")
+	require.NoError(t, err, output)
+
+	path := fmt.Sprintf("../../../build/zarf-package-test-helm-releasename-%s.tar.zst", e2e.arch)
+
+	// Deploy the charts
+	output, err = e2e.execZarfCommand("package", "deploy", path, "--confirm")
```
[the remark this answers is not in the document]
dsmithbauer: Added!

YrrepNoj:
The [Makefile](https://github.com/defenseunicorns/zarf/blob/fdaff86627fa1e2bba5bd5cf612aa6f1d50dce31/Makefile#L117) at the root of the project has a target called `test-e2e` that gets called when we run the tests. Part of that target builds the packages needed for the tests. 

Can you add a target to build the new helm-example package you just made (something like `make package-example-helm-releasename`) and then add a check to build that as part of the `test-e2e` target?

dsmithbauer:
@YrrepNoj I have fixed the Helm test as I was able to run it successfully locally.  Can you please retry the pipeline tests when you have a moment? Thanks.

dsmithbauer:
> @YrrepNoj I have fixed the Helm test as I was able to run it successfully locally. Can you please retry the pipeline tests when you have a moment? Thanks.

@jeff-mccoy It looks like the new test is passing and all that needs to be done when someone has a moment is to remove the needs-tests label and review.  Let me know if you need me to update anything else. Thanks!

YrrepNoj (approved):

jeff-mccoy:
@YrrepNoj I can do the git sign off process later on today/tonight.  Think we need a blocking flag for that in the future to avoid accidental merge (we did fine on this one just seeing it would be easy to mess that up). Something like "needs-git-sign-off"

jeff-mccoy (approved):
Lgtm, thanks for this contribution and thanks @YrrepNoj for the git sign-off on this.

PR #484, opened 2022-05-19
Update README.md

Madeline-UX:
@jeff-mccoy  to reflect our decision/ conversation when building website that Air Gap looks weird and AirGap is better.

## Description

Change Air Gap to AirGap

## Type of change

## Checklist before merging

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Zarf - DevSecOps for Air Gap Systems
+# Zarf - DevSecOps for AirGap Systems
 
 <img align="right" alt="zarf logo" src=".images/zarf-logo.png"  height="256" />
 


jeff-mccoy:
Looks like there are multiple instances in this document and a couple other places of "air gap" or "air-gapped"
Screen Shot 2022-05-19 at 5 27 31 PM
Screen Shot 2022-05-19 at 5 27 24 PM
.

Madeline-UX:
@jeff-mccoy yepp that is totally fair and something I only realized after I submitted. I was mostly using it as a test to show Jacob the UX flow for creating a PR against the read me - but I really do need to go fix all of them 

Or do we need to actually decide on which way is correct?

jeff-mccoy:
Closing this for now until we wrap up the larger https://docs.zarf.dev/docs/zarf-overview migration.

PR #486, opened 2022-05-19
Zarf Agent: ignore existing namespaces on zarf init

### LINKED ISSUE #485: Zarf Agent should not mutate existing resources in the cluster
If the mutating webhook has blanket authority to mutate everything (except Kube System) in the cluster, any existing K8s assets (including those deployed as a part of a K8s distro) will break when they cycle due to the images not being in the Zarf Registry.  Prior to applying the webhook, a label should be added to ignore existing namespaces.

Namespaces are the logical choice as it would be brittle (and much higher Kube API writing) to apply this to individual resources.  The drawback of this approach is that some namespaces such as `default` and `kube-system` should not be used or they will have problem with the webhook (if you don't manually specify the image paths and pull secrets configs as before).

jeff-mccoy:
Fixes #485

### DIFF
--- a/src/internal/helm/destroy.go
+++ b/src/internal/helm/destroy.go
@@ -3,6 +3,7 @@ package helm
 import (
 	"regexp"
 
+	"github.com/defenseunicorns/zarf/src/internal/k8s"
 	"github.com/defenseunicorns/zarf/src/internal/message"
 	"helm.sh/helm/v3/pkg/action"
 )
@@ -39,7 +40,7 @@ func Destroy(purgeAllZarfInstallations bool) {
 	// Iterate over all releases
 	for _, release := range releases {
 		if !purgeAllZarfInstallations && release.Namespace != "zarf" {
-			// Don't process releases outside the zarf namespace unless purgae all is true
+			// Don't process releases outside the zarf namespace unless purge all is true
 			continue
 		}
 		// Filter on zarf releases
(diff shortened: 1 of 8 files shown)


src/internal/packager/seed.go
```diff
@@ -65,6 +65,27 @@ func preSeedRegistry(tempPath tempPaths) {
 
 		// Setup zarf agent PKI
 		state.AgentTLS = pki.GeneratePKI(config.ZarfAgentHost)
+
+		namespaces, err := k8s.GetNamespaces()
+		if err != nil {
+			message.Fatalf(err, "Unable to get k8s namespaces")
+		}
+		// If this is a new Zarf cluster, mark existing namespaces as ignored for the zarf agent to prevent
```
YrrepNoj, line 73: super nit; `// If this ...` -> `// Since this ...`

src/internal/k8s/namespace.go
```diff
@@ -18,6 +18,15 @@ func GetNamespaces() (*corev1.NamespaceList, error) {
 	return clientset.CoreV1().Namespaces().List(context.TODO(), metaOptions)
 }
 
+func UpdateNamespace(namespace *corev1.Namespace) (*corev1.Namespace, error) {
+	message.Debugf("k8s.UpdateNamespace(%v)", namespace)
+
+	clientset := getClientset()
+	updateOptions := metav1.UpdateOptions{}
```
YrrepNoj, line 25: What do you think about adding this to the function parameters? 

One of the comments I had in #480 was pointing out that a nice-to-have update in the future is to add the options to the `GetPods()` call so the tests can do more specific queries instead of having to use kubectl.

Not sure it makes too much sense for an update call but wanted to mention it anyways.

src/internal/helm/destroy.go
```diff
@@ -57,5 +58,21 @@ func Destroy(purgeAllZarfInstallations bool) {
...
 
+	spinner.Updatef("Checking namespaces for zarf metadata")
+	if namespaces, err := k8s.GetNamespaces(); err != nil {
+		message.Error(err, "Unable to get k8s namespaces")
+	} else {
+		for _, namespace := range namespaces.Items {
+			if _, ok := namespace.Labels["zarf.dev/agent"]; ok {
+				spinner.Updatef("Removing Zarf Agent label for namespace %v", namespace.Name)
+				delete(namespace.Labels, "zarf.dev/agent")
+				if _, err = k8s.UpdateNamespace(&namespace); err != nil {
```
YrrepNoj, line 69: If `purgeAllZarfInstallations` is provided, does it make more sense to delete the namespace instead of just removing the label?

src/internal/k8s/namespace.go
jeff-mccoy: I like it, but let's keep it as a separate PR because I'd also like to refactor all the k8s work to leverage Go generics too.

src/internal/helm/destroy.go
jeff-mccoy: No this is a separate concept, we already delete namespaces, this is for mutating namespaces that existed outside of Zarf (prior to `zarf init`) that we need to put back.

jeff-mccoy: I did move the function though to a better location.

YrrepNoj (approved):
Looks good after the updates!

PR #488, opened 2022-05-23
implement component choice groups

### LINKED ISSUE #487: Add Configurable Way to Combine Component Deployment Prompts
Right now, each optional component is prompted one-by-one. There are some situations where it only makes sense to deploy one optional component out of a selection of options. (i.e. the package has a PLG component and an ELK component to serve as a logging solution but you only really need/want one of them). Right now, Zarf can be a little unintuitive since it will prompt for both, even if you already said yes to one of the options.

It would be nice if you could add the components into a logical group when creating the package that way the user is prompted to select one of the multiple components instead of prompted y/n for each one.

One potential solution is to add a `choice` key at the component level and then prompt all components that are a part of that 'choice' at the same time. 
Screen Shot 2022-05-19 at 12 21 55 AM

jeff-mccoy:
This PR provides a new deployment flow for components where the package author needs the ability to group a set of components and only allow a single one in the group to be deployed. Enforcement for CLI flags also exists to prevent a CLI flag from deploying multiple components from the same group. This PR also cleans up the business logic used to validate and select components to _hopefully_ be more readable. An example was added and Big Bang Core was moved to packages with a demonstration of this usage as well. Finally small pterm improvements were made for readability/improved flow including only showing credentials on init for components that were actually deployed.

Fixes #487 

## Type of change

## Checklist before merging


jeff-mccoy:
Prompt for choosing which component to deploy:
Screen Shot 2022-05-22 at 5 09 53 PM

After selection:
Screen Shot 2022-05-22 at 5 10 16 PM

Error message if attempting to specify more than one component from a choice group:
Screen Shot 2022-05-23 at 2 59 56 AM

jeff-mccoy:
Updated zarf init credential examples: 

`zarf init --confirm`
Screen Shot 2022-05-23 at 3 43 28 AM

`zarf init --confirm --components=logging`
Screen Shot 2022-05-23 at 3 46 10 AM

`zarf init --confirm --components=git-server`
Screen Shot 2022-05-22 at 5 28 40 PM

src/internal/packager/components.go
```diff
@@ -0,0 +1,207 @@
...
+			requested := isRequiredOrRequested(component, requestedComponentNames)
+
+			// If the user has not requested this component via CLI flag, then prompt them if not a choice group
+			if !requested && !userChoicePrompt {
+				requested = confirmOptionalComponent(component)
+			}
+
+			if requested {
+				// Mark deployment as appliance mode if this is an init config and the k3s component is enabled
+				if component.Name == "k3s" && config.IsZarfInitConfig() {
```
YrrepNoj, line 63: we have a `k8s.DistroIsK3s` const we can use here to check against instead of using a hardcoded string

src/internal/packager/components.go
```diff
@@ -0,0 +1,207 @@
...
+		} else {
+			// Otherwise, add the component name to the choice group list for later validation
+			choiceComponents = appendIfNotExists(choiceComponents, component.Name)
+		}
+
+		// Preserve component order
+		orderedKeys = appendIfNotExists(orderedKeys, key)
+
+		// Append the component to the list of components in the group
+		componentGroups[key] = append(componentGroups[key], component)
```
YrrepNoj, line 40: Reading through this just made me realize that we don't validate component names are unique within a package. Could you add that to this PR too (or write an issue if you don't think it fits in this PR's scope)?

YrrepNoj (changes requested):

packages/big-bang-core/zarf.yaml
```diff
@@ -78,3 +75,19 @@ components:
       # twistlock
       - registry1.dso.mil/ironbank/twistlock/console/console:22.01.840
       - registry1.dso.mil/ironbank/twistlock/defender/defender:22.01.840
+
+  - name: big-bang-core-demo
+    description: "Deploy a lightweight demo version of Big Bang Core not intended for production use"
```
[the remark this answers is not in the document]
jeff-mccoy: I don't want to have too many opinions around this and would like to leave it up to the community to figure out what makes sense.  That being said, I did change the wording a little bit.

PR #491, opened 2022-05-25
add sign-off label to banned labels list to prevent merge

YrrepNoj:
This flag will be used when an outside contributor makes a PR and a member of the team still needs to sign-off on their commits.


jeff-mccoy (approved):
lgtm.  we should do a PR at some point to just update https://github.com/yogevbd/enforce-label-action/blob/master/src/main.ts#L19 to support regex or at least prefix-matching so we could just do `needs-`

jeff-mccoy (approved):

PR #498, opened 2022-06-01
Use GoReleaser for GitHub release workflow

YrrepNoj:
Uses GoReleaser to build, release, and publish the tagged releases pushed to the repo.

Through this pipeline update, GoReleaser will publish a brew recipe to the [Defense Unicorns Tap](https://github.com/defenseunicorns/homebrew-tap) for the tagged release.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -50,6 +50,7 @@ brews:
     tap:
       owner: defenseunicorns
       name: homebrew-tap
+      token: "{{ .Env.HOMEBREW_TAP_GITHUB_TOKEN }}"
     commit_msg_template: "Brew formula update for {{ .ProjectName }} version {{ .Tag }}"
     homepage: "https://zarf.dev/"
     description: "An airgap delivery tool!"
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,0 +1,61 @@
+# TODO @JPERRY REMOVE THESE LINES AND ENSURE THE 'GITHUB_TOKEN' token is populated in the repo.
+# .goreleaser.yaml
+env_files:
+  github_token: ~/.path/to/my/github_token
+
+
+# Make sure to check the documentation at https://goreleaser.com
+before:
+  hooks:
+    - go mod tidy
+
+universal_binaries:
+  - replace: false
+
+builds:
+  - env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - darwin
+    dir: src
+    ldflags:
+      - -s -w -X github.com/defenseunicorns/zarf/src/config.CLIVersion={{.Version}}
+    goarch:
+      - amd64
+      - arm64
+archives:
+  -
+    format: binary
+    replacements:
+      darwin: Darwin
+      linux: Linux
+
+checksum:
+  name_template: 'checksums.txt'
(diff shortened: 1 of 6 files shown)


.goreleaser.yaml
```diff
@@ -0,0 +1,62 @@
...
+  use: github-native
+
+release:
+  github:
+    owner: defenseunicorns
+    name: zarf
+  prerelease: auto
+  mode: append
+
+brews:
```
jeff-mccoy, line 48: So if I'm reading this right, this is what will bump a new version on that repo, correct?

.goreleaser.yaml
```diff
@@ -0,0 +1,62 @@
...
+  - name: zarf
+    tap:
+      owner: defenseunicorns
+      name: homebrew-tap
+      token: "{{ .Env.HOMEBREW_TAP_GITHUB_TOKEN }}"
+    commit_msg_template: "Brew formula update for {{ .ProjectName }} version {{ .Tag }}"
+    homepage: "https://zarf.dev/"
+    description: "An airgap delivery tool!"
+
+blobs:
```
jeff-mccoy, line 58: This will drop all the same binaries in S3 as the prior workflow method, correct?

jeff-mccoy:
Overall this looks really good and will be a huge boost to our UX for getting started with Zarf. I do think we should include the init packages so users can jump right in to using Zarf if they want.  This will take a change in [cmd/initialize.go](https://github.com/defenseunicorns/zarf/blob/master/src/cmd/initialize.go) to check the CWD first for an init package and then try falling back to `os.Executable()` if that fails.

.goreleaser.yaml
YrrepNoj: This will replace the recipe in our tap repository so that `brew install zarf` will always install the latest release of zarf.
The user will probably have to do a `brew update` depending on the last time they naturally ran that.

YrrepNoj: This will replace the recipe in our tap repository so that `brew install zarf` will always install the latest release of zarf.
The user will probably have to do a `brew update` depending on the last time they naturally ran that.

.goreleaser.yaml
YrrepNoj: That is the end goal! 

This comment made me realize we accidentally stopped building the init-packages in the current PR since GoReleaser doesn't handle that for us. That will be fixed with the next commit, thanks!

YrrepNoj: That is the end goal! 

This comment made me realize we accidentally stopped building the init-packages in the current PR since GoReleaser doesn't handle that for us. That will be fixed with the next commit, thanks!

YrrepNoj:
`zarf init` attempts to use an init-package in the fully evaluated (following symlinks) `os.Executable` directory if there is no available package to use in the current working directory.

Screen Shot 2022-06-03 at 1 11 19 PM

YrrepNoj:
@jeff-mccoy This should be ready for review now. I left a small note in the 'nerd notes' section of the README that talks about the Homebrew Tap and how it was created. I'll add more docs about that as @Madeline-UX and I go through and improve the docs in both this repo and the docs website.

jeff-mccoy (approved):
great work, this will be important for a few features going forward.

PR #505, opened 2022-06-06
Better CLI Help Messages

YrrepNoj:
Improve the CLI `--help` message outputs.

### DIFF
--- a/src/cmd/initialize.go
+++ b/src/cmd/initialize.go
@@ -18,8 +18,14 @@ import (
 var initCmd = &cobra.Command{
 	Use:     "init",
 	Aliases: []string{"i"},
-	Short:   "Deploys the gitops service or appliance cluster on a clean linux box",
-	Long:    "Flags are only required if running via automation, otherwise the init command will prompt you for your configuration choices",
+	Short:   "Prepares a k8s cluster for the deployment of Zarf packages",
+	Long: "Uses a local 'zarf-init' package found either in the directory you executed " +
+		"the command from or the directory that the Zarf binaries exists. " +
+		"Injects a docker registry as well as other optional useful things (such as a git server " +
+		"and a logging stack) into a k8s cluster under the 'zarf' namespace " +
+		"to support future application deployments. \n" +
+		"If you do not have a k8s cluster already configured, this command will give you " +
+		"the ability to install a cluster locally.\n",
 	Run: func(cmd *cobra.Command, args []string) {
 		zarfLogo := message.GetLogo()
 		_, _ = fmt.Fprintln(os.Stderr, zarfLogo)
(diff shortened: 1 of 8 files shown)


src/cmd/prepare.go
```diff
@@ -72,7 +72,8 @@ var prepareComputeFileSha256sum = &cobra.Command{
 var prepareFindImages = &cobra.Command{
 	Use:     "find-images",
 	Aliases: []string{"prep"},
-	Short:   "evaluates components in a zarf file to identify images specified in their helm charts and manifests",
+	Short: "Evaluates components in a zarf file to identify images specified in their helm charts and manifests.\n" +
+		"Requires the `--repo-chart-path` flag to be provided.",
```
jeff-mccoy, line 76: this flag is optional, but needed for big bang and should be "-p ./chart".  Technically could be for any helm chart hosted in git, but we focus on Big Bang

src/cmd/initialize.go
```diff
@@ -46,8 +52,7 @@ var initCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(initCmd)
 	initCmd.Flags().BoolVar(&config.DeployOptions.Confirm, "confirm", false, "Confirm the install without prompting")
-	initCmd.Flags().StringVar(&config.DeployOptions.Components, "components", "", "Comma-separated list of components to install.  Adding this flag will skip the init prompts for which components to install")
-
+	initCmd.Flags().StringVar(&config.DeployOptions.Components, "components", "", "Comma-separated list of components to instal.")
```
jeff-mccoy, line 55: instal -> install

src/cmd/initialize.go
```diff
@@ -18,8 +18,14 @@ import (
 var initCmd = &cobra.Command{
 	Use:     "init",
 	Aliases: []string{"i"},
-	Short:   "Deploys the gitops service or appliance cluster on a clean linux box",
-	Long:    "Flags are only required if running via automation, otherwise the init command will prompt you for your configuration choices",
+	Short:   "Prepares a k8s cluster for the deployment of Zarf packages",
+	Long: "Uses a local 'zarf-init' package found either in the directory you executed " +
+		"the command from or the directory that the Zarf binaries exists. " +
```
jeff-mccoy, line 23: "the command from or the directory" reads a little odd I think

jeff-mccoy (changes requested):

src/cmd/prepare.go
YrrepNoj: Thanks, that makes sense. I updated the message (as well as corrected the message type).

jeff-mccoy (approved):

PR #508, opened 2022-06-09
fix kustomize path for big-bang core package

jeff-mccoy:
A small typo was introduced with the big bang core example -> packages move in #488, this corrects that.  Since we do not have the ability to run this in E2E right now due to IB credentials and how SUPER long it would take to fully test, including screenshots in the PR after testing from M1 -> Ubuntu AMD64 remote machine using https://github.com/alexellis/k3sup.

### DIFF
--- a/packages/big-bang-core/kustomization/core-light/kustomization.yaml
+++ b/packages/big-bang-core/kustomization/core-light/kustomization.yaml
@@ -1,5 +1,5 @@
 bases:
-  - ../core-prod
+  - ../core-standard
 
 configMapGenerator:
   - name: environment


jeff-mccoy:
Good package creation now:
Screen Shot 2022-06-09 at 1 48 12 PM

Working package deploy options:
Screen Shot 2022-06-09 at 1 55 33 PM

Successful package deployment:
Screen Shot 2022-06-09 at 1 58 07 PM

Helm releases pulled from gitea:
Screen Shot 2022-06-09 at 1 58 41 PM

PR #511, opened 2022-06-11
Debug Message Cleanup

jeff-mccoy:


Makefile
```diff
@@ -70,88 +70,31 @@ dev-agent-image:
...
-package-example-component-scripts: ## Create component script example
-	cd examples/component-scripts && ../../$(ZARF_BIN) package create --confirm && mv zarf-package-* ../../build/
+	@test -s ./build/zarf-package-dos-agmes-$(ARCH).tar.zst || $(ZARF_BIN) examples/game build -a $(ARCH) --confirm
+	
+	@test -s ./build/zarf-package-component-scripts-$(ARCH).tar.zst || $(ZARF_BIN) examples/component-scripts build -a $(ARCH) --confirm
 
-.PHONY: package-example-component-choice
-package-example-component-choice: ## Create component choice example
-	cd examples/component-choice && ../../$(ZARF_BIN) package create --confirm && mv zarf-package-* ../../build/
+	@test -s ./build/zarf-package-component-choice-$(ARCH).tar.zst || $(ZARF_BIN) examples/component-choice build -a $(ARCH) --confirm
```
YrrepNoj, line 86: NIT; can we turn `build` into `build/` to make it obvious that's a directory.

YrrepNoj, line 86: NIT (that I'm a little bit more passionate about); I think `$(ZARF_BIN) package create examples/{EXAMPLE} build/ -a $(ARCH) --confirm)` is a lot more clear. Shortcutting away the 'package create' doesn't really seem necessary in a Makefile.

YrrepNoj (changes requested):

YrrepNoj:
Right now we can do `zarf package deploy path/to/package`. I think it would be good UX if we were consistent and gave users the ability to do `zarf init path/to/package` as well.

examples/tiny-kafka/zarf.yaml
```diff
@@ -2,8 +2,6 @@ kind: ZarfPackageConfig
 metadata:
   name: kafka-strimzi-demo
```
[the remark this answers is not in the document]
jeff-mccoy: yes I did

src/internal/packager/compose.go
```diff
@@ -42,50 +46,105 @@ func validateOrBail(component *types.ZarfComponent) {
...
+
+	// Merge variables.
+	for key, variable := range override.Variables {
+		target.Variables[key] = variable
+	}
 
-	if src.Description != "" {
-		target.Description = src.Description
+	// Merge scripts.
+	target.Scripts.Before = append(target.Scripts.Before, override.Scripts.Before...)
```
YrrepNoj, line 79: @jeff-mccoy I'm ~~pretty sure~~ _really hoping_ the k3s tests are almost always failing on this branch because we aren't merging the `retry: true` value to the new k3s component. Since it's not retrying it's failing really early because the OS isn't ready to start up the k3s service the first couple 'attempts'.

jeff-mccoy: hmmm looks like retry is still running after all?

src/internal/k8s/tunnel.go
```diff
@@ -106,6 +106,8 @@ func (tunnel *Tunnel) Connect(target string, blocking bool) {
 		tunnel.resourceName = "zarf-loki-stack-grafana"
 		tunnel.localPort = PortLogging
 		tunnel.remotePort = 3000
+		// Start the logs with something useful
+		tunnel.urlSuffix = `/monitor/explore?orgId=1&left=%5B"now-12h","now","Loki",%7B"refId":"Zarf%20Logs","expr":"%7Bnamespace%3D%5C"zarf%5C"%7D"%7D%5D`
```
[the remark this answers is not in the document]
jeff-mccoy: It is generated by Grafana and the hope is to provide a direct link to something useful for users if they do `zarf connect logging`.  It should open (after login) to a view similar to this:

Screen Shot 2022-06-30 at 3 23 44 PM

jeff-mccoy:
All prior PRs are now merged, all that is left is some debug message cleanup 😆

src/internal/k8s/secrets.go
```diff
@@ -87,7 +87,8 @@ func GenerateRegistryPullCreds(namespace, name string) *corev1.Secret {
 }
 
 func GenerateTLSSecret(namespace, name string, conf types.GeneratedPKI) (*corev1.Secret, error) {
-	message.Debugf("k8s.GenerateTLSSecret(%s, %s, %v", namespace, name, conf)
+	message.Debugf("k8s.GenerateTLSSecret(%s, %s", namespace, name)
+	message.Debug(message.JsonValue(conf))
```
Racer159, line 91: Why not put this inline like replace TLS secret?

src/internal/k8s/state.go
```diff
@@ -35,14 +35,15 @@ func LoadZarfState() types.ZarfState {
...
+	message.Debug(message.JsonValue(state))
 
 	return state
 }
 
 // SaveZarfState takes a given state and makepersists it to the zarf/zarf-state secret
 func SaveZarfState(state types.ZarfState) error {
-	message.Debugf("k8s.SaveZarfState(%v)", state)
+	message.Debugf("k8s.SaveZarfState()")
+	message.Debug(message.JsonValue(state))
```
Racer159, line 46: Same here with the 2-line message.JsonValue()

src/internal/k8s/state.go
```diff
@@ -35,14 +35,15 @@ func LoadZarfState() types.ZarfState {
 		_ = json.Unmarshal(match.Data[ZarfStateDataKey], &state)
 	}
 
-	message.Debug(state)
+	message.Debug(message.JsonValue(state))
```
YrrepNoj, line 38: Probably should make this a `message.Debugf( ... )` so that we can describe what is getting logged. It would probably take me a little longer to realize that the random json is the zarf state

YrrepNoj (commented):
`grep -rnsI "%v" . --include="*.go"` returns a handful of places that are still using `%v` instead of `%#v`

None of them are logging anything more complex than a slice so it probably doesn't matter but it might be nice to update them just for consistency sake 🤷

jeff-mccoy:
> `grep -rnsI "%v" . --include="*.go"` returns a handful of places that are still using `%v` instead of `%#v`
> 
> None of them are logging anything more complex than a slice so it probably doesn't matter but it might be nice to update them just for consistency sake 🤷

Yeah I think those were missed with other merges, can update them too

jeff-mccoy:
Screen Shot 2022-07-18 at 5 05 15 PM

jeff-mccoy:
All requested changes made

Racer159 (approved):
Approved

YrrepNoj (approved):

Issue #489, opened 2022-05-23
Add a CLI flag to override the custom Zarf variables


YrrepNoj:
Right now we can define custom variables for the components that gets templated to the manifests/helm charts that the component uses. Currently, the value has to be defined in the zarf.yaml when the package is built.

An example of custom component variables is in [this example](https://github.com/defenseunicorns/zarf/blob/master/examples/component-variables/zarf.yaml)
```

kind: ZarfPackageConfig
metadata:
  name: component-variables
  description: "Test component to demonstrate script timeout feature"

components:
  # Demonstrates injecting custom variables into a K8s resource, e.g. ###ZARF_DOG###
  - name: variable-example
    required: true
    variables:
      dog: "woof"
      cat: "meow"
    manifests:
      - name: variable-example-configmap
        files:
          - simple-configmap.yaml
```

<br />
<br />

It would be very useful to be able to provide the values for the custom variables during runtime. Something like `zarf package deploy --values dog=woof2`.

> NOTE: Right now the  custom variables are component specific, this means that (at the current moment at least) the custom variable names do not need to be unique throughout that package. That might need to change depending on how the solution to this issue gets implemented.

YrrepNoj:
When this feature request was being discussed, it was pointed out that this feature would make it a lot easier to to rotate around passwords and certs of deployed packages.

jeff-mccoy:
I think https://www.terraform.io/language/values/variables#variable-definition-precedence has a lot of well-established patterns around doing runtime based variable overrides.

dsmithbauer:
Lol. We were just discussing the need for this exact feature again today on my team.  I had even searched the open issues earlier this morning before this was written. I'd love to see this feature implemented as it would save having to repackage  just to change some variables and eliminate the need to maintain different packages for different targets where the only difference is a few variables too. +1

anoncam:
@jeff-mccoy [this](https://github.com/shebashio/terraform-modules) you have access to this, I'll be finding the appropriate ones for a zarf initial set of modules and open source that repo

Issue #464, opened 2022-04-26
Anti-Template for values that need to be changed before building


YrrepNoj:
Zarf utilizes templates to hot-swap values when doing a `zarf package deploy`. We should add a way to have template/placeholder text that gets recognized and handled when doing a `zarf package create`.

For example, our eks package uses an [eks.yaml](https://github.com/defenseunicorns/zarf/blob/master/packages/distros/eks/eks.yaml) that has the name for the nodegroup and cluster commented out. This value is commented out because it needs to be unique each time the eks yaml is used which means before we create the package we need to remember to add/update the name field.

It would be very useful if we would check for an 'anti-template' (maybe ###ZARF_CHANGEME### ?) while creating packages to ensure things like this will get caught at build-time instead of running into errors during run-time.

Extra credit points if the `zarf package create` is able to prompt the user for what they want the value to be.

jeff-mccoy:
@YrrepNoj this is the issue we discussed for allowing package create time variables that must be supplied for the zarf agent changes on release.

jeff-mccoy:
Closing in favor #613

PR #516, opened 2022-06-15
Update target location for ekcstl for Darwin amd64 builds

YrrepNoj:
We determine the system architecture through 'uname -m' which outputs x86_64 for Mac's w/ intel chips.
This changes the target location name to match that output so the script can find/use it later.

### DIFF
--- a/packages/distros/eks/zarf.yaml
+++ b/packages/distros/eks/zarf.yaml
@@ -38,3 +38,5 @@ components:
         - "./eksctl create cluster --dry-run -f eks.yaml"
         - "sleep 15"
         - "./eksctl create cluster -f eks.yaml"
+      after:
+        - "./eksctl utils write-kubeconfig -c CHANGE_ME"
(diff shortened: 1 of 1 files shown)


YrrepNoj:
@jeff-mccoy This one should be safe to merge in whenever you have time

jeff-mccoy (approved):

packages/distros/eks/zarf.yaml
```diff
@@ -37,4 +37,6 @@ components:
       before:
         - "./eksctl create cluster --dry-run -f eks.yaml"
         - "sleep 15"
-        - "./eksctl create cluster -f eks.yaml"
\ No newline at end of file
+        - "./eksctl create cluster -f eks.yaml"
+      after:
+        - "./eksctl utils write-kubeconfig -c CHANGE_ME"
```
YrrepNoj, line 42: This is another place where the #464 and/or #489 would be super useful.

jeff-mccoy:
Insert mean words here.

jeff-mccoy (approved):

PR #519, opened 2022-06-15
Download and init script

### LINKED ISSUE #512: Getting started init script
As a team we created a script that it is useful for new engineers to setup and try Zarf really quickly. Usage is similar to:

```
curl -sL https://raw.githubusercontent.com/defenseunicorns/zarf/installZarf.sh | bash
```

It helps to:

* Create a /home/<user>/.zarf directory or /tmp/zarfxxx temporary directory
* Detects and downloads the proper binary and package depending on the platform (mac, linux, x64, 32, arm, etc.)
* Do check on the downloaded checksum files hashes
* Optionally, zarf init into the current k8s context

I see this as a helpful maybe in the getting started section. **Do you think this is something the project will benefit from?** We would be eager to create the initial PR with the script code.

mahomedalid:
## Description

We used this script in our CI/CD pipelines and to make it easier for new engineers to integrate zarf with their existing development loops.

## Related Issue

Fixes #512

## Type of change

- [X ] New feature (non-breaking change which adds functionality)

## Checklist before merging


jeff-mccoy:
Thanks this actually provides a lot of insight into pain points with getting started we should look at @Madeline-UX @YrrepNoj.  Going to sanity check this against our work with the `brew install` flow at https://zarf.dev/install.  Note the install page isn't complete right now as the sget package source isn't working yet.

It might take us a couple days to get this, but really appreciate the contribution!

mahomedalid:
> 

Cool! It is worth to mention two different ways this could be used, one as getting started, and the other as a CI/CD helper (as part of a CI/CD process). Although we could create examples (and extensions/tasks) of how to integrate this with Azure DevOps, GitHub and GitLab for the CI part, usually CD is very different depending on how the disconnected environment looks like. I guess, it could remain as reference of a way to do it.

jeff-mccoy:
Now that https://zarf.dev/install is complete and we have a way to easily install the Zarf CLI via https://brew.sh/, Zarf can auto-download the init tarball for you and deploy a package from an OCI registry via https://github.com/sigstore/cosign, I think we've met the intent of this script.  Before I close, would like your thoughts @mahomedalid.  For CI/CD on github, homebrew is already installed, on other CI systems I'm not sure.

mahomedalid:
I think it is fine even if it is a different approach, great work, thanks.

PR #544, opened 2022-06-19
Fix/nested composed components

YrrepNoj:
Fix nested components importing

### DIFF
--- a/src/internal/packager/compose.go
+++ b/src/internal/packager/compose.go
@@ -2,7 +2,6 @@ package packager
 
 import (
 	"path/filepath"
-	"strings"
 
 	"github.com/defenseunicorns/zarf/src/config"
 	"github.com/defenseunicorns/zarf/src/internal/message"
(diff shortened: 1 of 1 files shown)


src/internal/packager/compose.go
```diff
@@ -99,27 +105,36 @@ func getImportedComponent(importComponent types.ZarfComponent) (component types.
...
 		}
 	}
 
 	return component
 }
 
 // Reads the locally imported zarf.yaml
-func getSubPackage(component *types.ZarfComponent) (importedPackage types.ZarfPackage) {
-	path := filepath.Join(component.Import.Path, config.ZarfYAML)
-	utils.ReadYaml(path, &importedPackage)
```
YrrepNoj, line 112: @jeff-mccoy This is why it wasn't working. We were still using the relative path from the inner-composed perspective and weren't catching the `err` out of the `ReadYaml` call that was telling us the path was bad 🥲

jeff-mccoy: Nice! Quick look on mobile, this looks way cleaner, bless you for using filepath now 😂. I'll dig through it more in a couple hours.  Thanks!!

jeff-mccoy (approved):

PR #550, opened 2022-06-21
Add support for package deploy via cosign sget 

YrrepNoj:
Provides support for specifying a remote signed package (via cosign/sget) to deploy.  Initial use will be for https://zarf.dev/install/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,16 +41,16 @@ clean: ## Clean the build dir
 	rm -rf build
 
 build-cli-linux-amd: build-injector-registry
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="$(BUILD_ARGS)" -o build/zarf src/main.go
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="$(BUILD_ARGS)" -o build/zarf main.go
 
 build-cli-linux-arm: build-injector-registry
-	CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -ldflags="$(BUILD_ARGS)" -o build/zarf-arm src/main.go
+	CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -ldflags="$(BUILD_ARGS)" -o build/zarf-arm main.go
 
 build-cli-mac-intel: build-injector-registry
-	GOOS=darwin GOARCH=amd64 go build -ldflags="$(BUILD_ARGS)" -o build/zarf-mac-intel src/main.go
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -17,7 +17,7 @@ ifneq ($(UNAME_S),Linux)
 endif
 
 CLI_VERSION := $(if $(shell git describe --tags), $(shell git describe --tags), "UnknownVersion")
-BUILD_ARGS := -s -w -X 'github.com/defenseunicorns/zarf/src/config.CLIVersion=$(CLI_VERSION)'
+BUILD_ARGS := -s -w -X 'github.com/defenseunicorns/zarf/src/config.CLIVersion=$(CLI_VERSION)'       # TODO: Add some way to pass the sget string as a value here too
```
YrrepNoj, line 20: @jeff-mccoy Do you have any good ideas on how to do this dynamically?

src/internal/packager/common.go
```diff
@@ -163,6 +169,32 @@ func HandleIfURL(packagePath string, shasum string, insecureDeploy bool) (string
...
+	destinationFile, err := os.Create(localPackagePath)
+	if err != nil {
+		message.Fatal(err, "Unable to create the destination file")
+	}
+	defer destinationFile.Close()
+
+	// If this is a DefenseUnicorns package, use an internal sget public key
+	if strings.HasPrefix(sgetPackagePath, "sget://defenseunicorns") {
+		os.Setenv("DU_SGET_KEY", config.SGetPublicKey)
+		config.DeployOptions.SGetKeyPath = "env://DU_SGET_KEY"
```
YrrepNoj, line 188: Cosign can load the public key as an environment variable now. This must be new since when we last looked at it!

https://github.com/sigstore/cosign/blob/29361993239764ae63c3017198cc80ff5816c08f/pkg/blob/load.go#L46

Makefile
jeff-mccoy: we can remove this todo now, yes?

src/internal/packager/common.go
```diff
@@ -187,6 +187,11 @@ func handleSgetPackage(sgetPackagePath string) (string, func()) {
 		os.Setenv("DU_SGET_KEY", config.SGetPublicKey)
 		config.DeployOptions.SGetKeyPath = "env://DU_SGET_KEY"
 	}
+
+	// Remove the 'sget://' header for the actual sget call
+	sgetPackagePath = strings.Split(sgetPackagePath, "sget://")[1]
```
jeff-mccoy, line 192: Could we use https://pkg.go.dev/strings#TrimPrefix here?

jeff-mccoy (approved):

PR #575, opened 2022-06-29
QOL: Extending `zarf connect` usability 

YrrepNoj:
## Zarf Connect Default Stability
Add logic so default `zarf connect` commands will auto-detect when the standard ports are already being used and use a random open one instead of failing the command.

> Using a port other than 45003 since that port was already used on my host machine
Screen Shot 2022-06-29 at 3 46 13 PM

## Zarf Connect Usability Improvement
Add a utility function so users can figure out which `zarf connect` commands are available from packages that have already been installed. Super useful for people who are not familiar with the packages that has already been deployed to a cluster.

> List all connect options found within the cluster
Screen Shot 2022-06-29 at 3 43 06 PM

> Filter for connect options that contains part of an optional provided argument filter
Screen Shot 2022-06-29 at 3 44 17 PM


jeff-mccoy:
We should probably look at these together 😆 https://github.com/defenseunicorns/zarf/pull/511/commits/b350f4ec31399e12698e266bb72f6111994e0129

YrrepNoj:
Closing as this was also implemented in #586

PR #581, opened 2022-07-01
Converts E2E tests to user journey tests & adds `--no-progress` flag for cleaner CI logs

jeff-mccoy:
Migrate E2E tests to a [journey-style](https://martinfowler.com/bliki/UserJourneyTest.html) E2E test suite that focuses on categories of behavior/interaction with Zarf.  Tests are now grouped by either an outcome, i.e. "deploy the dos games" or a typical use of Zarf, "i.e." upload and interact with the git-server.

This PR also introduces the `--no-progress` flag to disable progress bars and spinners to keep logs more useful for CI.

NOTE: the E2E tests now execute against a single cluster much like a regular user might vs continually creating and destroying the Zarf assets in the cluster.  This isn't without risk as the shared runners we use are [limited to 2 CPU and 7 GBs of RAM](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) so we have to be aware of resource constraints.  For that reason each test setup dumps the current port bindings and cluster resource allocations.


.github/workflows/test-k3d.yml
```diff
@@ -22,7 +22,12 @@ jobs:
             ~/go/pkg/mod
           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
           restore-keys: |
-            ${{ runner.os }}-go-        
+            ${{ runner.os }}-go-
+
+      - name: "Install K3d and deploy a cluster"
+        run: |
+          wget -q -O - https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
```
YrrepNoj, line 29: Thoughts on using a specific k3d release tag (like we did w/ KinD)? 

It might be annoying to debug why our tests are failing if some breaking update happens in k3d main.

.github/workflows/test-k3s.yml
```diff
@@ -34,4 +34,4 @@ jobs:
         # NOTE: "PATH=$PATH" preserves the default user $PATH. This is needed to maintain the version of go installed
         #       in a previous step. This test run will use Zarf to create a K3s cluster, and a brand new cluster will be
         #       used for each test
-        run: sudo env "PATH=$PATH" TESTDISTRO=k3s make test-e2e
+        run: sudo env "PATH=$PATH" CI=true APPLIANCE_MODE=true make test-e2e
```
YrrepNoj, line 37: I'm surprised we needed this CI flag but my eyes don't lie from the last tests 🤷 

Although, is there are reason we're only providing it for k3s?

src/test/e2e/common.go
```diff
@@ -44,107 +42,41 @@ func getCLIName() string {
...
-		require.NoError(t, err, "unable to remove file when cleaning up after a test")
-	}
-	e2e.filesToRemove = []string{}
-
-	// Kill background processes spawned during the test
-	e2e.killBackgroundProcesses(t)
+// setup actions for each test
+func (e2e *ZarfE2ETest) setup(t *testing.T) {
+	t.Log("Test setup")
+	utils.ExecCommandWithContext(context.TODO(), true, "sh", "-c", "kubectl describe nodes |grep -A 99 Non\\-terminated")
```
YrrepNoj, line 48: Could you leave a comment describing what these two commands are doing?

src/test/e2e/common.go
```diff
@@ -44,107 +42,41 @@ func getCLIName() string {
...
+// teardown actions for each test
+func (e2e *ZarfE2ETest) teardown(t *testing.T) {
+	t.Log("Test teardown")
+	// Kill background processes spawned during the test
 	for _, cmd := range e2e.cmdsToKill {
 		if cmd.Process != nil {
-			err := cmd.Process.Kill()
-			require.NoError(t, err, "unable to kill background cmd when cleaning up after a test")
+			if err := cmd.Process.Kill(); err != nil {
+				t.Logf("unable to kill process: %v", err)
```
YrrepNoj, line 59: Just logging this instead of failing fast?

YrrepNoj (changes requested):

.github/workflows/test-k3s.yml
jeff-mccoy: Yeah sudo was dropping the env variable.  Not totally sure why.  I thought sudo env fixed that but I might be misremembering.

src/test/e2e/common.go
jeff-mccoy: I was torn on this one but it's more a ci env issue than zarf issue I think if this occurs.  I haven't seen it occur yet fwiw.

YrrepNoj: Fair enough. If it ends up being a problem we'll see it almost immediately later anyways.

.github/workflows/test-k3d.yml
jeff-mccoy: That's a good point I was just copy/pasting both of their official docs

jeff-mccoy: So I did make them consistent, but went the opposite direction.  I agree this could pose a more flaky CI/Cd, but it now matched how https://zarf.dev/install/ works and how a real user would experience Zarf, so that matches better with the change to Journey tests I think.  I'd rather have the test break here and slow our debugging down than break for a user.

YrrepNoj (approved):
Thanks for breaking this out of your other PR. I'm super excited that the tests are running so quickly now.

Issue #454, opened 2022-04-21
request to remove gitea's cpu limit or bump it from 1 --> 3 cpu cores


neoakris:
**Is your feature request related to a problem? Please describe.**
While working on a bigbang airgap quickstart

I'd run the following command to quickly populate gitea with a copy of repos sourced from repo1
```shell
time ls -1 ~/repos | grep -v bigbang | xargs --replace={item} -P 50 /bin/bash -c 'cd ~/repos/{item}; git remote add gitea https://$GIT_USER:$GIT_PASS@git.bigbang.dev/platform-one/{item}.git; git checkout main; git push gitea main; git push gitea --tags'
```
^-- The above uploads 32 repos sourced from bigbang's airgap artifact to gitea in parallel.

I had to create an nginx ingress controller ingress object with the following annotations to make gitea uploads run smoothly.
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: utility-ingress
  namespace: zarf
  annotations:
    nginx.ingress.kubernetes.io/proxy-body-size: "0" #removes size limit (for HTTPS uploads to gitea using git cli, otherwise there will be a HTTP 413 request entity too large error)
    nginx.ingress.kubernetes.io/proxy-buffering: "off" # (change logic from async to --> sync, and implicitly disables rate limiting)
    nginx.ingress.kubernetes.io/proxy-request-buffering: "off" # (same as above, and implicitly disables rate limiting)
    nginx.ingress.kubernetes.io/proxy-read-timeout: "900" #parallel git repo upload taxes gitea server 
    nginx.ingress.kubernetes.io/proxy-send-timeout: "900" #which slows down responses 
    nginx.ingress.kubernetes.io/proxy-connect-timeout: "900" #increasing timeouts to 15mins, avoids an ephemeral 504 error
...
```

The above annotations allowed me to stop all HTTP error codes (was having HTTP 413 & HTTP 504) error codes before finding the right combination of annotations to configure nginx-ingress controller that ships by default on an RKE2 install. 

The above parallel upload command takes 7 min to upload, where there's no network bottleneck.
k top node shows gitea is maxed out to 1 CPU core, it's limit. 

`k top pod -n=zarf`
```
NAME                                    CPU(cores)   MEMORY(bytes)
zarf-docker-registry-656c4f8c7d-lvlwb   1m           77Mi
zarf-gitea-0                            1000m        964Mi
```
[...]

PR #586, opened 2022-07-02
Disable docker for k3s test

### LINKED ISSUE #580: ERROR:  Unable to add login credentials for the gitops registry
### Environment
Device and OS: OSX v12.4 (Intel)
App version: v0.19.6
Kubernetes distro being used: v1.22.9-eks-a64ea69
Other:

### Steps to reproduce
1. ensure Docker desktop is NOT running
2. `zarf init`

### Expected result
zarf init does it's thing

### Actual Result
```     ERROR:  Unable to add login credentials for the gitops registry ```

image

### Severity/Priority
High

### Additional Context
Confirmed bug in debugging session with @jeff-mccoy and @YrrepNoj

### PR DESCRIPTION
**Pull request by jeff-mccoy**
## Description
Ensure docker isn't called from Zarf or Crane

## Related Issue
Fixes #580

## Type of change

## Checklist before merging

### DIFF
--- a/.github/workflows/test-k3s.yml
+++ b/.github/workflows/test-k3s.yml
@@ -24,6 +24,9 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
 
+      - name: "Disable Docker"
+        run: sudo systemctl stop docker --now
+
       - name: "Build CLI"
         run: make build-cli-linux
       

### CODE REVIEW

File: src/internal/packager/deploy.go [hunk 1]
```diff
@@ -251,12 +241,31 @@ func deployComponents(tempPath tempPaths, component types.ZarfComponent) {
 	}
 
 	if hasImages {
-		images.PushToZarfRegistry(tempPath.images, component.Images)
+		// Try image push up to 3 times
+		for retry := 0; retry < 3; retry++ {
```
YrrepNoj, line 245: Now that we're adding retries into the image & repo pushes, we should consider doing the pushes in a Goroutine. That seams like the ideal use case for a Goroutine.

We would just need to make sure all the routines are finished by the time we start applying the `component.Charts`

src/internal/k8s/tunnel.go
```diff
@@ -141,6 +136,9 @@ func (tunnel *Tunnel) Connect(target string, blocking bool) {
 			}
 		}
 
+		// Dump the tunnel URL to the console for other tools to use
+		fmt.Print(url)
```
YrrepNoj, line 140: Given that the spinner logs this url and that the function returns the url to the calling function, I don't think we need to be printing the url to stdout too.

src/test/e2e/21_logging_test.go
```diff
@@ -12,17 +14,20 @@ func TestLogginga(t *testing.T) {
 	e2e.setup(t)
 	defer e2e.teardown(t)
 
+	// Get a random local port for this instance
+	localPort, _ := k8s.GetAvailablePort()
+
 	// Establish the port-forward into the logging service
-	err := e2e.execZarfBackgroundCommand("connect", "logging", "--cli-only")
+	err := e2e.execZarfBackgroundCommand("connect", "logging", fmt.Sprintf("--local-port=%d", localPort), "--cli-only")
```
YrrepNoj, line 21: Can we have one of the tests not provide the `--local-port` flag for the 'default` connection. 

The tunnel logic is now handling that and dynamically using an open port and I want to make sure we are testing that functionality too.

src/internal/packager/deploy.go
YrrepNoj: Not asking this to be done in this PR. More so trying to see if you agree so I can create an issue in the backlog.

YrrepNoj (changes requested):

src/internal/message/spinner.go
```diff
@@ -60,10 +60,8 @@ func (p *Spinner) Debugf(format string, a ...any) {
 }
 
 func (p *Spinner) Stop() {
-	if p.spinner != nil && p.spinner.IsActive {
```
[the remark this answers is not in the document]
jeff-mccoy: Yeah this was really a debug thing for me when I was doing that because we had some random spinner duplications.  It was a real problem and we still should have some type of singleton logic here eventually because you actually can't have more than one spinner at once, but it feels like that should exist in the upstream vs here. Also, needed this to not be there so we could properly call a `Stop()` when `Success()` or `Fatal()` weren't applicable.

src/internal/packager/deploy.go
jeff-mccoy: Yeah I thought about that too, agree it needs to at least be in a separate issue--but I do have some concerns.  I know w #454 was talking about increasing CPU limits for gitea because they were parallelizing via a script for the git repos and saturate CPU. We already saturate CPU for the registry when pushing sequentially so think we need to look at that carefully.  Also note that a runner is [only 2 cores](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)--so we run out of CPU real fast if we try to get fancy with autoscaling.

src/internal/k8s/tunnel.go
jeff-mccoy: This is a particular issue for needing to push to `stdout` vs `stderr`, I agree I don't _love this_, but functionally speaking, we really needed to do this if someone wanted to script out `zarf connect registry` for example now that the ports aren't predictable.  I suppose you could argue they should just use a specific port like we do for tests, but then they'd need to guarantee port availability first and won't have the logic we do already handy.

jeff-mccoy: Very much open to ideas on that.

YrrepNoj: What if we hold the print until we return back to the `connect.go` command and only print if `--cli-only` is provided (or make a new flag that is explicit for returning the port to stdout?

jeff-mccoy: Updated to meet the intent of what you asked, could not push it back to the `cmd` package because the call needs to be blocking or the tunnel would die.

YrrepNoj: That works for me.

Sometime in the future I might propose moving the logic to handle when the `local-port == 0` to the `k8s.NewTunnel()` function instead of happening inside the `tunnel.Connect()` function that way we can print it out in the `cmd.Connect()` and just be more explicit in general.

YrrepNoj (approved):
I really like the usability of having the default connection commands using dynamic ports. Whoever came up with that had a really good idea 😉

src/internal/k8s/tunnel.go
jeff-mccoy: Yeah that's a totally valid point too, I think there's still more cleanup overall in that `cmd.Connect()` area.

PR #591, opened 2022-07-06
Feature: Add only filters for local OS and cluster architecture component filtering

### LINKED ISSUE #593: Add `only` filter to zarf components
Need to implement a filter for components based off of the cluster architecture and local operating system.  

- [ ] On `package create` filter out components with non-matching architectures 
- [ ] Always include components with OS filters, we will not know what the target runtime OS
- [ ] On `package deploy` filter by OS, matching the current runtime OS

Discussed with @RothAndrew, @YrrepNoj , @brianrexrode and @lucasrod16 to confirm the final structure below:

```yaml
components:
  - name: k3s
    only:
      localOS: linux
      cluster:
        architecture: amd64
```

jeff-mccoy:
## Description
Adds the ability to filter by local OS or cluster architecture.

## Related Issue
Fixes #593 

## Type of change

## Checklist before merging


src/internal/packager/deploy.go
```diff
@@ -52,7 +52,7 @@ func Deploy() {
 
 	// Load the config from the extracted archive zarf.yaml
 	spinner.Updatef("Loading the zarf package config")
-	if err := config.LoadConfig(tempPath.base + "/zarf.yaml"); err != nil {
+	if err := config.LoadConfig(tempPath.base+"/zarf.yaml", true); err != nil {
```
YrrepNoj, line 55: Can we do `filepath.Join()` here instead of concatenating strings. It'll make our lives easier once we start supporting the Windows binary.

src/internal/packager/inspect.go
```diff
@@ -32,7 +32,7 @@ func Inspect(packageName string) {
 	utils.ColorPrintYAML(text)
 
 	// Load the config to get the build version
-	if err := config.LoadConfig(tempPath.base + "/zarf.yaml"); err != nil {
+	if err := config.LoadConfig(tempPath.base + "/zarf.yaml", false); err != nil {
```
YrrepNoj, line 35: `filepath.Join()`

YrrepNoj (changes requested):

src/internal/packager/inspect.go
```diff
@@ -32,7 +34,7 @@ func Inspect(packageName string) {
 	utils.ColorPrintYAML(text)
 
 	// Load the config to get the build version
-	if err := config.LoadConfig(tempPath.base + "/zarf.yaml", false); err != nil {
+	if err := config.LoadConfig(tempPath.base+"/zarf.yaml", false); err != nil {
```
YrrepNoj, line 37: Did you mean to use the `configPath` var from above on line#25?

jeff-mccoy:
Demonstrate the only filter working on a mac (since we don't currently test that), `zarf init` does not present k3s as a component at all.

Screen Shot 2022-07-11 at 12 10 57 PM

src/internal/packager/compose.go
```diff
@@ -56,11 +56,22 @@ func getParentComponent(childComponent types.ZarfComponent, everGrowingComposePa
...
+				filterArch = childComponent.Only.Cluster.Architecture
+			}
+
+			// Only add this component if it is valid for the target architecture.
+			if filterArch == "" || filterArch == targetArch {
+				parentComponent = importedComponent
+				break
+			}
 		}
 	}
```
Racer159, line 77: Isn't there a chance that you could reach here without parentComponent being set?  Not sure what go would do here if you try to access parentComponent.Import.Path on a non-match.

jeff-mccoy: So I played around with this for a while.  There was an existing issue where an invalid component name could be specified, added an error handler for that.  As far as if you specify the wrong arch, it actually won't make it to this code at all, it would fail at the `config.isCompatibleComponent()` check in the config package instead.  That wouldn't be easy to add more logic around, but we do a couple things to help.  First, we specify in the JSON Schema what is valid and not as far as arch values, second, on package create the component would be removed and a log message in debug mode explaining why.  

## JSON Schema validation error
Screen Shot 2022-07-11 at 2 41 53 PM

## JSON Schema showing only valid options in drop-down
Screen Shot 2022-07-11 at 2 42 01 PM

## Debug warning of component removal
Screen Shot 2022-07-11 at 2 40 08 PM

YrrepNoj (approved):

Racer159 (approved):
Reviewed changes and approved

PR #599, opened 2022-07-07
Ignore non-code changes in CI for PRs

jeff-mccoy:
## Description
Do not run tests for docs-only changes.  See https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks for more details on how this works. This config should skip E2E tests and replace with a "no test required" job if changes only occurred to the following matching file patterns:

- '**.md'
- '**.jpg'
- '**.png'
- '**.gif'
- '**.svg'
- 'adr/**'
- 'docs/**'


.github/workflows/skip-k3d.yml
```diff
@@ -0,0 +1,17 @@
+name: e2e-k3d
+on:
+  pull_request:
+    paths:
```
YrrepNoj, line 4: I think the 'skip' tests need to use `paths-ignore` rather than `paths`.

The way I understand it, if someone commits a PR with docs AND code changes, the `skip-k3d` workflow will run because the docs changes trigger the `paths` conditional. This will pass the workflows and we could end up merging bad code because the real `e2e-k3d` test will still be running.

YrrepNoj (changes requested):

.github/workflows/skip-k3d.yml
YrrepNoj: And the real tests should be using `paths` instead of `paths-ignore`

YrrepNoj (approved):

jeff-mccoy:
## With docs-only changes:
Screen Shot 2022-07-08 at 6 00 11 PM

## With code-only changes:
Screen Shot 2022-07-08 at 6 02 46 PM

## With code and docs changes:
Screen Shot 2022-07-08 at 6 01 33 PM

PR #600, opened 2022-07-07
Add tmpdir command flag and refactor CommonOptions

### LINKED ISSUE #576: Add ability to specify an alternate tmp path folder
The Zarf deployment currently partially deploys to the `/tmp` directory and is several GBs in size. This may be an issue on machines with a small `/tmp` directory (especially ones that cannot be reconfigured to be larger).

Therefore an ability for the user to specify an alternate `/tmp` path folder would be helpful for these situations.

Racer159:
## Description

Added the ability to specify a `--tmpdir` in the places that `utils.MakeTempDir` is used.

## Related Issue

Fixes #576 

## Type of change

## Checklist before merging


Racer159:
Marked this PR as ready.  I didn't write a test for the init command --tmpdir because it would increase the time for e2e tests but I can add that if desired,

src/cmd/prepare.go
```diff
@@ -20,11 +21,11 @@ var prepareCmd = &cobra.Command{
 }
 
 var prepareTransformGitLinks = &cobra.Command{
-	Use:   "patch-git [HOST] [FILE]",
+	Use:     "patch-git [HOST] [FILE]",
```
jeff-mccoy, line 24: was this whitespace intentional, linting maybe?

src/config/config.go
```diff
@@ -49,6 +49,9 @@ var (
 	// CLIVersion track the version of the CLI
 	CLIVersion = "unset"
 
+	// GlobalOptions tracks user-defined values that apply across commands
+	GlobalOptions types.ZarfGlobalOptions
```
jeff-mccoy, line 53: I don't _love_ `GlobalOptions`, how much would you hate `CLIOptions`, `UserOptions`, `UserFlags`, anything besides `GlobalOptions`?  😆 I also don't have a super compelling reason, just global feels weird.

jeff-mccoy: I do like how the common options are now separate from the workflow-specific options

src/test/e2e/28_temp_directory_deploy_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// run `zarf package deploy` with a specified tmp location
+	otherTmpPath := "/tmp/othertmp"
+
+	e2e.cleanFiles(otherTmpPath)
+
+	e2e.setup(t)
+	defer e2e.teardown(t)
+
+	path := fmt.Sprintf("build/zarf-package-dos-games-%s.tar.zst", e2e.arch)
```
jeff-mccoy, line 25: I'd love to just move this test up to before the `20` init test because we can test against any of those packages that don't need a cluster.  The existing of a cluster is irrelevant for this test.

src/test/e2e/28_temp_directory_deploy_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	_ = os.Mkdir(otherTmpPath, 0750)
+
+	// Deploy the game
+	stdOut, stdErr, err := e2e.execZarfCommand("package", "deploy", path, "--confirm", "--tmpdir", otherTmpPath, "--log-level=debug")
+	require.Contains(t, stdErr, otherTmpPath, "The other tmp path should show as being created")
+	require.NoError(t, err, stdOut, stdErr)
+
+	// Get a random local port for this instance
+	localPort, _ := k8s.GetAvailablePort()
```
jeff-mccoy, line 35: Same as above, I'd look at using the same logic as `01_component_choice_test.go`, only without testing the variables themselves as they wouldn't matter for this test.

src/cmd/prepare.go
Racer159: Yes the linter did that (same with the other whitespace changes).  I included it in my commit since this file may have not been touched by the linter yet.  I think my linter settings are correct, but can work to correct it if not.

src/test/e2e/28_temp_directory_deploy_test.go
Racer159: Ok yeah was debating how "thorough" to be but that makes more sense.  I'll go ahead and fix this.

src/test/e2e/28_temp_directory_deploy_test.go
Racer159: 👍

src/cmd/prepare.go
Racer159: Played around with just running `pre-commit run -a` and on my machine it adds spaces to a lot of files...  Should we do a file cleanup PR to bring everything into standards with linting?

Racer159: (there are also a few more rules we may have to disable like `gochecknoglobals` as that is flagging on these config options for me)

src/config/config.go
jeff-mccoy: CommonOptions works for me, thanks!

src/cmd/prepare.go
Racer159: (discussed this offline, proceeding)

jeff-mccoy (approved):
LGTM, thanks!

PR #615, opened 2022-07-11
add zarf connect list for connect service lookup

### LINKED ISSUE #462: Add an easy way to see what configured 'connect' services are avaliable
Zarf packages can include service definitions with specific annotations and labels to allow create shortcuts for users to use to quickly do `zarf connect` commands. Right now those shortcuts are shown once the package is finish deploying but there is no way to use Zarf to figure out what those shortcuts are. After deploying multiple packages, or simply doing something after deploying a package, it's pretty easy to forget what all the existing shortcuts are.

A sub-command along the lines of `zarf connect list` that could print out all the connection shortcuts would be really helpful. Since Zarf uses the labels on the service to know what service to connect to, it should be pretty trivial to find all the services with the `zarf.dev/connect-name:` labels. 

Since at the shortcut is just a label, it is possible to get this information with a more robust kubectl command. I think having a quick and simple zarf command would still add a lot of value.

jeff-mccoy:
## Description
Adds new command `zarf connect list` to show all zarf connect targets (including zarf init-related targets).  Also updates (very stale docs) and switches to raw generator output for docs to improve doc overall consistency with the codebase.

## Related Issue
Fixes #462

## Type of change

## Checklist before merging


jeff-mccoy:
Screen Shot 2022-07-11 at 5 34 40 PM

jeff-mccoy:
The CLI docs were severely out of date, updated them all as a part of this as we needed to do so to update the new command CLI docs properly.

Racer159 (dismissed):
This change looks good to me.  We could add a check similar to the verify schema stuff for the CLI docs as well.  This may be worth an issue since there are other docs (like the CONTRIBUTING.md file) that can get out of sync.

src/cmd/internal.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/defenseunicorns/zarf/src/internal/k8s"
+	"github.com/defenseunicorns/zarf/src/internal/message"
+	"github.com/defenseunicorns/zarf/src/types"
+	"github.com/spf13/cobra"
+	"github.com/spf13/cobra/doc"
+)
+
+var internalCmd = &cobra.Command{
+	Use:     "internal",
+	Aliases: []string{"dev", "int"},
```
YrrepNoj, line 19: I suggest not including `int` since it could accidentally be trigger from a typo of `init` and lead to confusion. Up to your discretion though so feel free to resolve.

YrrepNoj (changes requested):

PR #621, opened 2022-07-13
Move variable definitions from components to packages

### LINKED ISSUE #489: Add a CLI flag to override the custom Zarf variables

### LINKED ISSUE #613: Variable all the things!
Move component-level variables to the package level and allow package string-templating for the Zarf.yaml and package variables templated out manifests.

### Zarf.yaml string templating
- [x] Validation must require uppercase and snake case
- [x] special type of string replace that can only be specified during `package create`
- [x] do not appear in the `variables` or `constants` lists
- [x] can only template string values in a `zarf.yaml`
- [x] run after component composition so cannot change component import paths or names
- [x] check for all matching variables and error if one is not set by the user on `package create`

```yaml
components:
  - name: zarf-agent
    required: true
    images:
      - "###ZARF_VAR_AGENT_IMAGE###"
    manifests:
      - name: zarf-agent
        namespace: zarf
```

### Constant declaration by package authors
- [x] validation must require uppercase and snake case for all variables
- [x] specified as an array under the `constants` key
- [x] only set during `package create`
- [x] cannot be overriden during `package deploy`
- [x] are used with `ZARF_CONST` prefix

```yaml
constants:
  # Must be all-caps, snake-case
  - name: AGENT_IMAGE
    value: "defenseunicorns/zarf-agent:v0.19.5"
```
In the manifest will appear as `###ZARF_CONST_AGENT_IMAGE###`

### Variable declaration by package deployers
- [x] validation must require uppercase and snake case for all variables
- [x] specified as an array under the `variables` key
- [x] set via `default` key
- [x] set via`--set` flags on `package deploy`
- [x] user prompt on `package deploy` if `prompt: true`
- [x] are used with `ZARF_VAR_` prefix

```yaml
variables:
  # Must be all-caps, snake-case
  - name: AGENT_IMAGE
    default: "defenseunicorns/zarf-agent:v0.19.5"

  # This variable will template manifests on deploy and will prompt if the user doesn't specify by flag
  - name: PROMPT_ON_DEPLOY
    prompt: true
```
In the manifest will appear as `###ZARF_VAR_AGENT_IMAGE###` and `###ZARF_VAR_PROMPT_ON_DEPLOY###`

Issue planning coordinated with @RothAndrew @YrrepNoj @Racer159

Racer159:
## Description

Move variable declarations to the package level and enable additional features such as `default` and `prompt`.

## Related Issue

Fixes #613 --> Moves variables to the package level instead of the component level
Fixes #489  --> Adds a cli flag to override variables in a package definition

## Type of change

## Checklist before merging


Racer159:
We'll see how far this gets :D (will probably fail but it should kinda work)

src/internal/packager/prepare.go
```diff
@@ -39,6 +39,10 @@ func FindImages(baseDir, repoHelmChartPath string) {
 		message.Fatal(err, "Unable to read the zarf.yaml file")
 	}
 
+	if err := config.SetActiveVariables(config.ZarfYAML, false); err != nil {
```
YrrepNoj, line 42: Should we add the `--set` flag to the `zarf prepare find-images` command? If the images are variable'd out it might get missed if there is no default value.

src/internal/template/template.go
```diff
@@ -87,33 +87,35 @@ func (values Values) Apply(component types.ZarfComponent, path string) {
...
-	for template, value := range mappings {
-		// Keys are always uppercase in the format ###ZARF_KEY###
-		template = strings.ToUpper(fmt.Sprintf("###ZARF_%s###", template))
-		utils.ReplaceText(path, template, value)
+	for key, value := range config.VariableMap {
+		// Variable keys are always uppercase in the format ###ZARF_VAR_KEY###
+		templateMap[strings.ToUpper(fmt.Sprintf("###ZARF_VAR_%s###", key))] = value
 	}
+
+	message.Debug(templateMap)
```
YrrepNoj, line 118: Can we also log the `builtinMap` like we originally were here before the variable name change?

Also, I personally think it's better to log the variableNames (or some type of added context) that way it's easier to figure out what something is when going through the logs
`message.Debugf("templateMap = %#v", templateMap)`

docs/4-user-guide/1-the-zarf-cli/100-cli-commands/zarf_tools_registry_login.md
```diff
@@ -6,6 +6,13 @@ Log in to a registry
 zarf tools registry login [OPTIONS] [SERVER] [flags]
 ```
 
+### Examples
```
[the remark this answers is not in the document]
jeff-mccoy: Until you give me a smarter way to do it 👀

src/internal/template/template.go
Racer159: The template map contains the builtinmap. Would you want to print the same info twice? (agree with `message.Debugf("templateMap = %#v", templateMap)`

src/internal/packager/prepare.go
Racer159: I tested this and asked about it here:  https://github.com/defenseunicorns/zarf/issues/613 What happens is it will print out the variable name (i.e. `###ZARF_AGENT_IMAGE###`).  `zarf prepare find-images` doesn't template the components (didn't before when it was just `AGENT_TAG`) so setting anything here has no effect on output without changing that.  The main reason variables are needed here though is if you variablize imports for components (like I did in the `examples/composable-package-variables` case)

docs/4-user-guide/1-the-zarf-cli/100-cli-commands/zarf_tools_registry_login.md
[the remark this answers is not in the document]
Racer159: Your wish is my command @jeff-mccoy

src/internal/packager/prepare.go
Racer159: (though I suppose it would be useful if you then `--set` those variablized imports to something else)

Racer159: And added

src/internal/template/template.go
YrrepNoj: Oh whoops I missed that `templateMap` contained `builtinMap`. Thanks!

YrrepNoj (approved):

src/config/config.go
```diff
@@ -276,3 +320,21 @@ func isCompatibleComponent(component types.ZarfComponent, filterByOS bool) bool
...
+
+	prompt := &survey.Input{
+		Message: "Please provide a value for '" + variable.Name + "'",
+	}
+
+	if variable.Default != nil {
+		prompt.Default = *variable.Default
+	}
+
+	_ = survey.AskOne(prompt, &value)
```
YrrepNoj, line 337: We probably shouldn't ignore the error here. A `ctrl+c` would return some form of an `&errors.errorString{s:"interrupt"}` error. Since the err is getting ignored this leads to the deployment continuing with the prompted value just being left as an empty string.

YrrepNoj (changes requested):

src/config/config.go
Racer159: Fixed

YrrepNoj:
@jeff-mccoy @Racer159 

I was working on updating the eks package to use the new package variables and ran into a bug (or at least something we should consider a little more):


The following `zarf package create` command output shows that the `CLUSTER_NAME` was modified in the `after` script section but doesn't actually get added as a default value in the `variables` 
`zarf package create packages/distros/eks --set CLUSTER_NAME=jperry-cluster-1`
Screen Shot 2022-07-19 at 12 04 30 AM


That means when I `zarf package deploy` this package and get prompted to input the value I don't get to see what the default is.
Screen Shot 2022-07-19 at 12 06 08 AM

---

This brings up two questions for me:
1. Should we be adding/updating the `--set` values as defaults during a `zarf package create`? (
   - I would say yes but I'm curious about other opinions
3. Should we throw an error when someone tries to `--set` a variable that has `prompt: true`?
   - In this case the `CLUSTER_NAME` is used in other manifests but since it already got templated into the `zarf.yaml` `after` section it would fail to deploy if the user didn't enter the SAME EXACT VALUE as when they ran the `zarf package create --set .....`

jeff-mccoy:
I'll wait to review until we get the other changes discussed in the issue reflected in this PR, thanks.

docs/4-user-guide/1-the-zarf-cli/100-cli-commands/zarf_completion.md
```diff
@@ -30,4 +30,4 @@ See each sub-command's help for details on how to use the generated script.
 * [zarf completion powershell](zarf_completion_powershell.md)	 - Generate the autocompletion script for powershell
 * [zarf completion zsh](zarf_completion_zsh.md)	 - Generate the autocompletion script for zsh
 
-###### Auto generated by spf13/cobra on 17-Jul-2022
+###### Auto generated by spf13/cobra on 25-Jul-2022
```
jeff-mccoy, line 33: I do wish there was a way to avoid triggering updates on every page--even ones without an actual update 😭

examples/package-variables/README.md
```diff
@@ -0,0 +1,71 @@
...
+    required: true
+    images:
+      - "###ZARF_PKG_VAR_PROMPT_ON_CREATE###"
+```
+
+> ⚠️ **NOTE:** *You can only template string values in this way as non-string values will not marshal/unmarshal properly*
+
+> ⚠️ **NOTE:** *If you use `--confirm` and do not `--set` all of the varaibles you will receive an error*
+
+> ⚠️ **NOTE:** *You cannot template the component import path using package variables*
```
jeff-mccoy, line 71: These are really fantastic docs for this new feature!  I wonder if we should consider and ADR for this and maybe just include some of the contents of these docs / the issue as it's a biggish change.  Thoughts @YrrepNoj / @Racer159?

src/cmd/tools.go
```diff
@@ -100,7 +100,11 @@ func init() {
 	archiverCmd.AddCommand(archiverDecompressCmd)
 
 	cranePlatformOptions := config.GetCraneOptions()
-	registryCmd.AddCommand(craneCmd.NewCmdAuthLogin())
+
+	craneLogin := craneCmd.NewCmdAuthLogin()
+	craneLogin.Example = ""
```
jeff-mccoy, line 105: Haha that's a good way to get rid of that nasty cobra output

src/config/config.go
```diff
@@ -202,10 +206,97 @@ func LoadConfig(path string, filterByOS bool) error {
...
+	for key, value := range packageVariables {
+		// Variable keys are always uppercase in the format ###ZARF_PKG_VAR_KEY###
+		templateMap[strings.ToUpper(fmt.Sprintf("###ZARF_PKG_VAR_%s###", key))] = *value
+	}
+
+	return utils.ReloadYamlTemplate(&active, templateMap)
+}
+
+// SetActiveVariables handles setting the active variables used to template component files.
+func SetActiveVariables() error {
```
jeff-mccoy, line 245: Can we make a new `variables.go` or something in the `config` pkg for all this logic?  It's getting a little messy.

jeff-mccoy:
Will finish reviewing this tomorrow, so far looks awesome!

src/config/config.go
Racer159: Created a new `variables.go` file and flattened the nested if statements to hopefully make this more readable.

YrrepNoj (dismissed):
Love the new feature and changes

jeff-mccoy:
sorry for the new merge conflicts @Racer159 👀

Racer159:
> sorry for the new merge conflicts @Racer159 eyes

Should be fixed now.

src/types/component.go
```diff
@@ -122,5 +116,5 @@ type ZarfDataInjection struct {
 // ZarfImport structure for including imported zarf components
 type ZarfComponentImport struct {
 	ComponentName string `yaml:"name,omitempty"`
-	Path          string `yaml:"path"`
+	Path          string `yaml:"path" jsonschema:"pattern=^(?!.*###ZARF_PKG_VAR_).*$"`
```
jeff-mccoy, line 119: One of the things we've added elsewhere in the codebase for potentially non-obvious regex is a link in the comments to a regex tool with some validations to show how it works.  

https://github.com/defenseunicorns/zarf/blob/v0.20.0/src/internal/utils/image.go#L5

I realize that we haven't done this for any other pattern matching in the schemas and that the schema doc generator links you to a regex test site, but if we don't have a link like above, could you put a small comment aabout what's happening/why for someone who isn't familiar with regex?

src/config/variables.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	return utils.ReloadYamlTemplate(&active, templateMap)
+}
+
+// SetActiveVariables handles setting the active variables used to template component files.
+func SetActiveVariables() error {
+	SetVariableMap = CommonOptions.SetVariables
+
+	for _, variable := range active.Variables {
```
jeff-mccoy, line 52: This reads a little odd to me with all the multiple negative conditins.  Some rewrite suggestions:

```golang
func SetActiveVariables() error {
	SetVariableMap = CommonOptions.SetVariables

	for _, variable := range active.Variables {
		_, present := SetVariableMap[variable.Name]

		// Variable is present, no need to continue checking
		if present {
			continue
		}

		// First set default if it exists
		if variable.Default != nil {
			SetVariableMap[variable.Name] = *variable.Default
		}

		// Prompt is enabled
		if variable.Prompt {
			// If Prompt and Confirm are set, throw an error
			if CommonOptions.Confirm {
				return fmt.Errorf("variable '%s' must be '--set' when using the '--confirm' flag", variable.Name)	
			}

			// Prompt the user for input 
			SetVariableMap[variable.Name], err := promptVariable(variable.Name, variable.Default)
			if err != nil {
				return err
			} 
		}
	}

	return nil
}
```

src/types/package.go
```diff
@@ -20,11 +22,24 @@ type ZarfMetadata struct {
...
 	User         string `yaml:"user"`
 	Architecture string `yaml:"architecture"`
 	Timestamp    string `yaml:"timestamp"`
 	Version      string `yaml:"version"`
 }
+
+// ZarfPackageVariable are variables that can be used to dynaically template K8s resources.
+type ZarfPackageVariable struct {
+	Name    string  `yaml:"name" jsonschema:"description=The name to be used for the variable,pattern=^[A-Z_]+$"`
+	Default *string `yaml:"default,omitempty" jsonschema:"description=The default value to use for the variable"`
```
jeff-mccoy, line 37: Can you explain why we need a pointer here?  It seems that the null value of "" would be fine to work with vs checking the pointer for nil.

Racer159: That depends on if we want to not template non-defaulted non-prompted values or not.  If a user does not specify something for one and default doesn't have the ability to be nil then you couldn't tell the difference between an unspecified `""` or an actual value defaulted to `""`.  This is probably an edge case though and we could just say everything is defaulted to "".

src/types/component.go
Racer159: Added a link

src/config/variables.go
Racer159: Refactored to something close to this.  Definitely agree it is better.

src/types/component.go
jeff-mccoy: I must be going blind...
Screen Shot 2022-07-28 at 1 11 17 PM
.

YrrepNoj: It's there for me. Since you're seeing the `Changed since last view` label you probably just need to refresh

Racer159: image
It is there for me too.

examples/package-variables/README.md
Racer159: Added a note in here about `prompt` and `default` defaults.

src/config/variables.go
Racer159: Removed the confirm check since default will always be set now.

src/types/package.go
Racer159: Removed the pointer here.

adr/0006-package-variables.md
```diff
@@ -0,0 +1,31 @@
...
+
+Date: 2022-07-28
+
+## Status
+
+Accepted
+
+## Context
+
+Currently, Zarf only allows variables to be specified within components whcich introduces the following limitations:
```
jeff-mccoy, line 11: whcich

adr/0006-package-variables.md
```diff
@@ -0,0 +1,31 @@
...
+
+ - Variables are scoped to the component making reuse between components impossible
+ - Variables cannot be used for elements within the package definition itself
+ - Variables can only be set at create time, not by a user at deploy time
+
+This forces a package creator to copy information within their package/component definitions and also requires them to make bespoke packages per environment even if only small changes are needed to the overal spec (such as changing a domain name).
```
jeff-mccoy, line 17: overal

adr/0006-package-variables.md
```diff
@@ -0,0 +1,31 @@
...
+- Package Variables (specified by using `###ZARF_PKG_VAR_*###` in the package definition) allow package creators to template the same information multiple times within the package definition or dynamically specify values or defaults in constants and variables.
+
+## Consequences
+
+This makes it easier to build a single package that will apply to multiple environments and helps package creators to develop automation around their packages while keeping their package definitions DRY.  Choosing to have constants *and* variables also allows us to reduce potential confusion from package deployers who would otherwise be able to accidentally override values that are meant to be static.
+
+As drawbacks, this method does tie us to only templating string values for the time being and means that we will have to think through what should be variablizable in the future (for example variables cannot be used in component import paths because this would introduce a lot of fragility for not much perceived user benefit).
```
jeff-mccoy, line 31: Could we mention something about the risk of being more imperative depending on how much is made available to users to change and the drawback of such a change?

src/internal/packager/compose.go
```diff
@@ -24,119 +27,115 @@ func GetComponents() (components []types.ZarfComponent) {
...
 
-func GetComposedComponent(childComponent types.ZarfComponent) types.ZarfComponent {
-	message.Debugf("packager.GetComposedComponent(%+v)", childComponent)
+// GetComposedComponent recursively retrieves a composed zarf component
+func GetComposedComponent(parentComponent types.ZarfComponent) types.ZarfComponent {
+	message.Debugf("packager.GetComposedComponent(%+v)", parentComponent)
 
 	// Make sure the component we're trying to import cant be accessed
-	validateOrBail(&childComponent)
+	validateOrBail(&parentComponent)
```
jeff-mccoy, line 37: I know we discussed this a lot and went back and forth on the philosophy of imported/inherited relationships.  Since this is literally the opposite wording here, can we put some substantive comment/explanation since we've all been confused over this for so long and just changing from child->parent will be even more confusing for future devs.

jeff-mccoy (changes requested):

adr/0006-package-variables.md
Racer159: fixed

adr/0006-package-variables.md
Racer159: fixed

adr/0006-package-variables.md
Racer159: Added

src/internal/packager/compose.go
Racer159: Added this:

```
// For composed components, we build the tree of components starting at the root and adding children as we go;
// this follows the composite design pattern outlined here: https://en.wikipedia.org/wiki/Composite_pattern
// where 1 component parent is made up of 0...n composite or leaf children.
```

jeff-mccoy (dismissed):
🥳  lgtm.  super excited to see this change.  This one is bigger, so think we should have me & @YrrepNoj approve.

jeff-mccoy (approved):

jeff-mccoy:
Everything lgtm, will wait for @YrrepNoj to approve since it's a bigger PR.  Thanks @Racer159--outstanding work!

YrrepNoj (approved):
Super excited for these changes!

PR #572, opened 2022-06-28
Fix/zarf agent flux mutation

jeff-mccoy:
## Description
- Proper binding of CREATE and UPDATE for Pods and Gitrepositories
- Add open source flux example for testing
- Add E2E for Flux
- Add retry logic for zarf connect/tunnel establishment
- Fix test stability issues with zarf connect actions and data injection tests
- Move to self-hosted runners

Fixes #564
Fixes #616
Fixes #610 

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist before merging
- [x] Tests have been added/updated as necessary (add the `needs-tests` label)

PR #624, opened 2022-07-15
Verify background command has started before continuing with tests

YrrepNoj:
The bottom of this file describes how `-1` is returned when the process has not started yet: https://go.dev/src/os/exec_posix.go

### DIFF
--- a/src/test/e2e/common.go
+++ b/src/test/e2e/common.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"context"
+	"errors"
 	"os"
 	"os/exec"
 	"runtime"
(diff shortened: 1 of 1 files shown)


Racer159:
Does this have an accompanying Issue #?

jeff-mccoy:
I actually don't think this is the right solution for this problem.  I have an alternative in #572--don't think an issue exists, this is related to the continual test flakes on background connections we've been seeing. The issue with this way is it doesn't actually address the core problem--when we "background" `zarf connect`, we can't wait for the return so that the test can keep running, and on the CLI the return won't occur until the connection closes.  The solution in the other PR does bypass the call in Cobra (a couple lines), but it tests the rest of the functionality properly.

jeff-mccoy:
Closed as #572 is now merged in.

PR #628, opened 2022-07-16
Gracefully handle zarf shorthand removal

jeff-mccoy:
This re-adds some behavior to tell users the `zarf zarf-package-thing-arch.tar.zst` is no longer valid.

### DIFF
--- a/src/cmd/root.go
+++ b/src/cmd/root.go
@@ -3,9 +3,11 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


jeff-mccoy:
## Current experience on master (not yet released)
image

## New behavior
Screen Shot 2022-07-15 at 11 16 00 PM
Screen Shot 2022-07-15 at 11 15 08 PM

Racer159 (approved):
Assuming the tests pass (which they should). This is simple and clean.

Issue #643, opened 2022-07-25
data injection logic is flakey


YrrepNoj:
As can be seen with a lot of our [recent workflow failures](https://github.com/defenseunicorns/zarf/actions?query=is%3Afailure+workflow%3Atest-k3s+workflow%3Atest-k3d+workflow%3Atest-kind) we are starting to consistently see failures due to the data-injection e2e test failing. 

This is most likely happening because the internal logic to handle injecting data into the pods is using a `kubectl` command through an `os/exec` call to the system. Since we are directly running the kubectl CLI we are passing off the execution to the host machine which can have inconsistent results due to the host runner having limited resources and trying to do its own thing to optimize process execution.

> deploy.go at line ~362
<img width="836" alt="Screen Shot 2022-07-25 at 10 45 29 AM" src="https://user-images.githubusercontent.com/18587823/180805079-35204dc6-fba5-4b38-a7b6-987568f29cc2.png">

Instead of executing the kubectl CLI we should switch over to handle the cp directly through the k8s api.

jeff-mccoy:
Going to take a peak at this for a little bit, not rewrite yet, but just stability

Racer159:
> Going to take a peak at this for a little bit, not rewrite yet, but just stability

🏔️

PR #153, opened 2021-11-03
Multi-distro support

jeff-mccoy:
Multi-distro support
- add ARM support
- adds in-memory embedded registry to serve registry image from "seed" config spec
- handles helm charts install/upgrade/rollback/uninstall via helm directly
- dynamically generate helm charts for raw manifests 
- add tunnel capability to access zarf endpoints without ingress config
- flattens image collection to single tarball
- remove root / linux requirements for zarf deployments
- add K8s distro-detection and distro-specific seed image behaviors
- document local registry per https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/generic/1755-communicating-a-local-registry
- reorganize logging, remove logrus, replace with pterm, add progress cli feedback
- add support for stacking helm values files
- normalize log levels: info, warn, debug, trace 
- update go mod deps
- move zarf state to a k8s secret in the zarf namespace
- support zarf destroy for 3rd-party clusters
- move crane image cache to sub folder under home
- add support for multiple users with htpasswd
- add path key for git-based helm charts
- remove need for PKI prompts on default init
- add storage class tracking in zarf state for multi-distro compatibility

PR #644, opened 2022-07-25
Data injection stability improvements

jeff-mccoy:
This PR addresses some race conditions present with data injection since the addition of the Multi-distro support in #153.  This also addresses some test flakes experienced for the data injection test.

- BREAKING: Removes the ability to specify data injections into generic pods vs pod + container
- Adds a `compress` flag for data injections to enable gzip compression on transfer
- Adds file copy activity logging while running the data injection
- Adds new Helm Chart config option `NoWait` to allow helm to apply manifests and keep going (needed to address data injection race condition)
- Adds final data injection pod wait loop after data transfer
- Changes from `kubectl cp` exec to `kubectl exec`
- Adds a dynamic sync completion marker so that data injections leveraging an initContainer don't stall on their second run
- Removes the arbitrary timeout conditions breaking large data syncs on slow networks

## Related Issue
Issue #643


src/internal/packager/data.go
```diff
@@ -0,0 +1,99 @@
...
+
+			// Wait until the pod we are injecting data into becomes available
+			pods := k8s.WaitForPodsAndContainers(data.Target, true)
+			if len(pods) < 1 {
+				continue
+			}
+
+			// Inject into all the pods
+			for _, pod := range pods {
+				cpPodExec := fmt.Sprintf("tar cz -C %s . | kubectl exec -i -n %s %s -c %s -- tar xzvf - -C %s",
```
YrrepNoj, line 54: We talked about making the `z` in `tar xzvf` a flagged input in the `dataInjections:` section of a zarf.yaml right? Do you want me to do that if you're busy with other things?

jeff-mccoy: Added now, thx for noting that.

jeff-mccoy:
@YrrepNoj @Racer159 this should be good for review now. Would like to merge in to stop breaking other PRs with the data injection flake.

YrrepNoj (dismissed):
Manually tested with a private package that was injecting >1.5GB of data

src/types/component.go
```diff
@@ -106,16 +106,17 @@ type ZarfComponentScripts struct {
 
 // ZarfContainerTarget defines the destination info for a ZarfData target
 type ZarfContainerTarget struct {
-	Namespace string `yaml:"namespace"`
-	Selector  string `yaml:"selector"`
-	Container string `yaml:"container,omitempty"`
-	Path      string `yaml:"path"`
+	Namespace string `yaml:"namespace" jsonschema:"description=The namespace to targer for data injection"`
```
Racer159, line 109: targer -> target

Racer159 (changes requested):

src/internal/template/template.go
```diff
@@ -86,6 +86,11 @@ func (values Values) Apply(component types.ZarfComponent, path string) {
 		"GIT_AUTH_PULL":      values.secret.gitPull,
 	}
 
+	// Include the data injection marker template if the component has data injections
+	if len(component.DataInjections) > 0 {
```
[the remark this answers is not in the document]
jeff-mccoy: No real reason but consistency, there is a scenario where someone might try to apply a data injection to a different component's manifests, but it wouldn't be a good idea since we change the helm wait logic based on if the current component has a data injection.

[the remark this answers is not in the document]
Racer159: Ok yeah I am fine with it just wanted to make sure there wasn't something I was missing.  It would only affect users in niche ways anyway.

Racer159 (approved):

YrrepNoj (approved):
Because @jeff-mccoy doesn't believe that I approve of his changes 👀

PR #646, opened 2022-07-28
570 external git server

### LINKED ISSUE #570: Support for using external Gitlab instead of gitea on zarf init
Similar to #560, it may be desirable for some to use an external Gitlab rather than the internal gitea instance that is shipped with Zarf.

In my case specifically my platform has a pre-existing instance of Gitlab that I would like the zarf sync'd code to reside in.

Could configure Gitlab to mirror the gitea instance but I couldn't figure a way to do this in an automated fashion.  This work-around still doesn't prevent data duplication.

YrrepNoj:
Fixes #570

### DIFF
--- a/.github/workflows/test-k3d.yml
+++ b/.github/workflows/test-k3d.yml
@@ -49,7 +49,7 @@ jobs:
           docker buildx build --push --platform linux/amd64 --tag zarfdev/agent:$GITHUB_SHA .
 
       - name: "Make Packages"
-        run: make build-examples ARCH=amd64 AGENT_IMAGE=zarfdev/agent:$GITHUB_SHA
+        run: make init-package build-examples ARCH=amd64 AGENT_IMAGE=zarfdev/agent:$GITHUB_SHA
 
       - name: "Run Tests"
         # NOTE: This test run will create its own K3d cluster. A single cluster will be used throughout the test run.
--- a/.github/workflows/test-k3d.yml
+++ b/.github/workflows/test-k3d.yml
@@ -11,7 +11,7 @@ on:
       - "docs/**"
 
 # Abort prior jobs in the same workflow / PR
-concurrency: 
+concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
@@ -49,7 +49,7 @@ jobs:
           docker buildx build --push --platform linux/amd64 --tag zarfdev/agent:$GITHUB_SHA .
 
       - name: "Make Packages"
-        run: make init-package build-examples ARCH=amd64 AGENT_IMAGE=zarfdev/agent:$GITHUB_SHA
+        run: make build-examples ARCH=amd64 AGENT_IMAGE=zarfdev/agent:$GITHUB_SHA
 
       - name: "Run Tests"
         # NOTE: This test run will create its own K3d cluster. A single cluster will be used throughout the test run.
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,8 @@ ci-release: init-package ## Create the init package
 build-examples:
 	@test -s $(ZARF_BIN) || $(MAKE) build-cli
 
+	@test -s ./build/zarf-init-$(ARCH).tar.zst || $(ZARF_BIN) package create -o build -a $(ARCH) --set AGENT_IMAGE=$(AGENT_IMAGE) --confirm .
+
 	@test -s ./build/zarf-package-dos-games-$(ARCH).tar.zst || $(ZARF_BIN) package create examples/game -o build -a $(ARCH) --confirm
 
 	@test -s ./build/zarf-package-component-scripts-$(ARCH).tar.zst || $(ZARF_BIN) package create examples/component-scripts -o build -a $(ARCH) --confirm
--- a/Makefile
+++ b/Makefile
@@ -91,8 +91,6 @@ ci-release: init-package ## Create the init package
 build-examples:
 	@test -s $(ZARF_BIN) || $(MAKE) build-cli
 
-	@test -s ./build/zarf-init-$(ARCH).tar.zst || $(ZARF_BIN) package create -o build -a $(ARCH) --set AGENT_IMAGE=$(AGENT_IMAGE) --confirm .
-
 	@test -s ./build/zarf-package-dos-games-$(ARCH).tar.zst || $(ZARF_BIN) package create examples/game -o build -a $(ARCH) --confirm
 
 	@test -s ./build/zarf-package-component-scripts-$(ARCH).tar.zst || $(ZARF_BIN) package create examples/component-scripts -o build -a $(ARCH) --confirm
@@ -115,4 +113,6 @@ build-examples:
 ## Requires an existing cluster for the env var APPLIANCE_MODE=true
 .PHONY: test-e2e
 test-e2e: init-package build-examples
+	@test -s ./build/zarf-init-$(ARCH).tar.zst || $(ZARF_BIN) package create -o build -a $(ARCH) --set AGENT_IMAGE=$(AGENT_IMAGE) --confirm .
+
 	cd src/test/e2e && go test -failfast -v -timeout 30m
--- a/Makefile
+++ b/Makefile
@@ -79,13 +79,12 @@ dev-agent-image:
 	$(eval arch := $(shell uname -m))
 	CGO_ENABLED=0 GOOS=linux go build -o build/zarf-linux-$(arch) main.go
 	DOCKER_BUILDKIT=1 docker build --tag $(tag) --build-arg TARGETARCH=$(arch) . && \
-	kind load docker-image zarf-agent:$(tag) && \
+	kind load docker-image $(tag) && \
 	kubectl -n zarf set image deployment/agent-hook server=$(tag)
 
 init-package: ## Create the zarf init package, macos "brew install coreutils" first
 	@test -s $(ZARF_BIN) || $(MAKE) build-cli
-
-	@test -s ./build/zarf-init-$(ARCH).tar.zst || $(ZARF_BIN) package create -o build -a $(ARCH) --set AGENT_IMAGE=$(AGENT_IMAGE) --confirm .
+	$(ZARF_BIN) package create -o build -a $(ARCH) ---set AGENT_IMAGE=$(AGENT_IMAGE) -confirm .
 
 ci-release: init-package ## Create the init package
 
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ dev-agent-image:
 
 init-package: ## Create the zarf init package, macos "brew install coreutils" first
 	@test -s $(ZARF_BIN) || $(MAKE) build-cli
-	$(ZARF_BIN) package create -o build -a $(ARCH) ---set AGENT_IMAGE=$(AGENT_IMAGE) -confirm .
+	$(ZARF_BIN) package create -o build -a $(ARCH) --set AGENT_IMAGE=$(AGENT_IMAGE) --confirm .
 
 ci-release: init-package ## Create the init package
 
--- a/packages/zarf-agent/manifests/deployment.yaml
+++ b/packages/zarf-agent/manifests/deployment.yaml
@@ -41,7 +41,13 @@ spec:
             - name: tls-certs
               mountPath: /etc/certs
               readOnly: true
+            - name: zarf-state
+              mountPath: /etc/zarf-state
+              readOnly: true
       volumes:
         - name: tls-certs
           secret:
             secretName: agent-hook-tls
+        - name: zarf-state
+          secret:
+            secretName: zarf-state
--- a/src/cmd/initialize.go
+++ b/src/cmd/initialize.go
@@ -118,8 +118,8 @@ func init() {
 	initCmd.Flags().StringVar(&config.InitOptions.NodePort, "nodeport", "", "Nodeport to access the Zarf container registry. Between [30000-32767]")
 
 	// Flags for using an external Git server
-	initCmd.Flags().StringVar(&config.InitOptions.GitServerInfo.GitAddress, "git-url", "", "External git server url to use for this Zarf cluster")
-	initCmd.Flags().StringVar(&config.InitOptions.GitServerInfo.GitPushUsername, "git-user", "", "Username to connect to the external git server. User must be able to create repositories via 'git push'")
-	initCmd.Flags().StringVar(&config.InitOptions.GitServerInfo.GitPushPassword, "git-password", "", "Password to connect to the external git server")
-	initCmd.Flags().IntVar(&config.InitOptions.GitServerInfo.GitPort, "git-port", 0, "Port to connect to the external git server")
+	initCmd.Flags().StringVar(&config.InitOptions.GitServer.Address, "git-url", "", "External git server url to use for this Zarf cluster")
+	initCmd.Flags().StringVar(&config.InitOptions.GitServer.PushUsername, "git-user", "", "Username to connect to the external git server. User must be able to create repositories via 'git push'")
+	initCmd.Flags().StringVar(&config.InitOptions.GitServer.PushPassword, "git-password", "", "Password to connect to the external git server")
+	initCmd.Flags().IntVar(&config.InitOptions.GitServer.Port, "git-port", 0, "Port to connect to the external git server")
 }
--- a/src/cmd/initialize.go
+++ b/src/cmd/initialize.go
@@ -119,9 +119,7 @@ func init() {
 
 	// Flags for using an external Git server
 	initCmd.Flags().StringVar(&config.InitOptions.GitServerInfo.GitAddress, "git-url", "", "External git server url to use for this Zarf cluster")
-	initCmd.Flags().StringVar(&config.InitOptions.GitServerInfo.GitPushUsername, "git-user", "", "Username to connect to the external git server")
+	initCmd.Flags().StringVar(&config.InitOptions.GitServerInfo.GitPushUsername, "git-user", "", "Username to connect to the external git server. User must be able to create repositories via 'git push'")
 	initCmd.Flags().StringVar(&config.InitOptions.GitServerInfo.GitPushPassword, "git-password", "", "Password to connect to the external git server")
-	initCmd.Flags().StringVar(&config.InitOptions.GitServerInfo.GitReadUsername, "git-read-user", "", "Username for a read-only user to the external git server")
-	initCmd.Flags().StringVar(&config.InitOptions.GitServerInfo.GitReadPassword, "git-read-password", "", "Password for a read-only user to the external git server")
 	initCmd.Flags().IntVar(&config.InitOptions.GitServerInfo.GitPort, "git-port", 0, "Port to connect to the external git server")
 }
--- a/src/cmd/initialize.go
+++ b/src/cmd/initialize.go
@@ -113,7 +113,13 @@ func init() {
 	initCmd.Flags().BoolVar(&config.CommonOptions.Confirm, "confirm", false, "Confirm the install without prompting")
 	initCmd.Flags().StringVar(&config.CommonOptions.TempDirectory, "tmpdir", "", "Specify the temporary directory to use for intermediate files")
 	initCmd.Flags().StringVar(&config.DeployOptions.Components, "components", "", "Comma-separated list of components to install.")
-	initCmd.Flags().StringVar(&config.DeployOptions.StorageClass, "storage-class", "", "Describe the StorageClass to be used")
-	initCmd.Flags().StringVar(&config.DeployOptions.Secret, "secret", "", "Root secret value that is used to 'seed' other secrets")
-	initCmd.Flags().StringVar(&config.DeployOptions.NodePort, "nodeport", "", "Nodeport to access the Zarf container registry. Between [30000-32767]")
+	initCmd.Flags().StringVar(&config.InitOptions.StorageClass, "storage-class", "", "Describe the StorageClass to be used")
+	initCmd.Flags().StringVar(&config.InitOptions.Secret, "secret", "", "Root secret value that is used to 'seed' other secrets")
+	initCmd.Flags().StringVar(&config.InitOptions.NodePort, "nodeport", "", "Nodeport to access the Zarf container registry. Between [30000-32767]")
+
+	// Flags for using an external Git server
+	initCmd.Flags().StringVar(&config.InitOptions.GitServerInfo.GitAddress, "git-url", "", "External git server url to use for this Zarf cluster")
+	initCmd.Flags().StringVar(&config.InitOptions.GitServerInfo.GitUsername, "git-user", "", "Username to connect to the external git server")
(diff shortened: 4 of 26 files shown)


packages/zarf-agent/zarf.yaml
```diff
@@ -7,9 +7,9 @@ components:
   - name: zarf-agent
     required: true
     images:
-      - "defenseunicorns/zarf-agent:992efd"
+      - "defenseunicorns/zarf-agent:remote-git-wip-agent"
     variables:
-      tag: "992efd"
+      tag: "remote-git-wip-agent"    # TODO: Replace with a new real build
```
YrrepNoj, line 12: Don't forget to change this

Makefile
```diff
@@ -70,13 +70,13 @@ dev-agent-image:
...
 	CGO_ENABLED=0 GOOS=linux go build -o build/zarf-linux-$(arch) main.go
 	DOCKER_BUILDKIT=1 docker build --tag $(tag) --build-arg TARGETARCH=$(arch) . && \
-	kind load docker-image zarf-agent:$(tag) && \
+	kind load docker-image $(tag) && \
 	kubectl -n zarf set image deployment/agent-hook server=$(tag)
 
 init-package: ## Create the zarf init package, macos "brew install coreutils" first
 	@test -s $(ZARF_BIN) || $(MAKE) build-cli
 
-	@test -s ./build/zarf-init-$(ARCH).tar.zst || $(ZARF_BIN) package create -o build -a $(ARCH) --confirm .
```
jeff-mccoy, line 79: I dont hate this change, except this will force local tests to always rebuild the init package--which often isn't necessary during dev, is there a particular reason to keep the behavior consistent with the examples?

packages/zarf-agent/manifests/deployment.yaml
```diff
@@ -41,7 +41,13 @@ spec:
             - name: tls-certs
               mountPath: /etc/certs
               readOnly: true
+            - name: zarf-state
```
jeff-mccoy, line 44: This shouldn't have been as exciting as it was for me to see.....

src/internal/git/push.go
```diff
@@ -18,12 +18,38 @@ const offlineRemoteName = "offline-downstream"
...
+		gitServerURL = gitServerInfo.GitAddress
+
+		// Add a port to the URL if it was provided
+		if gitServerInfo.GitPort != 0 {
+			gitServerURL = fmt.Sprintf("%s:%d", gitServerInfo.GitAddress, gitServerInfo.GitPort)
+		}
 
-	tunnelUrl := fmt.Sprintf("http://%s", tunnel.Endpoint())
+		// If this URL points to a resource within the cluster, create a tunnel to it
+		if strings.Contains(gitServerURL, "svc.cluster.local") {
```
Racer159, line 42: This is probably fine since this is super edge-casey, but my brain immediately goes to "but what about https://svc.cluster.local.example.com!"

Racer159 (approved):

packages/zarf-agent/zarf.yaml
Racer159: Approved assuming this is changed

Makefile
YrrepNoj: `make e2e` builds the example packages as a side affect.
`make init-package` builds the init-package as its primary purpose. 

When I was testing changes with the zarf-agent it took me longer than I'm proud to admit to realize that I wasn't actually rebuilding the init-package 😛

src/internal/git/push.go
YrrepNoj: That's a good edge-case! I changed the way we check the URL string to avoid situations like this.

Racer159: :tada:

Racer159 (approved):

src/test/external-git/secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: private-git-server
+  namespace: git-server
+type: Opaque
+data:
+  username: Z2l0LXVzZXI=
+  password: c3VwZXJTZWN1cmVQYXNzd29yZA==
```
jeff-mccoy, line 9: Some scanners will 🤮 on this username/password.  Clearly this is just a test entry, but could we leave come comment stating that in case someone's scanner trips up on it?

YrrepNoj: good point, will do 👍

Makefile
```diff
@@ -79,13 +79,12 @@ dev-agent-image:
...
 	DOCKER_BUILDKIT=1 docker build --tag $(tag) --build-arg TARGETARCH=$(arch) . && \
-	kind load docker-image zarf-agent:$(tag) && \
+	kind load docker-image $(tag) && \
 	kubectl -n zarf set image deployment/agent-hook server=$(tag)
 
 init-package: ## Create the zarf init package, macos "brew install coreutils" first
 	@test -s $(ZARF_BIN) || $(MAKE) build-cli
-
-	@test -s ./build/zarf-init-$(ARCH).tar.zst || $(ZARF_BIN) package create -o build -a $(ARCH) --set AGENT_IMAGE=$(AGENT_IMAGE) --confirm .
+	$(ZARF_BIN) package create -o build -a $(ARCH) --set AGENT_IMAGE=$(AGENT_IMAGE) --confirm .
```
jeff-mccoy, line 87: This still kinda kills me since in dev we're now forcing an init package rebuilding on every test run--this will slow down my flow a good bit, can you talk about how you see this flow working if you run e2e tests locally often?

src/cmd/initialize.go
```diff
@@ -103,7 +103,13 @@ func init() {
...
 	initCmd.Flags().StringVar(&config.DeployOptions.Components, "components", "", "Comma-separated list of components to install.")
-	initCmd.Flags().StringVar(&config.DeployOptions.StorageClass, "storage-class", "", "Describe the StorageClass to be used")
-	initCmd.Flags().StringVar(&config.DeployOptions.Secret, "secret", "", "Root secret value that is used to 'seed' other secrets")
-	initCmd.Flags().StringVar(&config.DeployOptions.NodePort, "nodeport", "", "Nodeport to access the Zarf container registry. Between [30000-32767]")
+	initCmd.Flags().StringVar(&config.InitOptions.StorageClass, "storage-class", "", "Describe the StorageClass to be used")
+	initCmd.Flags().StringVar(&config.InitOptions.Secret, "secret", "", "Root secret value that is used to 'seed' other secrets")
+	initCmd.Flags().StringVar(&config.InitOptions.NodePort, "nodeport", "", "Nodeport to access the Zarf container registry. Between [30000-32767]")
+
+	// Flags for using an external Git server
+	initCmd.Flags().StringVar(&config.InitOptions.GitServerInfo.GitAddress, "git-url", "", "External git server url to use for this Zarf cluster")
```
jeff-mccoy, line 111: Not a huge deal, but `config.InitOptions.GitServerInfo.Gitaddress` feels fairly verbose to me.  `config.InitOptions.GitServer.Address` I think captures the same info.

src/cmd/initialize.go
```diff
@@ -103,7 +103,13 @@ func init() {
...
+	initCmd.Flags().StringVar(&config.InitOptions.GitServerInfo.GitPushUsername, "git-user", "", "Username to connect to the external git server. User must be able to create repositories via 'git push'")
```
jeff-mccoy, line 112: Is there a way in the flags to set the read-only username?  I imagine someone using an external registry would actually be more likely to also have different accounts or tokens to access rw vs r.

src/cmd/tools.go
```diff
@@ -73,7 +73,7 @@ var readCredsCmd = &cobra.Command{
 		// Continue loading state data if it is valid
 		config.InitState(state)
 
-		fmt.Println(config.GetSecret(config.StateGitPush))
+		fmt.Println(state.GitServerInfo.GitPushPassword)
```
jeff-mccoy, line 76: Doesnt have to be in this PR overall, but would be nice to state (maybe via`message` so it's on stderr) that this is only the Git Server Push Password, not every credential.

src/internal/agent/hooks/flux.go
```diff
@@ -34,15 +36,34 @@ func NewGitRepositoryMutationHook() operations.Hook {
 func mutateGitRepository(r *v1.AdmissionRequest) (*operations.Result, error) {
 	var patches []operations.PatchOperation
 
+	zarfState, err := getZarfStateFromFileWithinAgentPod(zarfStatePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load zarf state from file: %v", err)
+	}
+
+	// Default to the InCluster gitURL
+	gitServerURL := config.ZarfInClusterGitServiceURL
```
jeff-mccoy, line 45: This seems strange to me, why wouldn't we just parse the state for the URL like we do for external?  I'd personally rather keep it generic and since this is breaking change anyway, feels like the right time to move to the new pattern.  Is there any benefit to tracking in-cluster git server otherwise?

src/internal/git/push.go
```diff
@@ -18,12 +18,38 @@ const offlineRemoteName = "offline-downstream"
...
 
 func PushAllDirectories(localPath string) error {
-	// Establish a git tunnel to send the repos
-	tunnel := k8s.NewZarfTunnel()
-	tunnel.Connect(k8s.ZarfGit, false)
-	defer tunnel.Close()
+	gitServerInfo := config.GetGitServerInfo()
+	gitServerURL := ""
+
+	if gitServerInfo.InternalServer {
```
jeff-mccoy, line 24: It feels like it would be cleaner to just use the gitServerInfo data to build a tunnel if it's in cluster (by detecting if the url ends with `svc.cluster.local`).

src/internal/git/push.go
```diff
@@ -37,55 +63,68 @@ func PushAllDirectories(localPath string) error {
...
 
 		// Add the read-only user to this repo
-		repoPathSplit := strings.Split(path, "/")
-		repoNameWithGitTag := repoPathSplit[len(repoPathSplit)-1]
-		repoName := strings.Split(repoNameWithGitTag, ".git")[0]
-		err = addReadOnlyUserToRepo(tunnelUrl, repoName)
-		if err != nil {
-			message.Warnf("Unable to add the read-only user to the repo: %s\n", repoName)
-			return err
+		if gitServerInfo.InternalServer {
```
jeff-mccoy, line 79: We should probably think about this some more--we intentionally kept the read only user creation in the git server component so it wasn't tightly-coupled/hard-coded.  

It may require a different PR, but I think I've found a better way that exists now in Gitea to handle read only users:  you can now create repos per-org on push and then assign a user to the org once--or just set the rules to "must be logged in" to gitea and even that isn't required.    https://docs.gitea.io/en-us/config-cheat-sheet/#repository-repository, `ENABLE_PUSH_CREATE_ORG`. 


Screen Shot 2022-08-01 at 1 42 19 PM
Screen Shot 2022-08-01 at 1 42 11 PM
Screen Shot 2022-08-01 at 1 41 34 PM
Screen Shot 2022-08-01 at 1 40 53 PM
Screen Shot 2022-08-01 at 1 39 55 PM
Screen Shot 2022-08-01 at 1 38 55 PM

src/internal/k8s/tunnel.go
```diff
@@ -88,6 +88,33 @@ func PrintConnectTable() error {
 	return nil
 }
 
+// NewTunnelFromServiceURL takes a serviceURL and parses it to create a tunnel to the cluster. The string is expected to follow the following format:
```
jeff-mccoy, line 91: I love this feature

src/internal/k8s/tunnel.go
```diff
@@ -88,6 +88,33 @@ func PrintConnectTable() error {
...
+// Example serviceURL: http://{SERVICE_NAME}.{NAMESPACE}.svc.cluster.local:{PORT}
+func NewTunnelFromServiceURL(serviceURL string) (*Tunnel, error) {
+
+	// Remove the protocol from the URL
+	serviceURL = strings.TrimPrefix(serviceURL, "http://")
```
jeff-mccoy, line 96: Think we'd be better served using a URL parsing library than manually hacking this:  https://golangbyexample.com/parse-url-golang/

Makefile
jeff-mccoy: Yeah that's fair, then can we make a new `make init-package-if-not-exist` or something that e2e calls--and we could do the same for examples too I suppose, but that will make things noisy in the makefile.  Alternatively, we could just dump a warning in the makefile output, that the file already exists so reusing it, this is just for devs after all.  You can change text colors in make pretty easy:

```shell
GREEN  := $(shell tput -Txterm setaf 2)
YELLOW := $(shell tput -Txterm setaf 3)
WHITE  := $(shell tput -Txterm setaf 7)
CYAN   := $(shell tput -Txterm setaf 6)
RESET  := $(shell tput -Txterm sgr0)
```

```shell
 ${YELLOW}words in yellow${RESET} ${GREEN}words in green${RESET}
```

jeff-mccoy (commented):
Will continue PR review after we work through these comments--this is going to be a great option for larger deployments and lays the needed groundwork for external registry support 🥳

src/internal/agent/hooks/flux.go
Racer159: FWIW I had the same thought reviewing this earlier, I just didn't write it down.

src/cmd/initialize.go
YrrepNoj: I left that out on purpose for now. 

We use the read-only user/pass when we create the secrets within the `post-render.go` logic. If we add flags for read-only user account(s) those credentials will not work because we directly use the Gitea API and we won't be able to do that with external registries.

Makefile
YrrepNoj: How about a middle ground where we keep the `init-package` explicit and always creates the package but then I update the workflows.yaml and the `test-e2e` target to only make the init-package if it's not already there?

src/cmd/tools.go
YrrepNoj: In the same vein, changing the command name might be useful `get-admin-password` is pretty generic and unclear unless your read the description in a help menu.

src/internal/agent/hooks/flux.go
YrrepNoj: I'll make this part more generic.

The only other time where it makes sense to track in-cluster git server is when we are creating a read-only user. Since we use Gitea api we can't do that for an external server.

Makefile
jeff-mccoy: that seems reasonable to me

Makefile
```diff
@@ -79,19 +79,20 @@ dev-agent-image:
...
 
 ci-release: init-package ## Create the init package
 
 build-examples:
 	@test -s $(ZARF_BIN) || $(MAKE) build-cli
 
+	@test -s ./build/zarf-init-$(ARCH).tar.zst || $(ZARF_BIN) package create -o build -a $(ARCH) --set AGENT_IMAGE=$(AGENT_IMAGE) --confirm .
```
jeff-mccoy, line 94: I don't really see `build-examples` depending on `init-package`, think this test probably makes more sense under the e2e section.

.github/workflows/test-k3d.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           docker buildx build --push --platform linux/amd64 --tag zarfdev/agent:$GITHUB_SHA .
 
       - name: "Make Packages"
-        run: make init-package build-examples ARCH=amd64 AGENT_IMAGE=zarfdev/agent:$GITHUB_SHA
+        run: make build-examples ARCH=amd64 AGENT_IMAGE=zarfdev/agent:$GITHUB_SHA
```
jeff-mccoy, line 52: Probably cleaner to keep this explicit as it's going to be created anyway since it won't exist in CI

src/cmd/prepare.go
```diff
@@ -37,7 +37,7 @@ var prepareTransformGitLinks = &cobra.Command{
 
 		// Perform git url transformation via regex
 		text := string(content)
-		processedText := git.MutateGitUrlsInText(host, text)
+		processedText := git.MutateGitUrlsInText(host, text, config.InitOptions.GitServer.PushUsername)
```
jeff-mccoy, line 40: This one is a little funny to me as the username is now dynamic and a `package deploy` configuration, not `package create` configuration.  Should this default to the internal user, but have an option to specify a custom one?  

Also curious how this works with external git servers--we can predict the path based off the username, is that always true in other git servers or is that something we need to talk about too?

jeff-mccoy: ... and scrolling down I see the flag for the pus user option, so disregard the first point.

src/internal/packager/seed.go
```diff
@@ -103,14 +103,32 @@ func preSeedRegistry(tempPath tempPaths) {
 	}
 
 	// CLI provided overrides that haven't been processed already
-	if config.DeployOptions.NodePort != "" {
-		state.NodePort = config.DeployOptions.NodePort
+	if config.InitOptions.NodePort != "" {
```
jeff-mccoy, line 106: Moving to InitOptions now made a lot of sense and will keep things saner. ✅

src/internal/packager/seed.go
```diff
@@ -103,14 +103,32 @@ func preSeedRegistry(tempPath tempPaths) {
...
+	if config.InitOptions.StorageClass != "" {
+		state.StorageClass = config.InitOptions.StorageClass
+	}
+
+	if config.InitOptions.GitServer.Address != "" {
+		// The user has provided external git server info, just use what they gave!
+		state.GitServer = config.InitOptions.GitServer
+
+		// For external clusters, the read-user is the same as the push-user
+		state.GitServer.ReadUsername = state.GitServer.PushUsername
```
jeff-mccoy, line 121: I still think we need to look at this--I don't see why wouldn't allow the user to specify a read-only user as well--it would likely be a robot account, PAT or similar to grant less-privilege and I think it would be a mistake to not allow them to do that.  Think this is related to assuming we can push to a different git server just like we do for gitea.

src/internal/packager/seed.go
```diff
@@ -103,14 +103,32 @@ func preSeedRegistry(tempPath tempPaths) {
...
+		state.GitServer.ReadPassword = state.GitServer.PushPassword
+	} else {
+		// Set the GitServerInfo for the internal Gitea server
+		state.GitServer.Address = config.ZarfInClusterGitServiceURL
+		state.GitServer.Port = config.ZarfInClusterGitServicePort
+		state.GitServer.PushUsername = config.ZarfGitPushUser
+		state.GitServer.PushPassword = utils.RandomString(24)
```
jeff-mccoy, line 128: this is a welcome change....

src/cmd/initialize.go
jeff-mccoy: I still don't think that's really an issue, we just have to communicate "if you specify a read-only user, you must ensure that repos zarf pushes can be accessed by that user"

jeff-mccoy (commented):
Left more comments, happy to discuss further--think it's getting pretty close and is a major enhancement to zarf's deployment flexibility, thanks!

src/internal/packager/seed.go
YrrepNoj: Let me know when you have a bit of time to chat about this. Just want to make sure we're both on the same page :)

src/cmd/prepare.go
YrrepNoj: It seems to be fairly consistent with all of the git servers I have used before. The path is the domain followed by either the username or org name. 

I reworded the flag message here to make it a little more obvious what to do if the repo is originally pushed under an org.

Racer159: That is how usernames work in GitLab, GitHub, BitBucket, and GitTea (and any other host based on those like CodeBerg) so it is likely good enough (and even if they do something like `https://example.com/scm/username/project.git` instead of a subdomain you should be able to use the full `https://example.com/scm/` as the `host` variable).

src/internal/agent/hooks/flux.go
```diff
@@ -34,15 +36,28 @@ func NewGitRepositoryMutationHook() operations.Hook {
 func mutateGitRepository(r *v1.AdmissionRequest) (*operations.Result, error) {
 	var patches []operations.PatchOperation
 
+	zarfState, err := getZarfStateFromFileWithinAgentPod(zarfStatePath)
```
Racer159, line 39: Could this be shortened to just `getZarfStateFromAgent` (or similar)? `getZarfStateFromFileWithinAgentPod` is a little long

src/internal/agent/hooks/flux.go
```diff
@@ -34,15 +36,28 @@ func NewGitRepositoryMutationHook() operations.Hook {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to load zarf state from file: %v", err)
+	}
+
+	// Form the gitServerURL from the state
+	gitServerURL := zarfState.GitServer.Address
+	if zarfState.GitServer.Port != 0 {
+		gitServerURL += fmt.Sprintf(":%d", zarfState.GitServer.Port)
```
Racer159, line 47: What I said above is not true though when using a non-default port.  If I hosted gittea behind an nginx server and rewrote a subpath to point to my git server then this would result in something like: https://example.com/my-scm:5000...

src/internal/git/push.go
```diff
@@ -18,12 +18,25 @@ const offlineRemoteName = "offline-downstream"
...
-	tunnel := k8s.NewZarfTunnel()
-	tunnel.Connect(k8s.ZarfGit, false)
-	defer tunnel.Close()
+	gitServerInfo := config.GetGitServerInfo()
+	gitServerURL := gitServerInfo.Address
 
-	tunnelUrl := fmt.Sprintf("http://%s", tunnel.Endpoint())
+	// Add a port to the URL if it was provided
+	if gitServerInfo.Port != 0 {
+		gitServerURL = fmt.Sprintf("%s:%d", gitServerInfo.Address, gitServerInfo.Port)
```
Racer159, line 26: Same thing here if address is not a root path

src/test/external-git/README.md
```diff
@@ -0,0 +1,33 @@
...
+> Note: In this case, the 'external' Git server is a server that already existed inside the k8s cluster
+
+This directory holds the tests that verify Zarf can initialize a cluster to use an already existing Git server that is external to the resources Zarf manages. The tests in this directory are currently only run when manually executed.
+
+
+## Running Tests Locally
+
+### Dependencies
+Running the tests locally have the same prerequisites as running and building Zarf:
+ 1. GoLang >= `1.18.x`
```
Racer159, line 11: 1.19.x

Racer159 (changes requested):

src/internal/agent/hooks/flux.go
YrrepNoj: In that case you could add the port as a string within the server URL and just not populate the port flag. If we ever actually need the port we could try to parse the port using `net/url`.

@Racer159 Do you think this works or do you think we need to come up with a more robust solution?

Racer159: At that point why specify the port as a separate thing at all?

YrrepNoj: That's a good point, I thought it would have been needed earlier but I'll just give the url

YrrepNoj:
Closing this as it's being duplicated by #666 along with other features (handling an external container registry)

PR #651, opened 2022-07-30
Always build & test Zarf Agent during pull requests; publish latest Zarf Agent on release

### LINKED ISSUE #386: Complete ARM Support

jeff-mccoy:
- Adds the Zarf Agent (mutating webook) automatic build in test with the latest code
- Fix issue with out of sync Zarf Agent image tag, always cut a new agent image before a release
- Test the release once more before publishing
- Full support for Linux ARM, closes #386


YrrepNoj (approved):

jeff-mccoy:
This is the related multi-arch Stage II Injector build: https://github.com/defenseunicorns/zarf/actions/runs/2771188396

Screen Shot 2022-08-01 at 11 57 48 AM

jeff-mccoy:
Every PR now will rebuild/publish a dev zarf agent image (using a different dockerhub account) so that every PR tests against the latest cut of the go codebase.

Screen Shot 2022-08-01 at 12 03 22 PM

YrrepNoj (approved):

PR #660, opened 2022-08-05
Add new "prepare" stage for components scripts to perform actions on "package create"

jeff-mccoy:
## Type of change

## Checklist before merging

```yaml
components:
  - name: prepare
    required: true
    scripts:
      # Note this file will be created in this directory, regardless of where the package create command is called from
      prepare:
        - "touch test-prepare.txt"
  # This script will pass after 1 second
  - name: deploy
    required: true
    scripts:
      before:
        - "touch test-deploy-before.txt"
      after:
        - "touch test-deploy-after.txt"

  # This script will fail after 1 seconds
  - name: timeout
    scripts:
      timeoutSeconds: 1
      before:
        - "sleep 30"
```


jeff-mccoy:
This one's for you @Racer159 😆 
Screen Shot 2022-08-05 at 2 46 51 AM

examples/component-scripts/zarf.yaml
```diff
@@ -4,14 +4,26 @@ metadata:
   description: "Test component to demonstrate script timeout feature"
 
 components:
+  - name: create
+    required: true
+    scripts:
+      # Note these files will be created in this directory, regardless of where the package create command is called from
+      create:
```
Racer159, line 11: do we need both before and after create functions? (might be weird to take actions after the .zst file is built?)

Racer159:
After discussion going toward:

```
  - name: nextcloud-apps
    scripts:
      before:
        - pwd
        - echo preparing inject
      after:
        - pwd
        - echo successfully injected
      prepare:
        - pwd
        - tar -xzf data/apps/*
      showOutput: true
    dataInjections:
      - source: apps
        target:
          namespace: nextcloud
          selector: app.kubernetes.io/name=nextcloud
          container: nextcloud
          path: /var/www/html/apps
        compress: true
```

Racer159 (approved):
lgtm!

examples/component-scripts/zarf.yaml
```diff
@@ -4,14 +4,23 @@ metadata:
   description: "Test component to demonstrate script timeout feature"
 
 components:
+  - name: prepare
+    required: true
+    scripts:
+      # Note these files will be created in this directory, regardless of where the package create command is called from
```
Racer159, line 10: Adding this as a comment since it is so minor, but there is only one file here. (this file)

Racer159 (approved):
woot!

PR #663, opened 2022-08-08
Handle release versions on init in a backwards compatible way

### LINKED ISSUE #529: `Zarf Init` should warn when the binary is out of sync with a release
If I `brew upgrade zarf`, I may forget to download the release associated with the updated binary.

Warn / remind users that the binary is out of sync with the release.

Remember to download the associated release when I update the binary

**Additional context**
Blame @jeff-mccoy

Racer159:
The leading v in the Zarf version was silently erroring the backwards compatibility for Zarf init package pulling.

## Related Issue

Fixes #529 

## Type of change

## Checklist before merging


Racer159:
@megamind came to this after some discussions with @YrrepNoj as it will pull the correct init release if we ever needed to go backward on a branch and build something (or if a breaking release was just pushed and someone was working off of an older version of the code).  We cannot completely get rid of the semver package as it is an indirect dependency of another library.

Racer159:
I also removed the spaces so that the Make versions are correct and do not contain a leading space.

jeff-mccoy:
> I also removed the spaces so that the Make versions are correct and do not contain a leading space.

okay now you're just showing off 😆.

src/cmd/initialize.go
```diff
@@ -58,16 +59,23 @@ var initCmd = &cobra.Command{
 					message.Fatalf(nil, "This command requires a zarf-init package, but one was not found on the local system.")
 				}
 
-				// If no CLI version exists (should only occur in dev or CI), try to get the latest release tag from Githhub
-				if _, err := semver.StrictNewVersion(config.CLIVersion); err != nil {
-					config.CLIVersion, err = utils.GetLatestReleaseTag(config.GithubProject)
+				// Parse the CLI version and extract its parts
+				initPackageVersion := strings.TrimLeft(config.CLIVersion, "v")
```
jeff-mccoy, line 63: Perhaps we should just do `strings.Trim()` for consistency instead here since the source left whitespace issue is resolved?

jeff-mccoy (approved):
Leave the strim thing up to you, otherwise lgtm

src/cmd/initialize.go
Racer159: Here we are trimming the `v` out and only expect it to be at the front of the string.  We could do a more precise version of this with something like:

```
if len(config.CLIVersion > 0) {
  initPackageVersion := config.CLIVersion[1 : len(config.CLIVersion)-1]
}
```

It felt simpler to use `strings.TrimLeft` even though that allows `vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv0.10.0`

jeff-mccoy: fair enough.  Since we don't have a test for this unique usecase (and it would be hard to test I think), can you drop a screenshot of it working as expected now?

Racer159: image

Here are most cases.  Top to bottom:

That branch as it is
A valid main release
An invalid semantic version
A valid older git commit

Issue #560, opened 2022-06-24
Support external registry on zarf init


jeff-mccoy:
In support of #376, Zarf should be able to use an external registry as an alternative to the injector, seed registry and registry components in the init package currently.  When an external registry is available, the connection info should be placed in the `zarf-state` secret and used everywhere appropriately.  

- [ ] Zarf init accepts a `host`, `user`, `passord` flags
- [ ] Update ZarfState to accept an override for the password in lieu of the calculated password 
- [ ] All calls, including the Zarf Agent, should use these variables vs hardcoding `127.0.0.1:31999`
- [ ] Docs should be updated to explain when this is useful such as when used with another Zarf Cluster acting as a #186 or in egress-restricted cloud environments

PR #666, opened 2022-08-09
Handle External Git Repository and Container Registry on `zarf init`

### LINKED ISSUE #561: Can't deploy Zarf package built with images from insecure repo
Environment
Device and OS: 2019 MacBook Pro, i9 16gb RAM, Big Sur 11.6
App version: 0.19.5
Kubernetes distro being used: minikube
Other:

### Steps to reproduce
1. `zarf package create --insecure`
2. `zarf package deploy --insecure`

### Expected result
Package deployment

### Actual Result
```
  ✔  Creating port forwarding tunnel available at http://127.0.0.1:45001                                                                                                                        
DEBUG   config.GetSecret(registry-push)                                                                                                                                                       
DEBUG   0x2c36f40                                                                                                                                                                             
DEBUG   parsing reference "127.0.0.1:45001/zarfregistrytest:5000/gateway:latest": could not parse reference: 127.0.0.1:45001/zarfregistrytest:5000/gateway:latest                             
ERROR:  Unable to push the image to the registry
```

Screen Shot 2022-06-24 at 10 48 24 AM

### Additional Context
Zarf successfully packages the images (or so it appears) with `zarf package create`. Unfortunately I cannot deploy the package on the same machine the package is created on. I haven't tried a separate machine but I would expect the same result. 

If the Zarf package contains images that were not pulled from the local reg (i.e. docker hub) it copies those fine; it only fails when it tries to copy the local images. It doesn't matter if the images were built locally, tagged, and pushed to the registry or pulled from outside, tagged, and pushed to local registry (tested with busybox).

YrrepNoj:
Fixes #561 #570 #560

Introduce new flags during `zarf init` to allow for having zarf push and pull from an external git repository and/or an external container registry. This registry can exist inside the cluster and have a url provided as a service url or can exist outside of the cluster assuming networking is configured to talk to the outside internet.

With this change, we are also now storing values that were derived from the `ZarfState.Secret` directly in the `ZarfState`.


src/types/k8s.go
```diff
@@ -6,9 +6,35 @@ type ZarfState struct {
...
+	ReadUsername string `json:"readUsername"`
+	ReadPassword string `json:"readPassword"`
+
+	Address string `json:"address"`
+	Port    int    `json:"port"`
+
+	InternalServer bool `json:"internalServer"`
+}
+
+type ContainerRegistryInfo struct {
```
jeff-mccoy, line 28: Nit, but we don't have any other concept of different registries types in zarf that I now of, would it make sense to use `RegistryInfo` so it's consistent with how we label registry everywhere else?  A library we use calls its `go-containerregistry`, but we just say `registry` for the most part.

src/types/k8s.go
```diff
@@ -6,9 +6,35 @@ type ZarfState struct {
...
-	NodePort      string       `json:"nodePort"`
 	AgentTLS      GeneratedPKI `json:"agentTLS"`
+
+	GitServer             GitServerInfo         `json:"gitServer"`
+	ContainerRegistryInfo ContainerRegistryInfo `json:"containerRegistryInfo"`
+	LoggingSecret         string                `json:"loggingSecret"`
+}
+
+type GitServerInfo struct {
+	PushUsername string `json:"pushUsername"`
```
jeff-mccoy, line 17: It probably doesn't matter as much since this isn't exposed in the `zarf.yaml`, but I've been trying to fill out at least the jsonschema description fields in the types pkg.

src/test/external-git/gitea-values.yaml
```diff
@@ -0,0 +1,34 @@
+persistence:
+  storageClass: "standard" # 'standard' is for kind
```
jeff-mccoy, line 2: This might cause us problems right?  Because k3s/k3d don't have that storageclass ootb, which is why we have the storage class template variable.

src/test/external-git/gitea-values.yaml
```diff
@@ -0,0 +1,34 @@
...
+gitea:
+  admin:
+    username: "git-user"
+    password: "superSecurePassword"
```
jeff-mccoy, line 6: should add the same password disclaimer here

src/test/external-git/external_git_server_test.go
```diff
@@ -0,0 +1,68 @@
...
+	initArgs := []string{"init", "--git-user", "git-user", "--git-password", "superSecurePassword", "--git-url", "http://gitea-http.git-server.svc.cluster.local:3000", "--confirm"}
+	_, _, err = utils.ExecCommandWithContext(context.TODO(), true, "../../../build/zarf-mac-intel", initArgs...)
+	require.NoError(t, err, "unable to initialize the k8s server with zarf")
+
+	// Deploy the flux example package
+	deployArgs := []string{"package", "deploy", "../../../build/zarf-package-flux-test-amd64.tar.zst", "--confirm"}
+	_, _, err = utils.ExecCommandWithContext(context.TODO(), true, "../../../build/zarf-mac-intel", deployArgs...)
+	require.NoError(t, err, "unable to deploy flux example package")
+
+	// Verify flux was able to pulll from the 'external' registry
```
jeff-mccoy, line 40: pulll

src/internal/template/template.go
```diff
@@ -45,22 +45,22 @@ func Generate() Values {
 	generated.seedRegistry = config.GetSeedRegistry()
 	generated.registry = config.GetRegistry()
 
-	generated.secret.registryPush = config.GetSecret(config.StateRegistryPush)
-	generated.secret.registryPull = config.GetSecret(config.StateRegistryPull)
-	generated.secret.registrySecret = config.GetSecret(config.StateRegistrySecret)
+	generated.secret.registryPush = config.GetContainerRegistryInfo().PushPassword
+	generated.secret.registryPull = config.GetContainerRegistryInfo().PullPassword
+	generated.secret.registrySecret = utils.RandomString(48)
```
jeff-mccoy, line 50: I do think if we are exposing this secret with templating we need to preserve it in the state too.  Otherwise things might get weird on a registry upgrade + template as you'd be introducing a different shared secret with the update.

jeff-mccoy (commented):
reviewed 17 of 34, will look at more tomorrow

src/test/external-git/gitea-values.yaml
YrrepNoj: Good catch, I was defiantly only using kind when I was testing since that is what we have the simple `make dev-agent-image` target to replace the container image for the  Zarf agent.

I originally intended for this test to be manually run for now since there wasn't a clear place for it to exist within the the e2e test for every PR (I didn't want to have to slow down the feedback loop by doing a `zarf destroy` and re-initializing another cluster.

Are you okay with keeping it 'dumb' and having assumptions over the distro? Or should I add it to the e2e maybe as its own standalone workflow?

jeff-mccoy:
Could you verify the contents of the zarfstate secret?

tworcester:
```
...
"registryInfo":{"pushUsername":"<>","pushPassword":"<>","pullUsername":"<>","pullPassword":"<>","address":"<>","nodePort":0,"internalRegistry":false,"secret":"<>"}
...
```

Values all look good there.

jeff-mccoy:
@YrrepNoj why is there still a secret value in the state? I thought we had removed the generic secret completely.

tworcester:
Interesting, seems to be something potentially hardcoded with the gitea?  Because the zarf-agent works fine:

```
$ k -n zarf get pods -oyaml | grep image:
    - image: <registry>/defenseunicorns/zarf-agent:992efd
      image: <registry>/defenseunicorns/zarf-agent:992efd
    - image: <registry>/defenseunicorns/zarf-agent:992efd
      image: <registry>/defenseunicorns/zarf-agent:992efd
      image: 127.0.0.1:31999/gitea/gitea:1.16.8
      image: 127.0.0.1:31999/gitea/gitea:1.16.8
      image: 127.0.0.1:31999/gitea/gitea:1.16.8
      image: 127.0.0.1:31999/gitea/gitea:1.16.8
    - image: 127.0.0.1:31999/gitea/gitea:1.16.8
    - image: 127.0.0.1:31999/gitea/gitea:1.16.8
    - image: 127.0.0.1:31999/gitea/gitea:1.16.8
    - image: 127.0.0.1:31999/gitea/gitea:1.16.8
```
```
$ k -n zarf get pods
NAME                          READY   STATUS                  RESTARTS   AGE
agent-hook-55bc978bdb-kdkrq   1/1     Running                 0          47s
agent-hook-55bc978bdb-zhn26   1/1     Running                 0          47s
zarf-gitea-0                  0/1     Init:ImagePullBackOff   0          43s
```

jeff-mccoy:
The zarf agent image is not modified by itself due to race conditions, it's managed on package deploy instead.

YrrepNoj:
> @YrrepNoj why is there still a secret value in the state? I thought we had removed the generic secret completely.

That is not the generic secret but is the registry secret that gets templated into the `registry-values.yaml` when we stand up our docker-registry.

YrrepNoj:
> ```
> ...
> "registryInfo":{"pushUsername":"<>","pushPassword":"<>","pullUsername":"<>","pullPassword":"<>","address":"<>","nodePort":0,"internalRegistry":false,"secret":"<>"}
> ...
> ```
> 
> Values all look good there.

Was the `--registry-url` an internal cluster route (i.e. `http://{SERVICE_NAME}.{NAMESPACE}.svc.cluster.local`? Your output of `$ k -n zarf get pods -oyaml | grep image:` is what gives me that impression. 

If that's the case you might need to specify a different nodeport. We use `31999` as the default for the registry we would stand up but depending on how you're standing up your own registry, that might be different.

tworcester:
Nope, my use case is two different kubernetes clusters managed by Rancher.  Registry is in Cluster A and Cluster B is where zarf is being initialized.

To use this registry I ran: `docker login docker-registry:443` and could use docker pull/push.
Configured zarf init like this (with credentials): `./zarf init --storage-class longhorn --registry-pull-password <> --registry-pull-username <> --registry-push-password <> --registry-push-username <> --registry-url docker-registry:443`

docker-registry is an DNS entry pointing to a nodeport svc in Cluster A.

jeff-mccoy:
Might be good to dump the zarf-agent logs too

tworcester:
Here ya go!
[zarf.log](https://github.com/defenseunicorns/zarf/files/9416139/zarf.log)

src/cmd/initialize.go
```diff
@@ -108,12 +113,39 @@ var initCmd = &cobra.Command{
...
-	initCmd.Flags().StringVar(&config.DeployOptions.NodePort, "nodeport", "", "Nodeport to access the Zarf container registry. Between [30000-32767]")
+	initCmd.Flags().StringVar(&config.InitOptions.StorageClass, "storage-class", "", "Describe the StorageClass to be used")
+
+	// Flags for using an external Git server
+	initCmd.Flags().StringVar(&config.InitOptions.GitServer.Address, "git-url", "", "External git server url to use for this Zarf cluster")
+	initCmd.Flags().StringVar(&config.InitOptions.GitServer.PushUsername, "git-push-username", config.ZarfGitPushUser, "Username to access to the git server Zarf is configured to use. User must be able to create repositories via 'git push'")
+	initCmd.Flags().StringVar(&config.InitOptions.GitServer.PushPassword, "git-push-password", "", "Password for the push-user to access the git server")
+	initCmd.Flags().StringVar(&config.InitOptions.GitServer.ReadUsername, "git-read-username", "", "Username for read-only access to the git server")
+	initCmd.Flags().StringVar(&config.InitOptions.GitServer.ReadPassword, "git-read-password", "", "Password for the read-only user to access the git server")
```
Racer159, line 139: I think this should be standardized to `pull` instead of `read` (or rename `push` to `write`) to maintain consistency.

src/internal/git/push.go
```diff
@@ -37,55 +45,68 @@ func PushAllDirectories(localPath string) error {
...
+func push(repo *git.Repository, localPath string, spinner *message.Spinner) error {
+
 	gitCred := http.BasicAuth{
-		Username: config.ZarfGitPushUser,
-		Password: config.GetSecret(config.StateGitPush),
+		Username: config.GetState().GitServer.PushUsername,
+		Password: config.GetState().GitServer.PushPassword,
 	}
 
 	// Since we are pushing HEAD:refs/heads/master on deployment, leaving
```
Racer159, line 112: Does line 120 just work in Gitea land? (pushing to a repo that doesn't exist?)  I tested this with GitLab and it returned a `422` (full log below):

```
  ✔  Storing images in the zarf registry                                                                                       
  DEBUG   tunnel.Close()
└ (/home/wayne/workspace/zarf/src/internal/k8s/tunnel.go:254)
  DEBUG   Rewrite git URL: https://github.com/defenseunicorns/zarf.git -> https://gitlab.com/Racer159/mirror__github.com__defenseunicorns__zarf.git
└ (/home/wayne/workspace/zarf/src/internal/git/utils.go:49)

  DEBUG   Remove head copies for /tmp/zarf-3968008433/components/baseline/repos/mirror__github.com__defenseunicorns__zarf.git__v0.15.0
└ (/home/wayne/workspace/zarf/src/internal/git/utils.go:134)

  DEBUG   Remove git references /tmp/zarf-3968008433/components/baseline/repos/mirror__github.com__defenseunicorns__zarf.git__v0.15.0
└ (/home/wayne/workspace/zarf/src/internal/git/utils.go:162)

 WARNING  Unable to push the git repo mirror__github.com__defenseunicorns__zarf.git__v0.15.0                                   
  DEBUG   unable to push repo to the gitops service: unexpected client error: unexpected requesting "https://gitlab.com/Racer159/mirror__github.com__defenseunicorns__zarf.git/info/refs?service=git-receive-pack" status code: 422
└ (/home/wayne/workspace/zarf/src/internal/packager/deploy.go:280)
 WARNING  Unable to push repos to the Zarf Registry, retrying in 5 seconds...
  DEBUG   Rewrite git URL: https://github.com/defenseunicorns/zarf.git -> https://gitlab.com/Racer159/mirror__github.com__defenseunicorns__zarf.git
└ (/home/wayne/workspace/zarf/src/internal/git/utils.go:49)

 WARNING  error when preping the repo for push.. failed to create offline remote: remote already exists                        

  DEBUG   failed to create offline remote: remote already exists
└ (/home/wayne/workspace/zarf/src/internal/packager/deploy.go:280)
 WARNING  Unable to push repos to the Zarf Registry, retrying in 5 seconds...
  DEBUG   Rewrite git URL: https://github.com/defenseunicorns/zarf.git -> https://gitlab.com/Racer159/mirror__github.com__defenseunicorns__zarf.git
└ (/home/wayne/workspace/zarf/src/internal/git/utils.go:49)

 WARNING  error when preping the repo for push.. failed to create offline remote: remote already exists                        

  DEBUG   failed to create offline remote: remote already exists
└ (/home/wayne/workspace/zarf/src/internal/packager/deploy.go:280)
 WARNING  Unable to push repos to the Zarf Registry, retrying in 5 seconds...
  ✔  Zarf deployment complete
```

(Also these errors should likely be more than just warnings since the deployment was not actually complete)

Racer159 (changes requested):
Tried testing the git repo stuff on Gitlab, but ran into some issues.

src/cmd/initialize.go
tworcester: I don't think the current changes will work for gitlab since the parameters in the snippet above say: `User must be able to create repositories via 'git push'`  Which is not a supported feature in Gitlab.

jeff-mccoy: That's an interesting problem we should address, would the user be expected to manually create those first, would we be on the hook to build api calls for various git servers?

YrrepNoj: I'll test it more later but GitLab docs seem to say it's possible if your user has the right permissions. https://docs.gitlab.com/ee/user/project/working_with_projects.html#create-a-new-project-with-git-push

src/internal/git/push.go
Racer159: This seems to be the reason it failed:

`Could not create project: Path must not start or end with a special character and must not contain consecutive special characters.`

There are two underscores in a row in the project name.

Racer159: Once we resolve that we should have a note in the docs that this will only work with Gitea and GitLab (git push to non-existent repos is not a thing on Bitbucket or Azure DevOps Services (the other two big `git` hosting players).  Source Hut is up and coming and also does not support this.

src/cmd/initialize.go
tworcester: Okay I see what happened in my case.  I used my personal account an generated a API token to pass to zarf-agent to do the git-operations.  It allowed the git push but it put it in my personal projects rather than the group structure that it was cloned with.

```
  DEBUG   Rewrite git URL: https://gitlab01.domain.com/devgroup/pipeline-templates.git -> https://paas-gitlab01/tworcester/mirror__paas-gitlab01.lakitu.com__devgroup__pipeline-templates.git
```

tworcester:
This one is likely user error, I had a component with an image tag like this:
```yaml
components:
  - name: baseline
    required: true
    images:
     # Harbor images
     - paas-docker-registry:443/ironbank/opensource/goharbor/chartmuseum:v2.5.1
```

And when I do my `zarf deploy` I see this in the logs:
```
  DEBUG   parsing reference "paas-docker-registry:443/paas-docker-registry:443/ironbank/opensource/goharbor/chartmuseum:v2.5.1": could not parse reference: paas-docker-registry:443/paas-docker-registry:443/ironbank/opensource/goharbor/chartmuseum:v2.5.1
```
Is it not supposed to strip out the old registry reference and retag it to the new one?

YrrepNoj:
> Is it not supposed to strip out the old registry reference and retag it to the new one?

We do our best to strip the old registry based on [this](https://regex101.com/library/PiL191) regex. Here is a link to the playground we have to see if a registry host is detected (and if Zarf will be able to successfully swap it). https://regex101.com/r/PiL191/1

Since your registry host didn't get swapped I'm guessing whatever host you're using is not getting matched by that regex. Could you give a more 'exact' example of what your host is so we can see if we can improve the regex to potentially identify it?

YrrepNoj:
> > @tworcester Was the `--registry-url` an internal cluster route (i.e. `http://{SERVICE_NAME}.{NAMESPACE}.svc.cluster.local`? Your output of `$ k -n zarf get pods -oyaml | grep image:` is what gives me that impression.
> > If that's the case you might need to specify a different nodeport. We use `31999` as the default for the registry we would stand up but depending on how you're standing up your own registry, that might be different.
> 
> @YrrepNoj Nope, my use case is two different kubernetes clusters managed by Rancher. Registry is in Cluster A and Cluster B is where zarf is being initialized.
> 
> To use this registry I ran: `docker login docker-registry:443` and could use docker pull/push. Configured zarf init like this (with credentials): `./zarf init --storage-class longhorn --registry-pull-password <> --registry-pull-username <> --registry-push-password <> --registry-push-username <> --registry-url docker-registry:443`
> 
> docker-registry is an DNS entry pointing to a nodeport svc in Cluster A.

@tworcester I feel silly for not realizing this earlier. The reason why you were seeing issues is because the `agent-image` (the mutating webhook that changes host of images the pods uses) version is hardcoded in the Makefile. This means you were likely using an old webhook that wasn't smart enough to look for the new `external-url` string when mutating the pods. I was caught up in my dev environment and forgot I changed that 🤦 

If you pull down the latest changes and rebuild I believe your errors with gitea should be resolved.

tworcester:
> > Is it not supposed to strip out the old registry reference and retag it to the new one?
> 
> We do our best to strip the old registry based on [this](https://regex101.com/library/PiL191) regex. Here is a link to the playground we have to see if a registry host is detected (and if Zarf will be able to successfully swap it). https://regex101.com/r/PiL191/1
> 
> Since your registry host didn't get swapped I'm guessing whatever host you're using is not getting matched by that regex. Could you give a more 'exact' example of what your host is so we can see if we can improve the regex to potentially identify it?

The values I posted were the exact values I was using.  Looks like having two dashes in the registry name breaks the regex.
image

tworcester:
Is there a reason you need to match on the registry itself?  Using a regex like `/(.*?)\/?(.+)$` to match just the trailing section should suffice.

The only downside that I see to this is if you have a registry in say, Nexus.  Where you may have a registry url like `https://awesome.nexus.com/repository/docker/<image path>:<image tag>`.  With the above regex you would get the /repository/docker as part of the match...which is probably harmless as it looks like your current regex achieves the same.

Another shortcoming of the regex I supplied above is that `https://` being at the front of an entry breaks the regex as well.

image

tworcester:
`(?im)^([a-z0-9\-.]+\.[a-z0-9\-]+:?[0-9]*|[a-z0-9\-.]+\-[a-z0-9\-]+:?[0-9]*)?/?(.+)$`

@YrrepNoj  Feels like a hack...but it was the only one I could get to work for all cases.
https://regex101.com/r/F6hx4h/1

YrrepNoj:
> `(?im)^([a-z0-9\-.]+\.[a-z0-9\-]+:?[0-9]*|[a-z0-9\-.]+\-[a-z0-9\-]+:?[0-9]*)?/?(.+)$`
> 
> @YrrepNoj Feels like a hack...but it was the only one I could get to work for all cases. https://regex101.com/r/F6hx4h/1

Wow that's some regex-fu. I'll take a look at this tomorrow, thanks!

YrrepNoj:
> Edit: I made the OR operator work without having to duplicate the whole group, I think this is cleaner now and less hacky :) `(?im)^([a-z0-9\-.]+[\.|\-][a-z0-9\-]+:?[0-9]*:?[0-9]*)?/?(.+)$` https://regex101.com/r/F6hx4h/2

Doesn't look like that works for `localhost:5000/user/image:tag` or when people put fully qualified url with a protocol `https://some-saas/some-user/some-image:some-tag`.

I took inspiration from your regex and made this https://regex101.com/r/c3qm3d/1. Can you double verify that this would work for your address that you've been using?

.github/workflows/test-external.yml
```diff
@@ -0,0 +1,59 @@
...
+    steps:
+      - name: "Dependency: Install Golang"
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19.x
+
+      - name: "Dependency: Install Docker Buildx"
+        id: buildx
+        uses: docker/setup-buildx-action@v2
+
```
jeff-mccoy, line 30: you'll need to update this test with the new steps to install k3d

tworcester:
> I took inspiration from your regex and made this https://regex101.com/r/c3qm3d/1. Can you double verify that this would work for your address that you've been using?

I think the issue with your latest iteration will be the impact of only matching the last two levels of the path.  That may cause some unwanted image path mismatches.

Further expanding the base regex that zarf was originally using we can cover all the cases with this:
`(?im)^((?:(?:https|http)://)?[\w\-.]+[\.|\-|:][\w\-]+)?/?(.+)$`
Demonstration: https://regex101.com/r/xVUd5j/1

This will use non-matching groups for `http://` `https://` and I collapsed the `:` matcher into the or grouping we created.

jeff-mccoy:
I don't think @YrrepNoj out the test of how this will be used.  The matching will only be a courtesy to users, the actual string will be "match-sha1sum" (shasum of the entire string to avoid edge cases and collisions).

tworcester:
> I don't think @YrrepNoj out the test of how this will be used. The matching will only be a courtesy to users, the actual string will be "match-sha1sum" (shasum of the entire string to avoid edge cases and collisions).

Was this in relation to the regex?  I don't follow.

jeff-mccoy:
Yes. The regex will be the prefix for the name internally, the remaining string will be a sha1 hash of the whole original string.

jeff-mccoy:
Given
ghcr.io/stefanprodan/podinfo

Would become
stefanprodan/podinfo-fde68284f960b126cd59af72cf18250f9d42d987

or

podinfo-fde68284f960b126cd59af72cf18250f9d42d987

Still some details to work out.

tworcester:
Oh, I was focusing on the Regex to allow the registry replacement that happens below since that is what my error was with.  Am I barking up the wrong tree?

```go
package utils

import "regexp"

// For further explanation see https://regex101.com/library/PiL191 and https://regex101.com/r/PiL191/1
var hostParser = regexp.MustCompile(`(?im)^([a-z0-9\-.]+\.[a-z0-9\-]+:?[0-9]*)?/?(.+)$`)

// SwapHost Perform base url replacment without the docker libs
func SwapHost(src string, targetHost string) string {
	var substitution = targetHost + "/$2"
	return hostParser.ReplaceAllString(src, substitution)
}
```

.github/workflows/test-external.yml
```diff
@@ -0,0 +1,72 @@
...
+      - "docs/**"
+
+# Abort prior jobs in the same workflow / PR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  validate:
+    runs-on: self-hosted
```
jeff-mccoy, line 20: this should now be `runs-on: [self-hosted, linux, x64, ubuntu]`

jeff-mccoy:
Looks like we'll need these merge conflicts resolved @YrrepNoj before review now

YrrepNoj:
Closing this PR (again..) in favor for #754

PR #667, opened 2022-08-09
split minio into own pkg & cleanup/update postgres/minio examples

jeff-mccoy:


YrrepNoj (approved):
I know this is still in 'draft' but I tested this locally for funzies and was able to stand up all the pods postgres/minio pods without any issues.

examples/postgres-operator/pgadmin/values.yaml
```diff
@@ -11,4 +11,4 @@ resources:
     memory: "512Mi"
 env:
   email: "zarf@example.local"
-  password: "###ZARF_GIT_AUTH_PUSH###"
+  password: "###ZARF_VAR_PGADMIN_PASSWORD###"
```
Racer159, line 14: This isn't declared anywhere in a package.

Racer159 (changes requested):
It appears to boot properly locally, but the PGADMIN password is not settable.

examples/postgres-operator/pgadmin/values.yaml
jeff-mccoy: `git stash` strikes again.....

Racer159 (approved):
Approved though this may be a problem?
image

examples/postgres-operator/zarf.yaml
```diff
@@ -2,63 +2,55 @@ kind: ZarfPackageConfig
...
       - name: postgres-operator-ui
         url: https://opensource.zalando.com/postgres-operator/charts/postgres-operator-ui
-        version: 1.7.0
+        version: 1.8.2
         namespace: postgres-operator
         valuesFiles:
-          - values/postgres-operator-ui.yaml
+          - postgres/operator-ui-values.yaml
+    manifests:
+      - name: zarf-connect
```
YrrepNoj, line 27: The manifest later on in the file (line 46) shares the same manifest name.

When we generate a helm chart from a raw manifest internally we [currently](https://github.com/defenseunicorns/zarf/blob/3a6436c1cf03d4bbb3c76996167d03e9e136effe/src/internal/helm/chart.go#L176) use the manifest name as the chart name.
```tmpChart.Metadata.Name = fmt.Sprintf("raw-%s", manifest.Name)```



This means when the second component gets deployed, the raw manifest is turned into a helm chart that then upgrades the first `zarf-connect` manifest that got installed earlier, removing the service the other component installed. This screenshot shows the `zarf connect postgres-operator-ui` option is displayed the package is done being deployed since it did get installed originally but it's not actually available in the cluster because the `raw-zarf-connect` chart got upgraded immediately afterwards.


Screen Shot 2022-08-15 at 10 01 45 PM

**Review by YrrepNoj — DISMISSED** [review]

File: examples/postgres-operator/zarf.yaml (hunk 2 above)
Comment (YrrepNoj) [reply examples/postgres-operator/zarf.yaml:27]: We could add some validation logic to make sure this doesn't happen again within a single package but we should probably think about how to do this for different packages.

`zarf-connect` seems like a very reasonable manifest name (and I'm sure there are thousands names for other manifests that can be used often) for single package to have, but if someone were to deploy two independent packages who both had the reasonable `zarf-connect` name then one of the manifests would essentially get uninstalled.

At the same time, doing something like adding a random salt to the end of the chart names will make it harder to update zarf packages that have been deployed in the future.

Comment (Racer159) [reply examples/postgres-operator/zarf.yaml:27]: Would it make sense to scope the manifests to the component or package (something like `postgres-operator.zarf-connect`)

Comment (YrrepNoj) [reply examples/postgres-operator/zarf.yaml:27]: That's a good point, doing something like that will probably be good enough. 

Since the chart name is something internal to the cluster and not something someone would see unless they were really looking for it, I would vote towards including the `{package}.{component}.{manifest}` as the chart name just to make it less likely to have an accidental collision.

Comment (YrrepNoj) [reply examples/postgres-operator/zarf.yaml:27]: I forgot that the `releaseName` is [limited to 53 characters](https://helm.sh/docs/chart_template_guide/getting_started/#:~:text=TIP%3A%20The%20name%3A%20field%20is,are%20limited%20to%2053%20characters.) so maybe not to my last suggestion 🤦

Comment (jeff-mccoy) [reply examples/postgres-operator/zarf.yaml:27]: Yeah...that was the issue, it's actually super dumb...but I do have a terrible idea...

**Comment by YrrepNoj** [comment]
Bit of an aside, I couldn't manage to sign into the pgadmin dashboard. I've tried different combinations of email/password based on what I deployed with and what the default values for the chart are but I kept getting invalid password messages. Has anyone else managed to sign in?

🤷

**Comment by jeff-mccoy** [comment]
seems that pgadmin didn't like the .local email domain? 

Screen Shot 2022-08-22 at 1 39 07 AM

File: examples/postgres-operator/zarf.yaml (hunk 2 above)
Comment (YrrepNoj) [inline examples/postgres-operator/zarf.yaml:27]: Can we change this manifest name to something slightly more unique? It will still conflict with the other manifest until the '~sha1' naming feature gets added.

File: examples/postgres-operator/zarf.yaml [hunk 3]
```diff
@@ -2,63 +2,55 @@ kind: ZarfPackageConfig
...
-        url: https://repo1.dso.mil/platform-one/big-bang/apps/application-utilities/minio.git
-        version: 4.2.3-bb.1
-        namespace: minio-operator
-        gitPath: chart
-        valuesFiles:
-          - values/minio-instance.yaml
-
+          - pgadmin/values.yaml
+    manifests:
+      - name: zarf-connect
```
YrrepNoj, line 46: Can we change this manifest name to something slightly more unique? It will still conflict with the other manifest until the '~sha1' naming feature gets added.

YrrepNoj (changes requested):

examples/postgres-operator/zarf.yaml
jeff-mccoy: updated, thanks.

examples/postgres-operator/zarf.yaml
jeff-mccoy: updated, thanks.

YrrepNoj (approved):

Racer159 (approved):

PR #689, opened 2022-08-23
Uninstall the things

YrrepNoj:
Saving information about which packages/components/charts have been installed into the cluster.

This will make it easier to perform upgrades of packages as well as being able to uninstall already installed components.


src/internal/packager/deploy.go
```diff
@@ -93,6 +96,22 @@ func Deploy() {
 		os.Exit(0)
 	}
 
+	// @JPERRY set up installed package secret state info
```
YrrepNoj, line 99: Check if this package has already been deployed.

We might need to add some more logic to know what to do when the package we're trying to install is already installed into the cluster.

jeff-mccoy:
close in favor of #736

PR #692, opened 2022-08-24
968/9 variables improvements

### LINKED ISSUE #675: Load package variables from a file
For Zarf package variables it would be nice to load them from a file so that you can override many of them at once without needing to manually type them or specify `--set` on the command line (which may be captured in shell history)

It would be nice to specify a variable override potentially to values.yaml overrides (or similar) in helm.

Considered loading from env vars but this would make things more complicated for end users to set the env vars first (rather than simply point to a file)

### LINKED ISSUE #968: Make prompt default or only behavior
We should make prompting for user variables the default or only accepted behavior when in interactive mode.  This would ensure that more things were presented to the user allowing them to more clearly control their outcomes.

Flipping of prompt: true/false to noPrompt: true/false or the removal of the prompt key from variables all together and always prompting

We could leave things as they are but this makes things less explicit overall.

**Additional context**
This issue is a break away from #675 loading of variables from files.

Racer159:
## Description

This introduces the ability to load package variables from files and refines the workflow to reduce confusion.

## Related Issue

Fixes #968, #969 

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist before merging


adr/0007-package-variables-files-revisit.md
```diff
@@ -0,0 +1,135 @@
...
+   - *Cons*: This relies on implementing a `default` pointer and gives package authors less control
+1. Remove the `prompt` key and always prompt
+   - *Pros*: This makes things simpler overall
+   - *Cons*: This could make things unweildy in large packages and gives package authors less control
+
+> Note that for the purposes of this ADR it was decided to exclude considering options making use of the `description` or `matches` fields as those ultimately solve different problems and would only be tangentially related to the above.
+
+## Decision
+
+It was decided to allow the `default` key to be nillable to detect its existence and error out if `--confirm` and `prompt` are used together with no existing default or `--set`.  This would result in the following behavior:
```
jeff-mccoy, line 98: I still disagree with this conclusion, but whatever we decide as a team, I think it would be good to document who agrees and who disagrees in this ADR and any dissenting comments. Regardless, this is a team decision, not mine.

I've also thought about this some more and might have another option. Can you clarify how this might change if we had a "validation" key as well? I think a viable option is to leave the current behavior as-is and provide a regex matcher so the package author can state what is acceptable. This way, in the case of the implied nil being used to force a failure, instead the failure can occur by setting the validation as requiring some content vs an empty string. 

For the first concern with the current implementation, allowing package authors to explicitly state what is a valid value will enable them to ensure the most important outcome: that an acceptable value was provided for the variable. Forcing this into nuance of whether the default key is specified by the package author seems far less obvious to me if I don't understand the details of yaml or golang typing.

For the second concern, the purpose of prompt really is to serve as a way to expose the variable as a choice for users in interactive mode. This is important because a persona only using interactive mode would have no way to set a variable that isn't set to prompt without using a flag. Prompt gives package authors a way to surface an important variable the user really should look at or change. 

Here's how I see the current and potential future keys working:

```yaml
name: the identifier for the variable
description: (not implemented yet), a message to explain to users the purpose of the variable or additional info to help decide what to enter
default: a value to suggest for users in interactive mode or the default value with --confirm; it's presence or absence shouldn't have nuanced meaning, it should just be a default value if provided, no other side effects
prompt: give users in interactive mode the ability to change the variable during package deploy, no meaning for --confirm as the key name itself requires interaction; this could have been interactivePrompt or similar I guess, but that would be longer
validation: (not implemented yet), regex to match whatever conditions the package author needs, including "not an empty string" 
```

Thoughts?

Racer159: Definitely want @YrrepNoj and @RothAndrew to weigh in too but I don't really see that `default` behavior as a nuance but rather as being more clear.  IMO if an author doesn't provide a default they shouldn't get a default (regardless of their knowledge of yaml/golang).  I agree validator or match would be a good thing to implement down the road, but IMO it is a bandaid to patch over on top of this problem.

adr/0007-package-variables-files-revisit.md
```diff
@@ -0,0 +1,135 @@
...
+   - *Pros*: This makes things simpler and allows us not to have to worry about abuses of this key.
+   - *Cons*: This removes options from package authors where they may want to set a legitimate default.  Overrides files also don't natively travel along with the package.
+1. Allow the `default` key to be unset if it does not exist (no implicit "")
+   - *Pros*: This allows package authors to set defaults that travel inside of a package and can be potentially carried into generated example override files.
+   - *Cons*: This would require the use of a pointer in the package types definitions.
+1. Keep the current `default` behavior and force an implicit ""
+   - *Pros*: This is not a breaking change and allows us to not have to have a pointer in our type definitions
+   - *Cons*: This forces an implicit default onto package creators.
+
+### Prompt
```
YrrepNoj, line 77: Want to quickly through out another option where we essentially turn `prompt` into an 'interactive only' configuration and ignore it during `--confirm` but then we add another field to the schema along the lines of `requireInput` where we would error out if someone tried to deploy with `--confirm` without setting that field.

- *Pros*: Non-breaking change, avoids NEEDing a pointer (if that's an issue), gives package creates even more control, not too different from what we currently do.
- *Cons*: Another addition to the schema, having `prompt` only really matter in 'interactive mode' feels weird at first

YrrepNoj: My main concern with this ADR is I think package creators should have the ability ensure a field is set during deploy without making deploying more cumbersome for the user (requiring every field be set, adding a new flag for using defaults, etc.). I am not too concerned with an implicit empty string if the package creator can dictate that specific variables are not allowed to be empty strings.

adr/0007-package-variables-files-revisit.md
RothAndrew: I agree with @Racer159 on all points. If an author didn't provide a default Zarf should not inject one behind the scenes.

adr/0007-package-variables-files-revisit.md
RothAndrew: > I am not too concerned with an implicit empty string if the package creator can dictate that specific variables are not allowed to be empty strings.

My problem with this is it would expect that the author will know there will be an implicit empty string, and forces them to add the the regex matcher to avoid unintentional side effects. It will be both error prone and annoying to the author.

Racer159: Yeah and I see the `prompt` and `default` keys as separate (but related) issues.  I am more firm on `default` being unset by default but could also support the addition of another `--ignore-prompt` flag or another schema key of `requireInput` (personally would lean toward another flag in that case).

adr/0007-package-variables-files-revisit.md
```diff
@@ -0,0 +1,135 @@
...
+# This would continue to work, setting NOTHING to "hello", NO_PROMPT to "no-prompt", NO_DEFAULT to "today" and BOTH to "both"
+
+zarf package deploy package.zst --confirm --set NOTHING=hello
+# This would not work, erroring out with: "Variable NO_DEFAULT does not have a default, when using `--confirm` you must specify it with `--set` or `--set-file`."
+
+zarf package deploy package.zst --confirm
+# This would not work, erroring out with: "Variable NOTHING does not have a default, when using `--confirm` you must specify it with `--set` or `--set-file`."
+
+zarf package deploy package.zst
+# This would not work, erroring out with: "Variable NOTHING does not have a default, and is not prompted for interactively. You must specify it with `--set` or `--set-file`."
```
RothAndrew, line 130: I'm not really getting how this case is valuable to the user. It would make more sense to me for Zarf to just go ahead and prompt for this variable if the user is in "interactive mode".

I see `prompt: true` as valuable only when paired with a default, in cases where the package author wants to provide a suggestion while also giving the user of the package 
image

RothAndrew: hang on that diagram is a big wrong. fixing it now..

RothAndrew: image

Racer159: Here is more what I was thinking from the ADR (this is also close to current behavior except there is no "Default -> No" step)
image

RothAndrew: ah yep, you're right that it would not cause a full cycle of the feedback loop if they can --set it on the CLI. As long as the error message says that (which is happening in the current proposal), then that's likely fine, though it would require them to use a flag when they were previously not using any

RothAndrew: I like the idea of changing `prompt:` which defaults to false to `suppressPrompt:` which defaults to false. Then it is an intentional action by the package author to let the user get down to the Fail at the bottom, rather than an unintended consequence of forgetting to add the field.

Racer159: And here is the full logic tree: (@Madeline-UX for context as well)

image

Racer159: And for more clarity, here is the currently implemented workflow:

image

Racer159: @YrrepNoj thoughts on the above and potential changing of `prompt` to `suppressPrompt` (or similar)?

YrrepNoj: I like both the logic tree and the change of `prompt` -> `suppressPrompt`

jeff-mccoy: A breaking change to make asking uses in interactive mode the default behavior seems logical to me, we could make it a little shorter with noPrompt as well.

adr/0007-package-variables-files-revisit.md
jeff-mccoy: I don't think validation is a bandaid, I think it's the right eventual solution in all cases. But either way, please update the ADR with this proposed option, even if the team doesn't select it, it needs to be listed as an option.

Racer159: I agree that it should be implemented regardless (and will definitely document it as a COA for these issues).  I'm also gonna add some logic images to this too to help break things down.  This is how I see validation alone working:

Validation Behavior drawio

jeff-mccoy: So two things, --confirm can never have additional user prompting. We currently break anything that is missing/incorrect if --confirm is provided. Reason for this is --confirm is used for automation and needs to fail vs just hang for an indefinite amount of time.

What I'm describing is exactly this flow without the "intentional default" box. I maintain it's nuance and implicit and lacks specificity for perceived convenience at the risk of further confusion. Granted, in this case I'm the only one of us 4 that sees it this way I think. But in the particular case that someone needs to have a value present for the template string to be valid, a valuation key is far more robust than "did the author remember to put a default key before creating the packaging and shipping it off."  

Across the 58 sting declarations in zarf types, we do not have a single case of key omission being nil. However, we are able to leverage json schema required for most of those keys. I think the issue at hand is we don't know if required is actually true or not.  At this point I'd rather just add "required" or "noEmpty" or something than use the presence or lack of a default key to mean that.  Though I maintain that validation also does this and more too, and would take half an hour to implement.

adr/0007-package-variables-files-revisit.md
RothAndrew: I think this captures the decision point between "prompt" and "noPrompt". The benefit of "noPrompt" is that the failure condition is pulled out of the "path of least resistance" in the case where both package author and user don't know or care about taking additional action to avoid failure.
image

adr/0007-package-variables-files-revisit.md
Racer159: --confirm skips prompting in the above example, I only made the Fail purple to distinguish an intentional failure from an unintentional one.

Without detecting the presence of a default key how do you know if the default was intentional and remove that logic point?

Also, `noEmpty` would also fail if an author intentionally wanted an empty default (i.e. a license file).

jeff-mccoy: yeah sorry I updated the comment, I misread the --confirm flow.

jeff-mccoy: In the license key example, where empty is valid, you wouldn't set `noEmpty` to true, or you wouldn't add additional validation because nothing is already valid.  I do not prefer `require` or `noEmpty` at all, I think that `validation` is far more robust and clear. I was only suggesting as a point of illustration how that could work, though not preferred.  

I think we all agree validation is useful and powerful.  I also think we need to just assume it exists now, and I can make the PR for it this afternoon if that's easier. My question then is, how are these concerns not addressed with validation in place.

Racer159: Ah ok I read that as being jsonschema rules.  I would argue they are not addressed (hence bandaid) because validation is only optional and still leaves the door open for the implicit default of `empty string`

Racer159: @YrrepNoj and @RothAndrew though for their thoughts

RothAndrew: > how are these concerns not addressed with validation in place.

Adding a validation will be an intentional action taken by the package author. If they don't add one, because they forgot, or they don't know it exists, the "unintentional default" logic branch will still be there.

The only way of removing that "unintentional default" logic branch while keeping the "nullability" of that field out, is to make the default validation something like `.+`.

adr/0007-package-variables-files-revisit.md
Racer159: Yeah I like the idea of switching to `noPrompt` and will update the ADR accordingly since it looks like we all agree.

adr/0007-package-variables-files-revisit.md
Racer159: IMO `.+` as a validation default, while allowing us to avoid `nil`s, is less intuitive to package authors (most validators that exist default to `.*`)

jeff-mccoy: That's a good point @RothAndrew, it's really hard for us right now to assume either way, whether on the utility/effectiveness of default or what the default  validation rule would be.  We are basing all of this off of very little data and relying instead on our current contexts to make assumptions about what is logical.  I know we're trying to get to majority / norm practice and doing that without user feedback/data is quite hard. We probably just need to solicit feedback from various user groups before moving forward, both package creators and deployers.  In particular I want to make sure we capture the low-level IT admin persona in interviews.  This sounds like a @Madeline-UX thing.

I honestly don't have a good answer right now on how often variables will actually be acceptable to empty strings or not. For example, if a package author is setting an optional license key, maybe an empty string is totally acceptable.  But they could also be doing something crazy (and we have no way to stop them) like defining custom yaml in the variable.  I don't think that will be the typical use-case, but it is possible I suppose.

RothAndrew: As a current user of Zarf, package creator and deployer (DI2-ME). I highly dislike the idea of having to remember to add a validation to every field that can't be an empty string, which will be ~99% of the variables that I create.

jeff-mccoy: That's useful feedback, though I wouldn't put you in the deployer persona as much as expert package creator.  On DI2-ME I have a separate concern with exposing too many variables to the user on package deploy and think we should dig into that more separately.  We could always expose the "default variable validation" in the schema somewhere, though I have no idea where.

RothAndrew: Making the validation field required could work as well to eliminate the "unintentional default". I'd prefer that over it being optional if the fall-through case is that the empty string becomes the implicit default.

RothAndrew: Side note: I propose `match:` as the name of the field. It gives better indication as to what the expected value is in the absence of any documentation.

adr/0007-package-variables-files-revisit.md
Racer159: ADR updated here: https://github.com/defenseunicorns/zarf/blob/675-load-package-variables-from-file/adr/0007-package-variables-files-revisit.md

adr/0007-package-variables-files-revisit.md
Racer159: Yeah that is another reason why (at least to me) I like keeping these things discrete `default` | `prompt` | `match`.  IMO if we try to solve one with another that introduces a lot of confusion.

Racer159: And do agree it is worth validating this ADRs flows with @Madeline-UX

adr/0007-package-variables-files-revisit.md
Madeline-UX: Another way to handle prompting. 

I recommend that we move error validation to the package create. 

Ex. Upon package create if Zarf identifies any defaults that are left empty, it returns a message to the package creator reading something like this and requiring them to confirm.

Zarf detected these variables are missing a default value. If you do not specify a default value a value of "empty string" will be automatically set as the default. 

- [values and missing default] [Add value or edit value button]- [values and missing default] [Add value or edit value button]

Racer159: Updated the ADR to remove this on package create: https://github.com/defenseunicorns/zarf/blob/675-load-package-variables-from-file/adr/0007-package-variables-files-revisit.md

adr/0007-package-variables-files-revisit.md
Madeline-UX: > That's a good point @RothAndrew, it's really hard for us right now to assume either way, whether on the utility/effectiveness of default or what the default validation rule would be. We are basing all of this off of very little data and relying instead on our current contexts to make assumptions about what is logical. I know we're trying to get to majority / norm practice and doing that without user feedback/data is quite hard. We probably just need to solicit feedback from various user groups before moving forward, both package creators and deployers. In particular I want to make sure we capture the low-level IT admin persona in interviews. This sounds like a @Madeline-UX thing.
> 
> I honestly don't have a good answer right now on how often variables will actually be acceptable to empty strings or not. For example, if a package author is setting an optional license key, maybe an empty string is totally acceptable. But they could also be doing something crazy (and we have no way to stop them) like defining custom yaml in the variable. I don't think that will be the typical use-case, but it is possible I suppose.

@jeff-mccoy I agree that this seems to be an area with a lot of both risk and assumptions and needs more validation.  I Just may need some technical support to come up with the best way to elicit the information we need to answer this question.

Madeline-UX: @jeff-mccoy If we have no validation for the need for nil at this time. How difficult would it be to build without it and then add in at a later point if validation shows that it is needed?

adr/0007-package-variables-files-revisit.md
```diff
@@ -0,0 +1,217 @@
...
+
+- Any variables that do not currently have a `default` key specified have a new error state to handle on `--confirm`
+- `noPrompt` (formerly `prompt: false`) must have a `default` key set in order to use that key.
+- Package authors must flip their `prompt` logic to `noPrompt`.
+
+This choice has the following benefits:
+
+- Prompting is now the default which helps our interactive persona
+- Errors asking the user to specify command flags only happen when they are already using the `--conrifm` flag
+- Implicit `""` defaults no longer are possible
```
RothAndrew, line 216: For clarity I would add to this line that explicit `""` is still able to be used if that is the author/user's desire

.hooks/verify-zarf-schema.sh
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env sh
 go run main.go internal config-schema > zarf.schema.json
+go run main.go internal override-schema > zarf-override.schema.json
```
RothAndrew, line 3: Looks like this is not the only place this is being done. See https://github.com/defenseunicorns/zarf/blob/master/.github/workflows/zarf-schema-check.yml#L39

We should either add this new stuff to there too, or make the other place use this shell script, so that we are more DRY

jeff-mccoy: Yeah this probably fits in the makefile and then called by the action/hook.

adr/0007-package-variables-files-revisit.md
Racer159: Added

.hooks/verify-zarf-schema.sh
Racer159: Changed the way this works, it now checks all docs (including CLI docs) and is integrated into make

Racer159:
@Madeline-UX from talking with @jeff-mccoy last night I think this PR has a need to validate with users (potentially just from the k8s slack) to gather input on what they expect.  With @jeff-mccoy, @RothAndrew and @YrrepNoj  being devs there can be a bias with the different ways that instantiating a "variable" can be done with different languages acting differently (`go` and `bash` will initialize a string as an empty string, `javascript`, `java`, `c`, and others will initialize as null or undefined, and `python` will force you to set it to None or an empty string).

Racer159:
Specifically:

```
variables:
  - name: "NOTHING"
    # default: ""   <- currently under the hood
    # prompt: false <- currently under the hood
  - name: "NO_PROMPT"
    default: "no-prompt"
    # prompt: false <- currently under the hood
  - name: "NO_DEFAULT"
    # default: ""   <- currently under the hood
    prompt: true
  - name: "BOTH"
    default: "both"
    prompt: true
```

```
zarf package deploy package.zst --confirm --set NOTHING=hello --set NO_PROMPT=world --set NO_DEFAULT=today --set BOTH=tomorrow

zarf package deploy package.zst --confirm --set NOTHING=hello --set NO_DEFAULT=today

zarf package deploy package.zst --confirm --set NOTHING=hello

zarf package deploy package.zst --confirm

zarf package deploy package.zst
```

Madeline-UX:
@Racer159 - I agree that more validation is needed. I am working off of the assumption that we can use the UI to get feedback from users from different skill levels.  I am working to integrate deploy time variables into the UI and have already been getting feedback on the feature (mostly questions at this time.) I would like some help with creating a more realistic prototype @RothAndrew that shows all the different options we are proposing here and see if that makes sense to users.

Madeline-UX:
> Specifically:
> 
> ```
> variables:
>   - name: "NOTHING"
>   - name: "NO_PROMPT"
>     default: "no-prompt"
>   - name: "NO_DEFAULT"
>     prompt: true
>   - name: "BOTH"
>     default: "both"
>     prompt: true
> ```
> 
> ```
> zarf package deploy package.zst --confirm --set NOTHING=hello --set NO_PROMPT=world --set NO_DEFAULT=today --set BOTH=tomorrow
> 
> zarf package deploy package.zst --confirm --set NOTHING=hello --set NO_DEFAULT=today
> 
> zarf package deploy package.zst --confirm --set NOTHING=hello
> 
> zarf package deploy package.zst --confirm
> 
> zarf package deploy package.zst
> ```

@Racer159 That looks like what the package definition would be - but what would the user see on deployment in the CLI?

RothAndrew:
Possible questions we could ask:

1. If you create a variable called "FOO" that does not have a default, and is not set to prompt the user for a value, what should happen?
    1. The system will automatically assign "" (an empty value / empty string) as the value of FOO and continue processing
    2. The system will stop processing and return an error asking the user to set a value for FOO using `<command> --set "FOO=YourValueHere"`
    3. Something else?

2. If you create a variable called "BAR" that has a default and is not set to prompt the user for a value, what should happen?
    1. It uses the default unless the user has used `<command> --set "BAR=YourValueHere"`. If `--set` was used it uses that as the value instead.
    2. Something else?

3. If you create a variable called "BAZ" that does not have a default and is set to prompt the user for a value, what should happen if the user uses `--confirm`, which overrides interactive prompts?
    1. The system will automatically assign "" (an empty value / empty string) as the value of BAZ and continue processing
    2. The system will stop processing and return an error asking the user to set a value for BAZ using `<command> --set "BAZ=YourValueHere"`
    3. Something else?

4. If you create a variable called "QUX" that has a default and is set to prompt the user for a value, what should happen if the user uses `--confirm`, which overrides interactive prompts?
    1. ?

Madeline-UX:
@RothAndrew This is helpful. I think to get the feedback we are looking for having common or easily understood examples will be helpful. The users I have discussed variables with so far do not understand what they are being asked at a general level but the context of (DNS, Ports, etc) helps. So is there a real/common example for each of those use cases outlined above?

Madeline-UX:
@RothAndrew @Racer159 @jeff-mccoy  Here are some of the mock up's I did for variables in the UI. Don't think I captured all the different ways variables work but trying to explore how we would present all the different types of variables to a user. Desired outcome is that the user completes all variables correctly and is able to successfully deploy as they intended without having to edit or redo the deployment due to a variable being incorrectly implemented. 
Screen Shot 2022-09-08 at 5 33 51 PM
Screen Shot 2022-09-08 at 6 02 02 PM
Screen Shot 2022-09-08 at 6 02 07 PM

Racer159:
> 

The top is the package definition, and then the bottom are the commands the user runs.

Madeline-UX:
> > 
> 
> The top is the package definition, and then the bottom are the commands the user runs.

zarf package deploy package.zst --confirm --set NOTHING=hello --set NO_PROMPT=world --set NO_DEFAULT=today --set BOTH=tomorrow

zarf package deploy package.zst --confirm --set NOTHING=hello --set NO_DEFAULT=today

zarf package deploy package.zst --confirm --set NOTHING=hello

zarf package deploy package.zst --confirm

zarf package deploy package.zst

- So I don't see how this invites user interaction. - It also does not follow the current Zarf model to my knowledge for promoting user input.

Racer159:
> > > 
> > 
> > 
> > The top is the package definition, and then the bottom are the commands the user runs.
> 
> zarf package deploy package.zst --confirm --set NOTHING=hello --set NO_PROMPT=world --set NO_DEFAULT=today --set BOTH=tomorrow
> 
> zarf package deploy package.zst --confirm --set NOTHING=hello --set NO_DEFAULT=today
> 
> zarf package deploy package.zst --confirm --set NOTHING=hello
> 
> zarf package deploy package.zst --confirm
> 
> zarf package deploy package.zst
> 
>     * So I don't see how this invites user interaction. - It also does not follow the current Zarf model to my knowledge for promoting user input.

The `--confirm` flag will just execute without needing any interaction, the last command example without a `--confirm` flag would currently prompt for the following (... is for truncated output):

```
...
Deploy this Zarf package? (y/N)
...
Please provide a value for 'NO_DEFAULT': 
Please provide a value for 'BOTH' (both):
...
```

RothAndrew:
Can I get a quick SWAG on when this might merge? DI2-ME updated to the latest version of Zarf but this is what we really need to start the transition to making it so users don't have to fork the repo to use the software factory package

jeff-mccoy:
I'll poke around this in the next couple of days but think we might go in a slightly different direction with viper.

RothAndrew:
Cool, yeah I don't really care how it's done, just want the capability.

Are we talking a week? a month? longer? Trying to figure out when to anticipate when the big push / refactor for DI2-ME might be able to be undertaken.

Racer159:
@jeff-mccoy @YrrepNoj @RothAndrew @Madeline-UX reset this PR now that the Viper PR has been merged.

Madeline-UX:
@jeff-mccoy @Racer159 I am working to find users to validate this. I have a few who I think will be a good fit.

RothAndrew:
I'd like to be one of the users pls

jeff-mccoy:
Closing this ancient wonder as stale until we can circle back to it at some future point.

PR #707, opened 2022-08-30
Add Or operator to properly handle unique hostParser case

tworcester:
Added Or operator to handle cases where hostnames have more than one dash.

Issue found and discussed here: https://github.com/defenseunicorns/zarf/pull/666#issuecomment-1230638470

/cc @YrrepNoj

Let me know if I should make a separate issue for this and I will update the PR :)

### DIFF
--- a/src/internal/utils/image.go
+++ b/src/internal/utils/image.go
@@ -2,8 +2,8 @@ package utils
 
 import "regexp"
 
-// For further explanation see https://regex101.com/library/PiL191 and https://regex101.com/r/PiL191/1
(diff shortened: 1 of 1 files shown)


tworcester:
@jeff-mccoy I couldn't figure out how to update the regex library entry that you created, I think you will need to do that if you all accept this change :)

tworcester:
Updated with the changes called out in https://github.com/defenseunicorns/zarf/pull/666#issuecomment-1233081552

tworcester:
@YrrepNoj Okay I think we got it!
Updated link showing the replacement being awesome https://regex101.com/r/zE4Klc/1

PR #716, opened 2022-09-07
Vendor syft/migrate witness -> syft direct

jeff-mccoy:
Adds `zarf tools syft` command for full access to https://github.com/anchore/syft.  Also migrates from 3rd party wrapper (Wtiness) to direct Syft integration.  Refactored code to prepare for adding file-based SBOM generation in addition to image-based.


jeff-mccoy:
Screen Shot 2022-09-07 at 3 31 16 AM

Racer159 (approved):
Tested locally and this worked fine.  Like that we are moving to a builder pattern and just fixed a small spelling mistake.

src/cmd/tools.go
```diff
@@ -109,4 +110,13 @@ func init() {
...
 	registryCmd.AddCommand(craneCmd.NewCmdCopy(&cranePlatformOptions))
 	registryCmd.AddCommand(craneCmd.NewCmdCatalog(&cranePlatformOptions))
+
+	syftCmd, err := cli.New()
+	if err != nil {
+		message.Fatal(err, "Unable to create syft CLI")
+	}
+	syftCmd.Use = "syft"
+	syftCmd.Short = "SBOM tools provided by Anchore Syft"
+	syftCmd.Aliases = []string{"s"}
```
Racer159, line 120: Do we want to include an alias called "sbom" (i.e. `zarf tools sbom ...`) as well so that it is more obvious what this is?

Racer159 (changes requested):

src/cmd/tools.go
jeff-mccoy: Yes we do ❤️ feel free to push the change

jeff-mccoy:
In fact sbom might make more sense and syft as the alias as we did with k9s iirc

jeff-mccoy:
Awesome, thanks for finishing that up.

Racer159:
I fixed the root examples but we may want to fix the sub command examples as well (looks like this would require looping through the cobra child commands and fixing them manually)

jeff-mccoy:
I think elsewhere we just didn't include the vendored-in examples, I think it might be fine to leave them off as it's more a power-user feature and not really our commands anyway.

Racer159 (approved):

Racer159:
Removed the sub command examples.  Left the high-level examples just as a quick start for folks

PR #719, opened 2022-09-07
Refactor CI to work on regular github runners + fix PLG support for K8s 1.25

jeff-mccoy:
Migrate back to regular-sized github runners.

- Remove notification/helm controllers from flux example (cut CPU need in half)
- Autoscale via `kubectl` coredns & zarf-agent to one pod
- Reset zarf-registry CPU req to original 100m
- Remove unnecessary build steps for requirements already present in the standard github runner 
- Updated PLG / disabled PSP to avoid breaking K8s 1.25+ (was breaking KinD on github runners)
- Updated gitea
- Re-added caching for go builds/modules (runners were taking > 5 mins to build the CLI without caching)
- Added automatic logfile creation on zarf commands
- Added runner artifact for downloading logfiles with complete debug info on pipeline runs


src/internal/message/message.go
```diff
@@ -46,10 +51,36 @@ func init() {
...
+		pterm.Debug.
+			WithShowLineNumber(true).
+			WithLineNumberOffset(offset).
+			WithDebugger(false).
+			WithWriter(logFile).
+			Println(a...)
+	}
 }
 
 func errorPrinter(offset int) *pterm.PrefixPrinter {
```
Racer159, line 86: If we are going to change our Debug printer, should we also standardize our other printers so that they are consistent when called?

jeff-mccoy: This is just a work-around for the debug printer, because if you don't do this, it will never print the debug message to the log file.

jeff-mccoy:
Okay you can see the log multi write operations in CI now.  CI will drop "debug-logs" artifact for each job that contains logs of all the runs with debug info included.  The CI runner logs will not have debug info (unless explicitly set that way), but the logs written to disk will.

Screen Shot 2022-09-08 at 4 54 07 PM

Screen Shot 2022-09-08 at 4 50 19 PM

Screen Shot 2022-09-08 at 4 51 54 PM

Screen Shot 2022-09-08 at 4 51 18 PM

.github/workflows/test-k3s.yml
```diff
@@ -52,7 +58,15 @@ jobs:
         # NOTE: "PATH=$PATH" preserves the default user $PATH. This is needed to maintain the version of go installed
         #       in a previous step. This test run will use Zarf to create a K3s cluster, and a brand new cluster will be
         #       used for each test
-        run: sudo env "PATH=$PATH" CI=true APPLIANCE_MODE=true make test-e2e ARCH=amd64
+        run: |
+          sudo env "PATH=$PATH" CI=true APPLIANCE_MODE=true make test-e2e ARCH=amd64
+          sudo chmod 644 /tmp/zarf-*.log
```
Racer159, line 63: I know this is because of the `sudo` stuff, but it did take a second looking at it to figure out why this was here.  Not required to change, but at least for me something like this would be more clear:

```
sudo chown $USER:$(id -gn) /tmp/zarf-*.log
```

src/cmd/prepare.go
```diff
@@ -30,7 +30,7 @@ var prepareTransformGitLinks = &cobra.Command{
 		host, fileName := args[0], args[1]
 
 		// Read the contents of the given file
-		content, err := ioutil.ReadFile(fileName)
+		content, err := os.ReadFile(fileName)
```
Racer159, line 33: Should we switch to os from now on? There are many other places ioutil still exists.

jeff-mccoy: Yeah the way we've dealt with other lint errors is to fix them when we touch a file vs go fix everywhere. I'd prefer to fix everywhere in another PR to keep this one less cluttered if that works.

.github/workflows/test-k3s.yml
jeff-mccoy: That's fair, I think we could even just change user and that would be enough for it to be read.

src/cmd/prepare.go
Racer159: That works

.github/workflows/test-k3s.yml
Racer159: Yeah that should work as well

Racer159 (approved):
lgtm

PR #735, opened 2022-09-15
razzle's web based ui additions

Noxsios:
fix some stuff


playwright.config.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+const config: PlaywrightTestConfig = {
+	testDir: './src/ui/tests',
```
jeff-mccoy, line 14: Can we move these tests to `src/test/ui` so all our testing is one place?

playwright.config.ts
```diff
@@ -0,0 +1,107 @@
...
+	timeout: 30 * 1000,
+	expect: {
+		/**
+		 * Maximum time expect() should wait for the condition to be met.
+		 * For example in `await expect(locator).toHaveText();`
+		 */
+		timeout: 5000
+	},
+	/* Run tests in files in parallel */
+	fullyParallel: true,
```
jeff-mccoy, line 25: This will be a problem since we will have to hit a real cluster for each test and those cant be run on top of each other.

src/ui/app.html
```diff
@@ -6,6 +6,13 @@
 		<meta name="viewport" content="width=device-width" />
 		%sveltekit.head%
 	</head>
+	<style>
+		html,
```
jeff-mccoy, line 10: I wonder if we should do some sort of CSS reset instead since we may have to deal with older browers?

src/ui/routes/+page.svelte
```diff
@@ -1,3 +1,63 @@
...
+	import bigZarf from '@images/zarf-bubbles.png';
+	let zarfStatus = false;
+</script>
+
+<svelte:head>
+	<title>Zarf</title>
+</svelte:head>
+
+{#if zarfStatus === false}
+	<section class="hero">
```
jeff-mccoy, line 12: Isn't hero a bootstrap term for this?  Am I crazy?

src/ui/lib/components/header.svelte
```diff
@@ -0,0 +1,52 @@
+<script>
+	import logo from '@images/zarf-logo.png';
+</script>
+
+<header class="header">
```
jeff-mccoy, line 5: This seems like it should use https://github.com/material-components/material-components-web/tree/master/packages/mdc-top-app-bar maybe?

src/ui/routes/+layout.svelte
```diff
@@ -18,26 +18,15 @@
 	}
 </script>
 
-<section>
-	<img alt="Zarf logo" id="logo" src={logo} />
+<Header />
 
+<!-- <section>
```
jeff-mccoy, line 23: Don't comment code out, just kill it with fire, that's what SCM is for.

src/ui/lib/api.ts
```diff
@@ -3,33 +3,17 @@ import type { ZarfState } from './api-types';
...
-): Promise<Response> {
-	try {
-		const response = await fetch(`/${BASE_URL}/${url}`, {
-			method,
-			headers: {
-				Authorization: `Bearer ${MAGIC}`,
-				'Content-Type': 'application/json'
-			},
-			body
+const getClusterState = async (): Promise<ZarfState> =>
```
jeff-mccoy, line 11: Going to play with this some more, not a fan of this rewrite of this file and it will lead to a lot of needless boilerplate.  This function wasn't even getting called from svelte, so this are doesn't really need to be rewritten.

playwright.config.ts
Noxsios: for sure, will move

src/ui/app.html
Noxsios: Stuff like [html conditional comments](https://www.quirksmode.org/css/condcom.html) to import polyfills / css compats?

src/ui/routes/+page.svelte
Noxsios: ya, `hero` is the classic term for a full centered page, usually a landing page with a "call to action" portion.  This case being the detection that zarf init has not run.

src/ui/routes/+layout.svelte
Noxsios: 🔥📕

src/ui/lib/components/header.svelte
Noxsios: ill play around w/ it

playwright.config.ts
Noxsios: https://playwright.dev/docs/test-parallel#disable-parallelism we can disable globally, and re-enable on a per test file basis to speed up tests we know can be parallelized

jeff-mccoy:
So the actually API call issue with nothing more than wrapping the promises again so svelte would rerun the call.  The actual change needed was `getState: request<ZarfState>('cluster/state'),` -> `getState: () => request<ZarfState>('cluster/state'),`.  I've largely rewritten the API back/front stuff now, will need your help to get the PR cleaned up to merge.

jeff-mccoy (approved):

src/ui/routes/deploy/+page.svelte
```diff
@@ -0,0 +1,38 @@
+<script>
+	import { Button, Stepper } from '@ui';
+	import Container from '$lib/components/container.svelte';
+	import PackageCard from '$lib/components/package-card.svelte';
+	import Icon from '$lib/components/icon.svelte';
+	import initConfig from '../packages/sample.json';
```
YrrepNoj, line 6: @Noxsios Did this file get committed?

jeff-mccoy: No he sent it to me, working on it 😘

PR #744, opened 2022-09-19
Check for CVEs with Grype in actions

Racer159:
## Description

This checks for CVEs in github actions to ensure we are updating deps properly.

## Related Issue

Fixes # N/A

## Type of change

## Checklist before merging


jeff-mccoy (approved):
Great addition 👏

PR #751, opened 2022-09-20
Make init package idempotent

jeff-mccoy:
## Description
This PR resolves issues with re-running `zarf init` on an existing zarf cluster.  Specifically, it resolves an issue with the injector not excluding itself from the mutating webhook and also the seed registry not using persistence causing issues with upgrades.


Racer159:
@jeff-mccoy @JasonvanBrackel @YrrepNoj thoughts on splitting the test for this into its own task under https://github.com/defenseunicorns/zarf/issues/774? If we get this in where we can run double inits we could add a new ci workflow for handling upgrade testing.

jeff-mccoy:
@Racer159 I'm totally fine with that

JasonvanBrackel:
Agree with @Racer159.  Fine with that

Racer159 (approved):
lgtm.  Will need https://github.com/defenseunicorns/zarf/issues/774 eventually

Issue #375, opened 2022-03-10
Define optimal multi-distro HA solution for the Docker Registry


jeff-mccoy:

jeff-mccoy:
## A. Require RWX StorageClass for HA
  ### Pros:
  - Simplest deployment complexity for Zarf
  - Requires only single push of images to the registry
  - Protected against pod eviction / node failure
  ### Cons:
  - Places HA burden on consumers
  - Could discourage HA which could be a real problem
  - May not be realistic for all systems
  - Not all RWX options are equally reliable, e.g. NFS 

&nbsp;

## B. Deploy minio or use existing S3-compatible storage
  ### Pros:
  - Standard charts exist to deploy minio
  - Well-established  / mature option
  - Requires only single push of images to the registry
  - Protected against pod eviction / node failure
  ### Cons:
  - Largest resource footprint
  - Non-trivial helm chart
  - May introduce additional cost/latency if use cloud object storage

&nbsp;

## C. Use Statefulset combined with InitContainer for automatic image syncing in-cluster
  ### Pros:
  - Very portable 
  - No resource overhead or deployment complexity to manage 
  - May not require any StorageClass or at least not RWX  
  - Requires only single push of images to the registry
  - Leverages only standard OCI capabilities 
  ### Cons:
  - Additional code must be written to handle pod-to-pod syncing using Crane
  ### Details:
  - Deploy 3+ pods with a Statefulset for predictable naming / rolling of pods
  - Push the image updates to **pod-0**
  - Start the pod roll ([reverse-order](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#rolling-updates))
  - Each pod (except **pod-0**, which syncs from any other pod in case of eviction) runs an init container first to sync to **pod-0** using the zarf injector go binary, this prevents nodes from trying to pull from that pod before it's updated
  - After all pods are updated, continue `zarf package deploy`

&nbsp;

## D. Use existing helm chart registry deployment and push copies of the images to each pod 
   ### Pros:
  - Very portable 
  - No resource overhead or deployment complexity to manage 
  - Leverages only standard OCI capabilities 
  ### Cons:
  - Requires pushing images o(n) based on number of registry pods deployed, some optimizations exist due to the way OCI Distribution works
  - Does not resolve pod eviction / node failure 
  ### Details:
  - Instead of using a service binding for `zarf connect`, bind to each pod individually and update every registry pod
  - After all pods are updated, continue `zarf package deploy`

[...]

PR #754, opened 2022-09-20
Feature: Support Using Zarf With an External Registry and Repository

### LINKED ISSUE #560: Support external registry on zarf init

### LINKED ISSUE #570: Support for using external Gitlab instead of gitea on zarf init

YrrepNoj:
## Description
This PR introduces the ability to connect to an already existing (and reachable) Container Registry and/or Git Repository during the `zarf init` command.

Closes #570 (Support using an external git server)
Closes #560 (Support using an external registry)
This implementation will serve as a good midway point on having a fully HA in-cluster registry #375.

## PR Feature List
- Added several flags to the `init` command to support using an external git repository
- Added several flags to the `init` command to support using an external container registry
- Update `zarf connect registry` to direct to `{HOST}/v2/_catalog` (this was confusing some other people since it would originally seem like the registry was returning an empty page)
- Add utility function to create a tunnel to a service URL
- Created slightly better regexp for replacing the host from a `containerImage` url
- semi-refactored the `zarf package deploy` logic

## Breaking Changes List
- We are changing the structure of the names of repos & containers we are pushing (we are simplifying the name and adding a sha1 hash of the original name to the end of the name)


src/internal/git/push.go
```diff
@@ -98,23 +128,23 @@ func push(localPath, tunnelUrl string, spinner *message.Spinner) error {
...
-		},
-	}
-
-	err = repo.Fetch(fetchOptions)
-	if errors.Is(err, transport.ErrRepositoryNotFound) {
-		message.Debugf("Repo not yet available offline, skipping fetch")
-	} else if err != nil {
-		return fmt.Errorf("unable to fetch remote cleanly prior to push: %w", err)
-	}
+	// // Fetch remote offline refs in case of old update or if multiple refs are specified in one package
```
YrrepNoj, line 131: @Racer159 Can I have some extra eyes / help debugging this later? 

Keep this in was failing the push when two separate refs existed in for the same repo but commenting it out seemed to work fine.

YrrepNoj, line 131: @Racer159 Can I have some extra eyes / help debugging this later? 

Keep this in was failing the push when two separate refs existed in for the same repo but commenting it out seemed to work fine.

Racer159 (changes requested):

Racer159:
@YrrepNoj just added a review.  We should probably also add a note to the actual docs on what is supported/not.  i.e. GitLab and Gitea will work fine, but other git hosts will not because they do not support push-to-create... (same for any registry gotchas, though I don't think there really are too many of those).

RothAndrew:
Will this work with AWS CodeCommit for repos and AWS ECR for registry? That's the primary avenue I'm seeing for stuff I'm working with. AFAIK anything else either wouldn't be approved (must be GovCloud inside the pre-authorized list of services), or would necessitate having to run some other stack which ultimately wouldn't really help any.

jeff-mccoy:
I think we should make testing with code commit and ECR a top priority, without that functionality this PR is far less valuable.

src/cmd/initialize.go
```diff
@@ -108,12 +121,45 @@ var initCmd = &cobra.Command{
...
+	initCmd.Flags().StringVar(&config.InitOptions.GitServer.Address, "git-url", "", "External git server url to use for this Zarf cluster")
+	initCmd.Flags().StringVar(&config.InitOptions.GitServer.PushUsername, "git-push-username", config.ZarfGitPushUser, "Username to access to the git server Zarf is configured to use. User must be able to create repositories via 'git push'")
+	initCmd.Flags().StringVar(&config.InitOptions.GitServer.PushPassword, "git-push-password", "", "Password for the push-user to access the git server")
+	initCmd.Flags().StringVar(&config.InitOptions.GitServer.PullUsername, "git-pull-username", "", "Username for pull-only access to the git server")
+	initCmd.Flags().StringVar(&config.InitOptions.GitServer.PullPassword, "git-pull-password", "", "Password for the pull-only user to access the git server")
+
+	// Flags for using an external registry
+	initCmd.Flags().StringVar(&config.InitOptions.RegistryInfo.Address, "registry-url", "", "External registry url address to use for this Zarf cluster")
+	initCmd.Flags().IntVar(&config.InitOptions.RegistryInfo.NodePort, "nodeport", 0, "Nodeport to access a registry internal to the k8s cluster. Between [30000-32767]")
```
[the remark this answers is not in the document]
YrrepNoj: This actually can be used for an internal Zarf registry. This flag already existed in case the nodeport 31999 was already being used in the k8s cluster.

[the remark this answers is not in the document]
Racer159: Ah ok gotcha that makes sense

src/internal/git/utils.go
```diff
@@ -26,28 +28,55 @@ type Credential struct {
...
 }
 
-func MutateGitUrlsInText(host string, text string) string {
+// MutateGitURlsInText Changes the giturl hostname to use the repository Zarf is configured to use
+func MutateGitUrlsInText(host string, text string, gitUser string) string {
 	extractPathRegex := regexp.MustCompilePOSIX(`https?://[^/]+/(.*\.git)`)
 	output := extractPathRegex.ReplaceAllStringFunc(text, func(match string) string {
-		if strings.Contains(match, "/"+config.ZarfGitPushUser+"/") {
+		// TODO @JPERRY: This might be less valid of a check now that people can specify the git username
+		if strings.Contains(match, "/"+gitUser+"/") {
```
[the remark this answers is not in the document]
YrrepNoj: Sorry, could you clarify a little more about what you mean by 'look for the URL instead'?

src/internal/git/push.go
YrrepNoj: Thank you! It's not intuitive to me why getting the error `git.ErrForceNeeded` while fetching doesn't lead to any issues while pushing but 🤷 

And I agree on logging the wildcard error as a warning and attempting to push. I'll do that.

YrrepNoj:
@RothAndrew @jeff-mccoy 

Using ECR ✅  -> Not tested yet but I don't foresee any issues with testing this eventually)

Using CodeCommit 🚫 
Zarf needs the ability to 'push to create repo'. This is something supported by [Gitea](https://github.com/go-gitea/gitea/pull/8419) and [GitLab](https://docs.gitlab.com/ee/user/project/working_with_projects.html#create-a-new-project-with-git-push) but I don't see anything in AWS docs that supports this (and the time I spent playing around didn't bear any good results)

RothAndrew:
It'll be kinda lame if we have to build in custom support for repo creation per provider, but look at the bright side, maybe we'll actually have a use case that is well suited to using Go's new generics support 😁

Racer159:
I would be for approving this for now and then having another issue to look at genericizing the provider for git to work with other things like BitBucket and AWS Code Commit.

src/internal/git/utils.go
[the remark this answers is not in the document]
YrrepNoj: @Racer159 I think I found a workaround for this. We only needed this check because we were using the same handler for `create` and `update` webhook triggers. 

Now, the webhook will only attempt to mutate the git URL if it's a `create`

YrrepNoj:
@Racer159 I'll work on adding a note to describe the gotchas for using an external repo and registry. 

I believe this is ready for another round of reviews in the meantime.

src/internal/agent/hooks/flux.go
```diff
@@ -26,35 +28,71 @@ type GenericGitRepo struct {
...
 	if err := json.Unmarshal(r.Object.Raw, &gitRepo); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal manifest: %v", err)
+		return nil, fmt.Errorf("failed to unmarshal manifest: %w", err)
 	}
 
-	message.Info(gitRepo.Spec.URL)
+	// Patch updates of the repo spec
+	patches = populatePatchOperations(gitRepo.Spec.URL, gitRepo.Spec.SecretRef.Name)
 
-	replacedURL := git.MutateGitUrlsInText("http://zarf-gitea-http.zarf.svc.cluster.local:3000", gitRepo.Spec.URL)
```
Racer159, line 45: What if someone updates the upstream URL for something they deploy?  i.e. instead of getting big bang from https://github.com/DoD-Platform-One/big-bang.git they later change that to https://repo1.dso.mil/platform-one/big-bang/bigbang.git?  (Theoretically, I haven't fully tested this) The name of the GitRepository resource wouldn't be different and so would trigger an update but that URL would never be swapped leading to people (potentially) scratching their heads.  I think it would be good to have a check like we used to have, but instead of looking for the git user, we look for the airgapped git url since we can be (reasonably) certain that that is unique to the air gap.

Racer159 (changes requested):
This worked for my test cases, and generally looks good. I think the only potential gotcha is if GitRepositoryUrls ever change we wouldn't catch the update

src/internal/agent/hooks/flux.go
YrrepNoj: Just to make sure I'm following, are you suggesting we check if the hostname is matching the hostname we have saved in the Zarf state?

Racer159: yes

YrrepNoj: Whoops I did this yesterday but forgot to push apparently... This should be checking the hostname like you recommended now.

YrrepNoj: That's a very good point, this would be much better.

Racer159 (approved):

Racer159 (approved):
lgtm

PR #755, opened 2022-09-21
Add support for git commit hashes

### LINKED ISSUE #747: Support fallback when `git` tags are not available
Some repos do not have well specified `git` tags and you may not have access to the upstream to properly tag it.  For this a fallback would be good to have.

Allow for pulling of `git` commit SHAs in addition to tags

You could pull other refs, such as heads, but every commit has a SHA so that would likely be the most robust mechanism

**Additional context**
N/A

Racer159:
## Description

This adds support for git commit hashes, and makes git slightly more robust.

## Related Issue

Fixes #747

## Type of change

## Checklist before merging


src/internal/git/push.go
```diff
@@ -96,6 +96,23 @@ func push(localPath, tunnelUrl string, spinner *message.Spinner) error {
...
+		RefSpecs: []goConfig.RefSpec{
+			"refs/heads/*:refs/heads/*",
+			onlineRemoteRefPrefix + "*:refs/heads/*",
+			"refs/tags/*:refs/tags/*",
+		},
+	}
+
+	err = repo.Fetch(fetchOptions)
+	if err != nil {
+		spinner.Debugf("%s", err.Error())
```
YrrepNoj, line 112: Shouldn't we return the err here instead of trying to push a repo we failed to fetch?

src/internal/git/checkout.go
```diff
@@ -29,13 +29,13 @@ func CheckoutTagAsBranch(path string, tag string, branch plumbing.ReferenceName)
...
-	checkoutHashAsBranch(path, tagRef.Hash(), branch)
+	CheckoutHashAsBranch(path, tagRef.Hash(), branch)
 }
 
-// checkoutHashAsBranch performs a `git checkout` of the commit hash associated
+// CheckoutHashAsBranch performs a `git checkout` of the commit hash associated
 // with the provided hash
 // It will delete the branch provided if it exists
-func checkoutHashAsBranch(path string, hash plumbing.Hash, branch plumbing.ReferenceName) {
+func CheckoutHashAsBranch(path string, hash plumbing.Hash, branch plumbing.ReferenceName) {
```
YrrepNoj, line 38: This function was made public but nothing outside of this package is calling it?

src/internal/git/fetch.go
```diff
@@ -8,8 +8,8 @@ import (
 	goConfig "github.com/go-git/go-git/v5/config"
 )
 
-// fetchTag performs a `git fetch` of _only_ the provided tag
-func fetchTag(gitDirectory string, tag string) {
+// FetchTag performs a `git fetch` of _only_ the provided tag.
+func FetchTag(gitDirectory string, tag string) {
```
YrrepNoj, line 12: This function was made public but nothing outside of this package is calling it?

src/internal/git/fetch.go
```diff
@@ -48,3 +48,42 @@ func fetchTag(gitDirectory string, tag string) {
 		message.Fatal(err, "Not a valid tag or unable to fetch")
 	}
 }
+
+// FetchHash performs a `git fetch` of _only_ the provided commit hash.
+func FetchHash(gitDirectory string, hash string) {
```
YrrepNoj, line 53: This function was made public but nothing outside of this package is calling it?

src/internal/git/pull.go
```diff
@@ -35,14 +36,14 @@ func pull(gitUrl, targetFolder string, spinner *message.Spinner) {
 	gitCred := FindAuthForHost(gitUrl)
 
 	matches := strings.Split(gitUrl, "@")
-	fetchAllTags := len(matches) == 1
+	fetchAll := len(matches) == 1
```
YrrepNoj, line 39: This var name is confusing me a bit, especially since we are always using the `!` of its value when checking it.

Thoughts on changing it to something like `fetchSpecificRef := len(matches) == 2`? or `onlyFetchRef`?

src/internal/git/fetch.go
```diff
@@ -48,3 +48,42 @@ func fetchTag(gitDirectory string, tag string) {
...
+
+	gitURL := remotes[0].Config().URLs[0]
+	message.Debugf("Attempting to find hash: %s for %s", hash, gitURL)
+
+	gitCred := FindAuthForHost(gitURL)
+
+	fetchOptions := &git.FetchOptions{
+		RemoteName: onlineRemoteName,
+		RefSpecs: []goConfig.RefSpec{
+			goConfig.RefSpec(hash + ":" + hash),
```
YrrepNoj, line 76: It looks like this line (and an error check for `git.ErrTagExists` is the only difference between `FetchHash(...)` and `FetchTag(...)` 

Could we refactor this into a single `FetchRef(...)` function where we set the `RefSpecs` based on the output of the regexp you wrote in `pull.pull(..)`?

src/internal/git/push.go
Racer159: If the repo doesn't exist we would expect it to fail.  We could add an explicit catch for that error though, but given we had no catch here earlier I felt it would be best just to log the error for now.

src/internal/git/checkout.go
Racer159: Was playing around with this to much and erroneously "standardized" it will fix all of these.

YrrepNoj: Go is pretty unique where variables/functions that start with a lower case are 'package private' and things that start with an uppercase are 'public'

Racer159: The classic trap of, well maybe there is another reason the world is this way...

examples/git-data/README.md
```diff
@@ -13,9 +13,21 @@ To view the example source code, select the `Edit this page` link below the arti
 
 ## Tag-Provided Git Repository Clone
 
-Tag-provided `git` repository cloning is the recommended way of cloning a `git` repository for air-gapped deployments. Tag-provided clones are defined using the `url.git@tag` format as seen in the example of the `defenseunicorns/zarf` repository (`https://github.com/defenseunicorns/zarf.git@v0.12.0`).
+Tag-provided `git` repository cloning is the recommended way of cloning a `git` repository for air-gapped deployments. Tag-provided clones are defined using the `url.git@tag` format as seen in the example of the `defenseunicorns/zarf` repository (`https://github.com/defenseunicorns/zarf.git@v0.15.0`).
 
-A tag-provided clone only mirrors the tag defined in the Zarf definition. The tag will appear on the `git` mirror as the default branch name of the repository being mirrored, along with the tag itself.
+A tag-provided clone only mirrors the tag defined in the Zarf definition. The tag will be applied on the `git` mirror to the default trunk branch of the repo (i.e. `master`, `main`, or the default when the repo is cloned).
```
JasonvanBrackel, line 18: Add a why this is important.  That is what happens when you grab no label or a label like 'stable' or 'beta', where the meaning of that evolves as the code does.

src/internal/git/push.go
JasonvanBrackel: I would add an explicit catch earlier and surface the error.  That way it's clear where the failure is an why

JasonvanBrackel (changes requested):
See @YrrepNoj's earlier comment.

JasonvanBrackel (approved):

PR #775, opened 2022-09-23
717 tmpdir flag silently fails when not a real directory

### LINKED ISSUE #717: tmpdir flag silently fails when not a real directory
If you pass in the `--tmpdir` flag but provide a directory that does not exist, instead of creating that as a new directory or throwing a warning/error that the directory does not exist, Zarf silently fails and uses an empty string as the tmpdir (effectively making. `/` the basePath of the tmpdir.

This might also be the case for the `--zarf-cache` flag.

(screenshot with extra debug statements for dramatic affect)

Screen Shot 2022-09-07 at 10 16 34 AM

Noxsios:
## Description

Add error handling to inside of `utils.makeTempDir` instead of handling each instance it is run, since all errors will produce the same outcome.  Users must __first__ create the `--tempdir` path if they specify it.

This means that users will have to do `mkdir -p <some temp path>` before they will be able to run using the `--tempdir` flag.  This is already handled in CI tests, which is the reason for the silent error during actual usage.

## Related Issue

Fixes #717

## Type of change


src/internal/utils/io.go
```diff
@@ -18,10 +18,13 @@ const dotCharacter = 46
 
 var TempPathPrefix = "zarf-"
 
-func MakeTempDir(tmpDir string) (string, error) {
+func MakeTempDir(tmpDir string) string {
```
YrrepNoj, line 21: @Noxsios I think we might have had a slight miscommunication earlier. It's not the end of the world if we use `message.fatal()` for now but I think we should do our best to move that to as early in the function hierarchy as reasonably possible.

Could you add back the error to this return?

Context:
Eventually (sooner than later hopefully..) we're going to refactor how we handle all errors across Zarf so that we are bubbling errors back to their calling function as reasonably as possible. This will be important to give the Zarf UI and more structured and predictable backbone for its API.

I might have misspoke earlier when I said instead of just using fatal logs that call `os.Exit()` we could call `panic()` (and then do extra logic with `defer` and `recover()`). While that might make sense in some places, I don't think we want that to be the norm.

src/internal/utils/io.go
```diff
@@ -20,7 +20,7 @@ var TempPathPrefix = "zarf-"
 
 func MakeTempDir(tmpDir string) (string, error) {
 	tmp, err := os.MkdirTemp(tmpDir, TempPathPrefix)
-	message.Debugf("Creating temp path %s", tmp)
+	message.Debugf("Attempted to create temp path %s", tmp)
```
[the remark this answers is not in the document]
Noxsios: kk, what would you prefer, since this log message is running _after_ the operation, and does not know the status (success or failure)

src/internal/api/packages/read.go
```diff
@@ -36,7 +37,7 @@ func readPackage(w http.ResponseWriter, path string) (pkg types.APIZarfPackage,
 
 	tmpDir, err := utils.MakeTempDir(config.CommonOptions.TempDirectory)
 	if err != nil {
-		return pkg, err
+		return pkg, fmt.Errorf("tmpdir does not exist: %w", err)
```
Racer159, line 40: This would be better stated as "unable to create tmpdir: ".  There are other reasons this may fail (permissions, bad characters, etc)

src/internal/git/pull.go
```diff
@@ -16,7 +16,10 @@ import (
 const onlineRemoteName = "online-upstream"
 
 func DownloadRepoToTemp(gitUrl string, spinner *message.Spinner) string {
-	path, _ := utils.MakeTempDir(config.CommonOptions.TempDirectory)
+	path, err := utils.MakeTempDir(config.CommonOptions.TempDirectory)
+	if err != nil {
+		message.Fatalf(err, "Specified tmpdir does not exist, please create it: %s", config.CommonOptions.TempDirectory)
```
Racer159, line 21: Same as above, there may be other reasons besides non-existence that this may fail.

src/internal/helm/post-render.go
```diff
@@ -43,7 +43,10 @@ func NewRenderer(options ChartOptions, actionConfig *action.Configuration) *rend
 func (r *renderer) Run(renderedManifests *bytes.Buffer) (*bytes.Buffer, error) {
 	message.Debugf("helm.Run(renderedManifests *bytes.Buffer)")
 	// This is very low cost and consistent for how we replace elsewhere, also good for debugging
-	tempDir, _ := utils.MakeTempDir(config.CommonOptions.TempDirectory)
+	tempDir, err := utils.MakeTempDir(config.CommonOptions.TempDirectory)
+	if err != nil {
+		return nil, fmt.Errorf("tmpdir does not exist: %w", err)
```
Racer159, line 48: Same

src/internal/packager/common.go
```diff
@@ -42,7 +42,10 @@ type tempPaths struct {
 }
 
 func createPaths() tempPaths {
-	basePath, _ := utils.MakeTempDir(config.CommonOptions.TempDirectory)
+	basePath, err := utils.MakeTempDir(config.CommonOptions.TempDirectory)
+	if err != nil {
+		message.Fatalf(err, "Specified tmpdir does not exist, please create it: %s", config.CommonOptions.TempDirectory)
```
Racer159, line 47: Same

Racer159 (changes requested):

Racer159:
@Noxsios your first commit in this PR is not signed: `[handle tmdir path dne](https://github.com/defenseunicorns/zarf/pull/775/commits/1e5192ceaa65d48a79d9b7a3db050b072a8f774e)`  I can help with that if you need.

YrrepNoj:
@Racer159 it looks like all the commits are signed from my perspective

Racer159:
> @Racer159 it looks like all the commits are signed from my perspective

Ah yeah you're right I was looking at the wrong thing.

jeff-mccoy (approved):
Lgtm

PR #781, opened 2022-09-26
Zarf config file support using Viper

jeff-mccoy:
This PR adds support for using [Viper](https://github.com/spf13/viper) to support declarative automation for the Zarf CLI. As more flags are added to Zarf, it can become more imperative at runtime, Viper allows Zarf to remain fully declarative operations supported by automation.


src/cmd/package.go
```diff
@@ -156,33 +156,37 @@ func cachePathClean(cachePath string) bool {
...
+	packageCreateCmd.Flags().StringVarP(&config.CreateOptions.OutputDirectory, "output-directory", "o", v.GetString("package.create.output_directory"), "Specify the output directory for the created Zarf package")
+	packageCreateCmd.Flags().BoolVar(&config.CreateOptions.SkipSBOM, "skip-sbom", v.GetBool("package.create.skip_sbom"), "Skip generating SBOM for this package")
+	packageCreateCmd.Flags().BoolVar(&config.CreateOptions.Insecure, "insecure", v.GetBool("package.create.insecure"), "Allow insecure registry connections when pulling OCI images")
 
+	//  Always require confirm flag (no viper)
 	packageDeployCmd.Flags().BoolVar(&config.CommonOptions.Confirm, "confirm", false, "Confirm package deployment without prompting")
-	packageDeployCmd.Flags().StringVar(&config.CommonOptions.TempDirectory, "tmpdir", "", "Specify the temporary directory to use for intermediate files")
 	packageDeployCmd.Flags().StringToStringVar(&config.CommonOptions.SetVariables, "set", map[string]string{}, "Specify deployment variables to set on the command line (KEY=value)")
```
Racer159, line 180: Was this overlooked for deploy? I see the package.create.set, but no package.deploy.set.

src/cmd/package.go
```diff
@@ -156,33 +156,37 @@ func cachePathClean(cachePath string) bool {
...
-	packageCreateCmd.Flags().StringVar(&config.CommonOptions.TempDirectory, "tmpdir", "", "Specify the temporary directory to use for intermediate files")
-	packageCreateCmd.Flags().StringToStringVar(&config.CommonOptions.SetVariables, "set", map[string]string{}, "Specify package variables to set on the command line (KEY=value)")
-	packageCreateCmd.Flags().StringVar(&zarfImageCache, "zarf-cache", config.ZarfDefaultImageCachePath, "Specify the location of the Zarf image cache")
-	packageCreateCmd.Flags().StringVarP(&config.CreateOptions.OutputDirectory, "output-directory", "o", "", "Specify the output directory for the created Zarf package")
-	packageCreateCmd.Flags().BoolVar(&config.CreateOptions.SkipSBOM, "skip-sbom", false, "Skip generating SBOM for this package")
-	packageCreateCmd.Flags().BoolVar(&config.CreateOptions.Insecure, "insecure", false, "Allow insecure registry connections when pulling OCI images")
+	packageCreateCmd.Flags().StringToStringVar(&config.CommonOptions.SetVariables, "set", v.GetStringMapString("package.create.set"), "Specify package variables to set on the command line (KEY=value)")
```
Racer159, line 172: We should add a viper toml test for package create/destroy to ensure these don't get mangled in the future (using most of the settings we want to check for)

Racer159 (changes requested):

src/cmd/package.go
jeff-mccoy: fixed thanks!

jeff-mccoy:
So ran into a really interesting thing that I don't think we thought about before... when we bind multiple flags to a single variable, each binding overwrites the default state with the very last binding to it.  Ran into this because `config.CommonOptions.SetVariables` default (which was provided by Viper) was being overridden by the empty default also provided by Viper later on.  🤦

jeff-mccoy:
Fortunately, the only place we use this that it would actually cause problems is for `config.CommonOptions.SetVariables`

jeff-mccoy:
### `zarf prep gc`:
```toml
architecture = ''
log_level = 'info'
no_log_file = false
no_progress = false
tmp_dir = ''

[init]
components = ''
storage_class = ''

[init.git]
pull_password = ''
pull_username = ''
push_password = ''
push_username = 'zarf-git-user'
url = ''

[init.registry]
nodeport = 0
pull_password = ''
pull_username = ''
push_password = ''
push_username = 'zarf-push'
secret = ''
url = ''

[package]
[package.create]
insecure = false
output_directory = ''
skip_sbom = false
zarf_cache = '.zarf-image-cache'

[package.create.set]

[package.deploy]
components = ''
insecure = false
sget = ''
shasum = ''

[package.deploy.set]
```

src/cmd/prepare.go
```diff
@@ -94,15 +95,43 @@ var prepareFindImages = &cobra.Command{
...
 	prepareCmd.AddCommand(prepareTransformGitLinks)
 	prepareCmd.AddCommand(prepareComputeFileSha256sum)
 	prepareCmd.AddCommand(prepareFindImages)
+	prepareCmd.AddCommand(prepareGenerateConfigFile)
+
+	v.SetDefault(V_PKG_CREATE_SET, map[string]string{})
 
 	prepareFindImages.Flags().StringVarP(&repoHelmChartPath, "repo-chart-path", "p", "", `If git repos hold helm charts, often found with gitops tools, specify the chart path, e.g. "/" or "/chart"`)
+	prepareFindImages.Flags().StringToStringVar(&config.CreateOptions.SetVariables, "set", v.GetStringMapString(V_PKG_CREATE_SET), "Specify package variables to set on the command line (KEY=value)")
 	prepareFindImages.Flags().StringVar(&config.CommonOptions.TempDirectory, "tmpdir", "", "Specify the temporary directory to use for intermediate files")
```
Racer159, line 134: Isn't this set at the root now?

src/cmd/prepare.go
```diff
@@ -94,15 +95,43 @@ var prepareFindImages = &cobra.Command{
...
 	rootCmd.AddCommand(prepareCmd)
 	prepareCmd.AddCommand(prepareTransformGitLinks)
 	prepareCmd.AddCommand(prepareComputeFileSha256sum)
 	prepareCmd.AddCommand(prepareFindImages)
+	prepareCmd.AddCommand(prepareGenerateConfigFile)
+
+	v.SetDefault(V_PKG_CREATE_SET, map[string]string{})
 
 	prepareFindImages.Flags().StringVarP(&repoHelmChartPath, "repo-chart-path", "p", "", `If git repos hold helm charts, often found with gitops tools, specify the chart path, e.g. "/" or "/chart"`)
+	prepareFindImages.Flags().StringToStringVar(&config.CreateOptions.SetVariables, "set", v.GetStringMapString(V_PKG_CREATE_SET), "Specify package variables to set on the command line (KEY=value)")
```
Racer159, line 133: It may be good to add hint text to the decription here since it might not be obvious that `zarf prepare find-images` uses the same configuration key as `zarf package create` (same goes for other reuses)

We could also add specific keys for these commands as well, but either way I would argue it should be communicated to the user.

src/internal/message/message.go
```diff
@@ -75,6 +81,9 @@ func debugPrinter(offset int, a ...any) {
 
 	// Always write to the log file
 	if logFile != nil && !pterm.PrintDebugMessages {
+		now := time.Now().Format(time.RFC3339)
```
Racer159, line 84: :heart_eyes:

Racer159 (changes requested):

src/cmd/prepare.go
jeff-mccoy: We actually have to load this like this (in create and prepare) due to how the functions initialize, I'll leave a comment

src/cmd/prepare.go
jeff-mccoy: added

src/cmd/prepare.go
Racer159: I think it should also be in the output the user sees (i.e.)

```
Specify package variables to set on the command line as done for package create (KEY=value)
```

src/cmd/prepare.go
Racer159: I think it should also be in the output the user sees (i.e.)

```
Specify package variables to set on the command line as done for package create (KEY=value)
```

src/cmd/prepare.go
Racer159: ok... that seems like a bug. Also create doesn't have tmpdir?

src/cmd/prepare.go
jeff-mccoy: I may have misunderstood, I only added it to the non-obvious one (prepare).  The `generate-config` was designed to give users an example.  Are you saying we should include more details about the config file beyond the docs?

Screen Shot 2022-10-04 at 7 31 34 PM

src/cmd/prepare.go
jeff-mccoy: It's not a bug perse, it's just a unique quirk of how we are stringing viper -> cobra flags -> config variables. And it's always been that way, we just didn't notice before.

jeff-mccoy: Tmpdir was moved to a global command

Screen Shot 2022-10-04 at 7 34 20 PM

Racer159: Removing that line doesn't change the docs...

```
// prepareFindImages.Flags().StringVar(&config.CommonOptions.TempDirectory, "tmpdir", "", "Specify the temporary directory to use for intermediate files")

```

jeff-mccoy: ah yes, fixed now--that was a product a bad merge conflict resolution 🤦

Racer159 (approved):
lgtm

PR #785, opened 2022-09-26
Support git repo caching on package create

### LINKED ISSUE #750: Support a git cache directory, similar to the image cache directory
Some git repositories can be very large to clone down each time you perform a package build, being able to clone them to a cache directory and then work on them locally would speed up the cycle time of these Zarf packages.

An option to provide a git repository cache directory with (ideally) a default cache location similar to how images work.

We could provide this as an optional option or provide some way to pull a git repo from an already local clone but this would provide the most flexibility for the future and would be similar to features already in Zarf.

**Additional context**
Some repos with a lot of history or files (or that improperly use LFS) can be many GiBs in size and can be difficult to download/clone when building a Zarf package.

Racer159:
## Description

Add the ability to locally cache large git repositories when pulling them with Zarf

## Related Issue

Fixes #750

## Type of change

## Checklist before merging


src/internal/images/pull.go
```diff
@@ -45,7 +46,7 @@ func PullAll(buildImageList []string, imageTarballPath string) map[name.Tag]v1.I
 		if err != nil {
 			spinner.Fatalf(err, "Unable to pull the image %s", src)
 		}
-		imageCachePath := config.GetImageCachePath()
+		imageCachePath := filepath.Join(config.GetCachePath(), "images")
```
YrrepNoj, line 49: Can we turn the "images" string into a const somewhere? It would be nice to track what subdirectories Zarf is utilizing within its cache.

src/internal/git/utils.go
```diff
@@ -28,6 +28,11 @@ type Credential struct {
 	Auth http.BasicAuth
 }
 
+var (
+	// For further explanation: https://regex101.com/library/UfILls and https://regex101.com/rary/UfILls
+	gitURLRegex = regexp.MustCompile(`^(.+)\/([\w\-\.]+?)(\.git)?(@([\w\-\.]+))?$`)
```
YrrepNoj, line 33: Now that I know named capture groups are a thing, could you do that here?

src/internal/git/pull.go
```diff
@@ -1,87 +1,84 @@
...
-		RemoteName: onlineRemoteName,
-	}
+func pull(gitURL, targetFolder string, spinner *message.Spinner, repoName string) {
+	spinner.Updatef("Processing git repo %s", gitURL)
 
-	if onlyFetchRef {
-		cloneOptions.Tags = git.NoTags
+	gitCachePath := targetFolder
+	if repoName != "" {
+		gitCachePath = filepath.Join(config.GetCachePath(), fmt.Sprintf("repos/%s", repoName))
```
YrrepNoj, line 46: Like the `images` directory in `images/pull.go` could we make this a const? Having a hardcoded string for directory traversing seems sketchy.

Even if not, we should be using a nested `filepath.Join(..)` instead of doing `fmt.Sprintf("repos/%s")`

src/internal/utils/exec.go
```diff
@@ -19,7 +19,7 @@ const colorCyan = "\x1b[36;1m"
 const colorWhite = "\x1b[37;1m"
 
 // nolint
-func ExecCommandWithContext(ctx context.Context, showLogs bool, commandName string, args ...string) (string, string, error) {
+func ExecCommandWithContext(ctx context.Context, dir string, showLogs bool, commandName string, args ...string) (string, string, error) {
```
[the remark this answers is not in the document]
Racer159: done

src/internal/git/utils.go
Racer159: done

src/internal/images/pull.go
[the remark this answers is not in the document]
Racer159: done

src/internal/git/pull.go
Racer159: done

src/cmd/tools.go
```diff
@@ -95,13 +95,33 @@ var k9sCmd = &cobra.Command{
 	},
 }
 
+var clearCacheCmd = &cobra.Command{
+	Use:     "clear-cache",
+	Aliases: []string{"c"},
+	Short:   "Clears the configured git and image cache directory",
+	Run: func(cmd *cobra.Command, args []string) {
+		if zarfCache != config.ZarfDefaultCachePath && cachePathClean(zarfCache) {
```
YrrepNoj, line 103: Since the `zarfCache` variable has the default path set by cobra, I think this function can get simplified down to 

```Go
if cachePathClean(zarfCache) {
    if err := os.RemovedAll(cachePath); err != nil {
        message.Fatalf(.....)
    }
}
```

src/test/e2e/00_use_cli_test.go
```diff
@@ -82,5 +84,26 @@ func TestUseCLI(t *testing.T) {
...
+	// Pull once to test git cloning
+	stdOut, stdErr, err = e2e.execZarfCommand("package", "create", "examples/git-data", "--confirm", "--zarf-cache", cachePath, "--tmpdir", otherTmpPath, "--log-level=debug")
+	require.NoError(t, err, stdOut, stdErr)
+
+	files, err = os.ReadDir(gitCachePath)
+	require.NoError(t, err, "Error when reading git cache path")
+	assert.Greater(t, len(files), 1)
+
+	// Pull twice to test git fetching (from cache)
+	stdOut, stdErr, err = e2e.execZarfCommand("package", "create", "examples/git-data", "--confirm", "--zarf-cache", cachePath, "--tmpdir", otherTmpPath, "--log-level=debug")
```
YrrepNoj, line 98: Do we have any way to verify that Zarf is actually using the repo cache instead of pulling again?

src/test/e2e/00_use_cli_test.go
```diff
@@ -82,5 +84,26 @@ func TestUseCLI(t *testing.T) {
...
+	require.NoError(t, err, stdOut, stdErr)
+
+	// Test removal of cache
+	stdOut, stdErr, err = e2e.execZarfCommand("tools", "clear-cache", "--zarf-cache", cachePath, "--log-level=debug")
+	require.NoError(t, err, stdOut, stdErr)
+
+	_, err = os.ReadDir(cachePath)
+	require.Error(t, err, cachePath+": no such file or directory")
```
YrrepNoj, line 106: This display message seems to be what the expected err would be. Should the display message be more along the lines of `Did not receive expected error when reading a directory that should not exist`?

src/internal/git/utils.go
```diff
@@ -43,33 +48,33 @@ func MutateGitUrlsInText(host string, text string, gitUser string) string {
 }
 
 func transformURLtoRepoName(url string) (string, error) {
-	// For further explanation: https://regex101.com/library/UfILls and https://regex101.com/rary/UfILls
-	findRegex := regexp.MustCompile(`\/([\w\-]+)(.git)?(@([\w\-\.]+))?$`)
-	substrings := findRegex.FindStringSubmatch(url)
-	if len(substrings) == 0 {
+	matches := gitURLRegex.FindStringSubmatch(url)
+	if len(matches) != 6 {
 		// the first element in the return substrings is
```
YrrepNoj, line 53: I must of borked up this comment when my PR got merged. Could you fix it up in your PR?

src/internal/git/pull.go
```diff
@@ -1,87 +1,83 @@
...
 
-	// Gracefully handle no git creds on the system (like our CI/CD)
-	if gitCred.Auth.Username != "" {
-		cloneOptions.Auth = &gitCred.Auth
+	substrings := gitURLRegex.FindStringSubmatch(gitURL)
+	if len(substrings) == 0 {
+		// Unable to find a substring match for the regex
+		message.Fatalf("unable to get extract the repoName from the url %s", gitURL)
 	}
+	onlyFetchRef := substrings[4] != ""
```
YrrepNoj, line 53: substrings is made from a regexp with capture groups, can we use the group name instead of indexes?

src/cmd/tools.go
Racer159: I'll simplify it but also need to test that special urls don't mess it up (there is ~ handling in those other spots)

src/test/e2e/00_use_cli_test.go
Racer159: We test that the gitCache directory gets populated, and you can't git clone into an already existing folder, though I suppose in code we could just delete the folder and reclone...  Not sure how to test that...

src/test/e2e/00_use_cli_test.go
Racer159: I mixed up the args here.  Changed it to ErrorContains and added that text to be more explicit (and correct)

src/cmd/tools.go
Racer159: It le worked

YrrepNoj (approved):
LGTM!

jeff-mccoy:
Looks like mac has issues with the given test location chosen 

Screen Shot 2022-10-01 at 1 54 57 PM

jeff-mccoy:
Didn't look into it any more, but `make test-e2e ARCH=arm64` produces this on my m1 mac:

Screen Shot 2022-10-01 at 1 58 38 PM

Screen Shot 2022-10-01 at 1 59 22 PM

[zarf-2022-10-01-13-58-49-3620131400.log](https://github.com/defenseunicorns/zarf/files/9690996/zarf-2022-10-01-13-58-49-3620131400.log)

src/internal/git/utils.go
```diff
@@ -43,33 +48,33 @@ func MutateGitUrlsInText(host string, text string, gitUser string) string {
...
 		return "", fmt.Errorf("unable to get extract the repoName from the url %s", url)
 	}
 
-	// NOTE: The first element in the returned substrings is the combination of all the rest of the substrings....
-	//       So just skip the first element so we can get a hash without the version tag
-	repoName := substrings[1]
+	repoName := matches[gitURLRegex.SubexpIndex("repo")]
+	// NOTE: We remove the .git so that https://zarf.dev/repo.git and https://zarf.dev/repo resolve to the same repo
+	// (as they would in real life)
+	sanitizedURL := matches[gitURLRegex.SubexpIndex("baseURL")] + "/" + repoName + matches[gitURLRegex.SubexpIndex("atRef")]
```
Racer159, line 60: @jeff-mccoy here is what changed to make the repo SHAs different.  We remove the .git so that zarf.git and zarf are the same.

src/cmd/package.go
```diff
@@ -46,8 +45,8 @@ var packageCreateCmd = &cobra.Command{
 			baseDir = args[0]
 		}
 
-		if zarfImageCache != config.ZarfDefaultImageCachePath && cachePathClean(zarfImageCache) {
-			config.SetImageCachePath(zarfImageCache)
+		if isCachePathClean(config.CreateOptions.CachePath) {
```
jeff-mccoy, line 48: As a general convention, if it's a public variable already in config, we don't pass those around, but instead use the variable directly.

src/cmd/package.go
```diff
@@ -46,8 +45,8 @@ var packageCreateCmd = &cobra.Command{
...
+			config.SetCachePath(config.CreateOptions.CachePath)
```
jeff-mccoy, line 49: We also try to reserve setters for non-public variables.  If the variable is exposed publicly it will likely be read/written publicly in some places (i.e. Cobra flags) so for consistency we only access it via the public read/write as any future getter/setter would not be used every time otherwise.

src/cmd/package.go
```diff
@@ -175,10 +174,10 @@ func choosePackage(args []string) string {
 	return path
 }
 
-func cachePathClean(cachePath string) bool {
-	var isCleanPath = regexp.MustCompile(`^[a-zA-Z0-9\_\-\/\.\~]+$`).MatchString
-	if !isCleanPath(cachePath) {
-		message.Warnf("Invalid characters in Zarf cache path, defaulting to ~/%s", config.ZarfDefaultImageCachePath)
+func isCachePathClean(cachePath string) bool {
+	var isCleanPathRegex = regexp.MustCompile(`^[a-zA-Z0-9\_\-\/\.\~]+$`).MatchString
```
jeff-mccoy, line 178: No need to change, this pattern sort of bothered me on one level because it's not something I've seen very much in go, reminds me of Javascript prototype reuse--but I do agree it is actually quite readable.  Just a comment that it _feels odd_, but is overall likely more readable so 🤷.

src/internal/git/utils.go
```diff
@@ -43,33 +48,33 @@ func MutateGitUrlsInText(host string, text string, gitUser string) string {
...
 
-func transformURL(baseUrl string, url string, username string) (string, error) {
-	replaced, err := transformURLtoRepoName(url)
+func transformURL(baseURL string, url string, username string) (string, error) {
+	repoName, err := transformURLtoRepoName(url)
 	if err != nil {
 		return "", err
 	}
-	output := baseUrl + "/" + username + "/" + replaced
+	output := baseURL + "/" + username + "/" + repoName
```
jeff-mccoy, line 77: Not a must change, as there are plenty of legacy offenders in this file already, but we've been trying to move to `fmt.Sprintf()` vs string concatenation whenever we touch older code.

src/internal/git/utils.go
```diff
@@ -43,33 +48,33 @@ func MutateGitUrlsInText(host string, text string, gitUser string) string {
...
-	repoName := substrings[1]
+	repoName := matches[gitURLRegex.SubexpIndex("repo")]
+	// NOTE: We remove the .git and protocol so that https://zarf.dev/repo.git and http://zarf.dev/repo
+	// resolve to the same repp (as they would in real life)
+	sanitizedURL := matches[gitURLRegex.SubexpIndex("hostPath")] + "/" + repoName + matches[gitURLRegex.SubexpIndex("atRef")]
 
 	// Add sha1 hash of the repoName to the end of the repo
 	hasher := sha1.New()
-	_, _ = io.WriteString(hasher, url)
+	_, _ = io.WriteString(hasher, sanitizedURL)
```
jeff-mccoy, line 64: Probably a good idea to leave a comment whenever explicitly ignoring an error.

src/internal/git/clone.go
```diff
@@ -0,0 +1,66 @@
...
+
+	if onlyFetchRef {
+		cloneOptions.Tags = git.NoTags
+	}
+
+	gitCred := FindAuthForHost(gitURL)
+
+	// Gracefully handle no git creds on the system (like our CI/CD)
+	if gitCred.Auth.Username != "" {
+		cloneOptions.Auth = &gitCred.Auth
```
jeff-mccoy, line 28: Thank you for this 😆

Racer159:
> Looks like mac has issues with the given test location chosen
> 
> Screen Shot 2022-10-01 at 1 54 57 PM

Switched to using os.TempDir instead of /tmp if that helps?

src/internal/git/utils.go
Racer159: But but 🐕🐕🛷

src/cmd/package.go
Racer159: Refactored this

src/cmd/package.go
Racer159: slightly refactored. I agree it is odd, we could include the logic in the actual command definition too without this separate function.

src/internal/git/utils.go
```diff
@@ -28,6 +28,11 @@ type Credential struct {
 	Auth http.BasicAuth
 }
 
+var (
+	// For further explanation: https://regex101.com/r/zq64q4/1
```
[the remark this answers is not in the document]
Racer159: 💓

src/internal/git/clone.go
Racer159: No problem lol

src/internal/git/pull.go
```diff
@@ -1,90 +1,87 @@
...
-	// Gracefully handle no git creds on the system (like our CI/CD)
-	if gitCred.Auth.Username != "" {
-		cloneOptions.Auth = &gitCred.Auth
+	matches := gitURLRegex.FindStringSubmatch(gitURL)
+	if len(matches) == 0 {
+		// Unable to find a substring match for the regex
+		message.Fatalf("unable to get extract the repoName from the url %s", gitURL)
 	}
+	onlyFetchRef := matches[gitURLRegex.SubexpIndex("atRef")] != ""
+	gitURLNoRef := matches[gitURLRegex.SubexpIndex("proto")] + matches[gitURLRegex.SubexpIndex("hostPath")] + "/" +
```
jeff-mccoy, line 57: This line is kind of hard to read. We don't do a lot of it (and maybe we should), but go convention actually uses a lot of short variable names, which would help with readability here. Also you could use the function -> variable pattern used above here too or a variable function.  Something like 

```go
matches := gitURLRegex.FindStringSubmatch(gitURL)
idx := gitURLRegex.SubexpIndex

if len(matches) != 6 {
	// Unable to find a substring match for the regex
	message.Fatalf("unable to get extract the repoName from the url %s", gitURL)
}
onlyFetchRef := matches[idx("atRef")] != ""
gitURLNoRef := fmt.Sprintf("%s/%s%s", matches[idx("baseURL")], matches[idx("repo")], matches[idx("git")])
```

or

```go
matches := gitURLRegex.FindStringSubmatch(gitURL)
match := func(test string) string {
	return matches[gitURLRegex.SubexpIndex(test)]
}

if len(matches) != 6 {
	// Unable to find a substring match for the regex
	message.Fatalf("unable to get extract the repoName from the url %s", gitURL)
}
onlyFetchRef := match("atRef") != ""
gitURLNoRef := fmt.Sprintf("%s/%s%s", match("baseURL"), match("repo"), match("git"))
```

src/internal/git/utils.go
Racer159: done

src/internal/git/utils.go
Racer159: Just caught the error here since we might as well.

src/cmd/package.go
Racer159: refactored

src/internal/git/pull.go
Racer159: I'll use the top one since I think it is a good compromise (I don't always like following weird helper functions around)

jeff-mccoy: haha yeah totally fair point..and agree, was just trying to throw ideas against the wall

Racer159: pushed up a change

src/cmd/package.go
Racer159: Removed that separate function

PR #790, opened 2022-09-26
Fix the k3s Unit File on older Systemd

Racer159:
## Description

This fixes an issue with older versions of systemd that require [Unit] at the top of a Unit file.

## Related Issue

Fixes #N/A

## Type of change


jeff-mccoy (approved):
Oops 😬 Lgtm

PR #814, opened 2022-09-29
unpack tarball and remove by name

### LINKED ISSUE #742: Remove deployed packages by pathing to the tarball
Right now the `zarf package remove` command requires the argument to be the package name. It would be nice if we expand out the argument functionality so that it can take a path to a Zarf package.

Currently only `zarf package remove dos-games --confirm` works, this issue is proposing to also allow `zarf package remove build/zarf-package-dos-games-amd64.tar.zst`.

Noxsios:
## Description

Adds functionality to provide a tarball path in `zarf package remove <name | path>`.  Unpacks given tarball temporarily to grab `zarf.yaml > .Metadata.Name` as the name to use in `packager.Remove(<name>)`.

## Related Issue

Fixes #742 

## Type of change


Racer159 (changes requested):
can you add a test?

Racer159 (approved):
lgtm! thanks!

PR #816, opened 2022-09-29
Add noWait field to yaml for manifests and charts

### LINKED ISSUE #766: post-install hooks do not get ran until all chart resources are ready/healthy
### Environment
Device and OS: All
App version: 0.21.3
Kubernetes distro being used: k3d/k3s
Other:

### Steps to reproduce
1. Get a helm chart that utilizes a post-install hook for a resource
2. Modify helm chart that a pod doesn't become healthy until that post-install hook is ran
3. Test using just helm - post-install hook runs once all resources are created and all pods end up healthy
4. Create zarf package using modified helm chart
5. Deploy package

### Expected result

Post-install hook resources are created once all other resources have been created, pod that depends on post-install resource becomes healthy.

### Actual Result

Post-install hook never runs so the zarf package never succeeds.

### Severity/Priority

Medium - Blocking but has a workaround.

### Additional Context

This can be worked around by adding a dataInjection since zarf doesn't wait for resources to be ready before running post-install hook resources.

corang:
## Description

Add field to zarf.yaml to support not waiting for charts and manifests.

Fixes #766 

## Type of change

## Checklist before merging


src/test/e2e/28_wait_test.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func TestWait(t *testing.T) {
+	t.Log("E2E: Helm Wait")
+	e2e.setup(t)
+	defer e2e.teardown(t)
+
+	path := fmt.Sprintf("build/zarf-package-test-helm-wait-%s.tar.zst", e2e.arch)
+
+	zarfChannel := make(chan zarfCommandResult, 1)
```
corang, line 33: @jeff-mccoy @Racer159 Was there a better way to go about setting a timeout for a function, I guess its not too bad now that I wrote it but figuring this out took a really long time lol

Racer159: I think this looks fine from an implementation perspective (especially for what you are trying to do).  The only other big place in tests we have that watches for a timeout is here: https://github.com/defenseunicorns/zarf/blob/2a6bdbd75a4d3d118f18ae20b8c3aed2e355a8cf/src/test/e2e/22_git_and_flux_test.go#L103 but that doesn't always work smoothly (and is checking something after the Zarf command has finished and Flux takes over)

examples/helm-no-wait/README.md
```diff
@@ -0,0 +1,17 @@
+# Helm Alt Release Name
```
Racer159, line 1: The title should be changed here

src/test/e2e/28_wait_test.go
```diff
@@ -0,0 +1,69 @@
...
+	// stdOut, stdErr, err := e2e.execZarfCommand("package", "deploy", path, "--confirm")
+	require.NoError(t, err, stdOut, stdErr)
+
+	// // Verify multiple helm installs of different release names were deployed
+	// kubectlOut, _ := exec.Command("kubectl", "get", "pods", "-n=helm-releasename", "--no-headers").Output()
+	// assert.Contains(t, string(kubectlOut), "zarf-cool-name-podinfo")
+
+	stdOut, stdErr, err = e2e.execZarfCommand("package", "remove", "test-helm-wait", "--confirm")
+	require.NoError(t, err, stdOut, stdErr)
+}
```
Racer159, line 69: There should be a newline at the end of the file (see above)

Racer159 (changes requested):

Racer159 (approved):

PR #817, opened 2022-09-29
Remove repo1 CICD dependency due to unreliable uptime

jeff-mccoy:


Racer159:
While I agree on removing repo 1, I do think that we should replace this test case rather than completely removing it... I see it as testing two code paths:

1. Repos that are still using go-git but use a sub-org underneath the owning org (which might fail if we screw up the git URL regex)
2. Repos that have tags on a branch other than master (which Zarf assumes as default right now)

jeff-mccoy:
I may be missing something but the test wasn't deleted, just changed the repo we were testing?

Racer159:
It is a test case in the test though.  Some codepaths will no longer be covered without a repo like this.

Racer159 (approved):
fine @jeff-mccoy

PR #832, opened 2022-10-03
windows build

### LINKED ISSUE #426: Make Windows Release for Zarf

Noxsios:
## Related Issue

Fixes #426

## Type of change

## Checklist before merging


.github/workflows/build-rust-injector.yml
```diff
@@ -22,6 +22,7 @@ jobs:
         run: |
           curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path
           echo "$HOME/.cargo/bin" >> $GITHUB_PATH
+          sudo apt install build-essential -y
```
jeff-mccoy, line 25: This is already installed, was there a reason it needed to be installed again?  https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#installed-apt-packages

.github/workflows/build-rust-injector.yml
```diff
@@ -31,6 +32,7 @@ jobs:
       - name: "Build Rust Binary for x86_64"
         working-directory: src/injector/stage1
         run: |
+          rustup target add x86_64-unknown-linux-musl
```
jeff-mccoy, line 35: good catch, this was missed when moving from self-hosted runners

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,66 @@
...
+        with:
+          go-version: 1.19.x
+
+      - name: "Dependency: Install Scoop+Make+NodeJS"
+        shell: pwsh
+        run: |
+          Set-ExecutionPolicy RemoteSigned -scope CurrentUser
+          iex "& {$(irm get.scoop.sh)} -RunAsAdmin"
+          Join-Path (Resolve-Path ~).Path "scoop\shims" >> $Env:GITHUB_PATH
+          scoop install make@4.3 nodejs@16.17.1
```
jeff-mccoy, line 33: Was there a reason we needed to install these again? https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md

.github/workflows/build-rust-injector.yml
Noxsios: no, but its explicit so that others reading the pipeline and attempting to replicate on their local machine can more easily do so

.github/workflows/test-windows.yml
Noxsios: I didn't know that it had all the deps we needed!  Was just working off assumption is was a clean Windows box and we had to handle all deps.

jeff-mccoy (commented):
I merged in all my changes via PR #860, will leave for @JasonvanBrackel and @Racer159 to sanity check before we merge it.

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,56 @@
...
+      - "docs/**"
+
+# Abort prior jobs in the same workflow / PR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  validate:
+    runs-on: windows-latest
```
JasonvanBrackel, line 20: Can you change this and spell out the version of Windows explicitly. The window-latest label can change.

Makefile
```diff
@@ -68,9 +72,12 @@ build-cli-mac-intel: build-injector-registry-amd build-ui ## Build the Zarf CLI
 build-cli-mac-apple: build-injector-registry-arm build-ui ## Build the Zarf CLI for macOS on ARM
 	GOOS=darwin GOARCH=arm64 go build -ldflags="$(BUILD_ARGS)" -o build/zarf-mac-apple main.go
 
+build-cli-windows-amd: build-injector-registry-amd build-ui
```
JasonvanBrackel, line 75: We should be supporting Window and Windows on ARM if that's possible.

src/test/e2e/common.go
```diff
@@ -27,6 +28,10 @@ func GetCLIName() string {
 		} else {
 			binaryName = "zarf-mac-intel"
 		}
+	} else if runtime.GOOS == "windows" {
```
JasonvanBrackel, line 31: Just repeating the comment about supporting Windows on ARM here as well.

JasonvanBrackel (dismissed):

examples/component-scripts/zarf.yaml
```diff
@@ -4,21 +4,61 @@ metadata:
...
+
+  - name: prepare-windows
+    only:
+      localOS: windows
+    required: true
+    scripts:
+      # Note prepare scripts run on the local system before the package is built,
+      # so this component will fail to build on Windows
+      prepare:
+        - "touch test-prepare.txt"
```
Racer159, line 51: Why not use `New-Item` here as well?  All this does is create a text file.

src/test/e2e/common.go
Racer159: FWIW I agree on WoA.  We should support if we can.

.github/workflows/test-windows.yml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          path: |
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+
+      - name: "Build CLI"
+        shell: pwsh
+        run: make build-cli-windows-amd
```
Racer159, line 48: It would be good to document the dependencies needed for this to run on windows somewhere in the developer guide since `make` and many of the `make` targets will not work OOB if someone tries this for themselves (even on pwsh 7).

Racer159 (changes requested):

.github/workflows/test-windows.yml
jeff-mccoy: I think this one might be okay as it changes pretty infrequently in github image land and is how we do all other CI rn (with ubuntu-latest).

Makefile
jeff-mccoy: Agreed, I did add it already to goreleaser, should just need to update the makefile.

.github/workflows/test-windows.yml
jeff-mccoy: Yeah I'm a bit torn if we really need to actually support developing the zarf codebase on windows too. We have no demand signal for that right now.

Makefile
jeff-mccoy: want to do some more makefile cleanup so I'll open a new PR for those changes + the arm make target

jeff-mccoy (approved):
Added all final corrections based on @Racer159 & @JasonvanBrackel comments.

PR #837, opened 2022-10-03
Add local helm chart capability

### LINKED ISSUE #836: Add capability to add a helm chart from local files
Currently the only way to include a helm chart is from a git repo or a remote helm repo, this introduces problems for helm charts that either aren't published or are published in private settings. This especially introduces complexity when trying to create a zarf package from a helm chart in a project's CI/CD pipeline where they both exist in the same repo. Proposed yaml below:
```yaml
components:
  - name: localChartComponent
    charts:
      - name: localChart
        version: 0.1
        namespace: default
        localPath: /path/to/chart/folder/
```

corang:
Closes #836 

See issue for details


Racer159 (changes requested):

src/test/e2e/29_helm_local_chart_test.go
```diff
@@ -0,0 +1,23 @@
...
+	e2e.setup(t)
+	defer e2e.teardown(t)
+
+	path := fmt.Sprintf("build/zarf-package-test-helm-local-chart-%s.tar.zst", e2e.arch)
+
+	// Deploy the charts
+	stdOut, stdErr, err := e2e.execZarfCommand("package", "deploy", path, "--confirm")
+	require.NoError(t, err, stdOut, stdErr)
+
+	stdOut, stdErr, err = e2e.execZarfCommand("package", "remove", "test-helm-releasename", "--confirm")
```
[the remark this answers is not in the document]
corang: I'm not sure what you mean, the changes I made only affect package creation, this test is verifying that a package created with a local chart still deploys

examples/helm-local-chart/README.md
```diff
@@ -0,0 +1,17 @@
+# Helm Alt Release Name
```
brandtkeller, line 1: Is this the intended README content?

src/test/e2e/29_helm_local_chart_test.go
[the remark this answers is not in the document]
corang: Test now passes

src/internal/packager/create.go
```diff
@@ -114,10 +114,17 @@ func addComponent(tempPath tempPaths, component types.ZarfComponent) {
 		re := regexp.MustCompile(`\.git$`)
 		for _, chart := range component.Charts {
 			isGitURL := re.MatchString(chart.Url)
+			URLLen := len(chart.Url)
 			if isGitURL {
 				_ = helm.DownloadChartFromGit(chart, componentPath.charts)
-			} else {
+			} else if URLLen > 0 {
```
brandtkeller, line 120: We'll likely be arguing pedantics over which takes priority (`URLLen`  else using the `LocalPath`) - but is there anything that prevents a user from using both and/or documentation that establishes precedence?

examples/helm-local-chart/README.md
corang: Readme content fixed

src/internal/helm/repo.go
```diff
@@ -15,6 +15,25 @@ import (
 	"helm.sh/helm/v3/pkg/repo"
 )
 
+// CreateChartFromLocalFiles creates a chart archive from a path to a chart on the host os
+func CreateChartFromLocalFiles(chart types.ZarfChart, destination string) string {
```
brandtkeller, line 19: Is there any value in running a high-level [validation](https://github.com/helm/helm/blob/v3.10.0/pkg/chart/chart.go#L124) of the chart prior to package? Maybe this is covered by the processes below and failing to create an archive if not a compliant chart.

brandtkeller: My thinking here is to fail early and fail fast. Packaging should attempt (in my opinion) to validate some aspects of what is going to be deployed such that time isn't wasted transferring that artifact to the target environment and finding out that it is broken. I may be incorrect in assuming Zarf wants to do this though.

jeff-mccoy: Zarf has a validation package that runs during package create for this purpose but mainly focuses on zarf isms right now, I have 0 complaints about making that more robust to avoid finding out in air gap something is a problem.

src/internal/packager/create.go
corang: Input is now validated to only have one of url, localpath or gitpath

src/internal/helm/repo.go
corang: Added a validation check to all functions that load helm charts

examples/helm-local-chart/chart/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.16.0"
```
Racer159, line 24: new line at end of file

Racer159 (approved):

Racer159 (approved):

PR #838, opened 2022-10-04
Rke2 distro support

JasonvanBrackel:
## Description

Add the ability to support initializing an RKE2 cluster with Zarf

Fixes # (issue)

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

## Checklist before merging

- [ ] Tests have been added/updated as necessary 
- [ ] Documentation has been updated as necessary


packages/distros/rke2/common/rke2-canal.conf
```diff
@@ -0,0 +1,2 @@
+[keyfile]
+unmanaged-devices=interface-name:cali*;interface-name:flannel*
```
Racer159, line 2: no newline at the end of the file

packages/distros/rke2/zarf.yaml
```diff
@@ -0,0 +1,29 @@
...
+  name: "distro-rke2"
+  description: "Used to establish a new Zarf cluster"
+
+components:
+  # AMD-64 version of the rke2 stack
+  - name: rke2
+    import:
+      path: common
+      name: rke2
+    only:
```
Racer159, line 12: Since RKE2 only supports AMD64 should we add a dummy component with an after script that returns a non-zero exit code and prints a message that RKE2 does not support ARM (given we have a (somewhat decent) number of ARM users?

packages/distros/rke2/common/rke2.service
```diff
@@ -0,0 +1,28 @@
...
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+Type=notify
+EnvironmentFile=-/etc/default/%N
+EnvironmentFile=-/etc/sysconfig/%N
+KillMode=process
+Delegate=yes
+# Having non-zero Limit*s causes performance problems due to accounting overhead
```
Racer159, line 16: Limit*s?

Racer159 (changes requested):

packages/distros/rke2/zarf.yaml
jeff-mccoy: Might be better to use the architecture key in the package for now as it will prevent building for that architecture and warn as well. E.g. https://github.com/defenseunicorns/zarf/blob/c15f371e26e10bc768522a98c316fed6eb7d1266/packages/big-bang-core/zarf.yaml#L9

packages/distros/rke2/common/rke2-canal.conf
JasonvanBrackel: Addressed

packages/distros/rke2/zarf.yaml
JasonvanBrackel: I've added both, I wanted to keep the placeholder there, and Jeff's change works too

packages/distros/rke2/common/rke2.service
JasonvanBrackel: fixed

packages/distros/rke2/zarf.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - name: rke2
+    description: >
+      ****** READ THIS ARM 64 NOT SUPPORTED ********* 
+      rke2 is not currently supported on arm64 architecture, due to upstream dependencies. This functionality will be 
+      added when rke2 supports arm64 architecture.
+
+      If you would like to buid the amd64 version of this package on a arm64 based machine use the -a flag.
+
+      ex: `zarf package create . -a amd64`
+    import:
```
Racer159, line 41: do we need the import and files?  I think we just need the scripts part no?

Racer159 (changes requested):

packages/distros/rke2/zarf.yaml
JasonvanBrackel: @jeff-mccoy would love your feedback here.  Left this commented out as a placeholder, but fine if it all goes away.

jeff-mccoy: You shouldn't have to re-import something already imported (or declared) by a component your importing.  However, I think for simplicities sake it might be worth forgoing the composable component here since there is no arm support anyway. The package-level `architecture` declaration at the top should be enough.

packages/distros/rke2/zarf.yaml
```diff
@@ -0,0 +1,72 @@
...
+        - common/empty-file
+    scripts:
+      retry: false
+      after:
+        # Force the return of a non-zero exit code.
+        - |
+          'echo "rke2 is not currently supported on arm64 architecture, due to upstream dependencies. This functionality will be 
+           added when rke2 supports arm64 architecture.'
+        - 'exit(1)'
+    # files:
```
jeff-mccoy, line 58: We will want to remove commented yaml before merging as well.

jeff-mccoy (commented):
The way we did K3s (not using the K3s installer) was a specific decision to work around some quirks with the k3s installer + SELinux at the time. I'd love to just use the provided upstream installers if possible here and probably look at the K3s one again as well.

JasonvanBrackel:
@jeff-mccoy working on it today, and tomorrow if I run out of time.  Will have the requested changes in no later than tomorrow. Hope to have docs written by EOD Thursday.

jeff-mccoy:
No worries, just wanted to check in.

JasonvanBrackel:
> The way we did K3s (not using the K3s installer) was a specific decision to work around some quirks with the k3s installer + SELinux at the time. I'd love to just use the provided upstream installers if possible here and probably look at the K3s one again as well.

Ok!  I'll look at more "official path" and update this accordingly.

jeff-mccoy:
Closing as stale.

PR #844, opened 2022-10-05
Add a description field for zarf variable / constant declarations

jeff-mccoy:


src/config/variables.go
```diff
@@ -105,17 +108,20 @@ func InjectImportedConstant(importedConstant types.ZarfPackageConstant) {
...
 	pterm.Println()
 
+	if variable.Description != "" {
+		message.Note(variable.Description)
+	}
+
 	prompt := &survey.Input{
-		Message: "Please provide a value for '" + varName + "'",
-		Default: varDefault,
+		Message: "Please provide a value for '" + variable.Name + "'",
```
Racer159, line 120: I know this is just pretty colors, but something like this is more pleasing/theme fitting than `message.Note`:

```
description := ""
	if variable.Description != "" {
		description = pterm.FgMagenta.Sprintf("\n - %s", variable.Description)
	}

	prompt := &survey.Input{
		Message: fmt.Sprintf("Please provide a value for '%s'%s", variable.Name, description),
		Default: variable.Default,
	}
```
At the very least we should change this to `fmt.Sprintf()` though

Racer159 (changes requested):

src/config/variables.go
jeff-mccoy: I'm totally fine changing it,I try to avoid direct color calls vs function wrappers but I'll look at it.

Racer159 (approved):
Approving for now, but I don't love the description above the prompt either (feels disconnected to me)...  Tagging @Madeline-UX in case she has opinions we could address in a future PR (since it would be non-breaking visual changes).  Would be nice to have a guide on the use of color, prompting and displaying info even for the CLI use case.

Madeline-UX:
> Approving for now, but I don't love the description above the prompt either (feels disconnected to me)... Tagging @Madeline-UX in case she has opinions we could address in a future PR (since it would be non-breaking visual changes). Would be nice to have a guide on the use of color, prompting and displaying info even for the CLI use case.

@Racer159 Can you provide a screen capture of what you are describing here and how it looks in the CLI?

jeff-mccoy:
Yeah I'm not sure what he's referring to, this is consistent color/placement with all other prompts / descriptions fields used in zarf cli today. It would also be very difficult to actually code the text below the input field with the current library.

Racer159:
@Madeline-UX @jeff-mccoy I may be thinking about this weirdly, but it just feels like the description is hanging out above the prompt.  Part of that could also be the way the prompt/description is worded too, but usually I am used to having a title that goes before the description and prompt to ground it more.

This is how it is:
194244628-e932c363-b34e-4928-b16f-f969e3cbdcc5

This is closer to how I (subjectively) feel it should be:
image


Not a big deal hence my approval, but something that felt weird so I mentioned it.

jeff-mccoy:
I think you might be getting wrapped around the wording.  We use this same format everywhere but the description field is not a question, that one just happened to be one in that example as a user can make it whatever they want. See component deployment prompts for how this works elsewhere.

Racer159:
I think it is the bolding hierarchy too

PR #845, opened 2022-10-05
Fix agent mutation of new registry

Racer159:
## Description

This is a hotfix related to the image SHASUMs introduced recently to not mutate the zarf registry on a zarf reinit.

## Related Issue

Fixes # N/A

## Type of change

## Checklist before merging


jeff-mccoy:
Is there a reason we wouldn't just add the correct agent ignore label to the pod all the time vs hardcode a special exemption that would still hit the webhook?

jeff-mccoy:
Actually thing we need to use ignore/skip "zarf.dev/agent=ignore" here instead.  Reason for this is if the agent and registry both died due to node failures, you'd be stuck in a pending state with "patched" waiting for the agent whereas ignore wouldn't depend on the agent existing at all.

jeff-mccoy (changes requested):
I think we need to use ignore/skip `zarf.dev/agent=ignore` here instead. Reason for this is if the agent and registry both died due to node failures, you'd be stuck in a pending state with "patched" waiting for the agent whereas ignore wouldn't depend on the agent existing at all.

Racer159:
> 

Changed

jeff-mccoy (approved):

PR #847, opened 2022-10-05
Fixup Makefile help and Speedup builds

Racer159:
## Description

This adds more help information to our Makefile and doesn't rebuild the UI unless explicitly asked for. Mostly a suggested improvement

## Related Issue

Fixes # N/A

## Type of change

## Checklist before merging


jeff-mccoy (changes requested):
I love this idea--but I think we need to tweak the git diff a bit, in testing this way won't capture staged files that are changed and I'd rather err on the side of rebuild vs not.

### Current method:
Screen Shot 2022-10-06 at 9 39 18 PM

### Using `--staged/--cached` (only gets staged):
Screen Shot 2022-10-06 at 9 41 34 PM


### Using `git diff HEAD` gets both staged & un-staged changes:
Screen Shot 2022-10-06 at 9 42 48 PM

Racer159:
> I love this idea--but I think we need to tweak the git diff a bit, in testing this way won't capture staged files that are changed and I'd rather err on the side of rebuild vs not.
> ### Current method:
> 
> Screen Shot 2022-10-06 at 9 39 18 PM
> ### Using `--staged/--cached` (only gets staged):
> 
> Screen Shot 2022-10-06 at 9 41 34 PM
> ### Using `git diff HEAD` gets both staged & un-staged changes:
> 
> Screen Shot 2022-10-06 at 9 42 48 PM

Good catch! Changed to `git diff HEAD`

jeff-mccoy (approved):
Love this feature!

PR #848, opened 2022-10-06
Support automatic multi-part packages based on a MaxPackageSize

### LINKED ISSUE #843: Splitting zarf packages into multiple files
Given that the military operates almost exclusively off of CDs and DVDs for data transfer between machines (especially to an air gapped machine) it would be a valuable feature of zarf to be able to split packages into multiple files to be able to fit on aforementioned storage mediums.

Either/both a flag to `zarf package create` to specify splitting size or/and a separate utility command `zarf tools split` & `zarf tools combine` to manage splitting and combining packages.

**Additional context**
There is already a pattern for this relating to injector configmap creation

jeff-mccoy:
This PR adds the `zarf package create --max-package-size=N`, where N is the maximum number of megabytes a part can be, with the default being 0. Above the size limit, a Zarf package is automatically broken into smaller chunks up to the limit. These chunks, starting with index 000, are then reassembled into a single package on `zarf package deploy` before running the deployment.  This means that if a user attempts to deploy a multi-part Zarf package but does not confirm the deployment, the multipart archive will now be just a single archive still. This was needed to avoid putting 3 copies of the data on the disk (the original multi-part, the assembled archive, and the temp folder where the archive is extracted).

## Related Issue

Fixes #843 

## Type of change

## Checklist before merging


jeff-mccoy:
Use `zarf package create --max-package-size=1` and then running `zarf package deploy` without a package name detects the first element of the mulit-part archive, but doesn't list the others. 
Screenshot 2022-11-15 at 1 02 56 PM

src/internal/packager/deploy.go
```diff
@@ -44,6 +45,47 @@ func Deploy() {
...
+			}
+			defer f.Close()
+
+			// Add the file contents to the package
+			if _, err = io.Copy(pkgFile, f); err != nil {
+				spinner.Fatalf(err, "Unable to copy partial package %s into package file", file)
+			}
+		}
+
+		// Remove the partial packages to reduce disk space before extracting
```
YrrepNoj, line 80: I was very much not expecting this when I was playing around with the branch (and should probably be documented somewhere / somewhere more noticeable)

src/cmd/package.go
```diff
@@ -202,11 +204,13 @@ func bindCreateFlags() {
 	v.SetDefault(V_PKG_CREATE_OUTPUT_DIR, "")
 	v.SetDefault(V_PKG_CREATE_SKIP_SBOM, false)
 	v.SetDefault(V_PKG_CREATE_INSECURE, false)
+	v.SetDefault(V_PKG_CREATE_MAX_PACKAGE_SIZE, 1024)
```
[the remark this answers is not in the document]
Racer159: As discussed I think pegging this at the size of a DVD could make sense

src/internal/packager/create.go
```diff
@@ -108,6 +108,34 @@ func Create(baseDir string) {
...
+			message.Fatal(err, "Unable to split the package archive into chunks")
+		}
+
+		message.Infof("Package split into %d files, original sha256sum is %s", len(chunks), sha256sum)
+		_ = os.RemoveAll(packageName)
+
+		for idx, chunk := range chunks {
+			path := strings.Replace(packageName, ".tar", fmt.Sprintf("-%03d.partial-tar", idx), 1)
+			if err := os.WriteFile(path, chunk, 0644); err != nil {
+				message.Fatalf(err, "Unable to write the file %s", path)
```
Racer159, line 135: We should add a shasum file as discussed

src/internal/packager/deploy.go
```diff
@@ -44,6 +45,47 @@ func Deploy() {
 		spinner.Fatalf(nil, "Unable to find the package on the local system, expected package at %s", config.DeployOptions.PackagePath)
 	}
 
+	// If packagePath has partial in the name, we need to combine the partials into a single package
+	if strings.Contains(config.DeployOptions.PackagePath, "-000.partial-tar") {
```
Racer159, line 49: I am more of a fan of following the convention to chain filenames here, like `.tar.gz` and doing something like `.tar.part000` since underneath the process of splitting it it is still a .tar or .zst file.

src/types/runtime.go
```diff
@@ -36,6 +36,7 @@ type ZarfCreateOptions struct {
 	Insecure        bool              `json:"insecure" jsonschema:"description=Disable the need for shasum validations when pulling down files from the internet"`
 	OutputDirectory string            `json:"outputDirectory" jsonschema:"description=Location where the finalized Zarf package will be placed"`
 	SetVariables    map[string]string `json:"setVariables" jsonschema:"description=Key-Value map of variable names and their corresponding values that will be used to template against the Zarf package being used"`
+	MaxPackageSize  int               `json:"maxPackageSize" jsonschema:"description=Size of chunks to use when splitting a zarf package into multiple files"`
```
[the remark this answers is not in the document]
Racer159: That would also help since this could be interpreted as "number of chunks" when read

src/internal/packager/deploy.go
Racer159: (that also prevents other tooling from thinking it can do something with just that one file alone)

Racer159 (changes requested):

src/cmd/package.go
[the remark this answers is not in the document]
jeff-mccoy: we'd need to leave some room for file formatting too, right? Would 4.5 GBs be sufficient?

src/internal/packager/deploy.go
jeff-mccoy: As discussed during office hrs, this will be verified by shasum prior to removal and is needed to not greatly increase disk pressure.

src/internal/packager/deploy.go
jeff-mccoy: that's a really good point, I'm down to change that

src/cmd/tools.go
```diff
@@ -156,11 +156,14 @@ func init() {
 	craneLogin := craneCmd.NewCmdAuthLogin()
 	craneLogin.Example = ""
 
+	craneCatalog := craneCmd.NewCmdCatalog(&cranePlatformOptions)
+	craneCatalog.Example = ""
```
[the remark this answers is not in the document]
jeff-mccoy: I kind of want to remove some of that verbiage too with the transition to the language file. The better thing is probably to look at building out a new feature in zarf, something like `zarf artifacts images list` or similar that automatically connect/authenticates and does the catalog equivalent for users. Including crane this way is really just a power user tool for advanced scenarios.

jeff-mccoy:
This should be ready for review again

blancharda:
Adding the ability to optionally split the artifact into chunks is great! That will make some niche transport cases much simpler.

Should splitting be the default behavior though?
IMO the defaults should support the simplest scenario, with options to provide more advanced usage.

YrrepNoj:
> Should splitting be the default behavior though? IMO the defaults should support the simplest scenario, with options to provide more advanced usage.

I find myself agreeing with this still (I made a similar comment earlier in the PR). 

I'm not sure the best way to word my thoughts but I feel like if I built a 500Mb package and it output with a single package, I would then expect that to be the 'default' behavior and would expect a single package when I built a 9Gb package. 

Being automatically opted-in to package splitting once a package reaches a specific size doesn't make sense to me. Imo flags should be used to change behavior but with this it seems like I have to use a flag to keep behavior consistent.

RothAndrew:
I agree as well

Madeline-UX:
Catching up on this - So it looks like there is an upper limit on package size. In order to fix that two options are being proposed

1. Automatically split package if it exceeds size limit (trigger by size) 
2. Default all packages to have the option to divide into smaller bits

May I recommend not assuming anything but also not making users have to do another step for small packages
When a package creator creates package < 4.3 GB 
Then Zarf creates a package (no prompt)

When a package creator creates package > 4.3 GB
Then Zarf CLI prompts with a warning message 
"Zarf cannot create a package of this size, would you like Zarf to automatically break up into smaller packages?"  Y/n
Yes - Zarf does functionality
No - user goes back and rearchitect package themselves (package is not created)

A few questions 
- will these be different Zarfballs? (ex. would the user have to move multiple files for the same package?

RothAndrew:
Zarf packages can be any size. The need for being able to split packages has to do with potential limits on storage media like USB drives or DVDs

Madeline-UX:
So I have another ideas 

3. Zarf tells user what the size of the package is upon creation > then there is a separate flag for splitting packages that allow user to set the upper limit to whatever they want. Does not interfere with or change the current package create flow. Just adds functionality

blancharda:
> 3\. Zarf tells user what the size of the package is upon creation
This feels like an extra step for something that could just be an optional flag on `zarf package create`..

My preference is no limit by default (meaning zarf will produce a single artifact by default), but a flag allowing the user to chunk the output if they want/need to.

jeff-mccoy:
Some random thoughts:

- Zarf at its core is an airgap tool. I don't know of any airgap system anywhere (including data diodes) that doesn't have a max file size limit. 
- Once [YOLO Mode](https://github.com/defenseunicorns/zarf/pull/1105) is merged in, there will be a very viable path for cloud/online systems to still uses Zarf but not bring any images or other large binaries so they would never see this split. 
- Changing the default behavior back forever on your system takes creating a file with 
  ```toml
  [package.create]
  max_package_size=0
  ```
- The process of splitting/merging files is automatic and the only change for users is the fact there is more than one file (if they didn't change the default behavior and had a package with > 4.3 GBs of compressed data
- This PR sets the groundwork for being able to package inspect multipart packages without having to decompress the archive first (there's a JSON file in the first file that stores metadata about the multipart package)
- The only potential negative I see is for those connected systems with no concern for file transfer size and are also very large, i.e. DI2ME. For that, I admit this would be an extra step and is part of the contention I believe.

Madeline-UX:
Prompt presented in the package create flow? Are there size limits to pacakge size Y/n?
That way user opts into additional functionality but is reminded to think of this limit before hand.

YrrepNoj:
I think my biggest concern is that it would now require a flag to enable consistent behavior.

If I'm iterating on a package that is is hovering around the 4.3GB size and add an extra component,  I now have to add a flag to get the same output as I would have before adding that component. That (self defined) arbitrary inconsistency is what concerns me.

Just as a person can set their config to not worry about package sizes. People who are building packages for air gap environments, and know what the limitations of their storage medium is, could set and forget their toml to
```toml
[package.create]
max_package_size=4300
```

RothAndrew:
Totally agree with Jon on this one

Racer159:
Reading through this, for my 2 pennies, I agree with no limits by default.  This will affect some of the customers that I know of and they will be perplexed by this limit as they use HDDs to transfer.    I think we can make the dvd case simpler for folks in the future by introducing keywords like dvd = 4300, bd = 25000, and bdxl = 100000 so people don't need to know those limits.

Racer159:
Also I think a tools command to split a package into chunks will be a very useful thing (especially if the package is coming from another place like our init package - right now we fit on a CD but if someone wanted to use something like zip disks they would need to split that up and forcing them to build a custom package sounds annoying)

src/cmd/package.go
```diff
@@ -234,12 +239,16 @@ func bindCreateFlags() {
 	v.SetDefault(V_PKG_CREATE_SKIP_SBOM, false)
 	v.SetDefault(V_PKG_CREATE_INSECURE, false)
 
+	// Set the max default size to 4.25 Gigabytes (4,563,402,752 bytes, DVD-5 is 4,700,000,000 bytes)
+	v.SetDefault(V_PKG_CREATE_MAX_PACKAGE_SIZE, 1024*4.25)
```
Racer159, line 243:
```suggestion
	v.SetDefault(V_PKG_CREATE_MAX_PACKAGE_SIZE, 0)
```

Racer159 (changes requested):

blancharda:
> This will affect some of the customers that I know of and they will be perplexed by this limit as they use HDDs to transfer.

Absolutely.
I understand I am only one user, but this will impact around ~50% of the packages I work with.

Further, it makes unnecessary assumptions about how I'm transferring data. I've carried plenty of encrypted drives (at least one direction) into air-gapped ATO'd systems. It's not always DVDs and diodes..

jeff-mccoy:
I don't see this change as negatively effecting users as it doesn't take anything away from someone transferring by HDD, for example. It would be the difference in a script of copying an exact name or name with a *. 

While I disagree with the PR comments above, I will make the changes based on the desire to not introduce breaking changes anymore. For consistency, I expect we'll apply that same criteria to future PRs.

YrrepNoj (dismissed):

jeff-mccoy:
Fall back to 0 on error (ctrl + C)
Screenshot 2022-12-09 at 10 27 12 PM

src/pkg/packager/interactive.go
```diff
@@ -44,7 +45,25 @@ func (p *Packager) confirmAction(userMessage string, sbomViewFiles []string) (co
 
 	// Prompt the user for confirmation, on abort return false
 	if err := survey.AskOne(prompt, &confirm); err != nil {
-		return false
+		confirm = true
+	}
+
+	// On create in interactive mode, prompt for max package size
+	if userMessage == "Create" {
+		value, _ := p.promptVariable(types.ZarfPackageVariable{
```
Racer159, line 53: We should not dogsled the error here.  An interrupt (Ctrl-C) should stop Zarf as soon as possible.

examples/multi-part/zarf.yaml
```diff
@@ -0,0 +1,13 @@
+kind: ZarfPackageConfig
+metadata:
+  name: multi-part
+  description: Example to demonstrate how to create multi-part zarf packages. By default, packages will not be split into multiple parts unless the package size is larger than 1 Gigabyte (1,024 Megabytes). This package contains a single 5 Megabyte file. To produce a multi-part package, use the command "zarf package create --max-package-size=1"
```
Racer159, line 4: The default has changed.

src/pkg/packager/interactive.go
```diff
@@ -44,7 +45,25 @@ func (p *Packager) confirmAction(userMessage string, sbomViewFiles []string) (co
...
+			Name:        "Maximum Package Size",
+			Description: "Specify a maximum file size for this package in Megabytes (MB). Above this size, the package will be split into multiple files. 0 will disable this feature.",
```
Racer159, line 55: Technically... This is Mebibytes (MiB) (1024*1024) not strictly Megabytes (MB) (1000*1000).  I know some of our personas might not know what that means, but we could hint at it by saying `Megabytes (MiB)` instead.

https://simple.wikipedia.org/wiki/Mebibyte
https://simple.wikipedia.org/wiki/Megabyte

src/pkg/packager/create.go
```diff
@@ -126,6 +129,45 @@ func (p *Packager) Create(baseDir string) error {
...
+			Sha256Sum: sha256sum,
+		})
+		if err != nil {
+			return fmt.Errorf("unable to marshal the partial package data: %w", err)
+		}
+
+		// Prepend the json data to the first chunk
+		chunks = append([][]byte{jsonData}, chunks...)
+
+		for idx, chunk := range chunks {
```
Racer159, line 163: Can we check  if the number of splits is > 1K?

Racer159 (changes requested):

jeff-mccoy:
all requested changes made

src/pkg/packager/inspect.go
```diff
@@ -21,6 +21,11 @@ func (p *Packager) Inspect(packageName string, includeSBOM bool, outputSBOM stri
 		return fmt.Errorf("invalid package name: %s", packageName)
 	}
 
+	// If packagePath has partial in the name, we need to combine the partials into a single package
+	if err := p.handleIfPartialPkg(); err != nil {
```
Racer159, line 25: Because DeployOpts is not set, this logs `No partial package detected`

And because it doesn't touch package name, results in an error of: `Extract error: format unrecognized by filename: zarf-package-*.tar.zst.part000` from the packager (this error is dogsledded though and should likely be returned to the inspect command for processing as well)

Racer159 (changes requested):
AGAIN!

src/pkg/packager/common.go
```diff
@@ -161,3 +166,138 @@ func getRequestedComponentList(requestedComponents string) []string {
...
+	// Load the config from the extracted archive zarf.yaml
+	spinner.Updatef("Loading the zarf package config")
+	configPath := filepath.Join(p.tmp.Base, config.ZarfYAML)
+	if err := p.readYaml(configPath, true); err != nil {
+		return fmt.Errorf("unable to read the zarf.yaml in %s: %w", p.tmp.Base, err)
+	}
+
+	// If SBOM files exist, temporary place them in the deploy directory
+	p.cfg.SBOMViewFiles, _ = filepath.Glob(filepath.Join(p.tmp.Sboms, "sbom-viewer-*"))
+	if err := sbom.WriteSBOMFiles(p.cfg.SBOMViewFiles); err != nil {
```
Racer159, line 203: :laughing: this will leave behind a zarf-sbom directory on package inspect.

src/pkg/packager/common.go
```diff
@@ -161,3 +166,138 @@ func getRequestedComponentList(requestedComponents string) []string {
...
+	}
+
+	// Load the config from the extracted archive zarf.yaml
+	spinner.Updatef("Loading the zarf package config")
+	configPath := filepath.Join(p.tmp.Base, config.ZarfYAML)
+	if err := p.readYaml(configPath, true); err != nil {
+		return fmt.Errorf("unable to read the zarf.yaml in %s: %w", p.tmp.Base, err)
+	}
+
+	// If SBOM files exist, temporary place them in the deploy directory
```
Racer159, line 201: temporary -> temporarily

src/pkg/packager/common.go
Racer159: Actually these are just no longer "temporary" (on deploy or inspect)

Racer159 (changes requested):

src/pkg/packager/interactive.go
```diff
@@ -44,7 +45,28 @@ func (p *Packager) confirmAction(userMessage string, sbomViewFiles []string) (co
...
-		return false
+		confirm = true
+	}
+
+	// On create in interactive mode, prompt for max package size
+	if userMessage == "Create" {
+		value, err := p.promptVariable(types.ZarfPackageVariable{
+			Name:        "Maximum Package Size",
+			Description: "Specify a maximum file size for this package in Megabytes (MB). Above this size, the package will be split into multiple files. 0 will disable this feature.",
+			Default:     "0",
```
Racer159, line 56: I'm sorry @jeff-mccoy ... I accidentally found this by running `zarf package create packages/big-bang-core -m 50` I expected the -m to be carried through (like we do for `components`)...

image
image

Can we carry forward the default through to here (either from the config toml or CLI?)

Racer159 (changes requested):

Racer159 (approved):
LGTM!

Madeline-UX:
Summary of user behavior
-- Confirm mode: (default set to no split) 
Interactive mode: Prompted to set the max file size (0 = no split)

Users can split any package after it is created through flag

PR #857, opened 2022-10-06
adds barebones terraform example

UncleGedd:
## Description

Adds a barebones example for deploying Terraform

## Related Issue

Fixes #856 

## Type of change
Documentation

## Checklist before merging

- [X] Documentation has been updated as necessary (add the `needs-docs` label)

PR #862, opened 2022-10-07
Update README.md

Madeline-UX:
## Description

I propose updating Zarf tagline to Kubernetes for AirGap per conversation with @JasonvanBrackel 

DevSecOps is more broad and really when we talk about Zarf we always reference the value in terms of bringing the power of Kubernetes to the AirGap


jeff-mccoy:
Not sure I agree with this one, think we need to have a broader conversation on this as Zarf already today does more than K8s, the recent terraform example we added, #857, is one that comes to mind.

Racer159:
> 

Throwing my 2 pennies in here as well, I do like keeping DevSecOps in there since Zarf enables that process over the airgap, and IMO Kubernetes is mostly an implementation detail in many use cases (if all we cared about was k8s, then why do SBOM as an example?)

jeff-mccoy:
Please open an issue to discuss this change and we can do a PR if a change is needed at that time. Make sure to keep in mind there will likely be other docs edits as well as the social media preview to update if we decide to change back to Kubernetes.

PR #876, opened 2022-10-10
Switch to distribution/distribution vs regex hacking for image url pa…

jeff-mccoy:
This PR removes the Frankenstein regex we were using for parsing docker image references and instead uses the same code that the docker registry itself uses for parsing references to handle various edges cases.  SHA1 hashes are also replaced with CRC32 for images and git URLs.  

The image replacement also should support image digests, but more work will need to be done on the usage of crane for image push/pull as digests are not yet preserved.


src/config/config.go
```diff
@@ -54,6 +54,8 @@ const (
 	ZarfInClusterContainerRegistryNodePort = 31999
 
 	ZarfInClusterGitServiceURL = "http://zarf-gitea-http.zarf.svc.cluster.local:3000"
+
+	ZarfSeedImage = "registry:2.8.1"
```
YrrepNoj, line 58: This feels like a weird thing to code into the binary. I see that you're doing this since you don't want to have the `seed` section for every package when it is only needed for initialization packages and I don't have any immediate suggestions but this feels off since it forces us to build a new Zarf binary if we want to update the registry component with a new container version.

YrrepNoj (commented):
I have some minor concerns with how we define the seed image tag. I don't have any real issues if this gets merged in as is, but I wanted to have a quick conversation about it first.

src/config/config.go
jeff-mccoy: We could change it to an env variable--but they've just started any useful updates again, it remained static for a couple years.  This will update probably every 2-3 minor bumps of zarf (every 3-4 months)

YrrepNoj: Maybe one day if this still bothers me in the future I can look into some form of env variable (or specialized Zarf Package Variable). But I'm fine with the way it is for now.

YrrepNoj (approved):

PR #885, opened 2022-10-11
Version Control Init Pkg & Use zarf-cache

### LINKED ISSUE #839: Issue installing Zarf Init - permission denied
### Environment
Device and OS: Debian 11
App version: 0.21.3
Kubernetes distro being used: Kind
Other:

### Steps to reproduce
When installing Zarf to /usr/local/bin on a linux box (Debian 11), and running as a user. It will try to write the zarf init pkg to /usr/local/bin/ and fail. with a permission denied.

### Expected result

Zarf place init pkg into the directory (I think)

### Actual Result

Fails with permission denied

### Severity/Priority

Low - CLI user was able to troubleshoot and fix without referencing docs

### Additional Context

Collected from user engagement

### LINKED ISSUE #889: Zarf exits with a zero exit code even though a package deployment failed.
### Environment
Device and OS: macbook Pro Intel chip / macOS Monterey v 12.6 
App version: v0.21.3
Other:

### Steps to reproduce
1. Build an application zarf package with missing deployment spec (i.e. no selector info or template.metadata)
2. Attempt to deploy package and receive a WARNING Unable to complete helm chart install/upgrade followed by "Zarf deployment complete" message

### Expected result
zarf exits with a non-zero exit code

### Actual Result
zarf exit with zero exit code

Screen Shot 2022-10-12 at 8 43 51 AM

jeff-mccoy:
## Related Issue
Fixes #839 
Fixes #889 
Related to #663 

## Type of change

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/spf13/cobra v1.6.0
 	github.com/spf13/viper v1.13.0
 	github.com/stretchr/testify v1.8.0
-	golang.org/x/crypto v0.0.0-20221010152910-d6f0a8c073c2
+	golang.org/x/crypto v0.0.0-20221012134737-56aed061732a
 	gopkg.in/yaml.v2 v2.4.0
 	helm.sh/helm/v3 v3.10.0
 	k8s.io/api v0.25.2
--- a/go.sum
+++ b/go.sum
@@ -2451,6 +2451,8 @@ golang.org/x/crypto v0.0.0-20221005025214-4161e89ecf1b h1:huxqepDufQpLLIRXiVkTvn
 golang.org/x/crypto v0.0.0-20221005025214-4161e89ecf1b/go.mod h1:IxCIyHEi3zRg3s0A5j5BB6A9Jmi73HwBIUl50j+osU4=
 golang.org/x/crypto v0.0.0-20221010152910-d6f0a8c073c2 h1:x8vtB3zMecnlqZIwJNUUpwYKYSqCz5jXbiyv0ZJJZeI=
 golang.org/x/crypto v0.0.0-20221010152910-d6f0a8c073c2/go.mod h1:IxCIyHEi3zRg3s0A5j5BB6A9Jmi73HwBIUl50j+osU4=
+golang.org/x/crypto v0.0.0-20221012134737-56aed061732a h1:NmSIgad6KjE6VvHciPZuNRTKxGhlPfD6OA87W/PLkqg=
+golang.org/x/crypto v0.0.0-20221012134737-56aed061732a/go.mod h1:IxCIyHEi3zRg3s0A5j5BB6A9Jmi73HwBIUl50j+osU4=
 golang.org/x/exp v0.0.0-20190121172915-509febef88a4/go.mod h1:CJ0aWSM057203Lf6IL+f9T1iT9GByDxfZKAQTCR3kQA=
 golang.org/x/exp v0.0.0-20190306152737-a1d7652674e8/go.mod h1:CJ0aWSM057203Lf6IL+f9T1iT9GByDxfZKAQTCR3kQA=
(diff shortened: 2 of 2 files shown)


jeff-mccoy:
Note most of the changes are in the `/docs/` because of the zarf-cache flag becoming a global flag.

jeff-mccoy:
init from the same directory
Screen Shot 2022-10-11 at 3 14 46 PM

jeff-mccoy:
init with the init package in the same dir as the binary
Screen Shot 2022-10-11 at 3 15 05 PM

jeff-mccoy:
init with the init package in the `zarf-cache` directory
Screen Shot 2022-10-11 at 5 38 36 PM

jeff-mccoy:
init with no local init package found
Screen Shot 2022-10-11 at 5 39 09 PM

src/cmd/initialize.go
```diff
@@ -52,73 +52,81 @@ var initCmd = &cobra.Command{
...
 		// Continue running package deploy for all components like any other package
-		initPackageName := fmt.Sprintf("zarf-init-%s.tar.zst", config.GetArch())
+		initPackageName := config.GetInitPackageName()
 		config.DeployOptions.PackagePath = initPackageName
 
 		// Try to use an init-package in the executable directory if none exist in current working directory
 		if utils.InvalidPath(config.DeployOptions.PackagePath) {
+			// Get the path to the executable
 			executablePath, err := utils.GetFinalExecutablePath()
-			if err != nil {
```
YrrepNoj, line 61: I know this is to combine the check below but I would still recommend checking errors sooner rather than later. It would be pretty simple for someone to come modify this code in the future and add some logic in-between these two sections not realizing that the error checking was being delayed.

jeff-mccoy: There are no error generators between it's declaration and usage, but you make a good point.  I'll go ahead and log the error too.

src/cmd/initialize.go
```diff
@@ -52,73 +52,84 @@ var initCmd = &cobra.Command{
...
+		}
+	} else {
+		// If CLI version exists then get the latest init package for the matching major, minor and patch
+		initPackageVersion = fmt.Sprintf("v%d.%d.%d", version.Major(), version.Minor(), version.Patch())
+	}
 
-				// Give the user the choice to download the init-package and note that this does require an internet connection
-				message.Question(fmt.Sprintf("It seems the init package could not be found locally, but can be downloaded from %s", url))
+	var confirmDownload bool
+	url := fmt.Sprintf("https://github.com/%s/releases/download/%s/%s", config.GithubProject, initPackageVersion, initPackageName)
```
Racer159, line 107: Now that the initPackageName contains the initPackageVersion this fails on anything other than a pure tagged release.  Previously (see line 103) it would chop off the -g0000000 commit piece when downloading which would still be nice to have for testing purposes and custom builds that do not change the agent.

Racer159: image

Racer159 (changes requested):

src/cmd/initialize.go
jeff-mccoy: But why would you want to d/l an old version of the init package. If you did that right now without this change it would break.  I'm not sure it's a valid use-case for a dev to download mismatched versions like that, except for upgrade testing, which should still be version matched anyway?

jeff-mccoy: ^ it would break because master has breaking changes since the last release

Racer159: It would break currently, but was a feature in the past.  Either that stripping code should be removed and it should be a non-feature or it should be implemented fully.

Racer159: There is also code to get the latest release tag that should probably be dropped. The latest release tag will be out of sync with master in this case too.

Racer159: The likely *best* course of action may actually be to strip the version and compare it to the CLI version.  If mismatched, show a warning that you are in dev, and should manually create an init package.

PR #886, opened 2022-10-12
Update the AGENT_IMAGE URL in the Makefile

YrrepNoj:
Update the agent_image so that we use an agent that calculates checksums of the images instead of shasums in local dev mode.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ else
 	endif
 endif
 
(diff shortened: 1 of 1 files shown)


jeff-mccoy:
Could you link to the build that produced this image so the PR has a record of it?

YrrepNoj:
> Could you link to the build that produced this image so the PR has a record of it?

Agent Image was built as part of this pipeline workflow: https://github.com/defenseunicorns/zarf/actions/runs/3226469820/jobs/5280023891

jeff-mccoy (approved):

Issue #897, opened 2022-10-12
OSSF Scorecard Recommendations


jeff-mccoy:
There are some pretty easy actions we can do to update our https://securityscorecards.dev/.  

- [ ] Add OpenSSF Badge, [link](https://github.com/ossf/scorecard/blob/7cd6406aef0b80a819402e631919293d5eb6adcf/docs/checks.md#cii-best-practices)
- [ ] Only use pinned dependencies, [link](https://github.com/ossf/scorecard/blob/7cd6406aef0b80a819402e631919293d5eb6adcf/docs/checks.md#pinned-dependencies)
- [ ] Use SAST, [link](https://github.com/ossf/scorecard/blob/7cd6406aef0b80a819402e631919293d5eb6adcf/docs/checks.md#sast)
- [ ] Define security policy, [link](https://github.com/ossf/scorecard/blob/7cd6406aef0b80a819402e631919293d5eb6adcf/docs/checks.md#security-policy)
- [ ] Sign releases, [link](https://github.com/ossf/scorecard/blob/7cd6406aef0b80a819402e631919293d5eb6adcf/docs/checks.md#signed-releases)
- [ ] Set token permissions, [link](https://github.com/ossf/scorecard/blob/7cd6406aef0b80a819402e631919293d5eb6adcf/docs/checks.md#token-permissions)

An output of a recent run using
`GITHUB_AUTH_TOKEN=<REPLACE_ME>  scorecard --repo=https://github.com/defenseunicorns/zarf`:

|  SCORE  |          NAME          |             REASON             |                                               DOCUMENTATION/REMEDIATION                                               |
|---------|------------------------|--------------------------------|-----------------------------------------------------------------------------------------------------------------------|
| 10 / 10 | Binary-Artifacts       | no binaries found in the repo  | [link](https://github.com/ossf/scorecard/blob/7cd6406aef0b80a819402e631919293d5eb6adcf/docs/checks.md#binary-artifacts)  |
| 10 / 10 | CI-Tests               | 30 out of 30 merged PRs checked by a CI test -- score normalized to 10       | [link](https://github.com/ossf/scorecard/blob/7cd6406aef0b80a819402e631919293d5eb6adcf/docs/checks.md#ci-tests)               |
| 0 / 10  | CII-Best-Practices     | no badge detected              | [link](https://github.com/ossf/scorecard/blob/7cd6406aef0b80a819402e631919293d5eb6adcf/docs/checks.md#cii-best-practices)     |
[...]

PR #898, opened 2022-10-12
Add CodeQL

jeff-mccoy:
Related to #897


jeff-mccoy:
Close in favor of #913

PR #906, opened 2022-10-13
Use Zarf user 1000 for Agent container

### LINKED ISSUE #903: Build Rootless/Non-Privileged Zarf Agent Container
Some security postures for Kubernetes clusters do not allow containers to run as root. These types of clusters often employ PodSecurityPolicies (PSPs) that block scheduling any pods that run as root.  The Zarf Agent container is built from `scratch` and as a result the default user is root.

Modify the Dockerfile to create a new user and ensure the container does not run as root or require any elevated privileges.

While it is possible to reduce or exclude certain applications from PSPs or create different PSPs for different applications or ServiceAccounts, I don't believe the Zarf Agent container requires any elevated privileges to run. It would be easier to simply set the user to a non-root user.

**Additional context**
I have already implemented a solution to this and deployed it to multiple clusters using a custom Zarf Agent image with a PSP that blocks containers running as root.

dsmithbauer:
Closes #903

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,17 @@
+FROM ubuntu:latest
+RUN useradd -mu 1000 zarf
+RUN mkdir /home/zarf/.config
+
 FROM scratch
 ARG TARGETARCH
 
 ADD "build/zarf-linux-$TARGETARCH" /zarf
 EXPOSE 8443
 
+COPY --from=0 /etc/passwd /etc/passwd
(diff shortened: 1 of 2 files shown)


jeff-mccoy:
Thanks for submitting this.  I opened a new issue #908 that I think will provide a broader solution than addressing scratch issues one-by-one (there are other concerns with tmp paths when moving from root in scratch).

jeff-mccoy:
Thanks for helping find the right solution, closing and will now be fixed in #913

PR #913, opened 2022-10-16
GH Pipeline Cleanup & Move to Distroless + GHCR

### LINKED ISSUE #113: Set up CODEOWNERS
As a maintainer of Zarf I'd like a CODEOWNERS configuration so that GitHub can automatically decide who should be an approver of a change. 

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

### LINKED ISSUE #903: Build Rootless/Non-Privileged Zarf Agent Container

jeff-mccoy:
Fixes #903 
Fixes #113 
Related to #897 

- Change zarf agent build to distroless vs scratch
- Push agent image to GHCR instead of Dockerhub
- Implement composite components to make workflows more DRY
- Add CodeQL scanning
- Normalize workflow / job names
- Disable logfile creation for zarf agent
- Fix issue with `--no-log-file` not loading properly
- Disable log file creation for cobra help commands, tools, version commands
- Disable svelte unused css warning noise
- Add CODEOWNERS


Racer159 (changes requested):
Overall supportive of this, just a few small thingz

svelte.config.js
```diff
@@ -2,6 +2,11 @@ import adapter from "@sveltejs/adapter-static";
 import preprocess from 'svelte-preprocess';
 
 const config = {
+  onwarn: (warning, handler) => {
```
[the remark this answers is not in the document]
jeff-mccoy: So I looked at this for a while it seems to be some issue with svelte + sass preprocessing.

.github/workflows/test-external.yml
```diff
@@ -19,22 +19,24 @@ jobs:
...
-      - run: "curl -s https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash"
+      - name: Setup golang
+        uses: ./.github/actions/golang
 
-      - run: make build-cli-linux-amd
+      - name: Setup NodeJS
+        uses: ./.github/actions/node
 
-      - uses: ./.github/actions/agent-image
+      - name: Build binary and zarf packages
```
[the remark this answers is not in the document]
jeff-mccoy: Yeah that was missed when rewriting, it's caught by the makefile I believe so it still passes. But I'll fix it.

jeff-mccoy:
Happy to discuss the codeowner section, I did a combination of orgs + individual GH accounts. I like the transparency/version control benefit of using actual GH accounts over just org accounts.

Makefile
```diff
@@ -44,7 +44,7 @@ clean: ## Clean the build directory
 	rm -rf build
 
 destroy: ## Run `zarf destroy` on the current cluster
-	$(ZARF_BIN) destroy --confirm --remove-components
+	$(ZARF_BIN) destroy --confirm --remove-components
```
Racer159, line 47: Added " "

packages/zarf-agent/manifests/deployment.yaml
```diff
@@ -21,7 +21,7 @@ spec:
         - name: private-registry
       containers:
         - name: server
-          image: "###ZARF_REGISTRY###/###ZARF_CONST_AGENT_IMAGE###"
+          image: "###ZARF_REGISTRY###/defenseunicorns/zarf/###ZARF_CONST_AGENT_IMAGE###"
```
Racer159, line 24: this seems hacky, but I get why it is happening.  Is there a way we could improve our regex matching for this in the future?

src/cmd/root.go
```diff
@@ -21,14 +22,11 @@ var v *viper.Viper
...
-
-		// Disable progress bars for CI envs
-		if os.Getenv("CI") == "true" {
-			message.Debug("CI environment detected, disabling progress bars")
-			message.NoProgress = true
+		// Don't add the logo to the help command
+		if cmd.Parent() == nil {
+			skipLogFile = true
 		}
+		alwaysRun()
```
Racer159, line 29: initialSetup() may be a better name for this function than alwaysRun() since that doesn't really explain what it is doing.

Racer159 (changes requested):

packages/zarf-agent/manifests/deployment.yaml
jeff-mccoy: So I actually thought about this a while and unfortunately no.  We include the host in the checksum by design and exclude it here intentionally.  The reason this is _hacky_ is because we aren't mutating it dynamically and it just so happens we previously could ignore the host on dockerhub. If anything we were more hacky before, it just wasn't as obvious.

jeff-mccoy: We could just do the tag, except that we use different images depending on if it's prod or dev.

Racer159 (approved):
unleash the bugs @jeff-mccoy hid in this!

packages/zarf-agent/manifests/deployment.yaml
Racer159: Yeah just kinda sucks

jeff-mccoy:
> unleash the bugs @jeff-mccoy hid in this!

...and there will be sooo many!

PR #921, opened 2022-10-17
920 viper package deploy variables do not work

### LINKED ISSUE #920: Viper Package Deploy Variables Do Not Work
### Environment
Device and OS:
App version:
Kubernetes distro being used:
Other:

### Steps to reproduce
1. Create a `zarf-config.toml` with `[package.deploy.set]` for a package that has deploy-time variables
2. Run zarf package deploy on that package

### Expected result
The set variables are used by Zarf.

### Actual Result
The set variables are ignored by Zarf.

image
image
image

### Severity/Priority
High

### Additional Context
This works fine for package create variables, but not package deploy variables.

Racer159:
## Description

Fixes a bug where deploy variables are not settable (and adds a test to catch it)

## Related Issue

Fixes #920

## Type of change

## Checklist before merging


examples/config-file/simple-configmap.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: simple-configmap
+  namespace: zarf
+data:
+  templateme.properties: |
+    zebra=###ZARF_VAR_DEPLOY_ZEBRA###
```
RothAndrew, line 8: Recommend changing the variable name. It confused me for a minute thinking `ZARF_VAR_DEPLOY` was the necessary prerequisite vs just `ZARF_VAR`.

Maybe different animals for the deploy vars?

Racer159: le fixed

jeff-mccoy (approved):
thx cleanup crew, lgtm!

src/test/e2e/30_config_file_test.go
```diff
@@ -12,7 +13,7 @@ import (
 
 func TestConfigFile(t *testing.T) {
 	t.Log("E2E: Config file")
-	e2e.setup(t)
+	e2e.setupWithCluster(t)
```
jeff-mccoy, line 16: As a general practice, we've tried to put as many tests as we can without needing a cluster for faster feedback. Is there any way to accomplish this test without needing a cluster? If you feel we need it to test templating in-cluster, than we can leave it.

jeff-mccoy: I didn't notice before that we had this as a < 20 test on my prior review

Racer159: I would really like to test with templating since this bug would not be caught otherwise and could be easy to reintroduce through a refactor if someone was not careful.  (and while this test won't run on Windows now, the time cost wasn't that significant to have the cluster).

In the future we could template `file` includes as well and refactor this and the package variables test.  I'll create an issue for that.

Racer159:
@jeff-mccoy mentioning this here to link these together: https://github.com/defenseunicorns/zarf/issues/922

YrrepNoj (approved):

PR #924, opened 2022-10-18
adds overwrite flag to archiver

### LINKED ISSUE #858: Archiver fails if files are already present
I'm working on building a Zarf package to deploy some terraform ([see package here](https://github.com/defenseunicorns/zarf/tree/adds-terraform-example/examples/terraform)), and I ran into an issue with the archiver. If I deploy the aforementioned package, everything works swimmingly, but if I deploy again (say I'm adding another TF provider), then it fails with the error message below. Tracing it back to the codebase and docs, it looks like the `mholt/archiver` will fail to extract if the files already exist, unless you set an option called `OverwriteExisting` in the archiver.

I know it's possible to simply delete the files that are making the archiver fail, but in the case of TF, it feels weird to have to delete the plugins directory. This is especially true if I want to run terraform commands in between updating the TF Zarf package.

I'd like a flag on the `zarf tools archiver decompress ...` command that would allow the archiver to overwrite files if they already exist

**Additional context**
Error message after running `zarf package deploy <myPackage>` twice
Screen Shot 2022-10-06 at 1 43 58 PM

UncleGedd:
## Description
🚧 

## Related Issue

Fixes #858 

## Type of change

## Checklist before merging

### DIFF
--- a/examples/.gitignore
+++ b/examples/.gitignore
@@ -1,1 +1,2 @@
 sync/
+tmp/
--- a/go.mod
+++ b/go.mod
@@ -79,12 +79,10 @@ require (
 	github.com/alibabacloud-go/tea-utils v1.4.4 // indirect
 	github.com/alibabacloud-go/tea-xml v1.1.2 // indirect
 	github.com/aliyun/credentials-go v1.2.3 // indirect
-	github.com/anchore/client-go v0.0.0-20210222170800-9c70f9b80bcf // indirect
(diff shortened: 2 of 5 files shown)


UncleGedd:
Closing because this is nigh impossible to do with the current version of `mholt/archiver` and the next version is still in alpha, more context [here](https://github.com/defenseunicorns/zarf/issues/858#issuecomment-1295581946)

PR #953, opened 2022-10-29
change enable --now to enable and start

Racer159:
## Description

Systemd versions before 2015 (which are still supported under 10 yr extended maintenance) do not respect the --now flag.  See https://github.com/systemd/systemd/blob/dd95b381b26369c02a3ca03585184cb2ea17295b/NEWS#L3043-L3046

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging


jeff-mccoy (approved):
great catch, lgtm

PR #960, opened 2022-10-31
wip- refactoring out the git package library code

YrrepNoj:


src/internal/packager/deploy.go
```diff
@@ -224,7 +224,7 @@ func (p *Packager) deployComponent(component types.ZarfComponent, noImgChecksum
 	}
 
 	if hasRepos {
-		p.pushReposToRepository(componentPath.Repos, component.Repos)
+		_ = p.pushReposToRepository(componentPath.Repos, component.Repos)
```
Racer159, line 227: This function returns errors already so this should probably be handled rather than ignored.

src/internal/packager/deploy.go
```diff
@@ -350,22 +350,33 @@ func (p *Packager) pushImagesToRegistry(componentImages []string, noImgChecksum
...
-			time.Sleep(5 * time.Second)
-			continue
-		} else {
-			break
-		}
+	for _, repoPath := range repos {
+		gitClient := git.New(p.cfg.InitOpts.GitServer)
+
+		utils.Retry(func() error {
+			if totalFailed < 3 {
```
Racer159, line 364: I know it is "one more thing" to add to this refactor but thoughts on making the retry amount a const in config?  This would make it easier to introduce a command flag to tune the number of retries if that is needed on some systems.

Racer159: It may even be nice to pass a function into a retry helper that is more centralized too so that retry logic can be the same everywhere.

Racer159 (changes requested):

Issue #804, opened 2022-09-28
URL Proxy Redirection


Racer159:
**Is your feature request related to a problem? Please describe.**
There are times when actual workloads inside of a Pod container want to reach out to resources that Zarf can bring across an airgap (like git repos) and it would be nice to have a seamless way to handle those requests once you are on the other side.

**Describe the solution you'd like**
A proxy that you can point online domains to (say github.com) that are then transformed and resolved into zarf-managed git repositories on the airgap side.

**Describe alternatives you've considered**
You could `sed` or `awk` your way to victory across all files that may mention an online remote, but depending on how an app or process is architected this may not be very easy or repeatable to do (especially for development use cases).

**Additional context**
N/A

PR #961, opened 2022-10-31
Bring the small fixes from the proxy pr into their own pr

Racer159:
## Description
This contains two small fixes that were noticed during the proxy poc pr.

## Related Issue

Related to #804 

## Type of change

## Checklist before merging


jeff-mccoy:
Since we don't full UI tests for gitea, could you drop a screenshot of the issue this fixes, also update the PR to specify the related issue.

Racer159:
This doesn't actually fix the UI.  It makes it so that calls internal to the cluster and interactions with gitea succeed (see this for an example response from the npm registry: 
image)

jeff-mccoy:
Ah copy thanks for the clarification

jeff-mccoy (approved):

PR #963, opened 2022-11-01
Use a regex manager to look at zarf.yamls and PR for their helm charts

Racer159:
## Description

This adds a regex manager for looking inside of zarf.yaml files and pulling out helm charts.  This will be useful to know when gitea (or any other charts we add) are out of date and in need of updating.

## Related Issue

Fixes # N/A

## Type of change

## Checklist before merging


Racer159:
You can see this working against this repo:

https://github.com/Racer159/zarf

```
renovate --token=<your_token> --include-forks true Racer159/zarf --dry-run
```

image

Also, some notes from this:

This is implemented using the recursive regex strategy so it will match the charts section first, then all of the charts in that section.  Also, renovate will ignore certain directories by default (which is why the regex here only detects 3 files with charts and 4 charts in those files) (see: https://docs.renovatebot.com/presets-default/#ignoremodulesandtests)

renovate.json
```diff
@@ -19,5 +19,16 @@
...
   ],
   "platformAutomerge": true,
-  "platformCommit": true
+  "platformCommit": true,
+  "regexManagers": [
+    {
+      "fileMatch": ["(^|/)zarf.yaml$"],
+      "matchStringsStrategy": "recursive",
+      "matchStrings": [
+        "charts:(.|\n)*?((^\\s{4}[\\w:]+)|\n$|^\\s{2}-)",
```
YrrepNoj, line 28: My regexp fu is failing me today. I'm having a little trouble understanding what the goal of the second capture group is `((^\\s{4}[\\w:]+)|\n$|^\\s{2}-)`

Racer159: That will try to catch the end of the charts section (either 4 space chars then word colon (i.e. the start of a new section `    scripts:`), an empty newline (i.e. the end of the file ``), or 2 space chars and a dash (i.e. the start of a new component `  - name:`))

Racer159: (though it may not be working correctly... lemme play around a bit more)

Racer159: NVM, forgot about the double slashes from JS land 

https://regex101.com/r/XhpCdJ/1

Racer159: (just removed an uneeded group btw)

jeff-mccoy:
What do we think about how this will affect things like the images in a chart? I know it's a problem when rennovate makes an update that has other downstream effects, this seems like one of those uses cases. Basically if someone lets rennovate update, it could make the package broken until someone runs `zarf prepare find-images`, which of course isn't fullproof either.  Thoughts?

Racer159:
> What do we think about how this will affect things like the images in a chart? I know it's a problem when rennovate makes an update that has other downstream effects, this seems like one of those uses cases. Basically if someone lets rennovate update, it could make the package broken until someone runs `zarf prepare find-images`, which of course isn't fullproof either. Thoughts?

I can definitely see that being a potential issue (especially if we unleashed this on something like the minio example package that has no tests), but right now this is limited to the `packages/` directory and the three packages with charts that it sees all have tests that would fail with ErrImagePullBackoff as soon as Renovate makes its PR if the image(s) changed (which would prevent the merge).

Also currently only @jeff-mccoy @YrrepNoj @Racer159 @JasonvanBrackel have access to make updates to approve changes to `packages/` which should curt down on the risk of someone else erroneously approving a renovate PR.

The only gotcha is that we would need to ensure that any new packages added to `packages/` have tests which IMO should be standard practice since `packages/` is our signal that these are actually supported vice just being `examples/`

Would be interested in @YrrepNoj's thoughts as well.

jeff-mccoy (approved):
Thanks for the additional PR comments, this will be a really helpful addition for us, lgtm

PR #971, opened 2022-11-02
Add versions to filenames if they exist

### LINKED ISSUE #970: Ability to specify package version as part of the filename
When I build a zarf package, the package's version is not included as part of its filename, even if I specify a value for `metadata.version` in the zarf.yaml file. This makes it more difficult to track which version of the package is being used on deployment, since they all are named the same way.

If `metadata.version` in zarf.yaml has a value, I'd like that value to appear in the package filename when it is created.

I could rename the package after it is created, but it would be nice if that happened for me.

**Additional context**
Since updating values inside files is a pain when using automation, I think it would also be great if we had a cli flag to set the version of the package on create. Something like `zarf package create --version 1.2.3 --confirm` which would result in a file named `zarf-package-foo-1.2.3-amd64.tar.zst`

Racer159:
## Description

This adds the version to the filename if it is specified in the zarf metadata as we do with the init package.

## Related Issue

Fixes #970

## Type of change

## Checklist before merging


RothAndrew:
Yahoo! That was fast thanks! cc @brianrexrode

YrrepNoj (approved):

YrrepNoj (approved):

Issue #986, opened 2022-11-08
UI - Update UnicornUI. 


mike-winberry:
Update unicorn-ui to v0.0.19

**Describe the solution you'd like**
Update unicorn-ui dependency version and usages of className to use the new default props supported in v0.0.19 (className -> class)

PR #994, opened 2022-11-09
#986. Bump unicorn-ui to v0.0.21. 

mike-winberry:
## Description
Update Props in references to UI components specifically change from className and wrapperClass to instead use class. Update imports and removed unused import.

## Related Issue
#986 

## Type of change
- [x] dependency update to prevent breaking changes.


jeff-mccoy (approved):
Oooh that's cleaner, lgtm

PR #1014, opened 2022-11-15
Fix UI failing silently if init package is not found

### LINKED ISSUE #1013: UI fails silently when init package is not found
### Environment
Device and OS: M1 Max Ventura
App version: main

### Steps to reproduce

1. `make dev`
2. `kind create cluster`
3.  Go to UI home
4. Click init
5. View Silent error

### Expected result

UI handles error and displays properly.

### Actual Result

Blank error page (console contains error)

Noxsios:
## Description

Fixes UI failing silently if init package is not found.

It is worth noting that this PR introduces some basic JSON error messaging to the API + JSON error message handling in the UI's `http.ts`.

## Related Issue

Fixes #1013 

## Type of change


src/internal/api/packages/find.go
```diff
@@ -38,7 +39,20 @@ func findPackage(pattern *regexp.Regexp, w http.ResponseWriter, setDir func() (s
 		message.ErrorWebf(err, w, "Error getting directory")
 	}
 
+	// TODO: decide as a team how JSON errors should be formatted
+	type JSONError struct {
```
YrrepNoj, line 43: Agreed that the todo should stay, but can you move this to `internal/api/common/common.go`? There is some other response writing stuff there already.

src/internal/api/packages/find.go
```diff
@@ -38,7 +39,20 @@ func findPackage(pattern *regexp.Regexp, w http.ResponseWriter, setDir func() (s
...
+		Error string `json:"error"`
+		Message string `json:"message"`
+	}
+
 	// Intentionally ignore errors
```
YrrepNoj, line 48: This comment is throwing me off a little.

Seems like there is a bit of mixed intentions if we are ignoring the error from the `RecursiveFileList` call but then also returning an http error if the call didn't return any files. 

I would vote that we either:
1. Check the error and add it to an error return if it's not nil
2. Continue to not check the error but then also not return an error if empty (and expect the frontend to check/handle that)

YrrepNoj: @jeff-mccoy You initially wrote the `// Intentionally ignore errors` so pinging you to pick your brain too

YrrepNoj (changes requested):

Madeline-UX (dismissed):

src/internal/api/packages/find.go
Noxsios: Chose to handle error

src/internal/api/packages/find.go
```diff
@@ -40,9 +41,14 @@ func findPackage(pattern *regexp.Regexp, w http.ResponseWriter, setDir func() (s
 	targetDir, err := setDir()
 	if err != nil {
 		message.ErrorWebf(err, w, "Error getting directory")
+		return
 	}
 
-	// Intentionally ignore errors
-	files, _ := utils.RecursiveFileList(targetDir, pattern)
+	files, err := utils.RecursiveFileList(targetDir, pattern)
+	if err != nil || len(files) == 0 {
```
YrrepNoj, line 48: Leaving a comment here just to denote the fact that we talked in person about eventually splitting this error return out into an bigger obj that can capture more detail on what specifically went wrong. Eventually...

YrrepNoj (approved):
Looks good to me!

Madeline-UX:
@Noxsios I think it would be helpful to add pictures or a video to show any UI fixes

PR #1039, opened 2022-11-25
Registry pod affinity

jeff-mccoy:


jeff-mccoy:
Screenshot 2022-11-30 at 10 55 18 PM
Screenshot 2022-11-30 at 10 55 56 PM

.github/workflows/test-windows.yml
```diff
@@ -9,7 +9,7 @@ on:
       - "**.svg"
       - "adr/**"
       - "docs/**"
-      - "CODEOWNERS"
```
Racer159, line 12: how do you even make these spaces @jeff-mccoy ?

jeff-mccoy: 😂 no idea...that was an autoformat...

jeff-mccoy: wait... no that was from _YOUR_ pr.... lol

packages/zarf-registry/zarf.yaml
```diff
@@ -3,12 +3,37 @@ metadata:
...
+    description: "The memory limit for the registry"
+    default: "2Gi"
+
 components:
   - name: zarf-seed-registry
     charts:
       - name: docker-registry
-        url: https://github.com/defenseunicorns/docker-registry.helm.git
-        version: 2.1.0-zarf
+        localPath: packages/zarf-registry/chart
```
Racer159, line 35: This was something that really bothered me for composable packages... I think this should be just "chart" since that is the relative directory from this yaml...  Not going to block this PR since this is a more widespread issue (and a breaking change) but something worth thinking about.

jeff-mccoy: I 500% agree; it's actually just a bug......if we had a decent PM he'd probably have already created an issue... 👀

jeff-mccoy:
Thanks for the review @Racer159 I'm going to keep this one open for comment tomorrow since I posted in K8s Slack asking for K8s feedback.

mikevanhemert (approved):

PR #1067, opened 2022-12-02
Fix retry logic for pushing git repos

YrrepNoj:
Remove stale state from the failed push so that we can attempt to push the repository.

### DIFF
--- a/src/pkg/packager/deploy.go
+++ b/src/pkg/packager/deploy.go
@@ -384,12 +384,14 @@ func (p *Packager) pushReposToRepository(reposPath string, repos []string) error
 			gitClient := git.New(p.cfg.State.GitServer)
 
 			// If this is a serviceURL, create a port-forward tunnel to that resource
-			if tunnel, err := cluster.NewTunnelFromServiceURL(gitClient.Server.Address); err != nil {
-				return err
-			} else {
-				tunnel.Connect("", false)
-				defer tunnel.Close()
-				gitClient.Server.Address = fmt.Sprintf("http://%s", tunnel.Endpoint())
+			if cluster.IsServiceURL(gitClient.Server.Address) {
+				if tunnel, err := cluster.NewTunnelFromServiceURL(gitClient.Server.Address); err != nil {
+					return err
+				} else {
+					tunnel.Connect("", false)
+					defer tunnel.Close()
+					gitClient.Server.Address = fmt.Sprintf("http://%s", tunnel.Endpoint())
+				}
 			}
 
 			// Convert the repo URL to a Zarf-formatted repo name
--- a/src/config/lang/english.go
+++ b/src/config/lang/english.go
@@ -199,3 +199,4 @@ const (
 
 // ErrInitNotFound
 var ErrInitNotFound = errors.New("this command requires a zarf-init package, but one was not found on the local system. Re-run the last command again without '--confirm' to download the package")
+var ErrNotAServiceURL = errors.New("the provided URL does not match service url format of http://{SERVICE_NAME}.{NAMESPACE}.svc.cluster.local:{PORT}")
--- a/src/internal/cluster/tunnel.go
+++ b/src/internal/cluster/tunnel.go
@@ -7,7 +7,6 @@ package cluster
 // Forked from https://github.com/gruntwork-io/terratest/blob/v0.38.8/modules/k8s/tunnel.go
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -22,6 +21,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/defenseunicorns/zarf/src/config/lang"
 	"github.com/defenseunicorns/zarf/src/types"
 
 	"github.com/defenseunicorns/zarf/src/config"
@@ -43,6 +43,9 @@ const (
 	ZarfLogging  = "LOGGING"
 	ZarfGit      = "GIT"
 	ZarfInjector = "INJECTOR"
+
+	// See https://regex101.com/r/OWVfAO/1
+	serviceURLPattern = `^(?P<name>[^\.]+)\.(?P<namespace>[^\.]+)\.svc\.cluster\.local$`
 )
 
 // Tunnel is the main struct that configures and manages port forwading tunnels to Kubernetes resources.
@@ -86,6 +89,21 @@ func (c *Cluster) PrintConnectTable() error {
 	return nil
 }
 
+// IsServiceURL will check if the provided string is a valid serviceURL based on if it properly matches a validating regexp
+func IsServiceURL(serviceURL string) bool {
+	parsedURL, err := url.Parse(serviceURL)
+	if err != nil {
+		return false
+	}
+
+	// Match hostname against local cluster service format
+	pattern := regexp.MustCompile(serviceURLPattern)
+	matches := pattern.FindStringSubmatch(parsedURL.Hostname())
+
+	// If incomplete match, return an error
+	return len(matches) == 3
+}
+
 // NewTunnelFromServiceURL takes a serviceURL and parses it to create a tunnel to the cluster. The string is expected to follow the following format:
 // Example serviceURL: http://{SERVICE_NAME}.{NAMESPACE}.svc.cluster.local:{PORT}
 func NewTunnelFromServiceURL(serviceURL string) (*Tunnel, error) {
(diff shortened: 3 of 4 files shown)


src/internal/packager/git/push.go
```diff
@@ -18,6 +18,10 @@ import (
 )
 
 func (g *Git) PushRepo(localPath string) error {
+	// Keeep a copy of the original address in case we need to retry this push
+	originalAddress := g.Server.Address
+	defer g.setServerAddress(originalAddress)
```
YrrepNoj, line 23: Style thoughts on doing the following block instead? (so we don't need to make a setter func)
```
defer func() { g.Server.Address = originalAddress }()
```

YrrepNoj: I think what's currently in the PR is more clear, but I generally dislike getter/setter functions so I could go either way.

Racer159: I prefer `defer func() { g.Server.Address = originalAddress }()` but also come from a dirty JavaScript background where that kind of thing is normal.  I like setters when they are going to be reused, but not when they aren't.

src/internal/packager/git/push.go
```diff
@@ -135,7 +142,7 @@ func (g *Git) push(repo *git.Repository, spinner *message.Spinner) error {
 	} else if errors.Is(err, git.NoErrAlreadyUpToDate) {
 		message.Debugf("Repo already up-to-date, skipping fetch...")
 	} else if err != nil {
-		message.Warnf("unable to fetch remote cleanly prior to push: %w", err)
+		message.Warnf("unable to fetch remote cleanly prior to push: %v", err)
```
Racer159, line 145: Why change this from a %w?

src/internal/packager/git/push.go
jeff-mccoy: I recommend we not do this at all and instead just recreate the variable in the outer loop to keep things clean. Defers are not guaranteed to run and this looks like a recipe for future bugginess.  I would love to rewrite the caller to something like this below (untested), thoughts?

Screenshot 2022-12-02 at 2 38 08 PM

```golang

// Push all of the components git repos to the configured git server
func (p *Packager) pushReposToRepository(reposPath string, repos []string) error {
	// Try repo push up to 3 times
	for _, repoURL := range repos {
		// Reusable function to attempt to push a git repo
		tryPush := func() error {
			gitClient := git.New(p.cfg.State.GitServer)

			// Convert the repo URL to a Zarf-formatted repo name
			if repoPath, err := gitClient.TransformURLtoRepoName(repoURL); err != nil {
				return fmt.Errorf("unable to get the repo name from the URL %s: %w", repoURL, err)
			} else {
				// On successful conversion, try to push the repo
				return gitClient.PushRepo(filepath.Join(reposPath, repoPath))
			}
		}

		// Try to push the repo up to 3 times
		if err := utils.Retry(tryPush, 3, 5*time.Second); err != nil {
			return fmt.Errorf("unable to push repo %s to the Git Server: %w", repoURL, err)
		}
	}

	return nil
}
```

src/internal/packager/git/push.go
jeff-mccoy: I think we should discuss using warns here now since we are bubbling all errors. I'd prefer to see these as `fmt.Errorf()` with `%w` (for error unwrapping) now.

YrrepNoj: @jeff-mccoy I'm a little confused by your comment, can you say that again?

> we should discuss using warns here now

and 

> prefer to see this as `fmt.Errorf()` 

seem to contradict each other.

YrrepNoj: Are you saying you don't want this to be a warning and you would rather just return the error immediately?

jeff-mccoy: Haha yes English is hard apparently 👀

src/internal/packager/git/push.go
YrrepNoj: Looking through the recommendation now, but

> Defers are not guaranteed to run

That is kind of news to me. I think the only time I know defers don't run is if an `os.Exit()` is called which would make this fix mute anyways.

YrrepNoj: This implementation would have the function `(g *Git) PushRepo(...)` modifying the `g.Server.Address` to an IP/PORT combo that no longer is accessible after the function returns. That is fine with this specific implementation since we are throwing away the `gitClient` right after this push is completed but that is not an intuitive thing and people using that function could easily slip up.

@jeff-mccoy What do you think about turning `(g *Git) PushRepo(...)` into a more pure wrapper around `git push` and force the caller to create the k8s tunneled address before calling?

YrrepNoj: ^ Above comment is assuming we remove the `defer g.setServerAddress()` like you were implying we should

Racer159: I would prefer that we purify the push / pull functions into more pure drivers of git (like clone does) and then we can call those functions in a retry loop more easily from a driver function (that way we still have one client and that client wont keep getting recalled in the retry loop.  IMO that provides better separation of concerns and won't unnecessarily create more and more clients in memory.

jeff-mccoy: Don't know if I'm tracking the clients in memory comment, yes new would be crated and GCd, but they are very small, it's not giant like packager.

Racer159: Yeah its not a big memory leak for sure, but I just feel it is generally cleaner to separate the the overall driving function, the receiver client and the wrapper around git (and then the retry logic can more happily live inside of the receiver instead of being mixed in with git operations).  Also I usually just prefer to not rely on GC if you don't have to

src/internal/packager/git/common.go
```diff
@@ -40,3 +40,7 @@ func NewWithSpinner(server types.GitServerInfo, spinner *message.Spinner) *Git {
 		Spinner: spinner,
 	}
 }
+
+func (g *Git) setServerAddress(address string) {
```
jeff-mccoy, line 44: Is this still used?

src/pkg/packager/deploy.go
```diff
@@ -377,19 +377,48 @@ func (p *Packager) pushImagesToRegistry(componentImages []string, noImgChecksum
 
 // Push all of the components git repos to the configured git server
 func (p *Packager) pushReposToRepository(reposPath string, repos []string) error {
+	// // Try repo push up to 3 times
```
jeff-mccoy, line 380: assume these will be deleted?

src/pkg/packager/deploy.go
```diff
@@ -377,19 +377,48 @@ func (p *Packager) pushImagesToRegistry(componentImages []string, noImgChecksum
...
+	// }
+
 	// Try repo push up to 3 times
 	for _, repoURL := range repos {
-		gitClient := git.New(p.cfg.State.GitServer)
-		repoPath, err := gitClient.TransformURLtoRepoName(repoURL)
-		if err != nil {
-			return fmt.Errorf("unable to get the repo name from the URL %s: %w", repoURL, err)
+
+		// Create an anonymous function to push the repo to the Zarf git server
```
jeff-mccoy, line 399: Yeah this feels less disgusting

src/internal/packager/git/common.go
YrrepNoj: nope! Thanks

src/pkg/packager/deploy.go
```diff
@@ -377,19 +377,32 @@ func (p *Packager) pushImagesToRegistry(componentImages []string, noImgChecksum
...
+		tryPush := func() error {
+			gitClient := git.New(p.cfg.State.GitServer)
+
+			// If this is a serviceURL, create a port-forward tunnel to that resource
+			if tunnel, err := cluster.NewTunnelFromServiceURL(gitClient.Server.Address); err != nil {
```
Racer159, line 387: Calling this every time like this breaks external git servers (it works for our "external" test because that is still a service inside the cluster...)

image

^ tested on https://gitlab.com

Racer159 (dismissed):
image

src/pkg/packager/deploy.go
Racer159: If you check this pattern first I think it would work for all cases again:

```
	pattern := regexp.MustCompile(`^(?P<name>[^\.]+)\.(?P<namespace>[^\.]+)\.svc\.cluster\.local$`)
```

src/pkg/packager/deploy.go
```diff
@@ -377,19 +377,34 @@ func (p *Packager) pushImagesToRegistry(componentImages []string, noImgChecksum
...
-		if err != nil {
-			return fmt.Errorf("unable to get the repo name from the URL %s: %w", repoURL, err)
+
+		// Create an anonymous function to push the repo to the Zarf git server
+		tryPush := func() error {
+			gitClient := git.New(p.cfg.State.GitServer)
+
+			// If this is a serviceURL, create a port-forward tunnel to that resource
+			if cluster.IsServiceURL(gitClient.Server.Address) {
+				if tunnel, err := cluster.NewTunnelFromServiceURL(gitClient.Server.Address); err != nil {
```
jeff-mccoy, line 388: It's a bit in the nerd-weeds but since you brought it up earlier, I don't actually believe there will be a GC issues with this setup. Brushed up on some golang GC behavior and I _think_ because there will be a goroutine that is still accessing the pointer internally that the value will not be GCd early.

jeff-mccoy (approved):
lgtm

PR #1070, opened 2022-12-02
Update cached repository with latest from upstream

YrrepNoj:
When using a cached repository, make sure we have all the latest changes from the upstream repository.


src/internal/packager/git/pull.go
```diff
@@ -119,3 +118,45 @@ func (g *Git) pull(gitURL, targetFolder string, repoName string) {
 		}
 	}
 }
+
+func (g *Git) pull() error {
```
YrrepNoj, line 122: @YrrepNoj write a test to verify this updates the workingdir HEAD refs

Racer159, line 122: Almost wonder if this should be separated into an actual pull.go and then rename this file to something like download.go or retrieve.go since it is driving other git actions.

src/internal/packager/git/pull.go
```diff
@@ -119,3 +118,45 @@ func (g *Git) pull(gitURL, targetFolder string, repoName string) {
...
+	if err != nil || len(remotes) == 0 {
+		return fmt.Errorf("unable to find remote repo: %w", err)
+	}
+
+	// Set auth if provided
+	gitURL := remotes[0].Config().URLs[0]
+	gitCred := g.FindAuthForHost(gitURL)
+	if gitCred.Auth.Username != "" {
+		pullOptions.Auth = &gitCred.Auth
+
```
Racer159, line 146: extra space

src/internal/packager/git/pull.go
jeff-mccoy: I'm not opposed to that at all

YrrepNoj:
Closing for now because we're going to need to fix #1079 before we can confidently say this is fixed too

PR #1077, opened 2022-12-03
Fix SBOMs in Zarf

### LINKED ISSUE #1076: SBOMs are no longer created in Zarf
### Environment
Device and OS: Dell Precision T1700 - Arch Linux
App version: latest on main
Other: N/A

### Steps to reproduce
1. Create a fresh package
2. Try to inspect it with the -s flag

### Expected result
Running Zarf package create (without --skip-sbom) generates SBOMs and I can inspect them

### Actual Result
Running Zarf package create (without --skip-sbom) skips SBOM creation and I cannot inspect them

image
image
image

### Severity/Priority
Critical

### Additional Context
N/A

Racer159:
## Description

This fixes simple errors that disabled SBOMs entirely for package create and inspect.

## Related Issue

Fixes #1076 

## Type of change

## Checklist before merging


src/pkg/packager/create.go
```diff
@@ -142,7 +142,7 @@ func (p *Packager) pullImages(imgList []string, path string) (map[name.Tag]v1.Im
 
 		pulledImages, err = imgConfig.PullAll()
```
YrrepNoj, line 144: nit; this empty line could be removed

YrrepNoj (approved):

src/test/e2e/00_use_cli_test.go
```diff
@@ -70,9 +70,11 @@ func TestUseCLI(t *testing.T) {
 	pkgName := fmt.Sprintf("zarf-package-dos-games-%s.tar.zst", e2e.arch)
 
 	stdOut, stdErr, err = e2e.execZarfCommand("package", "create", "examples/game", "--confirm", "--zarf-cache", cachePath)
+	require.Contains(t, stdErr, "Creating SBOMs for 1 images")
 	require.NoError(t, err, stdOut, stdErr)
 
-	stdOut, stdErr, err = e2e.execZarfCommand("package", "inspect", pkgName)
+	stdOut, stdErr, err = e2e.execZarfCommand("package", "inspect", pkgName, "-s")
```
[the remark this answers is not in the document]
Racer159: Yeah it works just fine.  I messed around with this quite a bit and what (I think) is happening is that because there is no standard in to read (we don't provide a reader to the command) it sees an EOF and the survey library we use just continues as if we had just hit enter.  The same thing will happen for other interactive commands (i.e. create without --confirm) it will give the error as if the user canceled it (not an interrupt or something else)

src/pkg/packager/create.go
Racer159: I think I will merge as is.  Maybe I am weird but I like newlines between blocks (we should talk about style though more broadly though)

PR #1085, opened 2022-12-05
Make data injection idempotent

jeff-mccoy:
This PR addresses some race conditions that can exist when running data injections or performing data injections. Key changes:

- Replace `WaitForPodsAndContainers()` second argument, `waitForAllPods` with a new filter function and simply return all pods found leaving it to the caller to determine any further action (only data injection needed more than one pod)
- Properly deal with matching the correct init containers based on the `dataInjectionMarker` so that we don't match other failed installs
- Properly validate after injection that at least one pod with the correct `dataInjectionMarker` is now ready
- No longer just warn and keep going on failure but actually retry the injection again
- Add 3 x data injection tests to ensure idempotence is working as expected


src/internal/cluster/data.go
```diff
@@ -111,3 +116,24 @@ iterator:
 		return
 	}
 }
+
+// Pod filter to ensure we only use the current deployment's pods
+func podFilterByInitContainer(data types.ZarfDataInjection) k8s.PodFilter {
+	return func(pod corev1.Pod) bool {
+		// Loop over all init containers
+		for _, container := range pod.Spec.InitContainers {
```
Racer159, line 124: What if someone is using the data injections without the init-container pattern (some apps that dynamically pickup things from disk won't need to have an init container for inject)

src/test/e2e/23_data_injection_test.go
```diff
@@ -22,20 +22,27 @@ func TestDataInjection(t *testing.T) {
...
 	path := fmt.Sprintf("build/zarf-package-data-injection-demo-%s.tar", e2e.arch)
 
-	// Limit this deploy to 5 minutes
-	ctx, cancel := context.WithTimeout(context.TODO(), 5*time.Minute)
-	defer cancel()
-
-	// Deploy the data injection example
-	stdOut, stdErr, err := utils.ExecCommandWithContext(ctx, true, e2e.zarfBinPath, "package", "deploy", path, "--confirm")
-	require.NoError(t, err, stdOut, stdErr)
+	// Repeat the inejction action 3 times to ensure the data injection is idempotent and doesn't fail to perform an upgrade
```
Racer159, line 25: inejction

src/internal/cluster/data.go
Racer159: IMO we would want to look at the init containers in one loop and the real containers in a second loop (potentially breaking the insides out into another function).  We could also splice the arrays together potentially since they likely are the same struct type.

Racer159 (changes requested):

jeff-mccoy:
updated as requested

Racer159 (approved):
LGTM!

PR #1095, opened 2022-12-06
First version of sbom output

### LINKED ISSUE #1075: Output SBOM folder on package create
It would be nice to output the sbom (optionally) into a folder on package create so that it could be referenced before bringing the package into a disconnected environment.

A new command line flag that allows you to output the SBOM on package create.

We could automatically do this on create or use inspect but it is easier to have it all in one command.

**Additional context**
This is useful so that you don't need to unarchive the package and rearchive it to pull out the SBOM

Racer159:
## Description

Add new flags to output SBOM files into specified directories **AND** change the sbom inspect behavior to also open the web browser automatically

## Related Issue

Fixes #1075

## Type of change

## Checklist before merging

### DIFF
--- a/src/internal/packager/sbom/tools.go
+++ b/src/internal/packager/sbom/tools.go
@@ -76,12 +76,14 @@ func ViewSBOMFiles(tmp types.TempPaths) {
 }
 
 // OutputSBOMFiles outputs the sbom files into a specified directory
-func OutputSBOMFiles(tmp types.TempPaths, outputDir string) error {
-	if err := os.RemoveAll(outputDir); err != nil {
+func OutputSBOMFiles(tmp types.TempPaths, outputDir string, packageName string) error {
+	packagePath := filepath.Join(outputDir, packageName)
+
+	if err := os.RemoveAll(packagePath); err != nil {
 		return err
 	}
 
-	if err := utils.CreatePathAndCopy(tmp.Sboms, outputDir); err != nil {
+	if err := utils.CreatePathAndCopy(tmp.Sboms, packagePath); err != nil {
 		return err
 	}
 
--- a/src/pkg/packager/create.go
+++ b/src/pkg/packager/create.go
@@ -128,7 +128,7 @@ func (p *Packager) Create(baseDir string) error {
 
 	// Output the SBOM files into a directory if specified
 	if p.cfg.CreateOpts.SBOMOutputDir != "" {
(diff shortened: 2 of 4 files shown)


src/types/runtime.go
```diff
@@ -41,6 +41,8 @@ type ZarfCreateOptions struct {
 	SkipSBOM        bool              `json:"skipSBOM" jsonschema:"description=Disable the generation of SBOM materials during package creation"`
 	Insecure        bool              `json:"insecure" jsonschema:"description=Disable the need for shasum validations when pulling down files from the internet"`
 	OutputDirectory string            `json:"outputDirectory" jsonschema:"description=Location where the finalized Zarf package will be placed"`
+	SBOM            bool              `json:"sbom" jsonschema:"description=Whether to pause to allow for viewing the SBOM post-creation"`
+	SBOMOutput      string            `json:"sbomOutput" jsonschema:"description=Location to output an SBOM into after package creation"`
```
YrrepNoj, line 45: thoughts on renaming these variables to something more like `SBOMOutputDir` and `ViewSBOM` ?

YrrepNoj (changes requested):

src/types/runtime.go
YrrepNoj: My main thought was that `SBOM bool` looks like it would just be the opposite of `SkipSBOM bool`

src/internal/packager/sbom/tools.go
```diff
@@ -46,3 +49,41 @@ func WriteSBOMFiles(sbomViewFiles []string) error {
...
+		}
+		_ = survey.AskOne(prompt, &value)
+	} else {
+		message.Note("There were no images with software bill-of-materials (SBOM) included.")
+	}
+}
+
+// OutputSBOMFiles outputs the sbom files into a specified directory
+func OutputSBOMFiles(tmp types.TempPaths, outputDir string) error {
+	if err := os.RemoveAll(outputDir); err != nil {
```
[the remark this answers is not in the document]
Racer159: My thought was that may get confusing since it would just dump all sboms into a single folder and you wouldn't know which sboms came from where anymore.  We could create sub directories for them potentially if we wanted (though we would need to be careful about name collisions there too).

src/types/runtime.go
Racer159: Yeah I wasn't comfortable with this either, will update.

YrrepNoj (approved):
lgtm

jeff-mccoy (approved):

PR #1100, opened 2022-12-07
showing configs failing in secrets

runyontr:
## Description

Showing error for #1101


Racer159:
Added the breaking change flag since new packages created past this PR will not be deployable with older Zarf versions, but older packages should still work with this version of Zarf (we could also drop this compatibility if statement but it will break more things)

src/pkg/packager/deploy.go
```diff
@@ -459,10 +460,29 @@ func (p *Packager) installChartAndManifests(componentPath types.ComponentPaths,
...
+				kPath := filepath.Join(componentPath.Manifests, k)
+
+				// Run the template engine against the kustomizations (if they exist) before they are built
+				valueTemplate, err := template.Generate(p.cfg)
+				if err != nil {
+					return installedCharts, err
+				}
+				template.ProcessYamlFilesInPath(kPath, component, valueTemplate)
+
+				if err := kustomize.BuildKustomization(kPath, destination, manifest.KustomizeAllowAnyDirectory); err != nil {
```
runyontr, line 479: I'm worried about this and remote bases in the airgap.  Does the Kustomize build properly replace a kustomize file like this one:

https://github.com/defenseunicorns/zarf/blob/main/packages/big-bang-core/kustomization/core-standard/kustomization.yaml#L2

It seems like it now may require connectivity for the kustomize build to access that address.

Racer159: We may need to run something like this when it is available or break the build into two steps: https://github.com/kubernetes-sigs/kustomize/issues/3980

PR #1102, opened 2022-12-07
1091 .netrc support inside zarf itself

### LINKED ISSUE #1091: .netrc support inside Zarf itself
It would be nice to add support for .netrc files inside of zarf's credential helper for git.  Git natively supports .netrc (as does curl and other tools) and this would make this more useful for people who use that over .git-credentials alone.

Add searching for .netrc files into the credential helper.

We could add this as a bigger refactor (.netrc would apply to file downloads as well potentially) but git seems like the simplest for now.

**Additional context**
If a person uses .netrc it will work but only in fallback mode.

Racer159:
## Description

This adds support for .netrc files into Zarf using the pre-curl 7.84.0 method: https://daniel.haxx.se/blog/2022/05/31/netrc-pains/

## Related Issue

Fixes #1091 

## Type of change

## Checklist before merging


Racer159:
Realize this will need a github workflow, but before I make that I wanted to get feedback on this approach/pattern/layout for unit tests within Zarf.  Was planning to add some notes to our docs on how we choose to unit test, and was thinking about the following criteria (at least for now):

1. Externally Defined Interfaces - There should be a clear interface to map to, ideally defined by another team/product
2. Compact Code Sections - Unit tests should test units, and not span too many files or functions

jeff-mccoy:
Adding needs-adr since this is introducing unit tests and all that entails. I've been thinking about this and think it seems reasonable to push for unit tests exclusively in the `/src/pkg/*` path where some of that code might eventually be pushed to its own library. This would mean we'd need to be willing to push things to some logical place as library code, including this git code. Thoughts?

Racer159:
Moved the git code to pkg and created an ADR.

.github/workflows/test-unit.yml
```diff
@@ -0,0 +1,40 @@
...
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup golang
+        uses: ./.github/actions/golang
+
+      - name: Setup NodeJS
```
jeff-mccoy, line 29: Is there a way to _not_ add javascript right now? I think that could easily turn into a slipper slope of randomly adding javascript unit tests....

jeff-mccoy (commented):
I'm not totally sure we want to move the entire `git` package like this to `pkg`, we should probably discuss keeping a chunk of it internal and just exposing a limited surface like we do with `pkg/packager`.  I think it could easily become a precedence to try and move it all to `pkg` and I really don't think we're ready to support that. When I was suggesting moving to `pkg` to make it testable I was thinking more of just the password parsing functions for git maybe under `utils/git.go` for example vs the entire package, which I don't think we really should expose as library code.

Racer159:
Changes made, though kept it called "auth" since it could apply to other things in the future (such as auth'ed file downloads). (can rename git if desired)

.github/workflows/test-unit.yml
Racer159: removed

CONTRIBUTING.md
```diff
@@ -53,9 +53,13 @@ In this repo we use [pre-commit](https://pre-commit.com/) hooks for automated va
 
 > ℹ️ **HINT:** *Consider [automatically enabling the hooks in every Git repository](https://pre-commit.com/#automatically-enabling-pre-commit-on-repositories)*
 
-### End2End Testing
+### Code Testing
 
-Our E2E tests can be found in the `/test` folder and follow the journey of someone as they would use the Zarf CLI.  In CI these tests run against our currently supported cluster distros.  You can learn more about testing of Zarf [here](https://docs.zarf.dev/docs/developer-guide/testing).
+Our E2E tests can be found in the `/test` folder and follow the journey of someone as they would use the Zarf CLI.  In CI these tests run against our currently supported cluster distros, and are the primary way that Zarf code is tested.
+
+Our Unit tests can be found as `*_test.go` files inside the package that they are designed to test.  These are also run in CI and are designed to test small functions with clear interfaces that would be difficult to test otherwise.
```
jeff-mccoy, line 60: Probably should mention we only have unit tests in `src/pkg`, especially since the make target reflects that as well.

adr/0008-begin-using-unit-tests.md
```diff
@@ -0,0 +1,44 @@
+# 8. Begin using unit tests
```
jeff-mccoy, line 1: great ADR

src/cmd/tools.go
```diff
@@ -121,13 +121,13 @@ var generatePKICmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		pki := pki.GeneratePKI(args[0], subAltNames...)
 		if err := os.WriteFile("tls.ca", pki.CA, 0644); err != nil {
-			message.Fatalf(err, lang.ErrWritingFile, "tls.ca", err)
```
jeff-mccoy, line 124: Was there a particular reason to change these from error unwrapping (as we do most places now) to simple string?

src/pkg/utils/auth.go
```diff
@@ -0,0 +1,174 @@
...
+	Auth http.BasicAuth
+}
+
+// FindAuthForHost finds the authentication scheme for a given host using .git-credentials then .netrc
+func FindAuthForHost(baseUrl string) Credential {
+	homePath, _ := os.UserHomeDir()
+
+	// Read the ~/.git-credentials file
+	credentialsPath := filepath.Join(homePath, ".git-credentials")
+	credentialsFile, _ := os.Open(credentialsPath)
```
jeff-mccoy, line 30: dogsledding without a comment... sus

src/pkg/utils/auth.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+
+		// Prepare our line to be tokenized
+		line = strings.ReplaceAll(line, "\t", " ")
+		line = strings.TrimSpace(line)
+
+		tokens := strings.Split(line, " ")
+
+		for _, token := range tokens {
+			if activeCommand != "" {
```
jeff-mccoy, line 123: I'm allergic to `else if` and this feels like it's really close to being just right for a `switch` statement instead

jeff-mccoy (commented):
netrc looks disgusting 😆 -- left some additional comments

src/cmd/tools.go
Racer159: Unit tests will fail since that method doesn't support error unwrapping...

image

(Pretty much) everywhere else is a fmt.Errorf function whereas this is the final message.Fatalf)

Racer159: Technically since they are in a different folder it would work now, but I felt it was better to correct those.

src/pkg/utils/auth.go
Racer159: :hand_over_mouth:

src/pkg/utils/auth.go
Racer159: changed

src/pkg/utils/auth.go
Racer159: added a comment

adr/0008-begin-using-unit-tests.md
Racer159: thanks!

CONTRIBUTING.md
Racer159: done

jeff-mccoy (approved):
sgtm, I maintain netrc is nasty 😆

PR #1103, opened 2022-12-08
Remove test now covered by zarf init usage of config file

jeff-mccoy:
Removes a test that was previously needed to make sure a package in a default path was read. Now that #1039 is merged, we have this covered on init package create / deploy + it was was trampling over that file during this test.


YrrepNoj (approved):

jeff-mccoy:
🙏

PR #1105, opened 2022-12-08
Add YOLO mode

### LINKED ISSUE #1046: [DOCS] Unable to find docs on the "ignore" label
I know the ability to tell zarf's mutating webhook to ignore image refs by applying a label to either the pod or the namespace exists, but a quick search of the docs site did not find any documentation on that capability.

docs on that capability

### LINKED ISSUE #1051: Support for Zarf in a Connected Environment
I'd like to use Zarf for declarative deployments, but I'm not deploying in an airgapped environment. In other words, I don't need the Gitea server, container registry or mutating webhook. 

I'd like to be able to create a Zarf package that doesn't depend on the Gitea server, container registry or Zarf agent. In other words, I want to use Zarf without its mutating webhook.

- I know I can add the `zarf-agent=ignore` label to namespaces, but I'd like to not deploy the agent at all
- Technically, I could just use the Gitea server and container registry, but the resulting package is >10GB and it was error-prone during creation (I had issues related to caching); this is not tenable for my non-platform dev audience to use.

**Additional context**
Why I want to use Zarf even though I'm not in an airgapped environment:
- Instead of having app teams fork my IaC/CaC, I want an immutable artifact that can be deployed that abstracts away some of the complexity around K8s, Flux, etc.
- This artifact needs to be immutable so I can ensure app teams don't change the IaC/CaC and inadvertently violate security controls
- I want non-platform devs to be able to easily perform cluster upgrades

jeff-mccoy:
Provides the ability to create Zarf packages intended to work in online-only environments, i.e., without the internal Zarf airgap components typically provided by `zarf init`.

## Related Issue
Fixes #1051 
Fixes #1046 
Related to #1134 

## Type of change

## Checklist before merging


jeff-mccoy:
Deployment of the yolo package to a clean cluster
Screenshot 2022-12-12 at 7 11 14 PM

Failure on OCI image inclusion
Screenshot 2022-12-12 at 7 08 52 PM

Failure on git repo inclusion
Screenshot 2022-12-12 at 7 09 41 PM

Racer159 (changes requested):

Racer159:
Not going to put this in the needs changes category, but this does feel like it would warrant more user messaging on deploy (i.e. a big warning note that this will attempt to pull things from the internet).  A real Zarf package could look this small, and the yolo part likely wouldn't really be understood by our deploy persona on an airgapped system.  In the future this could lead to confusion for what is (at least mostly still) an airgap tool @Madeline-UX for thoughts as well

jeff-mccoy:
> Not going to put this in the needs changes category, but this does feel like it would warrant more user messaging on deploy (i.e. a big warning note that this will attempt to pull things from an online source). A real Zarf package could look this small, and the yolo part likely wouldn't really be understood by our deploy persona on an airgapped system. In the future this could lead to confusion for what is (at least mostly still) an airgap tool @Madeline-UX for thoughts as well

Yeah that's a good point. But also don't want to totally scare ppl off either with some big scary message.  * thinking * 🤔

src/internal/packager/validate/validate.go
```diff
@@ -18,6 +18,10 @@ import (
 
 // Run performs config validations
 func Run(pkg types.ZarfPackage) error {
+	if pkg.Kind == "ZarfInitConfig" && pkg.Metadata.YOLO {
+		return fmt.Errorf("sorry, you can't YOLO an init package")
```
YrrepNoj, line 22: This seems like something that would be good for the `english.go` lang file. Low enough hanging fruit.

src/pkg/packager/deploy.go
```diff
@@ -305,7 +305,13 @@ func (p *Packager) getUpdatedValueTemplate(component types.ZarfComponent) (value
 
 	state, err := p.cluster.LoadZarfState()
 
-	// If no distro the zarf secret did not load properly
+	// YOLO mode but no state found so just fake it this run
+	if p.cfg.Pkg.Metadata.YOLO && state.Distro == "" {
```
YrrepNoj, line 309: If there is state here( ie. the cluster has been initialized with zarf), this function doesn't catch that and continues on to complete the package deployment. This can often lead to issues since the zarf-agent will mutate any urls to images that YOLO package is trying to use. 

Is that what we want, (acceptable risk given the whole 'yolo' namesake)?

src/pkg/packager/deploy.go
```diff
@@ -305,7 +305,13 @@ func (p *Packager) getUpdatedValueTemplate(component types.ZarfComponent) (value
...
+		state.Distro = "YOLO"
+		err = nil
+	}
+
+	// If there is no state, or the state is invalid, return early
 	if err != nil || state.Distro == "" {
```
YrrepNoj, line 315: I know you don't like `else if`, but I wanted to throw in my 2cents that I think this should be one.

Imo it's too easily for someone to accidentally throw code in-between this line and where the err was actually generated.

src/pkg/packager/deploy.go
YrrepNoj: One thing we can potentially play around with is pre-adding the `zarf.dev/agent=ignore` label to yolo packages if they specify their namespaces?

YrrepNoj (changes requested):

src/pkg/packager/deploy.go
jeff-mccoy: Thoughts? 
Screenshot 2022-12-15 at 3 02 10 PM

jeff-mccoy: Or maybe even simpler
Screenshot 2022-12-15 at 3 03 26 PM

jeff-mccoy: hmmm well that could maybe still have `err = nil` and `state.Distro = ""`, dang

jeff-mccoy: This should cover those different edge cases
Screenshot 2022-12-15 at 3 15 23 PM

src/pkg/packager/deploy.go
jeff-mccoy: So thought about this some more, I think it's mixing too many concerns, and we won't support a hybrid YOLO + initialize cluster for now. If they deploy some YOLO things and then later run `zarf init`, we already handle this by the fact that existing namespaces are exempted from the Zarf Agent. I'll add a warning about running a YOLO package on a Zarf cluster as a balance.

src/pkg/packager/deploy.go
```diff
@@ -304,10 +304,26 @@ func (p *Packager) getUpdatedValueTemplate(component types.ZarfComponent) (value
 	defer spinner.Stop()
 
 	state, err := p.cluster.LoadZarfState()
+	// Return on error if we are not in YOLO mode
```
YrrepNoj, line 307: I thought this block looked a little weird, wanted to suggest this (but won't block the PR on it):

``` Go
	// Return an error if we are unable to load state (assuming we are not in YOLO mode)
	if (err != nil && !p.cfg.Pkg.Metadata.YOLO) || (state.Distro == "" && !p.cfg.Pkg.Metadata.YOLO) {
		return values, fmt.Errorf("unable to load the Zarf State from the Kubernetes cluster: %w", err)
	}

	if p.cfg.Pkg.Metadata.YOLO {
		state.Distro = "YOLO"
		message.Warn("This package is in YOLO mode, ...")
	}
```

YrrepNoj (approved):

src/pkg/packager/deploy.go
```diff
@@ -304,10 +304,26 @@ func (p *Packager) getUpdatedValueTemplate(component types.ZarfComponent) (value
...
+	if err != nil && !p.cfg.Pkg.Metadata.YOLO {
+		return values, fmt.Errorf("unable to load the Zarf State from the Kubernetes cluster: %w", err)
+	}
```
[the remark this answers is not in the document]
Racer159: @YrrepNoj you can do multiline suggestions too

Racer159 (approved):

Madeline-UX:
> > Not going to put this in the needs changes category, but this does feel like it would warrant more user messaging on deploy (i.e. a big warning note that this will attempt to pull things from an online source). A real Zarf package could look this small, and the yolo part likely wouldn't really be understood by our deploy persona on an airgapped system. In the future this could lead to confusion for what is (at least mostly still) an airgap tool @Madeline-UX for thoughts as well
> 
> Yeah that's a good point. But also don't want to totally scare ppl off either with some big scary message. * thinking * 🤔

What about following the pattern that has already been established when Zarf downloads a package from the internet? It asks and double confirms that this is the action the user wishes to take. Considering the security posture of default to not harming a system I could see a warrant here for a check

Another option is to add a separate field or type to the packages? Maybe this has already been done and I missed it

PR #1107, opened 2022-12-08
Publish versioned homebrew releases without a 'v' in the VersionName

YrrepNoj:
Brew does not understand how to install `zarf@v0.22.2.rb` but does understand how to install `zarf@0.22.2.rb`

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -69,7 +69,10 @@ brews:
     commit_msg_template: "Brew formula update for {{ .ProjectName }} version {{ .Tag }}"
     homepage: "https://zarf.dev/"
     description: "DevSecOps for Air Gap"
-  - name: "zarf@{{ .Tag }}"
+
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -69,7 +69,10 @@ brews:
     commit_msg_template: "Brew formula update for {{ .ProjectName }} version {{ .Tag }}"
     homepage: "https://zarf.dev/"
     description: "DevSecOps for Air Gap"
-  - name: "zarf@{{ .Tag }}"
+
+  # NOTE: We are using .Version instead of .Tag because homebrew has weird semver parsing rules and won't be able to
```
Racer159, line 73: Technically it isn't weird, real semver doesn't support the v

Racer159 (approved):

.goreleaser.yaml
jeff-mccoy: Eh who cares https://semver.org/#is-v123-a-semantic-version

PR #1110, opened 2022-12-10
Remove all spinner.Debugf messages

Racer159:
## Description

This is a hotfix to remove spinner.Debugf messages from Zarf.  This was a little used feature, but it would often result in missing logs for the places that did use it and had a conflicting purpose already with message.Debugf.

## Related Issue

Fixes #N/A

## Type of change


jeff-mccoy (approved):
Thanks for cleaning this up, lgtm

PR #1112, opened 2022-12-10
Fix where we are deleting tags (don't edit cache\!)

### LINKED ISSUE #1079: Repository cache gets invalidated after several builds
Related to #1069 

When rebuilding a package, zarf will attempt to use repositories that already exist in the `~/.zarf-cache/repos` directory. When testing the fix for another (relatively similar issue) I realized that rebuilding the same package multiple times (3+ times) will result in a failure because the HEAD of the repository in the cache directory ends up picking up changes from `online-upstream` somehow.

```
# make sure the cache is clear
zarf tools clear-cache


# Build all the things!
zarf package create examples/git-data/ --confirm       # This will result with the git repositories being added to the cache
zarf package create examples/git-data/ --confirm       # This will result with the cached repositories having unstaged changes
zarf package create examples/git-data/ --confirm       # This will result in a creation failure
```

Screen Shot 2022-12-04 at 6 09 15 PM

Racer159:
## Description

Fixes a bug introduced from MZAL where cache could be messed with do to GitPath being changed in the git receiver.

## Related Issue

Fixes #1079 

## Type of change

### DIFF
--- a/src/internal/packager/git/fetch.go
+++ b/src/internal/packager/git/fetch.go
@@ -20,7 +20,7 @@ func (g *Git) fetchTag(tag string) {
 
 	refspec := goConfig.RefSpec("refs/tags/" + tag + ":refs/tags/" + tag)
 
-	err := g.fetch(refspec)
+	err := g.fetch(g.GitPath, refspec)
 
 	if err != nil {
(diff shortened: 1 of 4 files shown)


jeff-mccoy:
Is it possible to include a test that demonstrates this behavior and that is now fixed?

Issue #1122, opened 2022-12-14
Use variables for Gitea's values


RothAndrew:
**Is your feature request related to a problem? Please describe.**
Gitea's configuration values are still hard coded

**Describe the solution you'd like**
Gitea's configuration values use variables

**Describe alternatives you've considered**
none

**Additional context**
https://github.com/defenseunicorns/zarf/pull/1117

RothAndrew:
I recommend making the default value for cpu limit some arbitrarily large number, like "10". [CPU limits tend to constrain resources when they don't need to be constrained](https://home.robusta.dev/blog/stop-using-cpu-limits).

RothAndrew:
Screenshot demonstrating Gitea being throttled during a large zarf package deployment even though there was more than enough CPU to go around on the node.

![image](https://user-images.githubusercontent.com/16000938/207657093-51d7f5cd-2002-473f-be77-e462965ed26d.png)

YrrepNoj:
I know you said you picked a random arbitrarily large number as a recommendation. I would suggest we use the same default value we use for the container registry, which in this case [is 3](https://github.com/defenseunicorns/zarf/blob/3de6e411b7098f083a57fdee2c6d59230a8afcf9/packages/zarf-registry/zarf.yaml#L25).

Racer159:
I agree with Jeff's comments on the Variablization of the gitea values, it would be (relatively) trivial to make that happen.

RothAndrew:
Jeff's comment for context: https://github.com/defenseunicorns/zarf/pull/1117#pullrequestreview-1216941666

> I think we should follow the [pattern of the registry now](https://github.com/defenseunicorns/zarf/blob/main/packages/zarf-registry/zarf.yaml?rgh-link-date=2022-12-14T07%3A52%3A48Z#L15) and use variables to give users control of this rather than making an arbitrary very high CPU limit.

PR #1123, opened 2022-12-14
Increase Gitea's CPU limit

RothAndrew:
## Description

Updates the CPU limit for Gitea from "1" to "10". "10" is an arbitrarily large number that doesn't mean much other than that it is likely bigger than Gitea will ever need.

[CPU limits tend to constrain resources when they don't need to be constrained](https://home.robusta.dev/blog/stop-using-cpu-limits)

Until #1122 is merged this works as a decent halfway measure that will resolve the issue of Gitea being throttled unnecessarily. It's not the final solution, but it's a net positive so we shouldn't wait for #1122 to do this.

## Related Issue

#1122

## Type of change

## Checklist before merging

no new tests or docs needed. I can do a quick ADR if we feel it is needed that would address a generalized strategy for CPU limits.


RothAndrew:
Screenshot demonstrating Gitea being throttled even though there was plenty of CPU on the node

image

YrrepNoj:
I agree that the CPU limit should be increased, I also agree that a longer term solution for this is to [follow the pattern of the registry](https://github.com/defenseunicorns/zarf/blob/3de6e411b7098f083a57fdee2c6d59230a8afcf9/packages/zarf-registry/zarf.yaml#L23) and have this value be a variable that can be configured.


That being said, if we can come up with what we think the CPU Limit for Gitea should be, I think it makes perfect sense to get this PR merged in with whatever that value should be.


I recommend we use the same CPU limit we use for the container registry and set the limit to `3`.

RothAndrew:
changed value to 3

RothAndrew:
I'm going to wait a day or two to see if anyone that is new to development of Zarf picks up #1122 since I labeled it "good-first-issue". If nobody does I'll take care of it. I agree it is a fairly trivial change.

In the meantime, I'm not seeing any reason that this PR shouldn't merge in the interim, in accordance with the principles of DevOps best practices of making frequent small changes.

#1122 will be done as the "final" "good" solution. I'm making the argument that that should not block this from merging.

jeff-mccoy:
@RothAndrew, I think we can close this now; I opened #1133 with a default limit of 3 CPUs. We'll include it in the next release once approved. Thoughts?

RothAndrew:
Sounds like @gscallon picked up the issue for variables. Make sure you guys aren't doing the same work

jeff-mccoy:
Okay, not a problem; I'll let them know we already have it covered. I wanted to get this change in before the `0.23` release. I will go ahead and close this one.

PR #1135, opened 2022-12-16
Add SBOM Comparison Tool

Racer159:
## Description

This is an _experimental_ Zarf SBOM comparison tool.  Some folks have expressed interest in an easy way to compare zarf files across packages, and this lets them open json files from multiple places and see what was added, removed and changed.

## Type of change

## Checklist before merging


Madeline-UX:
@Racer159 Would you add a screen recording of how it works and what it looks like for my poor visually oriented mind?

Madeline-UX (changes requested):
I think for an MVP we can push with this one change
- move the export CSV button out of top nav. Recommend adding it top right above the table.
 
A few UX things I want to work on in future iterations
- SBOM tool navigation

Madeline-UX (dismissed):

crcady:
This has some sorta strange/not obvious behavior right now. When you deploy a package, `compare.html` isn't extracted into the `zarf-sboms` folder. If you open the viewer as described [here](https://docs.zarf.dev/docs/dashboard-ui/sbom-dashboard#viewing-sboms-when-deploying), you'll get a viewer that has a Compare Tool button, but that button will not work because `compare.html` doesn't exist. If you `zarf package inspect -s` then the `compare.html` file IS extracted into the `sboms` folder.

Racer159:
@crcady Thanks for testing! I pushed a _potential_ fix that will maintain the same behavior across all SBOM outputs but also want to loop in @Madeline-UX and @jeff-mccoy for their thoughts.  The basic gist with the last commit is that it makes it always show every file in the sbom dir (including jsons on deploy). I went this way because:

1. It would be technically more complicated to try to detect the existence of the compare.html file and remove the compare button if we wanted to exclude compare.html
2. If the compare tool were included without JSON files you couldn't easily use it (if the SBOM you wanted was in the package you would need to re-extract)

src/internal/packager/sbom/catalog.go
```diff
@@ -106,6 +106,12 @@ func Catalog(componentSBOMs map[string]*types.ComponentSBOM, tagToImage map[name
 		currImage++
 	}
 
+	if len(componentSBOMs) != 0 && len(tagToImage) != 0 {
```
jeff-mccoy, line 109: nit, style-wise we usually have used `len() < 1` vs `len() != 0`.

src/internal/packager/sbom/catalog.go
```diff
@@ -153,7 +159,7 @@ func (b *Builder) createImageSBOM(tag name.Tag) ([]byte, error) {
 	}
 
 	// Write the sbom to disk using the image tag as the filename
-	sbomFile, err := b.createSBOMFile("%s.json", tag.String())
+	sbomFile, err := b.createSBOMFile(fmt.Sprintf("%s.json", tag.String()))
```
jeff-mccoy, line 162: Nit. Most places in code we try to avoid `fmt.Sprintf()` inline, especially if it has to call a function too.

src/internal/packager/sbom/catalog.go
```diff
@@ -219,7 +225,7 @@ func (b *Builder) createFileSBOM(componentSBOM types.ComponentSBOM, component st
 	}
 
 	// Write the sbom to disk using the given name as the filename
-	sbomFile, err := b.createSBOMFile("%s.json", fmt.Sprintf("%s%s", componentPrefix, component))
+	sbomFile, err := b.createSBOMFile(fmt.Sprintf("%s%s.json", componentPrefix, component))
```
jeff-mccoy, line 228: Same comment on inline `fmt.Sprintf()`

src/internal/packager/sbom/viewer/viewer.js
```diff
@@ -45,55 +45,69 @@ function initData() {
...
-
+    });
 }
 
 function fileList(metadata) {
     if (metadata) {
         const list = (metadata.files || [])
             .map(file => {
-                return file.path || ''
+                return file.path || "";
```
jeff-mccoy, line 66: Not a big deal either way, but just curious if this was prettier or some linting changing these from single to double quotes? I assume this was since we introduced all the JS/TS config stuff with Zarf UI?

src/internal/packager/sbom/viewer/viewer.js
```diff
@@ -1,36 +1,36 @@
...
     ZARF_SBOM_LIST.sort().forEach(item => {
-        let selected = (url === item) ? 'selected' : '';
+        let selected = (url === item) ? "selected" : "";
         sbomSelector.add(new Option(item, item, selected, selected));
     });
 }
 
 function initData() {
-    const payload = ZARF_SBOM_DATA
+    const payload = ZARF_SBOM_DATA;
```
jeff-mccoy, line 23: Glad the `;` are being added, should they also be added in lines 1-7 if we're doing it?

src/internal/packager/sbom/viewer/viewer.js
```diff
@@ -1,36 +1,36 @@
...
+const modalFader = document.getElementById("modal-fader")
+const modalTitle = document.getElementById("modal-title")
+const modalContent = document.getElementById("modal-content")
+const artifactsTable = document.createElement("table")
 
-document.body.appendChild(artifactsTable)
+document.body.appendChild(artifactsTable);
 
 function initSelector() {
     window.location.href
```
jeff-mccoy, line 12: Know this isn't code you changed, but what the heck did we/I leave this here before?

src/internal/packager/sbom/viewer/template.gohtml
```diff
@@ -255,10 +259,18 @@
         <button class="modal-btn modal-hide" onclick="hideModal()">Close</button>
     </div>
     <div id="modal-fader"></div>
+
+    <div id="actions">
```
jeff-mccoy, line 263: Really glad we are adding this

src/internal/packager/sbom/viewer/compare.js
```diff
@@ -0,0 +1,173 @@
...
+    window.dt = new simpleDatatables.DataTable(dataTable, {
+        data,
+        perPage: 20,
+    });
+}
+
+function fileList(metadata) {
+    if (metadata) {
+        const list = (metadata.files || [])
+            .map(file => {
```
jeff-mccoy, line 124: Nit, but when I have very simple arrow functions like this i prefer to inline them:
`.map(file => file.path || "")`

src/internal/packager/sbom/viewer/compare.js
```diff
@@ -0,0 +1,173 @@
...
+function loadDataTable(artifacts, dataTable) {
+    const transformedData = artifacts.map(artifact => {
+        return [
+            diff(artifact.zarfDiff),
+            artifact.type,
+            artifact.name,
+            artifact.version,
+            fileList(artifact.metadata),
+            (artifact.metadata && artifact.metadata.description) || "-",
+            ((artifact.metadata && artifact.metadata.maintainer) || "-").replace(/\u003c(.*)\u003e/, `&nbsp;|&nbsp;&nbsp;<a href="mailto:$1">$1</a>`),
```
jeff-mccoy, line 92: this feels kind of gross, might be nice to at least bring the first part of the string into another variable maybe?

src/internal/packager/sbom/viewer/compare.js
```diff
@@ -0,0 +1,173 @@
+const sbomSelector = document.getElementById("sbom-selector")
```
jeff-mccoy, line 1: `;` at the end?

src/internal/packager/sbom/viewer/compare.js
```diff
@@ -0,0 +1,173 @@
...
+
+    let leftReader = new FileReader();
+    leftReader.readAsText(leftJson);
+
+    leftReader.onload = function() {
+        try {
+            let leftData = JSON.parse(leftReader.result);
+            const leftMap = {}
+            leftData.artifacts.map(artifact => {
+                if (!leftMap[artifact.name]) { leftMap[artifact.name] = {} }
```
jeff-mccoy, line 36: one-lining these probable aren't great, some other possible options (though I'm not certain what the JS engine efficiency is in variable assignment if re-assigning)

```javascript

// Same code, just wihtout linebreak
if (!leftMap[artifact.name]) {
  leftMap[artifact.name] = {};
}

// Same but shorten prop
let name = artifact.name
if (!leftMap[name]) {
  leftMap[name] = {};
}

// One-liner, but variable re-assingment
leftMap[artifact.name] = leftMap[artifact.name] || {};

// Use a helper function
ensure(leftMap, artifact.name);
function ensure(map, el) {
  if (!map[el]) {
    map[el] = {};
  }
}

```

src/internal/packager/sbom/viewer/compare.gohtml
```diff
@@ -0,0 +1,291 @@
...
+            </button>
+        </div>
+    </div>
+</body>
+
+<script type="text/javascript">
+    {{.CompareJS}}
+</script>
+
+</html>
```
jeff-mccoy, line 291: can't wait until we get all this in Zarf UI and kill this thing with 🔥

jeff-mccoy (changes requested):
left comments

src/internal/packager/sbom/viewer/viewer.js
Racer159: This was me trying to normalize it.  I'll actually change it to what the prettierrc says to do.

src/internal/packager/sbom/viewer/viewer.js
Racer159: I deleted it when I copied it but forgot to go back and clean it up.  Looked like a miscopy to me

src/internal/packager/sbom/viewer/compare.js
Racer159: Funny given you wrote that :wink:. Will change in both places

Racer159: We could also create a common.js file potentially to hold these things... I may do that real fast

src/internal/packager/sbom/catalog.go
Racer159: Changed to len() > 0

jeff-mccoy (commented):
code review all looks good, want to pull it down and play with the flow a little before approval. Will try to get to tonight.

jeff-mccoy:
I think we need to add some documentation or explanation on this view. 

Screenshot 2023-01-09 at 12 20 16 PM

Racer159:
> I think we need to add some documentation or explanation on this view.
> 
> Screenshot 2023-01-09 at 12 20 16 PM

Added some docs:
image

jeff-mccoy (approved):

jeff-mccoy:
g2g!

Issue #1022, opened 2022-11-17
Zarf Init =      ERROR:  Please ensure this hostname is valid according to https://www.ietf.org/rfc/rfc1123.txt.


n8ores:
Running Zarf Install process found on Zarf homepage, in WSL Ubuntu and getting the following error when attempting to do a Zarf Init. 

![image](https://user-images.githubusercontent.com/5061384/202579325-697bfa8e-04bb-455a-94b8-c018aa62225d.png)

  GNU nano 4.8                          /tmp/zarf-2022-11-17-22-06-47-192135821.log

Log file output
```
^[[33mSaving log file to /tmp/zarf-2022-11-17-22-06-47-192135821.log^[[0m
^[[33m^[[0m^[[30;100m^[[30;100m  DEBUG  ^[[0m^[[0m ^[[90m^[[90m2022-11-17T22:06:47Z  -  k8s.generateLogShim()^[[0m^[[0m>^[[90m└ (/home/runner/work/zarf/zarf/src/internal/k8s/common.go:157)^[[0m
^[[90m^[[0m^[[30;100m^[[30;100m  DEBUG  ^[[0m^[[0m ^[[90m^[[90m2022-11-17T22:06:47Z  -  Log level set to info^[[0m^[[0m>^[[90m└ (/home/runner/work/zarf/zarf/src/cmd/root.go:87)^[[0m
^[[90m^[[0m^[[30;100m^[[30;100m  DEBUG  ^[[0m^[[0m ^[[90m^[[90m2022-11-17T22:06:47Z  -  utils.GetExecutablePath()^[[0m^>^[[90m└ (/home/runner/work/zarf/zarf/src/internal/utils/io.go:146)^[[0m
^[[90m^[[0m^[[30;100m^[[30;100m  DEBUG  ^[[0m^[[0m ^[[90m^[[90m2022-11-17T22:06:47Z  -  packager.Deploy()^[[0m^[[0m^[[9>^[[90m└ (/home/runner/work/zarf/zarf/src/internal/packager/deploy.go:34)^[[0m
^[[90m^[[0m^[[30;100m^[[30;100m  DEBUG  ^[[0m^[[0m ^[[90m^[[90m2022-11-17T22:06:47Z  -  Using temp path: '/tmp/zarf-330>^[[90m└ (/home/runner/work/zarf/zarf/src/internal/packager/common.go:46)^[[0m
[...]

Issue #939, opened 2022-10-24
Zarf should manage IaC as a first class citizen


runyontr:
**Is your feature request related to a problem? Please describe.**
When deploying a full infra stack the wrapper around terraform creates confusion and doesn't integrate with the Kubernetes management zarf provides

**Describe the solution you'd like**

* IaC treated natively
* Integration between zarf state of IaC and zarf state in Kubernetes and state used for IaC tools
* Zarf deployment of Kubernetes would auto-zarf-init the cluster, enabling ease of use of zarf to manage the Kubernetes environment, or maybe install the zarf controller?
* Allow outputs from terrafrom be inputs to zarf packages

**Describe alternatives you've considered**
Zarf not integrating natively and having to tie the pieces together manually or with separate zarf packages, e.g. via a Plugin system.

**Additional context**

RothAndrew:
Recommend we break this down more into some specific use cases that have been identified as difficult without some of the capabilities that zarf brings to the table, as well as identifying the right persona. Are we going after someone who is trying to do IaC in an air gap or in a cloud environment? Depending on which, this could go in very different directions.

What problems are people facing now, using existing tools that are out there, that zarf may solve?

Additionally, I feel it is important as we work this to ensure we preserve the separation of concerns between different layers of the stack. Keeping the infra layer, the platform layer, and the app layer separate is, IMO, a best practice that should be maintained. They each have different forces at play, different upgrade cycles, different permissions boundaries, etc

Racer159:
Agree on having a use case to test against, and I think @jeff-mccoy was going to work a bit on what a plugin system might look like once the library PR is complete and I think @UncleGedd had some thoughts as well. We could probably pull something together from them to bounce things off of.

UncleGedd:
> What problems are people facing now, using existing tools that are out there, that zarf may solve?

Context from CMS (note we are in a connected, yet regulated, environment): I believe a potential use case for Zarf is in our Batcave platform delivery model (Batcave is Big Bang + third party apps + Terrform/Terragrunt). One of the potential delivery models is to hand the platform IaC over to app teams and have them deploy Batcave themselves. Instead of handing over the entire codebase, I think it could be better to give app teams an immutable, compiled Zarf package.

In this case, Zarf would need to run the Terraform (via Terragrunt) and then deploy the cluster. To achieve adoption though, Zarf needs a value add thats better than simply running the Terraform commands (or writing a script to run the Terraform commands).

[...]

PR #1144, opened 2022-12-27
Scripts -> Actions

### LINKED ISSUE #1137: Zarf Component Chart Values/Outputs
As a package developer, I would like to be able to reference objects that were created by previous components. 

For example, as part of #939, the ability to define a terraform component that creates cloud resources for an application, e.g. an RDS database and then inject the credentials for that database correctly into the application as one component, would provide the ability for a Zarf package creator to declare how the application should be run in production.

Explicitly, when deploying a terraform module like this,

```hcl
# This file would live in the ./terraform/ folder so the foobar component can call it correctly

resource "random_integer" "sample" {
  min = 1
  max = 50000
}

output "number" {
  value = random_integer.sample.result
}
```

I would like to be able to use the outputs from the terraform module.  

```yaml
kind: ZarfPackageConfig
metadata:
  name: outputs
  description: 'Test component to demonstrate how outputs from terraform modules can be used'


components:
  - name: foobar
    description: "Sample terraform that creates things that we'll need later"
    required: true
    terraform:
      module: ./terraform/

  - name: script
    description: "This is a zebra and they have ###ZARF_COMPONENT_FOOBAR_TF_OUTPUT_NUMBER###"
```

In this example, the component named `foobar` has terraform outputs available that get templated like variables.  This could be done via a new type of variable outlined by `###ZARF_COMPONENT_<NAME>_<TYPE>_OUTPUT_<OUTPUTNAME>###`.  

Another option could be to add definitions to have the outputs be saved as a variable with a definition like:

```yaml
  - name: foobar
    description: "Sample terraform that creates things that we'll need later"
    required: true
    terraform:
      module: ./terraform/
      output:
        baz: number

  - name: script
    description: "This is a zebra and they have ###ZARF_PKG_VAR_BAZ###"
```

Where the variable `baz` is assigned the output of the `number` output from the terraform module and then treated like a variable.

jeff-mccoy:
## Related Issues
Related to #939 
Fixes #1137, #1022

### Backwards compatible replacement of  simple `scripts` with complex `actions`
- [x] Define types
- [x] `package create` transparent migration
- [x] `package deploy` transparent migration
- [x] Introduce deprecation pattern into `src/pkg/packager`
- [x] Package`Create` actions: `Before`, `After`, `Success`, `Failure`
- [x] Package`Deploy` actions: `Before`, `After`, `Success`, `Failure` 
- [x] Package`Remove` actions: `Before`, `After`, `Success`, `Failure`
  
### Other refactors
- [x] Expand exec to support env variables, add config struct
- [x] Add common cmd defaults
- [x] Add set working dir support 
- [x] Inject deploy-time variables as env variables for `onDeploy`
- [x] Expand sha256sum function to support any generic crypto hash type


jeff-mccoy:
Automatic conversion of `scripts` -> `actions` on `zarf package create` or `zarf package deploy`
Screenshot 2022-12-26 at 11 03 31 PM

jeff-mccoy:
Added basic mermaid to show the cycle of component actions
Screenshot 2023-01-03 at 11 01 41 PM

jeff-mccoy:
It's ugly for now, but does warn the users.
Screenshot 2023-01-05 at 2 14 43 AM

examples/component-actions/README.md
```diff
@@ -0,0 +1,94 @@
+# Component Actions
+
+:::note
+
+Component Actions have replaced Component Scripts. Zarf will still read scripts entries, but will convert them to actions. This will be removed in a future release.
```
Racer159, line 5:
```suggestion
Component Actions have replaced Component Scripts. Zarf will still read scripts entries, but will convert them to actions. Component Scripts will be removed in a future release.
```

More clear on what "this" is

src/pkg/packager/actions.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	return cmd, nil
+}
+
+// Merge the actionset defaults with the action config
+func actionGetCfg(actionSet types.ZarfComponentActionSet, a types.ZarfComponentAction) types.ZarfComponentActionDefaults {
+	cfg := actionSet.Defaults
+
+	if !a.Mute {
```
Racer159, line 117: There is a problem here with relying simply on boolean operations (at least for what I would expect).  Consider:

1. If `a.Mute` is not specified if will default to the zero value of the boolean in Go (`false`) and will always override the default `cfg.Mute`
2. If `a.Mute` is `true` it will never override the default `cfg.Mute` since `!true` is false

Racer159: This may need to be a nillable field.

src/pkg/packager/actions.go
```diff
@@ -0,0 +1,165 @@
...
+		cfg.Mute = a.Mute
+	}
+
+	// Default is no timeout, but add a timeout if one is provided
+	if a.MaxTotalSeconds > 0 {
```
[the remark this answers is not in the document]
Racer159: The same would go for all of the other values.

src/pkg/packager/actions.go
```diff
@@ -0,0 +1,165 @@
...
+		)
+
+		cfg := actionGetCfg(actionSet, a)
+
+		if cmd, err = actionCmdMutation(a.Cmd); err != nil {
+			spinner.Errorf(err, "Error mutating command: %s", cmd)
+		}
+
+		// Otherwise, run the command with a timeout handler
+		spinner.Updatef("Waiting for command \"%s\" (timeout: %d seconds)", cmd, cfg.MaxTotalSeconds)
```
Racer159, line 42: On slower machines this may briefly flash a 0 second timeout message, then update to no timeout (would be better to set the messages in the if IMO.

src/pkg/packager/actions.go
```diff
@@ -0,0 +1,165 @@
...
+
+		duration := time.Duration(cfg.MaxTotalSeconds) * time.Second
+		timeout := time.After(duration)
+
+		// Keep trying until the max retries is reached
+		for remaining := cfg.MaxRetries + 1; remaining > 0; remaining-- {
```
Racer159, line 48: This will add 1 to the maxRetries giving one more than you actually expect.  If you wanted 0 to be equal to one then you would need to do something like this:

```suggestion
		for remaining := cfg.MaxRetries == 0 ? 1 : cfg.MaxRetries; remaining > 0; remaining-- {
```

src/pkg/packager/actions.go
```diff
@@ -0,0 +1,165 @@
...
+
+			// If no timeout is set, run the command and return
+			if cfg.MaxTotalSeconds < 1 {
+				spinner.Updatef("Waiting for command \"%s\" (no timeout)", cmd)
+
+				// Try running the command and return if it fails
+				if err := actionRun(context.TODO(), cfg, cmd); err != nil {
+					return fmt.Errorf("command \"%s\" failed: %w", cmd, err)
```
Racer159, line 56: This should continue the retry loop if it fails.

src/pkg/packager/compose.go
```diff
@@ -48,22 +51,22 @@ func (p *Packager) composeComponents() error {
 func (p *Packager) getComposedComponent(parentComponent types.ZarfComponent) (child types.ZarfComponent, err error) {
 	message.Debugf("packager.GetComposedComponent(%+v)", parentComponent)
 
-	// Make sure the component we're trying to import cant be accessed
+	// Make sure the component we're trying to import cant be accessed.
```
Racer159, line 54:
```suggestion
	// Make sure the component we're trying to import can't be accessed.
```

src/pkg/packager/deprecated/scripts-to-actions.go
```diff
@@ -0,0 +1,68 @@
...
+	defaults := types.ZarfComponentActionDefaults{
+		// ShowOutput (default false) -> Mute (default false)
+		Mute: !c.DeprecatedScripts.ShowOutput,
+		// TimeoutSeconds -> MaxSeconds
+		MaxTotalSeconds: c.DeprecatedScripts.TimeoutSeconds,
+	}
+
+	// Retry is now an integer vs a boolean (implicit infinite retries), so set to an absurdly high number
+	if c.DeprecatedScripts.Retry {
+		defaults.MaxRetries = 9999
```
Racer159, line 32: do we want to use something like `math.MaxInt32` instead of a magic number?

src/pkg/packager/remove.go
```diff
@@ -36,71 +36,89 @@ func (p *Packager) Remove(packageName string) error {
...
-					return err
-				}
+
+		// Remove the component we just removed from the array
+		packages.DeployedComponents = utils.RemoveMatches(packages.DeployedComponents, func(t types.DeployedComponent) bool {
+			return t.Name == c.Name
+		})
+
+		if len(packages.DeployedComponents) < 1 {
+			// All the installed components were deleted, there for this package is no longer actually deployed
```
Racer159, line 69:
```suggestion
			// All the installed components were deleted, therefore this package is no longer actually deployed
```

src/test/e2e/02_component_actions_test.go
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2021-Present The Zarf Authors
 
-// Package test provides e2e tests for Zarf.
+// Package test provides e2e tests for zarf
```
Racer159, line 4: Zarf should be caps here?

src/test/e2e/02_component_actions_test.go
```diff
@@ -11,45 +11,45 @@ import (
...
 
-	// Try creating the package to test the create scripts
-	stdOut, stdErr, err := e2e.execZarfCommand("package", "create", "examples/component-scripts", "--confirm")
+	// Try creating the package to test the onCreate actions
+	stdOut, stdErr, err := e2e.execZarfCommand("package", "create", "examples/component-actions", "--confirm")
 	require.NoError(t, err, stdOut, stdErr)
 
-	// Test for package create prepare artifact
-	require.FileExists(t, prepareArtifact)
+	// Test for package create prepare artirfact
```
Racer159, line 33:
```suggestion
	// Test for package create prepare artifact
```

src/pkg/packager/actions.go
Racer159: NVM on this.  I was thinking about the word "retries" wrong...

Racer159 (changes requested):
Made some comments on things that could be updated.

src/pkg/packager/actions.go
```diff
@@ -0,0 +1,195 @@
...
+	"time"
+
+	"github.com/defenseunicorns/zarf/src/internal/packager/template"
+	"github.com/defenseunicorns/zarf/src/pkg/message"
+	"github.com/defenseunicorns/zarf/src/pkg/utils/exec"
+	"github.com/defenseunicorns/zarf/src/types"
+)
+
+// Run commands that a component has provided.
+func (p *Packager) runAction(actionSet types.ZarfComponentActionSet, actions []types.ZarfComponentAction, valueTemplate *template.Values) error {
```
YrrepNoj, line 23: IMO it looks a lot clear to replace the `actionSet types.ZarfComponentActionSet` parameter with `actionSetDefaults types.ZarfComponentActionDefaults`.

That's all we're using the actionSet for and that way we're passing a struct and one of its own member fields to a singular function 😛

src/pkg/packager/actions.go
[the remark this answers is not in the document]
YrrepNoj: For this, we could treat negative numbers as 'infinity'. You could set the default to 60 seconds while seeding specific actions to `-1` to indicate that the specific action will ignore the default and run until completed.

src/pkg/packager/actions.go
```diff
@@ -0,0 +1,195 @@
...
+					continue
+				}
+
+				// If an output variable is defined, set it.
+				if a.SetVariable != "" {
+					p.setVariable(a.SetVariable, out)
+				}
+
+				// If the command ran successfully, continue to the next action.
+				continue ACTION
```
YrrepNoj, line 73: I'm pretty sure I'm missing something, but this looks like it will continue to outside of the original `for` loop, starting over from the beginning of the list of actions.

I'm just misunderstanding how continue's with labels work right?

YrrepNoj: After a little googling it looks like I did misunderstand labels 🤷 


Would you be opposed to splitting `runAction()` into an additional function of `runActions()` that way we can return from a function instead of doing this continue-to-label thing?

YrrepNoj (dismissed):

YrrepNoj:
A random thought that I'm not sure we've considered yet..

At the moment we batch up all the images that the components declare so that they are all pulled at the same time during the later stage of package creation. Depending on how specific someone is trying to be with a `ComponentAction.OnCreate` command that might cause a potential issue b/c that isn't intuitive unless you've looked through the code.

src/types/component.go
```diff
@@ -107,13 +110,48 @@ type ZarfComponentScripts struct {
 	After          []string `json:"after,omitempty" jsonschema:"description=Scripts to run after the component successfully deploys"`
 }
 
-// ZarfContainerTarget defines the destination info for a ZarfData target.
+// ZarfComponentActions are actionsets that map to different zarf package operations
+type ZarfComponentActions struct {
+	OnCreate ZarfComponentActionSet `json:"onCreate,omitempty" jsonschema:"description=Actions to run during package creation"`
```
YrrepNoj, line 115: A random thought that I'm not sure we've considered yet..

At the moment we batch up all the images that the components declare so that they are all pulled at the same time during the later stage of package creation. Depending on how specific someone is trying to be with a ComponentAction.OnCreate command that might cause a potential issue b/c that isn't intuitive unless you've looked through the code.

src/test/e2e/common.go
```diff
@@ -45,7 +45,7 @@ func (e2e *ZarfE2ETest) setup(t *testing.T) {
 	t.Log("Test setup")
 	// Output list of allocated cluster resources
 	if runtime.GOOS != "windows" {
-		_, _, _ = utils.ExecCommandWithContext(context.TODO(), true, "sh", "-c", "kubectl describe nodes |grep -A 99 Non\\-terminated")
+		_ = exec.CmdWithPrint("sh", "-c", "kubectl describe nodes |grep -A 99 Non\\-terminated")
```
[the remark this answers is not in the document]
jeff-mccoy: some of the tests need to pass in context for a timeout, but others, such as `exec.CmdWithContext(context.TODO(), exec.PrintCfg(), e2e.zarfBinPath, commandString...)` actually don't, and can be changed to  `exec.CmdWithPrint`. I don't have strong opinions either way and maybe having both ways is odd, no strong feelings either way.

src/pkg/utils/hash.go
```diff
@@ -1,20 +1,20 @@
...
 	"io"
 	"os"
 
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 )
 
-// GetSha256Sum returns the computed SHA256 Sum of a given file.
-func GetSha256Sum(path string) (string, error) {
+// GetCryptoHash returns the computed SHA256 Sum of a given file
+func GetCryptoHash(path string, hashName crypto.Hash) (string, error) {
```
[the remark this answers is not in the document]
jeff-mccoy: it was a holdover from before I broke the PR up, it just changed too many files for me to unwind it happily.

src/config/lang/english.go
```diff
@@ -202,6 +202,8 @@ const (
 
 // src/internal/packager/validate.
 const (
+	PkgValidateMustBeUppercase            = "variable name '%s' must be all uppercase and contain no special characters except _"
```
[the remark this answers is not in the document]
jeff-mccoy: Yeah I wasn't sure, it's one I want to reuse in a couple places, but feel free to change to what makes sense, 0 preference.

packages/distros/eks/zarf.yaml
```diff
@@ -7,14 +7,15 @@ metadata:
...
-        # Remove existing eksctl
-        - "rm -f eksctl"
-        # Extract the correct linux or mac binary from the tarball
-        - "./zarf tools archiver decompress archives/eksctl_$(uname -s)_$(uname -m).tar.gz ."
-        # Cleanup temp files
-        - "rm -fr archives"
+    actions:
+      onDeploy:
+        after:
+          # Remove existing eksctl
```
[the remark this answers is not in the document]
jeff-mccoy: Not sure I understand the q

examples/terraform/zarf.yaml
```diff
@@ -24,9 +25,11 @@ components:
         target: tmp/terraform_1.3.1_darwin_arm64.zip
 
   - name: execute-terraform
-    scripts:
-      timeoutSeconds: 60
-      showOutput: true
```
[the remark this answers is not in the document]
jeff-mccoy: this pr changes make the default show output == true (effectively), which was intentional as we've found users tend to want more running feedback vs silent magic. Keep this default of `mute:false` follows that same thinking.

src/internal/packager/validate/validate.go
```diff
@@ -119,6 +125,44 @@ func validateComponent(pkg types.ZarfPackage, component types.ZarfComponent) err
...
 	}
 
+	if err := validateOnDeployActions(component.Actions.OnDeploy); err != nil {
+		return fmt.Errorf(lang.PkgValidateErrAction, err)
+	}
+
+	return nil
+}
+
+func validateOnDeployActions(actions types.ZarfComponentActionSet) error {
```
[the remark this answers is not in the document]
jeff-mccoy: Yeah that's a good point I hadn't really thought of; technically we could have runtime variables at those stages--but the other variables are deploy-time only and won't exist. This is probably useful to keep--but we need to document that better if we do.

src/pkg/packager/variables.go
```diff
@@ -82,13 +82,17 @@ func (p *Packager) setActiveVariables() error {
...
-			p.cfg.SetVariableMap[variable.Name] = val
+			p.setVariable(variable.Name, val)
 		}
 	}
 
 	return nil
 }
 
+func (p *Packager) setVariable(name, value string) {
+	p.cfg.SetVariableMap[name] = strings.TrimSpace(value)
```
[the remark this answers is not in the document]
jeff-mccoy: Leading and trailing whitespace is problematic in string templating in general, but very obviously breaks when reading values from an exec output as there is usually a line break at the end of the output.

src/pkg/packager/actions.go
jeff-mccoy: Yeah that seems reasonable, I thought I might use more than that initially, but looks like I didn't.

src/pkg/packager/actions.go
jeff-mccoy: no that's fine, this just evolved into that and wasn't broken up better.

src/types/component.go
jeff-mccoy: I think that's a good point, we moved to a single image collection for binary efficiencies (reuse of layer refs in a single tarball), but iirc it wasn't significant. It probably would make more sense to do this per component if it's not a major loss in efficiency.

adr/0010-scripts-actions.md
```diff
@@ -0,0 +1,43 @@
...
+- `defaults` - action configuration that will be applied to all actions in the set
+- `before` - array of actions that will run before the component is processed for `create`, `deploy`, or `remove`
+- `after` - array of actions that will run after the component processed for `create`, `deploy`, or `remove` successfully
+- `onSuccess` - array of actions that will run after any `after` actions have successfully completed
+- `onFailure` - array of actions that will run after any error during the above actions or component operations
+
+Within each `before`, `after`, `onSuccess`, and `onFailure` action set, the following configurations are available:
+
+- `cmd` - (required) the command to run
+- `dir` - the directory to run the command in
```
runyontr, line 32: defaults to the folder with the zarf.yaml?

adr/0010-scripts-actions.md
```diff
@@ -0,0 +1,43 @@
...
+- `onCreate` - Runs during `zarf package create`
+- `onDeploy` - Runs during `zarf package deploy`
+- `onRemove` - Runs during `zarf package remove`
+
+Each of these action sets are optional and contain different hooks and events that can be used to customize the package lifecycle:
+
+- `defaults` - action configuration that will be applied to all actions in the set
+- `before` - array of actions that will run before the component is processed for `create`, `deploy`, or `remove`
+- `after` - array of actions that will run after the component processed for `create`, `deploy`, or `remove` successfully
+- `onSuccess` - array of actions that will run after any `after` actions have successfully completed
```
runyontr, line 26: after *any* of them have completed, or all of them have completed and returned a 0 exit code?

does this run after just `after` or after `before` is successful as well.

adr/0010-scripts-actions.md
```diff
@@ -0,0 +1,43 @@
...
+- `onFailure` - array of actions that will run after any error during the above actions or component operations
```
runyontr, line 27: error code being the return of a non-zero exit code for any command either the before or after?  or just after?

adr/0010-scripts-actions.md
```diff
@@ -0,0 +1,43 @@
...
+
+The `scripts` section of the `zarf.yaml` will be replaced with a new `actions` section. The `actions` section will be a map of action names to a list of commands to run. `actions` will contain action sets that map to the following lifecycle events:
+
+- `onCreate` - Runs during `zarf package create`
+- `onDeploy` - Runs during `zarf package deploy`
+- `onRemove` - Runs during `zarf package remove`
+
+Each of these action sets are optional and contain different hooks and events that can be used to customize the package lifecycle:
+
+- `defaults` - action configuration that will be applied to all actions in the set
```
runyontr, line 23: not sure what `configuration` means

adr/0010-scripts-actions.md
```diff
@@ -0,0 +1,43 @@
...
+
+Within each `before`, `after`, `onSuccess`, and `onFailure` action set, the following configurations are available:
+
+- `cmd` - (required) the command to run
+- `dir` - the directory to run the command in
+- `mute` - whether to mute the output of the command (default: `false`)
+- `maxTotalSeconds` - the maximum total time to allow the command to run (default: `0` - no limit)
+- `maxRetries` - the maximum number of times to retry the command if it fails (default: `0` - no retries)
+- `env` - an array of environment variables to set for the command in the form of `name=value`
+- `setVariable` - set the output of the command to a variable that can be used in other actions or components
```
runyontr, line 37: These commands run after the cmd is executed in this block?

adr/0010-scripts-actions.md
```diff
@@ -0,0 +1,43 @@
...
+
+Date: 2023-01-18
+
+## Status
+
+Accepted
+
+## Context
+
+Originally, the `scripts` noun was added to components to allow us to move hard-coded init business logic out of the codebase and into the package system. At the time there was only a `before` and `after` section with simply an array entry per command. Later, `prepare` was added as a way to do some thing during `zarf package create`. As teams began to find new ways to use the capabilities, their limitations became more obvious.
```
runyontr, line 11: I think something about what's being done with scripts is important.  It feels like we're saying we're replacing scripts, but no comment about how to update those nouns within a zarf.yaml or what it looks like for dealing with zarf.yaml versions from earlier versions of Zarf.

adr/0010-scripts-actions.md
runyontr: Does it handle onFailure for other components not succeeding?

adr/0010-scripts-actions.md
```diff
@@ -0,0 +1,43 @@
...
+- `before` - array of actions that will run before the component is processed for `create`, `deploy`, or `remove`
+- `after` - array of actions that will run after the component processed for `create`, `deploy`, or `remove` successfully
+- `onSuccess` - array of actions that will run after any `after` actions have successfully completed
+- `onFailure` - array of actions that will run after any error during the above actions or component operations
+
+Within each `before`, `after`, `onSuccess`, and `onFailure` action set, the following configurations are available:
+
+- `cmd` - (required) the command to run
+- `dir` - the directory to run the command in
+- `mute` - whether to mute the output of the command (default: `false`)
```
runyontr, line 33: should this return logs when it errors regardless of it being muted for being successful?

adr/0010-scripts-actions.md
```diff
@@ -0,0 +1,43 @@
...
+The `scripts` section of the `zarf.yaml` will be replaced with a new `actions` section. The `actions` section will be a map of action names to a list of commands to run. `actions` will contain action sets that map to the following lifecycle events:
+
+- `onCreate` - Runs during `zarf package create`
+- `onDeploy` - Runs during `zarf package deploy`
+- `onRemove` - Runs during `zarf package remove`
+
+Each of these action sets are optional and contain different hooks and events that can be used to customize the package lifecycle:
+
+- `defaults` - action configuration that will be applied to all actions in the set
+- `before` - array of actions that will run before the component is processed for `create`, `deploy`, or `remove`
```
runyontr, line 24: I think using the term `the component` is a little odd since the action is part of the component.  maybe a more global phrase like "the other parts of the component" or "non action components" or something to make it clear that its the same component the action is defined in.

src/types/component.go
YrrepNoj: Agreed. 

I think this can be done in a separate PR so that we don't grow this one to be too big.

adr/0010-scripts-actions.md
jeff-mccoy: Yes I guess we should clarify that, it will still drop the error on failure.

adr/0010-scripts-actions.md
jeff-mccoy: these are configs for the action, which requires a `cmd` and is optionally configured with the other keys

adr/0010-scripts-actions.md
jeff-mccoy: at any point one of the actions returns a non-zero of if a stage in the component lifecycle, i.e. images fails, it will call `onFailure`

adr/0010-scripts-actions.md
```diff
@@ -0,0 +1,43 @@
+# 10. Scripts -> Actions
```
runyontr, line 1: See #1223 to maybe change this number to 0011 for when that lands

adr/0010-scripts-actions.md
jeff-mccoy: that is actually covered in the consequences section, where I think it is most approrpiate

src/pkg/packager/remove.go
Racer159:
```suggestion
			// All the installed components were deleted, therefore this package is no longer actually deployed
```
Not sure why this was marked resolved.  `there for` is still in the comment.

src/pkg/packager/remove.go
```diff
@@ -36,60 +36,35 @@ func (p *Packager) Remove(packageName string) error {
...
-				if err != nil {
-					message.Errorf(err, "Unable to remove the installed helm chart (%s) from the namespace (%s) of component (%s) (were dependent components removed first?)",
-						installedChart.ChartName, installedChart.Namespace, installedComponent.Name)
-
-					return err
-				}
-
-				// Remove the uninstalled chart from the list of installed charts
-				deployedPackage.DeployedComponents[i].InstalledCharts = deployedPackage.DeployedComponents[i].InstalledCharts[:h]
-				p.updatePackageSecret(deployedPackage, secretName)
```
Racer159, line 86: This functionality does not exist in the new logic.  If a component has multiple helm charts, uninstalls the first one successfully and fails on the second, zarf's state will now be bad and will perma-fail on the now ghosted first chart.

src/pkg/packager/remove.go
```diff
@@ -114,3 +89,44 @@ func (p *Packager) updatePackageSecret(deployedPackage types.DeployedPackage, se
 		message.Warnf("Unable to update the %s package secret: %#v", secretName, err)
 	}
 }
+
+func (p *Packager) removeComponent(components []types.ZarfComponent, deployedComponent types.DeployedComponent, spinner *message.Spinner) error {
+	c := utils.Find(components, func(t types.ZarfComponent) bool {
```
Racer159, line 94: Because `OnRemove` is new functionality I would rather add `Actions.OnRemove` to `types.DeployedComponent`.  A missing key should initialize to the zero value (empty array) which would be fine for any existing deployed components without `OnRemove` actions.

src/pkg/packager/remove.go
```diff
@@ -114,3 +89,44 @@ func (p *Packager) updatePackageSecret(deployedPackage types.DeployedPackage, se
...
+		}
+	}
+
+	if err := p.runActions(onRemove.Defaults, onRemove.After, nil); err != nil {
+		onFailure()
+		return fmt.Errorf("unable to run the after action: %w", err)
+	}
+
+	if err := p.runActions(onRemove.Defaults, onRemove.OnSuccess, nil); err != nil {
+		onFailure()
```
Racer159, line 127: We run onFailure OnSuccess failure here :wink:

src/pkg/packager/variables.go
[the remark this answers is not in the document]
Racer159: Can we only do this on runtime set variables?  This could be problematic if someone needed something to start/end with a space.

src/test/e2e/common.go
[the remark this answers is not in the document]
Racer159: Personally I feel if we have the helper we should use the simplest form of it for consistency's sake.

src/test/e2e/02_component_actions_test.go
```diff
@@ -0,0 +1,80 @@
...
+	allArtifacts = append(allArtifacts, deployWithEnvVarArtifact)
+	e2e.cleanFiles(allArtifacts...)
+	defer e2e.cleanFiles(allArtifacts...)
+
+	/* Create */
+	// Try creating the package to test the onCreate actions.
+	stdOut, stdErr, err := e2e.execZarfCommand("package", "create", "examples/component-actions", "--confirm")
+	require.NoError(t, err, stdOut, stdErr)
+
+	// Test for package create prepare artifact.
```
Racer159, line 40:
```suggestion
	// Test for package create prepare artifacts.
```

src/test/e2e/02_component_actions_test.go
```diff
@@ -0,0 +1,80 @@
...
+	stdOut, stdErr, err = e2e.execZarfCommand("package", "deploy", path, "--confirm", "--components=on-deploy-with-variable", "-l=trace")
+	require.NoError(t, err, stdOut, stdErr)
+	require.Contains(t, stdOut, "the dog says ruff")
+
+	// Test using dynamic and multiple-variables
+	stdOut, stdErr, err = e2e.execZarfCommand("package", "deploy", path, "--confirm", "--components=on-deploy-with-dynamic-variable,on-deploy-with-multiple-variables", "-l=trace")
+	require.NoError(t, err, stdOut, stdErr)
+	require.Contains(t, stdOut, "the cat says meow")
+	require.Contains(t, stdOut, "the dog says ruff")
+	require.Contains(t, stdOut, "the snake says hiss")
```
Racer159, line 74: Did we want to explicitly test the TF_VAR stuff?  changing the package to be prefixed with `the terraform snake...` or similar would be an easy way to validate that functionality.

src/test/e2e/02_component_actions_test.go
```diff
@@ -0,0 +1,80 @@
...
+	stdOut, stdErr, err = e2e.execZarfCommand("package", "deploy", path, "--confirm", "--components=on-deploy")
+
+	// Check that the deploy artifacts were created.
+	for _, artifact := range deployArtifacts {
+		require.FileExists(t, artifact)
+	}
+
+	// Deploy the simple action that should fail the timeout.
+	stdOut, stdErr, err = e2e.execZarfCommand("package", "deploy", path, "--confirm", "--components=on-deploy-with-timeout")
+	require.Error(t, err, stdOut, stdErr)
```
Racer159, line 62: We need to check the output of this, not just that it failed to properly test OnFailure.

src/test/e2e/02_component_actions_test.go
```diff
@@ -0,0 +1,80 @@
...
+	stdOut, stdErr, err = e2e.execZarfCommand("package", "deploy", path, "--confirm", "--components=on-deploy-with-dynamic-variable,on-deploy-with-multiple-variables", "-l=trace")
+	require.NoError(t, err, stdOut, stdErr)
+	require.Contains(t, stdOut, "the cat says meow")
+	require.Contains(t, stdOut, "the dog says ruff")
+	require.Contains(t, stdOut, "the snake says hiss")
+
+	// Test using environment variables
+	stdOut, stdErr, err = e2e.execZarfCommand("package", "deploy", path, "--confirm", "--components=on-deploy-with-env-var")
+	require.NoError(t, err, stdOut, stdErr)
+	require.FileExists(t, deployWithEnvVarArtifact)
```
Racer159, line 79: We should add a test for the onRemove case as well.

src/test/e2e/02_component_actions_test.go
```diff
@@ -0,0 +1,80 @@
...
+	require.Contains(t, stdOut, "the dog says ruff")
+
+	// Test using dynamic and multiple-variables
+	stdOut, stdErr, err = e2e.execZarfCommand("package", "deploy", path, "--confirm", "--components=on-deploy-with-dynamic-variable,on-deploy-with-multiple-variables", "-l=trace")
+	require.NoError(t, err, stdOut, stdErr)
+	require.Contains(t, stdOut, "the cat says meow")
+	require.Contains(t, stdOut, "the dog says ruff")
+	require.Contains(t, stdOut, "the snake says hiss")
+
+	// Test using environment variables
```
Racer159, line 76: It would be nice to test the templated text file as well

Racer159 (changes requested):
Mostly spelling/grammar and test robustness.  There is also a linting error for a comment on `GetVariables` in `template.go`

src/test/e2e/02_component_actions_test.go
Racer159: Unresolving due to no comment.  If we include it and it is easy to test (which I believe it to be) then we should test it IMO.

src/test/e2e/02_component_actions_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Test to ensure the deploy scripts are not executed.
+	for _, artifact := range deployArtifacts {
+		require.NoFileExists(t, artifact)
+	}
+
+	/* Deploy */
+	path := fmt.Sprintf("build/zarf-package-component-actions-%s.tar.zst", e2e.arch)
+	// Deploy the simple script that should pass.
+	stdOut, stdErr, err = e2e.execZarfCommand("package", "deploy", path, "--confirm", "--components=on-deploy")
```
Racer159, line 53: stdOut, stdErr are not used here:
```suggestion
	_, _, err = e2e.execZarfCommand("package", "deploy", path, "--confirm", "--components=on-deploy")
```

src/test/e2e/02_component_actions_test.go
Racer159: Unresolving.  This should have something like this to fully test it:

```suggestion
	require.Error(t, err, stdOut, stdErr)
	require.Contains(t, stdOut, "this action failed because it took too long to run")
```

Racer159 (changes requested):
Adding back some changes

examples/component-actions/README.md
```diff
@@ -0,0 +1,11 @@
+# Component Actions
+
+:::note
+
+Component Actions have replaced Component Scripts. Zarf will still read scripts entries, but will convert them to actions. Component Scripts will be removed in a future release.
+
+:::
+
+This example demonstrates how to define actions within your package that can run either on `zarf package create`, `zarf package deploy` or `zarf package remove`. These actions will be executed with the context that the Zarf binary is executed with.
+
```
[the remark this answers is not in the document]
Racer159: This is related to another issue, and would be good practice going forward: https://github.com/defenseunicorns/zarf/issues/1215. I would argue not having a code snippet here could be confusing in the docs site.

src/test/e2e/02_component_actions_test.go
jeff-mccoy: we won't be able to easily test remove since that will require a cluster that won't exist yet.

src/test/e2e/02_component_actions_test.go
jeff-mccoy: lol or maybe be more patient, I was using resolutions to track the work

Racer159: It had been 9 hours 🤪. Just wanted to make sure it wasn't forgotten that's all.

src/test/e2e/02_component_actions_test.go
Racer159: We could add it to another test that could benefit from some remove behavior potentially. I do worry about not testing it because it is effectively a third of this PR (create, deploy, remove)

src/pkg/packager/deprecated/scripts-to-actions.go
```diff
@@ -0,0 +1,76 @@
...
+		hasScripts = true
+		c.Actions.OnDeploy.Defaults = defaults
+		for _, s := range c.DeprecatedScripts.After {
+			c.Actions.OnDeploy.After = append(c.Actions.OnDeploy.After, types.ZarfComponentAction{Cmd: s})
+		}
+
+		c.DeprecatedScripts.After = []string{}
+	}
+
+	// Leave deprecated scripts in place, but warn users
```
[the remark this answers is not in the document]
jeff-mccoy: left over from the change

src/test/e2e/02_component_actions_test.go
jeff-mccoy: I mean, it's a 3rd call of the majority same code, that's very different than "one-third of the PR". I'd strongly prefer to keep this test pre-cluster, maybe we could have a second actions test with-cluster or something. Not sure.

Racer159: Yeah that is what I was suggesting, and not 1/3 of code but 1/3 of functionality introduced (which could be taken out easily).

Racer159: I may be a little skittish but it is exactly that that completely disabled SBOMing and external registries in the past when simple if statements at higher levels got flipped.

Racer159: It wouldn't necessarily have to be its own test, but could be something we test as a side effect of another test (potentially just with a comment recording that).

Racer159: And case in point this functionality is partially broken (--set is not a thing on remove and it cannot / does not load variable defaults) 
image

Racer159: (here is that test case: https://github.com/defenseunicorns/zarf/tree/jeff-doing-stupid-jeff-things-remove-test)

Racer159: (differs slightly from the run above in that I added --set calls)

Racer159: (we could also store the vars used to deploy or explicitly not support variables here in the docs)

src/pkg/packager/remove.go
Racer159: mentioning in reference to not running it elasewhere.

src/pkg/packager/variables.go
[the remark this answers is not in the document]
YrrepNoj: I agree, do you have any objections @jeff-mccoy ?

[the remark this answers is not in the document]
jeff-mccoy: idk, it's not a document feature or omission at this point and I have a lot of concerns with allowing generic whitespace at the edge of variables--but I can accept moving it too.

src/test/e2e/02_component_actions_test.go
jeff-mccoy: in the case of create/remove the variables are only generated by actions, which I think makes sense.

jeff-mccoy: I see no problem with this behavior, though I suppose we could add more docs for that, too...even though defined variables are very clearly for package deploy only already

src/pkg/packager/remove.go
YrrepNoj: Pushed something out for this 👍

src/pkg/packager/remove.go
YrrepNoj: For my 2-cents;

I kind of like how the `types.Deployed{Package}/{Component}/{Chart}` all have the minimal amount of information necessary to know what is actually deployed for that type. At the moment it's purely state based information.

Adding a OnRemove lifecycle hook descriptor to components seems to be a bit overloady.

RothAndrew:
Does the schema file need to be updated?

image

Racer159:
> Does the schema file need to be updated?
> 
> Edit: no? I see `actions` in the schema file in this branch. Maybe I have something cached..
> 
> image

Not sure which IDE that is, but for VSCode at least it is in settings.json.  You also may need to restart your IDE after switching branches (at least VS Code requires a restart)

src/pkg/packager/remove.go
Racer159: I can see that.  I'll mark as resolved even though the loop still feels weird.

src/cmd/destroy.go
```diff
@@ -57,7 +57,7 @@ var destroyCmd = &cobra.Command{
 			// Iterate over all matching zarf-clean scripts and exec them
 			for _, script := range scripts {
 				// Run the matched script
-				_, _, err := utils.ExecCommandWithContext(context.TODO(), true, script)
+				err := exec.CmdWithPrint(script)
```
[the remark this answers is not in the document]
jeff-mccoy: So I actually tried to do this, but it gets super messy because you essentially need the original init package/scripts to do a proper `zarf destroy`. Keeping the scripts this way is actually more reliable for doing a final clean sweep. Also, since this function currently doesn't run package remove actions properly, I think it's out of scope to solve that problem in this PR too.

src/types/package.go
```diff
@@ -28,11 +28,12 @@ type ZarfMetadata struct {
...
-	User         string `json:"user"`
-	Architecture string `json:"architecture"`
-	Timestamp    string `json:"timestamp"`
-	Version      string `json:"version"`
+	Terminal     string   `json:"terminal"`
+	User         string   `json:"user"`
+	Architecture string   `json:"architecture"`
+	Timestamp    string   `json:"timestamp"`
+	Version      string   `json:"version"`
+	Migrations   []string `json:"migrations"`
```
YrrepNoj, line 36: When printing the zarf.yaml when deploying packages to get validation from the user that the deployment is correct the depreciated contents will be printed as well as the migrated contents. 

For this specific depreciation that might be ok (although still a little confusing), but I could guess that being less clear for some random potential future depreciation.

Thoughts on adding some utility function that simply removes the old deprecated content (if the package has already been migrated) to make the validation output make more sense?

I think this could be in a separate PR but this felt like a good place to start thinking about it.

jeff-mccoy: Adding a migration tool in a future pr makes sense to me. We already warn users about the deprecation when we show the yaml though, is that not enough?

YrrepNoj: We don't show that warning when deploying a package that has had deprecated features already migrated.

In the following screenshot shows me deploying a package, in t he `Deploy this Zarf package?` conformation I am being shown both the scripts and the actions but I'm not seeing a warning that there has been a migration b/c we skipped that process for deploy due to the migration tag existing in the build data.
Screen Shot 2023-01-24 at 11 10 10 AM

YrrepNoj: Someone could see this prompt and easily think the `touch test-deprecated-deploy-before-hook.txt` command is going to be run twice.

src/cmd/internal.go
```diff
@@ -113,6 +114,16 @@ var uiCmd = &cobra.Command{
 	},
 }
 
+var isValidHostname = &cobra.Command{
+	Use:   "is-valid-hostname",
+	Short: "Checks if the machine hostname is RFC1123 compliant",
```
Racer159, line 119: Per https://github.com/defenseunicorns/zarf/issues/1022 can we make it more clean that this is for the current machine?

```suggestion
	Short: "Checks if the current machine's hostname is RFC1123 compliant",
```

src/cmd/internal.go
```diff
@@ -113,6 +114,16 @@ var uiCmd = &cobra.Command{
...
+	Run: func(cmd *cobra.Command, args []string) {
+		if valid := utils.IsValidHostName(); !valid {
+			message.Fatal(nil, "The hostname is not RFC1123 compliant")
```
Racer159, line 122: Per the above (and this is what the user is more likely to see anyway).

```suggestion
			hostname, _ := os.Hostname()
			message.Fatalf(nil, "The hostname '%s' is not valid. Ensure the hostname meets RFC1123 requirements https://www.rfc-editor.org/rfc/rfc1123.html.", hostname)
```

src/types/package.go
Racer159: We could have the migration delete the scripts from the component if it sees that scripts to actions has taken place right?

https://github.com/defenseunicorns/zarf/blob/jeff-doing-stupid-jeff-things-smarter-migrations/src/pkg/packager/deprecated/scripts-to-actions.go

Racer159 (changes requested):

YrrepNoj (approved):
This PR has my unbiased approval.

Racer159 (approved):
:rocket:

PR #1145, opened 2022-12-27
update go.mod & package.json

jeff-mccoy:


Noxsios (dismissed):
SvelteKit 1.0 + Vite 4.0!

LGTM

jeff-mccoy:
SvelteKit is sad with 1.0 it looks like, can you take a look at this when you're back @Noxsios or @mike-winberry ?

Noxsios:
> SvelteKit is sad with 1.0 it looks like, can you take a look at this when you're back @Noxsios or @mike-winberry ?

I'll take a look first thing tomorrow

package.json
```diff
@@ -20,44 +20,44 @@
...
 		"@material/linear-progress": "14.0.0",
 		"@material/theme": "14.0.0",
-		"@playwright/test": "1.28.1",
-		"@sveltejs/adapter-static": "next",
-		"@sveltejs/kit": "next",
-		"@sveltejs/package": "next",
+		"material-symbols": "0.4.2",
+		"prismjs": "1.29.0",
+		"sanitize.css": "13.0.0",
+		"sass": "1.57.1",
```
Racer159, line 34: Why is sass a dependency now?

jeff-mccoy: I find the whole dep vs devdep weird with svelte tbh in how it is setup. They difference seems to be meaningless.

.grype.yaml
```diff
@@ -1,10 +1,18 @@
 # Ignore file for false positives from protobuf, see the following for more information:
 #   https://github.com/anchore/grype/issues/558
 ignore:
+  # False positives from CPE confusion of Go vs C (https://github.com/anchore/grype/issues/558)
   - vulnerability: CVE-2021-22570
   - vulnerability: CVE-2015-5237
 
-  # Bless your heart NPM, you're a mess
+  # False positive from CPE confusion of NPM vs Rust Crates (https://github.com/anchore/grype/issues/931)
```
jeff-mccoy, line 8: Mine was funnier though

package.json
Racer159: Yeah I think most stuff should be devDeps basically right?

.grype.yaml
Racer159: Mine was more useful :wink:

package.json
jeff-mccoy: Yeah think so, @Noxsios  or @mike-winberry may have thoughts.

Noxsios: It is 100% subjective, sometimes a dep prefers to be in one or the other. My logic is that things that help the build process (like typescript and any `@type`, etc...) go in devDep, vs deps that can be seen in the UI (ie material, prism) go in deps.  sass is a weird one where in my mind it could go in either.

Racer159: Sass is being used as a css pre-processor though right?  If so it would not be in the UI since that would only see regular CSS.  My personal opinion on deps vs devDeps is that if it is code that runs in the browser then it is a dep, but if it is code that generates code from other code that runs in the browser it is a devDep.  From what I know, in this case typescript and sass are doing the same thing

jeff-mccoy: yeah I'd lean towards deps are things that are needed at runtime (server or browser).

Racer159 (approved):
yay!

Issue #1153, opened 2023-01-02
Helm Chart localPath not resolving properly when used with Composable Packages


tworcester:
### Environment
Device and OS: Rocky 8
App version: v0.23.1
Kubernetes distro being used: k3s (zarf install)
Other:

### Steps to reproduce
Using the zarf composable-packages example try to create this package:
```yaml
kind: ZarfPackageConfig
metadata:
  name: compose-example
  description: "Demo Zarf package composability"

components:
  - name: games
    required: true
    description: "Example of a composed package with a unique description for this component"
    import:
      path: ../game
      # Example optional custom name to point to in the imported package
      name: baseline
  - name: local-chart
    required: false
    import:
      path: ../helm-local-chart
      name: demo-helm-local-chart
```

### Expected result
Bundled greatness

### Actual Result
```bash

  DEBUG   2023-01-02T09:42:08-05:00  -  packager.createComponentPaths({
            "name": "local-chart",
            "only": {
              "cluster": {}
            },
            "import": {
              "path": ""
            },
            "scripts": {},
            "charts": [
              {
                "name": "local-demo",
                "version": "0.1.0",
                "namespace": "local-chart",
                "localPath": "chart"
              }
            ],
            "images": [
              "nginx:1.16.0"
            ]
          })
  ⠋  Processing helm chart local-demo:0.1.0 from chart (0s)  DEBUG   2023-01-02T09:42:08-05:00  -  Chart.yaml file is missing                                                                               
     ERROR:  Validation failed for chart from chart (Chart.yaml file is missing)
  DEBUG   2023-01-02T09:42:08-05:00  -  goroutine 1 [running]:
          runtime/debug.Stack()
                /opt/hostedtoolcache/go/1.19.4/x64/src/runtime/debug/stack.go:24 +0x65
          github.com/defenseunicorns/zarf/src/pkg/message.Fatalf({0x301c220?, 0xc00044f608?}, {0x34ffef6, 0x28}, {0xc0014bdf28, 0x2, 0x2})
                /home/runner/work/zarf/zarf/src/pkg/message/message.go:149 +0x1df
          github.com/defenseunicorns/zarf/src/pkg/message.(*Spinner).Fatalf(0x34bebe5?, {0x4330f60?, 0xc00044f608?}, {0x34ffef6?, 0x28?}, {0xc0014bdf28?, 0x2?, 0x2?})
[...]

PR #1154, opened 2023-01-02
Update value of localPath for composable packages

tworcester:
## Related Issue

Fixes https://github.com/defenseunicorns/zarf/issues/1153

## Type of change


src/pkg/packager/compose.go
```diff
@@ -139,11 +139,14 @@ func (p *Packager) fixComposedFilepaths(parent, child types.ZarfComponent) types
 		child.Files[fileIdx].Source = p.getComposedFilePath(file.Source, parent.Import.Path)
 	}
 
-	// Prefix non-url composed component chart values files.
+	// Prefix non-url composed component chart values files and localPath.
 	for chartIdx, chart := range child.Charts {
 		for valuesIdx, valuesFile := range chart.ValuesFiles {
 			child.Charts[chartIdx].ValuesFiles[valuesIdx] = p.getComposedFilePath(valuesFile, parent.Import.Path)
 		}
+		if child.Charts[chartIdx].LocalPath != "" {
```
tworcester, line 147: Just for my own education, is this the proper way to check this variable when dealing with an `omitempty` field?
```go
LocalPath   string   `json:"localPath,omitempty" jsonschema:"oneof_required=localPath,description=The path to the chart folder"`
```

Racer159: Yep! That looks correct assuming you are checking that LocalPath has not been omitted (i.e. it contains a value).  By default (when not using pointers) Go will default to the zero value for that type https://go.dev/tour/basics/12

jeff-mccoy:
Thanks @tworcester we're waiting on #1173 to resolve issues around 3rd-party PRs

jeff-mccoy:
This is being weird, can you take a look at it again @tworcester?

tworcester:
> This is being weird, can you take a look at it again @tworcester?

Sure thing, the test failures?

tworcester:
@jeff-mccoy Please try again, I believe I fixed the issue. I ran the tests locally but not the `init-package` :facepalm:

tworcester:
Do we want to provide the ability to respect a full file path?  If people already using zarf have worked around this issue prior to this change, this would break them similar to what we saw above.

I think checking the provided `localPath` to see if it contains the zarf subcomponent folder already before adjusting would be defensive enough.  Thoughts?

jeff-mccoy:
Yeah I think that sounds reasonable and would prevent this from being a breaking change, even though it's also a bug fix.

jeff-mccoy (changes requested):
Add sanity check for existing behavior (component path in import path) to support existing broken behavior. Should leave a warning to the user noting this is incorrect.

jeff-mccoy:
Changing to draft to make tracking easier. Please complete the requested changes and then change back to ready for review, thanks!

tworcester:
> Changing to draft to make tracking easier. Please complete the requested changes and then change back to ready for review, thanks!

Thank you!  I likely won’t have time to get to this until next week.  (Just wanted to drop a note so you know I haven’t dropped it)

jeff-mccoy:
not a problem at all, we may pick up the rest this week if we can squeeze it in, thanks!

src/pkg/packager/compose.go
```diff
@@ -145,11 +145,14 @@ func (p *Packager) fixComposedFilepaths(parent, child types.ZarfComponent) types
...
+			child.Charts[chartIdx].LocalPath = p.getComposedFilePath(child.Charts[chartIdx].LocalPath, parent.Import.Path)
```
YrrepNoj, line 154:
```suggestion
			// Check if the localPath is relative to the parent Zarf package
            if _, err := os.Stat(child.Charts[chartIdx].LocalPath); os.IsNotExist(err) {
                // Since the chart localPath is not relative to the parent Zarf package, get the relative path from the composed child
                child.Charts[chartIdx].LocalPath = p.getComposedFilePath(child.Charts[chartIdx].LocalPath, parent.Import.path)
            }
```

jeff-mccoy (approved):

YrrepNoj (approved):
Eventually ™️, I think it would be a good idea to not even bother with the `os.Stat()` files from the parent directory perspective. Even though it's pretty unlikely, I think it's not that intuitive.

Of course since it's old behavior that we're trying not to break, it makes 110% sense to include in this PR for now!

tworcester:
Thank you for wrapping this up, apologies for being absent!

Issue #1141, opened 2022-12-20
Support Terraform Vars from file on Package Deploy


ntwkninja:
Problem:
Zarf currently only supports Vars in yaml files.

Potential Solution:
Given the locals.tf example below, support injecting different terraform [variable types](https://developer.hashicorp.com/terraform/language/expressions/types) on zarf package create and zarf package deploy.

Alternative Solutions:
- Duct tape (i.e. build a tf.yaml file to inject variables `cat tf.yaml | yq > locals.tf`)

**Additional context**
```
locals {

###########################################################
################## Global Settings ########################

  region                      = ###ZARF_VAR_REGION###  # target AWS region
  region2                     = ###ZARF_VAR_REGION2###  # RDS backup target AWS region
  account                     = ###ZARF_VAR_AWS_ACCOUNT###  # target AWS account
  aws_profile                 = ###ZARF_VAR_AWS_PROFILE###  # local AWS profile to be used for deployment


  tags = {
    Blueprint                 = "${replace(basename(path.cwd), "_", "-")}"  # tag names based on the directory name
    GithubRepo                = "github.com/aws-ia/terraform-aws-eks-blueprints"
  }
  
###########################################################
#################### VPC Config ###########################

  azs                         = ["${local.region}a", "${local.region}b", "${local.region}c"]
  vpc_cidr                    = ###ZARF_VAR_VPC_CIDR###
  vpc_name                    = ###ZARF_VAR_VPC_NAME###
  database_subnets            = ###ZARF_VAR_DB_SUBNETS###

  create_database_subnet_group       = true
  create_database_subnet_route_table = true


###########################################################
#################### EKS Config ###########################

  cluster_name                = ###ZARF_VAR_EKS_CLUSTER_NAME###
  eks_k8s_version             = ###ZARF_VAR_EKS_CLUSTER_VERSION###
  # list of admin's AWS account arn to allow control of KMS keys
  cluster_key_admin_arns      = [###ZARF_VAR_EKS_ADMIN_ARN_0###,###ZARF_VAR_EKS_ADMIN_ARN_1###]
  # list of admin's AWS account/group info to allow access to EKS cluster
  #16
  aws_auth_eks_map_users      = [
    {
      userarn  = ###ZARF_VAR_EKS_ADMIN_ARN_0###
      username = ""
      groups   = ["system:masters"]
    },
    {
      userarn  = ###ZARF_VAR_EKS_ADMIN_ARN_1###
      username = ""
[...]

Issue #1129, opened 2022-12-14
Traefik package in the init package


Racer159:
**Is your feature request related to a problem? Please describe.**
Currently we use `--disable-traefik` on our k3s server run leaving it up to customers to bring their own ingress.  This is fine on more complex deploys but it would be nice to have an easy default for those who don't need the complexity.

**Describe the solution you'd like**
A new (optional) init package that installs the Traefik chart into a cluster to setup ingress.

**Describe alternatives you've considered**
We could make the `--disable-traefik` flag configurable more directly but this could open things up to more complexity with our configuration.

**Additional context**
This will be useful in making the init package more a-la-carte (see #1062)

Racer159:
closing since we support `--set K3S_ARGS=""` now

PR #1171, opened 2023-01-07
Template files

jeff-mccoy:
This PR adds variable templating to text files included under the `files` key in a Zarf Package. This mirrors the existing behavior of Kustomizations, charts, and manifests, except for the case of files called before the Zarf package needs a cluster. In this case, some special variables, such as `###ZARF_REGISTRY###` will not be available. This was needed because some components with files will be called before a cluster is created, such as in the [K3s package](https://github.com/defenseunicorns/zarf/tree/main/packages/distros/k3s).

This PR also fixes a typo with the data-injection marker and permits the use of `0-9` for variable/constant names in addition to `A-Z_`.

Related to #1129 
Related to #1141


Racer159:
Added breaking change since the DATA_INJECTION_MARKER change will break any existing data injection packages.

src/internal/packager/template/template.go
```diff
@@ -91,9 +91,10 @@ func (values Values) Apply(component types.ZarfComponent, path string) {
 	}
 
 	// Include the data injection marker template if the component has data injections
-	// TODO: (@jeff-mccoy) DATA_INJECTON_MARKER is misspelled however it is in use by other packages and would be a breaking change!
 	if len(component.DataInjections) > 0 {
+		// Preserve existing misspelling for backwards compatibility
 		builtinMap["DATA_INJECTON_MARKER"] = config.GetDataInjectionMarker()
```
Racer159, line 96: Actually, can we track this into templating and warn the user that they should update their spelling?  (assuming we would want to remove this one day).

jeff-mccoy: It's a string replacement, little worried about breaking something if we do that, ideas?

Racer159: We could have a deprecated list of keys that it would search for first and just display a warning about them.  We could even have that be a string to string map so that we could loop through the keys to search and then display a message like:

`Warning:  This Zarf Package uses a deprecated variable: "DATA_INJECTON_MARKER" changed to "DATA_INJECTION_MARKER".  Please notify your package creator for an update.`

Racer159: @Madeline-UX on potential wording there

src/pkg/utils/io.go
```diff
@@ -191,3 +194,26 @@ func SplitFile(path string, chunkSizeBytes int) (chunks [][]byte, sha256sum stri
...
+	n, err := f.Read(data)
+	if err != nil && err != io.EOF {
+		return false, err
+	}
+
+	// Use http.DetectContentType to determine the MIME type of the file
+	mimeType := http.DetectContentType(data[:n])
+
+	// Check if the MIME type indicates that the file is text
+	return strings.HasPrefix(mimeType, "text/"), nil
```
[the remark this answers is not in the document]
jeff-mccoy: yeah that's fair, what about something like this?

Screenshot 2023-01-09 at 11 56 01 AM

referring to https://www.iana.org/assignments/media-types/media-types.xhtml#application

src/internal/packager/template/template.go
jeff-mccoy: so this is a string replace, so this would be adding an additional scan to the files?

Racer159: That would be the easiest way to do it, though ideally only for the limited subset of deprecations.  We are already scanning the file bytes many times over.

Racer159:
```
// ReplaceTextTemplate loads a file from a given path, replaces text in it and writes it back in place.
func ReplaceTextTemplate(path string, mappings map[string]string) {
	text, err := os.ReadFile(path)
	if err != nil {
		message.Fatalf(err, "Unable to load %s", path)
	}

	for deprecated, value := range template.Deprecations {
		if bytes.Contains(text, deprecated) {
			message.Warnf("This Zarf Package uses a deprecated variable: '%s' changed to '%s'.  Please notify your package creator for an update.", &deprecated, &value)
		}
	}

	for template, value := range mappings {
		text = bytes.ReplaceAll(text, []byte(template), []byte(value))
	}

	if err = os.WriteFile(path, text, 0600); err != nil {
		message.Fatalf(err, "Unable to update %s", path)
	}
}
```

src/pkg/utils/io.go
```diff
@@ -103,6 +103,12 @@ func ReplaceTextTemplate(path string, mappings map[string]string) {
 		text = bytes.ReplaceAll(text, []byte(template), []byte(value))
 	}
 
+	for old, new := range deprecations {
```
Racer159, line 106: Doing this here would never display the warning...  It would already be replaced

Racer159 (changes requested):

src/pkg/utils/io.go
jeff-mccoy: thanks, multitasking fail.

PR #1179, opened 2023-01-09
803 add component descriptions to the required components in the init config package

### LINKED ISSUE #803: 💡 Add component descriptions to the required components in the Init config package.
**Current State**
Currently, the Init Config Package does not have descriptions for 

- injector
- seed registry
- registry
- agent

Screen Shot 2022-09-27 at 5 00 33 PM
Screen Shot 2022-09-27 at 5 08 47 PM

**Suggestion**
Add component descriptions to the package configuration that describe what each component does to help users understand why it is required.

Noxsios:
## Description

Adds detailed descriptions to init package components. Also updates alignment of UI that displays these components in an accordion. Also add another line break between the YAML component printed to the console and the component description.

Component descriptions are taken from this comment by @Madeline-UX : https://github.com/defenseunicorns/zarf/issues/803#issuecomment-1294103006

## Related Issue

Fixes #803 

## Type of change


Noxsios:
Screenshot 2023-01-09 at 5 54 02 PM

Screenshot 2023-01-09 at 5 54 53 PM

Madeline-UX (dismissed):

packages/zarf-injector/zarf.yaml
```diff
@@ -5,6 +5,9 @@ metadata:
 
 components:
   - name: zarf-injector
+    description: |
+      Injects Zarf into the cluster so it can bootstrap other core Zarf components. 
+      Removed and destroyed after the registry is booted
```
Racer159, line 10: Is there supposed to be a period at the end here?

Racer159 (changes requested):

packages/zarf-injector/zarf.yaml
Racer159, line 10:
```suggestion
      Removed and destroyed after the registry is booted.
```

packages/zarf-injector/zarf.yaml
```diff
@@ -5,6 +5,9 @@ metadata:
 
 components:
   - name: zarf-injector
+    description: |
+      Injects Zarf into the cluster so it can bootstrap other core Zarf components.
```
runyontr, line 9: I think it might help to expand on this instead of saying it "injects Zarf".  There's bound to be some confusion about the Zarf Injector and the mutating webhook.  Perhaps having a reference back to a page in the docs, or something with more information about what the injector is, or using the injector name so someone using it could find it more easily in the docs would be helpful

Noxsios: perhaps `A tiny Rust binary that is used to setup the in-cluster seed registry when an external registry is unavailable.  See https://github.com/defenseunicorns/zarf/blob/main/adr/0007-use-rust-binary-for-both-injection-stages.md for more information.`

jeff-mccoy: I'd probably word it something like bootstraps a cluster with a single binary to load the Zarf Registry. Is probably more than sufficient for this. It's not like they can skip deploying it, if it's needed it will be deployed regardless.

YrrepNoj (changes requested):

packages/zarf-injector/zarf.yaml
Noxsios: > I'd probably word it something like bootstraps a cluster with a single binary to load the Zarf Registry. Is probably more than sufficient for this. It's not like they can skip deploying it, if it's needed it will be deployed regardless.

`Bootstraps a cluster using a tiny Rust binary to deploy the internal Zarf Registry`?

packages/zarf-agent/zarf.yaml
```diff
@@ -9,6 +9,9 @@ constants:
 
 components:
   - name: zarf-agent
+    description: |
+      Mutating webhook that enables Kubernetes (or cloud) resources to run in airgap.
+      Directs requests for external resources to internal Zarf-managed resources.
```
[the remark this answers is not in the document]
Noxsios: What would you recommend? @YrrepNoj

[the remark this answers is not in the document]
YrrepNoj: It's defiantly tough to come up with a description for this that is decently descriptive but also simple to understand...

Perhaps something like this?
```
Mutating webhook that enables Kubernetes (or cloud) resource to run in airgap.
This enables the cluster to redirect requests for external resources (namely images and git repositories) that are vital to deploying and updating Zarf Packages to internal Zarf-defined resources.
```
 - Part of me want's to say 'control plane' instead of cluster but I think 'cluster' is more generic and potentially easier to understand.
 - I changed `zarf-managed` to `zarf-defined` b/c the agent will still redirect requests to external repos/registries that get defined via flags

packages/zarf-injector/zarf.yaml
Madeline-UX: My thoughts: What would make the most sense to someone like Ashton? I think the above language is concise and clearly describes functionally. 

@runyontr I would be hesitant to include links to the docs since airgap users may not have access to those at this time.

runyontr: Agreed.  Perhaps the docs should be added to the ZarfUI for airgap help?

Madeline-UX: @runyontr It is something the team has discussed - there may be an issue for it if you would like to add your support. :) But I would want to heavily consider who would be using the docs in Airgap too. Right now the docs are written for a number of different users - not just Ashton

packages/zarf-agent/zarf.yaml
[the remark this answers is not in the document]
Madeline-UX: @YrrepNoj Where should we tell them it redirects to? I want to be sure that Ashton understands where that redirect is going to - and that it stays within their network

packages/zarf-agent/zarf.yaml
```diff
@@ -9,6 +9,11 @@ constants:
 
 components:
   - name: zarf-agent
+    description: |
+      A Kubernetes mutating webhook to enable automated URL rewriting for containers
```
Racer159, line 13:
```suggestion
      A Kubernetes mutating webhook to enable automated URL rewriting for container
```

packages/zarf-agent/zarf.yaml
```diff
@@ -9,6 +9,11 @@ constants:
...
+      images and git repositories references in Kubernetes manifests. This prevents
```
Racer159, line 14:
```suggestion
      images and git repository references in Kubernetes manifests. This prevents 
```

Racer159 (changes requested):
Too many plurals in one of the descriptions

Racer159 (approved):
lgtm!

PR #1183, opened 2023-01-10
Cleanup Icon component implementation

Noxsios:
## Description

The `icon.svelte` component has re-implementation of the `material-symbols` library.  Since the library exports a [typescript type](https://github.com/marella/material-symbols/blob/main/material-symbols/index.d.ts), a wrapper component around this should be used instead.  There is also a fallback for Zarf specific icons (such as `package` and `component`).

## Type of change


jeff-mccoy (approved):
SGTM.

Screenshot 2023-01-12 at 10 54 28 PM

PR #1196, opened 2023-01-12
Rename files to prevent being parsed as documentation

bdfinst:
## Description
The markdown files in hack/zarf-jsfh-template are breaking the build for zarf-docs. By prefixing the file names with "_", they are ignored by Docusaurus.


Racer159:
@bdfinst looks like you missed one in `hack/zarf-jsfh-template/_section_properties_details.md` on line 13:

```
{%- if sub_property.is_pattern_property %} Pattern Property{% endif %} {% with schema=sub_property %}{%- include "breadcrumbs.md" %} {% endwith %}
```

You can test this locally by running `make docs-and-schema` to make sure you got all of them without needing to wait for CI.

Racer159 (changes requested):
Per the above

PR #1197, opened 2023-01-12
Made show all namespace default for k9 dash

### LINKED ISSUE #1195: Default to all namespaces for zarf tools monitor (k9s)
Currently `zarf tools monitor` defaults to only showing the `zarf` namespace.  This shows people Zarf's state but is often not what people want to see (they want to see the state of additional packages installed on top as well).

`zarf tools monitor` shows all namespaces by default

You could force a different default but showing them the full state of the system (while potentially minorly overwhelming) shows a clearer picture of the full state of things without needing to know to press `0`

**Additional context**
This is something I have seen cause confusion with a few users not knowing where their stuff was however would need review by @Madeline-UX.  Below is the current default:
image

And this is what I would propose:
image

(note that there are pod statuses hidden from the user by default)

Madeline-UX:
## Description

This issue makes show all namespace the default for the K9 dash. It was made based on user insights and a few user errors that led to users believing their pods were note deployed due to not using the zarf namespace. 

This may also be important for packages that use the big-bang namespace or that deploy other resources that use a namespace other than Zarf

## Related Issue

Fixes #1195 1195 (issue)

## Type of change

## Checklist before merging

### DIFF
--- a/src/cmd/tools.go
+++ b/src/cmd/tools.go
@@ -94,7 +94,7 @@ var k9sCmd = &cobra.Command{
 	Short:   lang.CmdToolsMonitorShort,
 	Run: func(cmd *cobra.Command, args []string) {
 		// Hack to make k9s think it's all alone
-		os.Args = []string{os.Args[0], "-n", "zarf"}
+		os.Args = []string{os.Args[0]}
 		k9s.Execute()
 	},
 }


jeff-mccoy (approved):
lgtm

PR #1203, opened 2023-01-14
Push both a crc and non-crc image tag for zarf images

### LINKED ISSUE #1143: Option for non-CNC hashed images in Zarf registry
Our use cases for using Zarf have been broken since shifting to only CNC-hashed container images. We rely on Zarf as our airgap tool for things other than Kubernetes (Zarf agent handled) resources and adding in the CNC requirement forces us to refactor either additional logic to add image names/tags or to make our own solution from scratch.

We would like the option to have both the CNC hashed and original named images. We do still have the expected K8s usage so the agent is still useful, but we would like the ability to not have to rely on the agent for non-K8s usage.

**Additional context**
This is the result from these two Zarf Slack threads:
https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1669834296629069
https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1671548425862209

Racer159:
## Description

Pushes both a CRC hashed and non-CRC hashed image from Zarf.  Also improves testing of connect logic.

## Related Issue

Fixes #1143 

## Type of change

## Checklist before merging


jeff-mccoy (approved):
LGTM, and additional test logic is great!

PR #1205, opened 2023-01-15
Wrap the crane command to make it easier to use with Zarf

### LINKED ISSUE #1186: Make it easier to view contents of registry
It is difficult to view the contents of the zarf docker registry.

`zarf tools registry catalog` displays the contents of the registry with no other configuration or actions needed

I think this is what needs to happen?

* `zarf connect <the registry>`
* then `zarf tools registry login <the registry>`
* then `zarf tools registry catalog`

**Additional context**
image (3)

Racer159:
## Description

This wraps the crane command so that it is easier to use more directly with Zarf

## Related Issue

Fixes #1186 

## Type of change

## Checklist before merging

### DIFF
--- a/docs/4-user-guide/1-the-zarf-cli/100-cli-commands/zarf_tools_registry_catalog.md
+++ b/docs/4-user-guide/1-the-zarf-cli/100-cli-commands/zarf_tools_registry_catalog.md
@@ -6,6 +6,16 @@ List the repos in a registry
 zarf tools registry catalog [REGISTRY] [flags]
 ```
 
+### Examples
+
+```
+  # list the repos internal to Zarf
+  $ zarf tools registry catalog
+
+  # list the repos for reg.example.com
+  $ zarf tools registry catalog reg.example.com
+```
+
 ### Options
 
 ```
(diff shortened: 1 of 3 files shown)


jeff-mccoy:
This is a great short-term add, I was actually thinking maybe we should just add a more top-level command to interact with artifacts/assets zarf deploys.  Something like `zarf artifact images` or`zarf assets images`. I think it should probably be view-only since mutation would break the nature of declarative packages. Also `zarf artifact repos`. Thoughts?

Racer159:
> 

I think that could make sense, though we would likely need to pull that info from the zarf state instead of the actual service.  Doing that for git repos might get tricky across git hosts.

jeff-mccoy (changes requested):
There's a lint error in `tools.go`. Also, since we're adding custom logic outside of just crane, we should add a test for this. Fortunately we also need this function in the `src/test/e2e/21_connect_test.go` test, so might be able to just drop it in there.

Racer159:
> There's a lint error in `tools.go`. Also, since we're adding custom logic outside of just crane, we should add a test for this. Fortunately we also need this function in the `src/test/e2e/21_connect_test.go` test, so might be able to just drop it in there.

Fixed

jeff-mccoy (approved):
💥

PR #1206, opened 2023-01-15
Consolidate components flag

### LINKED ISSUE #1182: No longer able to `zarf package deploy` an init package
### Environment
Device and OS: Ubuntu EC2 instance
App version: v0.23.2
Kubernetes distro being used: built-in k3s
Other:

### Steps to reproduce
1. Download the init package that matches the version of zarf, rename it to something else like `init.tar.zst`
2. Run `./zarf package deploy init.tar.zst --components k3s --confirm`
3. Zarf will error, giving the following error:

image

### Expected result

Zarf deploys the init package, in the same way that running `zarf init` would have. This worked in previous versions of Zarf (the version I'm upgrading from is v0.21.3.

Suggestion: Read the zarf.yaml of the package and interrogate its `kind:` rather than relying on the filename.

### Actual Result

The above error

### Severity/Priority

Functionality that worked previously is broken, but it's not the end of the world for me to upgrade my automation. It may be a bigger issue for people that create custom init packages though.

### Additional Context

* This behavior is useful because without it, automation processes will need to have a greater understanding of what version of Zarf is being used. I was previously able to have a makefile target `make build/zarf-init-amd64.tar.zst`, which is great, because Make will automatically detect when the file is present or not, or old, and skip running that download when it doesn't need to.
* It was also useful if a user was creating a custom init package.

Log showing zarf thinks it isn't an init package because of the filename:
Screen Shot 2023-01-09 at 4 04 59 PM

Racer159:
## Description

This consolidates the components flag (following the pattern from set) so that you can Zarf Package Deploy and init package again.

## Related Issue

Fixes #1182 

## Type of change

## Checklist before merging

### DIFF
--- a/src/cmd/initialize.go
+++ b/src/cmd/initialize.go
@@ -77,7 +77,7 @@ func findInitPackage(initPackageName string) (string, error) {
 	// Create the cache directory if it doesn't exist
 	if utils.InvalidPath(config.GetAbsCachePath()) {
 		if err := os.MkdirAll(config.GetAbsCachePath(), 0755); err != nil {
-			message.Fatalf(err, "Unable to create cache directory: %s", config.GetAbsCachePath())
+			message.Fatalf(err, lang.CmdInitErrUnableCreateCache, config.GetAbsCachePath())
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


jeff-mccoy (approved):
lgtm

PR #1220, opened 2023-01-19
add tools get-creds command

### LINKED ISSUE #1098: Zarf CLI should easily show Git and Registry credentials
When looking to debug a deployment after doing a zarf init, if I didn't save the login credentials that get spit out durin gthe init phase, I can't find a clean way to discover the credentials.  This is how I've been finding it:

```shell
 k get secrets -n zarf zarf-state -o jsonpath="{ .data.state }" | base64 --decode | jq -r .
```

it would be nice if there were a zarf cli command to extract either the full secret, or at least the credentials that were returned as part of the init phase

andrewg-xyz:
## Description

Add command to easily get credentials printed during zarf init (i.e. git, logging, registry)

## Related Issue

Fixes #1098 

## Type of change

## Checklist before merging


andrewg-xyz:
Initial changes, will refactor get-git-password to support

`zarf tools get-creds <tool>` 

printing only the string secret, supporting users who will pipe the secret.

YrrepNoj:
Can you run `make docs-and-schema` to auto generate docs for the CLI help messages 😄 

(if a pre-commit linter complains when you try to commit please ignore the pre-commit and just commit with `--no-verify` for now)

Racer159 (changes requested):

.github/codeql.yaml
```diff
@@ -1,6 +1,7 @@
 paths-ignore:
   - src/pkg/packager/network.go
   - src/pkg/utils/network.go
+  - src/pkg/utils/credentials.go
```
Racer159, line 4: There is also a call that prints the push password in [src/cmd/tools.go](https://github.com/defenseunicorns/zarf/pull/1220/files#diff-69586b09df01b25b91a3d2d98bdec15c9104bbe387f369c4a46aae62fa027b03).  I would recommend making it so that this calls a credentials.go function to print this so both can be ignored.

src/cmd/tools.go
```diff
@@ -86,10 +89,33 @@ var readCredsCmd = &cobra.Command{
 		}
 
 		message.Note(lang.CmdToolsGetGitPasswdInfo)
+		message.Warn(lang.CmdToolGetGitDeprecation)
 		fmt.Println(state.GitServer.PushPassword)
```
Racer159, line 93: See other comment (recommend moving to credentials.go)

Racer159 (dismissed):

src/cmd/tools.go
andrewg-xyz: updated with [484cc98](https://github.com/defenseunicorns/zarf/pull/1220/commits/484cc980a35e6c9f496ff5239b3ac05a3054f98f)

.github/codeql.yaml
andrewg-xyz: updated with [484cc98](https://github.com/defenseunicorns/zarf/pull/1220/commits/484cc980a35e6c9f496ff5239b3ac05a3054f98f)

jeff-mccoy (approved):
This behaved as I would expect with the password only be sent to stdout for capture.  
Screenshot 2023-01-26 at 7 34 06 PM

Screenshot 2023-01-26 at 7 34 45 PM

PR #1230, opened 2023-01-20
Fix external registry issues

Racer159:
## Description

This refactors a lot of how we interact with registries (particularly external ones) to address the following bugs:

- The inability to use registries external to the current cluster
- The inability to set the NodePort of the internal Zarf registry

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging


Racer159:
Added breaking change since this will be minorly breaking as is for people who use an in cluster external registry

Racer159:
Dropping the breaking change label after some more testing.  It will update the state correctly on upgrade.

src/test/external-test/ext_out_cluster_init_test.go
```diff
@@ -0,0 +1,96 @@
...
+	giteaIP := "172.31.0.99"
+	giteaHost := "gitea.localhost"
+	registryHost := "registry.localhost"
+
+	zarfBinPath := path.Join("../../../build", test.GetCLIName())
+
+	// Teardown any leftovers from previous tests
+	_ = exec.CmdWithPrint("k3d", "cluster", "delete")
+	_ = exec.CmdWithPrint("k3d", "registry", "delete", registryHost)
+	_ = exec.CmdWithPrint("docker", "network", "remove", network)
```
jeff-mccoy, line 31: this would probably make a local dev real sad if their network were killed, should we put some sort of warning on that?

src/test/external-test/ext_out_cluster_init_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Setup a network for everything to live inside
+	err := exec.CmdWithPrint("docker", "network", "create", "--driver=bridge", "--subnet="+subnet, "--gateway="+gateway, network)
+	require.NoError(t, err, "unable to create the k3d registry")
+
+	// Install a k3d-managed registry server to act as the 'remote' container registry
+	err = exec.CmdWithPrint("k3d", "registry", "create", registryHost, "--port", "5000")
```
jeff-mccoy, line 38: big fan of using the k3d registry, clever way to test that behavior

src/test/external-test/configure-gitea.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Retry gitea migrate until the db is ready
+timeout 30 bash -c 'until gitea migrate; do sleep 2; done'
+
+echo '==== BEGIN GITEA CONFIGURATION ===='
+
+function configure_admin_user() {
```
jeff-mccoy, line 9: not a big deal for this pr, but would love to explore any other (lightweight) alternatives to gitea for this test since we already test just for diversity of git servers

jeff-mccoy (commented):
just a couple notes

src/test/external-test/ext_out_cluster_init_test.go
Racer159: It would only kill the named network and registry used by the test.  It would kill their k3d cluster though if they were using the default name.  We could name that too?

src/test/external-test/ext_out_cluster_init_test.go
Racer159: Thanks!

src/test/external-test/configure-gitea.sh
Racer159: Yeah it would have to be a push-to-create supporting service which we can look around for but to my knowledge is only gitea (and its derivatives) and gitlab right now.  @YrrepNoj may have found others though since he implemented the external stuff.

src/test/external-test/ext_out_cluster_init_test.go
jeff-mccoy: ah n/m I see this is removing only that k3d network, disregard.

src/test/external-test/configure-gitea.sh
jeff-mccoy: 👍

jeff-mccoy (approved):
🎉

PR #1233, opened 2023-01-21
fix agent building locally, in ci and on release

Racer159:
## Description

This PR hotfixes agent building in CI, on release and locally

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging


Racer159:
Marking as ready for review, but wanted feedback on how to best handle the windows case.  As I see it there would be a few ways to do it:

1. Drop the `if` statement entirely and always build the local version of the agent
2. Run both the linux/mac and windows versions of the command not caring that garbage will be output to the terminal
3. Create a shim to convert `if [ "agent:local" = "agent:local"]; then make build-agent-image; fi` to `if ( "agent:local" = "agent:local" ) { make build-agent-image }`
4. Now that we have fancy `cmd` syntax in the scripts to actions PR, add an OS only filter to that

src/pkg/packager/variables.go
```diff
@@ -15,37 +15,45 @@ import (
...
 			return fmt.Errorf("variable '%s' must be '--set' when using the '--confirm' flag", key)
 		}
 	}
 
 	templateMap := map[string]string{}
-	for key, value := range packageVariables {
-		// Variable keys are always uppercase in the format ###ZARF_PKG_VAR_KEY###
-		templateMap[strings.ToUpper(fmt.Sprintf("###ZARF_PKG_VAR_%s###", key))] = *value
+	for key, value := range setVariableMap {
+		templateMap[fmt.Sprintf("###ZARF_PKG_VAR_%s###", key)] = value
```
wirewc, line 56: I assume this means no more capitalized variables only for zarf vars?

wirewc:
LGTM!

packages/zarf-agent/zarf.yaml
```diff
@@ -25,3 +25,7 @@ components:
           - manifests/secret.yaml
           - manifests/deployment.yaml
           - manifests/webhook.yaml
+    scripts:
```
jeff-mccoy, line 28: should update to actions for this now.

.github/workflows/test-windows.yml
```diff
@@ -39,6 +39,12 @@ jobs:
         run: make build-cli-windows-amd
         shell: pwsh
 
+      # Builds an init package manually off of the last release
+      # Windows in GitHub cannot natively build linux containers
+      - name: Build init-package
+        run: |
+          make release-init-package ARCH=amd64 AGENT_IMAGE=agent:v0.23.6
```
jeff-mccoy, line 46: since we aren't actually using this for these tests, can we leave a note stating that, maybe we can even make a custom tag that just says "ignore-me" or something for this?

src/types/runtime.go
```diff
@@ -13,12 +13,13 @@ type ZarfCommonOptions struct {
...
-	SetVariables map[string]string `json:"setVariables" jsonschema:"description=Key-Value map of variable names and their corresponding values that will be used to template against the Zarf package being used"`
+	Insecure        bool              `json:"insecure" jsonschema:"description=Allow insecure connections for remote packages"`
+	Shasum          string            `json:"shasum" jsonschema:"description=The SHA256 checksum of the package to deploy"`
+	PackagePath     string            `json:"packagePath" jsonschema:"description=Location where a Zarf package to deploy can be found"`
+	Components      string            `json:"components" jsonschema:"description=Comma separated list of optional components to deploy"`
+	SGetKeyPath     string            `json:"sGetKeyPath" jsonschema:"description=Location where the public key component of a cosign key-pair can be found"`
+	SetVariables    map[string]string `json:"setVariables" jsonschema:"description=Key-Value map of --set variable names and their corresponding values that will be used to template against the Zarf package being used"`
```
jeff-mccoy, line 21: I probably need to chat about this one, want to understand the goal with this change.

jeff-mccoy (commented):
left a few notes

.github/workflows/test-windows.yml
Racer159: I'll add a note (someone poking in our package registry may get confused if they find the tag lol)

src/types/runtime.go
Racer159: Short answer is that it enables you to do this:

`zarf-config.toml`

```
[package.create.set]
# The image tag used for the zarf agent, defaults to a dev image tag
agent_image = 'agent:local'

# Tag for the zarf injector binary to use
injector_version = 'pr-948-e699899'
```

```
$ zarf init --set AGENT_IMAGE=agent:v0.23.6 --confirm
```

By default the `--set` array is _either_ the one from the config _or_ the one set on the command line (which in the above example doesn't contain `INJECTOR_VERSION`).   This allows the two to be merged with `--set` on the CLI taking precedent over the config

Racer159: Refactored this out of pkg as requested.

jeff-mccoy (approved):
🥳

PR #1240, opened 2023-01-21
Updated push.go

### LINKED ISSUE #1211: Provide more clarity on "unable to push the git repo" warning message
Zarf version v0.23.2

In the course of a `zarf package deploy` I receive several warning messages saying "Warning -- Unable to push the git repo", but zarf continues on deploying the package, and it eventually all comes up happy, which means to me that the git repo(s) were indeed actually pushed successfully. I suspect there is a retry that is happening, but it is not immediately obvious that that is the case.

If this warning appears, it informs the user that it is retrying, and then lets the user know what the ultimate result was (failure or success)

**Additional context**
image

Yash-C-330:
## Description
Changed warning message from "Unable to push the git repo" to "retrying" in the push.go
...

## Related Issue

Fixes #1211 

## Type of change

changed the warning message in the push.go


Racer159 (changes requested):
Requesting changes per the comments in the issue: https://github.com/defenseunicorns/zarf/issues/1211 (the message should include more than the word retrying)

Madeline-UX (changes requested):
Please review the language that was proposed in the original issue.

YrrepNoj:
Hey @Yash-C-330 really appreciate you staying on top of keeping this PR update to date :)

I'm going to take another look at the #1211 issue with the team today and come to a conclusion about what we want the exact verbiage to be since there were a couple good ideas discussed but we didn't see to come to a final conclusion yet.

Once we come to a conclusion, we might request you make a few additional change to this PR to match the verbiage we pick. Once again, we very much appreciate you picking up this issue for yourself and keeping the PR up to date 👍

Yash-C-330:
Hi, @YrrepNoj thank you for replying, I will be waiting for your request on this issue.

src/internal/packager/git/push.go
```diff
@@ -32,7 +32,7 @@ func (g *Git) PushRepo(localPath string) error {
 	}
 
 	if err := g.push(repo, spinner); err != nil {
-		spinner.Warnf("Unable to push the git repo %s", basename)
+		spinner.Warnf("Unable to push the git repo %s because of %w. Retrying the push.", basename, err)
```
Racer159, line 35: @Yash-C-330 apologies for the delay on this as we work through some other issues in the background, I think the overall flow is correct now but the error unwrapping doesn't print as expected (see below):

image

The following would unwrap this correctly:

```suggestion
		spinner.Warnf("Unable to push the git repo %s (%s). Retrying....", basename, err.Error())
```

Racer159: (here is an example of what that above suggestion outputs)
image

Racer159 (changes requested):
Thank you for your patience on this, should be gtg after this

Racer159 (approved):
lgtm
image

Madeline-UX (approved):
Yes, this message thread meets the desired outcomes for keeping the user informed to what is going on at all times.

PR #1242, opened 2023-01-22
fix compare tool generation and add better testing

Racer159:
## Description

this fixes compare tool generation for things that either don't have images or don't have component sboms.

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging

### DIFF
--- a/src/internal/packager/sbom/catalog.go
+++ b/src/internal/packager/sbom/catalog.go
@@ -106,6 +106,7 @@ func Catalog(componentSBOMs map[string]*types.ComponentSBOM, tagToImage map[name
 		currImage++
 	}
 
+	// Include the compare tool if there are any image SBOMs OR component SBOMs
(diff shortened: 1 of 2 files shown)


src/internal/packager/sbom/catalog.go
```diff
@@ -106,7 +106,7 @@ func Catalog(componentSBOMs map[string]*types.ComponentSBOM, tagToImage map[name
 		currImage++
 	}
 
-	if len(componentSBOMs) > 0 && len(tagToImage) > 0 {
+	if len(componentSBOMs) > 0 || len(tagToImage) > 0 {
```
jeff-mccoy, line 109: in git history this change might seem odd, can we leave a quick code comment above to explain this conditional?

Racer159: done

jeff-mccoy (approved):

Issue #854, opened 2022-10-06
RKE 2 Distro Support POC


JasonvanBrackel:
Create alpha-quality feature for supporting the RKE2 Distro at the same level as K3s

Use the K3s Zarf init package as a model for supporting RKE 2.  This is the first step of #853 and will give us something to show the various teams that are asking for this support, to gather feedback for a beta-target feature release.

codywyoming177:
Have a few folks on our end looking forward to this.  Happy to connect for user-feedback/testing.  I can also provide context on mission need/timeline if it'll help prioritize.

FruityWelsh:
What is the status of this effort?

brandtkeller:
I believe there is interest in this existing as a baseline - right now we are not focused on adding additional distro packages to the `zarf` repository and instead believe that we can establish a more suitable location for the community to collaborate on initiatives such as these.

PR #1264, opened 2023-01-25
Working Longhorn Example

JasonvanBrackel:
## Description
Add to Longhorn to Zarf Examples 

## Related Issue

Relates to https://github.com/defenseunicorns/zarf/issues/854 as longhorn will be used as the backing storage for this ZarfInitPackage

## Type of change

## Checklist before merging


wirewc:
Definitely should do the annotation for the service so a user may access the API.

jeff-mccoy:
NIce, I didn't realize how big it was 😆. Can we drop a couple of screenshots of this in action since we don't have a test for it right now?

JasonvanBrackel:
> Definitely should do the annotation for the service so a user may access the API.

I'm using the upstream helm chart, I'll have to see if there's a helm value to add the annotation.  

I'll think it through if not.

JasonvanBrackel:
> NIce, I didn't realize how big it was 😆. Can we drop a couple of screenshots of this in action since we don't have a test for it right now?

Sure thing

JasonvanBrackel:
Here's some screens from the twitch stream where I was working on it.
image
image
image
image

JasonvanBrackel:
@jeff-mccoy screenshots there and @wirewc connect added

jeff-mccoy (approved):
great add, lgtm

PR #1265, opened 2023-01-25
UI Package Deploy Routing Refactor

Noxsios:
## Description

This change completely reworks the navigation flow for deploying packages via the UI.

The current method is hardcoded to be `/initialize/{operation}`, the new flow enables for **both** the init package, as well as any package to be deployed via the UI utilizing the same flow.

1. UI opens to `/`
2. If uninitialized, user clicks `Initialize Cluster` link
3. --> `/package/deploy?path=init` (for init **ONLY**, the path provided is `init`, otherwise path is a URL encoded full path to a zarf package)
4. UI queries backend `/packages/find-init` and returns init package data
5. Package data is loaded into `pkgStore`
6. If no errors are encountered --> `/package/[name]/configure` where `[name]` is a URI encoded string from the package name; e.g. `zarf-package-init-package-git-server-arm64.tar.zst --> /package/init-package-git-server-arm64/configure`
7. Deployment flow from here is unchanged as the previous flow was 90% ready as a generic package deployment flow (still some work to be done with this flow, but for now it works well enough)

This opens the door to a potential use case where a user might only open the UI to deploy a single package, then close it, ie: `zarf package deploy [name].tar.zst --open-ui`, which would provide a URL: `http://localhost:{port}/deploy?path={path to package}`

## Type of change

## Checklist before merging


Noxsios:
https://user-images.githubusercontent.com/50058333/214733856-197efb76-c77c-460f-afe9-2f05f7d4614d.mp4

src/ui/routes/package/[name]/configure/+page.svelte
```diff
@@ -53,5 +54,5 @@
 
 <section class="actionButtonsContainer" aria-label="action buttons">
 	<Button href="/" variant="outlined" color="secondary">cancel deployment</Button>
-	<Button href="/initialize/review" variant="raised" color="secondary">review deployment</Button>
+	<Button href={`/package/${$page.params.name}/review`} variant="raised" color="secondary">review deployment</Button>
```
Comment (jeff-mccoy) [inline src/ui/routes/package/[name]/configure/+page.svelte:57]: why are we using path params in one spot and query params in another?

src/ui/routes/+page.svelte
```diff
@@ -34,13 +34,14 @@
 				</Typography>
 			{/if}
 
-			<Button variant="raised" color="secondary" href="/initialize/configure" id="init-cluster">
+			<Button variant="raised" color="secondary" href="/package/deploy?path=init" id="init-cluster">
```
jeff-mccoy, line 37: wouldn't this be `/package/init/deploy` to be consistent?

src/ui/routes/package/deploy/+page.svelte
```diff
@@ -0,0 +1,70 @@
...
+	let errMessage: string = '';
+	const pkgPath = $page.url.searchParams.get('path');
+	let pkgName: string;
+
+	if (pkgPath === null) {
+		errMessage = 'No package path provided';
+		status = LoadingStatus.Error;
+	} else {
+		pkgName = encodeURIComponent(
+			pkgPath
```
jeff-mccoy, line 27: this whole thing is super janky, would much rather see this an actual regex (what it's pretending to be rn)

src/ui/routes/package/deploy/+page.svelte
```diff
@@ -0,0 +1,70 @@
...
+				?.split('/')
+				.at(-1)
+				?.replaceAll('zarf-package-', '')
+				.replaceAll('.tar', '')
+				.replaceAll('.zst', '') ?? ''
```
jeff-mccoy, line 32: Related comment, but don't think I've ever seen anyone actually use `??` in a production codebase before.  There is almost always a more readable way to write things.

src/ui/routes/package/deploy/+page.svelte
```diff
@@ -0,0 +1,70 @@
...
+		);
+
+		if (pkgPath === 'init') {
+			Packages.findInit()
+				.catch(async (err: Error) => {
```
jeff-mccoy, line 37: why is the catch callback async? There is nothing async about that and adds a needless wrapper.  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch

Also I notice here we're doing `promise.catch.then` and below `promise.then.catch`, recommend we stay consistent (and I usually see `promise.then.catch.finally` as seen here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally.

src/ui/routes/package/deploy/+page.svelte
```diff
@@ -0,0 +1,70 @@
...
+							.then(() => {
+								status = LoadingStatus.Success;
+							});
+					}
+				});
+		} else {
+			Packages.read(pkgPath)
+				.then(pkgStore.set)
+				.then(() => {
+					status = LoadingStatus.Success;
```
jeff-mccoy, line 54: nit but these `promise.then` and `promise.catch` would be more readable as functions similar to how `pkgStore.set` was done.

jeff-mccoy (commented):
left some notes

src/ui/routes/+page.svelte
Madeline-UX: Agree with need to be consistent

Madeline-UX (approved):
Flow and url paths look good

src/ui/routes/+page.svelte
Noxsios: resolved with latest changes

jeff-mccoy (approved):
lgtm

PR #1191, opened 2023-01-12
Fixing remote resgistry insecure issues [fix for #1181, #1190, #561, #1015]

rishi-anand:
## Description

Adding support for insecure flag for external registry in `zarf init`. Fixing `zarf package create` and `zarf package deploy` failures with external self signed cert registry as mentioned in https://github.com/defenseunicorns/zarf/issues/1190 and https://github.com/defenseunicorns/zarf/issues/1181

## Related Issue

Fixes # https://github.com/defenseunicorns/zarf/issues/1190
Fixes # https://github.com/defenseunicorns/zarf/issues/1181
Fixes # https://github.com/defenseunicorns/zarf/issues/561
Fixes # https://github.com/defenseunicorns/zarf/issues/1015

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist before merging

- [ ] Tests have been added/updated as necessary (add the `needs-tests` label)
- [ ] Documentation has been updated as necessary (add the `needs-docs` label)
- [ ] An ADR has been written as necessary (add the `needs-adr` label) [ [1](https://github.com/joelparkerhenderson/architecture-decision-record) [2](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions) [3](https://adr.github.io/) ]

Issue #1190, opened 2023-01-11
Zarf init support external registry but doesn't accept any insecure flag for registry


rishi-anand:
Context
We are trying to leverage an external registry that has a self-signed certificate while deploying package via zarf but while doing zarf init, zarf-agent images doesn't gets uploaded to registry and it throws 'Certifficate Signed by unknown authority'. Also zarf init doesn't accept --insecure flag for registry.
In version 0.22.0 zarf init operation with internal registry wasn't failing even if image push fails. This seems to be a new behavior in v0.23.0, possibly because refactor in image push changes, which is the right behavior.

![image](https://user-images.githubusercontent.com/9067666/211942285-a5161244-4249-4bd2-93f6-d3c6f061ec14.png)

### Environment
Device and OS: MacOS, Ubuntu22.04
App version: 0.23.3
Kubernetes distro being used:
Other:

### Steps to reproduce

- Zarf init using an external registry `zarf init --registry-push-password=abcd --registry-push-username=abcd --registry-url=abcd.registry.local` and abcd.registry.local is a local registry with the self-signed certificate.
- zarf-init fails in 0.23.0 with error

### Expected result
Images in init package should get uploaded to external registry and zarf init should get completed successfully.

### Actual Result
```
  ⠼  Updating image ghcr.io/defenseunicorns/zarf/dev-agent:e32f41ab50f994302614adf62ab6f13a7ecfbb25 (1s)
     ERROR:  unable to deploy all components in this Zarf Package: unable to deploy component zarf-agent: unable to deploy component zarf-agent: unable to push images to the registry: Get "https://harbor-ehl.spectrocloud.com/v2/": x509: certificate signed by unknown authority
root@rishi-vm:~/work/zarf#
```

### Visual Proof (screenshots, videos, text, etc)
![image](https://user-images.githubusercontent.com/9067666/211942516-d429155c-4493-4c0c-9622-b228676d8e82.png)

### Severity/Priority
4 (on scale of 1-5)

### Additional Context
Building a zarf package using images from a self-signed registry also fails. There are open issues https://github.com/defenseunicorns/zarf/issues/561 and https://github.com/defenseunicorns/zarf/issues/1015 for creating package from self signed repo.

rishi-anand:
**_Update:_** I have added support for insecure flag in zarf init. I will raise PR very shortly for review. 
[...]

Issue #1181, opened 2023-01-09
Not able to deploy package to an external registry with self signed cert


rishi-anand:
### Context
We are trying to leverage an external registry that has a self-signed certificate while deploying package via zarf

### Environment
Device and OS: MacOS, Ubuntu22.04
App version: 0.23.0
Kubernetes distro being used: 
Other:

### Steps to reproduce
1. Zarf init using an external registry `zarf init --registry-push-password=abcd --registry-push-username=abcd --registry-url=abcd.registry.local` and `abcd.registry.local` is a local registry with the self-signed certificate.
2. Zarf init gets executed successfully
3. Run `zarf package deploy zarf-package-abc.tar.zst`

### Expected result
Images should get uploaded to the local registry and the package should get deployed.
### Actual Result
Image upload to registry fails with below error
```
  ✔  Loading the Zarf State from the Kubernetes cluster
  DEBUG   2023-01-09T23:21:13Z  -  images.PushToZarfRegistry(/tmp/zarf-3645564958/images.tar, [minio/console:v0.19.4 minio/operator:v4.4.28 quay.io/minio/minio:RELEASE.2022-05-26T05-48-41Z library/postgres:13 busybox:1.33.1 alpine quay.io/prometheus/prometheus:latest])
└ (/home/runner/work/zarf/zarf/src/internal/packager/deploy.go:340)
  DEBUG   2023-01-09T23:21:13Z  -  strconv.Atoi: parsing "": invalid syntax
└ (/home/runner/work/zarf/zarf/src/internal/images/push.go:29)
  DEBUG   2023-01-09T23:21:13Z  -  crane pushOptions = (crane.Option)(0x202c1a0)
└ (/home/runner/work/zarf/zarf/src/internal/packager/deploy.go:340)
  ⠋  Updating image minio/console:v0.19.4 (0s)  DEBUG   2023-01-09T23:21:13Z  -  crane.Push() /tmp/zarf-3645564958/images.tar:minio/console:v0.19.4 -> harbor-ehl.spectrocloud.com/minio/console-2853110726:v0.19.4)
└ (/home/runner/work/zarf/zarf/src/internal/packager/deploy.go:340)
  ⠦  Updating image minio/console:v0.19.4 (1s)
  DEBUG   2023-01-09T23:21:15Z  -  Get "https://harbor-ehl.spectrocloud.com/v2/": x509: certificate signed by unknown authority
[...]

Issue #1015, opened 2022-11-16
unable to create a zarf package containing images from a registry with a self-signed certificate


blancharda:
### Context
We are attempting to package a helm chart which uses an image from a private internal docker registry. The registry is configured with a self-signed cert. Our local development environments are configured to properly trust and verify the cert, we are able to interact with the registry using (secure) docker commands.

### Error Case
`zarf package create` fails to pull images -- even when using the `--insecure` flag.
This may also impact genuinely insecure registries (configured entirely without https), though that is untested.

### Environment
Device and OS: MacOS, Ubuntu22.04
Zarf version: `v0.22.2`

### Steps to reproduce
Run `zarf package create --insecure -l trace` on a package containing an image from a registry with an unverified or otherwise invalid cert.

### Expected result
Zarf/Crane will bypass any certificate validation errors while pulling images due to the `--insecure` flag, resulting in a successful zarf package creation..

### Actual Results
*MacOS*:
```
...
  ✔  Loading 4 K8s manifests
  DEBUG   Get "https://harbor.valkyrie.lan/v2/": x509: “Valkyrie” certificate is not standards compliant; Get "https://harbor.valkyrie.lan:443/v2/": x509: “Valkyrie” certificate is not standards compliant
└ (/home/runner/work/zarf/zarf/src/internal/message/spinner.go:116)
     ERROR:  Unable to pull the image "harbor.valkyrie.lan/valkyrie/arkime3:latest"
└ (/home/runner/work/zarf/zarf/src/internal/images/pull.go:48)
```
On MacOS, I believe this specific error is related to a [golang issue](https://github.com/golang/go/issues/51991) involving a change to the crypto package. Additional testing indicated this was not the root of the problem -- to validate, we also tested on Ubuntu, shown below.

*Ubuntu22.04*:
```
...
  ✔  Loading 4 K8s manifests
  DEBUG   Get "https://harbor.valkyrie.lan/v2/": x509: certificate is not valid for any names, but wanted to match harbor.valkyrie.lan; Get "https://harbor.valkyrie.lan:443/v2/": x509: certificate is not valid for any names, but wanted to match harbor.valkyrie.lan
└ (/home/runner/work/zarf/zarf/src/internal/message/spinner.go:116)
     ERROR:  Unable to pull the image "harbor.valkyrie.lan/valkyrie/arkime3:latest"
└ (/home/runner/work/zarf/zarf/src/internal/images/pull.go:48)
```

[...]

PR #1280, opened 2023-01-28
add common insecure flag

jeff-mccoy:
Migrated from #1191

## Description
Adding support for insecure flag for external registry in `zarf init`. Fixing `zarf package create` and `zarf package deploy` failures with external self signed cert registry as mentioned in https://github.com/defenseunicorns/zarf/issues/1190 and https://github.com/defenseunicorns/zarf/issues/1181

## Related Issue

Fixes # https://github.com/defenseunicorns/zarf/issues/1190
Fixes # https://github.com/defenseunicorns/zarf/issues/1181
Fixes # https://github.com/defenseunicorns/zarf/issues/561
Fixes # https://github.com/defenseunicorns/zarf/issues/1015


src/cmd/root.go
```diff
@@ -80,6 +81,7 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&message.NoProgress, "no-progress", v.GetBool(V_NO_PROGRESS), lang.RootCmdFlagNoProgress)
 	rootCmd.PersistentFlags().StringVar(&config.CommonOptions.CachePath, "zarf-cache", v.GetString(V_ZARF_CACHE), lang.RootCmdFlagCachePath)
 	rootCmd.PersistentFlags().StringVar(&config.CommonOptions.TempDirectory, "tmpdir", v.GetString(V_TMP_DIR), lang.RootCmdFlagTempDir)
+	rootCmd.PersistentFlags().BoolVar(&config.CommonOptions.Insecure, "insecure", v.GetBool(V_INSECURE), lang.RootCmdFlagInseure)
```
Racer159, line 84: Can we add a `-k` alias like we had discussed earlier in the week to signal to folks this more closely follows `curl`s paradigms? 

```suggestion
	rootCmd.PersistentFlags().BoolVarP(&config.CommonOptions.Insecure, "insecure", "k", v.GetBool(V_INSECURE), lang.RootCmdFlagInseure)
```

Racer159 (changes requested):
One small thing we had talked about earlier

src/config/lang/english.go
```diff
@@ -37,6 +37,7 @@ const (
 	RootCmdFlagNoProgress  = "Disable fancy UI progress bars, spinners, logos, etc"
 	RootCmdFlagCachePath   = "Specify the location of the Zarf cache directory"
 	RootCmdFlagTempDir     = "Specify the temporary directory to use for intermediate files"
+	RootCmdFlagInseure     = "Allow access to insecure registries and disable other recommended security enforcements. This flag should only be used if you have a specific reason and accept the reduced security posture."
```
Racer159, line 40: Not calling this out in the requested changes, but wanted to add a comment for @Madeline-UX to look at this verbiage to make sure it makes sense. `--insecure` here is doing two things:

1. Allowing connections to `http` image registries on `create` or `deploy`
2. Disabling Transport Layer Security (TLS) verification for connections to `https` registries

Technically both of these apply only to image registries at the moment and I do feel this message may require more balancing between `create` personas (who know what http/tls are) and `deploy` personas who may not.

jeff-mccoy: Yeah I agree, I just left it generic as "other security enforcements" because of that.

Racer159 (approved):
lgtm

src/config/lang/english.go
Madeline-UX: Would it be better to specify which security enforcements it disables (the TLS)? That way, it's specific and can be researched if needed. 

I am concerned that the vagueness of "and other security enforcements" may keep users from being able to accurately assess the risk.  Unless we are going for scary - vague. Don't do this ... unless you really need to.

Madeline-UX: @Racer159 @jeff-mccoy I could do a quick async UX test of the language - I need to know what we are wanting to test though. 

We assume that the creator understands and knows what these terms mean
We are not sure if the deploy persona "Ashton does?"

YrrepNoj (approved):

Racer159 (approved):
lgtm (again)

YrrepNoj (dismissed):

YrrepNoj (approved):

Racer159 (approved):
sad pandas on -k, but lgtm

PR #1297, opened 2023-01-30
Dep updates

jeff-mccoy:
## Type of change

## Checklist before merging


Racer159 (approved):
lgtm (if ci passes)

.github/workflows/test-ui.yml
```diff
@@ -42,6 +42,9 @@ jobs:
           path: "~/Library/Caches/ms-playwright/"
```
Noxsios, line 42: Is this path correct? https://github.com/defenseunicorns/zarf/actions/runs/4040760238/jobs/6946708240#step:8:19

docs-website/package.json
```diff
@@ -49,5 +37,8 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "engines": {
+    "node": ">=16.14"
```
Noxsios, line 42: If we are using Node 18 in CI, should we specify Node 18+ here as well?

jeff-mccoy: removed

.github/workflows/test-ui.yml
jeff-mccoy: updated

Racer159 (changes requested):

.github/workflows/release.yml
```diff
@@ -70,7 +71,7 @@ jobs:
 
       # Set up AWS credentials for GoReleaser to upload backups of artifacts to S3
       - name: Set AWS Credentials
-        uses: aws-actions/configure-aws-credentials@v1
+        uses: aws-actions/configure-aws-credentials@v1-node16
```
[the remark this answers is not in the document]
jeff-mccoy: That's what their docs offer--the default is actually node 12, but they added 16 in response to a gh deprecation warning.

https://github.com/aws-actions/configure-aws-credentials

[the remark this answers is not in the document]
jeff-mccoy: It has nothing to do with our use of node, it's what that action is written in as actions can be written in js.

Racer159 (approved):
lgtm

PR #1299, opened 2023-01-30
Refactor git repository caching logic

### LINKED ISSUE #1069: Potential for 'stale' git repo stash
If a Zarf Package does not specify a commit/tag for a repo, Zarf will include the entire repository with the package. If the upstream repository gets updates and you try to rebuild the package Zarf will fetch the changes but won't change the HEAD ref.

This results in the package being successfully created but the ref's of the repository will be weird so when you try to deploy the package you run into an internal error of `cannot lock ref 'refs/heads/master': reference already exists` when that repository is getting pushed to the Zarf Repository.

Desired Solution: 
The cache should have the latest upstream changes.

---

> Old cache AFTER `zarf package create`
Screen Shot 2022-12-02 at 12 23 01 PM

> Bad push for repo with old cache
Screen Shot 2022-12-02 at 12 23 14 PM

### LINKED ISSUE #1302: Additional git tags are produced when using repos from cache
When building a Zarf package that specifies a specific tag for a git repository, we expect ONLY that tag to be included within the package.

There is a bug right now where additional tags are being produced if the repo already exists within the zarf cache.

### LINKED ISSUE #1303: Git repo checksums are incorrectly including the specified references
We are using the references for git repos (hashes/tags) when producing the checksum for our git repositories. This is undesirable because that means `https://github.com/defenseunicorns/zarf.git` and `https://github.com/defenseunicorns/zarf.git@v0.15.0` will have different checksums. This results in duplicated caches and other potential issues.

The reference tag/hash should not be included in the checksum.

YrrepNoj:
## Description

MISC:
- Added more error handling to our git functions (one more step in our path away from using `fatal`.
---
Fixes #1069
- Fix a bug where cached repos HEADs were not getting updated when the upstream repo had new changes (we were fetching the changes but not updating the branch) so they were not getting deployed correctly
 ---
Fixes #1302
- Fix bug where additional unwanted tags where getting fetched when we were updating the HEAD of repos in `~/.zarf-cache/repos`
---
Fixes #1303
- Change how we calculate the checksums of git repos
- NOTE: This can be considered a breaking change...
---


src/internal/packager/git/checkout.go
```diff
@@ -21,28 +21,42 @@ func (g *Git) CheckoutTag(tag string) {
...
 // than checking out to a detached head, checks out to the provided branch ref
 // It will delete the branch provided if it exists.
-func (g *Git) checkoutTagAsBranch(tag string, branch plumbing.ReferenceName) {
+func (g *Git) checkoutTagAsBranch(tag string, branch plumbing.ReferenceName) error {
 	message.Debugf("git.checkoutTagAsBranch(%s,%s)", tag, branch.String())
 
 	repo, err := git.PlainOpen(g.GitPath)
 	if err != nil {
-		message.Fatal(err, "Not a valid git repo or unable to open")
+		message.Warnf("Not a valid git repo or unable to open")
```
jeff-mccoy, line 44: Should be using `message.Warf()` here or just return a formatted error like we do most other places in the codebase?

src/internal/packager/git/common.go
```diff
@@ -23,6 +25,8 @@ const onlineRemoteName = "online-upstream"
 const offlineRemoteName = "offline-downstream"
 const onlineRemoteRefPrefix = "refs/remotes/" + onlineRemoteName + "/"
 
+var IsHash = regexp.MustCompile(`^[0-9a-f]{40}$`).MatchString
```
jeff-mccoy, line 28: lint error

src/internal/packager/git/fetch.go
```diff
@@ -15,35 +15,49 @@ import (
...
 
-	err := g.fetch(g.GitPath, refspec)
+	return err
+}
 
-	if err != nil {
-		message.Fatal(err, "Not a valid tag or unable to fetch")
+// fetchTag performs a `git fetch` of _only_ the provided tag.
+func (g *Git) fetchTag(tag string) error {
+	message.Debugf("git.fetchTag(%s)", tag)
```
jeff-mccoy, line 33: I thought we decided we weren't adding debug message anymore?

src/internal/packager/git/fetch.go
```diff
@@ -15,35 +15,49 @@ import (
...
+	fetchOptions := &git.FetchOptions{
+		RemoteName: onlineRemoteName,
+		RefSpecs:   []goConfig.RefSpec{goConfig.RefSpec("refs/tags/" + tag + ":refs/tags/" + tag)},
```
jeff-mccoy, line 36: nit, would be good to see this string fun as a formatted string variable

src/internal/packager/git/fetch.go
```diff
@@ -15,35 +15,49 @@ import (
...
 
-	refspec := goConfig.RefSpec(hash + ":" + hash)
-
-	err := g.fetch(g.GitPath, refspec)
-
-	if err != nil {
-		message.Fatal(err, "Not a valid hash or unable to fetch")
+	fetchOptions := &git.FetchOptions{
+		RemoteName: onlineRemoteName,
+		RefSpecs:   []goConfig.RefSpec{goConfig.RefSpec(hash + ":" + hash)},
```
jeff-mccoy, line 50: string variable

src/internal/packager/git/pull.go
```diff
@@ -66,23 +68,42 @@ func (g *Git) pull(gitURL, targetFolder string, repoName string) {
...
+		if err != nil {
+			message.Debugf("unable to get the worktree for the repo: %s", gitURL)
+			return err
+		}
+		err = worktree.Pull(pullOptions)
 		if errors.Is(err, git.NoErrAlreadyUpToDate) {
 			message.Debug("Repo already up to date")
 		} else if err != nil {
-			g.Spinner.Fatalf(err, "Not a valid git repo or unable to fetch")
+			g.Spinner.Warnf("Not a valid git repo or unable to pull: %s", gitURL)
```
jeff-mccoy, line 88: same comment warn vs return formatted error

src/internal/packager/git/refs.go
```diff
@@ -23,6 +23,32 @@ func (g *Git) removeLocalBranchRefs() ([]*plumbing.Reference, error) {
 	)
 }
 
+// removeLocalTagRefs removes all tags in the local repo.
+// It returns a slice of tags deleted.
+func (g *Git) removeLocalTagRefs() ([]string, error) {
```
jeff-mccoy, line 28: do we have any testing for this logic?

jeff-mccoy (commented):
probably need to make sure we're testing these different implementations.

src/internal/packager/git/common.go
YrrepNoj: @jeff-mccoy 
What lint error are you seeing?

And how are you seeing it? I might not have something set up correctly b/c I'm not seeing it right now.

src/internal/packager/git/fetch.go
YrrepNoj: I didn't actually change this line but git thinks I did because of other changes above/below this so I didn't catch that. I'll remove it now :)

src/internal/packager/git/common.go
jeff-mccoy: Screenshot 2023-02-01 at 5 47 22 PM

src/test/e2e/22_git_and_flux_test.go
```diff
@@ -24,27 +25,29 @@ func TestGitAndFlux(t *testing.T) {
 	e2e.setupWithCluster(t)
 	defer e2e.teardown(t)
 
-	path := fmt.Sprintf("build/zarf-package-git-data-%s-v1.0.0.tar.zst", e2e.arch)
+	// path := fmt.Sprintf("build/zarf-package-git-data-%s-v1.0.0.tar.zst", e2e.arch)
```
jeff-mccoy, line 28: ?

YrrepNoj: If you comment out all the tests, they can't fail.

jeff-mccoy (approved):

PR #1307, opened 2023-01-31
Update Readme - K8 Slack link update

Madeline-UX:
## Description
Needed to update K8 slack link. The old one took users to a page that was only accessible if you had an account already
...

## Related Issue

Fixes #

## Type of change

## Checklist before merging


jeff-mccoy:
this change breaks auto-opening the slack channel for me, is there more info behind this change?

Madeline-UX:
@jeff-mccoy Uhg. So if you have an account it breaks the link. But if you do not yet have a K8 account then it takes you to a login page you cannot use to create an account. 

Given that they users of this page are more technical and likely have a K8 account already then I think leaving as is is fine

There is the link to create an account on the website

PR #1314, opened 2023-01-31
Update FAQ - Add license  information

Madeline-UX:
"What license does Zarf fall under?"

 is a common question we get when meeting with potential users/customers.


Noxsios (approved):
lgtm

docs/9-faq.md
```diff
@@ -1,13 +1,18 @@
 # FAQ
 
-## Do I have to use [Homebrew](https://brew.sh/) to install Zarf?
+## Do I have to use Homebrew to install Zarf?
```
jeff-mccoy, line 3: Any particular reason (besides style preference) to remove the link to homebrew?

Madeline-UX: Having a link in a header is usually frowned upon. If we want to link to homebrew i think we should include it in the description below. 

However, if a user is asking if they can use something other than homebrew, I think we can assume that the user is not interested int he homebrew link at that time.

jeff-mccoy (approved):

PR #1317, opened 2023-02-01
Smarter Zarf Registry HPA behavior

jeff-mccoy:
Make Registry HPA less aggressive and block from downscaling while running `zarf package deploy` if images exist.

## Type of change

## Checklist before merging


jeff-mccoy:
During deploy, scaleDown disabled:
Screenshot 2023-02-01 at 7 20 27 PM

After deploy, scaleDown reset:
Screenshot 2023-02-01 at 7 20 33 PM

Racer159 (changes requested):

src/pkg/packager/deploy.go
```diff
@@ -68,6 +71,9 @@ func (p *Packager) Deploy() error {
 	// Note: Not all packages need k8s; check if k8s is being used before saving the secret
 	if p.cluster != nil {
 		p.cluster.RecordPackageDeployment(p.cfg.Pkg, deployedComponents)
+		if hpaModified {
```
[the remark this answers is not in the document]
jeff-mccoy: no I need to address that, thx

src/pkg/packager/deploy.go
```diff
@@ -216,6 +222,15 @@ func (p *Packager) deployComponent(component types.ZarfComponent, noImgChecksum
 			}
 		}
 
+		// Disable the registry HPA scale down if we are deploying images and it is not already disabled
+		if hasImages && !hpaModified {
+			if err := p.cluster.EnableRegistryHPAScaleDown(false); err != nil {
```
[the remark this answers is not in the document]
jeff-mccoy: that's a really good point, will change

src/internal/cluster/zarf.go
```diff
@@ -98,3 +99,30 @@ func (c *Cluster) RecordPackageDeployment(pkg types.ZarfPackage, components []ty
 
 	c.Kube.CreateOrUpdateSecret(deployedPackageSecret)
 }
+
+// EnableRegistryHPAScaleDown enables or disables the HPA scale down for the Zarf Registry.
+func (c *Cluster) EnableRegistryHPAScaleDown(enable bool) error {
```
YrrepNoj, line 104: Nit; Having the function being named `EnableRegistryHPAScaleDown()` but then accepting a Boolean that affects if HPA is scaled down or not seems like an antipattern (or at least a bad name 😛)

Thoughts on changing the function header to something along the liens of `ConfigureRegistryHPAScaleDown(policy v2beta2.ScalingPolicySelect)` ?

YrrepNoj (dismissed):
I have a single 'nit' over a function name.

Since it seems like just a nit, I'll approve the PR and leave resolving that to your judgement :)

src/internal/cluster/zarf.go
Racer159: @jeff-mccoy and I don't like the word "toggle".  That implies that it remembers state and flips back and forth when you call it.  I would prefer Set or Configure personally

wirewc: @Racer159 I think toggle works great as most of the time you want it enabled to ensure a single core isn't getting hammered. Seeing that the issue originally reported by @corang shows that problem was an edge case for a specific deployment originally. Seeing the scope (as I understand it) is to temporarily change the behavior of HPA when deploying large packages, the word `toggle` seems accurate.

jeff-mccoy: looked at different strategies and ended up just doing to diff functions. Not as DRY, but more readable I suppose...

`EnableRegHPAScaleDown()`
`DisableRegHPAScaleDown()`

Screenshot 2023-02-03 at 12 12 07 PM

src/internal/cluster/zarf.go
```diff
@@ -98,3 +99,41 @@ func (c *Cluster) RecordPackageDeployment(pkg types.ZarfPackage, components []ty
 
 	c.Kube.CreateOrUpdateSecret(deployedPackageSecret)
 }
+
+// SetRegHPAScaleDown enables or disables the HPA scale down for the Zarf Registry.
+func (c *Cluster) EnableRegHPAScaleDown() error {
```
YrrepNoj, line 104: The docstring comment title got lost as we were experimenting with function names

YrrepNoj (approved):

Racer159 (approved):
Lgtm

PR #1331, opened 2023-02-03
1322 component tarballs

YrrepNoj:

PR #1336, opened 2023-02-06
Create a `zarf package publish` command

### LINKED ISSUE #381: Support Remote Package for Composable Packages
In order to expand on the current composability of zarf packages the user should be able to pull zarf.yaml and dependencies from the web.  Consideration should be taken for how versioning would be handled in addition to how a SHASUM or signature configuration will be required.

### LINKED ISSUE #823: Composable Packages from a URL
I'd like to build a Zarf package referencing a package via URL.

```
  - name: games
    description: "Games package"
    import:
      path: https://raw.githubusercontent.com/defenseunicorns/zarf/master/examples/game/
...
```

**Additional context**
Short slack discussion [here](https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1664653708490319)

Noxsios:
### 💲 `zarf package publish`

This PR adds a new `zarf package publish` command.

This command will publish a fully built (via `zarf package create`) Zarf package to an OCI compliant registry using the oras library.

The command's usage is based off of how Helm's `helm push` command operates and has many of the same restrictions.

<https://v3.helm.sh/docs/topics/registries/#the-push-subcommand>

```
$ zarf package publish [PATH_TO_TAR] [REGISTRY_URL]

$ zarf package publish zarf-package-strimzi.tar oci://localhost:666/defenseunicorns
```

#### CLI Syntax restrictions:

- The OCI URL _must_ start with `oci://`
- A basename/version _cannot_ be provided, as it is derived from information present in the built `zarf.yaml` where it _must_ be set.

- The command may return an error, permission denied, if the repository does not exist, ie `defenseunicorns/strimzi` as an example on Docker Hub.
- A `zarf tools registry login` must be done prior as this will inherit the credentials from docker's cred system (defaultly located at `~/.docker`), or users can create a Docker compatible `config.json` and point its directory:

```sh
# example symlinking podman
$ mkdir ~/.docker/
$ ln -s $XDG_RUNTIME_DIR/containers/auth.json ~/.docker/config.json

$ zarf package publish <...>
```

### 💲 `zarf package deploy oci://`

This PR also adds a new `deploy oci://` feature.

This command uses the existing `deploy` system, but does some slightly different behavior from `sget://`. oras is used to pull the package layer by layer to the temp directory. There is no need to unarchive, as the package is already in an uncompressed state. The temp path is then used by the packager to deploy the package as though it is a local package. Usage is denoted in the following example:

```
$ zarf package deploy oci://REGISTRY/NAMESPACE/NAME:VERSION

$ zarf package deploy oci://docker.io/defenseunicorns/strimzi:v0.24.0-arm64

$ zarf package deploy oci://localhost:666/strimzi:v0.24.0-arm64 --insecure
```

### 💲 `zarf package inspect oci://`

This PR also adds a new `inspect oci://` feature.

```
$ zarf package inspect oci://docker.io/defenseunicorns/strimzi

tags:
- v0.23.5-14-arm64
latest:
  tag: v0.23.5-14-arm64
  descriptor:
    mediaType: application/vnd.oci.image.manifest.v1+json
    digest: sha256:340f489a105e476f846203e6844b2738f2924fc608522711761a21b176d6b67f
    size: 41817
```

## Related Issue

Relates to #1298 

Fixes #381
Fixes #823

Blocked by #1331 

## Type of change

## Checklist before merging


src/pkg/packager/publish.go
```diff
@@ -0,0 +1,226 @@
...
+// - This is a WIP, not yet functional
+// - This is a copy of the oras example code, with some modifications
+//
+// replace comments w/ CLI args where appropriate
+// replace fmt.Printf w/ message.Debug
+
+var zarfMediaType = "application/vnd.zarf.layer.v1+tar.zst"
+
+func (p *Packager) Publish() error {
+	registry := "localhost:666"
```
Racer159, line 30: can we change this to something like localhost:443 instead?

Racer159 (changes requested):

src/pkg/packager/publish.go
Noxsios: was a holdover during transfer, disregard

wirewc:
For podman, make sure the following exists for the `docker login` to work as expected for podman. 

```bash
mkdir ~/.docker/
ln -s $$XDG_RUNTIME_DIR/containers/auth.json ~/.docker/config.json
```
I don't know if we should add that into a help doc or message, or if podman should just be that much of an outlier to ignore it.

Noxsios:
TODO: 2/9

- Create a `package pull` command that is able to pull a Zarf pkg out of an OCI registry
- Create the `skeleton` OCI artifact for pulls with composability
- Create an OCI `config.json` object during `publish` so that GCR does not get angry (and to be more in line w/ the OCI spec anyways)

Noxsios:
Awareness from @jeff-mccoy https://github.com/docker/docs/issues/8230

src/cmd/package.go
```diff
@@ -195,6 +195,38 @@ var packageRemoveCmd = &cobra.Command{
...
+	Example: "  zarf package publish my-package.tar oci://my-registry.com/my-namespace",
+	Args: cobra.ExactArgs(2),
+	Run: func(cmd *cobra.Command, args []string) {
+		pkgConfig.PublishOpts.PackagePath = choosePackage(args)
+
+		if !strings.HasPrefix(args[1], "oci://") {
+			message.Fatalf(nil, "Registry must be prefixed with 'oci://'")
+		}
+
+		registry := args[1]
```
Racer159, line 210: I would prefer this to be above the if so that the if statement is more readable:
```suggestion
		registry := args[1]
		
		if !strings.HasPrefix(registry, "oci://") {
			message.Fatalf(nil, "Registry must be prefixed with 'oci://'")
		}
```

src/pkg/packager/publish.go
```diff
@@ -0,0 +1,369 @@
...
+func (p *Packager) Publish() error {
+	p.cfg.DeployOpts.PackagePath = p.cfg.PublishOpts.PackagePath
+	if err := p.loadZarfPkg(); err != nil {
+		return fmt.Errorf("unable to load the package: %w", err)
+	}
+	spinner := message.NewProgressSpinner("")
+	defer spinner.Stop()
+
+	if p.cfg.PublishOpts.RegistryURL == "docker.io" {
+		// docker.io is commonly used, but is not a valid registry URL
```
Racer159, line 46: Why are we rewriting this to `registry-1`?

src/pkg/packager/publish.go
```diff
@@ -0,0 +1,369 @@
...
+		filepath.Join(p.tmp.Base, "checksums.txt"),
+		filepath.Join(p.tmp.Base, "zarf.yaml"),
+		filepath.Join(p.tmp.Base, "sboms.tar.zst"),
+	}
+	componentTarballs, err := filepath.Glob(filepath.Join(p.tmp.Base, "components", "*.tar.zst"))
+	if err != nil {
+		return err
+	}
+	paths = append(paths, componentTarballs...)
+	if p.cfg.PublishOpts.IncludeImages {
```
Racer159, line 61: I would rather handle this in a smarter way than including the images and failing if there are too many layers.  We could either include it as another artifact (tagging them differently in a way zarf could recombine them) or we could push the images back to the registry as they are (though this would require some rebuild logic to get them back into a package). @bburky @jeff-mccoy have some ideas.

Racer159: We can also split image handing into a separate issue since skelly packages do have value alone.

src/pkg/packager/publish.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+
+	if err = push(root); err != nil {
+		return err
+	}
+	spinner.Updatef("Published: %s [%s]", ref, root.MediaType)
+	message.SuccessF("Published: %s [%s]", ref, root.MediaType)
+	message.SuccessF("Digest: %s", root.Digest)
+	return nil
+}
```
Racer159, line 331: This function is several hundred lines long, I would recommend splitting it into at least private helper functions or if they are generic enough helper functions in the utils package.

Racer159 (changes requested):

src/pkg/packager/publish.go
Noxsios: This is due to `docker.io` not being the resolved registry URL when you perform a docker push. While `docker.io` is a popular alias, the real registry address is `registry-1`. This is my experience, but would love to be shot down (currently the publish will error due to this behavior.

src/pkg/packager/publish.go
Noxsios: Did some breakup of the big `publish` fn. Will take another look tonight.

src/pkg/packager/publish.go
Racer159: Yeah that is interesting: https://docs.docker.com/registry/introduction/

Noxsios: https://github.com/docker/cli/blob/master/man/src/image/pull.md

src/pkg/packager/publish.go
Noxsios: By default, only the skelly package is created, users must specify `--include-images` for the images to be included.  Once a determination is made on how we want to handle chunking, we can invert that flag.

Noxsios:
Note:

On one of the test pushes to ECR w/ many (150+ layers) there was a `504: Gateway TImeout` error. Since resumable uploads is part of the Docker+OCI spec, rerunning the `package publish` resumed successfully and the process continued. Might be worth leaving a note / handling.

Racer159:
Gonna dive into code in a bit, but added the needs_tests label to this for now.  A potentially easy(ish) way to test may be to publish a package to the zarf deployed registry and then deploy that same package back to the zarf init'ed cluster

src/pkg/packager/network.go
```diff
@@ -111,3 +119,51 @@ func (p *Packager) handleSgetPackage() error {
...
 	return nil
 }
+
+func (p *Packager) handleOciPackage() error {
+	message.Debug("packager.handleOciPackage()")
+	ref, err := registry.ParseReference(strings.TrimPrefix(p.cfg.DeployOpts.PackagePath, "oci://"))
+	if err != nil {
+		return fmt.Errorf("failed to parse OCI reference: %w", err)
+	}
+	// patch docker.io to registry-1.docker.io
```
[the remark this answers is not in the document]
YrrepNoj: Also can we double check if this is necessary to do here? It seems like this is also done in a couple nested calls when we run `p.pullOCIZarfPakage(....)` a few lines down.

src/pkg/packager/network.go
```diff
@@ -111,3 +119,51 @@ func (p *Packager) handleSgetPackage() error {
...
+	spinner.Successf("Pulled %s", ref.String())
+
+	p.cfg.DeployOpts.PackagePath = out
+	return nil
+}
+
+// isManifestUnsupported returns true if the error is an unsupported artifact manifest error.
+//
+// This function was copied verbatim from https://github.com/oras-project/oras/blob/main/cmd/oras/push.go
+func isManifestUnsupported(err error) bool {
```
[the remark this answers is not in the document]
Noxsios: Since this is a copy from `oras`, I feel it is fine to leave as is so that changes between this and the source can more easily be discerned.

Noxsios:
Going to be adding the tests in the main feature branch in the MZOCI PR: https://github.com/defenseunicorns/zarf/pull/1402

Issue #1337, opened 2023-02-07
Deploy K8s kustomizations Without Wrapping In A Helm Chart


UncleGedd:
**Is your feature request related to a problem? Please describe.**
I'm trying to deploy a Big Bang variant using a kustomization inside of a Zarf package. The Helm chart that Zarf wraps my kustomization with fails to install due a templating error. Note that the kustomization builds correctly with `kustomize build <dir>`
Example of my Zarf pkg
```
kind: ZarfPackageConfig
metadata:
  name: batcave
  architecture: amd64
  yolo: true

components:
- name: flux
  manifests:
  - name: local-flux
    namespace: flux-system
    kustomizations:
    - ./base/flux

- name: batcave
  manifests:
  - name: batcave-dev
    namespace: batcave
    kustomizations: # fails to install!
    - ./envs/dev
```

We have determined that the Helm chart fails to install due to a templating error related to the following values for the [Big Bang Monitoring](https://repo1.dso.mil/big-bang/apps/core/monitoring) chart:
```
    additionalPrometheusRulesMap:
      custom-rules:
        groups:
          - name: OOMKilled
            rules:
              - alert: OOMKilled
                expr: sum_over_time(kube_pod_container_status_terminated_reason{reason="OOMKilled"}[5m])>0
                for: 5m
                labels:
                  severity: critical
                annotations:
                  description: Pod {{$labels.pod}} in {{$labels.namespace}} got OOMKilled
```

Zarf fails with the error message:
```
parse error at (raw-batcave-batcave-batcave-dev/var/folders/sf/_jrk8zgs5tg33r_h1gzh_6t40000gn/T/zarf-1900317816/components/batcave/manifests/kustomization-batcave-dev-0.yaml:1263): undefined variable "$labels"
```

**Describe the solution you'd like**
I'd like the option to install the kustomization directly (like with `kustomization build <dir> | kubectl apply -f -`) instead of wrapping it in a Helm chart. Specifcally, I want to specify in the Zarf package to not wrap the kustomization in a Helm chart

Racer159:
For this (stated to @wirewc) we may want to make this apply for all manifests since that may be easier technically and would provide more flexibility on standard manifests as well

Racer159:
We will also have to think about how this affects the package lifecycle and user messaging on the key (since we use helm right now to uninstall for example and this may break that)

WarheadsSE:
[...]

PR #1347, opened 2023-02-09
1337 kustomize direct

UncleGedd:
## Description

...

## Related Issue

Fixes #

## Type of change

## Checklist before merging

### DIFF
--- a/examples/funny-helm/kustomization/configmap.yaml
+++ b/examples/funny-helm/kustomization/configmap.yaml
@@ -1,0 +1,20 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: monitoring-helmrelease-values-cm
+data:
+  values.yaml: |
+    monitoring:
+      values:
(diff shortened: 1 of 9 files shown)


Racer159:
Assuming this is still a Helm-related issue, can we do something like this instead (either on your side upstream or as a string literal wrapper to turn off Helm processing?)  Likely a much better way than diving into the lifecycle of Zarf:
image

jeff-mccoy:
What issue is this related to? I have concerns about moving away from helm for anything as it's a fundamental change to how we manage k8s resources and a complexity I don't think we can take on.

UncleGedd:
Related issue and more context/conversation: https://github.com/defenseunicorns/zarf/issues/1337

@wirewc and I discussed something similar yesterday. Either:
1. make the change my upsteam Helm chart (as @Racer159 suggests)
2. maybe use a Zarf action to do some string replacement

Racer159:
Closing in favor of #1350

PR #1349, opened 2023-02-10
Hotfix docker CRI support 

jeff-mccoy:
## Description

Fixes injector issues with the docker runtime and the oci seed image.

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging

### DIFF
--- a/src/injector/src/main.rs
+++ b/src/injector/src/main.rs
@@ -118,7 +118,7 @@ fn start_seed_registry() {
                     // on Content-Length, we respond the same as a GET
                     accept!(
                         request,
-                        "application/vnd.oci.image.manifest.v1+json" => {
+                        "application/vnd.docker.distribution.manifest.v2+json" => {
                             handle_get_manifest(&root)
                         },
                         "*/*" => Response::empty_406()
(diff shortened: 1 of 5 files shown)


src/injector/src/main.rs
```diff
@@ -130,7 +130,7 @@ fn start_seed_registry() {
                     // on Content-Length, we respond the same as a GET
                     accept!(
                         request,
-                        "application/vnd.oci.image.manifest.v1+json" => {
+                        "application/vnd.docker.distribution.manifest.v2+json" => {
```
YrrepNoj, line 133: My rust is rusty, but would it make sense to capture/perform on both `...oci.image...` and `...docker.distribution...`?

jeff-mccoy: The change is occurring because we moved to using crane vs a manually hacked source OCI for the seed registry. This passed with containerd, but seems to be sensitive with docker as CRI. I don't think supporting both would be necessary as this is purely an init package change: the rust injector logic + the oci objects in the same package.

YrrepNoj: I was that because when we're [creating the init package](https://github.com/defenseunicorns/zarf/blob/a0b13947b9b345b09ce23e6d848bd587012f1bc3/src/pkg/packager/create.go#L93) we are string replacing the `docker.distribution` with `oci.image`: https://github.com/defenseunicorns/zarf/blob/00736dd9d00d7b24f838612dc8c260dc5c41ee4c/src/internal/packager/images/pull.go#L186.


I think it makes sense to have endpoints that way we can handle both 'request types' but I'm not sure I 100% understand what went wrong here.

jeff-mccoy: It's a result of the change to crane vs hand-rolled OCI image. This is a permanent fix.

YrrepNoj (approved):

PR #1350, opened 2023-02-11
Resolve nested go templating in helm

### LINKED ISSUE #1337: Deploy K8s kustomizations Without Wrapping In A Helm Chart

jeff-mccoy:
## Description

Always wrap manifest/kustomizations in string literals within the helm chart generation process to avoid nested go template messes.

## Related Issue

Fixes #1337

## Type of change

## Checklist before merging


corang:
I think this will break if the templating being escaped has a backtick in it

src/internal/packager/helm/chart.go
```diff
@@ -172,44 +173,57 @@ func (h *Helm) GenerateChart(manifest types.ZarfManifest) (types.ConnectStrings,
...
 		data, err := os.ReadFile(manifest)
 		if err != nil {
 			return nil, "", fmt.Errorf("unable to read manifest file %s: %w", manifest, err)
 		}
+
+		// If the template contains the match "{{", wrap it in {{` and `}} to prevent helm from trying to parse it.
+		if bytes.Contains(data, match) {
+			data = append(prefix, data...)
+			data = append(data, suffix...)
+		}
```
Racer159, line 210: This doesn't account for backticks in the templat itself.

src/test/test-packages/evil-templates/configmap.yaml
```diff
@@ -0,0 +1,20 @@
...
+  values.yaml: |
+    monitoring:
+      values:
+        additionalPrometheusRulesMap:
+          custom-rules:
+            groups:
+              - name: OOMKilled
+                rules:
+                  - alert: OOMKilled {{ "{{ \"random.Values\" }}" }}
+                    expr: sum_over_time(kube_pod_container_status_terminated_reason{reason="OOMKilled"}[5m])>0
```
Racer159, line 15: Ideally this would include backticks to properly test escaping.

Racer159 (changes requested):

src/internal/packager/helm/chart.go
jeff-mccoy: Is there some real-world use case that includes nested templating and backticks? I haven't seen it.

jeff-mccoy: This is a 99% solution. We cannot optimize for every single possibility, real or imagined. The released version of Zarf is broken for nested templating, and since there is an edge case for double-nested templating, this is the best way to resolve it. I have not found any example of nested or multi-nested templating + backtick escaping. It's all generally not great behavior because [kyverno](https://kyverno.io/docs/writing-policies/variables/#variables-in-helm) made a bad choice and decided to use the same templating language as the much more well-established helm.

jeff-mccoy:
@corang please provide a real-world use of a combination of backtick and nested go templating.

src/internal/packager/helm/chart.go
jeff-mccoy: I think this is highly over-optimizing for an unknown edge case but added it while noting my disagreement.

src/internal/packager/helm/chart.go
```diff
@@ -172,44 +173,61 @@ func (h *Helm) GenerateChart(manifest types.ZarfManifest) (types.ConnectStrings,
...
 		}
+
+		// If the template contains the match "{{", wrap it in {{` and `}} to prevent helm from trying to parse it.
+		if bytes.Contains(data, match) {
+			// Replace backticks before wrapping in {{` and `}}.
+			data = bytes.ReplaceAll(data, backtick, backtickReplace)
+			data = append(prefix, data...)
+			data = append(data, suffix...)
+		}
+
```
Racer159, line 215: I personally feel that this would be much easier if this was changed to use something like strconv.Quote https://cs.opensource.google/go/go/+/go1.20:src/strconv/quote.go;l=128 - This would handle nearly all use cases for us without needing to manually do weird things.

```
var(
		prefix          = []byte("{{")
		suffix          = []byte("}}")
)
...
data = strconv.Quote(data)
data = append(prefix, data...)
data = append(data, suffix...)
```

Racer159 (changes requested):

src/internal/packager/helm/chart.go
jeff-mccoy: You can't escape backticks within a backtick so this wouldn't work.

Racer159:
> @corang please provide a real-world use of a combination of backtick and nested go templating.

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: dont-template-me
data:
  values.yaml: |
    monitoring:
      values:
        additionalPrometheusRulesMap:
          custom-rules:
            groups:
              - name: OOMKilled
                rules:
                  - alert: OOMKilled {{ "{{ \"random.Values\" }}" }} {{ `I wanted to use a backtick` }}
                    expr: sum_over_time(kube_pod_container_status_terminated_reason{reason="OOMKilled"}[5m])>0
                    for: 5m
                    annotations: 
                      backtick1: "content with backticks `some random things`"
                      backtick2: "nested templating with backticks {{` random.Values `}}"
                    labels:
                      severity: critical
                    annotations:
                      description: Pod {{$labels.pod}} in {{$labels.namespace}} got OOMKilled
```

src/internal/packager/helm/chart.go
jeff-mccoy: Oh I see you changed the prefix/suffix to not use backtick, I can try that.

jeff-mccoy: Confirmed that escaping is not sufficient to even deploy the chart, much less pass the test.

Screenshot 2023-02-12 at 6 33 44 PM

jeff-mccoy:
> > @corang please provide a real-world use of a combination of backtick and nested go templating.
> 
> ```
> apiVersion: v1
> kind: ConfigMap
> metadata:
>   name: dont-template-me
> data:
>   values.yaml: |
>     monitoring:
>       values:
>         additionalPrometheusRulesMap:
>           custom-rules:
>             groups:
>               - name: OOMKilled
>                 rules:
>                   - alert: OOMKilled {{ "{{ \"random.Values\" }}" }} {{ `I wanted to use a backtick` }}
>                     expr: sum_over_time(kube_pod_container_status_terminated_reason{reason="OOMKilled"}[5m])>0
>                     for: 5m
>                     annotations: 
>                       backtick1: "content with backticks `some random things`"
>                       backtick2: "nested templating with backticks {{` random.Values `}}"
>                     labels:
>                       severity: critical
>                     annotations:
>                       description: Pod {{$labels.pod}} in {{$labels.namespace}} got OOMKilled
> ```

lol that's NOT a real-world usage, that just shoving every possible weird combination (not seen in the wild) into one configmap to make testing faster.

Racer159:
> 

I would argue it is because there is not really a reason preventing someone from trying to use backticks in a template.  They don't _need_ to use backticks in that scenario, but it is something they certainly could try, that they would encounter an error on, file an issue we would need to respond and deal with as a possible bug.  We _could_ document this as a known limitation but that seems wrong when it is super easy to just make it work how people might expect.

Racer159 (approved):
acceptable work @jeff-mccoy

jeff-mccoy:
> > 
> 
> 
> 
> I would argue it is because there is not really a reason preventing someone from trying to use backticks in a template.  They don't _need_ to use backticks in that scenario, but it is something they certainly could try, that they would encounter an error on, file an issue we would need to respond and deal with as a possible bug.  We _could_ document this as a known limitation but that seems wrong when it is super easy to just make it work how people might expect.

Lol we already changed the behavior bad cop. 😂

jeff-mccoy:
> > 
> 
> 
> 
> I would argue it is because there is not really a reason preventing someone from trying to use backticks in a template.  They don't _need_ to use backticks in that scenario, but it is something they certainly could try, that they would encounter an error on, file an issue we would need to respond and deal with as a possible bug.  We _could_ document this as a known limitation but that seems wrong when it is super easy to just make it work how people might expect.

Lol we already changed the behavior bad cop. 😂

PR #1359, opened 2023-02-14
1324 images tar now expanded

YrrepNoj:
**Work Remaining:**
- [x] Find a good way to test backwards compatibility
    - backwards compatibility for this features is implicitly tested within `27_cosign_deploy_test.go` as that is deploying an older zarf package that still has the images in an `images.tar` tarball when deploying the `zarf-hello-world` games package.
- [x] Look into improving the progress bar spinner when pulling the images when building a package
- [x] Clean up some places where I got lazy with error handling (dogsledding errors from functions)

### DIFF
--- a/src/internal/packager/images/pull.go
+++ b/src/internal/packager/images/pull.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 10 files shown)


src/internal/packager/images/pull.go
```diff
@@ -168,3 +169,63 @@ func (i *ImgConfig) PullImage(src string, spinner *message.Spinner) (img v1.Imag
 
 	return img, nil
 }
+
+// IndexJSON represents the index.json file in an OCI layout.
+type IndexJSON struct {
```
Noxsios, line 174: In `features/oci-publish` I import the OCI spec library, once merged into the main feat branch, could replace this strcut w/ `ocispec.IndexManifest` (can't remember the specific name atm).

PR #1363, opened 2023-02-15
Patch the editUrl for the docs

bdfinst:
## Description

There’s a weird error in the docusaurus content-docs plugin. It’s dropping the last segment of the editUrl path
This editUrl: 'https://github.com/defenseunicorns/zarf/tree/main/x/' results in https://github.com/defenseunicorns/zarf/tree/main/ as the base path. For now, I’ll apply the hacky dummy x path. I may look at their code later to see why their plugin is broken.


docs-website/package.json
```diff
@@ -14,9 +14,9 @@
     "write-heading-ids": "docusaurus write-heading-ids"
   },
   "dependencies": {
-    "@docusaurus/core": "2.3.0",
-    "@docusaurus/preset-classic": "2.3.0",
-    "@docusaurus/theme-mermaid": "^2.3.0",
+    "@docusaurus/core": "^2.3.1",
```
jeff-mccoy, line 17: we can bump deps, but renovate will pin them again so can we drop the `^`?

jeff-mccoy (commented):
quick comment, but tested it on a few pages and it does seem to be working

Racer159:
> 

same, and agree with that comment @jeff-mccoy

docs-website/package.json
bdfinst: Yep. Didn't notice it added them back. Fixed.

jeff-mccoy (approved):

PR #1373, opened 2023-02-19
Add `tools kubectl`, `tools wait-for` & Wait Action.

jeff-mccoy:
Vendors in kubectl and adds a new `zarf tools wait-for` command that is wrapper around `kubectl get` + `kubectl wait`. Also adds a new concept of a `wait action`. 

Features:
- Wait for k8s resources to exist
- Wait for k8s resources to exist & have a specified status
- Use name or label selectors for k8s resources
- Wait for http(s) endpoints to have a specific response code
- Wait for generic network endpoints to appear to be up
- Specify a custom timeout window

Screenshot 2023-02-20 at 10 24 17 AM

## Type of change

## Checklist before merging


jeff-mccoy:
Custom kubectl actions (without `wait-for`)
Screenshot 2023-02-19 at 3 42 49 AM

Noxsios:
This change will remove the dependency of the `kubectl` CLI for Zarf operations, the docs will need to updated accordingly.

jeff-mccoy:
Switched back to draft until I can push up the action changes.

examples/flux-test/zarf.yaml
```diff
@@ -29,3 +29,20 @@ components:
       - https://github.com/stefanprodan/podinfo.git
     images:
       - ghcr.io/stefanprodan/podinfo:6.3.3
+    actions:
+      onDeploy:
+        after:
+          # Waiting using the wait-for command
+          - description: "Podinfo deployment to be available via cmd action"
+            cmd: "./zarf tools wait-for deployment podinfo available -n podinfo"
+            mute: true
```
Racer159, line 38: Do we want to leave this in here? I think we should push people to use `wait` in yaml rather than this command (even though they _can_)

src/cmd/tools/common.go
```diff
@@ -0,0 +1,60 @@
...
+	Short: lang.CmdToolsShort,
+}
+
+func Include(rootCmd *cobra.Command) {
+	rootCmd.AddCommand(toolsCmd)
+}
+
+// Check if the command is being run as a vendor-only command
+func CheckVendorOnly() bool {
+	vendorCmd := []string{
```
Racer159, line 31: Linting errors

src/config/lang/english.go
```diff
@@ -262,6 +262,19 @@ const (
 	CmdToolsSbomShort = "Generates a Software Bill of Materials (SBOM) for the given package"
 	CmdToolsSbomErr   = "Unable to create sbom (syft) CLI"
 
+	CmdToolsWaitForShort = "Waits for a given Kubernetes resource to be ready"
```
Racer159, line 265: There are linting errors above this that would be easy to fix while you are here.

src/types/component.go
```diff
@@ -137,14 +137,36 @@ type ZarfComponentActionDefaults struct {
...
+	Env             []string                 `json:"env,omitempty" jsonschema:"description=Additional environment variables to set for the command"`
+	Cmd             string                   `json:"cmd,omitempty" jsonschema:"description=The command to run. Must specify either cmd or wait for the action to do anything."`
+	SetVariable     string                   `json:"setVariable,omitempty" jsonschema:"description=The name of a variable to update with the output of the command. This variable will be available to all remaining actions and components in the package.,pattern=^[A-Z0-9_]+$"`
+	Description     string                   `json:"description,omitempty" jsonschema:"description=Description of the action to be displayed during package execution instead of the command"`
+	Wait            *ZarfComponentActionWait `json:"wait,omitempty" jsonschema:"description=Wait for a condition to be met before continuing. Must specify either cmd or wait for the action to do anything. If both are specified, the command will be run first, then the wait condition will be checked. See the 'zarf tools wait-for' command for more info."`
```
Racer159, line 148: I thought we talked about only allowing one or the other?

Racer159 (changes requested):

src/types/component.go
jeff-mccoy: Yes it says it in the docs, JSON schema and has a validator to run on `package create`, but our current setup doesn't support JSON Schema One-of (I tried it)

src/config/lang/english.go
jeff-mccoy: Indeed... 😆 

Screenshot 2023-02-21 at 11 41 52 AM

examples/flux-test/zarf.yaml
jeff-mccoy: Yeah, that was a holdover to test the path before I decided to use the actual wait-for command, so now it isn't needed.

docs/4-user-guide/5-component-actions.md
```diff
@@ -103,23 +103,44 @@ actions:
...
 ```
 
-## Action Configuration
+## Common Action Configuration Keys
 
-Within each of the `action` lists (`before`, `after`, `onSuccess`, and `onFailure`), the following action configurations are available:
+The following keys are common to all action configurations (wait or command):
 
-- `cmd` - (required) the command to run
+- `description` - a description of the action that will replace the text the users sees when the action is running, e.g. `description: "File to be created"` would show `Waiting for "File to be created"` instead of `Waiting for "touch test-create-before.txt"`
```
Racer159, line 110:
```suggestion
- `description` - a description of the action that will replace the text the user sees when the action is running, e.g. `description: "File to be created"` would show `Waiting for "File to be created"` instead of `Waiting for "touch test-create-before.txt"`
```

docs/4-user-guide/5-component-actions.md
```diff
@@ -103,23 +103,44 @@ actions:
...
 - `dir` - the directory to run the command in, defaults to the current working directory
 - `mute` - whether to mute the realtime output of the command, output is always shown at the end on failure (default: `false`)
-- `maxTotalSeconds` - the maximum total time to allow the command to run (default: `0` - no limit)
 - `maxRetries` - the maximum number of times to retry the command if it fails (default: `0` - no retries)
 - `env` - an array of environment variables to set for the command in the form of `name=value`
 - `setVariable` - set the standard output of the command to a variable that can be used in other actions or components
 
+## Wait Action Configuration
+
+A wait action holds the component stage it is triggered in until the condition is met, or triggers a failure if the timeout is exceeded (defaults to 5 minutes). Use the `wait` key to define the wait paramaters, _you cannot use `cmd` and `wait` in the same action_. A wait action is really just _yaml sugar_ for a call to `./zarf tools wait-for`, but in a more exciting yaml-sort-of-way. Within each of the `action` lists (`before`, `after`, `onSuccess`, and `onFailure`), the following action configurations are available:
```
Racer159, line 126:
```suggestion
A wait action pauses the component stage it is triggered in until the condition is met, or triggers a failure if maxTotalSeconds is exceeded (defaults to 5 minutes). Use the `wait` key to define the wait paramaters, _you cannot use `cmd` and `wait` in the same action_. A wait action is really just _yaml sugar_ for a call to `./zarf tools wait-for`, but in a more exciting yaml-sort-of-way. Within each of the `action` lists (`before`, `after`, `onSuccess`, and `onFailure`), the following action configurations are available:
```

src/types/component.go
Racer159: This text says: "If both are specified, the command will be run first" which cannot happen since they cannot both be specified then.

Racer159 (changes requested):

Racer159 (approved):
lgtm

PR #1427, opened 2023-03-08
Fix OCI pulling of helm chart versions

### LINKED ISSUE #1422: OCI Helm Charts Always Pull Latest
### Environment
Device and OS: System76 Darter Pro, PopOS 22.04
App version: v0.25.0-rc1
Other:

### Steps to reproduce
1. Create the `examples/helm-oci-chart` example
2. Deploy the example

### Expected result
Podinfo comes up without complaining about image pull backoff

### Actual Result
Podinfo wants the wrong version because Zarf pulls 6.3.4 instead of 6.3.3

image

### Severity/Priority
Medium

Racer159:
## Description

This resolves an issue where Zarf erroneously pulls the latest helm chart when pulling via OCI.

## Related Issue

Fixes #1422 

## Type of change

## Checklist before merging

### DIFF
--- a/src/internal/packager/helm/repo.go
+++ b/src/internal/packager/helm/repo.go
@@ -104,6 +104,8 @@ func (h *Helm) DownloadPublishedChart(destination string) {
 			spinner.Fatalf(err, "Unable to create a new registry client")
 		}
 		chartURL = h.Chart.URL
+		// Explicitly set the pull version for OCI
+		pull.Version = h.Chart.Version
 	} else {
 		// Perform simple chart download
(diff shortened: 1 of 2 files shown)


jeff-mccoy:
Would it be possible to have a test check for the exact version that was rolled out?

Racer159:
> Would it be possible to have a test check for the exact version that was rolled out?

We could though it will fail because the images will mismatch - that was what was happening before

jeff-mccoy (approved):
Lgtm, just weird they don't use the chart version 😂

PR #1447, opened 2023-03-17
Fix OCI test flakiness

### LINKED ISSUE #1435: Investigate + fix flaky OCI publish tests
Current E2E testing around OCI packages uses the in-cluster created registry for pushing / pulling packages.  I believe this causes test flakiness and should be changed to do a `docker run registry:2.8.1` during setup suite.

Example CI failure: https://github.com/defenseunicorns/zarf/actions/runs/4421331130/jobs/7752000718

This _could_ be for reasons mentioned above, it _could_ be our implementation of `http.Transport` in `oras.go`, or it _could_ be something else, more investigation is required.

### Severity/Priority

This should block the full release of the OCI publish feature.

Noxsios:
Fixes #1435 

## Type of change

## Checklist before merging


src/pkg/utils/oras.go
```diff
@@ -90,37 +91,84 @@ func (o *OrasRemote) withAuthClient(ref registry.Reference) (*auth.Client, error
...
+
+		timer := time.NewTimer(duration)
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			return nil, ctx.Err()
+		case <-timer.C:
+		}
+		attempt++
+	}
```
Racer159, line 155: Can this be moved to a better file (since it isn't oras specific and could be generally useful?

Also can you link to where this was from

Racer159 (changes requested):

Racer159 (approved):
LGTM

PR #1464, opened 2023-03-21
Update schema docs to include information about OCI registries for helm charts

### LINKED ISSUE #1463: Docs for the Zarf package schema do not specify that a chart URL can be an OCI registry
The docs for the Zarf package schema do not specify that a chart URL can be an OCI registry

https://docs.zarf.dev/docs/user-guide/zarf-schema#components--charts

### Additional context
Screenshot 2023-03-21 at 3 37 22 PM

lucasrod16:
## Description

- Updates schema docs to include information about OCI registries for helm charts
- Updates contributing guide to include information about regenerating schema and docs when an update to the docs is made

## Related Issue

Fixes #1463

## Type of change

## Checklist before merging


Racer159:
@lucasrod16 looks like you need to run `make docs-and-schema` and commit the generated changes

lucasrod16:
@Racer159 when I'm running `make docs-and-schema` it's stripping away the description after the first comma to be `The URL of the OCI registry`. I'll go ahead and put this PR in draft status until I resolve it.  Any suggestions?

Racer159:
@lucasrod16 yeah the reason it cuts it off is because the values there are comma separated - to do a description with commas you have to use `jsonschema_description` as in: https://github.com/defenseunicorns/zarf/blob/cbf3c04dbaa568f0010c4dfb2db1f317ead3f9c3/src/types/package.go#L30

And if you can add a reference to that in the contributing guide that would be great!  The purpose of make docs-and-schema is to update our auto generated docs - we for people to do this (rather than CI) so that there is a git commit signature from a person on the commit for better traceability (rather than an non-person entity Github CI token)

lucasrod16:
> @lucasrod16 yeah the reason it cuts it off is because the values there are comma separated - to do a description with commas you have to use `jsonschema_description` as in:
> 
> https://github.com/defenseunicorns/zarf/blob/cbf3c04dbaa568f0010c4dfb2db1f317ead3f9c3/src/types/package.go#L30
> 
> And if you can add a reference to that in the contributing guide that would be great! The purpose of `make docs-and-schema` is to update our auto generated docs - we force people to do this (rather than CI) so that there is a git commit signature from a person on the commit for better traceability (rather than an non-person entity Github CI token)

@Racer159 Thanks! I made the `jsonschema_description` update and added examples as well. I will also add a section about running `make docs-and-schema` in the contributing guide

CONTRIBUTING.md
```diff
@@ -46,6 +46,12 @@ You can learn more about the testing of Zarf [here](docs/6-developer-guide/2-tes
 
 ## Documentation
 
+### Updating Our Documentation
+
+Our documentation is auto-generated from the [Zarf package schema definition](https://github.com/defenseunicorns/zarf/blob/main/zarf.schema.json), so when an update is made you will need to run `make docs-and-schema` locally to regenerate the schema and documentation. CI checks if this was ran, and will fail if it wasn't.
```
Racer159, line 51: Technically there is more going on there.  Not a huge deal but this would be more accurate to say:

```suggestion
Our documentation is auto-generated from the `src/types` and `src/cmd` go packages.  This includes the [Zarf package jsonschema](https://github.com/defenseunicorns/zarf/blob/main/zarf.schema.json), the [Zarf schema docs](https://docs.zarf.dev/docs/user-guide/zarf-schema), the [Zarf CLI docs](https://docs.zarf.dev/docs/user-guide/the-zarf-cli/), and our [front-end API types](https://github.com/defenseunicorns/zarf/blob/main/src/ui/lib/api-types.ts).   When an update to types or the CLI commands is made you will need to run `make docs-and-schema` locally to regenerate the schema and documentation. CI checks if this was ran, and will fail if it wasn't.
```

Racer159 (changes requested):
tiny suggestion but figured more info could be valuable (in case someone ran into this on our command docs too)

CONTRIBUTING.md
lucasrod16: Changes made

Racer159 (approved):
lgtm! thanks!

Issue #23, opened 2021-09-07
Implement cryptographic validation of all payload and metadata in a zarf package


RothAndrew:
Implement hashing/signing for all elements of the package to ensure authenticity and integrity of the artifacts.

---

Migrated from https://repo1.dso.mil/platform-one/big-bang/apps/product-tools/zarf/-/issues/36

JasonvanBrackel:
@Racer159 Are you actively working on this?  If so could you please move it to doing now on the project board?  I may get rid of thing if we don't fine value in it long term, but I like the visual nature of it.

JasonvanBrackel:
https://github.com/defenseunicorns/zarf/issues/23 May be related / be a subtask of this.

Racer159:
Dropped myself as an assignee, this is still a backlog task, I just added myself to not forget about it at the time

PR #1467, opened 2023-03-22
Feature: Validate package contents and verify package signatures

YrrepNoj:
Fixes: #23


go.mod
```diff
@@ -57,6 +57,7 @@ require (
 	bitbucket.org/creachadair/shell v0.0.7 // indirect
 	cloud.google.com/go/compute v1.18.0 // indirect
 	cloud.google.com/go/compute/metadata v0.2.3 // indirect
+	cuelang.org/go v0.4.3 // indirect
```
Racer159, line 60: What did we change that added all of these libraries?

src/cmd/package.go
```diff
@@ -320,12 +322,14 @@ func bindDeployFlags() {
...
 
 	deployFlags.StringToStringVar(&pkgConfig.DeployOpts.SetVariables, "set", v.GetStringMapString(V_PKG_DEPLOY_SET), lang.CmdPackageDeployFlagSet)
 	deployFlags.StringVar(&pkgConfig.DeployOpts.Components, "components", v.GetString(V_PKG_DEPLOY_COMPONENTS), lang.CmdPackageDeployFlagComponents)
 	deployFlags.StringVar(&pkgConfig.DeployOpts.Shasum, "shasum", v.GetString(V_PKG_DEPLOY_SHASUM), lang.CmdPackageDeployFlagShasum)
 	deployFlags.StringVar(&pkgConfig.DeployOpts.SGetKeyPath, "sget", v.GetString(V_PKG_DEPLOY_SGET), lang.CmdPackageDeployFlagSget)
 	deployFlags.IntVar(&pkgConfig.PublishOpts.CopyOptions.Concurrency, "oci-concurrency", v.GetInt(V_PKG_PUBLISH_OCI_CONCURRENCY), lang.CmdPackagePublishFlagConcurrency)
+	deployFlags.StringVarP(&pkgConfig.DeployOpts.PublicKeyPath, "key", "k", v.GetString(V_PKG_DEPLOY_PUBLIC_KEY), lang.CmdPackageDeployFlagPublicKey)
 }
 
 func bindInspectFlags() {
```
Racer159, line 335: Can we do key validation on inspect as well?

src/config/lang/english.go
```diff
@@ -195,12 +195,14 @@ const (
...
 	CmdPackageCreateFlagSkipSbom        = "Skip generating SBOM for this package"
 	CmdPackageCreateFlagMaxPackageSize  = "Specify the maximum size of the package in megabytes, packages larger than this will be split into multiple parts. Use 0 to disable splitting."
+	CmdPackageCreateFlagSigningKey      = "Path to private key file for signing packages"
 
 	CmdPackageDeployFlagConfirm    = "Confirm package deployment without prompting"
 	CmdPackageDeployFlagSet        = "Specify deployment variables to set on the command line (KEY=value)"
 	CmdPackageDeployFlagComponents = "Comma-separated list of components to install.  Adding this flag will skip the init prompts for which components to install"
 	CmdPackageDeployFlagShasum     = "Shasum of the package to deploy. Required if deploying a remote package and \"--insecure\" is not provided"
 	CmdPackageDeployFlagSget       = "Path to public sget key file for remote packages signed via cosign"
+	CmdPackageDeployFlagPublicKey  = "Path to public key file for signed packages"
```
Racer159, line 205:
```suggestion
	CmdPackageDeployFlagPublicKey  = "Path to public key file for validating signed packages"
```

src/config/lang/english.go
```diff
@@ -209,6 +211,9 @@ const (
 	CmdPackageRemoveFlagComponents = "Comma-separated list of components to uninstall"
 
 	CmdPackagePublishFlagConcurrency = "Number of concurrent layer operations to perform when interacting with a remote package."
+	CmdPackagePublishFlagSigningKey  = "Path to private key file for signing packages"
+
+	CmdPackagePullPublicKey = "Path to public key file for signed packages"
```
Racer159, line 216:
```suggestion
	CmdPackagePullPublicKey = "Path to public key file for validating signed packages"
```

src/pkg/packager/common.go
```diff
@@ -352,3 +362,87 @@ func (p *Packager) handleIfPartialPkg() error {
 	p.cfg.DeployOpts.PackagePath = destination
 	return nil
 }
+
+// TODO: Still need to validate that there is no extra files in the package
```
Racer159, line 366:
```suggestion
// TODO: Still need to validate that there are no extra files in the package
```

src/pkg/packager/common.go
```diff
@@ -219,6 +222,13 @@ func (p *Packager) loadZarfPkg() error {
 		return fmt.Errorf("unable to read the zarf.yaml in %s: %w", p.tmp.Base, err)
 	}
 
+	// Validate the checksums of all the things!!!
+	if p.cfg.Pkg.Metadata.ChecksumSignature != "" {
```
Racer159, line 226: Can we call this something else since technically it isn't a signature (just a checksum of checksums):  `Metachecksum`, `AggregateChecksum` or something similar might make sense

src/pkg/packager/common.go
```diff
@@ -352,3 +362,87 @@ func (p *Packager) handleIfPartialPkg() error {
...
+	// Run pre-checks to make sure we have what we need to validate the checksums
+	_, err := os.Stat(p.tmp.Checksums)
+	if err != nil {
+		return fmt.Errorf("unable to validate checksums as we are unable to find checksums.txt file within the package")
+	}
+	if p.cfg.Pkg.Metadata.ChecksumSignature == "" {
+		return fmt.Errorf("unable to validate checksums because of missing metadata checksum signature")
+	}
+
+	// Verify the that checksums.txt file matches the signature provided
```
Racer159, line 379: Same thing here if we change "ChecksumSignature"

src/pkg/packager/common.go
```diff
@@ -352,3 +362,87 @@ func (p *Packager) handleIfPartialPkg() error {
...
+			return nil
+		}
+
+		// We were expecting a signature, but there wasn't one..
+		return fmt.Errorf("package is not signed but a key was provided")
+	}
+
+	// Validate the signature of the package
+	if publicKeyPath == "" {
+		return fmt.Errorf("package is signed but no key was provided")
```
Racer159, line 439: We may want to include the flag they need to pass here to help them with it

src/pkg/packager/publish.go
```diff
@@ -62,68 +61,34 @@ func parseZarfLayerMediaType(filename string) string {
...
-		filepath.Join(p.tmp.Images, "index.json"),
-		filepath.Join(p.tmp.Images, "oci-layout"),
-	}
-	// if checksums.txt file exists, include it
-	if !utils.InvalidPath(filepath.Join(p.tmp.Base, "checksums.txt")) {
-		paths = append(paths, filepath.Join(p.tmp.Base, "checksums.txt"))
-	}
-	// if p.tmp.SbomTar exists, include it
-	if !utils.InvalidPath(p.tmp.SbomTar) {
-		paths = append(paths, p.tmp.SbomTar)
```
Racer159, line 80: Without these checks we might have issues if someone tries to publish an older Zarf package.

Racer159 (changes requested):

src/cmd/package.go
YrrepNoj: Yes, but what are your thoughts on throwing this logic behind a flag for inspect?

When inspecting a published package, I would rather not have the default behavior be downloading the entire package to validate signatures/checksums.

src/pkg/packager/publish.go
YrrepNoj: I don't think so, we had to do these checks before because we had to manually add all the layers that we wanted to publish because we were being inefficient with how we were unarchiving/archiving the layers within a package.

Because of the changes above, we are only unarchiving the package itself, meaning we can add all the things (that we found via the `filepath.Glob()` call) that was already inside of that tarball as layers that we are publishing.


@Noxsios Does that sound right to you too?

go.mod
YrrepNoj: I was playing around with something in an early commit and had a merge conflict when I merged main back into this branch. After merging main back i ran a `go mod tidy` which is where this came from.

src/pkg/packager/publish.go
Noxsios: Going to run this locally to verify all the correct layers are still included

src/cmd/package.go
Racer159: I think that is fine to be behind a flag

src/cmd/package.go
```diff
@@ -320,18 +323,22 @@ func bindDeployFlags() {
...
 	inspectFlags := packageInspectCmd.Flags()
 	inspectFlags.BoolVarP(&includeInspectSBOM, "sbom", "s", false, lang.CmdPackageInspectFlagSbom)
 	inspectFlags.StringVar(&outputInspectSBOM, "sbom-out", "", lang.CmdPackageInspectFlagSbomOut)
+	inspectFlags.StringVarP(&pkgConfig.DeployOpts.PublicKeyPath, "key", "k", v.GetString(V_PKG_DEPLOY_PUBLIC_KEY), lang.CmdPackageDeployFlagPublicKey)
```
Racer159, line 340: Do we want to reuse deployopts here? We don't on pull and publish/create are different.

src/pkg/packager/common.go
```diff
@@ -352,3 +362,117 @@ func (p *Packager) handleIfPartialPkg() error {
...
+			return fmt.Errorf("unable to get the file info for a file within the package: %s - %w", path, err)
+		}
+
+		// If the glob item is a file, add it to the map
+		if !fileInfo.IsDir() {
+			filepathMap[path] = false
+		}
+	}
+	// Verify the that checksums.txt file matches the aggregated checksum provided
+	actualChecksumSig, err := utils.GetSHA256OfFile(p.tmp.Checksums)
```
Racer159, line 397:
```suggestion
	actualAggregateChecksum, err := utils.GetSHA256OfFile(p.tmp.Checksums)
```

src/pkg/packager/common.go
```diff
@@ -352,3 +362,117 @@ func (p *Packager) handleIfPartialPkg() error {
...
+	if err != nil {
+		return fmt.Errorf("unable to get the checksum of the checksums.txt file: %w", err)
+	}
+	if actualChecksumSig != p.cfg.Pkg.Metadata.AggregateChecksum {
```
Racer159, line 401:
```suggestion
	if actualAggregateChecksum != p.cfg.Pkg.Metadata.AggregateChecksum {
```

src/pkg/packager/common.go
```diff
@@ -352,3 +362,117 @@ func (p *Packager) handleIfPartialPkg() error {
...
+		if expectedShasum != actualShasum {
+			return fmt.Errorf("mismatch on the checksum of the %s file (expected: %s, actual: %s)", itemPath, expectedShasum, actualShasum)
+		}
+
+		filepathMap[filepath.Join(p.tmp.Base, itemPath)] = true
+	}
+
+	for path, processed := range filepathMap {
+		if !processed {
+			return fmt.Errorf("the file %s was not processed by the checksums.txt file", path)
```
Racer159, line 440: This error should be more explicit that an extra file was found inside of a Zarf.yaml

src/pkg/packager/common.go
```diff
@@ -352,3 +362,117 @@ func (p *Packager) handleIfPartialPkg() error {
...
+		}
+	}
+
+	message.Successf("All of the checksums matched!")
+	return nil
+}
+
+func (p *Packager) validatePackageSignature(publicKeyPath string) error {
+
+	// If the insecure flag we provided, ignore the signature validation
```
Racer159, line 450:
```suggestion
	// If the insecure flag was provided, ignore the signature validation
```

The docs for the insecure flag should note that it skips package validation as well.

src/pkg/packager/inspect.go
```diff
@@ -12,69 +12,68 @@ import (
...
+
 	}
 
 	pterm.Println()
 	pterm.Println()
 
 	utils.ColorPrintYAML(p.cfg.Pkg)
 
+	// Validate the package checksums and signatures if specified
+	if validatePackage || p.cfg.DeployOpts.PublicKeyPath != "" {
```
Racer159, line 59: This could maybe be one flag - if they provide a key attempt to validate, otherwise don't (having two is a little weird IMO)

src/pkg/packager/publish.go
```diff
@@ -62,68 +61,34 @@ func parseZarfLayerMediaType(filename string) string {
...
+	if err := p.readYaml(p.tmp.ZarfYaml, true); err != nil {
+		return fmt.Errorf("unable to read the zarf.yaml in %s: %w", p.tmp.Base, err)
 	}
-	componentTarballs := []string{}
 
-	// repackage the component directories into tarballs
-	for _, componentDir := range componentDirs {
-		dst := filepath.Join(p.tmp.Components, filepath.Base(componentDir)+".tar")
-		err = archiver.Archive([]string{componentDir}, dst)
+	if p.cfg.PublishOpts.SigningKeyPath != "" {
```
Racer159, line 74: Before signing it should make sure that the checksums.txt matches.

src/pkg/packager/common.go
```diff
@@ -352,3 +362,117 @@ func (p *Packager) handleIfPartialPkg() error {
...
+	_, err := os.Stat(p.tmp.Checksums)
+	if err != nil {
+		return fmt.Errorf("unable to validate checksums as we are unable to find checksums.txt file within the package")
+	}
+	if p.cfg.Pkg.Metadata.AggregateChecksum == "" {
+		return fmt.Errorf("unable to validate checksums because of missing metadata checksum signature")
+	}
+
+	// Get a glob of all the files within the package
+	packageGlob, err := filepath.Glob(filepath.Join(p.tmp.Base, "*"))
```
Racer159, line 378: This should be a filepath walk just like how the generate checksums does it - otherwise you could hide a file inside of a zarf package at a lower folder level.

Racer159 (changes requested):

src/pkg/packager/common.go
Noxsios: ^ because filepath.Glob does not have a `**` feature, it will only glob everything that is at the base level of the tmp dir.

src/pkg/packager/inspect.go
YrrepNoj: Having two is a bit weird, I was thinking the same thing but then I wasn't what to do about validating the checksums.

If we only have one flag, the checksums would either always get validate, or only get validated if the public key was provided.

I shied away from always validating the checksums because that ends up pulling a large amount of stuff when working with a remote package.

Racer159: Yeah we could make the public key path flag more explicit that you are both providing a public key and asking zarf to perform validation.

YrrepNoj: Cleaned it up to be a bit more intuitive while also being more direct about warning the user for checksum/signature validation given the two scenarios of a remote OCI package and a local on-disk package.

src/pkg/packager/publish.go
YrrepNoj: Doing this with a os.walk now to avoid this an issue.

go.mod
YrrepNoj: It appears these dependency updates are necessary for some of the functionality we are now doing with cosign. I tried to copy over the go.mod and go.sum from the current main branch and couldn't get a build to resolve.

src/cmd/package.go
```diff
@@ -300,13 +301,15 @@ func bindCreateFlags() {
 	v.SetDefault(V_PKG_CREATE_SBOM_OUTPUT, "")
 	v.SetDefault(V_PKG_CREATE_SKIP_SBOM, false)
 	v.SetDefault(V_PKG_CREATE_MAX_PACKAGE_SIZE, 0)
+	v.SetDefault(V_PKG_CREATE_SIGNING_KEY, "")
```
Racer159, line 304: Thoughts on adding the cosign command line as well so that people can more easily make keypairs?

src/pkg/utils/sget.go
```diff
@@ -150,3 +152,48 @@ func Sget(ctx context.Context, image, key string, out io.Writer) error {
...
+	certGithubWorkflowName := ""
+	certGithubWorkflowRepository := ""
+	certGithubWorkflowRef := ""
+	enforceSCT := false
+
+	return verify.VerifyBlobCmd(ctx, keyOptions, certRef, certEmail, certIdentity, certOidcIssuer, certChain, sigPath, blobPath, certGithubWorkflowTrigger, certGithubWorkflowSha, certGithubWorkflowName, certGithubWorkflowRepository, certGithubWorkflowRef, enforceSCT)
+}
+
+// CosignSignBlob signs the provide binary and returns the signature
+func CosignSignBlob(blobPath string, outputSigPath string, keyPath string) ([]byte, error) {
```
Racer159, line 176: This does not support keys with passwords :eyes:
image

Racer159 (changes requested):

src/pkg/packager/common.go
```diff
@@ -353,3 +363,110 @@ func (p *Packager) handleIfPartialPkg() error {
...
+		return fmt.Errorf("package is not signed but a key was provided")
+	}
+
+	// Validate the signature of the package
+	if publicKeyPath == "" {
+		return fmt.Errorf("package is signed but no key was provided, using signed packages requires a --key or --insecure flag to continue")
+	}
+
+	// Validate the signature with the key we were provided
+	if err := utils.CosignVerifyBlob(p.tmp.ZarfYaml, p.tmp.ZarfSig, publicKeyPath); err != nil {
```
Racer159, line 467: Is there a way to control the Verified OK text color?

YrrepNoj: As we discussed in person, we're going to leave the `Verified OK` from cosign but then make a green descriptive log afterwards to make it clear what has actually been verified successfully

src/cmd/package.go
YrrepNoj: Added a `gen-key` tools command to generate the keypairs

src/types/runtime.go
```diff
@@ -19,19 +19,31 @@ type ZarfCommonOptions struct {
...
-	PackagePath string             `json:"packagePath" jsonschema:"description=Location where a Zarf package to publish can be found"`
+	Reference          registry.Reference `jsonschema:"description=Remote registry reference"`
+	CopyOptions        oras.CopyOptions   `jsonschema:"description=Options for the copy operation"`
+	PackOptions        oras.PackOptions   `jsonschema:"description=Options for the pack operation"`
+	PackagePath        string             `json:"packagePath" jsonschema:"description=Location where a Zarf package to publish can be found"`
+	SigningKeyPassword string             `json:"signingKeyPassword" jsonschema:"description=Password to the private key signature file that will be used to sign the published package"`
+	SigningKeyPath     string             `json:"signingKeyPath" jsonschema:"description=Location where the private key component of a cosign key-pair can be found"`
+}
+
+// ZarfPublishOptions tracks the user-defined preferences during a package publish.
```
Racer159, line 40: comment on exported type ZarfPullOptions should be of the form "ZarfPullOptions ..." (with optional leading article)

Racer159: Will approve anyways but be careful about these linting issues - we will need to run a PR to clean them up.

Racer159 (approved):
Approving for now, but we will want to run a PR to cleanup linting issues - my PR had some, and this has some in it as well.

PR #1475, opened 2023-03-24
Adding Gitea Virtual Service to Big Bang example

dgershman:
## Description

I thought I would expand the example by showing a relevant and useful addition.  Since Gitlab is typically deployed with Big Bang, this would give somewhat of an equivalent feel to be able to not have to worry about setting up a proxy forward for the service for Gitea.

## Related Issue

N/A

## Type of change

## Checklist before merging

### DIFF
--- a/examples/big-bang/virtualservices/gitea.yaml
+++ b/examples/big-bang/virtualservices/gitea.yaml
@@ -1,0 +1,16 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: gitea
+  namespace: zarf
+spec:
+  gateways:
+    - istio-system/public
+  hosts:
+    - gitea.bigbang.dev
+  http:
+    - route:
+        - destination:
+            host: zarf-gitea-http.zarf.svc.cluster.local
+            port:
+              number: 3000
--- a/examples/big-bang/zarf.yaml
+++ b/examples/big-bang/zarf.yaml
@@ -27,3 +27,10 @@ components:
           - config/loki.yaml
           # Values are merged in order, so this would override the above and disable everything if uncommented
           # - config/disable-all.yaml
+  - name: extra-virtual-services
+    required: true
+    manifests:
+      - name: gitea
+        namespace: zarf
+        files:
+          - virtualservices/gitea.yaml


examples/big-bang/zarf.yaml
```diff
@@ -27,3 +27,10 @@ components:
           - config/loki.yaml
           # Values are merged in order, so this would override the above and disable everything if uncommented
           # - config/disable-all.yaml
+  - name: extra-virtual-services
+    required: true
```
Racer159, line 31: Can we name this more specifically and make it optional for those who want to keep Gitea internal or are using an external git-repo?

```suggestion
  - name: gitea-virtual-service
    description: >
      Expose the internal Zarf Gitea server through the Big Bang Istio deployment via a virtual service.
      (only applies if you are using the Zarf-provided Gitea deployment - not an externally configured git host)
```

examples/big-bang/virtualservices/gitea.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: gitea
+  namespace: zarf
+spec:
+  gateways:
+    - istio-system/public
+  hosts:
+    - gitea.bigbang.dev
```
Racer159, line 10: Can we use the DOMAIN variable here?
```suggestion
    - gitea.###ZARF_VAR_DOMAIN###
```

Racer159 (changes requested):
Just a few changes to make it more configurable

Racer159:
Thanks for the PR! Requested a few changes and this will also need your base branch to be updated prior to merge.

examples/big-bang/virtualservices/gitea.yaml
dgershman: done

Racer159 (approved):
lgtm, but you will need to update the base branch again

dgershman:
> lgtm, but you will need to update the base branch again

thanks for the heads up, done.

YrrepNoj (approved):

Racer159 (approved):
lgtm

PR #1477, opened 2023-03-24
Merge valuesFrom with BigBang compose

runyontr:
## Description

When using a zarf.yaml to extend an existing bigbang definition like:

```yaml
components:
  - name: cocowow
    required: true
    import:
      path: ../defense-unicorns-distro
      name: bigbang
    extensions:
      bigbang:
        version: "###ZARF_PKG_VAR_BIGBANG_VERSION###"
        valuesFiles:
        # look at imports and then merging here for composition
        - ../values/authservice.yaml
        - ../values/keycloak.yaml
   ```
Zarf ignores the provided values files in this zarf.yaml

## Related Issue

Relates to #1060 

## Type of change

## Checklist before merging

### CODE REVIEW

**Review by YrrepNoj — APPROVED** [review]

File: src/pkg/packager/compose.go [hunk 1]
```diff
@@ -203,6 +203,7 @@ func (p *Packager) mergeComponentOverrides(target *types.ZarfComponent, override
 	target.Images = append(target.Images, override.Images...)
 	target.Manifests = append(target.Manifests, override.Manifests...)
 	target.Repos = append(target.Repos, override.Repos...)
+	target.Extensions.BigBang.ValuesFiles = append(target.Extensions.BigBang.ValuesFiles, override.Extensions.BigBang.ValuesFiles...)
```
Racer159, line 206: ValuesFiles may not be instantiated (hence the nil pointer error) - also it may be better to implement a compose interface for extensions to implement (something like):

```
type Extension interface {
	Compose() interface{}
}
```

That way there is a pattern for future extensions to implement that isn't as coupled to Zarf code.

Racer159 (changes requested):

Racer159 (approved):
good 'nuf

PR #1489, opened 2023-03-27
update calculation for the expected bytes during image layer pulls

### LINKED ISSUE #1488: Incorrect expected size for progress bar on image downloads
When Zarf is pulling images to create a package it generates a progress bar to indicate.. well progress.

Right now the progress bar is displaying an incorrect value for the total bytes that are going to be downloaded because it is not accounting for the fact that some images may share layers.

When creating the progress bar, we should take into account duplicate layers so that we can provide a more accurate representation of the progress of the image pulls.

YrrepNoj:
## Description

The progress bar 'expected bytes' now takes duplicate layers into account

Fixes #1488

### DIFF
--- a/src/internal/packager/images/pull.go
+++ b/src/internal/packager/images/pull.go
@@ -70,6 +70,7 @@ func (i *ImgConfig) PullAll() error {
 	}
 
 	totalBytes := int64(0)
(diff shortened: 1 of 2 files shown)


Racer159 (approved):
lgtm

Issue #1491, opened 2023-03-27
Zarf silently fails when init package architecture doesn't match cluster architecture


lucasrod16:
### Environment
Device and OS: M1 Mac
App version: v0.25.0
Kubernetes distro being used: k3s
Other: Assumption is that this is agnostic to device/OS and k8s distro. This would likely occur on any mismatch between zarf init package architecture and target cluster architecture.

### Steps to reproduce
1. From an arm64 machine, run `zarf init`
2. If you don't already have an init package in your local `~/.zarf-cache` directory, zarf will prompt you to download an init package with the same architecture of the machine you're on (which may or may not be the same architecture of the target cluster). If it finds one already in the local cache, it's likely that most of the time this package is the same architecture as the machine you're on.
3. Initialize your cluster with `zarf init` that uses a different architecture for the init package than the target cluster. In this case use an `arm64` init package to initialize an `amd64` target cluster.
4. zarf hangs on opening a port-forwarding tunnel for the zarf injector service

### Expected result
Zarf performs pre-flight checks and lets me know that I can't initialize a cluster with an init package that has a different architecture

### Actual Result
Zarf hangs on the init stage, and doesn't produce any information that lets you know why

### Visual Proof (screenshots, videos, text, etc)
![Screenshot 2023-03-27 at 1 29 57 PM](https://user-images.githubusercontent.com/87675701/228034344-32fad843-cef7-4c4a-b85e-b15406eeea40.png)

![Screenshot 2023-03-27 at 1 33 51 PM](https://user-images.githubusercontent.com/87675701/228034686-b6064be2-9619-4960-b110-a44bdeb109e6.png)

### Severity/Priority
This isn't happening in a production environment, but arm64 machines for local dev are becoming more commonly used (thanks Apple), so I would consider this a medium priority.

lucasrod16:
Could possibly use the [`kubernetes.io/arch`](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-arch) label on nodes to query for the cluster architecture. It is technically possible to have nodes of different architectures in the same cluster, so I am not sure how or if we need to be able to handle that scenario.

Issue #955, opened 2022-10-31
Unable to Deploy Doom Game into Existing EKS Cluster - amd64 mismatch


vennemp:
### Environment
Device and OS: MBP M1Pro 12.6
App version: v0.22.2
Kubernetes distro being used: eks
Other:

### Steps to reproduce
1. Cloned repo.
2. Kubectl works fine with the cluster.
3. zarf init returns
ERROR:  The current Zarf package architecture arm64 does not match the cluster architecture amd64

jeff-mccoy:
The way zarf works by default is it assumes the cluster (EKS in this case) and the CLI binary are the same architecture. Because you're connecting to a different arch ARM -> AMD, you'll need to specify a flag to tell zarf to use a different architecture for the packages:   https://docs.zarf.dev/docs/user-guide/the-zarf-cli/cli-commands/zarf#options.

E.g.:
`zarf init -a=amd64`

vennemp:
thank you!

Racer159:
Added a few labels to mark this as an error message to look at during https://github.com/defenseunicorns/zarf/issues/631

mjnagel:
> The way zarf works by default is it assumes the cluster (EKS in this case) and the CLI binary are the same architecture. Because you're connecting to a different arch ARM -> AMD, you'll need to specify a flag to tell zarf to use a different architecture for the packages: https://docs.zarf.dev/docs/user-guide/the-zarf-cli/cli-commands/zarf#options.

@jeff-mccoy would there be value in adding a prompt to pull an init package with architecture matching the cluster? I'm thinking this could fit in well where the current prompt for downloading an init package is:
```
It seems the init package could not be found locally, but can be downloaded from
https://github.com/defenseunicorns/zarf/releases/download/x/x.tar.zst

Note: This will require an internet connection.
? Do you want to download this init package? (y/N) 
```
Could either before or after this check for a cluster connection and provide a similar prompt if an init package matching that architecture cannot be found locally. I think this would be beneficial since currently the error message for mismatched architecture doesn't pop up until further along in the process.

Noxsios:
I think this can safely be closed as completed. The mismatched arch error messaging is pretty good now due to #1495 and other UX improvements.

PR #1495, opened 2023-03-28
Fail when package architecture doesn't match cluster architecture

lucasrod16:
## Description
Fail `zarf init` and `zarf package deploy` when package architecture doesn't match cluster architecture

Deployment scenarios that this PR covers:

- If `zarf init` is ran in appliance mode(deploying k3s), verify that the package architecture matches the machine that zarf is attempting to initialize with a k3s cluster

- If `zarf init` or `zarf package deploy` is ran, verify that the package architecture matches the target cluster architecture

## Related Issue

Fixes https://github.com/defenseunicorns/zarf/issues/1491

Relates to https://github.com/defenseunicorns/zarf/issues/955

## Type of change

## Checklist before merging


lucasrod16:
This PR is ready for review whenever somebody gets the chance to take a look.

src/cmd/initialize.go
```diff
@@ -137,6 +152,30 @@ func downloadInitPackage(initPackageName, downloadCacheTarget string) error {
...
+	if !strings.Contains(components, "k3s") {
+		c := cluster.NewClusterOrDie()
+		systemArch, err = c.Kube.GetArchitecture()
+
+		if err != nil {
+			message.Fatal(err, err.Error())
+		}
+	} else {
+		systemArch = runtime.GOARCH
+	}
```
Racer159, line 172: Technically for this, the component could be selected interactively without having been specified on the CLI (components would only have k3s in it if it was from the CLI or config file).  A better way to handle this could be comparing the package arch (p.arch) to the cluster arch when the NewClusterWithWait is called

Racer159 (changes requested):

src/cmd/initialize.go
lucasrod16: The problem I was running into with not checking for k3s/appliance mode was that it was failing saying `Failed to connect to cluster` because `verifyArchitecture` was attempting to set up a k8s client, and if there was no existing cluster to connect to yet, then it fails : https://github.com/defenseunicorns/zarf/actions/runs/4556908999/jobs/8038155551#step:6:3228

So I added in this check to handle the cases where the user is using `zarf init` to deploy a k3s cluster. This current logic fails if the init package arch doesn't match the system arch before zarf actually ever prompts the user after running `zarf init`. If you execute `go run main.go init` on this branch with a mismatched arch package you'll see that it fails before prompting the user to select components. Is there a better way we can handle this part?

Updated this to use the `arch` from the Packager struct. Had to change it to `Arch` so that it would be exported and usable outside of the `packager` package. This required updating any references to this field as well, so there are more files changed now.

lucasrod16: I'm also noticing that using the arch field from the Packager struct to get the init package architecture isn't working. It's an empty string value, which is causing every case to fail. I'm not exactly sure why it's not populated

Racer159: At this point in the package lifecycle the package hasn't been loaded from disk yet - that happens in loadZarfPkg on L41 of deploy.go https://github.com/defenseunicorns/zarf/blob/main/src/pkg/packager/deploy.go#L41 goes to https://github.com/defenseunicorns/zarf/blob/9846cf4536cb657467be8b8b0a2ccbfdf0d5c7d0/src/pkg/packager/common.go#L192 which goes to https://github.com/defenseunicorns/zarf/blob/9846cf4536cb657467be8b8b0a2ccbfdf0d5c7d0/src/pkg/packager/yaml.go#L21 which sets p.arch

lucasrod16: Ahh I see. Thanks for pointing that out. Made changes to be able to use p.arch

lucasrod16:
@Racer159 this is ready for review when you get a chance to take a look

src/pkg/packager/common.go
```diff
@@ -440,6 +443,75 @@ func (p *Packager) validatePackageChecksums() error {
...
+func (p *Packager) validatePackageArchitecture() error {
+	var (
+		applianceMode bool
+		clusterArch   string
+		err           error
+		k8sTarget     bool
+	)
+
+	// Iterate over the package components.
+	components := p.getValidComponents()
```
Racer159, line 456: Doing this here will double prompt for components when done without the `--confirm` flag.  For the k3s component it would be better to have an `onDeploy` before action in the k3s Zarf yaml that checks to see if your current architecture matches the package architecture (since we won't truly know if it is selected until after package confirmation.

src/pkg/packager/common.go
```diff
@@ -440,6 +443,75 @@ func (p *Packager) validatePackageChecksums() error {
...
+	*/
+	if p.cluster != nil {
+		clusterArch, err = p.cluster.Kube.GetArchitecture()
+		if err != nil {
+			return err
+		}
+
+		if p.arch != clusterArch {
+			return fmt.Errorf(lang.CmdPackageDeployValidateArchitectureErr, p.arch, clusterArch)
+		}
```
Racer159, line 495: This is dead code - `p.cluster` is always `nil` here as it won't get set until lines 153 or 224 of `deploy.go`.

Now that we have it (it wasn't there when this PR started) I would look at the following in `interactive.go`:

https://github.com/defenseunicorns/zarf/blob/5666ae237cb088628d40af0b597e8fe4f38933af/src/pkg/packager/interactive.go#L40

We don't use p.cluster there since we are only trying to see if we have access and may not at that time but that is a way to get access to a cluster early for pre-checks like this.

src/pkg/packager/common.go
```diff
@@ -440,6 +443,75 @@ func (p *Packager) validatePackageChecksums() error {
...
+
+		clusterArch, err = client.Kube.GetArchitecture()
+		if err != nil {
+			return err
+		}
+
+		if p.arch != clusterArch {
+			return fmt.Errorf(lang.CmdPackageDeployValidateArchitectureErr, p.arch, clusterArch)
+		}
+	}
```
Racer159, line 510: This works, but should be called from that function in `interactive.go` to already have a cluster available.

src/pkg/packager/common.go
```diff
@@ -440,6 +443,75 @@ func (p *Packager) validatePackageChecksums() error {
...
+	// set the clusterArch to the machine we're running on.
+	if applianceMode {
+		clusterArch = runtime.GOARCH
+
+		if p.arch != clusterArch {
+			return fmt.Errorf(lang.CmdPackageDeployValidateArchitectureErr, p.arch, clusterArch)
+		}
+
+		return nil
+	}
```
Racer159, line 479: This should be moved to the `k3s` component of the init package as an `onDeploy` action

src/internal/cluster/common.go
```diff
@@ -28,7 +28,7 @@ var labels = k8s.Labels{
 
 // NewClusterOrDie creates a new cluster instance and waits up to 30 seconds for the cluster to be ready or throws a fatal error.
 func NewClusterOrDie() *Cluster {
-	c, err := NewClusterWithWait(defaultTimeout, true)
+	c, err := NewClusterWithWait(DefaultTimeout, true)
```
Racer159, line 31: DefaultTimeout should be replaced in other places that use the 30 second timeout for clusters like line 224 of `deploy.go`.

Racer159 (changes requested):
Sorry it took so long to get to.

src/pkg/packager/common.go
lucasrod16: Good catch. One of the other checks done in that loop over the components was to check if there are any container images in the package. The idea was to ignore verifying architectures match if there aren't, since that is (I believe) the root cause of problems with mismatched architectures. If there are no images in the package, should we be concerned about failing the deployment?

src/pkg/packager/common.go
lucasrod16: Do you see this being a `zarf internal` command, or using a shell script?

src/internal/cluster/common.go
```diff
@@ -18,7 +18,7 @@ type Cluster struct {
 }
 
 const (
-	defaultTimeout = 30 * time.Second
+	DefaultTimeout = 30 * time.Second
```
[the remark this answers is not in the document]
lucasrod16: Comment added

src/internal/cluster/common.go
lucasrod16: Changes made

src/pkg/packager/common.go
lucasrod16: Changes made

src/pkg/packager/common.go
lucasrod16: Changes made

src/pkg/packager/common.go
Racer159: Probably a shell script in the action - because `k3s` is linux only you could most likely use `uname -r` or `arch`.  Note that that won't quite match the goarch through so there may be some additional checking needed.

src/pkg/packager/common.go
Racer159: I think we can be simpler and just check architectures - packages without images are somewhat a can of worms.  We could use the YOLO mode metadata flag to never fail on YOLO packages though

src/pkg/packager/common.go
lucasrod16: Changes made. Will do some experimenting with using an `onDeploy` action

lucasrod16: @Racer159 the error message when `zarf init` fails doesn't clearly describe why. Is there a way to ensure the error message is clear as to why the deployment failed due to architecture mismatch when having this logic in the `onDeploy` action section of a Zarf init package? https://github.com/defenseunicorns/zarf/actions/runs/4854820097/jobs/8652653901?pr=1495#step:6:4473

lucasrod16: For further clarification, I already tried to `echo` a custom error message, but it doesn't get reported out on failure from the `onDeploy` action.

Racer159: Would something like this work?

```
        before:
          - cmd: "if [ \"hello\" != \"world\" ]; then echo \"I failed\" && exit 1; fi"
```

lucasrod16: https://github.com/defenseunicorns/zarf/actions/runs/4855678706/jobs/8654471359?pr=1495#step:6:4467

Looks like adding an echo statement doesn't do anything. The error gets reported as failing on `systemctl start k3s`.

Maybe checking for exit codes other than 0 when running actions and exiting immediately would fix this?

packages/distros/k3s/zarf.yaml
```diff
@@ -27,6 +27,12 @@ components:
       - source: https://github.com/k3s-io/k3s/releases/download/v1.24.1+k3s1/k3s-airgap-images-amd64.tar.zst
         shasum: 6736f9fa4d5754d60b0508bafb2f888170cb99a2d93a3a1617a919ca4ee74034
         target: /var/lib/rancher/k3s/agent/images/k3s.tar.zst
+    actions:
+      onDeploy:
+        before:
+          - cmd: if [ "$(arch)" != "x86_64" ]; then echo "this package architecture is amd64, but the target cluster has the arm64 architecture. These architectures must be the same" && exit 1; fi
```
Racer159, line 33:
```suggestion
          - cmd: if [ "$(arch)" != "x86_64" ]; then echo "this package architecture is amd64, but the target system has a different architecture. These architectures must be the same" && exit 1; fi
```

src/test/e2e/29_mismatched_architectures_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+// TestMismatchedArchitectures ensures that zarf produces an error
+// when the package architecture doesn't match the target cluster architecture.
+func TestMismatchedArchitectures(t *testing.T) {
+	t.Log("E2E: Mismatched architectures")
+	e2e.SetupWithCluster(t)
+	defer e2e.Teardown(t)
+
+	var (
+		mismatchedArch         = e2e.SetMismatchedArch()
```
Racer159, line 22:
```suggestion
		mismatchedArch         = e2e.GetMismatchedArch()
```
If the above suggestion is taken this sill be needed.

src/pkg/packager/common.go
```diff
@@ -443,6 +444,24 @@ func (p *Packager) validatePackageChecksums() error {
 	return nil
 }
 
+// validatePackageArchitecture validates that the package architecture matches the target cluster architecture.
+func (p *Packager) validatePackageArchitecture() error {
+	// Attempt to connect to a cluster to get the architecture.
+	if cluster, err := cluster.NewClusterWithWait(cluster.DefaultTimeout, false); err == nil {
```
Racer159, line 450:
```suggestion
	if cluster, err := cluster.NewCluster(); err == nil {
```

We should not wait for a cluster here since this would cause a 30s delay to any Zarf packages that didn't use a cluster or that deployed a cluster themselves.

src/test/e2e/20_zarf_init_test.go
```diff
@@ -27,13 +28,30 @@ func TestZarfInit(t *testing.T) {
...
+	// Build init package with different arch than the cluster arch.
+	stdOut, stdErr, err := e2e.ExecZarfCommand("package", "create", ".", "--architecture", mismatchedArch, "--confirm")
+	require.NoError(t, err, stdOut, stdErr)
+	defer e2e.CleanFiles(mismatchedInitPackage)
+
+	// Check that `zarf init` fails in appliance mode when we try to initialize a k3s cluster
+	// on a machine with a different architecture than the package architecture.
+	// We need to use the --architecture flag here to force zarf to find the package.
+	_, stdErr, err = e2e.ExecZarfCommand("init", "--architecture", mismatchedArch, "--components=k3s", "--confirm")
+	require.Error(t, err, stdErr)
```
Racer159, line 46: It would be good to check the error message here as you did in the below test.

Racer159 (changes requested):
sorry that took so long

src/test/e2e/20_zarf_init_test.go
lucasrod16: Changes made

Racer159 (approved):
lgtm!

Noxsios (approved):

PR #1520, opened 2023-03-30
Fix issues with Windows dataInjections

### LINKED ISSUE #1513: Data Injection - Windows Support
The data injection feature currently is not compatible when zarf is started from a Windows Client. In theory, the commands used should be windows compatible e.g. kubectl exec, tar, ...

The main reason it does not work however is that the responsible code block for "Data Injection" requires a shell to be present, it uses "sh -c" in order to start the kubectl commands. As windows does not have a shell by default, it fails.

https://github.com/defenseunicorns/zarf/blob/main/src/internal/cluster/data.go#L81

### Additional context
It would help to have table to see any known cross platform issues / not implemented features.

Racer159:
## Description

This selects the correct shell on Windows for dataInjections.

## Related Issue

Fixes #1513 

## Type of change

## Checklist before merging


src/internal/cluster/data.go
```diff
@@ -34,7 +34,7 @@ func (c *Cluster) HandleDataInjection(wg *sync.WaitGroup, data types.ZarfDataInj
 
 	tarCompressFlag := ""
 	if data.Compress {
-		tarCompressFlag = "z"
+		tarCompressFlag = " -z"
```
YrrepNoj, line 37: Putting a space at the beginning of this flags feels a bit weird to me 🤷‍♂️

src/internal/cluster/data.go
```diff
@@ -64,48 +72,50 @@ iterator:
...
 			}
 
-			kubectlExec := fmt.Sprintf("%skubectl exec -i -n %s %s -c %s ", binPath, data.Target.Namespace, pod, data.Target.Container)
-			tarExec := fmt.Sprintf("tar c%s", tarCompressFlag)
-			untarExec := fmt.Sprintf("tar x%svf - -C %s", tarCompressFlag, data.Target.Path)
+			kubectlCmd := fmt.Sprintf("%s exec -i -n %s %s -c %s ", kubectlBinPath, data.Target.Namespace, pod, data.Target.Container)
+
+			// Note that each command flag is separated to provide the widest cross-platform tar support
+			tarCmd := fmt.Sprintf("tar -c%s -f -", tarCompressFlag)
+			untarCmd := fmt.Sprintf("tar -x%s -v -f - -C %s", tarCompressFlag, data.Target.Path)
```
YrrepNoj, line 87: I know this works because we have a space in the flag (that gets set above earlier in the function), but since we have the power to string format this, I prefer something like `-c %s`  over `-c%s` since that more clearly shows that it's two different flags instead of trying to customize what the name of a single flag is.

YrrepNoj (approved):
I had a small nit, but it's more of a personal preference thing and not something I would prevent from merging in.

src/internal/cluster/data.go
Racer159, line 87:
```suggestion
			tarCmd := fmt.Sprintf("tar -c %s -f -", tarCompressFlag)
			untarCmd := fmt.Sprintf("tar -x %s -v -f - -C %s", tarCompressFlag, data.Target.Path)
```

src/internal/cluster/data.go
Racer159, line 37:
```suggestion
		tarCompressFlag = "-z"
```

YrrepNoj (approved):

Issue #1500, opened 2023-03-28
Add Zarf UI Navigation Drawer


Madeline-UX:
### Is your feature request related to a problem? Please describe.

**As** Ashton **I want** to see a left navigation drawer with tabs for each page **so that** I can easily know where I am in the application and navigate to other features easily

### Describe the solution you'd like

- [x] Left drawer with header for cluster context
- [x] Remove cluster context from global nav bar
- [x] Cluster context has error state with disconnected and shows cluster context when connected
- [x] Package page
- [x] Tools section - SBOM Compare
- [x] Ignore settings (for now)
- [x] ignore documentation (for now)

## Designs

Navigation component in Figma https://www.figma.com/file/MUItIzpzLBLuIyt225Bwgl/Zarf-UI?node-id=1081%3A54952&t=u5g95QRBwda6Qwgw-1

![Navigation](https://user-images.githubusercontent.com/92826525/228370426-5b471cef-7581-488e-9259-1d5dbbca5a15.png)

### Additional context

Related to #1396

Issue #1311, opened 2023-01-31
Landing page - Packages 


Madeline-UX:
## Big Components ##

- [x] Cluster section (states: not connected, connected not initialized, connected Initialized)
- [x] Table of packages (name, version, chips) 
- [x] Connect Cluster UX flow
- [x] Deploy Cluster UX Flow
- [x] Navigation (Global and left)

*Ignore - signed stuff

**As** Ashton **I want** to be able to connect to a cluster and deploy packages **so that** I can setup services that my organization needs.

### Connect a Cluster (Kubeconfig available)
**Given** - the user does not have a cluster connected (initialized with Zarf)
**Then** - show connect cluster button and disable the deploy package button
**When** - the user clicks connect cluster
**Then** - scan kubeconfig for available contexts
**And** - show a modal that lists the default cluster default (radio button to selected state)
**If** - User clicks Connect cluster - start to deploy wizard
**If** - User clicks cancel - return to the main screen. (Packages)

### Connect a Cluster (No Kubeconfig)
**Given** - the user does not have a cluster connected (initialized with Zarf)
**Then** - show connect cluster button and disable the deploy package button
**When** - the user clicks connect cluster
**Then** - scan kubeconfig for available contexts
**If**  - on Linux show a dialog with a button to deploy the init-package (to install k3s)
**when** - the user clicks deploy init
**Then** - start to deploy wizard
**If** - not on linux show a dialog message explaining user needs to create or connect to a cluster to connect.

### Initialize a Cluster
**Given** - the user has a cluster connected
**When** - a user clicks initialize cluster button
**Then** - they will be directed to step one of the deploy modal ([select package](https://github.com/defenseunicorns/zarf/issues/807))
**And** - the user should only see init packages (filter by Zarfinitconfig package type)
**When** - a user selects the init package
**Then** - they will be taken through the rest of the deploy flow
... [deploy flow](https://github.com/defenseunicorns/zarf/issues/807)
**When** - the user has deployed the init package 
**Then** - (if failed) stay on deployed page
**And** - show the error page and recommend user to go to the CLI to get logs
**Then** - (if successful) show success dialog
**Then** - redirect to package page (default)
**And** - Show Cluster Context and details
**And** - show deployed package is added to package table 
**And** - deploy button is now active

### Deploy Package
**Given** - a user has clicked the deploy package button
**Then** - they will be directed to step one of the deploy modal ([select package](https://github.com/defenseunicorns/zarf/issues/807))
[...]

Issue #964, opened 2022-11-01
Add information tooltips to elements in Zarf package description


Madeline-UX:
**Is your feature request related to a problem? Please describe.**
Desired outcome: 
When Ashton doesn't understand what a term means in the zarf package description, they can quickly and easily learn what they need to about that element to complete their task, without leaving the UI. 

**Describe the solution you'd like**
Add an information icon to package sections that provide more information on each Zarf package section

When Ashton hovers over the info button
the a tooltip with more information appears
Ashton reads the tool tip
then completes the section's task successfully.

Prototype
<img width="1161" alt="Screen Shot 2022-11-01 at 5 11 10 PM" src="https://user-images.githubusercontent.com/92826525/199351694-b634333d-9b77-4f89-92d7-9388300d938a.png">

**Additional context**
This feature is in response to usability testing
Users progress through task was interrupted by trying to understand what elements were in a package
The most confusing were:
- components
- SBOM
- Variables

Issue #807, opened 2022-09-28
Deploy package in Cluster UI


Madeline-UX:
## User Story
**As** Ashton **I want** to select from a list of packages **so that** I can visually see which packages I am able to deploy into my cluster.

**Given** I am on the main cluster overview page
**When** I click Deploy new package button on deployment page
**Then** the deployment wizard opens
**And** starts with the screen that shows me all of the local Zarf packages on my system
**And** I see the name of the package with a button next to it to deploy
**When** I click deploy package 
**Then** the wizard proceeds to step 2 (Step 1 of init wizard)

## Technical ##

- [x] Zarf needs to search for Zarf packages on the local system
- [x] When user selects package to deploy > $ Zarf package deploy [package name]

Madeline-UX:
@YrrepNoj @jeff-mccoy Can you add your thoughts about how to do this?

jeff-mccoy:
Good news, we already did the backend for that, just need to wire a ui to it 🤓

Madeline-UX:
yay!!

Racer159:
An api endpoint for this exists.

PR #1522, opened 2023-03-30
Zarf UI rewrite

mike-winberry:
## Description
Fixes: #1500 
Fixes: #1311 

## Closes 
Fixes: #964 
Fixes: #807 
Fixes: #794 
Fixes: #1500 
Fixes: #1311


src/internal/api/packages/find.go
```diff
@@ -6,18 +6,18 @@ package packages
...
 
 	"github.com/defenseunicorns/zarf/src/internal/api/common"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
-	"github.com/defenseunicorns/zarf/src/pkg/packager"
-	"github.com/defenseunicorns/zarf/src/pkg/utils"
 )
 
-var packagePattern = regexp.MustCompile(`zarf-package-.*\.tar`)
-var initPattern = regexp.MustCompile(packager.GetInitPackageName(""))
+var packagePattern = regexp.MustCompile(`zarf-package.*.tar\.zst$`)
```
Noxsios, line 19:
```suggestion
var packagePattern = regexp.MustCompile(`zarf-package-.*\.tar`)
```

Packages can be created uncompressed, (ie `zarf-package-name.tar`), so the `.zst` extension cannot be used an an identifier.

src/internal/api/packages/find.go
```diff
@@ -44,11 +44,44 @@ func findPackage(pattern *regexp.Regexp, w http.ResponseWriter, setDir func() (s
...
+		if err != nil {
+			// skip files and dirs we don't have permission to view.
+			if os.IsPermission(err) {
+				return nil
+			}
+			return err
+		}
+
+		// Skip hidden directories
+		if d.IsDir() && d.Name()[0] == dotcharater {
```
Noxsios, line 70: This will skip the default `~/.zarf-cache`, which is where the current init package is cached (and where future caching will be defaultly located)

src/ui/lib/components/build-providence.svelte
```diff
@@ -0,0 +1,51 @@
+<script lang="ts">
+	import { Box, Typography, type SSX, currentTheme } from '@ui';
+	import type { ZarfBuildData } from '$lib/api-types';
+
+	export let build: ZarfBuildData | undefined;
+	const labels = ['terminal', 'user', 'architecture', 'timestamp', 'version'];
```
Noxsios, line 6: Is there a way to extract the keys from the `ZarfBuildData` interface?

src/ui/lib/components/button-dense.svelte
```diff
@@ -0,0 +1,24 @@
...
+	<slot />
+</Button>
+
+<style global>
+	.box.button.dense {
+		height: 30px !important;
+		padding: 4px 10px !important;
+		font-size: small !important;
+	}
+</style>
```
Noxsios, line 24: Is there an issue / PR on unicorn UI for this custom component, doesn't look like it would be a huge lift for @TristanHoladay to implement.

src/ui/lib/components/cluster-info-table.svelte
```diff
@@ -0,0 +1,120 @@
...
+	<div class="cluster-info-table-column">
+		<Typography variant="overline">resources</Typography>
+		<Box class="label-values-container">
+			<Typography element="div" variant="caption" class="label-values">
+				<Typography variant="inherit">MEM:</Typography>
+				<Typography variant="inherit">CPU:</Typography>
+			</Typography>
+			<Typography element="div" variant="caption" class="label-values" style="font-weight:500;">
+				<Typography variant="inherit">n/a</Typography>
+				<Typography variant="inherit">n/a</Typography>
```
Noxsios, line 108: This should prob have a TODO on it

src/ui/lib/components/cluster-info.svelte
```diff
@@ -0,0 +1,73 @@
...
+	<Paper class="cluster-info-body" elevation={1}>
+		{#if !$clusterStore}
+			<div class="cluster-not-connected" in:fade={{ duration: 1000 }}>
+				<Typography variant="body1" color="blue-200">Searching for cluster.</Typography>
+				<Spinner color="blue-200" />
+			</div>
+		{:else if !showClusterInfo}
+			<Typography class="cluster-not-connected" variant="body1" element="span">
+				<span class="material-symbols-outlined" style="color:var(--warning);"> warning </span>
+				<span>&nbsp;Cluster not connected </span>
```
Noxsios, line 66: Is the `&nbsp;` needed here?

src/ui/lib/components/connect-cluster-dialog.svelte
```diff
@@ -0,0 +1,130 @@
...
+					backgroundColor="grey-300"
+					textColor="text-primary-on-light"
+				>
+					Connect Cluster
+				</Button>
+			</DialogActions>
+		{:else}
+			<Typography variant="caption" color="text-secondary-on-dark">
+				Zarf requires access to a cluster. Please install a cluster or login to your cluster
+				provider then try connecting to cluster again.
```
Noxsios, line 118: I think this verbiage can be improved.

src/ui/lib/components/deployment-actions.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts"></script>
```
Noxsios, line 1: This can be removed if there is no code.

src/ui/lib/components/deployment-actions.svelte
```diff
@@ -0,0 +1,12 @@
...
+
+<div class="deployment-actions" aria-label="action buttons">
```
Noxsios, line 3: When should `div` vs `Box` be used?

src/ui/lib/components/deployed-package-table.svelte
```diff
@@ -0,0 +1,158 @@
+<script lang="ts">
```
Noxsios, line 1: SPDX

src/ui/lib/components/deployed-package-menu.svelte
```diff
@@ -0,0 +1,113 @@
+<script lang="ts">
```
Noxsios, line 1: SPDX

src/ui/lib/components/connect-cluster-dialog.svelte
```diff
@@ -0,0 +1,130 @@
+<script lang="ts">
```
Noxsios, line 1: SPDX

src/ui/lib/components/cluster-info.svelte
```diff
@@ -0,0 +1,73 @@
+<script lang="ts">
```
Noxsios, line 1: SPDX

src/ui/lib/components/button-dense.svelte
```diff
@@ -0,0 +1,24 @@
+<script lang="ts">
```
Noxsios, line 1: SPDX

src/ui/lib/components/build-providence.svelte
```diff
@@ -0,0 +1,51 @@
+<script lang="ts">
```
Noxsios, line 1: SPDX

src/ui/lib/components/global-navigation-bar.svelte
```diff
@@ -0,0 +1,40 @@
+<!-- 
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+ -->
+<script lang="ts">
+	import ZarfLogoDark from '../../images/png/global-navigation-zarf-logo.png';
```
Noxsios, line 6: Use `@images` instead of relative paths.

src/ui/lib/components/deployment-actions.svelte
Noxsios: Also SPDX

src/ui/lib/components/local-package-table.svelte
```diff
@@ -0,0 +1,195 @@
+<script lang="ts">
```
Noxsios, line 1: SPDX

src/ui/lib/components/nav-drawer.svelte
```diff
@@ -0,0 +1,69 @@
+<script lang="ts">
```
Noxsios, line 1: SPDX

src/ui/lib/components/nav-link.svelte
```diff
@@ -0,0 +1,38 @@
+<script lang="ts">
```
Noxsios, line 1: SPDX

src/ui/lib/components/spinner.svelte
```diff
@@ -1,59 +1,26 @@
-<!-- 
-// SPDX-License-Identifier: Apache-2.0
-// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
- -->
```
Noxsios, line 4: SPDX

src/ui/lib/components/zarf-chip.svelte
```diff
@@ -0,0 +1,53 @@
+<script lang="ts">
```
Noxsios, line 1: SPDX + issue to Unicorn UI

src/ui/lib/http.ts
```diff
@@ -70,6 +74,33 @@ export class HTTP {
...
+		// 	.then((response) => {
+		// 		if (!response?.body) {
+		// 			throw new Error('Unable to get reader');
+		// 		}
+		// 		response.body.pipeThrough(new TextDecoderStream()).pipeTo(eventStream);
+		// 	})
+		// 	.catch((error) => {
+		// 		eventTarget.dispatchEvent(new CustomEvent('error', { detail: error }));
+		// 	});
+		// return eventTarget;
```
Noxsios, line 100: remove if dead code

src/ui/routes/+layout.svelte
```diff
@@ -1,40 +1,47 @@
-<!--
```
Noxsios, line 1: SPDX

src/ui/routes/+page.svelte
```diff
@@ -1,74 +1,48 @@
-<!--
```
Noxsios, line 1: SPDX

src/ui/routes/packages/+layout.svelte
```diff
@@ -1,7 +1,65 @@
-<!--
```
Noxsios, line 1: SPDX

src/ui/routes/packages/+page.svelte
```diff
@@ -1,52 +1,7 @@
-<!--
```
Noxsios, line 1: SPDX

src/ui/lib/components/nav-link.svelte
Noxsios: + Unicorn UI Issue

src/internal/api/packages/find.go
mike-winberry: Just curious when it begins being stored in the .zarf-cache is that a recent change? as this is derived from the io.go ListDirectorys that was previously here and didn't work but it currently is not having an issue. Was just curious about the inconsistency there and wether or not that ListDirectories method needs to be updated as well if the .zarf-cache is going to need to be explorable.

mike-winberry: Or is the .zarf-cache used when running through the brew package and not compiled from source?

Noxsios: No the `~/.zarf-cache` is the default location no matter where Zarf is built from and is governed by the `--zarf-cache` CLI flag / configuration.  When you do a `zarf init` and choose to download the init package, it will store it inside this folder. Currently, zarf packages are not stored in the cache as they are spit out to the location specified w/ the `-o` flag on `zarf package create`.

Noxsios: This list packages function will also be extended in the future to query OCI registries for Zarf packages stored.

src/internal/api/packages/find.go
mike-winberry: I think there might be bigger issues that will have to be looked into because the zarf cache gives this error
Unable to read the package at: `%2FUsers%2Fmikewinberry%2F.zarf-cache%2Fzarf-init-amd64-v0.25.0-20-g75b142a4.tar.zst` Its the only package that seems to be unreadable.

src/internal/api/packages/find.go
mike-winberry: I think there might be bigger issues that will have to be looked into because the zarf cache gives this error
Unable to read the package at: %2FUsers%2Fmikewinberry%2F.zarf-cache%2Fzarf-init-amd64-v0.25.0-20-g75b142a4.tar.zst Its the only package that seems to be unreadable.

src/ui/lib/components/build-providence.svelte
mike-winberry: I would either need to know which keys I want, or which keys I don't want to ignore the ones we aren't using each solution has the same result on just requires more comparisons.

mike-winberry: Those keys are also used as labels so I would say by explicitly declaring them is probably the cleaner solution imo than filtering the keys I don't want since that would be introducing explicit strings that are only used to not display something rather than strings that are explicitly relevant.

src/ui/lib/components/connect-cluster-dialog.svelte
mike-winberry: @Madeline-UX

src/ui/lib/components/deployment-actions.svelte
mike-winberry: I typically use box when I need things like ssx or event bubbling or any advanced features that would be necessary when composing this component, if it acts purely as a wrapper with no need to bring global in to the css I use div. But I would say it is mostly up to the developers discression, but the I would say that UnicornUI promotes componitization over semantic and global css classes. If I am not extending or taking advantage of the addition props than I don't need to use Box.

src/internal/api/packages/find.go
mike-winberry: Disregard seems to have been an issue with an old zarf-init package in my cache.

src/ui/lib/components/button-dense.svelte
mike-winberry: https://github.com/defenseunicorns/UnicornUI/issues/197

src/ui/lib/components/nav-link.svelte
mike-winberry: https://github.com/defenseunicorns/UnicornUI/issues/198

src/ui/lib/components/zarf-chip.svelte
mike-winberry: https://github.com/defenseunicorns/UnicornUI/issues/199

src/ui/lib/components/connect-cluster-dialog.svelte
mike-winberry: Madeline and I are gonna tag tomorrow for a design review Ill bring it up with her. Do you have suggestions if you are not around when we tag up?

Noxsios: My only note would be that the common verbiage is to `spin up` a cluster rather than `install` one, and to `authenticate` to a cluster rather than logging into a provider.

Madeline-UX: @Noxsios Thanks for the context. I updated the wording to include the more accurate term "spin up" cluser but kept "log in to cluster provider" as that is the actual task one would need to do if using Zarf with ex. AWS - if authenticate is separate and there is another step they need to do after login in we can add that. 

"Zarf requires access to a cluster. Please spin up a cluster or log in to your cluster provider, then try connecting to the cluster again. "

src/ui/lib/components/connect-cluster-dialog.svelte
```diff
@@ -0,0 +1,134 @@
...
+	$: hasDistro = $clusterStore?.distro;
+	$: {
+		if (hasDistro) {
+			titleText = 'Kubeconfig Found';
+			titleImage = ZarfKubeConfigFound;
+			titleAlt = 'Zarf, A happy axolotl has found a kubeconfig';
+		} else {
+			titleText = 'Kubeconfig Not Found';
+			titleImage = ZarfKubeConfigNotFound;
+			titleAlt = 'Zarf, A sat axolotl was unable to find a kubeconfig';
```
Racer159, line 74:
```suggestion
			titleAlt = 'Zarf, A sad axolotl was unable to find a kubeconfig';
```

src/ui/lib/components/connect-cluster-dialog.svelte
Racer159: deploy is another common word instead of "install"

src/ui/routes/packages/[name]/review/+page.svelte
```diff
@@ -0,0 +1,69 @@
...
+		>Supply chain is used to help determine if a package can be trusted. It includes declarative
+		data regarding how this package was built. Build providence includes metadata about the build
+		and where the package was created. SBOM includes information on all of the code, images, and
+		resources contained in this package.
+	</span>
+</SectionHeader>
+<BuildProvidence build={$pkgStore.zarfPackage.build} />
+
+<SectionHeader
+	>Components
```
Comment (Racer159) [inline src/ui/routes/packages/[name]/review/+page.svelte:42]:
```suggestion
<SectionHeader>
  Components
```
I've noticed this across a few files, is this a stylistic thing? Normally a tag is kept together rather than having its end split onto the next line.

src/ui/lib/components/package-section-header.svelte
```diff
@@ -6,13 +6,44 @@
 	import Icon from './icon.svelte';
 	import Tooltip from './tooltip.svelte';
 	import type { IconVariant } from './icon.svelte';
+	import { Paper, type SSX, Typography } from '@ui';
 	export let icon: IconVariant = 'package';
+
+	const ssx: SSX = {
```
Racer159, line 12: I can understand what it is and won't block this PR for it, but I have some concerns about how bespoke SSX is for potential newcomers to the project.

image

Racer159 (changes requested):

src/ui/routes/packages/[name]/review/+page.svelte
Comment (mike-winberry) [reply src/ui/routes/packages/[name]/review/+page.svelte:42]: This seems to be svelte issue that I have not found a solid solution for.

src/ui/lib/components/package-section-header.svelte
mike-winberry: https://unicorn-ui.netlify.app/scoped-styles

Madeline-UX (dismissed):
Design Review Complete and approved. Requested changes made and reviewed synchronously.

src/internal/api/packages/find.go
```diff
@@ -44,11 +44,38 @@ func findPackage(pattern *regexp.Regexp, w http.ResponseWriter, setDir func() (s
 		return
 	}
 
-	files, err := utils.RecursiveFileList(targetDir, pattern)
+	files, err := recursiveFileListSkipPermissionErrors(targetDir, pattern)
```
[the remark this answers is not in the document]
mike-winberry: Thats not the only thing thats changed for this, it is also using the entire file path against the regex instead of the title against the file. Should I bring that into the utils one as well.

src/internal/api/auth/token.go
```diff
@@ -12,7 +12,8 @@ import (
 func RequireSecret(validToken string) func(http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			if r.Header.Get("Authorization") != validToken {
+
+			if r.Header.Get("Authorization") != validToken && r.URL.Query().Get("auth") != validToken {
```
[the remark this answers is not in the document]
mike-winberry: This is a requirement for default ssr events, I do agree the implementation should be looked into and under less time restrictive circumstances, I also think the auth is pretty weak in general if the expectation is to open this and leave it deployed somewhere. especially since the token is just stored in session.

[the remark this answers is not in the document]
mike-winberry: I may just bring in the Microsoft fetch event source to solve this, I'll play with it and make sure it works but that should fix this and move back to header only

src/internal/api/packages/deploy.go
```diff
@@ -45,3 +50,37 @@ func DeployPackage(w http.ResponseWriter, r *http.Request) {
 
 	common.WriteJSONResponse(w, true, http.StatusCreated)
 }
+
+func StreamDeployPackage(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
mike-winberry: Adding them, Just curious if the linter is something I need to manually run?

src/internal/api/auth/token.go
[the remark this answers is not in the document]
mike-winberry: I ended up removing this and moving forward with @microsoft/fetch-event-source to resolve the need for the query param.

src/ui/lib/components/package-section-header.svelte
mike-winberry: https://defenseunicorns.blogin.co/posts/unicorn-ui-why-scopedstyles-200989?s=svelte

mike-winberry: architected after https://mui.com/system/getting-started/the-sx-prop/

src/ui/routes/packages/[name]/review/+page.svelte
Comment (mike-winberry) [reply src/ui/routes/packages/[name]/review/+page.svelte]: I fixed it, Had to set the htmlWhitespaceSensitivity to ignore in prettierrc

src/internal/api/packages/find.go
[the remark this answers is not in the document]
mike-winberry: I added the required flags and updated existing references to maintain their current behavior.

mike-winberry:
- [ ] Remove Confirmation needs to be added.

mike-winberry:
- [ ] Add remove package flow after selecting remove in the hamburger menu.

src/internal/api/packages/deploy.go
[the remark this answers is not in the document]
Racer159: It is something you can install via pre commit: https://github.com/defenseunicorns/zarf/blob/main/CONTRIBUTING.md#optional-pre-commit-hooks-and-linting

src/ui/lib/components/package-section-header.svelte
Racer159: Yep yep tracking our own implementation - just wary of bespoke things - even calling it the "sxProp" over ssx may be helpful for people coming to the project.

mike-winberry:
https://user-images.githubusercontent.com/86802655/232201606-152ec856-e834-481e-8a9a-7571e07f8556.mov

Racer159 (changes requested):

src/internal/api/packages/deploy.go
```diff
@@ -45,3 +50,42 @@ func DeployPackage(w http.ResponseWriter, r *http.Request) {
 
 	common.WriteJSONResponse(w, true, http.StatusCreated)
 }
+
+// StreamDeployPackage Establishes a stream that redirects pterm output to the stream
+// Resets the output to std.err after the stream connection is closed
+func StreamDeployPackage(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
mike-winberry: This has to do with the injector. It hijacks the output when `exec.CmdWithContext` is called. and in that method it looks like there is a deviation from from the common message which is called during deploy with the HandleDataInjection. Don't really have a solution right now, but can create an issue.

Racer159 (approved):
lgtm

Issue #1218, opened 2023-01-18
Pushing Images to zarf registry hangs forever


corang:
### Environment
Device and OS: Ubuntu 22
App version: 0.23.2+
Kubernetes distro being used: k3s in k3d
k8s version: v1.24.4+k3s1
Other:

### Steps to reproduce
1. Try to deploy zarf init package or any package, more images more chance of happening

### Expected result

All images are pushed to registry and package deployment continues

### Actual Result

Sometimes the package is pushed successfully (looking at registry logs) but the zarf CLI just hangs forever on "Updating Image xxx"

### Visual Proof (screenshots, videos, text, etc)

![image](https://user-images.githubusercontent.com/1349336/213303782-20489a17-bb4f-4034-b5ba-d843fc332509.png)
Has been on this image effectively the entire time

### Severity/Priority

High, I can't use zarf as it is now

corang:
The image push eventually finished but is happening on other images as well where they're taking way too long to push to registry for their size/this hardware

corang:
Last relevant registry log when CLI is hanging 
![image](https://user-images.githubusercontent.com/1349336/213308842-d00ad35a-a2d5-43c4-88ba-bbf5c76a56e7.png)

corang:
CLI while hanging with debug out 
![image](https://user-images.githubusercontent.com/1349336/213309207-0a048384-d8ac-416e-ba95-eb843afb8508.png)

corang:
Pod that was handling the image push was scaled out and deleted

corang:
CLI error `2023-01-18T16:36:24-06:00  -  "msg"="lost connection to pod\n" "error"=null`
The next attempt at pushing the image worked

corang:
Same error when pod isn't scaled out `2023-01-18T16:46:43-06:00  -  "msg"="lost connection to pod\n" "error"=null`

corang:
All http PUTs seem to work fine, every time a hang starts it's on an http PATCH but not every PATCH fails

corang:
PATCH log that never completes
```
[...]

PR #1558, opened 2023-04-05
Adding additional details for debugging about registry pushes

dgershman:
## Description

This enables some additional logging information for pushes to registries.

## Related Issue

Relates to #1218 

## Type of change

## Checklist before merging

### DIFF
--- a/src/internal/packager/images/push.go
+++ b/src/internal/packager/images/push.go
@@ -10,6 +10,7 @@ import (
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 	"github.com/defenseunicorns/zarf/src/pkg/utils"
 	"github.com/google/go-containerregistry/pkg/crane"
+	"github.com/google/go-containerregistry/pkg/logs"
 )
 
 // PushToZarfRegistry pushes a provided image into the configured Zarf registry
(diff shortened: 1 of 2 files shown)


src/internal/packager/images/push.go
```diff
@@ -52,6 +53,11 @@ func (i *ImgConfig) PushToZarfRegistry() error {
 	spinner := message.NewProgressSpinner("Storing images in the zarf registry")
 	defer spinner.Stop()
 
+	if message.GetLogLevel() >= message.DebugLevel {
+		logs.Warn.SetOutput(spinner)
```
Racer159, line 57:
```suggestion
	if message.GetLogLevel() >= message.DebugLevel {
		spinner.EnablePreserveWrites()
		logs.Warn.SetOutput(spinner)
```

If possible, can you also add this line in `src/internal/packager/images/pull.go` as well?  Looking at your code made me realize in many scenarios we are discarding this information when the spinner is destroyed.

Racer159 (changes requested):
Looks good overall, and thanks for the PR!  Requesting this line be added as well since it won't destroy the messages when the spinner completes (if it completes).

Also do you have any log snippets from when you saw the infinite spin with this in place?  (was testing locally over this past weekend and while I could significantly slow the Zarf install I never saw it actually infinite spin myself yet).

Racer159:
> 

Thanks, that is good to know - I was able to see the retry behavior happen in my tests by intentionally killing the connection to trigger them manually (by deleting or otherwise manipulating the registry pod) - what is odd is that is the exact opposite of what was reported by @corang (and others) https://github.com/defenseunicorns/zarf/issues/1218#issuecomment-1403703898

Racer159 (approved):
lgtm

YrrepNoj (approved):

dgershman:
> > 
> 
> Thanks, that is good to know - I was able to see the retry behavior happen in my tests by intentionally killing the connection to trigger them manually (by deleting or otherwise manipulating the registry pod) - what is odd is that is the opposite of what was reported by @corang (and others) [#1218 (comment)](https://github.com/defenseunicorns/zarf/issues/1218#issuecomment-1403703898) - more logs will definitely help since this is likely shaping up to be a multi-faceted issue

I started a new ticket to track my findings https://github.com/defenseunicorns/zarf/issues/1568.

PR #1559, opened 2023-04-05
Officially require 1.54.0 or greater of Big Bang now

dgershman:
## Description

This makes the requirement for the `bigbang` component officially 1.54.0 or greater.

## Related Issue

N/A

## Type of change

## Checklist before merging

### DIFF
--- a/src/extensions/bigbang/bigbang.go
+++ b/src/extensions/bigbang/bigbang.go
@@ -240,7 +240,7 @@ func isValidVersion(version string) bool {
 	minor, _ := strconv.Atoi(parts[1])
 
 	// This extension requires BB 1.54.0 or greater.
-	return major >= 1 && minor >= 54
+	return major == 1 && minor >= 54 || major > 2
 }
 
 // findBBResources takes a list of yaml objects (as a string) and
--- a/src/extensions/bigbang/bigbang.go
+++ b/src/extensions/bigbang/bigbang.go
@@ -142,7 +142,7 @@ func Run(tmpPaths types.ComponentPaths, c types.ZarfComponent) (types.ZarfCompon
 		}
 
 		// In Big Bang the metrics-server is a special case that only deploy if needed.
-		// The check it, we need to look for the APIService to exist instead of the HelmRelease, which
+		// The check it, we need to look for the existence of APIService instead of the HelmRelease, which
 		// may not ever be created. See links below for more details.
 		// https://repo1.dso.mil/big-bang/bigbang/-/blob/1.54.0/chart/templates/metrics-server/helmrelease.yaml
 		if hr.Metadata.Name == "metrics-server" {
--- a/src/extensions/bigbang/bigbang.go
+++ b/src/extensions/bigbang/bigbang.go
@@ -7,10 +7,9 @@ package bigbang
 import (
 	"fmt"
 	"path"
-	"strconv"
-	"strings"
 	"time"
 
+	"github.com/blang/semver/v4"
 	"github.com/defenseunicorns/zarf/src/internal/packager/helm"
 	"github.com/defenseunicorns/zarf/src/pkg/utils"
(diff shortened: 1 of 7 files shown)


Racer159 (changes requested):
Thanks for the PR! Can you also update the following?

https://github.com/defenseunicorns/zarf/blame/main/docs/13-walkthroughs/5-big-bang.md#L113
- Should say "This extension requires Big Bang version `1.54.0` or later."

https://github.com/defenseunicorns/zarf/blame/main/examples/git-data/README.md#L31
https://github.com/defenseunicorns/zarf/blob/main/examples/git-data/zarf.yaml#L34
- Should reference 1.54.x instead of 1.53.x

https://github.com/defenseunicorns/zarf/blob/main/src/extensions/bigbang/bigbang.go#L48
https://github.com/defenseunicorns/zarf/blob/main/src/extensions/bigbang/bigbang.go#L147
https://github.com/defenseunicorns/zarf/blob/main/src/extensions/bigbang/bigbang.go#L229
- Should reference 1.54.0 instead

https://github.com/defenseunicorns/zarf/blob/main/src/test/e2e/07_create_git_test.go#L63
https://github.com/defenseunicorns/zarf/blob/main/src/test/e2e/07_create_git_test.go#L68
- Because the `git-data` example would be changed above these should reference `release-1.54.x`

https://github.com/defenseunicorns/zarf/blob/main/src/test/e2e/07_create_git_test.go#L60
- Again because `git-data` changed, the folder `big-bang-2752174810` would change to `big-bang-2705706079` for the URL being updated to `https://github.com/DoD-Platform-One/big-bang.git@refs/heads/release-1.54.x`

dgershman:
I found an issue with this, when attempting to test against BB `1.57.0-rc.0` the `isValidVersion` function fails.  Testing a change out to support BB release candidate tags.

dgershman:
> I found an issue with this, when attempting to test against BB `1.57.0-rc.0` the `isValidVersion` function fails. Testing a change out to support BB release candidate tags.

Changes pushed.  I'm using a library already referenced by another package to parse the semantic version (based on Semver v2).  I also added some unit tests.

src/extensions/bigbang/bigbang.go
```diff
@@ -43,9 +43,15 @@ func Run(tmpPaths types.ComponentPaths, c types.ZarfComponent) (types.ZarfCompon
...
+
+	if err != nil {
+		return c, fmt.Errorf("invalid Big Bang version: %s, parsing issue %s", cfg.Version, err)
+	}
+
 	// Make sure the version is valid.
-	if !isValidVersion(cfg.Version) {
-		return c, fmt.Errorf("invalid Big Bang version: %s, must be at least 1.53.0", cfg.Version)
+	if !validVersionResponse {
+		return c, fmt.Errorf("invalid Big Bang version: %s, must be at least 1.54.0", cfg.Version)
```
Racer159, line 54:
```suggestion
		return c, fmt.Errorf("invalid Big Bang version: %s, must be at least %s", cfg.Version, bbMinRequiredVersion)
```
Since you already have `bbMinRequiredVersion` might as well do this.

src/extensions/bigbang/bigbang.go
```diff
@@ -7,10 +7,9 @@ package bigbang
 import (
 	"fmt"
 	"path"
-	"strconv"
-	"strings"
 	"time"
 
+	"github.com/blang/semver/v4"
```
Racer159, line 12: I know we pull in both, but can we use https://github.com/Masterminds/semver/v3 instead since it is a more active project?  (for context it is also the one we had previously directly used in Zarf in the past)

Racer159 (changes requested):
Just two small suggestions

Racer159 (approved):
lgtm!

YrrepNoj (approved):

PR #1573, opened 2023-04-07
Fix EKS package and add workflow for nightly EKS testing

YrrepNoj:
- Update our EKS package to create an ebs-csi-driver for the volume mounting necessary during a `zarf init` on an EKS cluster
- Add a couple more variables to the eks package so users have more control over the cluster they're creating
- Add commands during remove so that we can cleanup the eks cluster via a `zarf package remove` command

- Add the beginnings of a workflow for running nightly e2e tests on an EKS cluster


.github/workflows/night-ecr.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: make zarf-cli-linux-amd
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: "zarf-nightly-tester"
+          aws-region: us-east-1
+
+      - name: Login to the ECR Registry
+        run: aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/t8y5r5z5                                                                                                          ─╯
```
Racer159, line 37: public.ecr.aws/t8y5r5z5 this is kinda interesting - I guess that's a thing AWS does to obfuscate stuff

Racer159 (approved):
lgtm

Issue #1577, opened 2023-04-07
Package Create Image Registry Overrides


Racer159:
### Is your feature request related to a problem? Please describe.
When running package create in a controlled enterprise environment you may need to pull images on package create through pull-through image mirrors.  This is relatively easy to configure for internal packages that you maintain yourself, but for the init package or other skeleton packages (#1469) it may not be easy to override all of the downstream images.

### Describe the solution you'd like
A way on package create (likely through a new command flag) to provide a scoped list of mirrors to override, ideally based on domain name.  Something like:

```shell
$ zarf package create . --registry-override docker.io=dockerio-reg.enterprise.intranet --registry-override  ghcr.io=ghcrio-reg.enterprise.intranet
```

### Describe alternatives you've considered
We could have a more global override but that would not allow for the management of collisions (i.e. if nginx1:23.3 was pulled from both docker.io and ghcr.io - they could be different images technically) and the above pattern matches how many enterprise mirror services handle pull-through.

### Additional context
This is relatively common enterprise policy in large enough organizations to vet software at the edge of their network.

PR #1579, opened 2023-04-08
Refactor the hardcoded seed image out of the Zarf codebase

Racer159:
## Description

This refactors Zarf so that the `registry:2.8.1` image is no longer hard coded in the codebase (allowing it to be overridden and for the injector to be more flexible)

## Related Issue

Relates to #1577

## Type of change

## Checklist before merging


YrrepNoj (approved):
Poked through this a good bit and played around with it locally and it seemed to work out pretty well :)

PR #1581, opened 2023-04-08
Allow the selection between CMD and Powershell and prefer CMD for dataInjections

Racer159:
## Description

This allows package creators to select between cmd and powershell for actions and prefers cmd over powershell for dataInjections.

## Related Issue

Fixes #N/A

More info: https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1680880810057189?thread_ts=1680097325.822069&cid=C03B6BJAUJ3

## Type of change

## Checklist before merging


Racer159:
Also see https://github.com/orgs/PowerShell/discussions/16569

examples/component-actions/zarf.yaml
```diff
@@ -39,6 +39,8 @@ components:
           - cmd: touch test-create-after.txt
           - cmd: sleep 1
           - cmd: echo "I can print!"
+            # prefer to run the above command in "cmd" instead of "powershell" on Windows
+            preferLegacyShell: true
```
YrrepNoj, line 43: Instead of a `preferLegacyShell` toggle, what do you think about 'simply' having a `shell` field for the command where the user can specify the shell they want to use to run their command and if that shell doesn't exist we throw a warning and attempt to use the system default shell?

As it is right now, this field seems super specific and only has the niche use case which doesn't feel great

ntwkninja: Something similar to this @YrrepNoj ?

https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/ec2-windows-user-data.html

Racer159: I think that could work, but it would kinda be annoying to have a warning flash for a package deploying on linux that relates only to Windows...  Maybe a combination of OS/shell preference? Like

`windowsShell`, `darwinShell`, `linuxShell`?

YrrepNoj: @Racer159 That's a bit verbose but I agree that it's probably better to do it that way to avoid warnings when on different OS's.

Racer159: Changes made to add the following:

```
shell:
  windows: "pwsh"
  linux: "pwsh"
  darwin: "pwsh"
```

Noxsios (commented):
lgtm, do we want to add integration to `defaults` now or later?

Racer159:
> 

It is integrated with defaults.

YrrepNoj (approved):

YrrepNoj (approved):

Issue #1541, opened 2023-04-02
Show more user feedback when Zarf cannot connect to a cluster


Racer159:
### Is your feature request related to a problem? Please describe.
When Zarf cannot connect to a cluster it will stall after "Copying 1 files" for quite some time.  
![image](https://user-images.githubusercontent.com/3977569/229330111-536d4beb-7826-4a04-b18b-e2fbcab3f7ea.png)

### Describe the solution you'd like
We should show more clearly that we are trying to connect to a cluster but are unable to do so.  We should also likely lower the timeout for receiving a response.

### Describe alternatives you've considered
None

### Additional context
This is something that has impacted users with misconfigured `~/./kube/config` files in the real world.

Racer159:
Labeling this as a bug because the 30 second timeout we are setting is getting ignored.

Issue #1568, opened 2023-04-06
Intermittent Hangs on Registry Push


dgershman:
### Environment
Device and OS: AMD64 / Ubuntu 22.04 
App version: 0.25.2
Kubernetes distro being used: K3D K3S v1.25.7+k3s1
Other: Big Bang `1.57.1`

### Steps to reproduce
1. `zarf package deploy zarf-package-big-bang-example-amd64-1.57.1.tar.zst --confirm --log-level=trace`
2. Occasionally the process gets hung up "Updating image ..."

### Expected result
That things wouldn't get hung up, and continue along.

### Actual Result
Things get hung up and/or sometimes fail / timeout.

### Visual Proof (screenshots, videos, text, etc)

![Screenshot 2023-04-06 at 4 31 31 PM](https://user-images.githubusercontent.com/1850811/230488118-9a379604-83da-4da2-8aad-7dcc9c5e570c.png)

```
  DEBUG   2023-04-06T20:20:34Z  -  crane.Push() /tmp/zarf-1329914535/images:registry1.dso.mil/ironbank/fluxcd/kustomize-controller:v0.34.0 -> 127.0.0.1:34857/ironbank/fluxcd/kustomize-controller:v0.34.0)
└ (/home/runner/work/zarf/zarf/src/pkg/packager/deploy.go:403)
  ⠦  Updating image registry1.dso.mil/ironbank/fluxcd/kustomize-controller:v0.34.0 (2m5s))
  ⠧  Updating image registry1.dso.mil/ironbank/fluxcd/kustomize-controller:v0.34.0 (2m5s)
  DEBUG   2023-04-06T20:22:44Z  -  "msg"="error creating error stream for port 34857 -> 5000: Timeout occurred\n" "error"=null
└ (/home/runner/go/pkg/mod/github.com/go-logr/logr@v1.2.4/funcr/funcr.go:183)
  DEBUG   2023-04-06T20:24:01Z  -  "msg"="error creating error stream for port 34857 -> 5000: Timeout occurred\n" "error"=null
└ (/home/runner/go/pkg/mod/github.com/go-logr/logr@v1.2.4/funcr/funcr.go:183)
  DEBUG   2023-04-06T20:24:03Z  -  "msg"="error creating error stream for port 34857 -> 5000: Timeout occurred\n" "error"=null
└ (/home/runner/go/pkg/mod/github.com/go-logr/logr@v1.2.4/funcr/funcr.go:183)
[...]

Issue #1433, opened 2023-03-14
Intermitent Zarf (crane) image push failure (RKE2 + Longhorn)


mikevanhemert:
### Environment
Device and OS: 4 x 120-core, 512 GB servers, Ubuntu 22.04.2
App version: v0.25.0-rc1
Kubernetes distro being used: RKE2 v1.24.10+rke2r1

### Steps to reproduce
1. Stand up four-node RKE2 cluster with 1 master and 3 workers
2. Install Longhorn
3. Zarf init, install gitea
    ```bash
    zarf init --components git-server --confirm
    ```
4. Deploy [DCO Foundation](https://github.com/naps-dev/dco-foundation/tree/renew-certs) Zarf package on `renew-certs` branch
    ```bash
    zarf package deploy zarf-package-dco-foundation-amd64.tar.zst --components metallb --set BRANCH=renew-certs -l debug --confirm
    ```

### Expected result
Big Bang v1.52, KubeVirt, CDI, and MetalLB installed

### Actual Result
Intermittent Crane connection errors stop images from being loaded during Zarf package deploy

### Visual Proof (screenshots, videos, text, etc)
My terminal failed while filling out this ticket and I lost VPN and all the goods. I will circle back this evening.

### Additional Context
Failures don't happen on the same image every time. Re-running zarf package deploy several dozen times eventually gets us to success.

mikevanhemert:
Also worth noting that this same package deploys fine onto a `c5n.metal` AWS ec2 instance running single-node RKE2 cluster with localpath provisioner :upside_down_face:

corang:
@mikevanhemert Is this the same symptoms? https://github.com/defenseunicorns/zarf/issues/1218

mikevanhemert:
@corang Almost definitely! I will verify the behavior and report back

mikevanhemert:
Subsequent `zarf package deploy`
```
  ✔  Creating port forwarding tunnel at http://127.0.0.1:38489/v2/_catalog                                                                                                                                                                                                      
  DEBUG   2023-03-15T01:02:13Z  -  tunnel.Endpoint()
  DEBUG   2023-03-15T01:02:13Z  -  crane pushOptions = []crane.Option{(crane.Option)(0x1619260), (crane.Option)(0x2642b20)}                                                                                                                                                     
[...]

Issue #1364, opened 2023-02-16
Split image push into multiple tunnels


jeff-mccoy:
Although the zarf registry leverages HPA, if a single component (Big Bang in this case) has a large amount of images, there is no opportunity to take advantage of the multiple replicas since a single connection to the original pod will be reused. This dramatically reduces the value of HPA with a large set of images in one component.

![Screenshot 2023-02-16 at 3 07 40 AM](https://user-images.githubusercontent.com/882485/219319705-54d277a9-e0e7-4d8e-9986-f24ccb62e478.png)

PR #1590, opened 2023-04-11
Switch to per-image tunnels, CRC on tags, and progressbar for image pushes

Racer159:
## Description

This PR creates a tunnel per image push (making it easier to implement concurrency - may do that in this PR if we can confirm that issues are mitigated) moves the CRC from the image name to the tag and changes the UI to use a progressbar instead of a spinner for better user feedback.

## Related Issue

Relates to #1568 , #1433, #1218, #1364

This also will make #1594 slightly easier.

(See https://github.com/aws/containers-roadmap/issues/853)

Fixes: #1541

## Type of change

## Checklist before merging


src/pkg/transform/image.go
```diff
@@ -32,7 +32,7 @@ func ImageTransformHost(targetHost, srcReference string) (string, error) {
 	// Generate a crc32 hash of the image host + name
 	checksum := utils.GetCRCHash(image.Name)
 
-	return fmt.Sprintf("%s/%s-%d%s", targetHost, image.Path, checksum, image.TagOrDigest), nil
+	return fmt.Sprintf("%s/%s%s-zarf-%d", targetHost, image.Path, image.TagOrDigest, checksum), nil
```
YrrepNoj, line 35: Is this a breaking change? 

Is there something we can do to make this backwards compatible / lay out instructions as to how to upgrade?

Racer159: Yes, this is breaking - the upgrade path is to run `init` again and then redeploy zarf packages on top to reupload the images.

We might be able to pull image names/tags from their registry and check to see if they have the `-zarf-` tag and retag them if we wanted.

Racer159: I added the breaking change warning that we discussed.  I may wait on the warning rollup since that would need a bit more thought/design to make make sense (after playing with it simply reprinting them feels / looks weird)

src/internal/cluster/zarf.go
```diff
@@ -43,6 +43,21 @@ func (c *Cluster) GetDeployedZarfPackages() ([]types.DeployedPackage, error) {
 	return deployedPackages, nil
 }
 
+// GetDeployedZarfInitPackage gets the metadata information about the currently deployed init package in the cluster.
+// We determine what packages have been deployed to the cluster by looking for specific secrets in the Zarf namespace.
+func (c *Cluster) GetDeployedZarfInitPackage() (types.DeployedPackage, error) {
```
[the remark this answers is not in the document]
Racer159: that makes sense, refactored

src/pkg/packager/deprecated/common.go
```diff
@@ -37,3 +58,46 @@ func MigrateComponent(build types.ZarfBuildData, c types.ZarfComponent) types.Za
 	// Future migrations here.
 	return c
 }
+
+// PrintBreakingChanges prints the breaking changes between the provided version and the current CLIVersion
+func PrintBreakingChanges(deployedZarfVersion string) {
+	deployedSemver, err := semver.NewVersion(deployedZarfVersion)
+
```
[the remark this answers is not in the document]
Racer159: Fixed

YrrepNoj (dismissed):

YrrepNoj (approved):

PR #1622, opened 2023-04-18
Make the AWS role a repository secret

YrrepNoj:
Making the AWS Role we are assuming a repository secret so we can:
 1. Actually set it to the correct value
 2. Iterate on the role slightly easier (without needed more PRs)


Racer159 (changes requested):
While you're in here, can you fix these linting errors that got in as well
image

Racer159 (approved):

PR #1624, opened 2023-04-19
Fix linting and UI version check

### LINKED ISSUE #1614: zarf tools kubectl version returns error code
### Environment
Device and OS: OSX
App version: v0.25.2
Kubernetes distro being used: v1.23.16
Other:

When creating a pre-flight check for the kubernetes version, the zarf tools kubectl command was returning an error. It looks like the clientVersion is not getting set in a way the command likes, potentially based on how its being imported and used by Zarf tools.

### Steps to reproduce
Run the following:

```shell
$ zarf tools kubectl version --short        
Flag --short has been deprecated, and will be removed in the future. The --short output will become the default.
Client Version: v0.0.0-master+$Format:%H$
Kustomize Version: v4.5.7
Server Version: v1.23.16-eks-48e63af
error: could not parse pre-release/metadata (-master+$Format:%H$) in version "v0.0.0-master+$Format:%H$"
$ echo $?
1
```

### Expected result
For the return code for the command `zarf tools kubectl version --short ` to be zero.
### Severity/Priority

Simple work around by:

```yaml
  - name: one
    required: true
    actions:
      onDeploy:
        after:
        - cmd: "./zarf tools kubectl version --short || true"
          setVariable: OUTPUT
```

Racer159:
## Description

This cleans up the UI diff check script for local dev and a few linting issues.

## Related Issue

Fixes #1614 
 
## Type of change

## Checklist before merging

### DIFF
--- a/src/internal/packager/images/pull.go
+++ b/src/internal/packager/images/pull.go
@@ -7,6 +7,7 @@ package images
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -63,7 +64,7 @@ func (i *ImgConfig) PullAll() error {
 
 	// Create the ImagePath directory
 	err := os.Mkdir(i.ImagesPath, 0755)
-	if err != nil {
+	if err != nil && !errors.Is(err, os.ErrExist) {
 		return fmt.Errorf("failed to create image path %s: %w", i.ImagesPath, err)
 	}
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ jobs:
   push-resources:
     runs-on: ubuntu-latest
     permissions:
-      contents: read
+      contents: write
       packages: write
     steps:
       - name: Checkout
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -15,7 +15,7 @@ builds:
       - darwin
       - windows
     ldflags:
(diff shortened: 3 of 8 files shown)


src/internal/packager/images/pull.go
```diff
@@ -63,7 +64,7 @@ func (i *ImgConfig) PullAll() error {
 
 	// Create the ImagePath directory
 	err := os.Mkdir(i.ImagesPath, 0755)
-	if err != nil {
+	if err != nil && !errors.Is(err, os.ErrExist) {
```
YrrepNoj, line 67: This might be worth a warning, or double checking our workflow to make sure we are explicit on when we expect this directory to be created.

If this directory already exists when we are running `zarf package create` then there could be unintended contents in that directory that get added to the package (and the checksums since it was there during creation)

YrrepNoj: Super unlikely, but probably worth consideration

Racer159: Hmm... You would have to inject something into that directory after zarf create since the tmp dir everything is built from does not exist until that is ran.  If there was malware on the system that wanted to watch your commands all they would have to do now is move that watch to when the image directory is created... (and you would probably want to do this anyway since it would be better to manipulate images that are being run in the cluster rather than just having another image there (and any manifests or layers would get overwritten on crane pull).

This was specifically to fix the retry logic that will try to remake the dir and balk

PR #1633, opened 2023-04-20
1056 support (more) remote files in `zarf.yaml` manifests

### LINKED ISSUE #1056: Support remote files in zarf.yaml manifests
In the same vein as https://github.com/defenseunicorns/zarf/issues/1055, I want to ensure that when a Zarf package is being built, it's using the files that are checked into VCS, not on the user's local machine.

Consider the following `zarf.yaml`:
```
kind: ZarfPackageConfig
metadata:
  name: test
components:
  - name: upgrade-test
    required: true
    namespace: podinfo
    images:
      - ghcr.io/stefanprodan/podinfo:6.2.3
    repos:
      - https://github.com/UncleGedd/zarf-playground
      - https://github.com/stefanprodan/podinfo
    charts:
      - path-to-chart
    manifests:
      - name: podinfo
        namespace: podinfo
        files: 
            - https://github.com/UncleGedd/zarf-playground/blob/main/upgrade-testing/zarf.yaml # this is a remote file!
```
I'd like the `...manifest.files` to support remote files

Noxsios:
Support added for the following resources to be URLs:

- manifests + kustomizations
- data injectsions
- helm chart values files

files + helm charts already had support for URLs

additionally, URLs can be verified w/ SHA256 checksums, stored alongside the URL:

```
https://k8s.io/examples/application/deployment.yaml@c57f73449b26eae02ca2a549c388807d49ef6d3f2dc040a9bbb1290128d97157

github.com/stefanprodan/podinfo//kustomize?ref=0647aea75b85755411b007a290b9321668370be5
```

This also supports image discovery during `zarf prepare find-images`

## Related Issue

Fixes #1056 

## Type of change

## Checklist before merging


Noxsios:
leaving as a note for myself, the code used to test the URL parsing works properly for `@`:

```go
func DownloadToFile(src string) {
	parsed, err := url.Parse(src)
	if err != nil {
		message.Fatalf(err, "Unable to parse the URL: %s", src)
	}
	message.Infof("Downloading %s", src)
	hasChecksum := strings.Contains(parsed.Path, "@")
	var checksum string
	if hasChecksum {
		index := strings.LastIndex(parsed.Path, "@")
		checksum = parsed.Path[index+1:]
		parsed.Path = parsed.Path[:index]
		message.Infof("URL: %s", parsed.String())
		message.Infof("Checksum: %s", checksum)
	}
	message.Infof("hasChecksum: %t", hasChecksum)
}

func main() {
	testURLs := []string{
		"https://google.com",
		"https://hello.world/test@1234567890",
		"https://hello.world/test",
		"https://user:pass@hello.world/test@134509",
		"https://user:pass@hello.world/test",
	}
	for _, url := range testURLs {
		DownloadToFile(url)
		message.Info("-----")
	}
}
```

src/pkg/utils/network.go
```diff
@@ -70,26 +72,52 @@ func Fetch(url string) io.ReadCloser {
...
 	}
 
 	// Create the file
-	destinationFile, err := os.Create(target)
+	file, err := os.Create(dst)
 	if err != nil {
-		message.Fatal(err, "Unable to create the destination file")
+		return fmt.Errorf(lang.ErrUnableToCreateFile, dst)
 	}
-	defer destinationFile.Close()
```
Noxsios, line 85: note: re-add back in this defer file close

Racer159 (changes requested):

src/config/lang/english.go
```diff
@@ -378,5 +378,6 @@ const (
 
 // Collection of reusable error messages.
 var (
-	ErrInitNotFound = errors.New("this command requires a zarf-init package, but one was not found on the local system. Re-run the last command again without '--confirm' to download the package")
+	ErrInitNotFound       = errors.New("this command requires a zarf-init package, but one was not found on the local system. Re-run the last command again without '--confirm' to download the package")
+	ErrUnableToCreateFile = "unable to create file %s: %w"
```
Racer159: Extra %w, should be a constant

src/pkg/packager/create.go
```diff
@@ -310,45 +311,52 @@ func (p *Packager) addComponent(component types.ZarfComponent) (*types.Component
...
+					}
 				}
 			}
 		}
 	}
 
 	if len(component.Files) > 0 {
-		_ = utils.CreateDirectory(componentPath.Files, 0700)
+		if err = utils.CreateDirectory(componentPath.Files, 0700); err != nil {
+			return nil, fmt.Errorf("unable to create the component files directory: %w", err)
```
Racer159, line 329:
```suggestion
			return nil, fmt.Errorf("unable to create the component files directory: %s", err.Error())
```

src/pkg/packager/create.go
```diff
@@ -310,45 +311,52 @@ func (p *Packager) addComponent(component types.ZarfComponent) (*types.Component
 			}
 
 			for idx, path := range chart.ValuesFiles {
-				chartValueName := helm.StandardName(componentPath.Values, chart) + "-" + strconv.Itoa(idx)
-				if err := utils.CreatePathAndCopy(path, chartValueName); err != nil {
-					return nil, fmt.Errorf("unable to copy chart values file %s: %w", path, err)
+				dst := helm.StandardName(componentPath.Values, chart) + "-" + strconv.Itoa(idx)
+				if utils.IsURL(path) {
+					message.Debugf("Downloading %s", dst)
```
Racer159, line 316: This should probably be in utils.DownloadToFile

src/pkg/packager/create.go
```diff
@@ -358,19 +366,26 @@ func (p *Packager) addComponent(component types.ZarfComponent) (*types.Component
...
 			spinner.Updatef("Copying data injection %s for %s", data.Target.Path, data.Target.Selector)
-			destination := filepath.Join(componentPath.DataInjections, filepath.Base(data.Target.Path))
-			if err := utils.CreatePathAndCopy(data.Source, destination); err != nil {
-				return nil, fmt.Errorf("unable to copy data injection %s: %w", data.Source, err)
+			dst := filepath.Join(componentPath.DataInjections, filepath.Base(data.Target.Path))
+			if utils.IsURL(data.Source) {
+				utils.DownloadToFile(data.Source, dst, component.CosignKeyPath)
+			} else {
+				if err := utils.CreatePathAndCopy(data.Source, dst); err != nil {
+					return nil, fmt.Errorf("unable to copy data injection %s: %w", data.Source, err)
```
Racer159, line 374:
```suggestion
					return nil, fmt.Errorf("unable to copy data injection %s: %s", data.Source, err.Error())
```

src/pkg/packager/create.go
```diff
@@ -358,19 +366,26 @@ func (p *Packager) addComponent(component types.ZarfComponent) (*types.Component
...
 			pattern := regexp.MustCompile(`(?mi).+$`)
-			files, _ := utils.RecursiveFileList(destination, pattern, false, true)
+			files, _ := utils.RecursiveFileList(dst, pattern, false, true)
 			componentSBOM.Files = append(componentSBOM.Files, files...)
 		}
 	}
 
 	if len(component.Manifests) > 0 {
+		if err := utils.CreateDirectory(componentPath.Manifests, 0700); err != nil {
+			return nil, fmt.Errorf("unable to create manifest directory %s: %w", componentPath.Manifests, err)
```
Racer159, line 387:
```suggestion
			return nil, fmt.Errorf("unable to create manifest directory %s: %s", componentPath.Manifests, err.Error())
```

src/pkg/packager/create.go
```diff
@@ -382,15 +397,21 @@ func (p *Packager) addComponent(component types.ZarfComponent) (*types.Component
...
 		for _, manifest := range component.Manifests {
 			for idx, f := range manifest.Files {
 				// Copy manifests without any processing.
 				spinner.Updatef("Copying manifest %s", f)
+				if utils.IsURL(f) {
+					parsed, err := url.Parse(f)
+					if err != nil {
+						return nil, fmt.Errorf("unable to parse manifest URL %s: %w", f, err)
+					}
+					name := filepath.Base(parsed.Path)
```
Racer159, line 410: This won't necessarily be the filename... it could be nothing or a directory tree.

src/pkg/utils/network.go
```diff
@@ -70,26 +69,52 @@ func Fetch(url string) io.ReadCloser {
...
+	parsed, err := url.Parse(src)
+	if err != nil {
+		return fmt.Errorf("unable to parse the URL: %s", src)
+	}
+	// check if the parsed URL has a checksum
+	// if so, remove it and use the checksum to validate the file
+	hasChecksum := strings.Contains(parsed.Path, "@")
+	var checksum string
+	if hasChecksum {
+		index := strings.LastIndex(parsed.Path, "@")
```
Racer159, line 81: Pretty sure I could break this with a URL like:

```
https://example.com/manifest?type=yaml@348217fbca997689e9f98978ab
```

src/pkg/utils/network.go
```diff
@@ -70,26 +69,52 @@ func Fetch(url string) io.ReadCloser {
...
-		message.Fatal(err, "Unable to create the destination file")
+		return fmt.Errorf(lang.ErrUnableToCreateFile, dst)
 	}
-	defer destinationFile.Close()
 
-	// If the url start with the sget protocol use that, otherwise do a typical GET call
-	if strings.HasPrefix(url, SGETProtocol) {
-		sgetFile(url, destinationFile, cosignKeyPath)
+	// If the source url start with the sget protocol use that, otherwise do a typical GET call
+	if parsed.Scheme == "sget" {
```
Racer159, line 94: This should still reference the sget scheme as a variable

Racer159 (changes requested):

src/pkg/utils/network.go
Noxsios: now fixed in unit tests

src/pkg/utils/network.go
```diff
@@ -69,27 +67,69 @@ func Fetch(url string) io.ReadCloser {
...
+	if err != nil {
+		return err
 	}
 
 	// Create the file
-	destinationFile, err := os.Create(target)
+	file, err := os.Create(dst)
 	if err != nil {
-		message.Fatal(err, "Unable to create the destination file")
+		return fmt.Errorf("unable to create file %s", dst)
```
Racer159, line 106: This could use lang.ErrWritingFile

src/pkg/utils/network_test.go
```diff
@@ -0,0 +1,165 @@
...
+	paths []string
+}
+
+func (suite *TestNetworkSuite) SetupSuite() {
+	suite.urls.good = []string{
+		"https://zarf.dev",
+		"https://docs.zarf.dev",
+		"https://zarf.dev/docs",
+		"https://defenseunicorns.com",
+		"https://google.com",
```
Racer159, line 33: Can we add a URL with query params to the good URL list?

Racer159 (changes requested):

src/pkg/packager/create.go
```diff
@@ -385,15 +398,16 @@ func (p *Packager) addComponent(component types.ZarfComponent) (*types.Component
...
-		}
-
 		// Iterate over all manifests.
 		for _, manifest := range component.Manifests {
 			for idx, f := range manifest.Files {
 				// Copy manifests without any processing.
 				spinner.Updatef("Copying manifest %s", f)
+				if utils.IsURL(f) {
+					dst := filepath.Join(componentPath.Temp, fmt.Sprintf("manifest-%d", idx))
+					utils.DownloadToFile(f, dst, component.CosignKeyPath)
```
Racer159, line 408: errors are ignored here

src/pkg/packager/create.go
```diff
@@ -408,7 +422,13 @@ func (p *Packager) addComponent(component types.ZarfComponent) (*types.Component
 			for idx, k := range manifest.Kustomizations {
 				// Generate manifests from kustomizations and place in the package.
 				spinner.Updatef("Building kustomization for %s", k)
-				destination := fmt.Sprintf("%s/kustomization-%s-%d.yaml", componentPath.Manifests, manifest.Name, idx)
+				kname := fmt.Sprintf("kustomization-%s-%d.yaml", manifest.Name, idx)
+				destination := filepath.Join(componentPath.Manifests, kname)
+				if utils.IsURL(k) {
+					dst := filepath.Join(componentPath.Temp, kname)
+					utils.DownloadToFile(k, dst, component.CosignKeyPath)
```
Racer159, line 429: errors are ignored here

src/pkg/packager/create.go
```diff
@@ -313,45 +313,51 @@ func (p *Packager) addComponent(component types.ZarfComponent) (*types.Component
 			}
 
 			for idx, path := range chart.ValuesFiles {
-				chartValueName := helm.StandardName(componentPath.Values, chart) + "-" + strconv.Itoa(idx)
-				if err := utils.CreatePathAndCopy(path, chartValueName); err != nil {
-					return nil, fmt.Errorf("unable to copy chart values file %s: %w", path, err)
+				dst := helm.StandardName(componentPath.Values, chart) + "-" + strconv.Itoa(idx)
+				if utils.IsURL(path) {
+					utils.DownloadToFile(path, dst, component.CosignKeyPath)
```
Racer159, line 318: errors are ignored here.

src/pkg/packager/create.go
Racer159, line 408: This should just download to the correct manifest location instead of downloading to `temp` and then copying.  `temp` isn't guaranteed to exist here either and will break if not.

Racer159 (changes requested):

src/pkg/packager/create.go
```diff
@@ -385,30 +405,42 @@ func (p *Packager) addComponent(component types.ZarfComponent) (*types.Component
...
+				var destination string
 				// Copy manifests without any processing.
 				spinner.Updatef("Copying manifest %s", f)
-				// If using a temp directory, trim the temp directory from the path.
-				trimmedPath := strings.TrimPrefix(f, componentPath.Temp)
-				destination := path.Join(componentPath.Manifests, trimmedPath)
-				if err := utils.CreatePathAndCopy(f, destination); err != nil {
-					return nil, fmt.Errorf("unable to copy manifest %s: %w", f, err)
+				if utils.IsURL(f) {
+					destination = filepath.Join(componentPath.Manifests, fmt.Sprintf("manifest-%s-%d", manifest.Name, idx))
```
Racer159, line 416: Still broke 
image

src/pkg/utils/network.go
```diff
@@ -69,27 +68,74 @@ func Fetch(url string) io.ReadCloser {
...
+		}
+		if atSymbolCount == 1 && parsed.User != nil {
+			return src, checksum, nil
+		}
+
+		index := strings.LastIndex(src, "@")
+		checksum = src[index+1:]
+		src = src[:index]
+	}
+	return src, checksum, nil
```
Racer159, line 87: Also I think this will fail with a url that uses an @ but doesn't have a shasum.
i.e. `https://user:pass@hello.world?foo=bar`

Racer159 (changes requested):

src/pkg/utils/network.go
Noxsios: `parsed.User != nil` handles that

Noxsios:
working with the following `zarf.yaml` 

```yaml
kind: ZarfPackageConfig
metadata:
  name: remote-resources
  version: 0.0.1

components:
  - name: remote-manifests-and-kustomizations
    required: true
    manifests:
      - name: simple-nginx-deployment
        namespace: nginx
        files:
          - https://k8s.io/examples/application/deployment.yaml
      - name: podinfo
        namespace: podinfo
        kustomizations:
          # note this syntax
          - github.com/stefanprodan/podinfo//kustomize
    actions:
      onDeploy:
        after:
          - wait:
              cluster:
                kind: deployment
                name: nginx-deployment
                namespace: nginx
                condition: available
          - wait:
              cluster:
                kind: deployment
                name: podinfo
                namespace: podinfo
                condition: available
    images:
      - nginx:1.14.2
      - ghcr.io/stefanprodan/podinfo:6.3.5
```

src/internal/packager/kustomize/build.go
```diff
@@ -6,15 +6,45 @@ package kustomize
...
+	err = k.Unmarshal(b)
+	if err != nil {
+		return err
+	}
+	_, err = file.Write(b)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
Racer159, line 44: I don't think we need fetch... The bb extension calls the Build function with a remote url

Racer159 (changes requested):
Worked locally :hands-raised: but after the number of times this failed, I do think it would be worth having an e2e test for it.

src/internal/packager/kustomize/build.go
Noxsios: ah, well shucks. It was good learning a bit about krusty.

src/cmd/prepare.go
```diff
@@ -77,6 +78,19 @@ var prepareComputeFileSha256sum = &cobra.Command{
...
+			}
+			dst := filepath.Join(tmp, "file-to-be-hashed")
+			defer os.RemoveAll(tmp)
+			err = utils.DownloadToFile(fileName, dst, "")
+			if err != nil {
+				message.Fatalf(err, lang.ErrDownloading, fileName, err.Error())
+			}
+			fileName = dst
+		}
 		hash, err := utils.GetCryptoHash(fileName, crypto.SHA256)
```
Racer159, line 94: Doesn't GetCryptoHash already handle URLs?

examples/remote-manifests/README.md
```diff
@@ -0,0 +1,9 @@
+# remote-manifests
+
+This example shows how you to specify remote resources for a component's `manifests`.
+
+:::info
+
+To view the example source code, select the `Edit this page` link below the article and select the parent folder.
+
+:::
```
Racer159, line 9: I would like to see at least a short snippet of code on the example page (not the full zarf.yaml) but enough that people viewing it can get the gist without going to GitHub

examples/remote-manifests/README.md
```diff
@@ -0,0 +1,9 @@
+# remote-manifests
```
Racer159, line 1:
```suggestion
Remote Manifests
```
Matches our other examples better

Racer159 (changes requested):

src/cmd/prepare.go
Noxsios: derp

examples/remote-manifests/README.md
Noxsios: Also, I dont think this entire `Go to Github` flow works well at all, as even if we used kiwix it would not grab that, thereby breaking the ability to _air gap_ our docs.

Racer159: Oh yeah it is totally borked. If you make an issue I am good with that too - we really need to make examples more useful.

Racer159 (approved):

Issue #1570, opened 2023-04-06
Add Connect Action to Package actions menu


Madeline-UX:
## User Story

As Ashton, I need to be able to open a tunnel to the cluser to view an application or package resource from the UI

CLI command: zarf connect + connect command name

## Acceptance Criteria

**Given** - the user has packages deployed in their cluster
**When** - selects the connect... from the actions menu
**Then** - Open a dialog that lists the connect commands available for that package and have primary button disabled 
**When** - the user to selects one of the commands 
**Then** - make the primary button enabled
**When** -  the user clicks connect 
**Then** - Open tunnel in new tab (zarf connect + command)
**And** - Return user to previous screen

### Design Resources

**Figma File** - https://www.figma.com/file/MUItIzpzLBLuIyt225Bwgl/Zarf-UI?node-id=1347%3A57870&t=ojULNeuu0CxY2PM1-1

<img width="379" alt="Screenshot 2023-04-06 at 4 18 36 PM" src="https://user-images.githubusercontent.com/92826525/230495913-6c3a61d2-355a-4798-ac4e-af0b6cc8b44e.png">

![Dialog](https://user-images.githubusercontent.com/92826525/230495715-291de40d-eb00-4dcc-a5ea-1ee98c20c3e8.png)

PR #1638, opened 2023-04-21
#1570. Connect Action

mike-winberry:
## Description
Ashton should be able to connect to a pre-defined connection within a deployed package. 

## Related Issue
#1570 

## Type of change

## Checklist before merging


mike-winberry:
https://user-images.githubusercontent.com/86802655/235550908-f27a7f8a-5492-492f-8a88-ba9fa5128841.mov

src/ui/lib/components/DeployedPackageMenu/connect-deployed-package-dialog.svelte
```diff
@@ -0,0 +1,148 @@
...
+				color: 'var(--primary)',
+				'& .material-symbols-outlined': {
+					fontSize: '17px',
+				},
+			},
+		},
+	};
+
+	async function getPackageConnections(): Promise<
+		{ name: string; connectString: ConnectString }[]
```
Racer159, line 47: Can we make this a defined Type[] rather than an inline {...}[] here?

Racer159: (we could reuse that above as well on line 20)

src/ui/lib/components/DeployedPackageMenu/disconnect-deployed-package-dialog.svelte
```diff
@@ -0,0 +1,135 @@
...
+		} else {
+			tunnels[pkgName] = resources;
+		}
+		tunnelStore.set(tunnels);
+	}
+
+	async function disconnect(): Promise<void> {
+		try {
+			await Tunnels.disconnect(selectedConnection);
+			removeTunnel(pkg.name, selectedConnection);
```
Racer159, line 59: One thing to consider is if we want to return a map from the connect and disconnect APIs so that we don't have to have the `addTunnel` and `removeTunnel` functions in the front end.

src/ui/lib/components/DeployedPackageMenu/disconnect-deployed-package-dialog.svelte
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+</script>
+
+<ZarfDialog bind:open bind:toggleDialog {titleText} happyZarf={!failedToDisconnect}>
+	<svelte:fragment>
+		{#if !failedToDisconnect}
+			<Typography variant="body1" color="text-secondary-on-dark">
+				Select which resource you would like Zarf to disconnect from. Zarf will close and remove the
+				secure tunnel.
```
Racer159, line 90: "secure" might be misleading because while you do have to have cluster creds to create the tunnel it is not wrapped in encryption or anything.

Racer159 (changes requested):

src/internal/api/packages/list.go
```diff
@@ -28,3 +31,59 @@ func ListDeployedPackages(w http.ResponseWriter, _ *http.Request) {
...
+			connections[name] = types.ConnectString{
+				Description: svc.Annotations[config.ZarfConnectAnnotationDescription],
+				URL:         svc.Annotations[config.ZarfConnectAnnotationURL],
+			}
+		}
+
+	}
+	data.ConnectStrings = connections
+
+	common.WriteJSONResponse(w, data, http.StatusOK)
```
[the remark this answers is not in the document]
mike-winberry: Was able to remove this by adding the connection strings the pkg meta data before adding the "deployed package" secret to k8s.

src/ui/lib/components/DeployedPackageMenu/disconnect-deployed-package-dialog.svelte
mike-winberry: Removed from both the connect and disconnect dialogs.

src/types/package.go
```diff
@@ -16,19 +16,20 @@ type ZarfPackage struct {
...
+	YOLO              bool           `json:"yolo,omitempty" jsonschema:"description=Yaml OnLy Online (YOLO): True enables deploying a Zarf package without first running zarf init against the cluster. This is ideal for connected environments where you want to use existing VCS and container registries."`
+	Authors           string         `json:"authors,omitempty" jsonschema:"description=List of package authors (including contact info)"`
+	Documentation     string         `json:"documentation,omitempty" jsonschema:"description=Link to package documentation when online"`
+	Source            string         `json:"source,omitempty" jsonschema:"description=Link to package source code when online"`
+	Vendor            string         `json:"vendor,omitempty" jsonschema_description:"Name of the distributing entity, organization or individual."`
+	AggregateChecksum string         `json:"aggregateChecksum,omitempty" jsonschema:"description=Checksum of a checksums.txt file that contains checksums all the layers within the package."`
+	ConnectStrings    ConnectStrings `json:"connectStrings,omitempty" jsonschema:"description=List of connection strings for the package"`
```
Racer159, line 32: Can we add this to the `DeployedPackage` type instead? (having this here will make people think it is a thing for them to specify since this type is user-facing)

src/ui/lib/components/DeployedPackageMenu/connect-deployed-package-dialog.svelte
```diff
@@ -0,0 +1,127 @@
...
+
+<ZarfDialog bind:open bind:toggleDialog {titleText} happyZarf={!failedToConnect}>
+	<svelte:fragment>
+		{#if !failedToConnect}
+			<Typography variant="body1" color="text-secondary-on-dark">
+				Select which resource you would like Zarf to connect to. Zarf will create a tunnel and open
+				the connection in a new tab
+			</Typography>
+			<List ssx={listSSX}>
+				{#each connections as connection}
```
Racer159, line 86: There is no state for no connections available currently.

Racer159 (changes requested):

src/types/package.go
mike-winberry: Sounds good made that change!

src/ui/lib/components/DeployedPackageMenu/connect-deployed-package-dialog.svelte
mike-winberry: Screenshot 2023-05-03 at 5 09 54 PM

Racer159 (approved):
lgtm

Issue #1625, opened 2023-04-19
Help with ingress (nginx)


janluak:
Hey guys,

I can't get my head around the docs for using helm charts:

Due to not having traefik running as ingress controller pre-shipped (I think it's standard for k3s for having it included but it's not part of zart's k3s), I wanted to install nginx-ingress.

There is an [installation guide](https://kubernetes.github.io/ingress-nginx/deploy/#quick-start) presenting this helm command:
```
helm upgrade --install ingress-nginx ingress-nginx \
  --repo https://kubernetes.github.io/ingress-nginx \
  --namespace ingress-nginx --create-namespace
```

I tried tons of ways to let zarf know to download the required images and deploy the helm config afterwards but always something is missing.

I guess I am just missing out a key concept for using helm... Could you please help and let me know how to deploy a standard helm?

Thanks a lot!

---
the zarf.yaml
```
kind: ZarfPackageConfig
metadata:
  name: nginx-ingress
  description: "ingress based on nginx"

components:
  - name: nginx-ingress
    required: true
    charts:
      - name: ingress-nginx
        namespace: ingress-nginx
        version: 4.6.0
        url: https://kubernetes.github.io/ingress-nginx
    images:
      - nginx/nginx-ingress:3.1.0
```

Alternatively, I followed the full nginx-ingress guide to have a full k8s config to run the nginx-ingress. See the darf.yaml here with correlating file [ingress-controller.yaml.txt](https://github.com/defenseunicorns/zarf/files/11273991/ingress-controller.yaml.txt):
```
kind: ZarfPackageConfig
metadata:
  name: nginx-ingress
  description: "nginx-ingress"

components:
  - name: nginx-ingress
    required: true
    manifests:
      - name: nginx-ingress
        namespace: nginx-ingress
        files:
          - manifest/ingress-controller.yaml
    images:
      - nginx/nginx-ingress:3.1.0

```

Racer159:
Its not well documented but you should be able to turn back on traefik by specifying empty K3S_ARGS like this:

```
zarf init --set K3S_ARGS=""
```

You can see where that is pulled in here: https://github.com/defenseunicorns/zarf/blob/main/packages/distros/k3s/common/zarf.yaml#L8 (if it is not set it gets defaulted to `--disable traefik`)

---
That said, it looks like that is the wrong image for that chart.  This seemed to work for me:

```
# zarf.yaml
kind: ZarfPackageConfig
metadata:
  name: nginx-ingress
[...]

PR #1645, opened 2023-04-24
Discover jobs/hooks in find images and diff git repo patches

### LINKED ISSUE #1246: FindImages doesn't run any images found in Job Hooks
### Environment
Device and OS: Ubuntu 22.04
App version:  0.23.5
Kubernetes distro being used: KRE2

### Steps to reproduce
1.  Provide a helm chart to zarf.yaml (Like Flux) that contains a Job in the hook

kind: ZarfPackageConfig
metadata:
  name: test-helm-local-chart
  description: "Deploys Flux"
components:
  - name: flux
    required: true
    charts:
      - name: flux2
        localPath: charts/flux2
        namespace: flux-system
        version: 2.4.1

Because the Job is a hook, it is not included in the output Template from the Helm chart and as such is not included in the output image list.

### Expected result

Should include the image that the hook job uses that way when deployed air-gapped the container exists in that environment.

### Actual Result
Image is misssing.

Output of the above helm chart is:

- name: flux
    images:
      - ghcr.io/fluxcd/helm-controller:v0.27.0
      - ghcr.io/fluxcd/image-automation-controller:v0.27.0
      - ghcr.io/fluxcd/image-reflector-controller:v0.23.0
      - ghcr.io/fluxcd/kustomize-controller:v0.31.0
      - ghcr.io/fluxcd/source-controller:v0.32.1

### Additional Context

I think the issue is in chart.go in TemplateChart(), the templatedChart has a hook attribute but that is not included in the Manifest object.  Need to just append the Hook YAML to the existing manifests I believe and the rest of the code should work as it already does.

Racer159:
## Description

Implements hook discovery for images and shows changes from patch-git.

## Related Issue

Fixes #1246 
Relates to #1625 

## Type of change

## Checklist before merging


src/pkg/message/message.go
```diff
@@ -243,6 +244,17 @@ func Paragraphn(n int, format string, a ...any) string {
 	return pterm.DefaultParagraph.WithMaxWidth(n).Sprintf(format, a...)
 }
 
+// PrintDiff prints the differences between a and b with a as original and b as new
+func PrintDiff(textA, textB string) {
```
Noxsios, line 248: Does `diffmatchpatch` colorize the output like Git diff does?

Screenshot 2023-04-25 at 5 54 46 PM

Noxsios (approved):
lgtm, only thing I would change is colorizing the diff output, but thats a trivial change in the future.

src/pkg/packager/prepare.go
```diff
@@ -204,36 +205,38 @@ func (p *Packager) FindImages(baseDir, repoHelmChartPath string) error {
...
+			if len(validImages) > 0 {
+				componentDefinition += fmt.Sprintf("      # Possible images - %s - %s\n", p.cfg.Pkg.Metadata.Name, component.Name)
+				for _, image := range validImages {
+					componentDefinition += fmt.Sprintf("      - %s\n", image)
 				}
 			}
 		}
 	}
 
+	fmt.Println(componentDefinition)
```
Noxsios, line 238: Do we want to create a `message.Text` so that any STDERR/STDOUT redirections from pterm are used globally?

src/pkg/message/message.go
Racer159: It is not quite the same but yes it does - it does it based on character not line:
image

src/pkg/packager/prepare.go
Racer159: I think it is easy enough to use fmt.Println to print things to stdOut - I like that it is very explicit when we are printing to std out vs std err.  One thing I suppose though is that it doesn't currently print that output in the logs - though we only use this in more prepare or tools like scenarios.

PR #1662, opened 2023-04-30
Kube version override

### LINKED ISSUE #1607: helm kubeversion check fails if defined
### Environment
Device and OS: M1 Mac / MacOS
App version: Defined below
Other:

The default [helm KubeVersion](https://github.com/helm/helm/blob/912ebc1cd10d38d340f048efaf0abda047c3468e/pkg/chartutil/capabilities.go#L34) is set to `v1.20.0` by default. If a chart has `kubeVersion` defined and it is greater than `v1.20.0` then `zarf prep find-images` will fail during this validation (which is occurring during 

[example chart kubeVersion](https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/Chart.template.yaml#L6)

### Steps to reproduce
1. Use the following zarf.yaml
```
kind: ZarfPackageConfig
metadata:
  name: cert-manager
  description: Cloud native certificate management
components:
- name: cert-manager-component
  description: Cloud native certificate management
  required: true
  charts:
  - name: cert-manager
    version: v1.11.1
    namespace: cert-manager
    url: https://charts.jetstack.io/
```
2. run `zarf prep find-images`

### Expected result
Zarf finds the images it can detect
### Actual Result
```
  DEBUG   2023-04-17T10:28:41-07:00  -  error generating helm chart template: chart requires kubeVersion: >= 1.21.0-0 which is incompatible with Kubernetes v1.20.0
└ (/Users/brandtkeller/repos/defenseunicorns/zarf/src/pkg/packager/prepare.go:151)
 WARNING  Problem rendering the helm template for https://charts.jetstack.io/
```
### Additional Context

https://github.com/helm/helm/pull/11763

helm justification:
> "Any values that would normally be looked up or retrieved in-cluster will be faked locally.

In theory - this can be overwritten - but I will leave it up to the zarf team to decide what this should default to. Open to contributing with some consensus around options. 

```
	kv := chartutil.KubeVersion{
		Version: "v1.21.0",
		Major:   "1",
		Minor:   "21",
	}

	client.KubeVersion = &kv
```

[Possible placement](https://github.com/defenseunicorns/zarf/blob/v0.25.2/src/internal/packager/helm/chart.go#L165)

brandtkeller:
## Description
Adding a flag to the `zarf prepare find-images` command such that helm charts that implement a minimum `kubeVersion` field in the `Chart.yaml` can specify an override target Kubernetes version. That will allow `zarf prepare find-images` to specify a `--kube-version` that emulates the same activity that helm uses for `helm template`.

## Related Issue

Fixes #1607 

## Type of change

## Checklist before merging

### DIFF
--- a/examples/cert-manager/zarf.yaml
+++ b/examples/cert-manager/zarf.yaml
@@ -1,0 +1,13 @@
+kind: ZarfPackageConfig
+metadata:
+  name: cert-manager
+  description: Cloud native certificate management
+components:
+- name: cert-manager-component
+  description: Cloud native certificate management
+  required: true
+  charts:
+  - name: cert-manager
+    version: v1.11.1
+    namespace: cert-manager
+    url: https://charts.jetstack.io/
--- a/src/pkg/packager/prepare.go
+++ b/src/pkg/packager/prepare.go
@@ -21,6 +21,7 @@ import (
 	"github.com/defenseunicorns/zarf/src/pkg/utils"
 	"github.com/defenseunicorns/zarf/src/types"
 	"github.com/google/go-containerregistry/pkg/crane"
+	"helm.sh/helm/v3/pkg/chartutil"
 	v1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -28,7 +29,7 @@ import (
 )
 
 // FindImages iterates over a Zarf.yaml and attempts to parse any images.
-func (p *Packager) FindImages(baseDir, repoHelmChartPath string) error {
+func (p *Packager) FindImages(baseDir, repoHelmChartPath string, kubeVersionOverride string) error {
 
 	var originalDir string
 
@@ -143,9 +144,22 @@ func (p *Packager) FindImages(baseDir, repoHelmChartPath string) error {
 
 				var override string
 				var ok bool
+				var kubeVersion *chartutil.KubeVersion
 
 				if override, ok = chartNames[chart.Name]; ok {
 					chart.Name = "dummy"
+				}
+
+				if kubeVersionOverride != "" {
+					verSplit := strings.Split(kubeVersionOverride, ".")
+
+					kubeVersion = &chartutil.KubeVersion{
+						Version: kubeVersionOverride,
+						Major:   verSplit[0],
+						Minor:   verSplit[1],
+					}
+				} else {
+					kubeVersion = &chartutil.DefaultCapabilities.KubeVersion
 				}
 
 				// Generate helm templates to pass to gitops engine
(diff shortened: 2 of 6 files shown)


brandtkeller:
`example/cert-manager` is a temporary addition to allow testing with (success) and without (failure) the use of the new `--kube-version` flag

src/pkg/packager/prepare.go
```diff
@@ -143,16 +144,30 @@ func (p *Packager) FindImages(baseDir, repoHelmChartPath string) error {
...
 				}
 
+				if kubeVersionOverride != "" {
+					verSplit := strings.Split(kubeVersionOverride, ".")
+
+					kubeVersion = &chartutil.KubeVersion{
+						Version: kubeVersionOverride,
+						Major:   verSplit[0],
+						Minor:   verSplit[1],
+					}
```
tworcester, line 160: If you are relying on this being in a particular format should we guard it with a regex?

examples/cert-manager/zarf.yaml
```diff
@@ -0,0 +1,13 @@
+kind: ZarfPackageConfig
```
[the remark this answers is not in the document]
brandtkeller: I like the thought here... My current hang-up is how to ensure this doesn't break when if/when helm inevitably bumps the default versions and this stops failing by default. 

might still make for a good canary test in the event helm did ever bump the version - so I think I'll proceed and we can iterate from there. Relatively low overhead.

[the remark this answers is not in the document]
Racer159: You can write an inverse test - require has an assert that an error occurred that we use in a few places

src/test/e2e/00_use_cli_test.go
```diff
@@ -63,6 +63,15 @@ func TestUseCLI(t *testing.T) {
 	assert.NoError(t, err, stdOut, stdErr)
 	assert.Contains(t, stdOut, "nginx:1.16.0", "The chart image should be found by Zarf")
 
+	// Test `zarf prepare find-images` on a chart that has a `kubeVersion` declaration greater than the default (v1.20.0)
+	_, _, err = e2e.ExecZarfCommand("prepare", "find-images", "src/test/test-packages/00-kube-version-override")
+	assert.Error(t, err)
```
Racer159, line 68: This is where your tests are failing - the logic here makes sense but because we aren't actually returning a non-zero exit code the command itself doesn't fail.  This could be better written as a require.Contains to look for the warning message in stdErr instead.

Also I didn't realize this file was still using `assert` over `require` - `require` will fail immediately and is generally what we have been switching to because it is easier to debug than `assert` (they share the same interfaces so it should be as easy as a find and replace to fix that).

Racer159 (changes requested):

src/test/e2e/00_use_cli_test.go
brandtkeller: Good point!

PR #1664, opened 2023-05-01
Expose PVC accessMode as variable

willswire:
## Description

When attempting to upgrade our EKS Node AMIs in AWS, we noticed that the Zarf Registry deployment was unable to horizontally scale across nodes which needed to restart. We believe the culprit is the `accessMode` specification for the PersistentVolumeController. In order for multiple pods to have access to the same PersistentVolume, the `accessMode` must be set to "ReadWriteMany".

~~This PR proposes that when autoscaling is enabled for the Zarf Registry, the `accessMode` is set to "ReadWriteMany" by default; when autoscaling is disabled, it is set to "ReadWriteOnce". Due to the additional work required (i.e. using an existing PersistencVolumeController with a storage class compatible with RWX), we also propose that `autoscaling` be disabled by default.~~

This PR exposes the `REGISTRY_PVC_ACCESS_MODE` variable for the `zarf-registry` portion of the init package.

## Related Issue

- Relates to #375 

## Type of change

## Checklist before merging


packages/zarf-registry/chart/values.yaml
```diff
@@ -48,7 +48,7 @@ podDisruptionBudget:
   minAvailable: 1
 
 autoscaling:
-  enabled: true
+  enabled: false
```
Racer159, line 51:
```suggestion
  enabled: true
```

Would still like to keep autoscaling enabled.  I think the RWX change is a good one but autoscaling should be useable in EKS with the correct setup - see our eks distro we recently updated for nightly testing:

https://github.com/defenseunicorns/zarf/blob/main/packages/distros/eks/eks.yaml#L13

Note use of the EBS CSI driver will require `--disable-nodegroup-eviction` on teardown.

https://github.com/defenseunicorns/zarf/blob/main/packages/distros/eks/zarf.yaml#L69

Racer159 (changes requested):

packages/zarf-registry/chart/values.yaml
willswire: In our tests, we weren't able to bring down nodes with the Zarf Registry Deployment due to specified Pod Disruption Budgets.

willswire: For now, this PR will expose a variable for custom setting the PVC accessMode

jeff-mccoy:
We have pod affinity defined because this is node-attached storage and even a deployment rollout can break it if you don't tie the pod to the same node. HPA is no different than deployment rollout and does not require RWX.

willswire:
Reopening this PR after some `git rebase` shenanigans.

packages/zarf-registry/zarf.yaml
```diff
@@ -11,6 +11,10 @@ variables:
     description: "The size of the persistent volume claim for the registry"
     default: "20Gi"
 
+  - name: REGISTRY_PVC_ACCESS_MODE
+    description: "The access mode of the persistent volume claim for the registry"
+    default: "ReadWriteOnce"
```
Racer159, line 16: Can we add back in the switch on HPA enablement? RWX is probably what we will still use for HPA across most clusters for now so RWO should only be if HPA is disabled.

Racer159 (changes requested):

packages/zarf-registry/zarf.yaml
Racer159: Actually thinking about this again, I think this configuration option is fine since it will be more about what the storage plugin supports - but can we keep the default RWX?

Racer159: NVM I misread this when I first saw it - this is probably fine to allow for configuration

Racer159 (approved):
lgtm

willswire:
Twas missing `###` on my Zarf Var... fixed!

Racer159 (approved):

YrrepNoj (approved):

PR #1681, opened 2023-05-08
Add Big Bang YOLO mode support 

### LINKED ISSUE #1670: Big Bang YOLO mode support
Would like to have support for YOLO mode added to the Big Bang extension to allow online versions of big bang to be deployed without images or repos.

Passing the YOLO mode package metadata flag to the bb extension so that it knows not to include images.

We could have a new extension key but this would require more work for what is already a global package key.

### Additional context
This was requested from our k8s slack channel: https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1683059584720439?thread_ts=1682704735.709779&cid=C03B6BJAUJ3

willswire:
## Description

Add Big Bang YOLO mode support 

## Related Issue

Addresses #1670 

## Type of change

## Checklist before merging


src/extensions/bigbang/manifests.go
```diff
@@ -63,12 +63,27 @@ kyvernoPolicies:
...
-func manifestZarfCredentials(version string) corev1.Secret {
+const bbV2ZarfYoloModeCredentialsValues = `
+# -- Big Bang v2 Kyverno Support
+kyvernoPolicies:
+  values:
+    exclude:
+      any:
+      - resources:
+          namespaces:
+          - zarf # don't have Kyverno prevent Zarf from doing zarf things
```
Racer159, line 74: Thanks for the PR! Technically if you were running YOLO mode you shouldn't need this either since the Zarf agent would not be running in cluster.

packages/zarf-registry/registry-values.yaml
```diff
@@ -2,6 +2,7 @@ persistence:
   storageClass: "###ZARF_STORAGE_CLASS###"
   size: "###ZARF_VAR_REGISTRY_PVC_SIZE###"
   existingClaim: "###ZARF_VAR_REGISTRY_EXISTING_PVC###"
+  accessMode: "###ZARF_VAR_REGISTRY_PVC_ACCESS_MODE###"
```
Racer159, line 5: sneaky ;)

Racer159 (changes requested):

packages/zarf-registry/registry-values.yaml
willswire: removed! lol

src/extensions/bigbang/manifests.go
willswire: dropped these values in lieu of setting `registryCredentials` instead.

willswire:
using [this repo](https://github.com/willswire/yolo-big-bang) to test against compiled version with changes. so far can verify working package creation and deployment on local cluster

src/extensions/bigbang/manifests.go
```diff
@@ -63,12 +63,23 @@ kyvernoPolicies:
           - zarf # don't have Kyverno prevent Zarf from doing zarf things
           `
 
-func manifestZarfCredentials(version string) corev1.Secret {
+const bbV2ZarfYoloModeCredentialsValues = `
+registryCredentials:
+  registry: "registry1.dso.mil"
+  username: "###ZARF_VAR_REGISTRY1_USERNAME###"
+  password: "###ZARF_VAR_REGISTRY1_CLI_SECRET###"
```
Racer159, line 70: While doing this works, it requires the user to create `variables` at the package level which might not be intuitive.  This may be better done via an override of the `manifests` key for flux's secrets and an extra values.yaml for the YOLO mode.  That would give more flexibility too for different ways to bring in these secrets than these explicit ###ZARF_VARs

Racer159 (changes requested):

src/extensions/bigbang/manifests.go
Racer159: To explain that more when you specify the `bigbang` noun you can also specify `manifests` on the same component as well and they will be merged together which should work for flux.

willswire: I see where you're coming from - having variables deep within the Zarf source code feels a little bit weird to me as well. Trying your suggestion (see below) yielded similar results, where the initial deployment hangs as it is unable to send the flux `private-registry` secret beforehand. 

```yaml
...
components:
  - name: bigbang
    required: true
    manifests:
      - name: private-registry
        namespace: flux-system
        files:
          - private-registry.yaml
    extensions:
      bigbang:
        version: 2.0.0
        valuesFiles:
          - ingress.yaml
          - kyverno.yaml
          - loki.yaml
```

willswire: Something like the following seems to work:

```yaml
components:
  - name: flux
    manifests:
      - name: flux-system
        namespace: flux-system
        files:
          - private-registry.yaml
  - name: bigbang
    required: true
    extensions:
      bigbang:
        version: 2.0.0
        valuesFiles:
          - ingress.yaml
          - kyverno.yaml
          - loki.yaml
```

willswire: Removed internal use of Zarf Variables in latest commits

Racer159 (dismissed):
lgtm

willswire:
rebase w/ signed commits

Racer159:
There are still some commits in the middle that aren't signed (if needed we can open a new PR and sign - maintaining your authorship)
image

willswire:
> There are still some commits in the middle that aren't signed (if needed we can open a new PR and sign - maintaining your authorship)
> image

I force-pushed signed commits over those - they should go thru

Racer159 (approved):
approving, we'll see! (GH may be being weird with what it shows there - usually old  commits disappear)

Noxsios (approved):
lgtm

PR #1694, opened 2023-05-11
Remove more files before running goreleaser

Racer159:
## Description

This removes more files (including the new init package from the architecture test) to reduce the disk pressure before goreleaser runs.

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging


Racer159:
This is not the problem

PR #1710, opened 2023-05-15
Allow downloading the init-package

### LINKED ISSUE #1654: zarf tools init-download
You can get an init package from running `zarf init` without --confirm but for dev/test in things like CI pipelines it would be nice to have a `zarf tools init-download` that would pull down the init package for the current zarf version.

This would pull the init package from the releases page on GitHub into the current working directory - It could also look for the init package in cache first and copy it into the CWD.

We could have a flag on zarf init to force download but this would not be as friendly for moving the init package to other machines.

### Additional context
This was brought up by package creators/testers and would not be likely to be used in prod

Racer159:
## Description

Adds a command to download the init package for your version and architecture of Zarf.

## Related Issue

Fixes #1654 

## Type of change

## Checklist before merging


Makefile
```diff
@@ -25,7 +25,7 @@ else
 	endif
 endif
 
-CLI_VERSION := $(if $(shell git describe --tags),$(shell git describe --tags),"UnknownVersion")
+CLI_VERSION ?= $(if $(shell git describe --tags),$(shell git describe --tags),"UnknownVersion")
```
Noxsios, line 28: what does changing this line do?

src/cmd/initialize.go
Noxsios, line 1: lint, github won't let me comment on line 106

```go
func downloadInitPackage(initPackageName, downloadCacheTarget string) error {
```

src/config/lang/english.go
```diff
@@ -290,6 +290,10 @@ zarf init --git-push-password={PASSWORD} --git-push-username={USERNAME} --git-ur
 	CmdToolsClearCacheSuccess       = "Successfully cleared the cache from %s"
 	CmdToolsClearCacheFlagCachePath = "Specify the location of the Zarf artifact cache (images and git repositories)"
 
+	CmdToolsDownloadInitShort               = "Download the init package for the current Zarf version into the specified directory."
+	CmdToolsDownloadInitFlagOutputDirectory = "Specify a directory to place the init package in."
```
Noxsios, line 294: nit, what is our policy on which messages are sentences (end in `.`) and which are not?

Noxsios (changes requested):

Makefile
Racer159: It makes it so that the CLI_VERSION is only defaulted to that value.  If you export `CLI_VERSION` or include it as an env var just for the make command it will take that value instead - I did it because I wanted an easy way to test the init download logic and give us a way to override the value if we wanted.

src/config/lang/english.go
Racer159: I made an issue here: https://github.com/defenseunicorns/zarf/issues/1712

Noxsios (approved):
lgtm

PR #1714, opened 2023-05-16
Handle OCI Imported Components When Creating a Differential Package

### LINKED ISSUE #1426: Differential Updates for Composable Components
As a package creator using composable components, it is frustrating to rebuild / update large zarf packages that apply only small updates / incremental changes. 

I would like 'zarf package create` on a composed package to only rebuild the component packages that have changed. This would encourage smaller, more frequent updates more efficiently. Conversely, the current state of zarf may incentivize larger batch updates (which carry more risk) or decentralized zarf packages(requires documentation outside of the packages / introduces more opportunities for human error).

Planning long walks and engagements during package build pipelines, using decoupled / decentralized zarf packages, naps, or potentially looking at zarf actions to help build differential Updates smartly (introduces additional complexity and doesn't afford reusability for others).

YrrepNoj:
When creating a differential package, Zarf now considers OCI imported components as versioned assets. Since the OCI imported components are versioned assets, if Zarf is performing a differential build and the version of the OCI component we're important hasn't changed, Zarf will no longer include that OCI-Imported Component in the finalized differential Package.

Closes #1426


src/test/e2e/52_oci_compose_differential_test.go
```diff
@@ -0,0 +1,140 @@
...
+	Remote    *utils.OrasRemote
+	Reference registry.Reference
+}
+
+func (suite *OCIDifferentialSuite) SetupSuite() {
+
+	suite.Reference.Registry = "localhost:555"
+
+	// spin up a local registry
+	registryImage := "registry:2.8.1"
```
Racer159, line 42: vulnerable

examples/podinfo-flux/zarf.yaml
```diff
@@ -2,6 +2,7 @@ kind: ZarfPackageConfig
 metadata:
   name: podinfo-flux
   description: "Deploy flux and then podinfo via flux"
+  version: 0.0.1
```
Racer159, line 5: would prefer this to match podinfo's version or not be here.

src/types/component.go
```diff
@@ -216,3 +218,17 @@ type ZarfComponentImport struct {
 	// For further explanation see https://regex101.com/r/nxX8vx/1
 	URL string `json:"url,omitempty" jsonschema:"description=The URL to a Zarf package to import via OCI,pattern=^oci://(?!.*###ZARF_PKG_TMPL_).*$"`
 }
+
+// isEmpty returns if the component is empty of all content that would be deployed.
```
Racer159: lint

src/pkg/packager/create.go
```diff
@@ -625,10 +677,15 @@ func (p *Packager) removeCopiesFromDifferentialPackage() error {
 
 	// Loop through all of the components to determine if any of them are using already included images or repos
 	componentMap := make(map[int]types.ZarfComponent)
+	componentsToOCI := make(map[string]string)
+	for ociURL, componentName := range p.cfg.Pkg.Build.OCIImportedComponents {
+		componentsToOCI[componentName] = ociURL
```
Racer159, line 682: is this used?

src/test/e2e/52_oci_compose_differential_test.go
Racer159: also we could create a helper function that spins up a local registry

src/test/e2e/52_oci_compose_differential_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (suite *OCIDifferentialSuite) TearDownSuite() {
+	_, _, err := exec.Cmd("docker", "rm", "-f", "registry")
+	suite.NoError(err)
+
+	os.RemoveAll(tmpPath)
+}
+
+func (suite *OCIDifferentialSuite) Test_0_Create_Differential_OCI() {
+	suite.T().Log("E2E: Skeleton Package Publish oci://")
```
Racer159, line 78: accuratize this log

src/test/e2e/52_oci_compose_differential_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Build without differential
+	anotherPackagePath := "src/test/test-packages/oci-differential"
```
Racer159, line 81: anotherPackagePath is also above - could be global

src/types/component.go
```diff
@@ -216,3 +218,17 @@ type ZarfComponentImport struct {
...
+	listsAreEmpty := len(c.Charts) == 0 && len(c.Images) == 0 && len(c.DataInjections) == 0 && len(c.Files) == 0 && len(c.Manifests) == 0 && len(c.Repos) == 0
+	extensionsAreEmpty := reflect.ValueOf(c.Extensions).IsZero()
+	actionsAreEmpty := reflect.ValueOf(c.Actions).IsZero()
+	scriptsAreEmpty := reflect.ValueOf(c.DeprecatedScripts).IsZero()
+	importIsEmpty := ignoreImport || reflect.ValueOf(c.Import).IsZero()
+	importGroupIsEmpty := c.Group == ""
+	cosignKeyPathIsEmpty := c.CosignKeyPath == ""
+
+	return listsAreEmpty && extensionsAreEmpty && actionsAreEmpty && scriptsAreEmpty && importIsEmpty && importGroupIsEmpty && cosignKeyPathIsEmpty
+}
```
Racer159, line 234: If we can invert the logic here that would be better for the future.

Racer159 (changes requested):

docs/3-create-a-zarf-package/1-zarf-packages.md
```diff
@@ -67,6 +67,9 @@ When executed, the `zarf package create` command locates the `zarf.yaml` file in
 
 The process of defining and creating a package is also elaborated on in detail in the [Creating a Zarf Package Tutorial](../5-zarf-tutorials/0-creating-a-zarf-package.md).
 
+### Creating Differential Packages
+If you have already created a Zarf package and want to create a new package to capture updates that you have made to your package you would normally have to re-create the entire package from scratch, including all of the parts of the package that might not have changed at all. Depending on your workflow, you may only want to create a package that contains the artifacts that have changed since the last time you built your package. This can be achieved by using the `--differential` flag while running the `zarf package create` command. You can use this flag to point to an already built package you have locally or to a package that has been previously [published](../5-zarf-tutorials/7-publish-and-deploy.md#publish-package).
```
Racer159, line 71:
```suggestion
If you already have a Zarf package and you want to create an updated package you would normally have to re-create the entire package from scratch, including things that might not have changed. Depending on your workflow, you may  want to create a package that only contains the artifacts that have changed since the last time you built your package. This can be achieved by using the `--differential` flag while running the `zarf package create` command. You can use this flag to point to an already built package you have locally or to a package that has been previously [published](../5-zarf-tutorials/7-publish-and-deploy.md#publish-package) to a registry.
```
package package package package package lol

src/pkg/packager/create.go
```diff
@@ -41,8 +41,10 @@ func (p *Packager) Create(baseDir string) error {
 	}
 
 	// Load the images and repos from the 'reference' package
-	if err := p.loadDifferentialData(); err != nil {
-		return err
+	if p.cfg.CreateOpts.DifferentialData.DifferentialPackagePath != "" {
+		if err := p.loadDifferentialData(); err != nil {
+			return err
+		}
```
Racer159, line 47: I prefer to have one `nil` check (which one already exists in `loadDifferentialData`) since over time it can be confusing as to which function is responsible for that check - if this were used in multiple places and someone did not check all of them but decided to remove the check from the sub function then that could break things if they weren't careful.

src/pkg/packager/create.go
```diff
@@ -59,6 +61,13 @@ func (p *Packager) Create(baseDir string) error {
 		p.cfg.IsInitConfig = true
 	}
 
+	// Before we compose the components (and render the imported OCI components), we need to remove any components that are not needed for a differential build
+	if p.cfg.CreateOpts.DifferentialData.DifferentialPackagePath != "" {
+		if err := p.removeDifferentialComponentsFromPackage(); err != nil {
+			return err
+		}
+	}
```
Racer159, line 69: Same here - the length check in `removeDifferentialComponentsFromPackage` will have the same effect as this `nil` check

src/pkg/packager/create.go
```diff
@@ -576,6 +585,9 @@ func (p *Packager) loadDifferentialData() error {
 		return nil
 	}
 
+	// Save the fact that this is a differential build into the build data of the package
+	p.cfg.Pkg.Build.Differential = true
```
Racer159, line 589: This is being set twice and actually whatever value you set here will be overwritten based on later logic.

src/test/common.go
```diff
@@ -101,3 +104,21 @@ func (e2e *ZarfE2ETest) GetLogFileContents(t *testing.T, stdErr string) string {
 	require.NoError(t, err)
 	return string(logContents)
 }
+
+func (e2e *ZarfE2ETest) SetupDockerRegistry(t *testing.T, port int) *configfile.ConfigFile {
```
Racer159, line 108: lint

src/test/test-packages/52-oci-differential/README.md
```diff
@@ -0,0 +1,7 @@
+# OCI Differential
+
+This test package is used to test the functionality of creating a Zarf Package while using the `--differential` flag where one of the components in the package we're creating is a component that has been imported from an OCI registry.
+
+This test package demonstrate that OCI imported components will not be included during a differential package creation and that the proper build metadata will be added to the finalized package to ensure users of the package know which OCI imported components were not included.
```
Racer159, line 5:
```suggestion
This test package demonstrates that OCI imported components will not be included during differential package creation and that the proper build metadata will be added to the finalized package to ensure users of the package know which OCI imported components were not included.
```

Racer159 (changes requested):

src/pkg/packager/create.go
YrrepNoj: I am removing it from the other location it's being set.

Racer159 (approved):
lgtm

PR #1728, opened 2023-05-17
Ensure variables are uppercase 

### LINKED ISSUE #1726: intermittent issue: zarf package deploy --set value is not honored
### Environment
Device and OS: Ubuntu 22.10
App version: Zarf v0.26.3
Kubernetes distro being used: k3d v1.24.4+k3s1
Other:

### Steps to reproduce
```bash
# clone branch for podinfos
git clone --branch zarf-vars https://github.com/MxNxPx/zarf-podinfos && cd $_

# create k3d cluster and init
k3d cluster create
zarf init --confirm --components git-server

# package podinfos
zarf package create .

# loop thru deploying with different zarf vars via --set
export colors=("#ffffff" "#000000" "#9c9c9c"); while :; do for i in "${colors[@]}"; do echo; echo; date; echo "COLOR: ${i}"; zarf package deploy zarf-package-podinfos-amd64-0.0.1.tar.zst --confirm --set podinfo_too_color=${i}; sleep 5; deployed_color=$(kubectl get deploy -n podinfo-too podinfo-too -ojsonpath='{.spec.template.spec.containers[0].env[0]}' | jq -r .value); test "$i" = "$deployed_color" && echo "all good" || break 2; done; done
```

### Expected result

It would use the value passed in via --set for the helm chart and upgrade the helm release

### Actual Result

Intermittently, it does not honor the --set value provided via the zarf package deploy command

https://github.com/MxNxPx/zarf-podinfos/blob/zarf-vars/docs/README.md

### Additional Context

YrrepNoj:
If there are variables in a config.json/toml and variables provided via the `--set` flag we need to ensure both the variable keys are uppercase before processing and templating them. At the moment we were only uppercasing the config file variable keys which resulted in inconsistent behavior  because of the indeterministic ordering of maps as we were looking for the variable's value.

Fixes #1726

### DIFF
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -81,8 +81,11 @@ var packageDeployCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		pkgConfig.DeployOpts.PackagePath = choosePackage(args)
 
-		// Ensure uppercase keys from viper
+		// Ensure uppercase keys from viper and CLI --set
 		viperConfigSetVariables := utils.TransformMapKeys(v.GetStringMapString(V_PKG_DEPLOY_SET), strings.ToUpper)
+		pkgConfig.DeployOpts.SetVariables = utils.TransformMapKeys(pkgConfig.DeployOpts.SetVariables, strings.ToUpper)
+
+		// Merge the viper config file variables and provided CLI flag variables (CLI takes precedence))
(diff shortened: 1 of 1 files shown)


Noxsios (approved):
approved! @YrrepNoj w/ the fastest debugging in the west 🥇 🍪

PR #1729, opened 2023-05-18
Log Error message for failed config loads

runyontr:
## Description

When the config file fails to load correctly, we get an error message without help:

```shell
$ zarf package create .
 WARNING  failed to load config file: %!w(MISSING)

....
```

## Type of change


src/cmd/viper.go
```diff
@@ -117,7 +117,7 @@ func initViper() {
 	if err != nil {
 		// Config file not found; ignore
 		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
-			message.Error(err, lang.CmdViperErrLoadingConfigFile)
+			message.Error(err, lang.CmdViperErrLoadingConfigFile, err)
```
YrrepNoj, line 120:
```suggestion
			message.Error(err, fmt.Sprintf(lang.CmdViperErrLoadingConfigFile, err.Error()))
```

NOTE: This will require importing `fmt` in this package and modifying the `lang.CmdViperErrLoadingConfigFile` variable to be `%s` instead of `%w`

YrrepNoj (changes requested):

src/cmd/viper.go
Racer159, line 120:
```suggestion
			message.Errorf(err, lang.CmdViperErrLoadingConfigFile, err)
```
message.Errorf would be even better/simpler to us

Racer159 (changes requested):

Racer159:
This has been resolved in #1749

PR #1731, opened 2023-05-18
feat: add scheduling logic to zarf injector

### LINKED ISSUE #905: The CreatePod Function Does Not Consider Nodes that May Have Taints
It is common in some cluster configurations to taint the control plane nodes to not allow most workloads to be deployed there.  For example RKE2 HA documentations suggests adding the following taint that disallows most pods from being scheduled on control plane nodes (see <https://docs.rke2.io/install/ha/#2a-optional-consider-server-node-taints>):

```
node-taint:
  - "CriticalAddonsOnly=true:NoExecute"
```

When Zarf attempts to deploy the agent to a cluster like this, if a tainted control plane node is the first in the list the pod will fail to deploy and startup.  We've been able to work around this for our purposes by modifying the CreatePod function like so:
```
// CreatePod inserts the given pod into the cluster
func CreatePod(pod *corev1.Pod) (*corev1.Pod, error) {
	pod.Spec.Tolerations = []corev1.Toleration{
		{
			Key: "CriticalAddonsOnly",
			Operator: "Exists",
		},
	}
	message.Debugf("k8s.CreatePod(%#v)", pod)

	clientset, err := getClientset()
	if err != nil {
		return nil, err
	}

	createOptions := metav1.CreateOptions{}
	return clientset.CoreV1().Pods(pod.Namespace).Create(context.TODO(), pod, createOptions)
}
```

This, however, is a specific solution to our use case. It would be better if nodes with taints that would affect the deployment of critical Zarf components could be excluded or a configurable list of taints could be passed in to filter out the ineligible nodes.

See workaround above.

**Additional context**
There could be many different taint values used that could prevent deployment so a configurable option/list is probably best.

### LINKED ISSUE #1730: BUG: Zarf Injector can attempt to spin up on nodes that have a NoSchedule taint
### Environment
Device and OS:
App version:
Kubernetes distro being used:
Other:

### Steps to reproduce
1. Find a kubernetes cluster with master nodes that are "schedulable" but have the `node-role.kubernetes.io/master:NoSchedule` taint. I was testing with a kubeadm provisioned cluster on VMWare

### Expected result
When running `zarf init` the injector pod picks a node that it is able to be scheduled on

### Actual Result
Sometimes the zarf injector pod manifest will pick a nodeSelector on a node that has the `node-role.kubernetes.io/master:NoSchedule` taint, and the zarf injector pod does not configure a toleration for this taint. This results in the zarf injector pod being stuck in a perpetually "pending" state. 

I believe this would eventually end up working, due to the retry logic in the zarf tunnel.Create() function. However what could end up happening is that the next image that is tried could land on another node that has that same `NoSchedule` taint, causing another retry, etc etc etc. Sometimes it just works, sometimes it can take a long time!

I have submitted a merge request which fixes this issue if you'd like to use that. If not, please let me know if I could be of assistance!

ragingpastry:
## Description

Checks to ensure that a node does not have a `NoSchedule` taint before choosing that node/image for the Zarf injector pod

## Related Issue

Fixes #1730
Fixes #905

## Type of change

## Checklist before merging

Gave this a WAG. It seems to work locally, but am unsure if this is the direction you guys want to go. Let me know!

### DIFF
--- a/src/pkg/k8s/images.go
+++ b/src/pkg/k8s/images.go
@@ -19,8 +19,8 @@ type ImageMap map[string]bool
 type ImageNodeMap map[string][]string
 
 // GetAllImages returns a list of images and their nodes found in pods in the cluster.
-func (k *K8s) GetAllImages() (ImageNodeMap, error) {
-	timeout := time.After(5 * time.Minute)
+func (k *K8s) GetAllImages(timeoutDuration time.Duration) (ImageNodeMap, error) {
+	timeout := time.After(timeoutDuration)
 
 	for {
 		// Delay check 2 seconds.
@@ -44,7 +44,8 @@ func (k *K8s) GetAllImages() (ImageNodeMap, error) {
 	}
 }
 
-// GetImagesWithNodes returns all images and their nodes in a given namespace.
+// GetImagesWithNodes checks for images on schedulable nodes and returns
+// a map of these images and their nodes in a given namespace.
 func (k *K8s) GetImagesWithNodes(namespace string) (ImageNodeMap, error) {
 	result := make(ImageNodeMap)
 
@@ -53,16 +54,27 @@ func (k *K8s) GetImagesWithNodes(namespace string) (ImageNodeMap, error) {
 		return nil, fmt.Errorf("unable to get the list of pods in the cluster")
 	}
 
+	findImages:
 	for _, pod := range pods.Items {
-		node := pod.Spec.NodeName
+		nodeName := pod.Spec.NodeName
+		nodeDetails, err := k.GetNode(nodeName)
+		if err != nil {
+			return nil, fmt.Errorf("unable to get the node %s", pod.Spec.NodeName)
+		}
+
+		for _, taint := range nodeDetails.Spec.Taints {
+			if (taint.Effect == corev1.TaintEffectNoSchedule || taint.Effect == corev1.TaintEffectNoExecute) {
+				continue findImages
+			}
+		}
 		for _, container := range pod.Spec.InitContainers {
-			result[container.Image] = append(result[container.Image], node)
+			result[container.Image] = append(result[container.Image], nodeName)
 		}
 		for _, container := range pod.Spec.Containers {
-			result[container.Image] = append(result[container.Image], node)
+			result[container.Image] = append(result[container.Image], nodeName)
 		}
 		for _, container := range pod.Spec.EphemeralContainers {
-			result[container.Image] = append(result[container.Image], node)
+			result[container.Image] = append(result[container.Image], nodeName)
 		}
 	}
 
--- a/src/pkg/k8s/nodes.go
+++ b/src/pkg/k8s/nodes.go
@@ -16,3 +16,9 @@ func (k *K8s) GetNodes() (*corev1.NodeList, error) {
 	metaOptions := metav1.ListOptions{}
 	return k.Clientset.CoreV1().Nodes().List(context.TODO(), metaOptions)
 }
+
+// GetNode returns a node from the k8s cluster.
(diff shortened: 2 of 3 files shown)


src/pkg/k8s/nodes.go
```diff
@@ -16,3 +16,8 @@ func (k *K8s) GetNodes() (*corev1.NodeList, error) {
 	metaOptions := metav1.ListOptions{}
 	return k.Clientset.CoreV1().Nodes().List(context.TODO(), metaOptions)
 }
+
+// GetNode returns a node from the k8s cluster.
+func (k *K8s) GetNode(nodeName string) (*corev1.Node, error) {
+	return k.Clientset.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
+}
```
Racer159, line 23: New line at end of files is required per linting

Racer159 (changes requested):

src/pkg/k8s/images.go
```diff
@@ -53,16 +54,27 @@ func (k *K8s) GetImagesWithNodes(namespace string) (ImageNodeMap, error) {
...
 	for _, pod := range pods.Items {
-		node := pod.Spec.NodeName
+		nodeName := pod.Spec.NodeName
+		nodeDetails, err := k.GetNode(nodeName)
+		if err != nil {
+			return nil, fmt.Errorf("unable to get the node %s", pod.Spec.NodeName)
+		}
+
+		for _, taint := range nodeDetails.Spec.Taints {
+			if taint.Effect == "NoSchedule" {
```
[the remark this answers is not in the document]
Racer159: This enum would also be better to use than strings 
image

src/pkg/k8s/images.go
```diff
@@ -53,16 +54,27 @@ func (k *K8s) GetImagesWithNodes(namespace string) (ImageNodeMap, error) {
...
 	for _, pod := range pods.Items {
-		node := pod.Spec.NodeName
+		nodeName := pod.Spec.NodeName
+		nodeDetails, err := k.GetNode(nodeName)
+		if err != nil {
+			return nil, fmt.Errorf("unable to get the node %s", pod.Spec.NodeName)
+		}
+
+		for _, taint := range nodeDetails.Spec.Taints {
+			if (taint.Effect == "NoSchedule" || taint.Effect == "NoExecute") {
```
Racer159, line 66: Sorry added this in between the last review but these should be from the taint enum
image

Racer159 (changes requested):

src/pkg/k8s/images.go
ragingpastry: No worries! I saw your comment right after I pushed haha. Something like that maybe?

Racer159: Yep t hat looks correct!

Racer159 (changes requested):
I can't request it as a suggestion in this PR but one tiny thing after testing locally: Can you make the spinner for gathering images notify the user of the 5 minute timeout since they may hit that if all nodes have taints (similar to how waiting for cluster connection works)?

image
image

https://github.com/defenseunicorns/zarf/blob/main/src/internal/cluster/injector.go#L44
```go
	timeout := 5 * time.Minute
	spinner.Updatef("Getting the list of existing cluster images (%s timeout)", timeout.String())")
	if images, err = c.Kube.GetAllImages(timeout); err != nil {
```

https://github.com/defenseunicorns/zarf/blob/main/src/pkg/k8s/images.go#L22
```go
func (k *K8s) GetAllImages(timeoutDuration time.Duration) (ImageNodeMap, error) {
	timeout := time.After(timeoutDuration)
```

Racer159 (approved):
lgtm

PR #1736, opened 2023-05-19
Save/Update deployed packages state per component

### LINKED ISSUE #1732: Write the Package secret after each deployed component.
We should write the package secret after after each component deploys so that we can keep track of components throughout a deployment and more easily recover from failed deployments.

https://github.com/defenseunicorns/zarf/blob/78a3cb57cec9c49160c35f43e0d899d437dff472/src/pkg/packager/deploy.go#LL97C21-L97C21

### Additional context
We already do this on remove, but should extend this to deploy.

YrrepNoj:
Prior to this PR, we only save a `zarf-package-*` after a package has been fully deployed. This behavior led to issues when a deployment failed midway through as the users cluster would have some artifacts from the deployment but no secret to reference to see what was actually deployed via Zarf. This led to individuals not being able to execute a `zarf package remove ...` command (since there was no secret) and having to manually look through their cluster to see what was deployed and manually remove/modify them.

By updating the `zarf-package-*` secret throughout the deployment, even if a later component fails, there will still be some valuable information that a user can reference. This will also make it so users can run `zarf package remove ...` to clean up in a failed deployment easier.

Fixes #1732


src/test/test-packages/22-git-and-flux/zarf.yaml
```diff
@@ -0,0 +1,70 @@
+kind: ZarfPackageConfig
```
YrrepNoj, line 1: @Racer159 @Noxsios 

I don't think I like what I did here but I want to get your thoughts on it. I didn't want to update an `example` with logic that was strictly for testing so I decided to make a test package. 

On the other hand, I didn't want to create a new test workflow that touched the cluster since that would be added test time when it isn't really necessary...

Racer159 (approved):

PR #1748, opened 2023-05-23
Rename helmrelease manifest file for BB extension

mjnagel:
## Description

Flux has both `HelmRelease` and `HelmRepository` custom resources. Renaming this manifest for consistency with the kind of thing being created here. Since Big Bang now has an option to deploy with `HelmRepository` source this could be confusing to end users when deploying and reviewing the manifests in the zarf package.

## Related Issue

No issue, just something I noticed while deploying.

## Type of change

## Checklist before merging


Racer159 (approved):
good catch!

Issue #956, opened 2022-10-31
Better user feedback for tarball writing


Racer159:
**Is your feature request related to a problem? Please describe.**
We should show a progress bar for tarball writing like we to for image downloads.  For large tarballs (10s of GiBs) this process can take several minutes during which Zarf appears to freeze to the user (even in trace logging).

**Describe the solution you'd like**
A progress bar or a spinner during tarball writing

**Describe alternatives you've considered**
We could print a message like bytes/bytes or percentages but this would pollute the terminal window and wouldn't match the rest of the UI

**Additional context**
An example of one of these packages is the https://github.com/defenseunicorns/zarf-package-software-factory package

Racer159:
Dropping good first issue as this actually may be harder than thought for newcomers

Racer159:
Dropping good first issue as this actually may be harder than thought for newcomers

Issue #1763, opened 2023-05-31
OCI Layer Optimizations


docandrew:
### Is your feature request related to a problem? Please describe.

We're building a monolithic (>50 GB) Zarf package containing multiple components, and storing it in an OCI image registry. Using `zarf package publish` to push the repo upstream pushes every layer each time, even if layers are unchanged between pushes. Using `zarf package deploy` from the OCI with a subset of the package's components likewise results in downloading every layer, not just the layers necessary to deploy the selected components.

### Describe the solution you'd like

- Given a call to `zarf package publish`, if a layer with matching hash is already present in the upstream repo, I would like Zarf to skip uploading that layer, similar to how the Docker cli tool does with the `layer already exists` message.

- Given a call to `zarf package deploy` from an `oci://` URL, with a list of `--component` options, I'd like Zarf to pull only those layers necessary to deploy the selected components.

- Additionally, if possible `zarf package create` might be able to calculate the overall checksum of the contents for an individual component and avoid re-packaging it if it when the intended destination is an OCI repo and the component layers already exist.

### Describe alternatives you've considered

We've attempted to reduce unnecessary build & publish steps by baking in the commit hash into the Zarf package metadata and skipping a build of the package if it already exists upstream. However this only works for the package as a whole, which is a less common situation than us modifying individual components.

Noxsios:
@docandrew 

I am interested in:

> Using zarf package publish to push the repo upstream pushes every layer each time, even if layers are unchanged between pushes.

As Zarf does not push layers that have not changed.

During publish the same message is used when a layer exists, or has successfully been pushed.

```✔  46564f0eff85 images/blobs/sha256/46564f0...06008f762391a7bb7d58f339ee```

**Comment by Noxsios** [comment]
For point 3, would something like work?:

```
zarf package create <path> -o oci://repo:ref
```

This would mean that the destination is OCI, and the created package _will_ not be saved to disk.

docandrew:
@Noxsios skipping the on-disk save would be great, that might enable us to use smaller CI runners as well!

Racer159:
Closing since the majority of this issue has been addressed and bundles mitigate the last point.  We can revisit later but currently our SBOM logic makes the latter point difficult to do directly in Zarf.

Noxsios:
i concur

PR #1764, opened 2023-06-01
Some OCI Optimizations

Noxsios:
## Description

- [x] `zarf deploy oci:// --components=...` Given a call to zarf package deploy from an oci:// URL, with a list of --component options, I'd like Zarf to pull only those layers necessary to deploy the selected components.
- [x] `zarf package create . -o oci://...` calling this guy inline publising. This is essentially the same thing as doing a `zarf package create && zarf package publish`, but in the future this could also be smart about images, and only fetch images if their ref from the OCI package is different.
- [x] re-architect + restructure our OCI library code to actually be usable as a library
- [x] `zarf package remove oci://`

## Related Issue

Closes https://github.com/defenseunicorns/zarf/issues/1779
Relates to #1763 
Closes https://github.com/defenseunicorns/zarf/issues/956

## Type of change

## Checklist before merging


src/pkg/packager/common.go
```diff
@@ -474,7 +474,13 @@ func (p *Packager) validatePackageChecksums() error {
 		itemPath := strs[1]
 		expectedShasum := strs[0]
 
-		actualShasum, err := utils.GetSHA256OfFile(filepath.Join(p.tmp.Base, itemPath))
+		path := filepath.Join(p.tmp.Base, itemPath)
+		if utils.InvalidPath(path) {
+			message.Debugf("skipping checksum of %s as it does not exist", path)
+			continue
+		}
```
Racer159, line 481: Thinking about this some more I think it would be better to have some logic that checks / skips the components /images that it doesn't expect to be there locally - we already have the YAML at this point in loading so if we passed the component list to this function (or just accessed the options) we could figure this out.

That would likely make sense in a helper function too since very similar logic would be needed / exists in the `getLayersToPullFromRequestedComponents` function (and it would help make that one smaller too).

src/pkg/packager/network.go
```diff
@@ -272,3 +285,87 @@ func pullLayer(dst *utils.OrasRemote, desc ocispec.Descriptor, out string) error
...
+			layers = append(layers, filepath.Join(blobsDir, strings.TrimPrefix(manifest.Config.Digest.String(), "sha256:")))
+			for _, layer := range manifest.Layers {
+				pathInPkg := filepath.Join(blobsDir, strings.TrimPrefix(layer.Digest.String(), "sha256:"))
+				layers = append(layers, pathInPkg)
+			}
+		}
+	}
+
+	return layers, nil
+}
```
Racer159, line 371: I think this function could be refactored to be shorter with some helpers (including potentially the one mentioned above)

Racer159 (changes requested):

src/pkg/packager/network.go
Noxsios: i did a little more than that

src/pkg/utils/oras.go
```diff
@@ -92,27 +179,247 @@ func (o *OrasRemote) withAuthClient(ref registry.Reference) (*auth.Client, error
...
+		}
+	}
+	for _, component := range pkg.Components {
+		// If we requested this component, or it is required, we need to pull its images and tarball
+		if SliceContains(requestedComponents, component.Name) || component.Required {
+			images = append(images, component.Images...)
+			layers = append(layers, root.Locate(filepath.Join(zarfconfig.ZarfComponentsDir, fmt.Sprintf(tarballFormat, component.Name))))
+		}
+	}
+	if len(images) > 0 {
```
YrrepNoj, line 306: make sure the images is a unique list (unique or map)

src/pkg/utils/oras.go
```diff
@@ -92,27 +179,247 @@ func (o *OrasRemote) withAuthClient(ref registry.Reference) (*auth.Client, error
...
+
+// PullPackage pulls the package from the remote repository and saves it to the given path.
+//
+// layersToPull is an optional parameter that allows the caller to specify which layers to pull.
+//
+// The following layers will ALWAYS be pulled if they exist:
+//   - zarf.yaml
+//   - checksums.txt
+//   - zarf.yaml.sig
+func (o *OrasRemote) PullPackage(out string, concurrency int, layersToPull ...ocispec.Descriptor) error {
```
YrrepNoj, line 349: Can we make `out` more obvious that it's a directory path and not a singular output binary of the package <3

src/pkg/packager/deploy.go
```diff
@@ -100,7 +108,35 @@ func (p *Packager) Deploy() error {
...
+			for _, component := range componentsToPull {
+				tarball := filepath.Join(p.tmp.Components, fmt.Sprintf("%s.tar", component))
+				err := archiver.Unarchive(tarball, p.tmp.Components)
+				if err != nil {
+					return deployedComponents, fmt.Errorf("unable to extract the component: %w", err)
+				}
+				_ = os.Remove(tarball)
+			}
+		}
+	}
```
Racer159, line 139: Can we avoid this logic for now and do the partial stuff on `--confirm`?

src/pkg/packager/inspect.go
```diff
@@ -23,19 +24,24 @@ func (p *Packager) Inspect(includeSBOM bool, outputSBOM string, inspectPublicKey
...
 
-		message.Debugf("Pulling layers %v from %s", layersToPull, p.cfg.DeployOpts.PackagePath)
-		if err := p.pullPackageLayers(p.cfg.DeployOpts.PackagePath, p.tmp.Base, layersToPull); err != nil {
-			return fmt.Errorf("unable to pull layers for inspect: %w", err)
+		message.Debugf("Pulling layers %v from %s", requestedFiles, p.cfg.DeployOpts.PackagePath)
+
+		client, err := oci.NewOrasRemote(p.cfg.DeployOpts.PackagePath)
+		if err != nil {
+			return err
+		}
```
Racer159, line 38: common remote

src/pkg/packager/pull.go
```diff
@@ -12,15 +12,18 @@ import (
...
-	"oras.land/oras-go/v2/registry"
 )
 
 // Pull pulls a Zarf package and saves it as a compressed tarball.
 func (p *Packager) Pull() error {
-	err := p.handleOciPackage(p.cfg.DeployOpts.PackagePath, p.tmp.Base, p.cfg.PullOpts.CopyOptions.Concurrency)
+	client, err := oci.NewOrasRemote(p.cfg.PullOpts.PackageSource)
+	if err != nil {
+		return err
+	}
```
Racer159, line 25: common remote

Racer159 (changes requested):

src/pkg/oci/pull.go
```diff
@@ -0,0 +1,238 @@
...
+	estimatedBytes := int64(0)
+	if isPartialPull {
+		for _, path := range AlwaysPull {
+			desc := manifest.Locate(path)
+			layersToPull = append(layersToPull, desc)
+			estimatedBytes += desc.Size
+		}
+		for _, desc := range layersToPull {
+			estimatedBytes += desc.Size
+		}
```
Racer159, line 140: This is doubling estimated bytes size for some layers

Racer159 (changes requested):
Overall looks good, but did find one issue.

src/pkg/oci/pull.go
```diff
@@ -136,7 +136,10 @@ func (o *OrasRemote) PullPackage(destinationDir string, concurrency int, layersT
 			estimatedBytes += desc.Size
 		}
 		for _, desc := range layersToPull {
-			estimatedBytes += desc.Size
+			path := desc.Annotations[ocispec.AnnotationTitle]
+			if !utils.SliceContains(AlwaysPull, path) {
+				estimatedBytes += desc.Size
+			}
 		}
```
Racer159: estimatedBytes should just be dropped from line 136 - no need for this slice thing.

Racer159 (changes requested):

src/cmd/viper.go
```diff
@@ -69,14 +72,12 @@ const (
...
-	V_PKG_PUBLISH_OCI_CONCURRENCY      = "package.publish.oci_concurrency"
 	V_PKG_PUBLISH_SIGNING_KEY          = "package.publish.signing_key"
 	V_PKG_PUBLISH_SIGNING_KEY_PASSWORD = "package.publish.signing_key_password"
 
 	// Package pull config keys
-	V_PKG_PULL_OCI_CONCURRENCY = "package.pull.oci_concurrency"
-	V_PKG_PULL_OUTPUT_DIR      = "package.pull.output_directory"
-	V_PKG_PULL_PUBLIC_KEY      = "package.pull.public_key"
+	V_PKG_PULL_OUTPUT_DIR = "package.pull.output_directory"
+	V_PKG_PULL_PUBLIC_KEY = "package.pull.public_key"
```
Racer159, line 80: Something like this will maintain the output directory for backwards compat:
```
outputDirectory := v.GetString("package.create.output_directory")
	output := v.GetString("package.create.output")
	if outputDirectory != "" && output == "" {
		v.Set("package.create.output", outputDirectory)
	}
```

src/internal/api/components/list.go
```diff
@@ -7,12 +7,19 @@ package components
...
+	"github.com/defenseunicorns/zarf/src/internal/cluster"
+	"github.com/defenseunicorns/zarf/src/pkg/message"
+	"github.com/go-chi/chi/v5"
 )
 
 // ListDeployingComponents writes a list of packages that have been deployed to the connected cluster.
-func ListDeployingComponents(w http.ResponseWriter, _ *http.Request) {
-	deployingPackages := config.GetDeployingComponents()
-	common.WriteJSONResponse(w, deployingPackages, http.StatusOK)
+func ListDeployingComponents(w http.ResponseWriter, r *http.Request) {
```
Racer159, line 18:
```suggestion
func ListDeployedComponents(w http.ResponseWriter, r *http.Request) {
```

Deployed makes more sense than deploying with the refactor

src/pkg/oci/common.go
```diff
@@ -92,38 +155,13 @@ func (o *OrasRemote) withAuthClient(ref registry.Reference) (*auth.Client, error
...
-	if title != "" {
-		format = fmt.Sprintf("%s %s", desc.Digest.Encoded()[:12], First30last30(title))
-	} else {
-		format = fmt.Sprintf("%s [%s]", desc.Digest.Encoded()[:12], desc.MediaType)
-	}
-	message.Successf(format)
-	return nil
+	reg.PlainHTTP = zarfconfig.CommonOptions.Insecure
+	reg.Client = o.client
+	return reg.Ping(o.Context)
```
Racer159, line 166: We might want to spin off an issue for this function looking for both push and pull scopes since we don't need both of them in all cases.

src/pkg/oci/pull.go
```diff
@@ -0,0 +1,241 @@
...
+				images[image] = true
+			}
+			layers = append(layers, root.Locate(filepath.Join(config.ZarfComponentsDir, fmt.Sprintf(tarballFormat, component.Name))))
+		}
+	}
+	if len(images) > 0 {
+		// Add the image index and the oci-layout layers
+		layers = append(layers, root.Locate(root.indexPath), root.Locate(root.ociLayoutPath))
+		// Append the sbom.tar layer if it exists
+		sbomDescriptor := root.Locate(config.ZarfSBOMTar)
```
Racer159, line 82: components can also contain SBOMs (not just images).

src/pkg/oci/pull.go
```diff
@@ -0,0 +1,241 @@
...
+//
+// The following layers will ALWAYS be pulled if they exist:
+//   - zarf.yaml
+//   - checksums.txt
+//   - zarf.yaml.sig
+func (o *OrasRemote) PullPackage(destinationDir string, concurrency int, layersToPull ...ocispec.Descriptor) error {
+	isPartialPull := len(layersToPull) > 0
+	ref := o.Reference
+	message.Debugf("Pulling %s", ref.String())
+	message.Infof("Pulling Zarf package from %s", ref)
```
Racer159, line 124: newline above this per above

src/pkg/oci/pull.go
```diff
@@ -0,0 +1,241 @@
...
+			layersToPull = append(layersToPull, desc)
+			estimatedBytes += desc.Size
+		}
+		for _, desc := range layersToPull {
+			path := desc.Annotations[ocispec.AnnotationTitle]
+			if !utils.SliceContains(AlwaysPull, path) {
+				estimatedBytes += desc.Size
+			}
+		}
+	} else {
```
Racer159, line 144:
```suggestion
	if isPartialPull {
		for _, path := range AlwaysPull {
			desc := manifest.Locate(path)
			layersToPull = append(layersToPull, desc)
		}
		for _, desc := range layersToPull {
			estimatedBytes += desc.Size
		}
	} else {
```

src/pkg/oci/pull.go
```diff
@@ -0,0 +1,241 @@
...
+			if layer.Annotations[ocispec.AnnotationTitle] != "" {
+				layersToCheck = append(layersToCheck, layer.Annotations[ocispec.AnnotationTitle])
+			}
+		}
+	}
+	return utils.ValidatePackageChecksums(destinationDir, layersToCheck)
+}
+
+// PullFileLayer pulls a file layer from the remote repository and saves it to `destinationDir/annotationTitle`.
+func (o *OrasRemote) PullFileLayer(desc ocispec.Descriptor, destinationDir string) error {
```
Racer159, line 206:
```suggestion
func (o *OrasRemote) PullLayer(desc ocispec.Descriptor, destinationDir string) error {
```

This could probably be PullLayer to be cleaner since File has a different meaning in Zarf.

src/pkg/oci/pull.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+	}
+	var pkg types.ZarfPackage
+	err = utils.ReadYaml(filepath.Join(destinationDir, config.ZarfYAML), &pkg)
+	if err != nil {
+		return err
+	}
+	if pkg.Metadata.AggregateChecksum != "" {
+		return utils.ValidatePackageChecksums(destinationDir, AlwaysPull)
+	}
```
Racer159, line 239: Noticed you wrap this with this empty string check in some places but don't in others.

src/pkg/packager/common.go
```diff
@@ -235,9 +235,13 @@ func createPaths() (paths types.TempPaths, err error) {
 	return paths, err
 }
 
-func getRequestedComponentList(requestedComponents string) []string {
+func (p *Packager) getRequestedComponentList(requestedComponents string) []string {
```
Racer159, line 238: This doesn't need to be a receiver

src/pkg/packager/common.go
```diff
@@ -249,6 +253,8 @@ func (p *Packager) loadZarfPkg() error {
 		return fmt.Errorf("unable to handle the provided package path: %w", err)
 	}
 
+	wasPulled := p.cfg.DeployOpts.PackagePath == p.tmp.Base
```
Racer159, line 256:
```suggestion
	extractedToTmp := p.cfg.DeployOpts.PackagePath == p.tmp.Base
```

I think this could have a better name to explain why you are doing this.

src/pkg/utils/checksum.go
```diff
@@ -0,0 +1,159 @@
...
+		actualSHA, err := GetSHA256OfFile(path)
+		if err != nil {
+			return fmt.Errorf("unable to get checksum of: %s", err.Error())
+		}
+
+		if sha != actualSHA {
+			return fmt.Errorf("invalid checksum for %s: (expected: %s, received: %s)", path, sha, actualSHA)
+		}
+
+		pathCheckMap[path] = true
```
Racer159, line 100:
```suggestion
		checkedMap[path] = true
```

CheckedMap may be more distinct from pathsToCheck (and implies better that they have already been checked)

Racer159 (changes requested):

src/pkg/oci/pull.go
Noxsios: eh? idk what you mean here.

Noxsios: ah you mean a component could have `file` SBOMs but no image SBOMs

src/internal/api/components/list.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 // ListDeployingComponents writes a list of packages that have been deployed to the connected cluster.
-func ListDeployingComponents(w http.ResponseWriter, r *http.Request) {
+func ListDeployedComponents(w http.ResponseWriter, r *http.Request) {
```
Racer159, line 18: lint

src/pkg/oci/pull.go
```diff
@@ -203,7 +203,7 @@ func (o *OrasRemote) PullPackage(destinationDir string, concurrency int, layersT
 }
 
 // PullFileLayer pulls a file layer from the remote repository and saves it to `destinationDir/annotationTitle`.
-func (o *OrasRemote) PullFileLayer(desc ocispec.Descriptor, destinationDir string) error {
+func (o *OrasRemote) PullLayer(desc ocispec.Descriptor, destinationDir string) error {
```
Racer159, line 206: lint

Racer159 (changes requested):
lint

src/pkg/oci/common.go
Racer159: https://github.com/defenseunicorns/zarf/issues/1821

PR #1770, opened 2023-06-02
Refactors and library additions to support `bundle`s

Noxsios:
## Description

This PR performs some refactoring in `packager`, `oci` and a few other places to allow for orchestration of Zarf packages in `bundles`.  A POC of this bundle effort is/was contained in this PR, but will/has been migrated to a new CLI, and as such been removed from this PR. The final state of this PR will allow for easier consumption of certain functionality in `packager`, more powerful `oci` library features, and is one of the first true tests of Zarf as a library. 2 months well spent.

Closes https://github.com/defenseunicorns/zarf/issues/1908

## Type of change

## Checklist before merging


docs/gen-cli-docs.sh
Noxsios, line 1: TODO: temporary for now, will revert before I fully merge

src/cmd/bundle.go
```diff
@@ -0,0 +1,171 @@
...
+var bundleInspectCmd = &cobra.Command{
+	Use:     "inspect [BUNDLE]",
+	Aliases: []string{"i"},
+	Short:   lang.CmdBundleInspectShort,
+	Args:    cobra.ExactArgs(1),
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		if !utils.IsOCIURL(args[0]) && !bundler.IsValidTarballPath(args[0]) {
+			return fmt.Errorf("first argument must either be a valid OCI URL or a valid path to a bundle tarball")
+		}
+		return oci.ValidateReference(args[0])
```
Racer159, line 85: This should probably be a helper function since it exists in a few spots. `PreRunE` can be set to a named function as well.

src/cmd/bundle.go
```diff
@@ -0,0 +1,171 @@
...
+	},
+}
+
+var bundleRemoveCmd = &cobra.Command{
+	Use:     "remove [BUNDLE_NAME|BUNDLE_TARBALL|OCI_REF]",
+	Aliases: []string{"u"},
+	Args:    cobra.ExactArgs(1),
+	Short:   lang.CmdBundleRemoveShort,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if !utils.IsOCIURL(args[0]) && utils.InvalidPath(args[0]) {
```
Racer159, line 105: (case in point to the above, not sure why this line uses a different InvalidPath fuction)

src/pkg/bundler/common.go
```diff
@@ -0,0 +1,184 @@
...
+		message.Fatalf(err, "bundler unable to setup, bad config: %s", err.Error())
+	}
+	return bundler
+}
+
+// ClearPaths clears out the paths used by Bundler
+func (b *Bundler) ClearPaths() {
+	b.FS.ClearPaths()
+}
+
```
Racer159, line 73: This function seems uneeded - FS is exposed and is the only operation.

src/pkg/bundler/common.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	b.bundle.Build.Terminal = hostname
+
+	// TODO: investigate the best way forward for determining arch
+	b.bundle.Metadata.Architecture = runtime.GOARCH
+	b.bundle.Build.Architecture = runtime.GOARCH
+
+	b.bundle.Build.Timestamp = now.Format(time.RFC1123Z)
+
+	b.bundle.Build.Version = config.CLIVersion
```
Racer159, line 153: since this reuses a lot of the logic from packager common it might be good to have the common bits in a separate struct / utility function.

src/pkg/bundler/fs.go
```diff
@@ -0,0 +1,89 @@
...
+// CD is a wrapper around os.Chdir
+func (bfs *BFS) CD(path string) error {
+	message.Debugf("bfs.CD - %s", path)
+	return os.Chdir(path)
+}
+
+// ReadBundleYaml is a wrapper around utils.ReadYaml
+func (bfs *BFS) ReadBundleYaml(path string, bndl *types.ZarfBundle) error {
+	return utils.ReadYaml(path, bndl)
+}
```
Racer159, line 65: These wrappers seem uneeded.

Racer159 (changes requested):
a few small notes

src/pkg/bundler/common.go
```diff
@@ -0,0 +1,309 @@
...
+			}
+			if pkg.OptionalComponents[0] == "*" {
+				// a wildcard has been given, so all optional components will be included
+				for _, c := range zarfYAML.Components {
+					if c.Only.Cluster.Architecture == "" || c.Only.Cluster.Architecture == b.bundle.Metadata.Architecture {
+						pkg.OptionalComponents = append(pkg.OptionalComponents, c.Name)
+					}
+				}
+				continue
+			}
```
Noxsios, line 163: must also expand `aws-*` wildcard syntax here. Will fail if NO components satisfy arch. Continue if at least 1 satisfies and warn about the others?

src/pkg/oci/pull.go
```diff
@@ -136,9 +140,17 @@ func (o *OrasRemote) PullPackage(destinationDir string, concurrency int, layersT
...
+			exists := false
+			for _, layer := range layersToPull {
+				if layer.Annotations[ocispec.AnnotationTitle] == path {
+					exists = true
+				}
+			}
+			if !exists {
+				desc := manifest.Locate(path)
+				layersToPull = append(layersToPull, desc)
+			}
```
Noxsios, line 153: this needs to use similar SHA logic as implemented later. As it stands you could have duplication of paths, and the wrong size is used, resulting in a wrong progress bar.

src/pkg/bundler/common.go
```diff
@@ -0,0 +1,325 @@
...
+	if len(b.bundle.Packages) == 0 {
+		return fmt.Errorf("zarf-bundle.yaml is missing required list: packages")
+	}
+	// validate access to packages as well as components referenced in the package
+	for idx, pkg := range b.bundle.Packages {
+		url := fmt.Sprintf("%s:%s-%s", pkg.Repository, pkg.Ref, b.bundle.Metadata.Architecture)
+
+		if strings.Contains(pkg.Ref, "@sha256:") {
+			url = fmt.Sprintf("%s:%s", pkg.Repository, pkg.Ref)
+		}
```
Noxsios, line 121: need to double check this logic

src/pkg/packager/common.go
```diff
@@ -498,15 +530,16 @@ var (
...
-func (p *Packager) validatePackageSignature(publicKeyPath string) error {
+// ValidatePackageSignature validates the signature of a package
+func ValidatePackageSignature(directory string, publicKeyPath string) error {
 
-	// If the insecure flag was provided, or there is no aggregate checksum, ignore the signature validation
-	if config.CommonOptions.Insecure || p.cfg.Pkg.Metadata.AggregateChecksum == "" {
+	// If the insecure flag was provided, ignore the signature validation
+	if config.CommonOptions.Insecure {
 		return nil
 	}
```
Noxsios, line 539: need a consensus on this

src/cmd/package.go
```diff
@@ -37,7 +37,7 @@ var packageCmd = &cobra.Command{
 var packageCreateCmd = &cobra.Command{
 	Use:     "create [ DIRECTORY ]",
 	Aliases: []string{"c"},
-	Args:    cobra.MaximumNArgs(1),
+	Args:    cobra.ExactArgs(1),
```
Racer159, line 40:
```suggestion
	Args:    cobra.MaximumNArgs(1),
```


`zarf package create` is a thing (with no args)

src/cmd/package.go
```diff
@@ -75,18 +75,18 @@ var packageDeployCmd = &cobra.Command{
 	Aliases: []string{"d"},
 	Short:   lang.CmdPackageDeployShort,
 	Long:    lang.CmdPackageDeployLong,
-	Args:    cobra.MaximumNArgs(1),
+	Args:    cobra.ExactArgs(1),
```
Racer159, line 78:
```suggestion
	Args:    cobra.MaximumNArgs(1),
```

`zarf package deploy` is a thing (with no args)

src/cmd/package.go
```diff
@@ -104,9 +104,9 @@ var packageInspectCmd = &cobra.Command{
 	Aliases: []string{"i"},
 	Short:   lang.CmdPackageInspectShort,
 	Long:    lang.CmdPackageInspectLong,
-	Args:    cobra.MaximumNArgs(1),
+	Args:    cobra.ExactArgs(1),
```
Racer159, line 107:
```suggestion
	Args:    cobra.MaximumNArgs(1),
```

`zarf package inspect` is a thing (with no args)

src/cmd/prepare.go
```diff
@@ -90,7 +90,7 @@ var prepareComputeFileSha256sum = &cobra.Command{
 var prepareFindImages = &cobra.Command{
 	Use:     "find-images [ PACKAGE ]",
 	Aliases: []string{"f"},
-	Args:    cobra.MaximumNArgs(1),
+	Args:    cobra.ExactArgs(1),
```
Racer159, line 93:
```suggestion
	Args:    cobra.MaximumNArgs(1),
```

`zarf prepare find-images` is a thing (no args)

src/cmd/root.go
```diff
@@ -48,7 +48,7 @@ var rootCmd = &cobra.Command{
 	},
 	Short: lang.RootCmdShort,
 	Long:  lang.RootCmdLong,
-	Args:  cobra.MaximumNArgs(1),
+	Args:  cobra.ExactArgs(1),
```
Racer159, line 51:
```suggestion
	Args:  cobra.MaximumNArgs(1),
```

`zarf` is a thing - it prints the logo and help

src/cmd/version.go
```diff
@@ -52,11 +52,7 @@ var versionCmd = &cobra.Command{
 		buildMap["platform"] = runtime.GOOS + "/" + runtime.GOARCH
 		buildMap["goVersion"] = runtime.Version()
 		ver, err := semver.NewVersion(config.CLIVersion)
-		if err != nil {
-			buildMap["minor"] = ""
-			buildMap["patch"] = ""
-			buildMap["prerelease"] = ""
-		} else {
+		if err == nil {
```
Racer159, line 55:
```suggestion
    validSemver := err == nil
		if validSemver {
```


This is a pattern that is used in `src/internal/api/cluster.go` and I would advocate we look to switch these err == nils to reduce confusion since the more common pattern is `err != nil`

Racer159: (we could also decide on another pattern just think it should be visually distinct)

src/config/config.go
```diff
@@ -59,15 +50,30 @@ const (
...
+
+	ZarfRegistryPushUser                   string = "zarf-push"
+	ZarfRegistryPullUser                   string = "zarf-pull"
+	ZarfInClusterContainerRegistryNodePort int    = 31999
+
+	ZarfGitPushUser string = "zarf-git-user"
+	ZarfGitReadUser string = "zarf-git-read-user"
+
+	ZarfInClusterGitServiceURL      string = "http://zarf-gitea-http.zarf.svc.cluster.local:3000"
+	ZarfInClusterArtifactServiceURL string = ZarfInClusterGitServiceURL + "/api/packages/" + ZarfGitPushUser
```
Racer159, line 74:
```suggestion
	IPV4Localhost  = "127.0.0.1"

	ZarfArtifactTokenName  = "zarf-artifact-registry-token"

	ZarfImagePullSecretName  = "private-registry"
	ZarfGitServerSecretName  = "private-git-server"

	ZarfRegistryPushUser                    = "zarf-push"
	ZarfRegistryPullUser                    = "zarf-pull"
	ZarfInClusterContainerRegistryNodePort     = 31999

	ZarfGitPushUser  = "zarf-git-user"
	ZarfGitReadUser  = "zarf-git-read-user"

	ZarfInClusterGitServiceURL       = "http://zarf-gitea-http.zarf.svc.cluster.local:3000"
	ZarfInClusterArtifactServiceURL  = ZarfInClusterGitServiceURL + "/api/packages/" + ZarfGitPushUser
```

The types are a bit redundant

src/internal/cluster/injector.go
```diff
@@ -44,11 +42,9 @@ func (c *Cluster) StartInjectionMadness(tempPath types.TempPaths, injectorSeedTa
 	// Get all the images from the cluster
 	timeout := 5 * time.Minute
 	spinner.Updatef("Getting the list of existing cluster images (%s timeout)", timeout.String())
-	if images, err = c.Kube.GetAllImages(timeout); err != nil {
+	if images, err = c.GetAllImages(timeout); err != nil {
 		spinner.Fatalf(err, "Unable to generate a list of candidate images to perform the registry injection")
 	}
-	message.Debugf("Found %d images in the cluster", len(images))
-	message.Debugf("Images: %#v", images)
```
Racer159, line 51: I would argue these are useful debug statements - I have used them in the past.

src/pkg/interactive/interactive.go
Racer159, line 1: This should probably be named `prompt.go` to avoid a double interactive.

src/pkg/k8s/pods.go
```diff
@@ -142,7 +141,6 @@ func (k *K8s) WaitForPodsAndContainers(target PodLookup, include PodFilter) []co
 
 			}
 
-			k.Log("Ready pods %#v", readyPods)
```
Racer159, line 145: I would argue this is useful since it could help show a stuck cluster.

src/pkg/oci/common.go
```diff
@@ -63,6 +66,10 @@ func NewOrasRemote(url string) (*OrasRemote, error) {
 
 // WithRepository sets the repository for the remote as well as the auth client.
 func (o *OrasRemote) WithRepository(ref registry.Reference) error {
+	if o.root != nil {
```
Racer159, line 69: I feel like it is a weird pattern to have a non-reusable item on an otherwise reusable struct.

Racer159 (changes requested):
partial review

src/internal/api/packages/deploy.go
```diff
@@ -24,8 +24,7 @@ import (
 
 // DeployPackage deploys a package to the Zarf cluster.
 func DeployPackage(w http.ResponseWriter, r *http.Request) {
```
cmwylie19, line 26:
```suggestion
// Deploy deploys a package to the Zarf cluster.
func Deploy(w http.ResponseWriter, r *http.Request) {
```

Since this is the packages package should we removed Packages here? and update [‎src/internal/api/start.go](https://github.com/defenseunicorns/zarf/blob/f2fc5513580c67dc441b69b06c283194466840d2/src/internal/api/start.go#L91) I know it is different than package but just wondering your opinion

src/config/config.go
Noxsios: the types of these things are not shown in VSCode unless you put them there

src/pkg/oci/common.go
Noxsios: Since a lot of operations call `FetchRoot`, I wanted to cache this request so that future calls do not need to perform an extraneous fetch. Could just always set `o.root = nil` here.

cmwylie19 (commented):
This looks clean

src/config/config.go
Racer159: It shows them for me...
image


If we want to have them we should have them everywhere to stay consistent.

Noxsios: im fine w/ that

src/pkg/packager/common.go
Racer159: We should not remove `|| p.cfg.Pkg.Metadata.AggregateChecksum == ""` as it will cause backwards compatibility for older packages to fail.

src/pkg/oci/pull.go
```diff
@@ -216,20 +224,31 @@ func (o *OrasRemote) PullLayer(desc ocispec.Descriptor, destinationDir string) e
 	return utils.WriteFile(filepath.Join(destinationDir, desc.Annotations[ocispec.AnnotationTitle]), b)
 }
 
-// PullPackageMetadata pulls the package metadata from the remote repository and saves it to `destinationDir`.
-func (o *OrasRemote) PullPackageMetadata(destinationDir string) (err error) {
+// PullMultipleFiles pulls multiple files from the remote repository and saves them to `destinationDir`.
+func (o *OrasRemote) PullMultipleFiles(paths []string, destinationDir string) ([]ocispec.Descriptor, error) {
```
Racer159, line 228:
```suggestion
func (o *OrasRemote) PullPackagePaths(paths []string, destinationDir string) ([]ocispec.Descriptor, error) {
```
PullPackagePaths might be a better way of naming this

src/pkg/oci/utils.go
```diff
@@ -26,34 +28,36 @@ import (
...
+	message.Debug("Raw OCI reference from metadata:", raw)
 
 	ref, err := registry.ParseReference(raw)
 	if err != nil {
-		return nil, err
+		return "", err
 	}
 
-	return &ref, nil
+	return ref.String(), nil
```
Racer159, line 53: why not return the full reference if it is this easy to get a string out?

src/pkg/oci/utils.go
```diff
@@ -93,33 +99,46 @@ func (o *OrasRemote) FetchLayer(desc ocispec.Descriptor) (bytes []byte, err erro
...
+	return FetchYAML[types.ZarfPackage](o.FetchLayer, manifest, config.ZarfYAML)
+}
+
+// FetchImagesIndex fetches the images/index.json file from the remote repository.
+func (o *OrasRemote) FetchImagesIndex(manifest *ZarfOCIManifest) (index *ocispec.Index, err error) {
+	return FetchJSON[*ocispec.Index](o.FetchLayer, manifest, manifest.indexPath)
+}
+
+// FetchJSON fetches the given JSON file from the remote repository.
+func FetchJSON[T any](fetcher func(desc ocispec.Descriptor) (bytes []byte, err error), manifest *ZarfOCIManifest, path string) (result T, err error) {
```
Racer159, line 111:
```go
// FetchJSON fetches the given JSON file from the remote repository.
func FetchJSON[T any](fetcher func(desc ocispec.Descriptor) (bytes []byte, err error), manifest *ZarfOCIManifest, path string) (result T, err error) {
	return FetchUnmarshal[T](fetcher, json.Unmarshal, manifest, path)
}

// FetchYAML fetches the given JSON file from the remote repository.
func FetchYAML[T any](fetcher func(desc ocispec.Descriptor) (bytes []byte, err error), manifest *ZarfOCIManifest, path string) (result T, err error) {
	return FetchUnmarshal[T](fetcher, goyaml.Unmarshal, manifest, path)
}

// FetchUnmarshal fetches the given descriptor from the remote repository and unmarshals it.
func FetchUnmarshal[T any](fetcher func(desc ocispec.Descriptor) (bytes []byte, err error), unmarshaler func(data []byte, v interface{}) error, manifest *ZarfOCIManifest, path string) (result T, err error) {
	descriptor := manifest.Locate(path)
	if IsEmptyDescriptor(descriptor) {
		return result, fmt.Errorf("unable to find %s in the manifest", path)
	}
	bytes, err := fetcher(descriptor)
	if err != nil {
		return result, err
	}
	err = unmarshaler(bytes, &result)
	if err != nil {
		return result, err
	}
	return result, nil
}
```
If you wanna be real cool

src/pkg/packager/common.go
```diff
@@ -75,8 +77,18 @@ func New(cfg *types.PackagerConfig) (*Packager, error) {
 		}
 	)
 
+	if config.CommonOptions.TempDirectory != "" {
+		if err := utils.CreateDirectory(config.CommonOptions.TempDirectory, 0700); err != nil {
+			return nil, fmt.Errorf("unable to create temp directory: %w", err)
+		}
```
Racer159, line 83: I think I would rather do the actual create of the directory in `createPaths`

src/pkg/packager/common.go
```diff
@@ -264,14 +286,24 @@ func (p *Packager) loadZarfPkg() error {
...
+
+	if alreadyExtracted && pathsToCheck == nil {
+		paths, err := utils.RecursiveFileList(p.tmp.Base, nil, false)
+		if err != nil {
+			return fmt.Errorf("unable to get a list of files in the package: %w", err)
+		}
+		for _, path := range paths {
+			pathsToCheck = append(pathsToCheck, strings.TrimPrefix(path, p.tmp.Base+string(os.PathSeparator)))
+		}
+	}
```
Racer159, line 299: I would prefer if we didn't have this shim here since it is janky - assuming this is because library code could do a partial pull into a directory and Zarf may not know about that?

Racer159: If so it might be fine (for now) but we should remove it quickly

src/pkg/packager/deploy.go
```diff
@@ -31,18 +31,14 @@ import (
 )
 
 var (
-	stateInitialized bool
-	hpaModified      bool
-	valueTemplate    *template.Values
-	connectStrings   = make(types.ConnectStrings)
+	hpaModified    bool
+	connectStrings = make(types.ConnectStrings)
```
Racer159, line 35: Both of these also are likely not best done here and should likely be put on packager (even though they are less important than state and values they could cause bugs with reuse)

Racer159: (for example hpa modified is never set to false but it is reverted once a package deployment is completed)

src/pkg/packager/deploy.go
```diff
@@ -113,12 +109,12 @@ func (p *Packager) Deploy() error {
...
-	if valueTemplate, err = template.Generate(p.cfg); err != nil {
-		return deployedComponents, fmt.Errorf("unable to generate the value template: %w", err)
-	}
-
 	for _, component := range componentsToDeploy {
+		// Generate a value template
+		if p.valueTemplate, err = template.Generate(p.cfg); err != nil {
+			return deployedComponents, fmt.Errorf("unable to generate the value template: %w", err)
+		}
+
```
Racer159, line 117: Why do this for every component?

src/test/e2e/00_use_cli_test.go
```diff
@@ -119,12 +119,12 @@ func TestUseCLI(t *testing.T) {
 	t.Run("zarf package create with tmpdir and cache", func(t *testing.T) {
 		t.Parallel()
 		tmpdir := t.TempDir()
-		cachePath := filepath.Join(tmpdir, ".cache-location")
-		stdOut, stdErr, err := e2e.Zarf("package", "create", "examples/dos-games", "--zarf-cache", cachePath, "--tmpdir", tmpdir, "--log-level=debug", "--confirm")
+		cachedir := filepath.Join(t.TempDir(), ".cache-location")
```
Racer159, line 122:
```suggestion
		cacheDir := filepath.Join(t.TempDir(), ".cache-location")
```

Racer159 (changes requested):
More changes

src/pkg/oci/copier.go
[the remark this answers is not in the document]
Noxsios: The primary function of this code is is allows for the mirroring / copying all or some layers of one Zarf package in OCI to be transferred into another. While the UDS CLI is the primary and sole consumer at this time, it is still better suited in the `oci` library of Zarf IMO as that is easier to consume for the community. I can write some unit tests for this before this PR is merged.

src/pkg/oci/utils.go
Noxsios: This function is only really used to feed into `OrasRemote`, which accepts a string parameter, and not for true full manipulation of the ref.

src/pkg/oci/utils.go
Noxsios: I wasn't sure if this level of abstraction was worth doing, sacrificing readability even more than what exists. I will think about it.

src/pkg/packager/common.go
Noxsios: Correct, as `packager` only accepts a directory as a source, not a collection of file paths.

src/pkg/packager/common.go
Noxsios: I agree, but this is something we should mark for removal before GA

src/pkg/k8s/pods.go
Noxsios: What specific information is helpful from this debug message? I would rather debug only the pertinent info (like `name:ns:uuid:image` etc...

src/internal/cluster/injector.go
Noxsios: What specific information is helpful from this debug message? I would rather debug only the pertinent info (like `name:ns:uuid:image` etc...

src/pkg/packager/common.go
Racer159: I'm good with that - created an issue to track and add to here: https://github.com/defenseunicorns/zarf/issues/1956

src/internal/cluster/injector.go
Racer159: Actually looks like the same info is logged in the spinner in the `range images` function - knowing the images is useful though to figure out which ones it found / is trying to use since some images will not work such as those on nodes with taints.

src/pkg/k8s/pods.go
Racer159: Knowing the status of the pods is useful when figuring out what Zarf saw as it looked at them - if readyPods is empty or doesn't contain the pods we expect then we know that the pod was likely pending or in some bad state and whatever failure we say isn't less likely due to bad logic.

Racer159: We could be more verbose about what we are seeing by potentially logging more status info in the container section (since we check phase in the else block)

src/cmd/root.go
```diff
@@ -70,6 +70,11 @@ func Execute() {
 	cobra.CheckErr(rootCmd.Execute())
 }
 
+// RootCmd returns the root command.
+func RootCmd() *cobra.Command {
+	return rootCmd
+}
+
```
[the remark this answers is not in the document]
Noxsios: This allows for some interesting stuff like: https://github.com/Noxsios/docs-pdf-zarf/blob/778f7374866f7e79b0e53d226282b3f34a488328/main.go#L38C26-L38C26

src/pkg/packager/common.go
Noxsios: refactor noted in https://github.com/defenseunicorns/zarf/issues/1943#issuecomment-1676007461

src/pkg/packager/actions.go
```diff
@@ -200,6 +200,16 @@ func actionCmdMutation(cmd string, shellPref types.ZarfComponentActionShell) (st
 		return cmd, err
 	}
 
+	// if another binary is calling Zarf, use the system Zarf
+	// See https://regex101.com/r/O4Ngia/1
+	regex, err := regexp.Compile(`^.*zarf(\.exe)?$`)
```
Racer159, line 205: This will not work for `zarf_v0.28.4_Darwin_all ` - we should probably do something where we try to get package information from the binary.

src/pkg/packager/common.go
```diff
@@ -75,8 +79,15 @@ func New(cfg *types.PackagerConfig) (*Packager, error) {
...
+	if config.CommonOptions.TempDirectory != "" {
+		// If the cache directory is within the temp directory, warn the user
+		if strings.HasPrefix(config.CommonOptions.CachePath, config.CommonOptions.TempDirectory) {
+			message.Warnf("The cache directory (%q) is within the temp directory (%q) and will be removed when the temp directory is cleaned up", config.CommonOptions.CachePath, config.CommonOptions.TempDirectory)
+		}
+	}
+
 	// Create a temp directory for the package
-	if pkgConfig.tmp, err = createPaths(); err != nil {
+	if pkgConfig.tmp, err = createPaths(config.CommonOptions.TempDirectory); err != nil {
```
Racer159, line 90:
```suggestion
	if err = p.SetTempDirectory(config.CommonOptions.TempDirectory); err != nil {
```

src/pkg/packager/common.go
```diff
@@ -225,10 +242,18 @@ func isValidFileExtension(filename string) bool {
...
+		basePath, err = utils.MakeTempDir()
+		if err != nil {
+			return paths, err
+		}
+	} else {
+		if err := utils.CreateDirectory(basePath, 0700); err != nil {
+			return paths, fmt.Errorf("unable to create temp directory: %w", err)
+		}
+	}
+	message.Debug("Using temp", basePath)
```
Racer159, line 256:
```suggestion
	message.Debug("Using temporary directory:", basePath)
```

src/pkg/packager/common.go
Racer159: fine

src/pkg/packager/common.go
```diff
@@ -500,15 +535,19 @@ var (
...
-
+// ValidatePackageSignature validates the signature of a package
+func ValidatePackageSignature(directory string, publicKeyPath string) error {
+	var pkg types.ZarfPackage
+	if err := utils.ReadYaml(filepath.Join(directory, config.ZarfYAML), &pkg); err != nil {
+		return fmt.Errorf("unable to read the package config: %w", err)
+	}
 	// If the insecure flag was provided, or there is no aggregate checksum, ignore the signature validation
-	if config.CommonOptions.Insecure || p.cfg.Pkg.Metadata.AggregateChecksum == "" {
+	if config.CommonOptions.Insecure || pkg.Metadata.AggregateChecksum == "" {
```
Racer159, line 545:
```suggestion
	if config.CommonOptions.Insecure {
```

Racer159: This is a breaking change but I get your point

src/pkg/packager/network.go
```diff
@@ -22,9 +22,7 @@ import (
 
 // handlePackagePath If provided package is a URL download it to a temp directory.
 func (p *Packager) handlePackagePath() (partialPaths []string, err error) {
-	message.Debug("packager.handlePackagePath()")
-
-	opts := p.cfg.DeployOpts
+	opts := p.cfg.PkgOpts
 
 	// Check if the user gave us a remote package
 	providedURL, err := url.Parse(opts.PackagePath)
```
Racer159, line 28:
```suggestion
	providedURL, err := url.Parse(p.cfg.PkgOpts.PackagePath)
```

src/pkg/packager/network.go
```diff
@@ -22,9 +22,7 @@ import (
 
 // handlePackagePath If provided package is a URL download it to a temp directory.
 func (p *Packager) handlePackagePath() (partialPaths []string, err error) {
-	message.Debug("packager.handlePackagePath()")
-
-	opts := p.cfg.DeployOpts
+	opts := p.cfg.PkgOpts
```
Racer159, line 25:
```suggestion
```

src/pkg/packager/network.go
```diff
@@ -123,16 +120,16 @@ func (p *Packager) handleSgetPackage() error {
 	// If this is a DefenseUnicorns package, use an internal sget public key
 	if strings.HasPrefix(opts.PackagePath, fmt.Sprintf("%s://defenseunicorns", utils.SGETURLScheme)) {
```
Racer159, line 121:
```suggestion
	if strings.HasPrefix(p.cfg.PkgOpts.PackagePath, fmt.Sprintf("%s://defenseunicorns", utils.SGETURLScheme)) {
```

src/pkg/packager/network.go
```diff
@@ -104,10 +102,9 @@ func (p *Packager) handlePackagePath() (partialPaths []string, err error) {
 }
 
 func (p *Packager) handleSgetPackage() error {
-	message.Debug("packager.handleSgetPackage()")
 	message.Warn(lang.WarnSGetDeprecation)
 
-	opts := p.cfg.DeployOpts
+	opts := p.cfg.PkgOpts
```
Racer159, line 107:
```suggestion
```

src/pkg/utils/helpers/hash.go
```diff
@@ -26,3 +26,8 @@ func GetCRCHash(text string) uint32 {
 	table := crc32.MakeTable(crc32.IEEE)
 	return crc32.Checksum([]byte(text), table)
 }
+
+// GetSHA256 returns the computed SHA256 Sum of a given file
+func GetSHA256(data io.ReadCloser) (string, error) {
+	return GetCryptoHash(data, crypto.SHA256)
```
Racer159, line 32:
```suggestion
// GetSHA256Hash returns the computed SHA256 Sum of a given file
func GetSHA256Hash(data io.ReadCloser) (string, error) {
	return GetCryptoHash(data, crypto.SHA256)
```

src/pkg/utils/io.go
```diff
@@ -406,16 +398,23 @@ func IsDir(path string) bool {
...
 		return "", err
 	}
-	defer file.Close()
+	defer f.Close()
 
-	hash := sha256.New()
-	if _, err := io.Copy(hash, file); err != nil {
-		return "", err
-	}
+	return helpers.GetCryptoHash(f, crypto.SHA256)
```
Racer159, line 407:
```suggestion
	return helpers.GetSHA256Hash(f)
```

src/test/e2e/00_use_cli_test.go
```diff
@@ -119,12 +119,12 @@ func TestUseCLI(t *testing.T) {
 	t.Run("zarf package create with tmpdir and cache", func(t *testing.T) {
 		t.Parallel()
 		tmpdir := t.TempDir()
```
Racer159, line 121:
```suggestion
		tmpDir := t.TempDir()
```

src/types/runtime.go
```diff
@@ -30,15 +30,19 @@ type ZarfCommonOptions struct {
 	OCIConcurrency int    `jsonschema:"description=Number of concurrent layer operations to perform when interacting with a remote package"`
 }
 
-// ZarfDeployOptions tracks the user-defined preferences during a package deployment.
+// ZarfPackageOptions tracks the user-defined preferences during common package operations.
+type ZarfPackageOptions struct {
+	Shasum        string            `json:"shasum" jsonschema:"description=The SHA256 checksum of the package to deploy"`
```
Racer159, line 35:
```suggestion
	Shasum        string            `json:"shasum" jsonschema:"description=The SHA256 checksum of the package"`
```

src/types/runtime.go
```diff
@@ -30,15 +30,19 @@ type ZarfCommonOptions struct {
...
+	PackagePath   string            `json:"packagePath" jsonschema:"description=Location where a Zarf package to deploy can be found"`
```
Racer159, line 36:
```suggestion
	PackagePath   string            `json:"packagePath" jsonschema:"description=Location where a Zarf package to deploy can be found"`
```
```suggestion
	PackagePath   string            `json:"packagePath" jsonschema:"description=Location where a Zarf package can be found"`
```

src/types/runtime.go
```diff
@@ -30,15 +30,19 @@ type ZarfCommonOptions struct {
...
+	Components    string            `json:"components" jsonschema:"description=Comma separated list of optional components to deploy"`
```
Racer159, line 37:
```suggestion
	OptionalComponents    string            `json:"optionalComponents" jsonschema:"description=Comma separated list of optional components"`
```

Racer159 (changes requested):

Racer159 (approved):
lgtm

PR #1782, opened 2023-06-02
Use action defaults MaxTotalSeconds during Big Bang deploy

### LINKED ISSUE #1780: Big Bang Extension: Allow for customizing wait time on helmreleases

Currently when deploying a zarf package using the BB extension a series of wait actions are automatically injected to `onDeploy.after` to wait 10 minutes for each package inside of Big Bang (in order). However this is currently hardcoded with a timeout of 10 minutes - https://github.com/defenseunicorns/zarf/blob/main/src/extensions/bigbang/bigbang.go#L149 . While Big Bang does use a default 10 minute timeout on each package this is customizable, and if I raised it to 15 minutes in Big Bang I would like to also edit it in zarf to match. There are also cases where the 10 minute time period does not line up with Flux's 10 minutes due to dependencies, etc.

Allow for customizing the timeout, with a default of 10 minutes. One possibility to implement this:
```
    extensions:
      bigbang:
        wait_timeout: 900s
```

Remove the automatic waits from the extension (or allow them to be toggled off) so that I can build my own custom waits. This would also be acceptable, but would be more work for the end user to rebuild the functionality in a clean way. 

### Additional context

As an aside - the waits as-is are not very functional across upgrades due to timing of flux's reconciliation loop. Zarf will hit the wait functions before flux has actually triggered the upgrade across the packages, meaning that everything is in a ready state (pre-upgrade) and a deploy will almost always succeed quickly, without waiting. Having these type of waits in the extension may lead users to a false sense of security about the deployment, unless they are more robust.

Noxsios:
## Description

This PR makes it so that the Big Bang extension respects and uses the `onDeploy.defaults.maxTotalSeconds` field during its `onSuccess` actions to wait for HelmReleases to be ready.

The default time is still ten minutes (600s), but if a default is specified _and is greater than 600_, it will use that instead.

```yaml
components:
  - name: bigbang
    extensions:
      bigbang:
        version: 2.0.0
        valuesFiles:
          - config/disable-all.yaml
    actions:
      onDeploy:
        defaults:
          maxTotalSeconds: 1200 # 20 minutes in seconds
 ```

## Related Issue

Fixes #1780 

## Type of change

## Checklist before merging

### DIFF
--- a/src/extensions/bigbang/bigbang.go
+++ b/src/extensions/bigbang/bigbang.go
@@ -139,14 +139,20 @@ func Run(YOLO bool, tmpPaths types.ComponentPaths, c types.ZarfComponent) (types
 		return c, fmt.Errorf("unable to sort Big Bang HelmReleases: %w", err)
 	}
 
-	tenMinsInSecs := 10 * 60
+	// ten minutes in seconds
+	maxTotalSeconds := 10 * 60
+
+	defaultMaxTotalSeconds := c.Actions.OnDeploy.Defaults.MaxTotalSeconds
+	if defaultMaxTotalSeconds != 0 {
+		maxTotalSeconds = defaultMaxTotalSeconds
+	}
 
 	// Add wait actions for each of the helm releases in generally the order they should be deployed.
 	for _, hrNamespacedName := range namespacedHelmReleaseNames {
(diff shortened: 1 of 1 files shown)


Racer159 (approved):
lgtm

Issue #1760, opened 2023-05-31
Fail fast when handling incompatible package


timblaktu:
### Is your feature request related to a problem? Please describe.
`zarf package deploy`ing a package that was `zarf package create`d with a version of `zarf` that uses/used a different packaging format, can cause `deploy`-time errors related to the breaking schema change. These sorts of indirect errors can be some of the hardest to diagnose. This feature request aims to save users/developers time by catching these earlier and providing explicit clear feedback.

### Describe the solution you'd like
As early as possible, `zarf package deploy pkg.zst` will glean the `zarf` version used to create `pkg.zst`, compare that to the running `zarf` version, and if the delta includes known breaking schema change(s), emit an informative ERROR and `exit 1` immediately.

This presupposes that `zarf` release process uses something like semver, which makes it trivial to check for a breaking transition using arithmetic. If this is not the case, perhaps now would be the time to better define its versioning scheme and compatibility guarantees.

### Describe alternatives you've considered
Since we found this in CI originally, we considered implementing this check ourselves in the pipeline steps.

### Additional context
@lucasrod16 discovered this to be the cause of some [errors UDS team was seeing in `zarf-package-big-bang` CI](https://github.com/defenseunicorns/zarf-package-big-bang/actions/runs/5126388478/jobs/9236934135#step:13:2655). Specifically, our deploy pipeline was using 0.26.4 to deploy a package built with v0.27.0. [The 0.27.0 release notes](https://github.com/defenseunicorns/zarf/releases/tag/v0.27.0) clearly highlight the breaking change:

> Internal package structure changes will require packages built with v0.27.0 to be deployed with v0.27.0 or higher by @Noxsios in https://github.com/defenseunicorns/zarf/pull/1469 and by @Racer159 https://github.com/defenseunicorns/zarf/pull/1696
> 
>     Previously built packages are still compatible with v0.27.0, only newly created packages will require the higher version

We would love to have seen `zarf deploy` instead fail fast with something like:

```
ERROR: `zarf` v0.26.4 cannot `deploy` a package created with `zarf` v0.27.0 due to a breaking change.
```

...perhaps followed with some guidance and a built-in reminder about `zarf`'s versioning scheme and compatibility guarantees:

```
Zarf uses a variant of semver and cannot guarantee inter-operability between instances with different major or minor versions.
```

..or whatever is appropriate for the actual versioning/compatibility contract `zarf` prescribes.

Racer159:
[...]

PR #1788, opened 2023-06-05
Update package file locations to be more consistent

### LINKED ISSUE #1787: Allow templates in OCI manifest names (Standardize Package Internals)
**As** a package creator **I want** to have PKG_TMPLs in my manifest names **so that** I can better control what helm chart names are generated for my manifests.

- **Given** I have a `zarf.yaml` with a `manifest` with a `name` set to a template (i.e. `MY_NAME`)
- **And** the package is published as a skeleton
- **When** I import that skeleton package into another package
- **And** run `zarf package create --set MY_NAME=thename`
- **Then** Zarf successfully creates the package with my name templated

### Additional context
This would be easiest to fix by using the same layout as files and datainjections ('idx/filpath.Base()`) - we should also clarify how helm charts are named - it is a combination of package name component name **and** manifest name.

Racer159:
## Description

This updates filepaths to be more consistent within Zarf (but is a breakage for older versions of Zarf deploying new packages).

## Related Issue

Fixes #1787

## Type of change

## Checklist before merging

### DIFF
--- a/src/internal/cluster/data.go
+++ b/src/internal/cluster/data.go
@@ -56,10 +56,10 @@ iterator:
 	// The eternal loop because some data injections can take a very long time
 	for {
 		message.Debugf("Attempting to inject data into %s", data.Target)
-		source := filepath.Join(componentPath.DataInjections, filepath.Base(data.Target.Path))
+		source := filepath.Join(componentPath.DataInjections, strconv.Itoa(dataIdx), filepath.Base(data.Target.Path))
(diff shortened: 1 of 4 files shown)


Racer159:
Adding needs docs because we should document our compatibility strategy (namely that packages created with newer versions of Zarf may not be deployable with older ones - and we should address the concerns of https://github.com/defenseunicorns/zarf/issues/1760)

Racer159:
closing this for now

PR #1813, opened 2023-06-13
initial ADR for zarf hooks

YrrepNoj:


adr/0016-hooks.md
```diff
@@ -0,0 +1,60 @@
...
+Hooks lifecycle options will be before/after a package deployment and before/after a component deployment.
+ - NOTE: The order of hook execution is nearly random. If there are multiple hooks for a lifecycle there is no guarantee that they will be executed in a certain order.
+ - NOTE: The `package` lifecycle might be changed to a `run-once` lifecycle. This would benefit packages that don't have kube context information when the deployment starts.
+
+Hooks have to be 'installed' onto a cluster before they are used. When Zarf is deploying a package onto a cluster, it will look for any secrets with the `zarf-hook` label in the `zarf` namespace. If hooks are found, Zarf will run any 'package' level hooks before deploying a component and run any 'component' level hook for each component that is getting deployed.
+
+
+## Consequences
+
+- External hooks will likely not be implemented in the first pass of this feature. Handling external hooks will be an interesting challenge as we'll have to download the hook image and run execute it. Security will be something we consider heavily when implementing this feature.
```
Noxsios, line 52:
```suggestion
- External hooks will likely not be implemented in the first pass of this feature. Handling external hooks will be an interesting challenge as we'll have to download the hook image and run / execute it. Security will be something we consider heavily when implementing this feature.
```

Noxsios (changes requested):

adr/0016-hooks.md
```diff
@@ -0,0 +1,113 @@
...
+
+1. The package deployer solves the problem pre-deployment (creating needed repos, secrets, etc...)
+2. The package itself solves these problems with `actions` that are custom built for ECR clusters.
+
+Neither one of these current solutions are ideal. We don't want to require overly complex external + prior actions for Zarf package deployments, and we don't want package creators to have to create and distribute packages that are specific to ECR.
+
+## Approaches Considered
+
+### Internal Zarf Implementation
+ Clusters that have hooks will have `zarf-hook-*` secret(s) in the 'zarf' namespace. This secret will contain the hook's configuration and any other required metadata. As part of the package deployment process, Zarf will check if the cluster has any hooks and run them if they exist. Given the scenario above, there is no longer a need for an ECR specific Zarf package to be created. An ECR hook would perform the proper configuration for any package deployed onto that cluster; thereby requiring no extra manual intervention from the package deployer.
```
Racer159, line 34:
```suggestion
Clusters that have hooks will have `zarf-hook-*` secret(s) in the 'zarf' namespace. This secret will contain the hook's configuration and any other required metadata. As part of the package deployment process, Zarf will check if the cluster has any hooks and run them if they exist. Given the scenario above, there is no longer a need for an ECR specific Zarf package to be created. An ECR hook would perform the proper configuration for any package deployed onto that cluster; thereby requiring no extra manual intervention from the package deployer.
```

Racer159: Extra space

adr/0016-hooks.md
```diff
@@ -0,0 +1,113 @@
...
+lifecycle: before-component
+hookData:
+  registryURL: public.ecr.aws/abcdefg/zarf-ecr-registry
+  region: us-east-1
+  repositoryPrefix: ecr-zarf-registry
+```
+
+For this solution, hooks have to be 'installed' onto a cluster before they are used. When Zarf is deploying a package onto a cluster, it will look for any secrets with the `zarf-hook` label in the `zarf` namespace.  If hooks are found, Zarf will run any 'package' level hooks before deploying a component and run any 'component' level hook for each component that is getting deployed. The hook lifecycle options will be before/after a package deployment and before/after a component deployment.
+ - NOTE: The order of hook execution is nearly random. If there are multiple hooks for a lifecycle there is no guarantee that they will be executed in a certain order.
+ - NOTE: The `package` lifecycle might be changed to a `run-once` lifecycle. This would benefit packages that don't have kube context information when the deployment starts.
```
Racer159, line 61: We should probably say what all of the lifecycle events we expect to be are (I don't see a list of them).

adr/0016-hooks.md
```diff
@@ -0,0 +1,113 @@
...
+
+Zarf hooks will have two forms of execution via `Internal` and `External` hooks.
+### Internal Hooks
+Internal hooks will be hooks that are built into the Zarf CLI and run internal code when executed. The logic for these hooks would be built into the Zarf CLI and would be updated with new releases of the CLI.
+
+### External Hooks
```
Racer159, line 67:
```suggestion
#### Internal Hooks
Internal hooks will be hooks that are built into the Zarf CLI and run internal code when executed. The logic for these hooks would be built into the Zarf CLI and would be updated with new releases of the CLI.

#### External Hooks
```
Should have another indent on the header

adr/0016-hooks.md
```diff
@@ -0,0 +1,113 @@
...
+
+
+**PROS**
+ - Implementing Hooks internally means we don't have to deal with any bootstrapping issues.
+ - Internally managed hooks can leverage Zarf internal code.
+
+**CONS**
+ - Since 'Internal' hooks are built into the CLI, the only way to get updates for the hook is to update the CLI.
+ - External hooks will have a few security concerns that we will have to work through.
+
```
Racer159, line 86:
```suggestion
- Implementing Hooks internally adds more complexity to the Zarf CLI
```

adr/0016-hooks.md
```diff
@@ -0,0 +1,113 @@
...
+
+
+### Pepr
+Pepr is a K8s controller that enables Kubernetes mutations. We are (or will be) considering using Pepr to replace the `Zarf Agent`. Pepr is capable to accomplishing most of what Zarf wants to do with the concept of Hooks. Zarf hook configuration could be saved as secrets that Zarf will be able to use
+
+**PROS**
+ - Pepr as a solution would be more ?flexible?? than internal Zarf implementation of Hooks.
```
Racer159, line 93:
```suggestion
 - Pepr as a solution would be more flexible than the internal Zarf implementation of Hooks since the webhook could be anywhere.
```

adr/0016-hooks.md
```diff
@@ -0,0 +1,113 @@
...
+ - Implementing Hooks internally means we don't have to deal with any bootstrapping issues.
+ - Internally managed hooks can leverage Zarf internal code.
+
+**CONS**
+ - Since 'Internal' hooks are built into the CLI, the only way to get updates for the hook is to update the CLI.
+ - External hooks will have a few security concerns that we will have to work through.
+
+
+
+### Pepr
```
Racer159, line 89: We may want to just say "Webhooks" - this _could_ be Pepr or it could be some other random API.  It might also be worth showing what the secrets / lifecycle parts would look like for this too (i.e. instead of `OCIReference` it would show a `URL` and / or a `Selector` of some kind)

adr/0016-hooks.md
Racer159: (Especially if we do WASM)

adr/0016-hooks.md
```diff
@@ -0,0 +1,113 @@
...
+ - Lifecycle management would be a lot easier with a webhook solution like Pepr.
+
+**CONS**
+ - Pepr is a new project that hasn't been stress tested in production yet (but neither has Hooks).
+ - The Pepr image needs to be pushed to an image registry before it is deployed. This will require a new bootstrapping solution to solve the ECR problem we identified above.
+
+
+## Consequences
+
+- External hooks will likely not be implemented in the first pass of this feature. Handling external hooks will be an interesting challenge as we'll have to download the hook image and run / execute it. Security will be something we consider heavily when implementing this feature.
```
Racer159, line 105: I think it might be worth exploring both - if we don't want to deal with Pepr right away we could have an uglier expressjs server instead (since that is what Pepr would ultimately do under the hood) - that way we can get feedback / evaluate both approaches with a bit more data.

Racer159 (changes requested):

adr/0016-hooks.md
YrrepNoj: The new parent to that note is:
> Hooks lifecycle options will be before/after a package deployment and before/after a component deployment.

Does that satisfy what you'd like to see for expected lifecycle events?

adr/0016-hooks.md
YrrepNoj: I agree, that it'd be worth exploring both, but I feel like we might as well try with Pepr since that seems like the most flushed out solution that we would actually want to use.

Was there something you wanted me to add to the ADR here?

Racer159 (approved):
Approving as pending / consequences may change

adr/0016-hooks.md
```diff
@@ -0,0 +1,118 @@
+# 16. Zarf Hooks
```
Racer159, line 1: number update

Racer159 (changes requested):

Racer159 (approved):
lgtm

Noxsios (approved):

Issue #1711, opened 2023-05-16
Reduce Verbosity to import multi-component Zarf package


runyontr:
### Is your feature request related to a problem? Please describe.
When importing an entire Zarf package, each component needs to be identified and imported individually.   Currently it requires importing multiple components:

https://github.com/defenseunicorns/zarf-package-big-bang/blob/main/aws/zarf.yaml#L114-L126

### Describe the solution you'd like

- **Given** this [zarf package](https://github.com/defenseunicorns/zarf-package-big-bang/blob/main/defense-unicorns-distro/zarf.yaml)
- **When**creating a zarf package that uses this zarf package as "part" of it
- **Then** we could import the entire package at once:

```yaml
components:
...
- name: bigbang
  import:
    path: oci://ghcr.io/defenseunicorns/uds/bigbang:0.1.3
```

which would import all the components.

### Alternative

Something like this solution to performing a zarf init on the cluster:

https://github.com/runyontr/zarf-package-k3d/blob/main/zarf.yaml#L31-L39

Racer159:
@runyontr would "bundling" work in your use case: https://github.com/defenseunicorns/zarf/pull/1770

Noxsios:
Superseded by UDS Bundles + overall better UX around skeleton packages.

PR #1820, opened 2023-06-14
`zarf bundle` ADR

Noxsios:
## Description

Creates a `pending` ADR for initial work on `zarf bundle`.

## Related Issue

Relates to https://github.com/defenseunicorns/zarf/issues/1711

## Type of change

## Checklist before merging


adr/0017-zarf-bundle.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Status
+
+Pending
+
+## Context
+
+Orchestrating Zarf packages into meta-packages is a current weakpoint of Zarf. The core of Zarf was oriented around components as capabilities, but as Zarf packages have scaled, it has been discovered that entire packages are now rapidly becoming modularized capabilities.
+
+Currently there is no official way to enable the deployment, publishing, pulling, and creation of these Zarf packages. Due to this shortfalling, it has led to the community developing such antipatterns as:
```
Racer159, line 13:
```suggestion
Orchestrating capabilities from multiple Zarf packages into meta-packages is a current weakpoint for Zarf. The core of Zarf was built around components as capabilities, but as Zarf packages have scaled, there has been a need to create a new boundary to manage these capabilities efficiently.

Currently there is no official way to enable the deployment, publishing, pulling, and creation of multiple Zarf packages together, and due to this some in the community have resorted to patterns such as:
```
A bit better wording on some of the why (efficiency) and also wanted to stray away from the "antipattern" wording since it isn't really the communities' fault that they are doing that - to me "antipattern" is something that is well communicated as something to not do which up until recently the Zarf team has not done.

adr/0017-zarf-bundle.md
```diff
@@ -0,0 +1,93 @@
...
+  files:
+    - source: zarf-init-amd64-v0.27.0.tar.zst
+      target: zarf-init-amd64-v0.27.0.tar.zst
+  actions:
+    onDeploy:
+      after:
+        - cmd: zarf package deploy zarf-init-amd64-v0.27.0.tar.zst --components git-server --confirm -l warn
+```
+
+While this _does_ fulfill the need to deploy two packages in one command, it does so in such a way that is very brittle and creates bloated and inefficient packages.
```
Racer159, line 27:
```suggestion
While this _does_ fulfill the need to deploy two packages in one command, it does so in such a way that is verbose within the `zarf.yaml`, brittle across Zarf versions, inefficient within the package structure (it doesn't share layers), and is difficult to use `variables` with.
```
Expanding on a few more *ilities to be considered

adr/0017-zarf-bundle.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Proposed Solutions
+
+There are currently three proposed solutions to this, each with their own pros and cons:
+
+#### Zarf Bundle
+
+Uses OCI and a separate zarf schema / declarative YAML definition to pull packages into a single artifact and orchestrate them together.
```
Racer159, line 35:
```suggestion
Uses OCI and a separate Zarf schema / declarative YAML definition to pull packages into a single artifact and orchestrate them together.
```

adr/0017-zarf-bundle.md
```diff
@@ -0,0 +1,93 @@
...
+
+Biggest problem to solve is how to handle variables and whether/how those should be shared between packages.
+
+Package sources in bundles would be OCI _only_, and would not support local packages.
```
Racer159, line 39:
```suggestion
Pros:
- maintains efficient OCI layering / deduping of shared package resources
- allows flexibility in defining what `zarf bundle` (or a separate command) would look like as its own command without polluting `zarf package`

Cons:
- variables set within packages with `setVariables` may be difficult to share across packages
- package sources in bundles would be OCI _only_, and would not support local packages.
```
This is only cons and a separate pros/cons list may be easier to parse.

adr/0017-zarf-bundle.md
```diff
@@ -0,0 +1,93 @@
...
+
+#### Super Zarf Packages
+
+Add a packages key or another way to overlay packages into a larger package with the same internal structure.
+
+Biggest problem here is how we would scope things after layering things together such as how we might need to change Helm chart name generation or other package level things (like deployed package secrets)
```
Racer159, line 45:
```suggestion
Adds a packages key or another way to overlay packages into a larger package with the same internal structure as current Zarf packages.

Pros:
- packages would maintain the same syntax under `zarf package` between normal / meta packages.

Cons:
- it would be difficult to properly scope things like variables and helm chart names properly across packages.
- this continues to add to `zarf package` making it more complex and harder to test
```

adr/0017-zarf-bundle.md
```diff
@@ -0,0 +1,93 @@
...
+
+#### Zarf Package Manager
+
+Have a separate binary pull and manage packages together - this would also likely include dependency declaration and resolution between packages.
+
+Biggest problem is how do we define dependencies and know what is "installed" for a Zarf package.  Variable orchestration would also be an issue. The added workload of supporting an entire separate binary would also place strain on the Zarf team.
```
Racer159, line 51:
```suggestion
Uses a separate binary (not `zarf`) to pull and manage packages together - this would also include dependency declaration and resolution between packages.

Pros:
- this is a familiar/expressive way to solve the package problem and would be familiar to developers and system administrators
- allows flexibility in defining what the package manager would look like as its own command without polluting `zarf package`

Cons:
- dependencies may be difficult to determine whether they are "installed"/"deployed" - particularly for pre-cluster resources
- variables set within packages with `setVariables` may be difficult to share across packages
- this would necessitate a separate binary with it's own CLI that would need its own release schedule and maintenance

> :warning: **NOTE**: The package manager could also be made to be OCI-only but would come with the same OCI pros/cons.
```

Racer159: Again pros+cons

adr/0017-zarf-bundle.md
```diff
@@ -0,0 +1,93 @@
...
+
+packages:
+  - repository: ghcr.io/defenseunicorns/packages/dubbd
+    ref: 0.0.1 # OCI spec compliant reference
+    # arch is not needed as it will use w/e arch is set in the bundle's metadata
+    components:
+      - "*" # grab all components
+  - repository: docker.io/<namespace>/<name>
+    ref: 0.0.1
+    components:
```
Racer159, line 74:
```suggestion
    optionalComponents:
```
I think this should explicitly be optional components or we should add a note somewhere in the ADR that this works like the `--components` flag

adr/0017-zarf-bundle.md
```diff
@@ -0,0 +1,93 @@
...
+- `zarf bundle list` --> will probably just show the same as `zarf package list`
+- ~~`zarf bundle publish`~~ --> Bundles will be OCI only, so there is no need for a publish command, `create` will handle that.
+- `zarf bundle pull oci://<reference> -o <dir>`
+- `zarf bundle remove oci://<reference>|<path>`
+
+## Consequences
+
+This does add complexity to the Zarf codebase, as it is the addition of an entire suite of commands, JSON schema, schema docs, CLI docs, and chunk of library code + tests.  It is a good litmus test of the current packager and OCI codebase to see how ready it is to be consumed as a library.
+
+There is also the hard requirement that packages bundled must be first published to a registry available to the person performing the bundle operation. This removes some ability to develop bundles on an air gapped environment, but the team believes that in such scenarios, the air gapped environment should be _receiving_ a bundle, rather than developing one internal.
```
Racer159, line 93:
```suggestion
There is also the hard requirement that packages bundled must be first published to a registry available to the person performing the bundle operation. This removes some ability to develop bundles on an air gapped environment, but the team believes that in such scenarios, the air gapped environment should be _receiving_ a bundle, rather than developing one internally.  If this assumption is incorrect however there are options for us to allow the creation of bundles from OCI directories on local systems if we need to or we could provide more provisions within Zarf to make it easier to connect to airgap registries to mirror bundles.
```
May be worth noting that there are things we could do to help in this scenario.

Racer159 (changes requested):

corang:
I have a concern and a question. If the proposed `zarf bundle` is the decision that we go with it may necessitate some weird packaging, say if I want to deploy 2 packages in a zarf bundle but in between I have some manifest I need to deploy then I'd have to create and publish a whole other 3rd package that's just that manifest, while verbose I'd say it's on the line of too verbose and a lot more work than if there was a `packages` key in the existing zarf.yaml package definition. 

My question is can you provide variables to packages being deployed via a zarf bundle in the yaml file, and will zarf bundles use `###ZARF_PKG_TMPL_NAME###` / `###ZARF_VAR_NAME###` to be able to customize them at create/deploy time.

adr/0017-zarf-bundle.md
```diff
@@ -0,0 +1,119 @@
...
+
+Date: 2023-06-13
+
+## Status
+
+Pending
+
+## Context
+
+Orchestrating capabilities from multiple Zarf packages into meta-packages is a current weakpoint for Zarf. The core of Zarf was built around components as capabilities, but as Zarf packages have scaled, there has been a need to create a new boundary to manage these capabilities efficiently.
```
cmwylie19, line 11:
```suggestion
Orchestrating capabilities from multiple Zarf packages into meta-packages is a current weak point for Zarf. The core of Zarf was built around components as capabilities, but as Zarf packages have scaled, there has been a need to create a new boundary to manage these capabilities efficiently.
```

cmwylie19: i think weak point is two words, just nit picking

adr/0017-zarf-bundle.md
```diff
@@ -0,0 +1,119 @@
...
+- `zarf bundle deploy oci://<reference>|<path>`
+- `zarf bundle inspect oci://<reference>|<path>`
+- `zarf bundle list` --> will probably just show the same as `zarf package list`
+- ~~`zarf bundle publish`~~ --> Bundles will be OCI only, so there is no need for a publish command, `create` will handle that.
+- `zarf bundle pull oci://<reference> -o <dir>`
+- `zarf bundle remove oci://<reference>|<path>`
+
+## Consequences
+
+This does add complexity to the Zarf codebase, as it is the addition of an entire suite of commands, JSON schema, schema docs, CLI docs, and chunk of library code + tests.  It is a good litmus test of the current packager and OCI codebase to see how ready it is to be consumed as a library.
```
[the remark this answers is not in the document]
cmwylie19: a chunk of..

zachariahmiller:
I am curious on the points raised by @corang as well. A more generic question variable issues as it relates to this deicsion:

Is the thought that since variables are an issue in each of the options that overcoming limitations to that end would be addressed separately? Would it be fair to say from your perspective that the variable issues would be the same in each option and as such are inconsequential to this specific decision? If not could the assumed/known differences in those limitations for each option be elaborated on?

JasonvanBrackel:
I think I agree that "bundle" is the best of the 3 ideas.  That being said, I'd try to find a way to support local file patterns.  Even consider files -> OCI locally -> bundle

Noxsios:
@corang 

Since bundles are orchestrating packages post `package create`, there would not be any more `PKG_TMPL` as that is a create time only variable. `PKG_VAR` _would_ operate the same way as if you deployed that sub-package on its own. Sharing variables will be interesting (perhaps for deploying via bundles variables will get prefixed with the package name, or we add a `shared-variables` key to denote which variables in a package are to be shared across multiple packages).

corang:
> @corang
> 
> Since bundles are orchestrating packages post `package create`, there would not be any more `PKG_TMPL` as that is a create time only variable. `PKG_VAR` _would_ operate the same way as if you deployed that sub-package on its own. Sharing variables will be interesting (perhaps for deploying via bundles variables will get prefixed with the package name, or we add a `shared-variables` key to denote which variables in a package are to be shared across multiple packages).

But can I put a zarf variable in a zarf bundle (something like this):

```yaml
metadata:
  name: omnibus
  description: an example Zarf bundle
  version: 0.0.1
  arch: amd64
packages:
  - repository: ghcr.io/defenseunicorns/packages/dubbd
    ref: ###ZARF_VAR_DUBBD_VERSION###
```

And can I provide variable values to packages in a bundle (something like this):

```yaml
metadata:
  name: omnibus
  description: an example Zarf bundle
  version: 0.0.1
  arch: amd64
packages:
  - repository: ghcr.io/defenseunicorns/packages/dubbd
    ref: 0.0.1 # OCI spec compliant reference
    variables:
    	- name: DOMAIN
    	  value: bigbang2.dev
```

Noxsios:
@corang 

Yes, templating the `zarf-bundle.yaml` would operate nearly (not fully commited to 100%) the same as templating `zarf.yaml`.

For variables, I envision either what you have shown - essentially an embedded `zarf-config.yaml .package.deploy.set` or:

```yaml
packages:
  - repository: ghcr.io/defenseunicorns/packages/dubbd
    ref: 0.0.1 # OCI spec compliant reference
    variables:
      - dubbd-config.yaml
```

cmwylie19 (approved):

Noxsios:
> I think I agree that "bundle" is the best of the 3 ideas. That being said, I'd try to find a way to support local file patterns. Even consider files -> OCI locally -> bundle

@JasonvanBrackel For the initial POC, I believe targeting only OCI will deliver the most value, local file patterns can be a future ask if the community requires it.

adr/0017-zarf-bundle.md
```diff
@@ -0,0 +1,119 @@
...
+
+Pros:
+
+- maintains efficient OCI layering / deduping of shared package resources
+- allows flexibility in defining what `zarf bundle` (or a separate command) would look like as its own command without polluting `zarf package`
+
+Cons:
+
+- variables set within packages with `setVariables` may be difficult to share across packages
+- package sources in bundles would be OCI _only_, and would not support local packages.
```
YrrepNoj, line 45: This read like we are technically incapable of using local packages when creating bundles. That is not the case.


I get that we might have a preference for not using local packages but we should be explicit and list out our reasons for not wanting to rather than just stating that bundles won't support local packages.

adr/0017-zarf-bundle.md
```diff
@@ -0,0 +1,119 @@
...
+
+Date: 2023-06-13
+
+## Status
+
+Pending
+
+## Context
+
+Orchestrating capabilities from multiple Zarf packages into meta-packages is a current weak point for Zarf. The core of Zarf was built around components as capabilities, but as Zarf packages have scaled, there has been a need to create a new boundary to manage these capabilities efficiently.
```
Madeline-UX, line 11: Recommended change to this line "Orchestrating capabilities from multiple Zarf packages into a single artifact(package) for deployment is not currently supported in Zarf."

adr/0017-zarf-bundle.md
YrrepNoj: <3

adr/0017-zarf-bundle.md
```diff
@@ -0,0 +1,119 @@
...
+
+Pros:
+
+- maintains efficient OCI layering / deduping of shared package resources
+- allows flexibility in defining what `zarf bundle` (or a separate command) would look like as its own command without polluting `zarf package`
+
+Cons:
+
+- variables set within packages with `setVariables` may be difficult to share across packages
+- package sources in bundles would be OCI _only_, and would not support local packages. (this may change in the future, but for an intial POC it would be out of scope)
```
Racer159, line 45:
```suggestion
- package sources in bundles would be best to keep as OCI _only_, without support for local packages. This would help us ensure there are versions for packages and would help with efficiency by taking advantage of things like layer deduping.
```

Racer159 (changes requested):

Racer159 (approved):
lgtm

PR #1826, opened 2023-06-15
Improve the component actions page and related examples.

### LINKED ISSUE #1743: Document more about how zarf handles modifications to actions
Currently we don't have any documentation on `./zarf` or how Zarf modifies variables and commands to run on Windows systems.  This should be added to our docs page as tips admonitions so that they are more discoverable.

https://docs.zarf.dev/docs/create-a-zarf-package/component-actions
https://github.com/defenseunicorns/zarf/blob/main/packages/gitea/zarf.yaml#L53

### Additional context
`./zarf` is used a lot internally and is pretty powerful to not be documenting

Racer159:
## Description

This improves the component actions docs page and related examples.

## Related Issue

Fixes #1743 

## Type of change

## Checklist before merging


examples/kiwix/manifests/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+    spec:
+      # Kiwix can hot-load files from the filesystem, but if your application cannot, this example shows how you can use an initContainer to bootstrap the injected files.
+      initContainers:
+        - name: data-loader
+          image: alpine:3.18
+          command:
+            [
+              "sh",
+              "-c",
+              'while [ ! -f /data/###ZARF_DATA_INJECTION_MARKER### ]; do echo "waiting for zarf data sync" && sleep 1; done; echo "we are done waiting!"',
```
Noxsios, line 25: What is the `DATA_INJECTION_MARKER`?

examples/kiwix/manifests/persistence.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: kiwix-data
+spec:
+  storageClassName: "###ZARF_STORAGE_CLASS###"
```
Noxsios, line 6: Where does `STORAGE_CLASS` come from?

Noxsios (changes requested):

examples/kiwix/manifests/deployment.yaml
Racer159: Added a comment explaining it

examples/kiwix/manifests/persistence.yaml
Racer159: This is a built in var - I copied it in from @jeff-mccoy 's original Kiwix example, but this should be a regular variable instead.

Racer159: NVM I forgot that this comes from `init` - basically init queries the k8s distro and selections a storage class or (optionally) it uses the override from the --storage-class flag.

https://github.com/defenseunicorns/zarf/blob/main/src/internal/cluster/state.go#L120

Racer159: Made that a bit clearer in a comment

YrrepNoj (dismissed):
I think these put our docs / examples a lot more clear

YrrepNoj (approved):

YrrepNoj (approved):

Issue #1606, opened 2023-04-17
Add Package commands to deployed package context menu


Madeline-UX:
## User Story ##

As Ashton, I would like to be able to view the yaml package definitions for a package so that I can get more technical information about what is deployed.

Each command has it's own story, this issue is just for adding the menu component

CLI command: zarf inspect

Acceptance Criteria

Given - the user has packages deployed in their cluster
When - selects view code... from the actions menu
Then - Open a drawer that shows the Zarf.yaml for the package.
DESIGN THINGS..
And - Return user to previous screen

## Designs ## 

<img width="382" alt="Screenshot 2023-04-06 at 4 22 14 PM" src="https://user-images.githubusercontent.com/92826525/230496941-57445d1b-c992-4682-8c6b-d6d72714ca70.png">

Racer159:
@Madeline-UX this one needs a design too

Issue #1582, opened 2023-04-10
Error handling - for package deployment 


Madeline-UX:
## User Flow ## 

*Given* - Package deploys correctly
*Then* - modal shown with a success message and auto redirects back to the main page
*If* - Package fails to deploy
*Then* - Step that package failed in changes to error state AND the error message from CLI is in secondary text
*And* - Action buttons at the bottom of the screen are (secondary) Open Logs (primary) Return to Packages.

## Updates ##

- [x] Stay on the deploy page when the package fails
- [x] Change state for stepper that package failed on to error state
- [x] Change action buttons to reflect error state options
- [x] Add error message as secondary text to stepper *if possible*

## Design file ##

https://www.figma.com/file/MUItIzpzLBLuIyt225Bwgl/Zarf-UI?node-id=1369%3A58453&t=I6ve2QRHORWzJXZO-1

![Image](https://user-images.githubusercontent.com/92826525/236059497-a8ce7c51-be09-4094-bb61-fff894d05dc9.png)

### Further UX Research Questions ###

Driving questions for future iterations:

1. How much information does a user want to see upon success?
2. What information is the user looking for upon failure?
3. How does the user prefer to view that information CLI/Modal? More or less info
4. How do we ensure the user knows what state the deployment is in if they walk away and come back and the deployment is completed while they were gone?
5. If is deployment failed, how might we support quick and easy resolution?

Madeline-UX:
Design is ready

Issue #1312, opened 2023-01-31
Error handling for UI


Madeline-UX:
Error message 

When/what errors should be

- [ ] Stop progress > Dialog 
- [ ] Snackbars >
- [ ] Badges >

mike-winberry:
- [ ] Searching for packages and the handling of corrupted or "unreadable" packages. 
- [ ] Errors during the deploy screen.

Racer159:
Related to https://github.com/defenseunicorns/zarf/issues/1582

Racer159:
Another thing to consider while breaking this down is that our ErrorWebf function always returns `http.StatusInternalServerError` which depending on the error may not be correct.

PR #1828, opened 2023-06-16
1582 error handling for package deployment

mike-winberry:
## Description
#1582 
#1606 

## Type of change

## Checklist before merging


src/internal/cluster/zarf.go
```diff
@@ -31,13 +31,13 @@ func (c *Cluster) GetDeployedZarfPackages() ([]types.DeployedPackage, error) {
 	for _, secret := range secrets.Items {
 		var deployedPackage types.DeployedPackage
 		err := json.Unmarshal(secret.Data["data"], &deployedPackage)
+		// Don't break out of the loop if we can't unmarshal a secret, just log it
 		if err != nil {
-			message.Warnf("Unable to unmarshal package secret")
-
-			return deployedPackages, err
```
YrrepNoj, line 37: I think we should save the error and return it later. If we are unable to unmarshal a secret then something is probably really wrong with either Zarf or the secret. Simply warning it probably isn't enough.


I think it makes sense to continue parsing through the secrets so that the caller can get a list of all package secrets we were able to parse and error(s?) for the bad package secrets and decide how to handle it themselves.

YrrepNoj: @mike-winberry Is this error something you ran into yourself?

mike-winberry: Yeah so I think it should be tagged for https://github.com/defenseunicorns/zarf/issues/1312. Cause right now basically it just bailed and nothing was happening on the ui, so not even the working packages were showing up as a return for this method. And since there is no error handling for the front end right now, I still need it to show the good packages and then log the failures until there is something in place on the front end to handle those errors. But I can go ahead and at least capture the errors for the front end now even if the logic isn't in place to handle them.

mike-winberry: If theres an error when deploying the init package it kinda borks the whole ui unless I ignore the un-marshalling error.

mike-winberry: Its something I need to dig deeper into, but should be handled in #1312. I agree that we shouldn't be ignoring, Ill rewrite to add least pass the errors found back from this routine so they will be there for capturing in #1312.

mike-winberry: It mostly came up because I was having to force errors for this story in order to ensure they were being captured in the deploy page, then I started running into the unmarshalling error that was causing any package after a broken package to not be returned from this routine.

mike-winberry: Screenshot 2023-06-21 at 9 52 57 PM

mike-winberry: The specifics are logged to the log file for unmarshaling errors.

src/internal/cluster/zarf.go
```diff
@@ -18,29 +19,30 @@ import (
...
+			errorList = append(errorList, fmt.Errorf("unable to unmarshal the secret %s/%s", secret.Namespace, secret.Name))
+		} else {
+			deployedPackages = append(deployedPackages, deployedPackage)
 		}
 
-		deployedPackages = append(deployedPackages, deployedPackage)
 	}
 
-	return deployedPackages, nil
+	return deployedPackages, errorList
```
YrrepNoj, line 45:
```suggestion
    // TODO: If we move this function out of `internal` we should return a more standard singular error.
	return deployedPackages, errorList
```

src/config/lang/english.go
```diff
@@ -218,6 +218,7 @@ const (
 
 	CmdPackageListShort         = "Lists out all of the packages that have been deployed to the cluster (runs offline)"
 	CmdPackageListNoPackageWarn = "Unable to get the packages deployed to the cluster"
+	CmdPackageListUnmarshalErr  = "Unable to unmarshal all of the packages deployed to the cluster"
```
YrrepNoj, line 221:
```suggestion
	CmdPackageListUnmarshalErr  = "Unable to read all of the packages deployed to the cluster"
```

src/internal/api/packages/list.go
```diff
@@ -20,8 +20,8 @@ func ListDeployedPackages(w http.ResponseWriter, _ *http.Request) {
 		return
 	}
 
-	deployedPackages, err := c.GetDeployedZarfPackages()
-	if err != nil {
+	deployedPackages, errs := c.GetDeployedZarfPackages()
+	if len(errs) > 0 && len(deployedPackages) == 0 {
```
YrrepNoj, line 24:
```suggestion
    // TODO #1312: Handle errors where some deployedPackages were able to be parsed
	if len(errs) > 0 && len(deployedPackages) == 0 {
```

src/pkg/message/message.go
```diff
@@ -39,6 +39,9 @@ const (
 // NoProgress tracks whether spinner/progress bars show updates.
 var NoProgress bool
 
+// LogStream is the stream to write logs to.
+var LogStream io.Writer = os.Stderr
```
YrrepNoj, line 43:
```suggestion
// LogWriter is the stream to write logs to.
var LogWriter io.Writer = os.Stderr
```

(as well as all the other places <3)

src/ui/package.json
```diff
@@ -32,11 +33,11 @@
 		"@floating-ui/dom": "1.3.0",
 		"@fontsource/roboto": "5.0.3",
 		"@microsoft/fetch-event-source": "^2.0.1",
+		"ansi-to-html": "^0.7.2",
 		"material-symbols": "0.7.0",
 		"prismjs": "1.29.0",
 		"sanitize.css": "13.0.0",
-		"xterm": "^5.2.1",
-		"xterm-addon-fit": "^0.7.0",
+		"strip-ansi": "^7.1.0",
```
YrrepNoj, line 40:
```suggestion
```

YrrepNoj (changes requested):

YrrepNoj (approved):

Racer159 (approved):
Approving for now since this is a step forward though after playing with it a bit some of the design pieces are missing (particularly red error handling).

image
Top Left - Design File
Top Right - As Implemented in WebUI
Bottom Left - The Web UI log output
Bottom Right - The CLI log output for that failing package

PR #1835, opened 2023-06-21
Adding crane ls command

dgershman:
## Description

This adds support for `crane ls`.  This feature is very useful when attempting to do some automation of updating images.  The only other way to do this is to run `zarf connect registry` and then request the tags endpoint of the registry.

## Related Issue

N/A

## Type of change

## Checklist before merging


src/cmd/tools/crane.go
```diff
@@ -88,3 +91,54 @@ func zarfCraneCatalog(cranePlatformOptions *[]crane.Option) *cobra.Command {
 
 	return craneCatalog
 }
+
+// Wrap the original crane list with a zarf specific version
+func zarfCraneList(cranePlatformOptions *[]crane.Option) *cobra.Command {
+	craneList := craneCmd.NewCmdList(cranePlatformOptions)
+
+	eg := `  # list the repos internal to Zarf
```
YrrepNoj, line 99:
```suggestion
	eg := `  # list the tags for a repo internal to Zarf
```

src/cmd/tools/crane.go
```diff
@@ -88,3 +91,54 @@ func zarfCraneCatalog(cranePlatformOptions *[]crane.Option) *cobra.Command {
...
+  $ zarf tools registry ls internal/stefanprodan/podinfo
+
+  # list the repos for reg.example.com
```
YrrepNoj, line 102:
```suggestion
  # list the tags for a repo hosted at reg.example.com
```

YrrepNoj (changes requested):

src/cmd/tools/crane.go
```diff
@@ -88,3 +91,54 @@ func zarfCraneCatalog(cranePlatformOptions *[]crane.Option) *cobra.Command {
...
+	craneList.Args = nil
+
+	originalListFn := craneList.RunE
+
+	craneList.RunE = func(cmd *cobra.Command, args []string) error {
+		if len(args) < 1 {
+			message.Fatal(nil, lang.CmdToolsCraneListNoRepoSpecified)
+		}
+
+		if !strings.HasPrefix(args[0], "internal/") {
```
YrrepNoj, line 115: Instead of having an `internal/` prefix, we should have the user provide the full host and check if the host matches the value in the `zarfState.RegistryInfo.URL` secret that gets loaded later in this function.

If the registry hostname matches, we can use the username/password from the state to configure the `authOption`. Otherwise we will not provide an  `authOption` and let crane use the values it finds on the hosts local credential store to make the request.


I.e.
```bash
zarf init --registry-url="my-registry.com"

zarf tools crane registry ls myregistry/myrepo/myimage      # use username/password from secret in state
```

src/cmd/tools/crane.go
```diff
@@ -88,3 +91,54 @@ func zarfCraneCatalog(cranePlatformOptions *[]crane.Option) *cobra.Command {
...
+			return originalListFn(cmd, args)
+		}
+
+		// Load Zarf state
+		zarfState, err := cluster.NewClusterOrDie().LoadZarfState()
```
YrrepNoj, line 120: Instead of `cluster.NewClusterOrDie()` we will need to use `cluster.NewCluster()` and then manually check if we are connected to the cluster.

Right now there is no way to call the original `crane list` function without having an initialized k8s cluster.

YrrepNoj (changes requested):

src/cmd/tools/crane.go
dgershman: Can you clarify the 2nd part of your comment?  You run `crane list` against a registry, not a k8s cluster.  Are you speaking specifically about a Zarf initialized cluster?

dgershman: I believe I understood what you said.  Please review my latest changes.

src/cmd/tools/crane.go
```diff
@@ -82,3 +86,62 @@ func zarfCraneCatalog(cranePlatformOptions *[]crane.Option) *cobra.Command {
...
+
+// Wrap the original crane list with a zarf specific version
+func zarfCraneList(cranePlatformOptions *[]crane.Option) *cobra.Command {
+	craneList := craneCmd.NewCmdList(cranePlatformOptions)
+
+	eg := `  # list the tags for a repo internal to Zarf
+  $ zarf tools registry ls 127.0.0.1:31999/stefanprodan/podinfo
+
+  # list the tags for a repo hosted at reg.example.com
+  $ zarf tools registry ls reg.example.com/stefanprodan/podinfo`
```
Racer159, line 98: Can this be under `lang` like `lang.CmdToolsRegistryCatalogExample`?

src/cmd/tools/crane.go
```diff
@@ -82,3 +86,62 @@ func zarfCraneCatalog(cranePlatformOptions *[]crane.Option) *cobra.Command {
...
+
+		// Open a tunnel to the Zarf registry
+		tunnelReg, err := cluster.NewZarfTunnel()
+		if err != nil {
+			return err
+		}
+		err = tunnelReg.Connect(cluster.ZarfRegistry, false)
+		if err != nil {
+			return err
+		}
```
Racer159, line 135:
```suggestion
    if  zarfState.RegistryInfo.InternalRegistry {
		  // Open a tunnel to the Zarf registry
		  tunnelReg, err := cluster.NewZarfTunnel()
		  if err != nil {
			  return err
		  }
		  err = tunnelReg.Connect(cluster.ZarfRegistry, false)
		  if err != nil {
			  return err
		  }
		  
		  givenAddress := fmt.Sprintf("%s/", zarfState.RegistryInfo.Address)
		  tunnelAddress := fmt.Sprintf("%s/", tunnelReg.Endpoint())
		  args[0] = strings.Replace(args[0], givenAddress, tunnelAddress, 1)
		}
```
We should only open a tunnel if it is an internal reg

src/cmd/tools/crane.go
```diff
@@ -82,3 +86,62 @@ func zarfCraneCatalog(cranePlatformOptions *[]crane.Option) *cobra.Command {
...
+
+		// Add the correct authentication to the crane command options
+		authOption := config.GetCraneAuthOption(zarfState.RegistryInfo.PullUsername, zarfState.RegistryInfo.PullPassword)
+		*cranePlatformOptions = append(*cranePlatformOptions, authOption)
+		registryEndpoint := tunnelReg.Endpoint()
```
Racer159, line 140:
```suggestion
```
Dropping this here per the above

src/cmd/tools/crane.go
```diff
@@ -82,3 +86,62 @@ func zarfCraneCatalog(cranePlatformOptions *[]crane.Option) *cobra.Command {
...
+
+		return originalListFn(cmd, []string{strings.Replace(args[0],
+			fmt.Sprintf("%s/", zarfState.RegistryInfo.Address), fmt.Sprintf("%s/", registryEndpoint), 1)})
```
Racer159, line 143:
```suggestion
		return originalListFn(cmd, args)
```
This can be simplified if the above is accepted.

Racer159 (changes requested):
I think the logic overall looks good - added some feedback on internal vs external registry support. Also you'll need to run `make docs-and-schema` to update the CLI docs.  Also as a heads up / note this may hit some small merge conflicts with https://github.com/defenseunicorns/zarf/pull/1848 as well

Racer159 (approved):
lgtm

Racer159 (changes requested):
`make docs-and-schema` looks like it still needs to be ran

Racer159 (approved):
lgtm (ah GitHub race conditions...)

dgershman:
I am seeing an issue with `make docs-and-schema` where its updating all the docs with an extra line break at the end of each markdown file.

Racer159:
> I am seeing an issue with `make docs-and-schema` where its updating all the docs with an extra line break at the end of each markdown file.

Hmm that might be because of your OS...  What were you devving on?

dgershman:
> 

Ventura 13.4

Racer159:
> > 
> 
> macOS Ventura 13.4

Hmm I mostly run flavors of Linux - might be something with the `sed` command we run under `docs/gen-cli-docs.sh`

Racer159:
Looks like if you remove one newline from the doc you added that would make it pass.

dgershman:
> > > 
> > 
> > 
> > macOS Ventura 13.4
> 
> Hmm I mostly run flavors of Linux - might be something with the `sed` command we run under `docs/gen-cli-docs.sh`

Yah.  Fixed on my end with an alias to gsed.  https://superuser.com/questions/307165/newlines-in-sed-on-mac-os-x

Racer159 (changes requested):
image

Looks like catalog and registry also changed (since ls is a new subcommand of registry).

Racer159:
> image
> 
> Looks like catalog and registry also changed (since ls is a new subcommand of registry).

(might still need catalog to be updated but we'll see... GH doesn't do a good job of showing updates when they happen)

Racer159 (approved):
lgtm! :tada:

YrrepNoj (approved):

dgershman:
PS, I plan to work in `crane push` and `crane pull` so they also don't require establishing a tunnel if they are referencing an internal image.  Is this cool?

Racer159:
Yep! That works following the pattern established here.  Will definitely make the crane commands more useful.

PR #1853, opened 2023-06-26
[FEATURE] add metrics to zarf-agent

### LINKED ISSUE #1849: Zarf Agent to emit Prom Metrics
In the enterprise, it is common for companies to request that applications emit Prom style metrics. Due to this reason, i think an _optional_ `/metrics` endpoint here in the [zarf-agent server component](https://github.com/defenseunicorns/zarf/blob/6fca25a747f94cd7409ad144f6645ddca7811aed/src/internal/agent/http/server.go#L29) could be beneficial.

- **Given** the zarf agent is deployed and there is a `command` added to the [Deployment](https://github.com/defenseunicorns/zarf/blob/6fca25a747f94cd7409ad144f6645ddca7811aed/packages/zarf-agent/manifests/deployment.yaml#L24) like `zarf serve --metrics` 
- **When** the agent is deployed
- **Then** based on the flags supplied to the zarf binary, a `/metrics` endpoint is exposed.

This is a nice to have, and I propose a very simple solution to see if it would get traction in the community. Furthermore, i think the solution should be _configurable_, so that metrics would not be turned on by deafult. This would be coupled with another optional zarf package with the prometheus operator, `ServiceMontors`, `PromRules` and a `README.md` explaining how to [granularly](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/) scrape only the relevant metrics that are needed.

OR maybe it would be easier to pass metrics as an ENVVAR

Alternatively, we could only implement this in Pepr (pepr-zarf-agent) to keep the size of Zarf down

cmwylie19:
## Description

Expose Prometheus metrics through `/metrics` HTTP endpoint. 
Add example of how to scrape metrics.

The example valids:
- Named Ports PR
- Git Tag and Chart Version Assumed Match PR
- Wait-For PR (jsonPath)

## Related Issue

Fixes #1849 

Relates to #1849 

## Type of change

## Checklist before merging


cmwylie19:
There will be functionality added to the Big Bang example to show how to scrape the zarf-agent metrics with Prometheus.

examples/big-bang/README.md
```diff
@@ -42,3 +42,125 @@ To view the example in its entirety, select the `Edit this page` link below the
 :::
 
 <ExampleYAML example="big-bang/yolo" showLink={false} />
+
+## Big Bang Scrape Zarf Agent Metrics
```
Racer159, line 46:
```suggestion
## Big Bang Zarf Agent Metrics Support
```
Makes this more similar to the other YOLO Mode heading

examples/big-bang/README.md
```diff
@@ -42,3 +42,125 @@ To view the example in its entirety, select the `Edit this page` link below the
...
+    matchLabels:
+      app: agent-hook
+EOF
+```
+
+At this point, we can curl against the Prometheus API to ensure the Zarf Agent target has been picked up by the Prometheus Operator.
+
+```bash
+# terminal 1
+$ kubectl port-forward svc/prometheus-operated 9090 -n monitoring
```
Racer159, line 86: This can actually be done with `zarf connect` as well:
```
zarf connect --name=prometheus-operated --namespace monitoring --remote-port 9090
```

examples/big-bang/README.md
```diff
@@ -42,3 +42,125 @@ To view the example in its entirety, select the `Edit this page` link below the
...
 <ExampleYAML example="big-bang/yolo" showLink={false} />
+
+## Big Bang Scrape Zarf Agent Metrics
+
+The Zarf Agent emits Prometheus metrics that can be scraped by Big Bang's Prometheus Operator. To enable this, set `monitoring.enabled` to true in the `config/disable-all.yaml`, and uncomment the `disable-all.yaml` under the `components.extentions.bigbang.valuesFiles` section in `zarf.yaml`.
+
+Finally, create a `ServiceMonitor` for the Zarf Agent. Since this the Zarf Agent exposes an `https` port, we need to provide the `bearerTokenFile` and `tlsConfig` to specify the TLS settings for scraping agaist the service. 
+
+```yaml
+kubectl create -f -<<EOF
```
Racer159, line 53: We wrap `kubectl` under `zarf tools` and could re-plug that here.  I also wonder if this is worth being a sub-example like the YOLO mode package is where we show the Zarf YAML that would apply this instead of the raw service monitor with a `kubectl apply`

Racer159 (changes requested):

examples/big-bang/README.md
cmwylie19: sounds like a good idea to me

cmwylie19: ✅

cmwylie19: hey @Racer159 , I was diving into the YOLO mode example and I don't know if we can give an example in this scenario as the `agent-hook` will not be around to scrape. I could be missing something, though. Let me know your thoughts around this, thanks! 🦄

cmwylie19: Suggestion's from @Racer159 after 1on1:
- [ ] update the bundle to install in monitoring ns
- [ ] refactor that zarf package now understanding that array and potential for an array of manifests should it be needed taking out the cmd
- [ ] simplify the port-forwarding piece by adding labels
consolidate the demos so that there is not 2 different demos - pointing the the scrape - [ ] Zarf Agent demo from the big bang demo 
- [ ] get a test in around prometheus running
- [ ] take out hello zarf since it does not have much to do with scraping zarf agent

Racer159: On num 3 this might be able to be achieved by adding a secondary service that is not controlled by the operator

examples/scraping-zarf-agent/zarf.yaml
```diff
@@ -0,0 +1,38 @@
+kind: ZarfPackageConfig
+metadata:
+  name: scrape-zarf-agent
+  yolo: false
```
Racer159, line 4:
```suggestion
```
We don't need to call out yolo being false

src/internal/cluster/tunnel.go
```diff
@@ -44,6 +44,7 @@ const (
 	ZarfLogging  = "LOGGING"
 	ZarfGit      = "GIT"
 	ZarfInjector = "INJECTOR"
+	Prometheus   = "PROMETHEUS"
```
Racer159, line 47: We shouldn't need these if we add our own service to connect to prometheus

examples/big-bang/README.md
```diff
@@ -42,3 +42,9 @@ To view the example in its entirety, select the `Edit this page` link below the
 :::
 
 <ExampleYAML example="big-bang/yolo" showLink={false} />
+
+## Big Bang Zarf Agent Metrics Support
+
+Using Big Bang's Prometheus instance, we can easily add a `ServiceMonitor` to scrape the Zarf Agent's metrics. Check out the [scraping-zarf-agent example](../scraping-zarf-agent/README.md) for more details. Make sure monitoring in enabled.
```
Racer159, line 48:
```suggestion
:::info

Using Big Bang's Prometheus instance, we can easily add a `ServiceMonitor` to scrape the Zarf Agent's metrics. Check out the [scraping-zarf-agent example](../scraping-zarf-agent/README.md) for more details. Make sure monitoring in enabled.
```
Without :::info the admonition won't show correctly

examples/scraping-zarf-agent/README.md
```diff
@@ -0,0 +1,76 @@
+import ExampleYAML from "@site/src/components/ExampleYAML";
+
+# Scrape Zarf Agent
```
Racer159, line 3:
```suggestion
# Zarf Agent Metrics (Prometheus)
```
We may want to rename this to be more explicit in what we are scarping / how (similar to the `Data Injection (Kiwix)` example)

examples/scraping-zarf-agent/manifests/prometheus-operator/clusterrole.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - ""
+  resources:
+  - pods
+  - pods/status
+  - endpoints
+  - services
+  verbs:
+  - watch
+  - get
+  - list
```
Racer159, line 17: The linter complains when the file doesn't end in a newline
image

examples/scraping-zarf-agent/manifests/prometheus-operator/clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+  creationTimestamp: null
+  name: scrape-cr-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: scrape-cr
+subjects:
+- kind: ServiceAccount
+  name: prometheus-operator
+  namespace: monitoring
```
Racer159, line 13: newline

examples/scraping-zarf-agent/manifests/prometheus-operator/servicemonitor-agent.yaml
```diff
@@ -0,0 +1,25 @@
...
+      insecureSkipVerify: true
+      # host name for the TLS handshake
+      serverName: agent-hook.zarf.svc
+  jobLabel: zarf-agent
+  namespaceSelector:
+    matchNames:
+    - zarf
+  selector:
+    matchLabels:
+      app: agent-hook
```
Racer159, line 25: newline

examples/scraping-zarf-agent/manifests/prometheus-deps/bundle.yaml
Racer159, line 1: This is quite a large file to have here - can we reference it from the internet instead?  We should be able to do so with a link to its URL and a SHASUM at the end.

(see the remote tab here: https://docs.zarf.dev/docs/create-a-zarf-package/zarf-components?chart-examples=URL+%28git%29&file-examples=Remote+with+SHA+sums&manifest-examples=Remote#kubernetes-manifests)

Racer159 (changes requested):

examples/scraping-zarf-agent/manifests/prometheus-deps/bundle.yaml
Racer159: (That *should* make it easier to keep up to date / version)

cmwylie19: That's a good idea. Only obstacle is that the bundle installs in the `default` namespace. 

- What if we pull it from the [URL](https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/main/bundle.yaml) and do a sed replace on `s/namespace: default/namespace: monitoring/g`? Would something like that work with some Zarf sorcery?

- Or, I could install it from the official helm chart if that makes more sense.

examples/big-bang/README.md
cmwylie19: great idea

examples/scraping-zarf-agent/manifests/prometheus-deps/bundle.yaml
cmwylie19: We are installing this from the helm chart now once this [PR](https://github.com/defenseunicorns/zarf/pull/1892) merges

src/internal/cluster/tunnel.go
cmwylie19: fixed thanks

Issue #1839, opened 2023-06-22
Slim init package for external registry and git repository


laurentiusoica:
With external registry and repository, I don't need the zarf registry or the git components. But if I build a custom init package and I don't include the registry replated components, the cluster is not initialized; the state secret object is not there under zarf namespace.

Racer159:
Thanks for the issue!

Definitely going to keep this one open as we look to modularize our init process more and more.

Currently Zarf looks for an init component named `zarf-seed-registry` as a marker that it is passed the `k3s` component and has a K8s cluster that it can talk to. To get around this currently you can create an empty required component that is called `zarf-seed-registry` as below:

```
kind: ZarfInitConfig
metadata:
  name: init
  description: Used to establish a new Zarf cluster

components:
  # Creates the temporary seed-registry
  - name: zarf-seed-registry
    required: true

  # Creates the pod+git mutating webhook
  - name: zarf-agent
    required: true
    import:
      path: packages/zarf-agent
```

Racer159:
(once #1855 is merged zarf state will now be created before the first init component that requires the cluster (in your case the `zarf-agent`))

PR #1855, opened 2023-06-27
Add only filter support and init package removal by package name

Racer159:
## Description

This adds `only` filter support to `zarf package remove` as well as allowing removal of init packages by their `.tar.zst`

## Related Issue

Fixes: #1859
Fixes: #1839

## Type of change

## Checklist before merging


Racer159:
Added `needs-tests` since this should have a final init package remove step at the very end now that that is possible.

src/pkg/packager/components.go
```diff
@@ -251,3 +251,17 @@ func (p *Packager) appendIfNotExists(slice []string, item string) []string {
 	}
 	return append(slice, item)
 }
+
+func (p *Packager) requiresCluster(component types.ZarfComponent) (confirmComponent bool) {
```
YrrepNoj, line 255: nit; it's a bit weird to have a named return but not use it.

YrrepNoj (approved):
LGTM. Left a nit comment up to you.

src/pkg/packager/components.go
Racer159: Good find!  Fixed

YrrepNoj (approved):

PR #1860, opened 2023-06-27
include empty files on package creation

### LINKED ISSUE #1791: Empty files under the `files` key don't get put into the package
### Environment
Device and OS: ubuntu 22 VM
App version: 0.27.0

### Steps to reproduce
1. create a zarf.yaml with a `files` component
2. Make the file that the component refernces empty
3. build the package
4. deploy the package

### Expected result

The empty file is placed where specified

### Actual Result

Package fails to deploy with

```
 Saving values.yaml                                                                                                                                                                 
  ⠋  Saving values.yaml (0s)     ERROR:  Failed to deploy package: unable to deploy all components in this Zarf Package: unable to deploy                                           
             component values: unable to process the component files: unable to copy file
             /tmp/zarf-2295005221/components/values/files/0 to values.yaml: lstat
             /tmp/zarf-2295005221/components/values/files/0: no such file or directory
```

### Severity/Priority

Meh, I don't know why you'd want an empty file

### Additional Context

YrrepNoj:
fixes #1791


src/pkg/utils/io.go
```diff
@@ -62,6 +62,13 @@ func CreateDirectory(path string, mode os.FileMode) error {
 	return nil
 }
 
+// CreateFile creates an empty file at the given path.
+func CreateFile(filepath string) error {
+	f, err := os.Create(filepath)
+	defer f.Close()
+	return err
+}
```
Noxsios, line 70: nit: call this `Touch` instead as it closely mirrors the `touch` CLI

Racer159: I disagree - `CreateFile` is likely a better name given we already have `CreateDirectory` and `CreatePathAndCopy` - I would say that we should probably match the functionality in `CreateDirectory` of checking `InvalidPath(path)`in the function.

Racer159: If we do go with `Touch` we should update everything to the POSIX equivalent

Racer159 (approved):

src/pkg/utils/io.go
Noxsios: tbh I am not opposed to that. I do agree it would break consistency with what is already there, so I will rescind my claim

Racer159: After thinking about it more, changing it to a request changes (for the invalid path stuff InvalidPath(path))

Racer159 (changes requested):

src/pkg/utils/io.go
```diff
@@ -207,7 +219,17 @@ func CreatePathAndCopy(source string, destination string) error {
...
-	return copy.Copy(source, destination)
+	// Copy all the source data into the destination location
+	if err := copy.Copy(source, destination); err != nil {
+		return err
+	}
+
+	// If the path doesn't exist yet then this is an empty file and we should create it
+	if InvalidPath(destination) {
+		return CreateFile(destination)
+	}
```
Racer159, line 230:
```suggestion
	return CreateFile(destination)
```
We don't need the double check

Racer159 (changes requested):

Racer159 (approved):
lgtm

PR #1875, opened 2023-07-02
Adds support to mutate repoURL in Application CRD and Repositories for ArgoCD

dgershman:
## Description

This adds support to have the Zarf Agent mutate the `repoURL` field inside of the `Application` CRD of ArgoCD.  This would enable those who are utilizing Argo as part of their GitOps deployment strategy to gain the same benefits as the current support that the Agent has  present for Flux.

## Related Issue

[Slack Discussion](https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1682172181361569)

## Type of change

## Checklist before merging


dgershman:
I have to make another change here as ArgoCD also requires Repository configuration that matches which also defines the authentication.  This is actually stored as a `Secret`.

dgershman:
> I have to make another change here as ArgoCD also requires Repository configuration that matches which also defines the authentication. This is actually stored as a `Secret`.

Ok added this.  Running some more tests.

dgershman:
> > I have to make another change here as ArgoCD also requires Repository configuration that matches which also defines the authentication. This is actually stored as a `Secret`.
> 
> Ok added this. Running some more tests.

Testing looks good.  This is ready for review.

src/internal/agent/hooks/argocd-application.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/defenseunicorns/zarf/src/pkg/utils"
+	"github.com/defenseunicorns/zarf/src/types"
+	v1 "k8s.io/api/admission/v1"
+)
+
+type Source struct {
+	Path           string `json:"path"`
+	RepoURL        string `json:"repoURL,omitempty"`
+	TargetRevision string `json:"targetRevision,omitempty"`
+}
```
Racer159, line 24:
```suggestion
type Source struct {
	RepoURL        string `json:"repoURL"`
}
```

This doesn't entirely match the upstream struct that defines this object: https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L163 (`repoURL` is required and `omitempty` will make it optional), and technically we only use RepoURL so we should only need that in this struct.

src/internal/agent/hooks/argocd-application.go
```diff
@@ -0,0 +1,104 @@
...
+
+type ArgoApplication struct {
+	Spec struct {
+		Source Source `json:"source"`
```
Racer159, line 28: In addition to `Source` we may also want to edit `Sources` https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L84

packages/zarf-agent/manifests/webhook.yaml
```diff
@@ -91,3 +91,91 @@ webhooks:
...
+          - "argoproj.io"
+        apiVersions:
+          - "v1alpha1"
+        resources:
+          - "applications"
+    admissionReviewVersions:
+      - "v1"
+      - "v1beta1"
+    sideEffects: None
+  - name: agent-argocd-repository.zarf.dev
```
Racer159, line 136:
```suggestion
  - name: agent-argocd-repocreds.zarf.dev
```
It may be better to call these repo creds since that is what they are called in the upstream code (and they are distinct from `Repository`) - https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/repository_types.go#L16

Racer159 (changes requested):
Overall the code looks good! - just a few small suggested changes.

Also you mentioned you had some tests setup, can you add a test under `src/test/e2e/22_git_and_flux_test.go` for ArgoCD? (potentially renaming that test file to `22_git_and_gitops_test.go`)

Also as a note, we have been standardizing around `podinfo` for most tests to keep things fast / lean (since we only have to pull / push a single image like `podinfo:6.4.0` once) but if that is too far out from what you have we can pull in other simple examples for now as well (like `guestbook`).  If you want you could put the package as an example under `examples` (if you want to add a README.md to it for Argo) or under `src/test/packages` with a prefix of `22` to just tie it to that test.

src/internal/agent/http/server.go
```diff
@@ -12,20 +12,24 @@ import (
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 )
 
-// NewAdmissionServer creates an http.Server for the mutating webhook admission handler.
```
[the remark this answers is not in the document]
dgershman: I blame IntelliJ

packages/zarf-agent/manifests/webhook.yaml
dgershman: Technically they are not repocreds.  Repocreds are credential templates that enable for reuse across multiple Repositories.  We are indeed catching a Repository.

Racer159: Ah I see the correct struct now: https://github.com/argoproj/argo-cd/blob/master/util/settings/settings.go#L286

Racer159: Ah yeah here is the magic function: https://github.com/argoproj/argo-cd/blob/d7632df378577229c8228b6e2d4c8aeeb31e7f7e/util/db/repository_secrets.go#L302 (they don't just marshal it over like I expected)

dgershman:
While attempting to build a package for argo-cd I'm getting an error because the chart contains a subchart that is not already present in the `charts/` path.  Is there a way to tell the `charts` component to run `helm dependency update` to ensure the subcharts are included?

dgershman:
> While attempting to build a package for argo-cd I'm getting an error because the chart contains a subchart that is not already present in the `charts/` path. Is there a way to tell the `charts` component to run `helm dependency update` to ensure the subcharts are included?

Here is the specific dependency https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/Chart.yaml#L20.

Racer159:
With that merge ^ there is now support for building dependencies (which is to update as npm ci is to npm install)

dgershman:
> With that merge ^ there is now support for building dependencies (which is to update as npm ci is to npm install)

I have a workaround for this by not using the Git repository and using the Helm repo instead.  However, it looks like it's failing before it can have a chance to install dependencies with the Git Repository method.  It's not a blocker for me, but something to perhaps be aware of.

```
 WARNING  Helm is unable to save the archive and create the package
  DEBUG   2023-07-12T22:17:03-05:00  -  found in Chart.yaml, but missing in charts/ directory: redis-ha
  DEBUG   2023-07-12T22:17:03-05:00  -  unable to add component: error creating chart archive, unable to pull the chart from git: found in Chart.yaml, but missing in charts/ directory: redis-ha
     ERROR:  Failed to create package: unable to add component: error creating chart archive, unable to pull the
             chart from git: found in Chart.yaml, but missing in charts/ directory: redis-ha
  DEBUG   2023-07-12T22:17:03-05:00  -  goroutine 1 [running]:
          runtime/debug.Stack()
          	/usr/local/opt/go/libexec/src/runtime/debug/stack.go:24 +0x64
          github.com/defenseunicorns/zarf/src/pkg/message.Fatal({0x108e2ca40?, 0x1400095c3c0?}, {0x140009f8480, 0xb3})
          	/Users/dgershman/Projects/go/src/github.com/defenseunicorns/zarf/src/pkg/message/message.go:176 +0x128
          github.com/defenseunicorns/zarf/src/pkg/message.Fatalf({0x108e2ca40, 0x1400095c3c0}, {0x1070da294?, 0x0?}, {0x14001ad3b80?, 0x1400056c140?, 0x14000492c30?})
          	/Users/dgershman/Projects/go/src/github.com/defenseunicorns/zarf/src/pkg/message/message.go:183 +0x7c
          github.com/defenseunicorns/zarf/src/cmd.glob..func15(0x10b7089a0?, {0x140005d1d10?, 0x3?, 0x3?})
          	/Users/dgershman/Projects/go/src/github.com/defenseunicorns/zarf/src/cmd/package.go:68 +0x4ac
          github.com/spf13/cobra.(*Command).execute(0x10b7089a0, {0x140005d1cb0, 0x3, 0x3})
          	/Users/dgershman/Projects/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:944 +0x5b0
          github.com/spf13/cobra.(*Command).ExecuteC(0x10b709800)
          	/Users/dgershman/Projects/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1068 +0x35c
          github.com/spf13/cobra.(*Command).Execute(...)
          	/Users/dgershman/Projects/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:992
          github.com/defenseunicorns/zarf/src/cmd.Execute()
          	/Users/dgershman/Projects/go/src/github.com/defenseunicorns/zarf/src/cmd/root.go:70 +0x28
          main.main()
          	/Users/dgershman/Projects/go/src/github.com/defenseunicorns/zarf/main.go:23 +0x9c
```

dgershman:
Here is the yaml snippet from zarf.yaml to reproduce the issue:

```yaml
  - name: argocd-git-chart
    required: true
    charts:
      - name: argocd
        version: argo-cd-5.37.1
        namespace: podinfo-from-git
        url: https://github.com/argoproj/argo-helm.git
        gitPath: charts/argo-cd
    images:
      - quay.io/argoproj/argocd:v2.7.7
      - public.ecr.aws/docker/library/redis:7.0.11-alpine
```

Racer159:
Ah I see - it is actually trying to work but is returning the wrong error - I will push up a fix in a PR shortly.  This worked for the git chart for me:

```yaml
kind: ZarfPackageConfig
metadata:
  name: argocd

components:
  - name: argocd-git-chart
    required: true
    charts:
      - name: argocd
        version: 5.37.1
        namespace: argocd
        url: https://github.com/argoproj/argo-helm.git@argo-cd-5.37.1
        gitPath: charts/argo-cd
    images:
      - quay.io/argoproj/argocd:v2.7.7
      - public.ecr.aws/docker/library/redis:7.0.11-alpine
```

(not the `version` and the `@` at the end of the git repo)

**Note:** I did have to run `helm repo add dandydeveloper https://dandydeveloper.github.io/charts/` before building so that Helm inside Zarf could properly build the dependencies.  We had an error message that was trying to be helpful but this was not being returned correctly here.

dgershman:
It seems strange to me that we’d have to add the helm repo if we are simply using git.  We worked around it, by directly referencing the helm repo, so I’m not blocked by this.  I am at the point where I can automate this for the tests, hadn’t gotten around to it quite yet.  I hope to pick it up soon.

From: Wayne Starr ***@***.***>
Date: Thursday, July 20, 2023 at 1:19 PM
To: defenseunicorns/zarf ***@***.***>
Cc: Danny Gershman ***@***.***>, Author ***@***.***>
Subject: Re: [defenseunicorns/zarf] Draft: Adds support to mutate repoURL in Application CRD and Repositories for ArgoCD (PR #1875)

Ah I see - it is actually trying to work but is returning the wrong error - I will push up a fix in a PR shortly. This worked for the git chart for me:

kind: ZarfPackageConfig

metadata:

  name: argocd



components:

  - name: argocd-git-chart

    required: true

    charts:

      - name: argocd

        version: 5.37.1

        namespace: argocd

        url: ***@***.***

        gitPath: charts/argo-cd

    images:

      - quay.io/argoproj/argocd:v2.7.7

      - public.ecr.aws/docker/library/redis:7.0.11-alpine

(note the version and the @ at the end of the git repo)

Note: I did have to run helm repo add dandydeveloper https://dandydeveloper.github.io/charts/ before building so that Helm inside Zarf could properly build the dependencies. We had an error message that was trying to be helpful but this was not being returned correctly here.

dgershman:
> Overall the code looks good! - just a few small suggested changes.
> 
> Also you mentioned you had some tests setup, can you add a test under `src/test/e2e/22_git_and_flux_test.go` for ArgoCD? (potentially renaming that test file to `22_git_and_gitops_test.go`)
> 
> Also as a note, we have been standardizing around `podinfo` for most tests to keep things fast / lean (since we only have to pull / push a single image like `podinfo:6.4.0` once) but if that is too far out from what you have we can pull in other simple examples for now as well (like `guestbook`). If you want you could put the package as an example under `examples` (if you want to add a README.md to it for Argo) or under `src/test/packages` with a prefix of `22` to just tie it to that test.

Went ahead and built an example for this.  I created a new sample app (as the guestbook example seemed a bit fishy).  The new sample application is called [tracerbullet](https://github.com/radiusmethod/tracerbullet)—essentially it's nginx with a custom message.  Let me know if you'd like to have written permissions for this.  

I have done as you suggested and renamed the test to be more generic to gitops.

examples/argocd/baseline/values.yaml
```diff
@@ -0,0 +1,15 @@
...
+    github-podinfo:
+      name: podinfo
+      url: https://github.com/stefanprodan/podinfo.git
+      username: ""
+      password: ""
+    github-tracerbullet:
+      name: tracerbullet
+      url: https://github.com/radiusmethod/tracerbullet.git
+      username: ""
+      password: ""
```
Racer159, line 15: I would rather keep the tests simpler and not have tracerbullet in it (since, while simple, it would be one differentiator from other tests that use podinfo that might introduce flakes (our tests are pretty tight especially on disk, and additional images or charts could cause issues)

examples/argocd/zarf.yaml
```diff
@@ -0,0 +1,59 @@
...
+    images:
+      - ghcr.io/stefanprodan/podinfo:6.4.0
+      - docker.io/nginx:1.25.1
+    actions:
+      # This prepares the default namespace for the Zarf agent for the sample applications.
+      onDeploy:
+        before:
+          - cmd: "zarf tools kubectl label namespaces default zarf.dev/agent-"
+          - cmd: "zarf tools kubectl delete secret private-registry --ignore-not-found=true -n default"
+          - cmd: "zarf tools kubectl create secret docker-registry private-registry --docker-server=${ZARF_REGISTRY} --docker-username=zarf-pull --docker-password=${ZARF_REGISTRY_AUTH_PULL}"
```
Racer159, line 43: While this looks like it would work is it possible to deploy into a non-default namespace instead?  If you trigger Zarf to make the namespace in a manifest or chart it will add these things automatically in a helm post render.

src/test/e2e/22_git_and_gitops_test.go
```diff
@@ -50,6 +50,39 @@ func TestGitAndFlux(t *testing.T) {
...
+
+	testGitServerConnect(t, tunnel.HTTPEndpoint())
+	testGitServerReadOnly(t, tunnel.HTTPEndpoint())
+	testGitServerTagAndHash(t, tunnel.HTTPEndpoint())
+	waitArgoDeployment(t)
+
+	stdOut, stdErr, err = e2e.Zarf("package", "remove", "argocd", "--confirm")
+	require.NoError(t, err, stdOut, stdErr)
+}
+
```
Racer159, line 85: I don't think this is actually testing argocd (I just see the zarf.yaml at this directory was moved without changes).

Racer159 (changes requested):

examples/argocd/baseline/values.yaml
dgershman: removed

src/test/e2e/22_git_and_gitops_test.go
cmwylie19: I think you could potentially call the Argo App and assert on the repoURLS are mutated here as expected

cmwylie19 (commented):
great work, Danny!

The two changes I think need to be made are:
- Remove the podinfo from the default namespace as per convention of other examples
- [src/test/e2e/22_git_and_gitops_test.go](https://github.com/defenseunicorns/zarf/pull/1875/files#diff-f71e281eb826507b267ab0cb65e6c7664c37155a5e8bb9a57268d684c1221176) needs just a little tweaking to asset that Argo deployed the app and mutated the [repoURLS](https://github.com/defenseunicorns/zarf/pull/1875/files#r1281783699).

examples/argocd/zarf.yaml
```diff
@@ -0,0 +1,51 @@
...
+        url: https://argoproj.github.io/argo-helm
+        releaseName: argocd-baseline
+        valuesFiles:
+          - baseline/values.yaml
+    images:
+      - quay.io/argoproj/argocd:v2.7.7
+      - ghcr.io/dexidp/dex:v2.37.0
+      - public.ecr.aws/docker/library/redis:7.0.11-alpine
+  - name: argocd-apps
+    required: false
```
Racer159, line 22:
```suggestion
    required: true
```
Can we make this a required component in this example?

Racer159 (changes requested):

src/test/e2e/22_git_and_gitops_test.go
dgershman: This makes sense, however, looking at the Flux test, I don't see anywhere it does anything like this as an example.  I can certainly write a test for this, but calling out that I don't see any obvious place where it tests the mutation of a repository URL via flux.

dgershman: Nevermind, I see it now.

dgershman: I take that back.  It looks like `TestGitAndFlux` is doing two things.  The first part tests the mutation of URLs without flux.  The 2nd part is installing Flux and then waiting for the pod to come online and then tearing it down.  I don't see anywhere where it tests the mutation of the Pod Info Git URL.  Theoretically, it checks to see if the PodInfo pod comes online, which isn't necessarily wrong.  And that is what I was copying for ArgoCD.

I think we could probably split Git, Flux, and ArgoCD into their tests.  Flux needs to have a test that checks the URL mutation (and ArgoCD, the same).  

Separately, on the latest `main`, one of the dependencies is causing immediate test failures.  `https://github.com/defenseunicorns/golang-tekton-hello-world.git`

Racer159: Yeah sorry, I highlighted a bit too far down / should have been clearer - we don't need to repeat the git tests for the argo test - these could/should probably be 3 test functions as you mentioned.

Also agreed checking the GitRepository / Application CRDs would be a good add to both tests if you wanted to do that to tighten up the tests.

Are you sure you have the latest main? That repo was added because it is a single branch, no tag repo that on a full repo clone will cause an already up to date error to be returned from go-git during the fetch after clone (related to this line: https://github.com/defenseunicorns/zarf/blob/main/src/internal/packager/git/clone.go#L66

dgershman: I have decoupled these into 3 separate tests.  I have also added some checks to validate that mutation is working for the "custom resources".

src/test/e2e/22_git_and_gitops_test.go
```diff
@@ -17,20 +18,20 @@ import (
...
+func TestGit(t *testing.T) {
+	t.Log("E2E: Git")
 	e2e.SetupWithCluster(t)
 	tmpdir := t.TempDir()
 	cachePath := filepath.Join(tmpdir, ".cache-location")
 
-	buildPath := filepath.Join("src", "test", "packages", "22-git-and-flux")
-	stdOut, stdErr, err := e2e.Zarf("package", "create", buildPath, "--zarf-cache", cachePath, "-o=build", "--confirm")
+	buildPath := filepath.Join("src", "test", "packages", "22-git-data")
+	stdOut, stdErr, err := e2e.Zarf("package", "create", buildPath, "--zarf-cache", cachePath, "-o=build", "--confirm", "-a=arm64")
```
Racer159, line 28:
```suggestion
	stdOut, stdErr, err := e2e.Zarf("package", "create", buildPath, "--zarf-cache", cachePath, "-o=build", "--confirm")
```
The architecture should be removed since the test will use e2e.Arch (and in GitHub that is `amd64`)

src/test/e2e/22_git_and_gitops_test.go
```diff
@@ -108,7 +122,38 @@ func waitFluxPodInfoDeployment(t *testing.T) {
 	stdOut, stdErr, err := e2e.Zarf("package", "deploy", path, "--confirm")
 	require.NoError(t, err, stdOut, stdErr)
 
+	// Tests the URL mutation for GitRepository CRD for Flux.
+	stdOut, stdErr, err = e2e.Kubectl("get", "gitrepositories", "podinfo", "-n", "flux-system", "-o", "jsonpath={.spec.url}")
+	expectedMutatedRepoUrl := fmt.Sprintf("%s/%s/podinfo-1646971829.git", config.ZarfInClusterGitServiceURL, config.ZarfGitPushUser)
+	require.Equal(t, expectedMutatedRepoUrl, stdOut)
```
Racer159, line 128:
```suggestion
	expectedMutatedRepoURL := fmt.Sprintf("%s/%s/podinfo-1646971829.git", config.ZarfInClusterGitServiceURL, config.ZarfGitPushUser)
	require.Equal(t, expectedMutatedRepoURL, stdOut)
```
linting

src/test/e2e/22_git_and_gitops_test.go
```diff
@@ -108,7 +122,38 @@ func waitFluxPodInfoDeployment(t *testing.T) {
...
 	require.NoError(t, err, stdOut, stdErr)
 }
+
+func waitArgoDeployment(t *testing.T) {
+	// Deploy the argocd example and verify that it works
+	path := fmt.Sprintf("build/zarf-package-argocd-%s.tar.zst", e2e.Arch)
+	stdOut, stdErr, err := e2e.Zarf("package", "deploy", path, "--components=argocd-apps", "--confirm")
+	require.NoError(t, err, stdOut, stdErr)
+
+	expectedMutatedRepoUrl := fmt.Sprintf("%s/%s/podinfo-1646971829.git", config.ZarfInClusterGitServiceURL, config.ZarfGitPushUser)
```
Racer159, line 141:
```suggestion
	expectedMutatedRepoURL := fmt.Sprintf("%s/%s/podinfo-1646971829.git", config.ZarfInClusterGitServiceURL, config.ZarfGitPushUser)
```

src/test/e2e/22_git_and_gitops_test.go
```diff
@@ -108,7 +122,38 @@ func waitFluxPodInfoDeployment(t *testing.T) {
...
+
+	// Tests the mutation of the private repository Secret for ArgoCD.
+	stdOut, stdErr, err = e2e.Kubectl("get", "secret", "argocd-repo-github-podinfo", "-n", "argocd", "-o", "jsonpath={.data.url}")
+	require.NoError(t, err, stdOut, stdErr)
+
+	expectedMutatedPrivateRepoUrlSecret, err := base64.StdEncoding.DecodeString(stdOut)
```
Racer159, line 147:
```suggestion
	expectedMutatedPrivateRepoURLSecret, err := base64.StdEncoding.DecodeString(stdOut)
```

src/test/e2e/22_git_and_gitops_test.go
```diff
@@ -108,7 +122,38 @@ func waitFluxPodInfoDeployment(t *testing.T) {
...
+	require.NoError(t, err, stdOut, stdErr)
+	require.Equal(t, expectedMutatedRepoUrl, string(expectedMutatedPrivateRepoUrlSecret))
```
Racer159, line 149:
```suggestion
	require.Equal(t, expectedMutatedRepoURL, string(expectedMutatedPrivateRepoURLSecret))
```

src/test/e2e/22_git_and_gitops_test.go
```diff
@@ -108,7 +122,38 @@ func waitFluxPodInfoDeployment(t *testing.T) {
...
+
+	// Tests the mutation of the repoURL for Application CRD source(s) for ArgoCD.
+	stdOut, stdErr, err = e2e.Kubectl("get", "application", "apps", "-n", "argocd", "-o", "jsonpath={.spec.sources[0].repoURL}")
+	require.NoError(t, err, stdOut, stdErr)
+	require.Equal(t, expectedMutatedRepoUrl, stdOut)
```
Racer159, line 154:
```suggestion
	require.Equal(t, expectedMutatedRepoURL, stdOut)
```

src/internal/agent/hooks/argocd-repository.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/defenseunicorns/zarf/src/internal/agent/operations"
+	"github.com/defenseunicorns/zarf/src/internal/agent/state"
+	"github.com/defenseunicorns/zarf/src/pkg/message"
+	"github.com/defenseunicorns/zarf/src/pkg/transform"
+	"github.com/defenseunicorns/zarf/src/pkg/utils/helpers"
+	"github.com/defenseunicorns/zarf/src/types"
+	v1 "k8s.io/api/admission/v1"
+)
+
+type ArgoRepository struct {
```
Racer159, line 22:
```suggestion
// ArgoRepository represents a subset of the Argo Repository object needed for Zarf Git URL mutations
type ArgoRepository struct {
```
lint

src/internal/agent/hooks/argocd-application.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/defenseunicorns/zarf/src/pkg/utils/helpers"
+	"github.com/defenseunicorns/zarf/src/types"
+	v1 "k8s.io/api/admission/v1"
+)
+
+type Source struct {
+	RepoURL string `json:"repoURL"`
+}
+
+type ArgoApplication struct {
```
Racer159, line 24:
```suggestion
// ArgoApplication represents a subset of the Argo Application object needed for Zarf Git URL mutations
type ArgoApplication struct {
```

src/internal/agent/hooks/argocd-application.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/defenseunicorns/zarf/src/internal/agent/operations"
+	"github.com/defenseunicorns/zarf/src/internal/agent/state"
+	"github.com/defenseunicorns/zarf/src/pkg/message"
+	"github.com/defenseunicorns/zarf/src/pkg/transform"
+	"github.com/defenseunicorns/zarf/src/pkg/utils/helpers"
+	"github.com/defenseunicorns/zarf/src/types"
+	v1 "k8s.io/api/admission/v1"
+)
+
+type Source struct {
```
Racer159, line 20:
```suggestion
// Source represents a subset of the Argo Source object needed for Zarf Git URL mutations
type Source struct {
```

src/internal/agent/hooks/argocd-repository.go
```diff
@@ -0,0 +1,107 @@
...
+// NewRepositoryMutationHook creates a new instance of the ArgoCD Repository mutation hook.
+func NewRepositoryMutationHook() operations.Hook {
+	message.Debug("hooks.NewRepositoryMutationHook()")
+	return operations.Hook{
+		Create: mutateRepository,
+		Update: mutateRepository,
+	}
+}
+
+// mutateGitRepoCreate mutates the git repository url to point to the repository URL defined in the ZarfState.
```
Racer159, line 37:
```suggestion
// mutateRepository mutates the git repository URL to point to the repository URL defined in the ZarfState.
```

src/internal/agent/hooks/argocd-application.go
```diff
@@ -0,0 +1,126 @@
...
+// NewApplicationMutationHook creates a new instance of the ArgoCD Application mutation hook.
+func NewApplicationMutationHook() operations.Hook {
+	message.Debug("hooks.NewApplicationMutationHook()")
+	return operations.Hook{
+		Create: mutateApplication,
+		Update: mutateApplication,
+	}
+}
+
+// mutateGitRepoCreate mutates the git repository url to point to the repository URL defined in the ZarfState.
```
Racer159, line 48:
```suggestion
// mutateApplication mutates the git repository url to point to the repository URL defined in the ZarfState.
```

Racer159 (changes requested):
Looks super close! Only a few linting errors and there is a hardcoded arch causing test failures in GitHub.

src/test/e2e/22_git_and_gitops_test.go
dgershman: ah yes, left over from testing, my bad.

dgershman:
All linting suggestions implemented.  Ran `make go-lint` to verify.

src/test/e2e/22_git_and_gitops_test.go
Racer159: This one still is here (won't show on linting since it is not an instantiation of the var)

Racer159 (changes requested):

Racer159:
Looks like everything worked except for what I feared about storage space - worse comes to worse we can split the build and validate stages across all of our tests (like what Upgrade does) but one thing to try first would be removing images using the new `zarf tools registry prune` command.  I put an example on `main` for what that would look like with the `podinfo` flux test (the current merge conflict).

If that doesn't work I'll make another PR to split all of the workflows (will just mean that it will take a few more minutes for us).

dgershman:
> Looks like everything worked except for what I feared about storage space - worse comes to worse we can split the build and validate stages across all of our tests (like what Upgrade does) but one thing to try first would be removing images using the new `zarf tools registry prune` command. I put an example on `main` for what that would look like with the `podinfo` flux test (the current merge conflict).
> 
> If that doesn't work I'll make another PR to split all of the workflows (will just mean that it will take a few more minutes for us).

Nice command.  Done.

dgershman:
I have also disabled `dex` (not required) and there was a left over image (`nginx`), that isn't needed anymore (for tracerbullet).

Racer159 (approved):
lgtm

cmwylie19 (approved):
lgtm

Issue #1882, opened 2023-07-05
Improve interaction with Zarf messaging for library users


Racer159:
### Describe what should be investigated or refactored
Currently our message library has the following shortcomings for library users:

1. it is relatively large and pulls in more than it should - it is also used in other `utils` and `pkg` code where it should use a logger interface instead (like the `k8s` package) - this would help library users keep their code small and light.
2. it prints more than it should impacting the UX of other CLIs / programs that use Zarf as a library.

### Links to any relevant code
![image](https://github.com/defenseunicorns/zarf/assets/3977569/bcbf31a5-d2af-41a0-b131-c49f38887f02)

### Additional context
This would help with other projects using Zarf like compiling Zarf as a library in WASM modules for Pepr

PR #1883, opened 2023-07-05
Reduce utils size

cmwylie19:
## Description

This reduces the size of the dependencies in the transform lib by over ~45Mb.

## Related Issue

Fixes #

Relates to #1882

## Type of change

## Checklist before merging


src/pkg/utils/io.go
```diff
@@ -27,6 +29,26 @@ const (
 	tmpPathPrefix = "zarf-"
 )
 
+func GetCryptoHashFromFile(path string, hashName crypto.Hash) (string, error) {
```
Racer159, line 32: Lint

Racer159 (changes requested):
one little lint issue, but otherwise looks fine

Racer159 (approved):
lgtm!

Issue #1766, opened 2023-06-01
Git Tag and Chart Version assumed to match


runyontr:
### Steps to reproduce
Create a zarf.yaml with this component:

```yaml
  - name: dcgm-exporter
    required: true
    charts:
    - name: dcgm-exporter
      version: 3.1.7-3.1.4 # <- this tag doesn't match the chart version
      url: https://github.com/NVIDIA/dcgm-exporter.git
      gitPath: deployment
      namespace: "leapfrogai"
    images:
      - "nvcr.io/nvidia/k8s/dcgm-exporter:3.1.7-3.1.4-ubuntu20.04"
   ```

This particular application has a mismatch between the git tag and the chart version.  This causes an error in either one of two ways since those two are assumed to match:

1. Build error: If using the version matching the chart version, the package doesn't build since Zarf can't find the Git Tag
2. Deploy Error:  If using the git tag, zarf can't deploy the chart since it can't find a chart with the provided version

### Expected result

Be able to deploy an arbitrary chart version from a git repository

### Additional Context
Add any other context or screenshots about the technical debt here.

Ideas/for implementation:

1. Add new property to the `charts` that allows for setting `gitVersion` separately from `version`, but have the default `gitVersion` match the `version` field
2. Add more robust regex on the `url` to allow for passing branch/tag versions in the URL.  Some ambiguity here on backwards compatibility.  If nothing is set, does it default to the default branch, or the `version`?
3. Don't have zarf check to ensure the version matches when deploying a chart from git.

PR #1886, opened 2023-07-07
[TASK] TransformGitURL helm helper

cmwylie19:
## Description

What we could do (to keep backwards compat in tact and address https://github.com/defenseunicorns/zarf/issues/1766) is see if there is an @ ref on the git URL and if there is, leave it on there, and if not append the @tag as we do today. Dedup some code in src/pkg/packager/prepare.go, and src/pkg/packager/create.go.
## Related Issue

Fixes #

Relates to #1766

## Type of change

## Checklist before merging

### DIFF
--- a/src/internal/packager/helm/repo.go
+++ b/src/internal/packager/helm/repo.go
@@ -12,6 +12,7 @@ import (
 	"github.com/defenseunicorns/zarf/src/config"
 	"github.com/defenseunicorns/zarf/src/internal/packager/git"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
+	"github.com/defenseunicorns/zarf/src/pkg/transform"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/registry"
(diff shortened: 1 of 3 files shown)


src/pkg/packager/create.go
```diff
@@ -326,7 +326,9 @@ func (p *Packager) addComponent(index int, component types.ZarfComponent, isSkel
 
 	// If any helm charts are defined, process them.
 	for chartIdx, chart := range component.Charts {
-		re := regexp.MustCompile(`\.git$`)
+		// https://regex101.com/r/jYLoUy/1
+		re := regexp.MustCompile(`\.git@`)
```
YrrepNoj, line 330: Should the `@` be an optional? How were we getting away with not having it before?

src/internal/packager/helm/repo.go
```diff
@@ -165,3 +166,18 @@ func (h *Helm) DownloadChartFromGitToTemp(spinner *message.Spinner) (string, err
 
 	return gitCfg.GitPath, nil
 }
+
+// Helper function src/pkg/packager/prepare.goL118 src/pkg/packager/create.goL328
+func (h *Helm) TransformGitURL(charts string) (string, error) {
```
YrrepNoj, line 171: Can we rename this function? Since it's calling `h.PackageChartFromGit(...)` it is doing a lot more than just transforming the git URL.

src/pkg/packager/create.go
cmwylie19: this was resolved in https://github.com/defenseunicorns/zarf/pull/1892

src/internal/packager/helm/repo.go
cmwylie19: https://github.com/defenseunicorns/zarf/pull/1892 (We deleted this function). Thanks for the feedback!

PR #1890, opened 2023-07-07
Add ExitOnInterrupt utils function; catch interrupts when using wait

### LINKED ISSUE #1864: Cannot stop process when in the middle of a `wait`
When a Zarf package has a `wait` target described, the CLI does not receive `ctrl + c` commands to stop the process. The default timeout for a `wait` is five minutes meaning a user who wants to stop the deployment is unable to cancel the process during those five minutes.

Example `wait` specification:
```yaml
components:
  - name: wait-component
     actions:
      onDeploy:
        after:
          # This will use a wait action to wait for the podinfo pod to be ready
          - description: Podinfo pods to be ready via wait action
            wait:
              cluster:
                kind: pod
                name: app=podinfo
                namespace: podinfo
                condition: ready
```

caesarshift:
Add ExitOnInterrupt utils function; catch interrupts when using wait

## Description

Add ability to catch interrupts when using wait command.

## Related Issue

Fixes #1864

## Type of change

## Checklist before merging

### DIFF
--- a/src/cmd/tools/wait.go
+++ b/src/cmd/tools/wait.go
@@ -29,6 +29,7 @@ var waitForCmd = &cobra.Command{
 	Example: lang.CmdToolsWaitForExample,
 	Args:    cobra.MinimumNArgs(2),
 	Run: func(cmd *cobra.Command, args []string) {
+		exec.ExitOnInterrupt(1, lang.ErrInterrupt)
 		// Parse the timeout string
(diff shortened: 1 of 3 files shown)


caesarshift:
The ticket (and this PR) is specific to the `wait` command, but this same approach could be applied to other commands as well (e.g., `zarf connect registry` has the same issue)

caesarshift:
Branch is misnamed. Closing/reopening under corrected name.

Issue #1518, opened 2023-03-30
Graceful shutdown to clean up canceled deployment


vanakema:
### Is your feature request related to a problem? Please describe.
I encountered an issue when using Zarf to deploy a failing application. During the deployment, I used Ctrl-C to interrupt the process, but this did not allow Zarf to clean up the state it put in the cluster. As a result, I am forced to wait for the deployment to fail multiple times before Zarf can properly clean up the failed deployment, which can take a significant amount of time (up to 15 minutes per timeout).

### Describe the solution you'd like
I propose implementing a graceful shutdown signal when using Ctrl-C during deployment. This will allow Zarf to clean up the cluster state if the deployment is interrupted. Additionally, if Ctrl-C is used during the graceful shutdown, it should trigger a hard stop like it currently does.

### Describe alternatives you've considered
An alternative solution would be to provide a separate command to manually trigger a cleanup of a failed deployment in the cluster. This would allow users to clean up the state without waiting for multiple deployment timeouts. The current `zarf package delete` command fails to clean up a failed deploy due to some secret not existing that it expects.

### Additional context
This improvement would save users time and frustration when dealing with failed deployments in Zarf. By implementing a graceful shutdown signal upon using Ctrl-C, users can have more control over their deployment process and avoid waiting for extended periods due to multiple timeouts.

AustinAbro321:
This was addressed by #3031

PR #1891, opened 2023-07-07
Add ExitOnInterrupt utils function; catch interrupts when using wait

### LINKED ISSUE #1864: Cannot stop process when in the middle of a `wait`

caesarshift:
Add ExitOnInterrupt utils function; catch interrupts when using wait

## Description

Add ability to catch interrupts when using wait command.

## Related Issue

Fixes #1864

## Type of change

## Checklist before merging

### DIFF
--- a/src/cmd/tools/wait.go
+++ b/src/cmd/tools/wait.go
@@ -10,6 +10,7 @@ import (
 	"github.com/defenseunicorns/zarf/src/config/lang"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 	"github.com/defenseunicorns/zarf/src/pkg/utils"
+	"github.com/defenseunicorns/zarf/src/pkg/utils/exec"
 	"github.com/spf13/cobra"
 
 	// Import to initialize client auth plugins.
@@ -29,6 +30,7 @@ var waitForCmd = &cobra.Command{
 	Example: lang.CmdToolsWaitForExample,
 	Args:    cobra.MinimumNArgs(2),
 	Run: func(cmd *cobra.Command, args []string) {
+		exec.ExitOnInterrupt(1, lang.ErrInterrupt)
 		// Parse the timeout string
 		timeout, err := time.ParseDuration(waitTimeout)
 		if err != nil {
--- a/src/cmd/tools/wait.go
+++ b/src/cmd/tools/wait.go
@@ -10,7 +10,6 @@ import (
 	"github.com/defenseunicorns/zarf/src/config/lang"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 	"github.com/defenseunicorns/zarf/src/pkg/utils"
-	"github.com/defenseunicorns/zarf/src/pkg/utils/exec"
 	"github.com/spf13/cobra"
 
 	// Import to initialize client auth plugins.
@@ -30,7 +29,6 @@ var waitForCmd = &cobra.Command{
 	Example: lang.CmdToolsWaitForExample,
 	Args:    cobra.MinimumNArgs(2),
 	Run: func(cmd *cobra.Command, args []string) {
-		exec.ExitOnInterrupt(1, lang.ErrInterrupt)
 		// Parse the timeout string
 		timeout, err := time.ParseDuration(waitTimeout)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


caesarshift:
The ticket (and this PR) is specific to the `wait` command, but this same approach could be applied to other commands as well (e.g., `zarf connect registry` has the same issue)

src/cmd/tools/wait.go
```diff
@@ -29,6 +30,7 @@ var waitForCmd = &cobra.Command{
 	Example: lang.CmdToolsWaitForExample,
 	Args:    cobra.MinimumNArgs(2),
 	Run: func(cmd *cobra.Command, args []string) {
+		exec.ExitOnInterrupt(1, lang.ErrInterrupt)
```
Racer159, line 33: Can we make this call in the `rootCmd` `PersistentPreRun` to handle all exits for Zarf commands (after the check for vendored commands)?

https://github.com/defenseunicorns/zarf/blob/4d4ae8fe077b03ef06f2674cdcb5d310c04bb125/src/cmd/root.go#L39

That would allow it to work for this command and be a good starting point for https://github.com/defenseunicorns/zarf/issues/1518

src/pkg/utils/exec/exec.go
```diff
@@ -207,3 +209,17 @@ func GetOSShell(shellPref types.ZarfComponentActionShell) (string, string) {
...
+// ExitOnInterrupt catches an interrupt and exits with ErrorCode
+func ExitOnInterrupt(ErrorCode int, ErrorMessage string) chan os.Signal {
+	c := make(chan os.Signal, 2)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		<-c
+		if ErrorMessage != "" {
+			message.Warnf(ErrorMessage)
+		}
+		os.Exit(ErrorCode)
```
Racer159, line 222:
```suggestion
		message.Fatal(lang.ErrInterrupt, lang.ErrInterrupt.Error())
```

This might be simpler for now and would show be more consistent with other interrupts (like when a confirmation is canceled with an interrupt).

This changes Warnf to Fatal to print a red error rather than a warning and remove formatting (since that can cause garbled text to show if the text contains formatting shortcodes).  It also changes `lang.ErrInterrupt` to be a reusable error rather than a message (see the bottom of `lang/english.go`)

image

src/pkg/utils/exec/exec.go
```diff
@@ -207,3 +209,17 @@ func GetOSShell(shellPref types.ZarfComponentActionShell) (string, string) {
 func IsPowershell(shellName string) bool {
 	return shellName == "powershell" || shellName == "pwsh"
 }
+
+// ExitOnInterrupt catches an interrupt and exits with ErrorCode
+func ExitOnInterrupt(ErrorCode int, ErrorMessage string) chan os.Signal {
```
Racer159, line 214:
```suggestion
func ExitOnInterrupt() chan os.Signal {
```
Per the below this likely can be simplified to just use a reusable error

src/config/lang/english.go
```diff
@@ -27,6 +27,7 @@ const (
 	ErrCreatingDir         = "failed to create directory %s: %s"
 	ErrRemoveFile          = "failed to remove file %s: %s"
 	ErrUnarchive           = "failed to unarchive %s: %s"
+	ErrInterrupt           = "Failed due to interrupt"
```
Racer159, line 30: Per the below this may be better as a reusable error.

Racer159 (changes requested):
Thanks for the PR! Just a few small change but otherwise looks good.

caesarshift:
Thank you for the feedback. Beyond your changes, I've pushed the following additional changes:
- Minor cleanup to the `ExitOnInterrupt` function: removed unused return variable; changed size of channel to 1
- Added the `exec.ExitOnInterrupt` to the `PersistentPreRun` in `rootCmd`, `toolsCmd`, and `registryCmd`. Adding it to the `toolsCmd` and `registryCmd` was required because they define their own `PersistentPreRun` functions that override the rootCmd's function.

YrrepNoj (approved):

PR #1892, opened 2023-07-07
1766 - Git Tag and Chart Version assumed to match

### LINKED ISSUE #1766: Git Tag and Chart Version assumed to match

cmwylie19:
## Description

What we could do (to keep backwards compat in tact and address https://github.com/defenseunicorns/zarf/issues/1766) is see if there is an @ ref on the git URL and if there is, leave it on there, and if not append the @tag as we do today. Dedup some code in src/pkg/packager/prepare.go, and src/pkg/packager/create.go.

## Related Issue

Fixes #1766

Relates to 

## Type of change

## Checklist before merging

### DIFF
--- a/src/internal/packager/helm/repo.go
+++ b/src/internal/packager/helm/repo.go
@@ -34,7 +34,7 @@ func (h *Helm) PackageChart(destination string) error {
 		}
 
 		if isGitURL {
-			// if it is append chart version as if its a tag
+			// if it is git url append chart version as if its a tag
 			if refPlain == "" {
 				h.Chart.URL = fmt.Sprintf("%s@%s", h.Chart.URL, h.Chart.Version)
 			}
@@ -68,7 +68,7 @@ func (h *Helm) PackageChartFromLocalFiles(destination string) (string, error) {
 		spinner.Errorf(err, "Validation failed for chart from %s (%s)", h.Chart.LocalPath, err.Error())
 		return "", err
 	}
-
+	h.buildChartDependencies(spinner)
 	client := action.NewPackage()
 
 	client.Destination = destination
@@ -180,4 +180,34 @@ func (h *Helm) DownloadChartFromGitToTemp(spinner *message.Spinner) (string, err
 	}
 
 	return gitCfg.GitPath, nil
+}
+
+// buildChartDependencies builds the helm chart dependencies
+func (h *Helm) buildChartDependencies(spinner *message.Spinner) error {
+	regClient, err := registry.NewClient(registry.ClientOptEnableCache(true))
+	if err != nil {
+		spinner.Fatalf(err, "Unable to create a new registry client")
+	}
+
+	settings := cli.New()
+
+	man := &downloader.Manager{
+		Out:            os.Stdout,
+		ChartPath:      h.Chart.LocalPath,
+		Getters:        getter.All(settings),
+		RegistryClient: regClient,
+
+		RepositoryConfig: settings.RepositoryConfig,
+		RepositoryCache:  settings.RepositoryCache,
+		Debug:            false,
+	}
+	// Verify the chart
+	man.Verify = downloader.VerifyIfPossible
+
+	// Build the deps from the helm chart
+	err = man.Build()
+	if e, ok := err.(downloader.ErrRepoNotFound); ok {
+		return fmt.Errorf("%s. Please add the missing repos via 'helm repo add'", e.Error())
+	}
+	return nil
 }
--- a/src/internal/packager/helm/repo.go
+++ b/src/internal/packager/helm/repo.go
@@ -188,17 +188,15 @@ func (h *Helm) buildChartDependencies(spinner *message.Spinner) error {
 	if err != nil {
 		spinner.Fatalf(err, "Unable to create a new registry client")
 	}
-
-	settings := cli.New()
-
+	h.Settings = cli.New()
 	man := &downloader.Manager{
 		Out:            os.Stdout,
 		ChartPath:      h.Chart.LocalPath,
-		Getters:        getter.All(settings),
+		Getters:        getter.All(h.Settings),
 		RegistryClient: regClient,
 
-		RepositoryConfig: settings.RepositoryConfig,
-		RepositoryCache:  settings.RepositoryCache,
+		RepositoryConfig: h.Settings.RepositoryConfig,
+		RepositoryCache:  h.Settings.RepositoryCache,
 		Debug:            false,
 	}
 	// Verify the chart
--- a/src/internal/packager/helm/repo.go
+++ b/src/internal/packager/helm/repo.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/defenseunicorns/zarf/src/config"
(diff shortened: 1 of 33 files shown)


cmwylie19:
I am removing the `needs-tests` label because I am going to address the tests in the [metrics pr](https://github.com/defenseunicorns/zarf/pull/1853) as this current PR is a prereq along with the [named port pr](https://github.com/defenseunicorns/zarf/pull/1894) and the waitFor pr which has been merged.

src/internal/packager/helm/repo.go
```diff
@@ -22,65 +24,81 @@ import (
 	"helm.sh/helm/v3/pkg/repo"
 )
 
+func (h *Helm) PackageChart(destination string) error {
```
Racer159, line 27: lint

src/internal/packager/helm/repo.go
```diff
@@ -22,65 +24,81 @@ import (
...
+	if len(h.Chart.URL) > 0 {
+		url, refPlain, err := transform.GitTransformURLSplitRef(h.Chart.URL)
+		// check if the chart is a git url with a ref (if an error is returned url will be empty)
+		isGitURL := strings.HasSuffix(url, ".git")
+		if err != nil {
+			message.Debugf("unable to parse the url, continuing with %s", h.Chart.URL)
+		}
+
+		if isGitURL {
+			// if it is git url append chart version as if its a tag
```
Racer159, line 37:
```suggestion
			// if it is a git url append chart version as if its a tag
```

Racer159 (changes requested):

src/internal/packager/helm/repo.go
cmwylie19: thanks, done!

src/internal/packager/helm/repo.go
```diff
@@ -22,65 +24,82 @@ import (
 	"helm.sh/helm/v3/pkg/repo"
 )
 
+// PackageChart creates a chart archive from a path to a chart on the host os and builds chart dependencies
+func (h *Helm) PackageChart(destination string) error {
+	if len(h.Chart.URL) > 0 {
+		url, refPlain, err := transform.GitTransformURLSplitRef(h.Chart.URL)
```
Noxsios, line 30: Generally go-convention is to not have the package name in any public function names (meh)

src/pkg/transform/git.go
```diff
@@ -12,8 +12,8 @@ import (
 	"github.com/defenseunicorns/zarf/src/pkg/utils/helpers"
 )
 
-// For further explanation: https://regex101.com/r/YxpfhC/3
-var gitURLRegex = regexp.MustCompile(`^(?P<proto>[a-z]+:\/\/)(?P<hostPath>.+?)\/(?P<repo>[\w\-\.]+?)(?P<git>\.git)?(?P<atRef>@(?P<force>\+)?(?P<ref>[\/\+\w\-\.]+))?(?P<gitPath>\/(?P<gitPathId>info\/.*|git-upload-pack|git-receive-pack))?$`)
+// For further explanation: https://regex101.com/r/YxpfhC/5
+var gitURLRegex = regexp.MustCompile(`^(?P<proto>[a-z]+:\/\/)(?P<hostPath>.+?)\/(?P<repo>[\w\-\.]+?)?(?P<git>\.git)?(\/)?(?P<atRef>@(?P<force>\+)?(?P<ref>[\/\+\w\-\.]+))?(?P<gitPath>\/(?P<gitPathId>info\/.*|git-upload-pack|git-receive-pack))?$`)
```
Noxsios, line 16: Do we have this tested somewhere in a unit test?

cmwylie19: more test-cases added. LMK if this is what you had in mind. Thanks!

src/internal/packager/helm/repo.go
cmwylie19: renamed

src/pkg/transform/git.go
cmwylie19: Did you want a plan regex case or more [here](https://github.com/defenseunicorns/zarf/blob/0d0cac8496e397ef63adcc276956b4fcdd50811f/src/pkg/transform/git_test.go#L74)?

src/pkg/transform/git_test.go
```diff
@@ -87,6 +88,7 @@ func TestGitTransformURLSplitRef(t *testing.T) {
 		{"https://github.com/defenseunicorns/zarf.helm.git", ""},
 		{"https://github.com/defenseunicorns/zarf.git", "refs/tags/v0.16.0"},
 		{"https://github.com/DoD-Platform-One/big-bang.git", "refs/heads/release-1.54.x"},
+		{"https://github.com/prometheus-community/helm-charts.git", "kube-prometheus-stack-47.3.0"},
```
Racer159, line 91: The two test cases we would need to cover were:

Trailing slash after a path
```
https://github.com/prometheus-community/
```

Trailing slash with no path
```
https://github.com/
```

src/pkg/transform/git.go
```diff
@@ -29,8 +29,8 @@ func MutateGitURLsInText(logger Log, targetBaseURL string, text string, pushUser
 	return output
 }
 
-// GitTransformURLSplitRef takes a git url and returns a separated source url and zarf reference.
-func GitTransformURLSplitRef(sourceURL string) (string, string, error) {
+// GitURLSplitRef takes a git url and returns a separated source url and zarf reference.
+func GitURLSplitRef(sourceURL string) (string, string, error) {
```
Racer159, line 33: If we do this rename we should do it on all `GitTransform*` functions.

Racer159 (changes requested):

Noxsios (dismissed):

src/pkg/transform/git_test.go
cmwylie19: misinterpreted this somehow, now should be addressed. lmk

Racer159 (approved):
lgtm

PR #1896, opened 2023-07-10
Fix IB registry injector failure

### LINKED ISSUE #1895: Injector fails to select a pod when using images other than `library/registry`
### Environment
Device and OS: Dell T1700 - Arch Linux 
App version: v0.28.0
Other:

### Steps to reproduce
1. Create an init package with the registry v2 image from Iron Bank: https://registry1.dso.mil/harbor/projects/3/repositories/opensource%2Fdocker%2Fregistry-v2/artifacts-tab
2. Run `zarf package init`

### Expected result
Zarf successfully initializes

### Actual Result
Zarf fails to select a pod to initialize from

### Additional Context
This is due to an issue with how the injector handles serving images that it loads from configmaps

Racer159:
## Description

This fixes an injector issue where the registry-v2 image was not able to be loaded via the injector due to the way it handled namespaces - this PR makes the injector much more flexible for different images.

## Related Issue

Fixes #1895

## Type of change

## Checklist before merging


src/injector/src/main.rs
```diff
@@ -105,43 +105,111 @@ fn start_seed_registry() {
...
+
+                (GET) (/v2/{n1 :String}/{n2 :String}/{n3 :String}/{name :String}/manifests/{tag :String}) => {
+                    let namespaced_name :String = n1 + "/" + &n2 + "/" + &n3 + "/" + &name;
+                    handle_get_manifest(&root, &namespaced_name, &tag)
+                },
+
+                (GET) (/v2/{n1 :String}/{n2 :String}/{n3 :String}/{n4 :String}/{name :String}/manifests/{tag :String}) => {
+                    let namespaced_name :String = n1 + "/" + &n2 + "/" + &n3 + "/" + &n4 + "/" + &name;
+                    handle_get_manifest(&root, &namespaced_name, &tag)
+                },
```
Noxsios, line 130: I hate this, but it will work for most namespaced containers.

Noxsios: A potentially better pattern would be to use the `!find_route` macro https://docs.rs/rouille/latest/rouille/macro.find_route.html and create our own custom router so that we could handle arbitrary path segments / wildcards on our own.

src/injector/src/main.rs
```diff
@@ -105,43 +105,111 @@ fn start_seed_registry() {
...
+                    )
+                },
+
+                (HEAD) (/v2/{namespace :String}/{name :String}/manifests/{tag :String}) => {
+                    let namespaced_name :String = namespace + "/" + &name;
+                    // a normal HEAD response has an empty body, but due to rouille not allowing for an override
+                    // on Content-Length, we respond the same as a GET
+                    accept!(
+                        request,
+                        "application/vnd.docker.distribution.manifest.v2+json" => {
```
Noxsios, line 150: string constants like this should prob be top level constants

src/injector/src/main.rs
```diff
@@ -154,23 +222,39 @@ fn start_seed_registry() {
...
+                    .to_owned();
+            }
+        }
+    }
+
+    if sha_manifest != "" {
+        let file = File::open(&root.join("blobs").join("sha256").join(&sha_manifest)).unwrap();
+        Response::from_file("application/vnd.docker.distribution.manifest.v2+json", file)
+            .with_additional_header(
+                "Docker-Content-Digest",
```
Noxsios, line 250: string headers like this that are common / the same across requests should prob be top level constants

.github/actions/packages/action.yaml
```diff
@@ -1,9 +1,19 @@
...
   using: composite
   steps:
     - run: |
-        make build-cli-linux-amd init-package build-examples ARCH=amd64
+        make build-cli-linux-amd ARCH=amd64
+      shell: bash
+    - run: |
+        make init-package build-examples ARCH=amd64
       shell: bash
+      if: ${{ inputs.build-packages == 'true' }}
```
Noxsios, line 19: finally

src/injector/src/main.rs
Noxsios, line 1: I humbly request that we add an E2E test for the injector now that we are supporting arbitrary registry images

https://github.com/defenseunicorns/static-registry-rs/blob/433f8518d11d6ef99020560b5e93a7898ed5c2fe/src/main.rs#L209

Noxsios (changes requested):

src/injector/src/main.rs
Racer159: Lol I hate it too - I'll give find_route a try though

src/injector/src/main.rs
```diff
@@ -100,77 +102,91 @@ fn start_seed_registry() {
...
+                );
+            }
+        } else if object_type == "blobs" && request.method() == "GET" {
+            let digest = url_segments[tag_index].to_owned();
+            return handle_get_digest(&root, &digest);
+        }
+    }
+
+    Response::empty_404()
+}
```
Noxsios, line 154: Can you add comments that specify which code is handling which route ie:

```
// this route handles (GET) /v2/**/manifests/<reference>
```

Noxsios (changes requested):

src/injector/src/main.rs
Racer159: Resolved

src/injector/src/main.rs
Racer159: done

Noxsios:
note: to use a locally built version of the injector in the `init` package, do something like this in `packages/zarf-registry/zarf.yaml`:

```yaml
    files:
      # Rust Injector Binary
      - source: ../../src/injector/target/x86_64-unknown-linux-musl/release/zarf-injector
        target: "###ZARF_TEMP###/zarf-injector"
        executable: true

Noxsios (approved):
lgtm 🦀

Issue #1632, opened 2023-04-20
Replace Zarf Agent with Pepr


Racer159:
### Describe what should be investigated or refactored
We should replace the Zarf Agent with Pepr so that we have more capabilities for mutating cluster objects in the airgap.

Ideally we wouldn't lose much functionally in this transition, and to that end would like to see:
- Pepr's container be around the same size as the Zarf Agent container (or smaller)
- Pepr be available in Iron Bank or at least on track to get into Iron Bank

This should likely take one of two paths:

1. If Pepr quickly meets the above we could just swap it in place and most people would keep on trucking (there may be some who do not have node apps in there environments though and that could cause consternation)

2. If Pepr cannot meet the above but we still want to get feedback, we could implement it as an optional part of the init package - not fully replacing the zarf agent, but instead living alongside it as a choice you can make.  As it becomes more mature it could become the default option and then fully replace the Zarf Agent

### Links to any relevant code
https://github.com/defenseunicorns/zarf/blob/main/Dockerfile
https://github.com/defenseunicorns/pepr

### Additional context
Personally #2 seems more reasonable and would allow us to get feedback faster without needing to re-implement all of the Zarf agent *ilities right away.  The init package is also much more modular than it has been in the past and so providing the option should be reasonably easy now and can be made even easier over time (a lot of that was already on our roadmap).

Racer159:
@jeff-mccoy @bdw617 @YrrepNoj @Noxsios @mike-winberry for thoughts on the above.

jeff-mccoy:
I prefer #2

Racer159:
We should focus on core Zarf Agent logic (Mutating Web-hook) for now - leaving the proxy in the Zarf Agent for now and spinning off a separate issue / ADR cycle for that.

The hardest thing will be keeping logic in sync between the CLI and the Pepr Capability (exploring a WASM module may be useful here to avoid diverging implementations).

cmwylie19:
Updated:
- [x] Pods
- [x] GitOps
- [ ] Flux WIP
- [ ] Include WASM - WIP

outofrange:
Hi,

I wanted to ask if Pepr instead of Zarf Agent is coming / if there is any timeline.
We'd like to adjust the hook mechanism (like whitelisting of registries, described in a different issue), and I've seen that there already is a POC linked [here](https://github.com/defenseunicorns/zarf/issues/2056#issuecomment-1749407177).

[...]

PR #1900, opened 2023-07-11
1862 adr

cmwylie19:
## Description

ADR for decoupling the transform library from go code and allowing it to be used in TypeScript with Pepr.

## Related Issue

Fixes #

Relates to #1632

## Type of change

## Checklist before merging


Racer159 (approved):
Approving as is though these small things should be fixed when the ADR is finalized

Racer159 (approved):
lgtm

PR #1913, opened 2023-07-21
zarf package mirror command introduction

Racer159:
## Description

This adds a command that mirrors artifacts from a package and pushes them to their airgap equivalents

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging


src/pkg/packager/unload.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/defenseunicorns/zarf/src/config"
+	"github.com/defenseunicorns/zarf/src/pkg/message"
+	"github.com/defenseunicorns/zarf/src/pkg/utils"
+	"github.com/defenseunicorns/zarf/src/pkg/utils/helpers"
+	"github.com/defenseunicorns/zarf/src/types"
+)
+
+// Unload pulls resources from a package (images, git repositories, etc) and pushes them to remotes in the air gap without deploying them
+func (p *Packager) Unload(packageName string) (err error) {
+	spinner := message.NewProgressSpinner("Unloading zarf package %s", packageName)
```
YrrepNoj: Unsure how this is intended to be called, but we should probably use the package name from the `p *Packager` metadata.

Racer159: yep yep - refactored

Madeline-UX:
@Racer159 Created a test for naming of new command. Results of the test will appear on the [Maze Results page]( https://app.maze.co/report/New-Feature-Naming-Text/7fpilkd04gok/intro) Planning to get participants early next week -

laurentiusoica:
Are there plans to include mirroring for helm charts as well? And also configuration to control the git repository names when mirroring?

src/cmd/package.go
```diff
@@ -102,6 +103,28 @@ var packageDeployCmd = &cobra.Command{
...
+		pkgConfig.PkgSource = pkgConfig.PkgOpts.PackagePath
+
+		// Configure the packager
+		pkgClient := packager.NewOrDie(&pkgConfig)
+		defer pkgClient.ClearTempPaths()
+
+		// Deploy the package
+		if err := pkgClient.Mirror(noImgChecksum); err != nil {
+			message.Fatalf(err, lang.CmdPackageDeployErr, err.Error())
+		}
```
Noxsios, line 124: For consistency `noImgChecksum` should be made into `MirrorOpts`, as all other top level packager receivers do not take arguments and instead grab from the opts struct.

src/cmd/package.go
```diff
@@ -334,6 +359,27 @@ func bindDeployFlags(v *viper.Viper) {
...
 
+func bindMirrorFlags(v *viper.Viper) {
+	mirrorFlags := packageMirrorCmd.Flags()
+
+	// Always require confirm flag (no viper)
+	mirrorFlags.BoolVar(&config.CommonOptions.Confirm, "confirm", false, lang.CmdPackageDeployFlagConfirm)
+
+	mirrorFlags.BoolVar(&noImgChecksum, "no-img-checksum", false, lang.CmdPackageMirrorFlagNoChecksum)
+
+	mirrorFlags.StringVar(&pkgConfig.PkgOpts.OptionalComponents, "components", v.GetString(common.VInitComponents), lang.CmdPackageMirrorFlagComponents)
```
Noxsios, line 370: Mirror resources by default uses `v.GetString(common.VInitComponents)`?

src/pkg/packager/mirror.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/defenseunicorns/zarf/src/config"
+	"github.com/defenseunicorns/zarf/src/pkg/message"
+	"github.com/defenseunicorns/zarf/src/pkg/utils/helpers"
+	"github.com/defenseunicorns/zarf/src/types"
+)
+
+// Mirror pulls resources from a package (images, git repositories, etc) and pushes them to remotes in the air gap without deploying them
+func (p *Packager) Mirror(noImgChecksum bool) (err error) {
+	spinner := message.NewProgressSpinner("Mirroring zarf package %s", p.cfg.PkgOpts.PackagePath)
```
Noxsios, line 19:
```suggestion
	spinner := message.NewProgressSpinner("Mirroring Zarf package %s", p.cfg.PkgOpts.PackagePath)
```

src/pkg/packager/mirror.go
```diff
@@ -0,0 +1,92 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+// mirrorComponent mirrors a Zarf Component.
+func (p *Packager) mirrorComponent(component types.ZarfComponent, noImgChecksum bool) error {
+	message.Debugf("packager.mirrorComponent(%#v, %#v", p.tmp, component)
```
Noxsios, line 66:
```suggestion
```

src/pkg/packager/mirror.go
```diff
@@ -0,0 +1,92 @@
...
+		if err := p.pushImagesToRegistry(component.Images, noImgChecksum); err != nil {
+			return fmt.Errorf("unable to push images to the registry: %w", err)
+		}
+	}
+
+	if hasRepos {
+		if err = p.pushReposToRepository(componentPath.Repos, component.Repos); err != nil {
+			return fmt.Errorf("unable to push the repos to the repository: %w", err)
+		}
+	}
```
Noxsios, line 89: this is where I would put debug statements if I wanted them, as all of the previous operations in this fn are near guaranteed to not fail.

Noxsios (changes requested):

src/cmd/package.go
Racer159: Le Copy Le Pasta

src/cmd/package.go
Racer159: Yeah that was from back when that wasn't the case - will refactor

src/config/config.go
```diff
@@ -66,6 +66,7 @@ const (
 
 	ZarfDeployStage = "Deploy"
 	ZarfCreateStage = "Create"
+	ZarfUnloadStage = "Unload"
```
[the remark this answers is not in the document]
Racer159: The way these are used the actual text of the string is printed and would need to be resolved (so an enum that is an integer would require a secondary slice to decode it).

src/pkg/packager/mirror.go
Racer159: I think it is fine to not add debug statements - removed them from the other places (was mostly copy pasta at that point)

cmwylie19 (approved):
lgtm

Noxsios (approved):

Issue #1888, opened 2023-07-07
Allow templating COMPONENT_NAME during package create


Racer159:
### Is your feature request related to a problem? Please describe.
When creating reusable importable components it would be useful to have a `###ZARF_COMPONENT_NAME###` template / variable that can be used to scope that component properly.

### Describe the solution you'd like

- **Given** I have a `zarf.yaml` with `###ZARF_COMPONENT_NAME###` in a string value
- **When** I run `zarf package create` on that package
- **Then** Zarf will template that value with that component's name

### Describe alternatives you've considered
We could do this on deploy time or in another way, but most scoping would be around variables or actions and this would be easier / more declarative to do create-time.

### Additional context
This would make things like a reusable S3 bucket component easier to make.

Racer159:
Closing as completed by #1923

PR #1923, opened 2023-07-26
1888 - Template Component Fields into Components on Zarf.yaml

cmwylie19:
## Description

This PR allows for templating component name during package create and give the ability to template with more fields in the component struct. 

While the PR is roughly similar to [FindYamlTemplates](https://github.com/defenseunicorns/zarf/blob/02150cfda1a32d5c6ebb4c446d731dc217e3961e/src/pkg/utils/yaml.go#L164)  there is extra complexity in components that required a new function [FindComponentTemplatesAndReload](https://github.com/defenseunicorns/zarf/blob/9b3e396abc39a395ceb7999565bfb09088cd39c5/src/pkg/utils/yaml.go#L166). Each component may have a different name that the string map would overwrite. Due to this nature, the straightforward path was a new function to mutate in place, FindComponentTemplatesAndReload. It operates on the ZarfPackage object passed in by interface by iterating over the Components and mutating by updating the templated files. This opens up the possibility to template more fields in components rather than just [component names](https://github.com/defenseunicorns/zarf/blob/9b3e396abc39a395ceb7999565bfb09088cd39c5/src/pkg/utils/yaml.go#L186). 

Currently there is a place-holder example of [Component-Name-Template](https://github.com/defenseunicorns/zarf/blob/1888/examples/component-name-template/zarf.yaml) which will need to be changed before removing draft status from PR. 

This PR is in a holding pattern while waiting from feedback from @Racer159 

## Related Issue

Fixes 

Relates to #1888

## Type of change

## Checklist before merging


src/test/e2e/51_oci_compose_test.go
```diff
@@ -109,6 +109,33 @@ func (suite *SkeletonSuite) Test_1_Compose() {
...
+		"import-helm-local case was here",
+		"import-helm-local-relative",
+		"import-helm-oci case was here",
+		"import-repos case was here",
+		"import-images case was here",
+	}
+
+	for _, target := range targets {
+		suite.Contains(stdErr, target)
+	}
```
Noxsios, line 135: this does properly test that component names were replaced, but does not test that the replacement mapped properly

src/test/packages/51-import-everything/zarf.yaml
```diff
@@ -6,12 +6,14 @@ metadata:
 
 components:
   - name: import-component-local
+    description: "###ZARF_COMPONENT_NAME### case was here"
```
Noxsios, line 9:
```suggestion
  - name: import-component-local
    description: "import-component-local == ###ZARF_COMPONENT_NAME###"
```

implements changes proposed above

Noxsios (changes requested):

src/test/packages/51-import-everything/zarf.yaml
Noxsios: this would be replicated across all of the components + the targets array

src/test/e2e/51_oci_compose_test.go
cmwylie19: this is already on a [per component basis](https://github.com/defenseunicorns/zarf/blob/304d5d7041eaaad7c30cbccd4d9a556f8168592e/src/pkg/packager/variables.go#L159), which uses a [map of strings](https://github.com/defenseunicorns/zarf/blob/304d5d7041eaaad7c30cbccd4d9a556f8168592e/src/pkg/packager/variables.go#L160) to ensure one Component name is mapped each type. 

Do you need more granularity than that said in the above comment?

src/test/packages/51-import-everything/zarf.yaml
cmwylie19: no problem making this change, just want to get your opinion on if you think this is needed as each component would all have the same ###ZARF_COMPONENT_NAME### as it is on a per component basis. The nature of the function using the string map of strings ensures this, or is this something that you want to unit test?

src/test/e2e/51_oci_compose_test.go
cmwylie19: the other hesitation i have around this is the way that we are testing. We are using stdErr which has colors and it not easily unmarshaled to a Component. Would need to change the way we are testing this

Noxsios (approved):
lgtm

PR #1927, opened 2023-07-28
return images as an array of strings for all components

### LINKED ISSUE #1918: Zarf prepare findImages method should return the string of images not only printing them to stdout.
We are consuming Zarf a library, especially the FindImages method from the pkgClient *packager.Packager object. This method computes the list of images that are found inside a ZarfConfig. 
The problem is that the method only prints in a structured yaml format the list of images to stdout but does not return them using the function return variables.

Solutions 1:

The FindImages function should change the signature to:
func (p *Packager) FindImages(baseDir, repoHelmChartPath string, kubeVersionOverride string) (string, error) {}

The returned string should be the list of images

Solutions 2:

The FindImages function should change the signature to:
func (p *Packager) FindImages(baseDir, repoHelmChartPath string, kubeVersionOverride string) (*ZarfConfiguration, error) {}

type ZarfComponent struct {
	Name   string   `yaml:"name"`
	Images []string `yaml:"images"`
}

type ZarfConfiguration struct {
	Components []IndigoZarfComponent `yaml:"components"`
}

The return should represent a domain model and not a simple string.

cmwylie19:
## Description

Users using library code can benefit from receiving a [list of images ](https://github.com/defenseunicorns/zarf/blob/8d67f5fb2935ee9355786dfc0fd3b83efb6a8142/src/pkg/packager/prepare.go#L31). This was a feature request from the community. This provides users with the ability to pre-cache/pre-download images. 

## Related Issue

Fixes #1918 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/packager/prepare.go
+++ b/src/pkg/packager/prepare.go
@@ -28,10 +28,10 @@ import (
 )
 
 // FindImages iterates over a Zarf.yaml and attempts to parse any images.
-func (p *Packager) FindImages(baseDir, repoHelmChartPath string, kubeVersionOverride string) ([]string, error) {
+func (p *Packager) FindImages(baseDir, repoHelmChartPath string, kubeVersionOverride string) (map[string][]string, error) {
 
 	var originalDir string
-	var images []string
+	imagesMap := make(map[string][]string)
 
 	// Change the working directory if this run has an alternate base dir
 	if baseDir != "" {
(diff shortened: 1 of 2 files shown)


src/pkg/packager/prepare.go
```diff
@@ -28,9 +28,10 @@ import (
 )
 
 // FindImages iterates over a Zarf.yaml and attempts to parse any images.
-func (p *Packager) FindImages(baseDir, repoHelmChartPath string, kubeVersionOverride string) error {
+func (p *Packager) FindImages(baseDir, repoHelmChartPath string, kubeVersionOverride string) ([]string, error) {
```
Noxsios, line 31: Instead of returning an `[]string` it would be more useful to return either a map of `component name --> []string`, or an `[]ZarfComponent` I believe this more properly fulfills the intent of the original issue.

Noxsios (changes requested):

src/pkg/packager/prepare.go
cmwylie19: Agreed, makes sense

cmwylie19: Let me know if the change addresses your recommendation sufficiently.

Racer159 (approved):
lgtm

Noxsios (approved):

Issue #1932, opened 2023-07-31
BB Extension Flux deprecation on 2.7.0+


mjnagel:
### Environment
Device and OS: ec2 instance, Ubuntu
App version: zarf 0.28.3, BB 2.7.0
Kubernetes distro being used: k3d

### Steps to reproduce

Deploy BB 2.7.0 via the extension and examine the debug log. You will see a warning about the deprecation:

```
└ (/home/runner/work/zarf/zarf/src/pkg/message/message.go:137)
  ⠦  Attempting chart installation (3s)  DEBUG   2023-07-31T16:15:56Z  -  "level"=0 "msg"="v1beta2 GitRepository is deprecated, upgrade to v1\n"                                                                                                                                 
└ (/home/runner/work/zarf/zarf/src/pkg/message/message.go:137)
  ⠦  creating 13 resource(s)  DEBUG   2023-07-31T16:15:56Z  -  "level"=0 "msg"="v1beta2 GitRepository is deprecated, upgrade to v1\n"                                                                                                                                            
└ (/home/runner/work/zarf/zarf/src/pkg/message/message.go:137)
```

### Expected result

No deprecated API Versions should be used.

### Actual Result

Warning about gitrepo being on deprecated version

PR #1933, opened 2023-07-31
Big Bang extension flux api version update

mjnagel:
## Description

Updates the Flux GitRepo version to v1.

Not sure if this is the right/desired fix or if this should be more backwards compatible (similar to the way credentials are handled for BB 1/2.

## Related Issue

Fixes https://github.com/defenseunicorns/zarf/issues/1932

## Type of change

## Checklist before merging


src/extensions/bigbang/manifests.go
```diff
@@ -91,7 +91,7 @@ func manifestGitRepo(cfg *extensions.BigBang) fluxSrcCtrl.GitRepository {
 	return fluxSrcCtrl.GitRepository{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       fluxSrcCtrl.GitRepositoryKind,
-			APIVersion: "source.toolkit.fluxcd.io/v1beta2",
```
Racer159, line 94: I do think it would be good to do some backwards compatibility handling here - it shouldn't be too difficult to check the semver since you have the full extension configuration.

Racer159 (changes requested):

src/extensions/bigbang/manifests.go
mjnagel: Let me know if I addressed this well enough.

src/extensions/bigbang/manifests.go
```diff
@@ -88,10 +88,22 @@ func manifestZarfCredentials(version string) corev1.Secret {
 
 // manifestGitRepo generates a GitRepository object for the Big Bang umbrella repo.
 func manifestGitRepo(cfg *extensions.BigBang) fluxSrcCtrl.GitRepository {
+	apiVersion := "source.toolkit.fluxcd.io/v1beta2"
+	semverVersion, err := semver.NewVersion(cfg.Version)
+	if err == nil {
+		c, err := semver.NewConstraint(">= 2.7.0")
+		if err == nil {
+			updateFlux, err := c.Validate(semverVersion)
+			if err == nil && updateFlux && !cfg.SkipFlux {
```
mjnagel, line 97: Calling attention to this because I'm not sure what zarf wants to do here...there is the potential someone with `skipFlux` could have an older version of flux that doesn't support this newer v1 apiVersion so I added this check. Not sure if that's a good approach or not...I guess depends on zarf's posture towards skipping flux?

cmwylie19: so if the semantic version of cfg.Version is >= 2.7.0 then BB uses "source.toolkit.fluxcd.io/v1" else it uses "source.toolkit.fluxcd.io/v1beta2"?

cmwylie19 (changes requested):
https://github.com/defenseunicorns/zarf/blob/f2fc5513580c67dc441b69b06c283194466840d2/packages/zarf-agent/manifests/webhook.yaml#L85

I think we may need to tell the `MutatingWebhookConfiguration` to look at `v1` here too since we are bumping the version of Flux. @Racer159 is that correct?

src/extensions/bigbang/manifests.go
mjnagel: Yeah that's correct...as long as the BB extension config _does not_ skip flux. If the flux install is being skipped in the extension then this logic would leave it as `source.toolkit.fluxcd.io/v1beta2` given the unknown about the version they are using.

cmwylie19:
im running this k3s test again, i think it should pass, it seems to be timing at different places in the CI but it is passing locally.

cmwylie19 (approved):
lgtm

src/extensions/bigbang/manifests.go
Racer159, line 97:
```suggestion
	apiVersion := "source.toolkit.fluxcd.io/v1beta2"

	// Set apiVersion to v1 on BB v2.7.0 or higher falling back to v1beta2 as needed
	semverVersion, _ := semver.NewVersion(cfg.Version)
	if semverVersion != nil {
		c, _ := semver.NewConstraint(">= 2.7.0")
		if c != nil {
			updateFlux, _ := c.Validate(semverVersion)
			if updateFlux && !cfg.SkipFlux {
```

Racer159 (changes requested):

src/extensions/bigbang/manifests.go
mjnagel: Updated!

PR #1940, opened 2023-08-01
Migrating bundle code out of Zarf

UncleGedd:
Removing bundle code as it is migrated over to [uds-cli](https://github.com/defenseunicorns/uds-cli)


Racer159:
Looks like you will need to run `make docs-and-schema`

src/pkg/packager/actions.go
```diff
@@ -200,6 +200,15 @@ func actionCmdMutation(cmd string, shellPref types.ZarfComponentActionShell) (st
 		return cmd, err
 	}
 
+	// if another binary is calling Zarf, use the system Zarf
+	regex, err := regexp.Compile("^.*zarf(\\.exe)?$")
```
Racer159, line 204:
```suggestion
	// See https://regex101.com/r/O4Ngia/1
	regex, err := regexp.Compile(`^.*zarf(\.exe)?$`)
```
Just a small change to formatting - we usually provide regex101 links for regex's and use ` instead of " for the regex string.

Racer159 (changes requested):

Noxsios (approved):

PR #1947, opened 2023-08-04
Use standard admonition syntax

bdfinst:
The overview file used a mix of syntax for admonitions. I simplified it to prepare for conversion.


Racer159 (approved):
lgtm

Issue #1750, opened 2023-05-24
Explore places to add more unit tests to library code


Racer159:
### Describe what should be investigated or refactored
We should look at targeted areas to add unit tests (specifically in `src/pkg`) to provide better assurances over the interfaces in this code.

### Links to any relevant code
https://github.com/defenseunicorns/zarf/tree/main/src/pkg

### Additional context
This is something we should look at as we move toward GA and get Zarf's interfaces (both schema and library code) to be more stable.

schristoff-du:
Closing as a dupe

PR #1957, opened 2023-08-07
Add unit test for validatePackageArchitecture() method

lucasrod16:
## Description

- Replaces use of the concrete `*kubernetes.Clientset` implementation with the `kubernetes.Interface` for client operations for easier mocking/testing

- Adds unit test for `validatePackageArchitecture()` method

## Related Issue

Relates to https://github.com/defenseunicorns/zarf/issues/1750

## Type of change

## Checklist before merging


src/internal/packager/helm/chart.go
```diff
@@ -389,7 +389,7 @@ func (h *Helm) loadChartData() (*chart.Chart, chartutil.Values, error) {
 
 func (h *Helm) migrateDeprecatedAPIs(latestRelease *release.Release) error {
 	// Get the Kubernetes version from the current cluster
-	kubeVersion, err := h.Cluster.Kube.Clientset.ServerVersion()
+	kubeVersion, err := h.Cluster.Kube.Clientset.Discovery().ServerVersion()
```
Racer159, line 392:
```suggestion
	kubeVersion, err := h.Cluster.Kube.GetServerVersion()
```

src/pkg/k8s/version.go
```diff
@@ -0,0 +1,17 @@
...
+
+// GetServerVersion retrieves and return the k8s revision.
+func (k *K8s) GetServerVersion() (version string, err error) {
+	versionInfo, err := k.Clientset.Discovery().ServerVersion()
+	if err != nil {
+		return "", fmt.Errorf("unable to get Kubernetes version from the cluster : %w", err)
+	}
+
+	return versionInfo.String(), nil
+}
```
Racer159, line 17: This could probably be added to what is currently `distro.go` and then that file could be renamed `info.go` since it already does more than get the Distro (it also has an architecture function)

src/pkg/packager/deploy.go
```diff
@@ -38,9 +38,16 @@ var (
...
+func (p *Packager) Deploy() (err error) {
 	message.Debug("packager.Deploy()")
 
+	// Attempt to connect to a Kubernetes cluster.
+	// Not all packages require Kubernetes, so we only want to log a debug message rather than return the error when we can't connect to a cluster.
+	p.cluster, err = cluster.NewCluster()
+	if err != nil {
+		message.Debug(err)
+	}
+
```
Racer159, line 50: Since we are doing this now we should also refactor line 60 of `src/pkg/packager/interactive.go` to use p.cluster and simply check if it is nil instead of instantiating it's own cluster.

Racer159 (changes requested):
a few changes, but overall looks good!

Racer159 (approved):
lgtm

PR #1962, opened 2023-08-09
Zarf files archive path support

### LINKED ISSUE #1928: Zarf files archive path support
Often binaries or other files that need to be included in a Zarf package are contained inside of archives (zip / tar) and need to be extracted to be used.  This currently requires an additional script or action which clutters the `zarf.yaml`.

- **Given** I have a `zarf.yaml` that specifies a file archive
- **And** I specify an `archivePath`

```yaml
    files:
      - source: https://github.com/weaveworks/eksctl/releases/download/v0.146.0/eksctl_Darwin_amd64.tar.gz
        target: ~/.zarf/bin/eksctl
        shasum: 2e3faa90e3adf639aed4dbefd7691f73f4b2a21d64d907e6bdbc086cd79bf275
        archivePath: eksctl
```

- **When** I create the package
- **Then** the archive is extracted and only the internal file or directory is included in the package
- **And** the SHASUM I provide would be that of the source archive

We could add an addition unarchive action type (similar to wait) but this action is not normally life cycle specific and likely would not make sense being done in this way.

### Additional context
Common binaries like flux are distributed this way and in use in zarf packages today: https://github.com/defenseunicorns/uds-package-dubbd/blob/6c00497870b7bed3ce1b03554705b20dda746c41/defense-unicorns-distro/zarf.yaml#L138

cmwylie19:
## Description

Add support for ArchivePath which can be a file or a folder in Zarf Package

## Related Issue

Fixes #1928 

Relates to #

## Type of change

## Checklist before merging


examples/archive-path/zarf.yaml
cmwylie19, line 1: this is just for local testing

packages/distros/eks/zarf.yaml
cmwylie19, line 1: I added a new component because I didn't want to interfere with the actions of `load-eksctl`

src/types/component.go
```diff
@@ -79,11 +79,12 @@ type ZarfComponentOnlyCluster struct {
...
-	Executable bool     `json:"executable,omitempty" jsonschema:"description=(files only) Determines if the file should be made executable during package deploy"`
-	Symlinks   []string `json:"symlinks,omitempty" jsonschema:"description=List of symlinks to create during package deploy"`
+	Source      string   `json:"source" jsonschema:"description=Local folder or file path or remote URL to pull into the package"`
+	Shasum      string   `json:"shasum,omitempty" jsonschema:"description=(files only) Optional SHA256 checksum of the file"`
+	Target      string   `json:"target" jsonschema:"description=The absolute or relative path where the file or folder should be copied to during package deploy"`
+	Executable  bool     `json:"executable,omitempty" jsonschema:"description=(files only) Determines if the file should be made executable during package deploy"`
+	Symlinks    []string `json:"symlinks,omitempty" jsonschema:"description=List of symlinks to create during package deploy"`
+	ArchivePath string   `json:"archivePath,omitempty" jsonschema:"description=Local folder or file to be extracted into the package"`
```
Racer159, line 87:
```suggestion
	ArchivePath string   `json:"archivePath,omitempty" jsonschema:"description=Local folder or file to be extracted from a 'source' archive"`
```

src/pkg/utils/helpers/url.go
```diff
@@ -40,3 +41,14 @@ func DoHostnamesMatch(url1 string, url2 string) (bool, error) {
 
 	return parsedURL1.Hostname() == parsedURL2.Hostname(), nil
 }
+
+// ExtractFilenameFromURL returns filename from URL string
+func ExtractFilenameFromURL(urlStr string) (string, error) {
```
Racer159, line 46:
```suggestion
func ExtractBasePathFromURL(urlStr string) (string, error) {
```

I think BasePath might be more accurate than filename since that is the function we are running under the hood.

src/pkg/utils/helpers/url_test.go
```diff
@@ -90,6 +90,29 @@ func (suite *TestURLSuite) Test_2_DoHostnamesMatch() {
 	suite.False(b)
 }
 
+func (suite *TestURLSuite) Test_3_ExtractFilenameFromURL() {
+	urls := []string{
```
Racer159, line 94: We should have some things that will generate errors as well to ensure those return properly

packages/distros/eks/zarf.yaml
```diff
@@ -20,6 +20,16 @@ variables:
...
 components:
+  - name: test-archive
+    description: Test archive
+    required: true
+    files:
+      - source: https://github.com/weaveworks/eksctl/releases/download/v0.147.0/eksctl_Darwin_arm64.tar.gz
+        target: eksctl
+        shasum: 1d7dd5b9907de1cb3fa7832659db29f50530444d10e77b4a8eb27aa648da6fab
+        archivePath: eksctl
+        executable: true
```
Racer159, line 31: This should be integrated into `load-eksctl`

src/pkg/utils/helpers/files.go
```diff
@@ -0,0 +1,50 @@
...
+				return err
+			}
+
+		}
+		// Remove the compressed file
+		err := os.Remove(archivePath)
+		if err != nil {
+			return err
+		}
+		return nil
```
Racer159, line 46: To be a more discrete helper this should probably be done in the parent caller since not all users of this function might want the archive deleted.

src/pkg/utils/helpers/files.go
```diff
@@ -0,0 +1,50 @@
...
+				return err
+			}
+			defer destinationFile.Close()
+
+			// Write the data to the destination file
+			_, err = destinationFile.Write(data)
+			if err != nil {
+				return err
+			}
+
```
Racer159, line 39: We should do an `io.Copy` instead of a read then write.  This reads the file into RAM and will cause issues if the file size exceeds available RAM.

src/pkg/utils/helpers/files.go
```diff
@@ -0,0 +1,50 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/mholt/archiver/v3"
+)
+
+// FindAndCopyFileFromArchive inspects an archive file, and copies the target file to the destination
+func FindAndCopyFileFromArchive(archivePath, targetFile, destinationDir string) error {
+	err := archiver.Walk(archivePath, func(f archiver.File) error {
+		if strings.HasSuffix(f.Name(), targetFile) {
```
Racer159, line 19: This won't work for directories - I believe it would need to be:

```suggestion
		if strings.HasPrefix(f.Name(), targetFile) {
```

src/test/e2e/00_use_cli_test.go
```diff
@@ -116,6 +116,26 @@ func TestUseCLI(t *testing.T) {
...
+		stdOut, stdErr, err := e2e.Zarf("package", "create", "packages/distros/eks", "--zarf-cache", "--confirm")
+		require.NoError(t, err, stdOut, stdErr)
+	})
+
+	t.Run("zarf package inspect for archive path", func(t *testing.T) {
+		t.Parallel()
+		path := "build/zarf-package-distro-eks-multi-0.0.2.tar.zst"
+		stdOut, stdErr, err := e2e.Zarf("package", "inspect", path)
+		require.NoError(t, err, stdOut, stdErr)
+	})
```
Racer159, line 130: This is not needed since it doesn't touch the `file`

src/test/e2e/00_use_cli_test.go
```diff
@@ -116,6 +116,26 @@ func TestUseCLI(t *testing.T) {
 		require.Error(t, err, stdOut, stdErr)
 	})
 
+	t.Run("zarf package create for archive path", func(t *testing.T) {
```
Racer159, line 119: This could be incorporated into a single test that creates and deploys.

src/test/e2e/00_use_cli_test.go
```diff
@@ -116,6 +116,26 @@ func TestUseCLI(t *testing.T) {
...
+		path := "build/zarf-package-distro-eks-multi-0.0.2.tar.zst"
+		stdOut, stdErr, err := e2e.Zarf("package", "inspect", path)
+		require.NoError(t, err, stdOut, stdErr)
+	})
+
+	t.Run("zarf package deploy for archive path", func(t *testing.T) {
+		t.Parallel()
+		path := "build/zarf-package-distro-eks-multi-0.0.2.tar.zst"
+		stdOut, stdErr, err := e2e.Zarf("package", "deploy", path, "--confirm")
+		require.NoError(t, err, stdOut, stdErr)
```
Racer159, line 136: This should check that the right `eksctl` is spit out and should clean up the file when done.

Racer159 (changes requested):

src/pkg/utils/helpers/files.go
cmwylie19: i definitely hear you on that. The problem that I ran into was that io.Copy requires a `*os.File` and that is why i ended up going this route. Is there a workaround for this that you know of?

src/pkg/utils/helpers/files.go
cmwylie19: image

I could be wrong but i think even in the case of a directory it must be suffix because the prefix is a `.`

You brought up a good point that this does not work with directories. I am going to re-work this to check if the file is a directory

cmwylie19: Nevermind, i found a much easier way to do this

src/pkg/utils/helpers/files.go
cmwylie19: found a better way

src/test/e2e/00_use_cli_test.go
cmwylie19: https://github.com/defenseunicorns/zarf/blob/314e07cd895ecbd2d9deaa2185e6b4a6843355af/packages/distros/eks/zarf.yaml#L29 the package's actions delete eksctl

cmwylie19:
I thought this was RTR but this is now in a draft state because I have some questions around the tests. They were passing locally but new complexities are introduced in the e2e test around the folder where the package is being built

cmwylie19:
Tested with a Variety of usecases, can accept local file or URLs.

```
    files:
      # - source: https://github.com/fluxcd/flux2/releases/download/v2.0.1/flux_2.0.1_darwin_arm64.tar.gz
      #   target:  flux
      #   executable: true
      #   shasum: 9b17a9d0ee8e0f0cf071e74df05e3e264320c7ce415cb706791b5773de7747f0
      #   archivePath: flux
      # - source: https://github.com/cmwylie19/archive-path/raw/main/archive-path.tar.gz
      #   target:  archive-path
      #   # executable: true
      #   # shasum: 9b17a9d0ee8e0f0cf071e74df05e3e264320c7ce415cb706791b5773de7747f0
      #   archivePath: archive-path
      # - source: archive-path.tar.gz
      #   target:  archive-path
      #   # executable: true
      #   # shasum: 9b17a9d0ee8e0f0cf071e74df05e3e264320c7ce415cb706791b5773de7747f0
      #   archivePath: archive-path
      - source: hi.tar.gz
        target:  something/hi.txt
        # executable: true
        shasum: aadc1955c030f723e9d89ed9d486b4eef5b0d1c6945be0dd6b7b340d42928ec9
        archivePath: hi.txt
```

cmwylie19:
I excluded this from the source but I am adding it for any questions that I may get.
```go
				// get the compressedFileName from the source
				compressedFileName, err := helpers.ExtractBasePathFromURL(file.Source)
				if err != nil {
					return fmt.Errorf(lang.ErrFileNameExtract, file.Source, err.Error())
				}

				compressedFile := filepath.Join(componentPath.Temp, compressedFileName)

				// If the file is an archive, download it to the componentPath.Temp
				if err := utils.DownloadToFile(file.Source, compressedFile, component.DeprecatedCosignKeyPath); err != nil {
					return fmt.Errorf(lang.ErrDownloading, file.Source, err.Error())
				}
				// dst = /var/folders/v0/slmrzc4s6kx4n7jb77ch9fc80000gn/T/zarf-1676087642/components/load-eksctl/files/1/eksctl_Darwin_x86_64

				dirDst := filepath.Dir(dst)
				err = archiver.Extract(compressedFile, file.ExtractPath, dirDst)
				if err != nil {
					return fmt.Errorf(lang.ErrFileExtract, file.ExtractPath, compressedFileName, err.Error())
				}
				// this extracts to /var/folders/v0/slmrzc4s6kx4n7jb77ch9fc80000gn/T/zarf-3537766959/components/load-eksctl/files/1/eksctl
				updatedExtractedFileOrDir := filepath.Join(dirDst, file.ExtractPath)
				if updatedExtractedFileOrDir != dst {
					err = os.Rename(updatedExtractedFileOrDir, dst)
					if err != nil {
						return fmt.Errorf(lang.ErrWritingFile, dst, err)
					}
				}
```

src/pkg/packager/create.go
```diff
@@ -389,23 +388,74 @@ func (p *Packager) addComponent(index int, component types.ZarfComponent, isSkel
...
+				if updatedExtractedFileOrDir != dst {
+					err = os.Rename(updatedExtractedFileOrDir, dst)
+					if err != nil {
+						return fmt.Errorf(lang.ErrWritingFile, dst, err)
+					}
+				}
+			} else {
+				if err := utils.CreatePathAndCopy(file.Source, dst); err != nil {
+					return fmt.Errorf("unable to copy file %s: %w", file.Source, err)
+				}
```
Racer159, line 444: This logic is the same as `407-425` and could be simplified.

Racer159 (changes requested):
not a full review but some things

src/test/e2e/00_use_cli_test.go
```diff
@@ -116,6 +116,16 @@ func TestUseCLI(t *testing.T) {
...
 
+	t.Run("zarf package to test archive path", func(t *testing.T) {
+		t.Parallel()
+		stdOut, stdErr, err := e2e.Zarf("package", "create", "packages/distros/eks", "--confirm")
+		require.NoError(t, err, stdOut, stdErr)
+
+		path := "zarf-package-distro-eks-multi-0.0.2.tar.zst"
+		stdOut, stdErr, err = e2e.Zarf("package", "deploy", path, "--confirm")
+		require.NoError(t, err, stdOut, stdErr)
+	})
```
Racer159, line 127:
```suggestion

		require.FileExists(t, "binaries/eksctl_Darwin_x86_64")
		require.FileExists(t, "binaries/eksctl_Darwin_arm64")
		require.FileExists(t, "binaries/eksctl_Linux_x86_64")

		e2e.CleanFiles("binaries/eksctl_Darwin_x86_64", "binaries/eksctl_Darwin_arm64", "binaries/eksctl_Linux_x86_64")
	})
```

src/pkg/packager/create.go
```diff
@@ -384,28 +382,74 @@ func (p *Packager) addComponent(index int, component types.ZarfComponent, isSkel
...
+			}
+		}
+
 		// Abort packaging on invalid shasum (if one is specified).
 		if file.Shasum != "" {
-			if actualShasum, _ := utils.GetCryptoHashFromFile(dst, crypto.SHA256); actualShasum != file.Shasum {
+			actualShasum, _ := utils.GetSHA256OfFile(dst)
+			if actualShasum != file.Shasum {
 				return fmt.Errorf("shasum mismatch for file %s: expected %s, got %s", file.Source, file.Shasum, actualShasum)
```
Racer159, line 450:
```suggestion
				return fmt.Errorf("shasum mismatch for file %s: expected %s, got %s", dst, file.Shasum, actualShasum)
```

src/pkg/packager/create.go
```diff
@@ -384,28 +382,74 @@ func (p *Packager) addComponent(index int, component types.ZarfComponent, isSkel
...
+				}
+			} else {
+				if err := utils.CreatePathAndCopy(file.Source, dst); err != nil {
+					return fmt.Errorf("unable to copy file %s: %w", file.Source, err)
+				}
 			}
+
 			if isSkeleton {
 				p.cfg.Pkg.Components[index].Files[filesIdx].Source = rel
 			}
```
Racer159, line 432:
```suggestion
```

src/pkg/packager/create.go
```diff
@@ -384,28 +382,74 @@ func (p *Packager) addComponent(index int, component types.ZarfComponent, isSkel
...
+			// Make sure dst reflects the actual file or directory.
+			updatedExtractedFileOrDir := filepath.Join(destinationDir, file.ExtractPath)
+			if updatedExtractedFileOrDir != dst {
+				err = os.Rename(updatedExtractedFileOrDir, dst)
+				if err != nil {
+					return fmt.Errorf(lang.ErrWritingFile, dst, err)
+				}
+			}
+		}
+
```
Racer159, line 445:
```suggestion

		if isSkeleton {
			// Change the source to the new relative source directory (any remote files will have been skipped above)
			p.cfg.Pkg.Components[index].Files[filesIdx].Source = rel
			// Remove the extractPath from a skeleton since it will already extract it
			p.cfg.Pkg.Components[index].Files[filesIdx].ExtractPath = ""
		}

```

Racer159 (changes requested):
This breaks for skeletons - the suggestions should fix it.

Racer159 (approved):
lgtm

PR #1973, opened 2023-08-15
Fix deploying packages w/ no component tarballs

### LINKED ISSUE #1963: Can't deploy an actions-only Zarf package pulled from an OCI registry
### Environment
Device and OS: Mac m1
App version:  v0.28.3

### Steps to reproduce
Given the following `zarf.yaml`:
```
kind: ZarfPackageConfig
metadata:
  name: no-cluster-pkg
  version: 0.0.1

variables:
  - name: COUNTRY
    default: Scotland
components:
  - name: echo
    required: true
    actions:
      onDeploy:
        before:
          - cmd: echo "Unicorns are the national animal of "${ZARF_VAR_COUNTRY}""
```
1. Build the Zarf pkg and push to an OCI registry
2. Pull the pkg down and deploy

### Expected result
The package deploys successfully!

### Actual Result
Pulling the Zarf pkg down and running `zarf p deploy zarf-package-no-cluster-pkg-arm64-0.0.1.tar.zst ` yields:
```
ERROR:  Failed to deploy package: unable to load the Zarf Package: unable to get a list of components...
             open /var/folders/sf/_jrk8zgs5tg33r_h1gzh_6t40000gn/T/zarf-2374842217/components: no such file or
             directory
```

### Additional Context
- At some point in the OCI reg, the `components` folder is removed from the Zarf pkg because it is an empty folder that a digest can't be computed for, so it's left out of the Zarf tarball artifact.
- Noting that the error comes from [here](https://github.com/defenseunicorns/zarf/blob/3d600c6dfd97212bbbbec0bd119aa6de517f302f/src/pkg/packager/common.go#L304)

Noxsios:
## Related Issue

Fixes #1963 

## Type of change


Racer159 (approved):
lgtm

Issue #1972, opened 2023-08-15
Error when deploying v0.28.4 of init pkg using library Zarf


UncleGedd:
### Environment
Device and OS: Mac M1
App version: 0.28.4
Kubernetes distro being used: k3d

### Steps to reproduce
When calling the [packager.Deploy](https://github.com/defenseunicorns/zarf/blob/5b5396c1466a123e5fb83adb2f02ac040a2c3e5b/src/pkg/packager/deploy.go#L41) function, I get the following error:
```
unable to parse Zarf CLI version 'unset' : Invalid Semantic Version
```
The stack trace leads to the [validateLastNonBreakingVersion](https://github.com/defenseunicorns/zarf/blob/5b5396c1466a123e5fb83adb2f02ac040a2c3e5b/src/pkg/packager/common.go#L465) function. Some notes on that function:

On v0.28.3 of the init pkg:
- the `lastNonBreakingVersion` var gets set to `""`
- the `cliVersion` var is set to "unset"

On v0.28.4 of the init pkg
- the `lastNonBreakingVersion` var gets set to `"v0.27.0"`
- the `cliVersion` var is set to "unset"

Racer159:
This is expected behavior - Zarf is trying to determine if the package that is being deployed might have issues deploying and is looking at it's version to do so - `unset` is the default value to flag that this value should be set - setting it to UnknownVersion will skip this check if you don't want to have a version at all (though this will throw errors in other places) or the version can be set to the version of Zarf (which is recommended since this value is used in checks elsewhere).

You should be able to get the info needed from `debug.BuildInfo` for now and we may take a look at refactoring this to be cleaner in the future (hence marking as tech debt instead of possible bug).

PR #1981, opened 2023-08-21
Throw warning instead of error when default CLIVersion is used when validating last non-breaking version

lucasrod16:
## Description
Print a warning rather than return an error when an invalid CLIVersion is used when validating last non-breaking version

## Related Issue

Fixes https://github.com/defenseunicorns/zarf/issues/1972

## Type of change

## Checklist before merging

### DIFF
--- a/src/config/lang/english.go
+++ b/src/config/lang/english.go
@@ -256,6 +256,7 @@ const (
 	CmdPackageDeployFlagPublicKey                      = "Path to public key file for validating signed packages"
 	CmdPackageDeployValidateArchitectureErr            = "this package architecture is %s, but the target cluster has the %s architecture. These architectures must be the same"
 	CmdPackageDeployValidateLastNonBreakingVersionWarn = "the version of this Zarf binary '%s' is less than the LastNonBreakingVersion of '%s'. You may need to upgrade your Zarf version to at least '%s' to deploy this package"
+	CmdPackageDeployUnsetCLIVersionWarn                = "CLIVersion is set to the default value of '%s'. This could potentially cause issues with package creation and deployment. To avoid such issues, please set the value to a valid semantic version."
 	CmdPackageDeployErr                                = "Failed to deploy package: %s"
 
 	CmdPackageInspectFlagSbom      = "View SBOM contents while inspecting the package"
--- a/src/config/lang/english.go
+++ b/src/config/lang/english.go
@@ -256,7 +256,7 @@ const (
 	CmdPackageDeployFlagPublicKey                      = "Path to public key file for validating signed packages"
 	CmdPackageDeployValidateArchitectureErr            = "this package architecture is %s, but the target cluster has the %s architecture. These architectures must be the same"
 	CmdPackageDeployValidateLastNonBreakingVersionWarn = "the version of this Zarf binary '%s' is less than the LastNonBreakingVersion of '%s'. You may need to upgrade your Zarf version to at least '%s' to deploy this package"
-	CmdPackageDeployUnsetCLIVersionWarn                = "CLIVersion is set to '%s' which can cause issues with package creation and deployment. To avoid such issues, please set the value to the valid semantic version for this version of Zarf."
+	CmdPackageDeployInvalidCLIVersionWarn              = "CLIVersion is set to '%s' which can cause issues with package creation and deployment. To avoid such issues, please set the value to the valid semantic version for this version of Zarf."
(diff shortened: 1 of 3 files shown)


src/pkg/packager/common.go
```diff
@@ -503,7 +503,7 @@ func (p *Packager) validateLastNonBreakingVersion() (err error) {
 	cliVersion := config.CLIVersion
 	lastNonBreakingVersion := p.cfg.Pkg.Build.LastNonBreakingVersion
 
-	if lastNonBreakingVersion == "" || cliVersion == "UnknownVersion" {
+	if lastNonBreakingVersion == "" || cliVersion == "UnknownVersion" || cliVersion == "unset" {
 		return nil
 	}
```
Racer159, line 509: This should use a warning instead and match print breaking changes

cmwylie19 (dismissed):
lgtm

src/pkg/packager/common.go
```diff
@@ -507,6 +505,12 @@ func (p *Packager) validateLastNonBreakingVersion() (err error) {
...
 
+	if cliVersion == "unset" {
+		unsetWarning := fmt.Sprintf(lang.CmdPackageDeployUnsetCLIVersionWarn, config.CLIVersion)
+		p.warnings = append(p.warnings, unsetWarning)
+		return nil
+	}
+
 	lastNonBreakingSemVer, err := semver.NewVersion(lastNonBreakingVersion)
```
Racer159, line 514: would rather have this logic in what is now line 521 instead of returning an error there.  This could also be `UnknownVersion` in addition to unset and they might also set it to a bad semantic version which we might not want to error out for.

src/config/lang/english.go
```diff
@@ -256,6 +256,7 @@ const (
 	CmdPackageDeployFlagPublicKey                      = "Path to public key file for validating signed packages"
 	CmdPackageDeployValidateArchitectureErr            = "this package architecture is %s, but the target cluster has the %s architecture. These architectures must be the same"
 	CmdPackageDeployValidateLastNonBreakingVersionWarn = "the version of this Zarf binary '%s' is less than the LastNonBreakingVersion of '%s'. You may need to upgrade your Zarf version to at least '%s' to deploy this package"
+	CmdPackageDeployUnsetCLIVersionWarn                = "CLIVersion is set to the default value of '%s'. This could potentially cause issues with package creation and deployment. To avoid such issues, please set the value to a valid semantic version."
```
Racer159, line 259:
```suggestion
	CmdPackageDeployUnsetCLIVersionWarn                = "CLIVersion is set to '%s' which can cause issues with package creation and deployment. To avoid such issues, please set the value to the valid semantic version for this version of Zarf."
```

Racer159 (changes requested):

src/pkg/packager/common.go
lucasrod16: Changes made

src/pkg/packager/common.go
```diff
@@ -496,14 +496,18 @@ func (p *Packager) validatePackageArchitecture() (err error) {
...
 func (p *Packager) validateLastNonBreakingVersion() (err error) {
 	cliVersion := config.CLIVersion
 	lastNonBreakingVersion := p.cfg.Pkg.Build.LastNonBreakingVersion
 
-	if lastNonBreakingVersion == "" || cliVersion == "UnknownVersion" {
+	if lastNonBreakingVersion == "" {
+		return nil
+	}
+
+	if cliVersion == "unset" || cliVersion == "UnknownVersion" {
```
Racer159, line 508: I don't think we explicitly need this as a separate check since the language it returns would suffice for an invalid semantic version as well - generally fewer if blocks means easier to understand logic.

Racer159 (changes requested):

src/pkg/packager/common.go
lucasrod16: Removed this check

src/pkg/packager/common.go
```diff
@@ -525,6 +525,7 @@ func (p *Packager) validateLastNonBreakingVersion() (err error) {
 			lastNonBreakingVersion,
 		)
 		p.warnings = append(p.warnings, warning)
+		return nil
```
Racer159, line 528:
```suggestion
```
Unnecessary logic

Racer159 (changes requested):
very close though would rather this nil return not be there

Racer159 (approved):

PR #1987, opened 2023-08-25
Ensure Zarf image manifests are pulled as blobs

UncleGedd:
## Description
When pulling from GHCR we noticed that Zarf image manifests were coming from the `/manifests/` path when they should actually be coming from `/blobs/`. Even though they are technically image manifests, Zarf stores in them OCI as blobs, so this PR changes the MediaType on the image manifest to be a Zarf blob


UncleGedd:
Confirmed that I can now deploy bundles containing packages from GHCR

Racer159 (approved):
lgtm

PR #1989, opened 2023-08-25
Fix potential for double transform of images (i.e with operators that self-reference)

### LINKED ISSUE #1988: Zarf image mutation happens twice in some scenarios
### Environment
Device and OS: macOS, m1 mac, deployment on Ubuntu 20
App version: 0.29.0
Kubernetes distro being used: RKE2 1.26

### Steps to reproduce
1. Create a deployment or job that already has a mutated image (a "127.0.0.1:31999/something:1.2.3-zarf-xxxxx" image rather than "docker.io/something:1.2.3"). This is probably difficult to do in practice 🙃 
2. Validate that the image is "double mutated" by the zarf agent. The pod created will end up having a double "zarf-xxxx" tag (with two separate hashes based on the different starting images). The image will also fail to pull as expected due to the wrong tag.

It might be easier to try and reproduce with [this package](https://github.com/defenseunicorns/uds-package-rook-ceph/blob/docs-values-updates/zarf.yaml) which is where I experienced the issue. More details on that in the additional context section below.

### Expected result

Zarf skips over images it has already mutated and not mutate their tags again.

### Actual Result

If a "parent resource" of a pod (deployment, job, etc) has an already mutated image specified then the pod will pass through zarf's mutation and end up with a doubled zarf-hash tag mutation.

Logs from zarf agent showing a patch with 2 zarf-hash tags:
```
  DEBUG   2023-08-25T18:33:17Z  -  PATCH:  [{"op":"replace","path":"/spec/imagePullSecrets","value":[{"name":"private-registry"}]},{"op":"replace","path":"/spec/initContainers/0/image","value":"127.0.0.1:31999/ironbank/rook/ceph:v1.12.2-zarf-3113775139-zarf-1434334155"},{"op":"replace","path":"/spec/containers/0/image","value":"127.0.0.1:31999/ironbank/opensource/ceph/ceph:v17.2.6-zarf-2774537164"},{"op":"replace","path":"/metadata/labels/zarf-agent","value":"patched"}]
```

### Severity/Priority

Low-medium. This is a weird edge case but could be encountered without any workarounds in some scenarios.

### Additional Context

This is probably a really wild edge case. The scenario I ran into this with was a zarf package deploying the rook operator. The operator will spin up several jobs based on the image it is running (which happens to be an already mutated pod in zarf's registry). The resulting pod for that job then experiences the behavior described above.

This package is definitely WIP but if you want to experience the error you should be able to build and deploy - https://github.com/defenseunicorns/uds-package-rook-ceph/blob/docs-values-updates/zarf.yaml

cmwylie19:
## Description

Checks if image.Host has already been patched

## Related Issue

Fixes #1988

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/transform/image.go
+++ b/src/pkg/transform/image.go
@@ -31,7 +31,7 @@ func ImageTransformHost(targetHost, srcReference string) (string, error) {
 	}
 
 	// check if image has already been transformed
-	if targetHost == image.Host || strings.HasPrefix(targetHost, image.Host) {
+	if strings.HasPrefix(targetHost, image.Host) {
 		return srcReference, nil
 	}
 
@@ -54,7 +54,7 @@ func ImageTransformHostWithoutChecksum(targetHost, srcReference string) (string,
 	}
 
 	// check if image has already been transformed
-	if targetHost == image.Host || strings.HasPrefix(targetHost, image.Host) {
+	if strings.HasPrefix(targetHost, image.Host) {
 		return srcReference, nil
 	}
 
--- a/src/pkg/transform/image.go
+++ b/src/pkg/transform/image.go
@@ -30,7 +30,11 @@ func ImageTransformHost(targetHost, srcReference string) (string, error) {
 	}
 
 	// check if image has already been transformed
-	if targetHost == image.Host {
+	isPatched, err := helpers.DoHostnamesMatch(targetHost, image.Host)
+	if err != nil {
+		return "", err
+	}
+	if isPatched {
 		return srcReference, nil
 	}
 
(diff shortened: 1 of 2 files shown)


src/pkg/transform/image.go
```diff
@@ -29,6 +29,11 @@ func ImageTransformHost(targetHost, srcReference string) (string, error) {
 		return "", err
 	}
 
+	// check if image has already been transformed
+	if targetHost == image.Host {
```
Racer159, line 33: This branch of code should have a unit test case added for it

Racer159 (changes requested):
looks good but this should probably be unit tested.

cmwylie19:
@Racer159 I took the needs test label off because I think i have covered the newly implemented logic in image_test.go. LMK if you would like to see changes.

Racer159 (approved):
lgtm

PR #1990, opened 2023-08-25
1773 Publish Zarf init + skeleton as OCI on release

### LINKED ISSUE #1773: Publish Zarf Init Skeleton
I would like to import the zarf init components as skeletons so that I can bring them into a larger Zarf package.

- **Given** a desire to deploy IaC+Zarf
- **When** creating a zarf package
- **Then** I can easily import components to do it

Something like this:
https://github.com/runyontr/zarf-package-k3d/blob/main/zarf.yaml#L31-L50

### Additional Context:
This should also swap the `zarf tools download-init` and `zarf init` prompts to pull the init package via `oci://`.
Init packages should still be published to GitHub releases though as before.

Noxsios:
## Related Issue

Fixes #1773 

## Type of change

## Checklist before merging

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,6 +63,11 @@ jobs:
           make release-init-package ARCH=amd64 AGENT_IMAGE_TAG=$GITHUB_REF_NAME
           make release-init-package ARCH=arm64 AGENT_IMAGE_TAG=$GITHUB_REF_NAME
 
+      - name: Publish Init Package as OCI and Skeleton
+        run: |
+          make publish-init-package ARCH=amd64 REPOSITORY_URL=ghcr.io/defenseunicorns/packages
+          make publish-init-package ARCH=arm64 REPOSITORY_URL=ghcr.io/defenseunicorns/packages
+
       # Create a CVE report based on this build
       - name: Create release time CVE report
         run: "make cve-report"
@@ -141,18 +146,6 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_GOV_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_GOV_SECRET_ACCESS_KEY }}
           aws-region: us-gov-west-1
-
-      - name: Login to GHCR
-        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a # v2.1.0
-        with:
-          registry: ghcr.io
-          username: dummy
-          password: ${{ github.token }}
-
-      - name: Publish Init Package as OCI and Skeleton
-        run: |
-          make publish-init-package ARCH=amd64 REPOSITORY_URL=ghcr.io/defenseunicorns/packages
-          make publish-init-package ARCH=arm64 REPOSITORY_URL=ghcr.io/defenseunicorns/packages
 
       - name: Skip brew latest for pre-release tags
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -142,6 +142,18 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_GOV_SECRET_ACCESS_KEY }}
           aws-region: us-gov-west-1
 
+      - name: Login to GHCR
+        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a # v2.1.0
+        with:
+          registry: ghcr.io
+          username: dummy
+          password: ${{ github.token }}
+
+      - name: Publish Init Package as OCI and Skeleton
+        run: |
+          make publish-init-package ARCH=amd64 REPOSITORY_URL=ghcr.io/defenseunicorns/packages
+          make publish-init-package ARCH=arm64 REPOSITORY_URL=ghcr.io/defenseunicorns/packages
+
       - name: Skip brew latest for pre-release tags
         run: |
           if [[ $GITHUB_REF_NAME == *"rc"* ]]; then
--- a/Makefile
+++ b/Makefile
@@ -132,7 +132,6 @@ ib-init-package:
 
 # INTERNAL: used to publish the init package
 publish-init-package:
-	@test -s $(ZARF_BIN) || $(MAKE) build-cli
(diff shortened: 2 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -142,6 +142,18 @@ jobs:
...
+        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a # v2.1.0
+        with:
+          registry: ghcr.io
+          username: dummy
+          password: ${{ github.token }}
+
+      - name: Publish Init Package as OCI and Skeleton
+        run: |
+          make publish-init-package ARCH=amd64 REPOSITORY_URL=ghcr.io/defenseunicorns/packages
+          make publish-init-package ARCH=arm64 REPOSITORY_URL=ghcr.io/defenseunicorns/packages
```
Racer159, line 155: I think we could publish these in the build stage like we do the agent image - if it fails a fix wouldn't be too far behind and would use the same tag just updating it - would also mean we wouldn't need to login to GHCR twice.

Makefile
```diff
@@ -130,6 +130,12 @@ ib-init-package:
 		--set REGISTRY_IMAGE="ironbank/opensource/docker/registry-v2" \
 		--set REGISTRY_IMAGE_TAG="2.8.2"
 
+# INTERNAL: used to publish the init package
+publish-init-package:
+	@test -s $(ZARF_BIN) || $(MAKE) build-cli
```
Racer159, line 135:
```suggestion
```

I think I would rather this line be left out similar to how `release-init-package` is.  That way there is less chance a dev build of zarf does something weird.

src/cmd/tools/zarf.go
```diff
@@ -174,9 +174,9 @@ var downloadInitCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		initPackageName := packager.GetInitPackageName("")
 		target := filepath.Join(outputDirectory, initPackageName)
-		url := packager.GetInitPackageRemote("")
-		err := utils.DownloadToFile(url, target, "")
-		if err != nil {
+		url := oci.GetInitPackageURL(config.CLIArch, config.CLIVersion)
```
Racer159, line 177:
```suggestion
		url := oci.GetInitPackageURL(config.GetArch(), config.CLIVersion)
```

src/cmd/initialize.go
```diff
@@ -114,7 +115,7 @@ func downloadInitPackage(downloadCacheTarget string) error {
 	}
 
 	var confirmDownload bool
-	url := packager.GetInitPackageRemote("")
+	url := oci.GetInitPackageURL(config.CLIArch, config.CLIVersion)
```
Racer159, line 118:
```suggestion
	url := oci.GetInitPackageURL(config.GetArch(), config.CLIVersion)
```

src/pkg/oci/utils.go
```diff
@@ -91,3 +94,36 @@ func RemoveDuplicateDescriptors(descriptors []ocispec.Descriptor) []ocispec.Desc
 	}
 	return list
 }
+
+// GetInitPackageURL returns the URL for the init package for the given architecture and version.
+func GetInitPackageURL(arch, version string) string {
+	return fmt.Sprintf("ghcr.io/defenseunicorns/packages/init:%s-%s", arch, version)
```
Racer159, line 100:
```suggestion
	return fmt.Sprintf("ghcr.io/defenseunicorns/packages/init:%s-%s", version, arch)
```

src/pkg/oci/utils.go
```diff
@@ -91,3 +94,36 @@ func RemoveDuplicateDescriptors(descriptors []ocispec.Descriptor) []ocispec.Desc
...
+}
+
+// DownloadPackage downloads the given OCI package and saves as a tarball.
+func DownloadPackage(url, destinationTarball string, concurrency int) error {
```
Racer159, line 104:
```suggestion
func PullTarball(url, destinationTarball string, concurrency int) error {
```
Would this make sense as `PullTarball` and then being moved to `pull.go` that way it is more discoverable to library users who may want it?

Racer159 (changes requested):

src/pkg/oci/utils.go
Noxsios: Functions related to OCI / OrasRemote that are not receiver methods on `OrasRemote` are contained in `utils.go`. Pull operations usually have hook-ins related to the way layers are treated. This function is very "dumb", which is why it is called `DownloadPackage`, and does not re-use verbiage contained elsewhere.

.github/workflows/release.yml
Noxsios: done

src/pkg/oci/utils.go
Racer159: Ok that makes sense - can we call the function `DownloadPackageTarball`` or `DownloadTarball`? Just to de-conflict with `PullPackage` since that pulls package layers into a dir which is a bit different than this.

Noxsios: sure

src/cmd/initialize.go
```diff
@@ -114,7 +115,7 @@ func downloadInitPackage(downloadCacheTarget string) error {
 	}
 
 	var confirmDownload bool
-	url := packager.GetInitPackageRemote("")
+	url := oci.GetInitPackageURL(config.GetArch(), config.CLIVersion)
 
 	// Give the user the choice to download the init-package and note that this does require an internet connection
 	message.Question(fmt.Sprintf(lang.CmdInitDownloadAsk, url))
```
Racer159, line 121: One tiny little thing here, when we print this to the user can we show them the `oci://` url as they would use it on package deploy?

i.e.

```
It seems the init package could not be found locally, but can be pulled from
oci://docker.io/racer159/init:v0.29.0-17-g046e436a-amd64

 NOTE  Note: This will require an internet connection.
? Do you want to pull this init package? Yes
```
image

(Not this adds `oci://` and changes `download` to `pull` in this context

Racer159 (changes requested):

src/cmd/initialize.go
Noxsios: lmk if this works

Racer159 (approved):
lgtm

PR #2000, opened 2023-09-01
Resolve edge cases with `--tmpdir` behavior

Racer159:
## Description

This PR resolves edge cases with `--tmpdir` behavior by always using the specified temp dir in commands and always making a unique sub folder for each zarf run.

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/utils/io.go
+++ b/src/pkg/utils/io.go
@@ -60,13 +60,13 @@ type TextTemplate struct {
 
 // MakeTempDir creates a temp directory with the zarf- prefix.
 func MakeTempDir(basePath string) (string, error) {
-	if basePath != "" && InvalidPath(basePath) {
+	if basePath != "" {
 		if err := CreateDirectory(basePath, 0700); err != nil {
 			return "", err
 		}
 	}
 	tmp, err := os.MkdirTemp(basePath, tmpPathPrefix)
-	message.Debugf("Creating temp path: '%s'", tmp)
+	message.Debug("Using temporary directory:", basePath)
 	return tmp, err
 }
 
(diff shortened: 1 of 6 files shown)


cmwylie19 (dismissed):
lgtm

src/pkg/utils/io.go
```diff
@@ -59,8 +59,13 @@ type TextTemplate struct {
 }
 
 // MakeTempDir creates a temp directory with the zarf- prefix.
-func MakeTempDir() (string, error) {
-	tmp, err := os.MkdirTemp("", tmpPathPrefix)
+func MakeTempDir(basePath string) (string, error) {
+	if basePath != "" && InvalidPath(basePath) {
```
Noxsios, line 63:
```suggestion
	if basePath != "" {
```

`utils.CreateDirectory` already performs a `InvalidPath` check

src/pkg/utils/io.go
```diff
@@ -59,8 +59,13 @@ type TextTemplate struct {
...
+		if err := CreateDirectory(basePath, 0700); err != nil {
+			return "", err
+		}
+	}
+	tmp, err := os.MkdirTemp(basePath, tmpPathPrefix)
 	message.Debugf("Creating temp path: '%s'", tmp)
```
Noxsios, line 69:
```suggestion
	message.Debugf("Created temp directory: %q", tmp)
```

Noxsios (changes requested):

src/pkg/utils/io.go
Racer159: Didn't exactly do this but normalized the verbiage across things - noticed we were double printing the path in some cases as well.

Noxsios (approved):

PR #2003, opened 2023-09-01
adds a hint when the release might be empty

dgershman:
## Description

This assists when you may include a `bigbang` package but for some reason a dependency is not included.  A good example of this is when you try to include `mattermost` and `mattermostOperator` but you forget, or don't know to include `minioOperator`.  The extension attempts to install it, but doesn't include the name of what it's trying to install.  This gives a hint by including the namespace in the output.  There might be a better way to solve this, but this seems to help me figure out what to do next.

Without this the output is:

```
Big Bang Helm Release `` to be ready
```

With this change the output is:

```
Big Bang Helm Release `` to be ready (Namespace: bigbang.minio-operator)
```

## Related Issue

N/A

## Type of change

## Checklist before merging


src/extensions/bigbang/bigbang.go
```diff
@@ -151,7 +151,7 @@ func Run(YOLO bool, tmpPaths types.ComponentPaths, c types.ZarfComponent) (types
 	for _, hrNamespacedName := range namespacedHelmReleaseNames {
 		hr := hrDependencies[hrNamespacedName]
 		action := types.ZarfComponentAction{
-			Description:     fmt.Sprintf("Big Bang Helm Release `%s` to be ready", hr.Metadata.Name),
+			Description:     fmt.Sprintf("Big Bang Helm Release `%s` to be ready (Namespace: %s)", hr.Metadata.Name, hrNamespacedName),
```
Racer159, line 154:
```suggestion
			Description:     fmt.Sprintf("Big Bang Helm Release `%s` to be ready", hrNamespacedName),
```
Because `hrNamespacedName` already contains the HR name we could just use it instead.  I would be fine with the above _but_ we could also add a warning on the lookup of `hr := hrDependencies[hrNamespacedName]`.  If it was `hr, ok := hrDependencies[hrNamespacedName]` the `ok` boolean could be checked like below:

```
hr, ok := hrDependencies[hrNamespacedName]
if !ok {
message.Warnf("Discovered a missing dependent Helm release `%s`, reconfigure your Big Bang values file to add it.", hrNamespacedName)
}
```

Racer159 (changes requested):

Racer159 (approved):
lgtm!

Noxsios (approved):

PR #2005, opened 2023-09-04
feat: flux OCI support in Zarf Agent

### LINKED ISSUE #1974: Zarf does not support deploying Flux HelmReleases with OCI Helm chart sources

Zarf does not support deploying Flux HelmReleases with OCI Helm chart sources

_[Related issue](https://github.com/defenseunicorns/zarf/issues/2021)_

- **Given**
    * A `zarf.yaml` with `manifests:` including a Flux HelmRepository referencing an `oci://` chart and a HelmRelease referencing it.
    * And including the OCI helm chart in `images:` in `zarf.yaml`.
- **When** creating the package
- **Then**
   - The package cannot be created because `images:` does not support arbitrary (non-container) OCI images
      ```
      ERROR:  Failed to create package: unable to pull images after 3 attempts: error when trying to save the img                                                                   
              (ghcr.io/stefanprodan/charts/podinfo:6.4.1): mismatched fs layers (1) and diff ids (0)
      ```
   - Additionally, Zarf's mutating admission webhooks do not support mutating HelmRepository and HelmRelease resources to reference the Zarf registry

Example Flux manifests referencing an OCI Helm chart: https://fluxcd.io/flux/cheatsheets/oci-artifacts/#helm-oci

A potential issue: Zarf uses an insecure registry and Flux source-controller may not support insecure OCI registries with Helm: https://github.com/fluxcd/source-controller/issues/807

   * A little unclear on what's affected and if this issue is still current
   * OCIRepository (consumed by kustomize-controller, not helm-controller) _does_ support [`.spec.insecure`](https://fluxcd.io/flux/components/source/ocirepositories/#insecure)
   * Need to check the current status of HelmRepository and HelmReleases with an insecure localhost registry (sometimes localhost is special cased and allowed)

Zarf's built in Helm support does allow OCI, see `demo-helm-oci-chart` in this example: https://docs.zarf.dev/examples/helm-charts/

However, this does not support using Flux.

### Additional context

The scenario prompting this is a vendor only provides OCI packaged Helm charts. We do not need to modify the upstream vendor charts, and would like to copy them unmodified, and deploy them using Flux.

Additionally, Big Bang now distributes Helm charts in OCI format. If Zarf supported this, the dependency on a Gitea git repository could be dropped in many cases.

Support for generic OCI artifacts may be generally useful for other non-Flux use cases too.

cmwylie19:
## Description

This features add support for Zarf to delivery [Flux's](https://fluxcd.io/flux/cheatsheets/oci-artifacts/#helm-oci) `OCIRepo`, and `HelmRepo` that references an OCI URL. 

## Related Issue

Fixes #1974 

## Type of change

## Checklist before merging


gadido30:
can it have oci support also for ArgoCD?

cmwylie19:
> can it also have support for ArgoCD?

There is already support for Argo https://github.com/defenseunicorns/zarf/tree/main/examples/argocd

AustinAbro321:
This is still blocked on the flux side as trying to connect to the internal registry for login is still going over https when repo.spec.insecure is set to true.

We have created a PR on the flux side and are awaiting a response from them https://github.com/fluxcd/source-controller/pull/1412

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/defenseunicorns/zarf
 
-go 1.21.8
```
phillebaba, line 3: Are we bumping the Go version on purpose or is there a new dependency which requires Go 1.22?

AustinAbro321: Yes flux 1.3.0 requires go 1.22

site/src/content/docs/ref/init-package.mdx
```diff
@@ -151,13 +151,21 @@ Notably, the `REGISTRY_AFFINITY_CUSTOM` variable overrides the default pod anti-
...
 
-This allows the cluster to pull images from the Zarf Registry instead of the internet without having to modify the original image references.
+The `zarf-agent` modifies the following [flux](https://fluxcd.io/flux/) resources: [GitRepository](https://fluxcd.io/docs/components/source/gitrepositories/), [OCIRepository](https://fluxcd.io/flux/components/source/ocirepositories/), & [HelmRepository](https://fluxcd.io/flux/components/source/helmrepositories/) to point to the local Git Server or Zarf Registry. HelmRepositories are only mutated if the `type` key is set to `oci`.
 
-The `zarf-agent` also modifies [Flux GitRepository](https://fluxcd.io/docs/components/source/gitrepositories/) objects to point to the local Git Server.
+> Support for mutating OCIRepository and HelmRepository objects is in [`alpha`](/roadmap#beta) and should be tested on non-production clusters before being deployed to production clusters.
```
AustinAbro321, line 158: Calling features alpha or beta in the docs feels a bit weird since we don't have a list of what features are what maturity. Anyone have ideas suggestions on how we could be more clear on this? Should that be a page in our docs?

src/internal/agent/hooks/flux-helmrepo.go
```diff
@@ -0,0 +1,112 @@
...
+			return mutateHelmRepo(ctx, r, cluster)
+		},
+		Update: func(r *v1.AdmissionRequest) (*operations.Result, error) {
+			return mutateHelmRepo(ctx, r, cluster)
+		},
+	}
+}
+
+// mutateHelmRepo mutates the repository url to point to the repository URL defined in the ZarfState.
+func mutateHelmRepo(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster.Cluster) (result *operations.Result, err error) {
```
phillebaba, line 40: Dont use named returns.

src/internal/agent/hooks/flux-helmrepo.go
```diff
@@ -0,0 +1,112 @@
...
+	var (
+		zarfState *types.ZarfState
+		patches   []operations.PatchOperation
+	)
```
phillebaba, line 44: Why define these variables early if they will be assigned later?

src/internal/agent/hooks/flux-helmrepo_test.go
```diff
@@ -0,0 +1,166 @@
...
+					ClusterIP: "10.11.12.13",
+				},
+			},
+			code: http.StatusOK,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
```
phillebaba, line 157: Make the sub test parallel.

src/internal/agent/hooks/flux-ocirepo.go
```diff
@@ -0,0 +1,134 @@
...
+			return mutateOCIRepo(ctx, r, cluster)
+		},
+		Update: func(r *v1.AdmissionRequest) (*operations.Result, error) {
+			return mutateOCIRepo(ctx, r, cluster)
+		},
+	}
+}
+
+// mutateOCIRepo mutates the oci repository url to point to the repository URL defined in the ZarfState.
+func mutateOCIRepo(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster.Cluster) (result *operations.Result, err error) {
```
phillebaba, line 39: No named returns.

src/internal/agent/hooks/flux-ocirepo.go
```diff
@@ -0,0 +1,134 @@
...
+	var (
+		zarfState *types.ZarfState
+	)
```
phillebaba, line 42: Ditto with defining the zarf state variable here.

src/internal/agent/hooks/flux-ocirepo.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	if src.Labels != nil && src.Labels["zarf-agent"] == "patched" {
+		return &operations.Result{
+			Allowed:  true,
+			PatchOps: []operations.PatchOperation{},
+		}, nil
+	}
+
+	if zarfState, err = cluster.LoadZarfState(ctx); err != nil {
```
phillebaba, line 62: This logic here makes things a bit confusing. Because the zarfState is defined earlier the value will be assigned if error is not returned. I would much rather split the function calling and if statement to separate lines to make this more readable.

src/test/external/ext_in_cluster_test.go
```diff
@@ -123,18 +125,33 @@ func (suite *ExtInClusterTestSuite) Test_1_Deploy() {
...
-	deployArgs := []string{"package", "deploy", "../../../build/zarf-package-podinfo-flux-amd64.tar.zst", "--confirm"}
+	deployArgs := []string{"package", "deploy", filepath.Join(temp, "zarf-package-podinfo-flux-amd64.tar.zst"), "--confirm"}
 	err = exec.CmdWithPrint(zarfBinPath, deployArgs...)
 	suite.NoError(err, "unable to deploy flux example package")
 
-	// Verify flux was able to pull from the 'external' repository
-	podinfoWaitCmd := []string{"wait", "deployment", "-n=podinfo", "podinfo", "--for", "condition=Available=True", "--timeout=3s"}
-	errorStr := "unable to verify flux deployed the podinfo example"
+	// Verify flux was able to pull from the 'external' git repository
+	podinfoWaitCmd := []string{"wait", "deployment", "-n=podinfo-git", "podinfo", "--for", "condition=Available=True", "--timeout=3s"}
```
phillebaba, line 138: Instead of using wait you can use the new Kubernetes helper package.

src/pkg/cluster/tunnel.go
```diff
@@ -267,42 +267,42 @@ func (c *Cluster) findPodContainerPort(ctx context.Context, svc corev1.Service)
 }
 
 // TODO: Refactor to use netip.AddrPort instead of a string for nodePortURL.
-func serviceInfoFromNodePortURL(services []corev1.Service, nodePortURL string) (string, string, int, error) {
+func serviceInfoFromNodePortURL(services []corev1.Service, nodePortURL string) (corev1.Service, int, error) {
```
phillebaba, line 270: What is the purpose for changing the function signature?

phillebaba (changes requested):

src/pkg/cluster/tunnel.go
AustinAbro321: I need more information from the service, rather than returning yet another string it felt easier to just return a service

src/internal/agent/hooks/flux-helmrepo_test.go
AustinAbro321: I didn't make it parrallel since I'm creating a service in the last test. Though it is probably better to defer deletion of that service either way.

AustinAbro321: Just created a new client for every test

phillebaba (dismissed):

phillebaba (approved):

PR #2008, opened 2023-09-04
feat: add REGISTRY_CA_BUNDLE variable to registry package and chart

### LINKED ISSUE #2007: Support custom root CAs in registry helm chart

I would like to be able to supply a custom root ca to the registry pod.  

Add the ability to provide configmaps, volumes, and volumeMounts to the registry helm chart.

### Additional context
This is necessary for backing the registry with S3 in some partitions or using another object storage solution with a self-signed ca over HTTPS.

AbrohamLincoln:
## Description

Allows users to specify a CA bundle to the Registry when performing a `zarf init`.  

An example of how to use this to supply a Root CA Bundle in commercial AWS (although you don't need to):

Create a `zarf-config.yaml` similar to the following:

```yaml
package:
  deploy:
    set:
      REGISTRY_CA_BUNDLE: my-custom-ca.pem
```

Initialize the cluster:
`ZARF_CONFIG=./zarf-config.yaml zarf init --confirm`

## Related Issue

Fixes #2007

## Type of change

## Checklist before merging

### DIFF
--- a/packages/zarf-registry/chart/values.yaml
+++ b/packages/zarf-registry/chart/values.yaml
@@ -54,6 +54,23 @@ autoscaling:
   maxReplicas: 5
   targetCPUUtilizationPercentage: 80
 
+caBundle: ""
+## One or more concatenated certificates
+## Will be mounted to /etc/ssl/certs/ca-certificates.crt
+# caBundle: |
+#   # Root CA 1
+#   -----BEGIN CERTIFICATE-----
+#   ...
+#   -----END CERTIFICATE-----
+#   # Intermediate CA 1
+#   -----BEGIN CERTIFICATE-----
+#   ...
+#   -----END CERTIFICATE-----
+#   # Root CA 2
+#   -----BEGIN CERTIFICATE-----
+#   ...
+#   -----END CERTIFICATE-----
+
 extraEnvVars: []
 ## Additional ENV variables to set
 # - name: REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY
--- a/packages/zarf-registry/zarf.yaml
+++ b/packages/zarf-registry/zarf.yaml
@@ -47,6 +47,11 @@ variables:
     description: Enable the Horizontal Pod Autoscaler for the registry
     default: "true"
 
+  - name: REGISTRY_CA_BUNDLE
+    description: Bundle of trusted certificates to mount into the registry container
+    default: ""
+    autoIndent: true
+
   - name: REGISTRY_EXTRA_ENVS
     description: Array of additional environment variables passed to the registry container
     default: ""
--- a/packages/zarf-registry/zarf.yaml
+++ b/packages/zarf-registry/zarf.yaml
@@ -48,9 +48,10 @@ variables:
     default: "true"
 
   - name: REGISTRY_CA_BUNDLE
-    description: Bundle of trusted certificates to mount into the registry container
+    description: Filepath to a bundle of trusted certificates to mount into the registry container
     default: ""
     autoIndent: true
+    type: file
 
   - name: REGISTRY_EXTRA_ENVS
     description: Array of additional environment variables passed to the registry container
--- a/packages/zarf-registry/chart/templates/_helpers.tpl
+++ b/packages/zarf-registry/chart/templates/_helpers.tpl
@@ -22,3 +22,15 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Merge all configmaps
+*/}}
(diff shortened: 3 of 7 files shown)


packages/zarf-registry/chart/values.yaml
```diff
@@ -58,3 +58,24 @@ extraEnvVars: []
...
+#    readOnly: true
+
+extraConfigMaps: []
+## Additional configMaps to create in the release namespace
+# - name: custom-ca-bundle
+#   data:
+#     ca-certificates.crt: |
+#       -----BEGIN CERTIFICATE-----
+#       ...
+#       -----END CERTIFICATE-----
```
Racer159, line 81: Unless there is another use case you had in mind for this it may be better / more user friendly to just accept the certificate bundle as text without providing direct access to the volumes / config maps, etc (which requires more k8s knowledge).  It may also be better to do it with a `subPath` that way the volume mount places the single file in the volume without removing the other CA certs that may be there.

Racer159 (changes requested):

packages/zarf-registry/chart/values.yaml
```diff
@@ -54,7 +54,43 @@ autoscaling:
...
+#  - mountPath: /my/custom/file/path/
+#    name: my-custom-volume
+
+extraConfigMaps: []
+## Additional configMaps to create in the release namespace
+# - name: my-custom-configmap
+#   data:
+#     my-file.txt: |
+#       foo
+#       bar
```
Racer159, line 96: Just to keep this simpler I would rather not expose these unless there is a common reason to do so (people can always make a custom init package if they _really_ need something special).  The default init package is geared toward IT admins who have less k8s experience and these options may end up being confusing (since they will show up in package inspection and are settable at deploy time).  We are working to get these personas published as well for external contributors to see (like GitLab does) but haven't completed that yet.

Racer159 (changes requested):

packages/zarf-registry/chart/values.yaml
Racer159: Sorry for the delayed response the REGISTRY_CA_BUNDLE variable is what I was asking for - on the subPath the other CAs are in nearly all cases likely not needed I just usually prefer to add rather than replace to provide more flexibility

packages/zarf-registry/chart/values.yaml
Racer159: [ashton_gru.pdf](https://github.com/defenseunicorns/zarf/files/12579047/ashton_gru.pdf)

packages/zarf-registry/zarf.yaml
```diff
@@ -47,6 +47,11 @@ variables:
     description: Enable the Horizontal Pod Autoscaler for the registry
     default: "true"
 
+  - name: REGISTRY_CA_BUNDLE
+    description: Bundle of trusted certificates to mount into the registry container
+    default: ""
+    autoIndent: true
```
Racer159, line 53:
```suggestion
    description: Filepath to a bundle of trusted certificates to mount into the registry container
    default: ""
    autoIndent: true
    type: file
```

After testing this a bit more locally everything looked pretty good!  One thing that _may_ make sense though (want to hear what this looks like in your use case) would be using the `type` field to pull in a file from a filename rather than a CA bundle as a multiline field.  This would provide the following benefits:

1. On super large CA bundles (an edge case but still possible) actions could fail due to exceeding the maximum size for environment variables (on Windows this is 32,767 characters)
2. It would allow using a CA bundle in the more familiar pem format rather than having to template it into a config file yourself manually.

Might not make sense if you aren't getting a file back in your use case, and requires an additional file to manage but may be worth it.  This would also require an added check to change line `194` of `src/pkg/utils/io.go` to prevent an erroneous warning on an empty string (which will never be a valid file)

```
				if template.Type == types.FileVariableType && value != "" {
```

Racer159: Just following up to see if you had thoughts on the above - I think it may be a bit cleaner but want to make sure that also makes sense for your use case.

Racer159: No problem at all, just wanted to make sure you had what you needed.

Racer159: Just checking back on this one - if needed we can open a new PR (maintaining your authorship) to help get this across the finish line (assuming you are good with the type: file suggestion)

Noxsios (approved):

Racer159 (approved):
lgtm

Racer159 (changes requested):
Code looks good but looks like there are a few commits that aren't signed

Noxsios (approved):

Racer159 (approved):
Lgtm

PR #2012, opened 2023-09-11
Adjust timeout messaging to be more correct on timeout errors with 0 retries

Racer159:
## Description

This fixes an issue where the timeout message wouldn't show if retries was set to 0.

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging


Noxsios (approved):
lgtm, for anyone who may be confused like I was at first:

In go `break` and `continue` can optionally use a label to change the code execution order of nested loops.

`retryCmd:` creates essentially a named for-loop that `break` and `continue` can interact with, restarting/ending the loop as desired.

PR #2016, opened 2023-09-12
Initial implementation of pattern

### LINKED ISSUE #1997: Variable validation by a regex `pattern`
It would be nice to have validation for variables to ensure they meet certain parameters (I.e. being numbers for ports, or simply not being an empty string) so that packages can fail early when bad variables are given.

- **Given** I have a deploy time variable
- **And** have specified a regex in a new `pattern` key
- **When** The package is deployed
- **Then** Zarf will validate the provided value against the pattern
- **And** the deployment will fail or pass based on the result

We could have a more generic field like `required` but `pattern` allows more flexibility for control of validation.

### Additional context
This is something that has been requested to reduce the chance for errors when a package creator is not the package deployer.

Racer159:
## Description

This PR introduces a new variable key that checks a set value against a regex pattern before being allowed to proceed.

## Related Issue

Fixes #1997

## Type of change

## Checklist before merging

### DIFF
--- a/docs/3-create-a-zarf-package/4-zarf-schema.md
+++ b/docs/3-create-a-zarf-package/4-zarf-schema.md
@@ -2213,6 +2213,22 @@ Must be one of:
 
 <details>
 <summary>
+<strong> <a name="components_items_actions_onCreate_before_items_setVariables_items_pattern"></a>pattern</strong>
+</summary>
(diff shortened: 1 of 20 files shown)


cmwylie19 (approved):
lgtm

PR #2025, opened 2023-09-21
Decouples agent image build process in zarf repository specific makefile

dgershman:
## Description

This change decouples the specific logic about whether or not `AGENT_IMAGE_TAG` is set to `local` or not, as the description indicates.  When you build the package outside the zarf repository, you get an error message stating it can't find the make target.  The [logic in the Makefile](https://github.com/defenseunicorns/zarf/blob/main/Makefile#L105-L106) is such that it would determine whether not this is set as such.  By adding this change, you can use the `zarf-agent` package unchanged if wanted, potentially integrating improvements a little easier to track.

## Related Issue

N/A

## Type of change

## Checklist before merging


packages/zarf-agent/zarf.yaml
```diff
@@ -30,7 +30,7 @@ components:
     actions:
       onCreate:
         before:
-          - cmd: "make init-package-local-agent AGENT_IMAGE_TAG=\"###ZARF_PKG_TMPL_AGENT_IMAGE_TAG###\" ARCH=\"###ZARF_PKG_ARCH###\""
+          - cmd: "test \"###ZARF_PKG_TMPL_AGENT_IMAGE_TAG###\" != \"local\" || make init-package-local-agent AGENT_IMAGE_TAG=\"###ZARF_PKG_TMPL_AGENT_IMAGE_TAG###\" ARCH=\"###ZARF_PKG_ARCH###\""
```
Racer159, line 33:
```suggestion
          - cmd: "test \"###ZARF_PKG_TMPL_AGENT_IMAGE_TAG###\" != \"local\" || make build-local-agent-image AGENT_IMAGE_TAG=\"###ZARF_PKG_TMPL_AGENT_IMAGE_TAG###\" ARCH=\"###ZARF_PKG_ARCH###\""
```

Racer159: It would be better if we are moving the check up here to point to the `build-local-agent-image` Make target instead - the `init-package-local-agent` shim target should also be removed since it was only for this check and isn't used elsewhere.

Racer159 (changes requested):

packages/zarf-agent/zarf.yaml
```diff
@@ -30,7 +30,7 @@ components:
     actions:
       onCreate:
         before:
-          - cmd: "make init-package-local-agent AGENT_IMAGE_TAG=\"###ZARF_PKG_TMPL_AGENT_IMAGE_TAG###\" ARCH=\"###ZARF_PKG_ARCH###\""
+          - cmd: "test \"###ZARF_PKG_TMPL_AGENT_IMAGE_TAG###\" != \"local\" || make build-local-agent-image AGENT_IMAGE_TAG=\"###ZARF_PKG_TMPL_AGENT_IMAGE_TAG###\" ARCH=\"###ZARF_PKG_ARCH###\""
```
Racer159, line 33: Hehehe - `The token '||' is not a valid statement separator in this version.`

```suggestion
          - cmd: "test \"###ZARF_PKG_TMPL_AGENT_IMAGE_TAG###\" != \"local\" || make build-local-agent-image AGENT_IMAGE_TAG=\"###ZARF_PKG_TMPL_AGENT_IMAGE_TAG###\" ARCH=\"###ZARF_PKG_ARCH###\""
            shell:
              windows: cmd
```

Setting a preferred windows shell should fix it.

Racer159 (changes requested):

packages/zarf-agent/zarf.yaml
Racer159: (although `test` may not exist on GH runners since it is not default in Windows)

Racer159: Following up on this `mingw` is available on the runners so it should work - https://github.com/actions/runner-images/blob/main/images/win/Windows2019-Readme.md#tools

Racer159 (approved):
lgtm

Noxsios (approved):

Issue #475, opened 2022-05-06
Images should Include Signatures/SBOMS/Attestations


runyontr:
**Is your feature request related to a problem? Please describe.**
As a system owner, I would like to obtain all artifacts for an image produced by IronBank and other image vendors.

**Describe the solution you'd like**

As an end user, when an image is included as part of the Zarf.yaml, I would like the associated objects to be included in the Zarf bundle:

```bash
$ cosign tree registry1.dso.mil/ironbank/redhat/ubi/ubi8:8.5
📦 Supply Chain Security Related artifacts for an image: registry1.dso.mil/ironbank/redhat/ubi/ubi8:8.5
└── 💾 Attestations for an image tag: registry1.dso.mil/ironbank/redhat/ubi/ubi8:sha256-20cc7b5f918a4b545093e84f8d514d18734577e4eac74acedec2e6285976150a.att
   └── 🍒 sha256:c4825a56acbe4b241e34ad741bece768cf6f6dd9665a7a5e6747189df9c08b23
└── 🔐 Signatures for an image tag: registry1.dso.mil/ironbank/redhat/ubi/ubi8:sha256-20cc7b5f918a4b545093e84f8d514d18734577e4eac74acedec2e6285976150a.sig
   └── 🍒 sha256:7d5ed1fe5df8e5fb32c35ab9c0c8b09a5cd30f20683acc73e67a4d7dfbb4e884
└── 📦 SBOMs for an image tag: registry1.dso.mil/ironbank/redhat/ubi/ubi8:sha256-20cc7b5f918a4b545093e84f8d514d18734577e4eac74acedec2e6285976150a.sbom
```

Currently they provide vulnerability information in the attestation as can be seen here:

```bash
cosign download attestation registry1.dso.mil/ironbank/redhat/ubi/ubi8:8.5 | jq -r .payload | base64 --decode | jq .
```

**Describe alternatives you've considered**
The Zarf.yaml could define each one of these objects individually, which seems time consuming, but very explicit about which version of the image should be downloaded.

**Additional context**
IronBank is currently providing SBOM, signatures and justification, but as they expand the metadata around the image they could/may provide it would be good to dynamically bring more of the data with in Zarf without explicit reference changes.

https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/tree.go

jeff-mccoy:
[...]

PR #2027, opened 2023-09-25
feat: support cosign signatures / attestations and discover in `zarf prepare find-images`

mjnagel:
## Description

Output from @rjferguson21 and my dash day's explorations of this.

This PR includes:
- Logic pulled from [cosign tree](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/tree.go#L50) to detect and include signatures/attestations when using `zarf prepare find-images`
- Skips SBOMs and caching for "non-images" by checking layer mediatypes at pull time

Several TODOs for follow-on work based on the issue/other needs:
- Add in SBOM pulling when available, skip over syft creation of SBOM when applicable.
- Mutating digests as needed - in order for tools to identify the signature the tag for it must be tagged `<digest>.sig`. When zarf does its [`AddImageAnnotation`](https://github.com/defenseunicorns/zarf/blob/main/src/pkg/utils/image.go#L43) this could change the image digest, making it so that the signature is no longer at the correct tag. Images already having that annotation are unaffected (which is why this works OK for Ironbank).

## Related Issue

Relates to https://github.com/defenseunicorns/zarf/issues/475

## Type of change

## Checklist before merging


Racer159 (changes requested):
I know it is still draft but figured I would add some initial comments

src/pkg/packager/create.go
```diff
@@ -90,6 +92,64 @@ func (p *Packager) Create(baseDir string) error {
...
+				ref, err := name.ParseReference(image, nameOpts...)
+
+				if err != nil {
+					return err
+				}
+
+				var remoteOpts []ociremote.Option
+				simg, err := ociremote.SignedEntity(ref, remoteOpts...)
+				if err != nil {
+					return err
```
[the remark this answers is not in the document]
mjnagel: Interesting, I'll have to test this again. We had packages that were built with both signed and non-signed images and didn't fail create.

src/cmd/common/viper.go
```diff
@@ -67,6 +67,7 @@ const (
 	VPkgCreateSbom               = "package.create.sbom"
 	VPkgCreateSbomOutput         = "package.create.sbom_output"
 	VPkgCreateSkipSbom           = "package.create.skip_sbom"
+	VPkgCreateSkipSigs           = "package.create.skip_sigs"
```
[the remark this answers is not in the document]
mjnagel: Renamed, this makes sense to me.

src/pkg/packager/create.go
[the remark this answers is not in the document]
mjnagel: Not sure why this happens in some cases...random dockerhub images seem to work fine despite no signatures. I think I got it fixed by just doing a `continue` if the `SignedEntity` is `nil`. Let me know if you see any other issues with it.

examples/cosign-signatures/zarf.yaml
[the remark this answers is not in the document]
mjnagel: Moved - I added a basic test that checks to ensure the number of images post create is 3, using a chainguard image with a signature and attestation. Not sure if we might want to use a different image that zarf has control over (agent image?) but I thought it might be useful to test attestation pulling, and the agent image has no attestations currently.

Also not sure if the test I added is the best way to test this - I saw a similar test being used for the differential package so hopefully alright.

src/internal/packager/images/pull.go
```diff
@@ -486,8 +486,11 @@ func (i *ImageConfig) PullImage(src string, spinner *message.Spinner) (img v1.Im
 	}
 
 	spinner.Updatef("Preparing image %s", src)
-	imageCachePath := filepath.Join(config.GetAbsCachePath(), layout.ImagesDir)
-	img = cache.Image(img, cache.NewFilesystemCache(imageCachePath))
+
+	if !strings.HasSuffix(src, ".sig") && !strings.HasSuffix(src, ".att") {
```
Racer159, line 490: At this point you have access to the `img` object and it's `MediaType()` which you should be able to check to see if this is actually an image manifest or not (instead of fuzzily checking the "extension")

src/internal/packager/sbom/catalog.go
```diff
@@ -71,6 +72,11 @@ func Catalog(componentSBOMs map[string]*layout.ComponentSBOM, imageList []transf
 	for _, refInfo := range imageList {
 		builder.spinner.Updatef("Creating image SBOMs (%d of %d): %s", currImage, imageCount, refInfo.Reference)
 
+		if strings.HasSuffix(refInfo.Tag, ".sig") || strings.HasSuffix(refInfo.Tag, ".att") {
```
Racer159, line 75: You could try again looking at the `img` object's MediaType that is returned from `utils.LoadOCIImage` to check if we should skip or not.

Racer159: We may also want to say "non-image artifact" as well since that check wouldn't be cosign specific anymore

src/pkg/packager/create.go
```diff
@@ -90,6 +93,50 @@ func (p *Packager) Create() (err error) {
...
+				if err == nil {
+					layers, _ := atts.Layers()
+					if len(layers) > 0 {
+						imageSigList = append(imageSigList, attRef.String())
+					}
+				}
+			}
+			p.cfg.Pkg.Components[i].Images = append(p.cfg.Pkg.Components[i].Images, imageSigList...)
+		}
+	}
```
Racer159, line 138: This should be in a helper (and probably really 2 helpers).

- The first that deals with looping through images should be a receiver on `Packager` that handles the control flow of adding the signatures and attestations to the package.
- The second should be in `src/utils/sget` (which should be renamed `src/utils/cosign` to do the signature and attestations for a given image reference.

Racer159 (changes requested):

src/pkg/packager/create.go
mjnagel: I think I got this adjusted in the latest commits - I wasn't sure where to place the first one so I just left it in create for now.

src/pkg/packager/create.go
```diff
@@ -776,3 +789,20 @@ func (p *Packager) removeCopiesFromDifferentialPackage() error {
 
 	return nil
 }
+
+func (p *Packager) addExtraArtifacts() (err error) {
+	for i, c := range p.cfg.Pkg.Components {
+		var extraArtifactList []string
+		for _, image := range c.Images {
+			if !p.cfg.CreateOpts.SkipCosignLookup {
```
Racer159, line 797: This if statement should probably wrap the whole for loop.

src/pkg/utils/cosign.go
```diff
@@ -217,3 +217,38 @@ func CosignSignBlob(blobPath string, outputSigPath string, keyPath string, passw
 
 	return sig, err
 }
+
+func GetCosignArtifacts(image string) (cosignList []string, err error) {
```
Racer159, line 221: lint

src/pkg/utils/cosign.go
```diff
@@ -217,3 +217,38 @@ func CosignSignBlob(blobPath string, outputSigPath string, keyPath string, passw
...
+		return cosignArtifactList, err
+	}
+
+	var remoteOpts []ociremote.Option
+	simg, _ := ociremote.SignedEntity(ref, remoteOpts...)
+	if simg == nil {
+		return cosignArtifactList, nil
+	}
+	sigRef, _ := ociremote.SignatureTag(ref, remoteOpts...)
+	attRef, _ := ociremote.AttestationTag(ref, remoteOpts...)
```
Racer159, line 236: It is probably ok to dogsled the error here given these functions always return a struct (which won't be `nil`), but would be worth a comment.

src/test/packages/10-cosign-lookup/zarf.yaml
Racer159, line 1: It may be good to just use the existing tests since they already look for the .sig files - doing this will add another image and cost ~130MiBs which is a lot for our tests that are heavily disk constrained.

src/pkg/packager/create.go
```diff
@@ -182,10 +188,17 @@ func (p *Packager) Create() (err error) {
 			return fmt.Errorf("unable to pull images after 3 attempts: %w", err)
 		}
 
-		for _, img := range pulled {
+		for transform, img := range pulled {
 			if err := p.layout.Images.AddV1Image(img); err != nil {
 				return err
 			}
+			isImage, err := utils.HasImageLayers(img)
```
Racer159, line 195: How expensive is this operation? If it is fast it might be ok to do it twice but if it takes some time (even a seemingly small amount) it could be worth returning from the pull instead.

Racer159 (changes requested):

Racer159 (commented):
It would be good to also update the fuzzy image regex to this while you touch the find images code:

`["|=]([a-z0-9\-.\/:]+:[\w.\-]*[a-z\.\-][\w.\-]*)"`

Welcome feedback on whether it makes sense and the full context is here: https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1696639896292219?thread_ts=1696585484.114289&cid=C03B6BJAUJ3 (edited)

src/test/packages/10-cosign-artifacts/zarf.yaml
```diff
@@ -0,0 +1,9 @@
+kind: ZarfPackageConfig
+metadata:
+  name: cosign-artifacts
+  description: Test package for cosign image artifact pulls
+
+components:
+  - name: test
+    images:
+      - ghcr.io/defenseunicorns/zarf/agent:sha256-90863c246da361499e8f59dfae728c34b50ee2057e61e06dacdcfad983425c32.sig
```
Racer159, line 9: Can we integrate this into a podinfo example instead (they also sign their image): https://github.com/stefanprodan/podinfo/pkgs/container/podinfo (that may keep things simpler and more discoverable if it is just in another example and it has a comment explaining why it is there)

src/pkg/packager/prepare.go
```diff
@@ -249,6 +259,12 @@ func (p *Packager) FindImages() (imgMap map[string][]string, err error) {
 					// Otherwise, add to the list of images
 					message.Debugf("Imaged digest found: %s", descriptor.Digest)
 					validImages = append(validImages, image)
+					cosignArtifacts, err := utils.GetCosignArtifacts(image)
+					if err != nil {
+						message.WarnErrf(err, "Problem looking up cosign artifacts for %s: %s", image, err.Error())
+						erroredCosignLookups = append(erroredCosignLookups, image)
+					}
+					validImages = append(validImages, cosignArtifacts...)
```
Racer159, line 267: Instead of copying this code twice could we do it like the maybes as a separate section / loop to keep the logic self-contained?

src/internal/packager/images/pull.go
```diff
@@ -426,15 +430,23 @@ func (i *ImageConfig) PullAll() (map[transform.Image]v1.Image, error) {
...
 		spinner.Updatef("Reading image tarball: %s", src)
-		return crane.Load(src, config.GetCraneOptions(true, i.Architectures...)...)
+		img, err = crane.Load(src, config.GetCraneOptions(true, i.Architectures...)...)
+		if err != nil {
+			return nil, false, err
+		}
+		hasImageLayers, err = utils.HasImageLayers(img)
+		if err != nil {
+			return nil, false, fmt.Errorf("failed to check image %s layer mediatype: %w", src, err)
+		}
```
Racer159, line 448: I know that caching causes issues with these artifacts but there may be ways to simplify the logic to not have to state this check 3 times.

Racer159 (changes requested):

src/test/packages/10-cosign-artifacts/zarf.yaml
mjnagel: Moved to the podinfo helm-charts example and removed the separate test for this.

src/pkg/packager/prepare.go
mjnagel: Moved to a separate loop, hopefully accomplishes this in the same way, also added a comment differentiating these as cosign artifacts.

src/internal/packager/images/pull.go
mjnagel: I think I tackled this - mapping out the logic I believe the 3 separate spots were effectively an if/else-if/else so I wrapped them in that structure and moved to a single return statement for all success cases.

src/pkg/utils/cosign.go
```diff
@@ -217,3 +217,40 @@ func CosignSignBlob(blobPath string, outputSigPath string, keyPath string, passw
...
+		}
+	}
+
+	atts, err := simg.Attestations()
+	if err == nil {
+		layers, _ := atts.Layers()
+		if len(layers) > 0 {
+			cosignArtifactList = append(cosignArtifactList, attRef.String())
+		}
+	}
```
Racer159, line 254:
```suggestion
	sigs, err := simg.Signatures()
	if err != nil {
		return cosignArtifactList, err
	}
	layers, err := sigs.Layers()
	if err != nil {
		return cosignArtifactList, err
	}
	if len(layers) > 0 {
		cosignArtifactList = append(cosignArtifactList, sigRef.String())
	}

	atts, err := simg.Attestations()
	if err != nil {
		return cosignArtifactList, err
	}
	layers, err := atts.Layers()
	if err != nil {
		return cosignArtifactList, err
	}
	if len(layers) > 0 {
		cosignArtifactList = append(cosignArtifactList, attRef.String())
	}
```

Racer159: I think we should return the errors where applicable.

Racer159 (changes requested):

Noxsios (approved):
lgtm now the error was handled

Racer159 (approved):

Issue #1814, opened 2023-06-13
Migration of the Zarf UI to UDS CLI


Racer159:
### Describe what should be investigated or refactored
Currently the Zarf UI is available under the `zarf dev ui` command however Zarf is not likely a good long term home for the UI since it runs from a go binary that is not easy to get running inside of a server and doesn't mesh well with the Zarf CLI lifecycle.

We should migrate the Zarf UI over to UDS CLI and potentially as a first step move the existing code into a separate repository that imports Zarf as a library.

### Links to any relevant code
https://github.com/defenseunicorns/zarf/blob/main/src/cmd/internal.go#L145

### Additional context
The Zarf UI is currently in beta and some community folks use it so it shouldn't be discarded and should have a clear migration path for them.

PR #2042, opened 2023-09-30
Mark the Zarf UI as [Deprecated] and update tunnel logic for library use

Racer159:
## Description

This PR marks the Zarf UI as deprecated within Zarf as its transition over toward UDS CLI begins.  It also points to the new https://github.com/defenseunicorns/zarf-ui repository that will be the new temporary home for the Zarf UI using Zarf as a library.

## Related Issue

Relates to #1814

## Type of change

## Checklist before merging


Noxsios (approved):
lgtm but I would update the description if not already done so to also highlight some of the tweaks on `tunnel` (like `exec.SupressGlobalInterrupt`) so git history is a bit more expressive.

PR #2047, opened 2023-10-03
Publish / pull UX improvement and minor fixes

### LINKED ISSUE #2044: Improve OCI pull / publish UX
The current UX for publish and pull is ok for showing progress but puts a lot of content on the screen - in `zarf package pull` and `zarf package publish` this can be tolerable, however during an init or deploy process it introduces a lot of output to both the screen and the logs.

- **Given** I have a package that I would like to pull or publish
- **When** I execute a command that will pull or publish that artifact
- **Then** I will see a streamlined pull process UX that no longer prints layers information but still has a progress bar

image

We could switch to a spinner that shows layer information but this is not something that is really necessary for the user.

### Additional context
We should still try to be helpful in the debug logs but should be careful to not fall back to the old UX because this could still inflate the size of the log file making it difficult to find useful information.

Racer159:
## Description

This improves the pull user experience and also introduces minor fixes for other UX bugs.

## Related Issue

Fixes #2044 

## Type of change

## Checklist before merging

### DIFF
--- a/src/extensions/bigbang/bigbang.go
+++ b/src/extensions/bigbang/bigbang.go
@@ -268,6 +268,26 @@ func Skeletonize(tmpPaths *layout.ComponentPaths, c types.ZarfComponent) (types.
 		c.Extensions.BigBang.ValuesFiles[valuesIdx] = rel
 	}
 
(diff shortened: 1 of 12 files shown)


Noxsios (changes requested):
one minor tweak

Noxsios (dismissed):

PR #2049, opened 2023-10-03
Move find port function into library code

Racer159:
## Description

This moves a helper function for finding a containers port from a service pod into library code to be used by other binaries.

## Related Issue

Relates to #1814 

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/k8s/pods.go
+++ b/src/pkg/k8s/pods.go
@@ -161,7 +161,9 @@ func (k *K8s) WaitForPodsAndContainers(target PodLookup, include PodFilter) []co
 	return []corev1.Pod{}
 }
 
-// FindPodContainerPort will find a pod's container from a service and return it.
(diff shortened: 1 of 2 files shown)


src/pkg/k8s/pods.go
```diff
@@ -160,3 +160,25 @@ func (k *K8s) WaitForPodsAndContainers(target PodLookup, include PodFilter) []co
 
 	return []corev1.Pod{}
 }
+
+// FindPodContainerPort will find a pod's container from a service and return it.
```
Noxsios, line 164:
```suggestion
// FindPodContainerPort will find a pod's container port from a service and return it.
//
// Returns 0 if no port is found.
```

Noxsios (changes requested):

Noxsios (approved):

Issue #1845, opened 2023-06-23
Multi-layered imports do not adjust filepaths correctly


mjnagel:
### Environment
Device and OS: MBP, macOS
App version: 0.27.1
Kubernetes distro being used: N/A

### Steps to reproduce
1. Clone https://github.com/mjnagel/layers-of-cake
2. `zarf package create`
3. Observe during confirmation that only the "bottom" import has its paths adjusted correctly.

### Expected result

All layers of imports should have filepaths adjusted to be relative to the import path.

### Actual Result

Only the last imported component will be adjusted correctly. Note that this was reproduced with both 3 & 4 layers, and I assume this could be extended indefinitely.

### Visual Proof (screenshots, videos, text, etc)

```
kind: ZarfPackageConfig
metadata:
  name: parent
components:
- name: file
  required: true
  files:
  - source: leaf/lfile.txt
    target: lfile.txt
  - source: cfile.txt
    target: cfile.txt
  - source: sfile.txt
    target: sfile.txt
  - source: file.txt
    target: file.txt
```

### Severity/Priority

No clear workaround outside of symlinks, etc to make paths work.

### Additional Context

This was originally seen with the bigbang extension values files, but it seems to affect all places where imported components need to adjust filepaths.

Issue #1827, opened 2023-06-15
Inception: Importing a local component when local component imports a URL


benmountjoy111:
### Environment
Device and OS: MacOS Ventura 13.2.1
App version: zarf v0.27.1
Kubernetes distro being used: N/A
Other:

### Steps to reproduce

```bash
# create zarf.yaml (to be locally imported)
mkdir bigbang

cat <<EOF> bigbang/zarf.yaml
kind: ZarfPackageConfig
metadata:
  name: dubbd
  description: "Defense Unicorns BigBang Distro"
  architecture: amd64

components:
  - name: load-certs
    required: true
    import:
      name: load-certs
      url: oci://ghcr.io/defenseunicorns/packages/dubbd:0.2.2-skeleton

  - name: preflight
    required: true
    import:
      name: preflight
      url: oci://ghcr.io/defenseunicorns/packages/dubbd:0.2.2-skeleton
EOF

# create top-level zarf.yaml which will import components from the one above
cat <<EOF> zarf.yaml
kind: ZarfPackageConfig
metadata:
  name: parent
  architecture: amd64

components:
  - name: load-certs
    required: true
    import:
      path: './bigbang/'

  - name: preflight
    required: true
    import:
      path: './bigbang/'
EOF

# create parent package
zarf package create
# Observe no contents for load-certs and preflight (first screenshot)

# create sub package
cd bigbang
zarf package create
# Observe populated/correct contents for load-certs and preflight (second screenshot)

```

### Expected result
Doing a 'zarf package create', which includes a locally imported component (which the locally imported component then imports a url) will correctly include the url's component's contents.

### Actual Result
Doing a 'zarf package create', which includes a locally imported component (which the locally imported component then imports a url), creates 'empty' components.

### Visual Proof (screenshots, videos, text, etc)
![Screenshot 2023-06-15 at 2 10 42 PM](https://github.com/defenseunicorns/zarf/assets/118748385/c2ca494b-6efb-49f4-89b2-937b4d255b91)

![Screenshot 2023-06-15 at 2 11 05 PM](https://github.com/defenseunicorns/zarf/assets/118748385/23542236-3837-4443-bed2-086b3d43b3d3)

PR #2058, opened 2023-10-07
chore: refactor Zarf Component composition

### LINKED ISSUE #2037: Refactor component composability and skeletons to address bugs and update to newer patterns
There have been a few bugs around component compose that need to be addressed as well as some improved:

Bugs:

Fixes https://github.com/defenseunicorns/zarf/issues/1827
Fixes https://github.com/defenseunicorns/zarf/issues/1845

Refactors:
Move skeleton to package build data so that we can tell what is a skeleton package when it is outside of an OCI ref
Simplify the interactions with the new `layout` struct from the provider / sources refactor.

https://github.com/defenseunicorns/zarf/blob/main/src/pkg/packager/compose.go

### Additional context
This would greatly simplify component composability and make is more understandable.

Noxsios:
## Description

```go
chain, err := composer.NewImportChain(pkg.Components[0], "amd64")
if err != nil {
    panic(err)
}

fmt.Println(chain.String())
```

results in:

```
component "k3s" imports "k3s" in packages/distros/k3s, which imports "k3s" in common
```

## Related Issue

Fixes #2037 
Fixes https://github.com/defenseunicorns/zarf/issues/1827
Fixes https://github.com/defenseunicorns/zarf/issues/1845

## Type of change

## Checklist before merging


src/pkg/packager/composer/list.go
```diff
@@ -0,0 +1,173 @@
...
+func (ic *ImportChain) append(c types.ZarfComponent) {
+	node := &Node{ZarfComponent: c, prev: nil, next: nil}
+	if ic.head == nil {
+		ic.head = node
+		ic.tail = node
+	} else {
+		p := ic.head
+		for p.next != nil {
+			p = p.next
+		}
```
Racer159, line 41: You should be able to do this based on the tail pointer instead of traversing forward through the list like this.

src/pkg/packager/composer/list.go
```diff
@@ -0,0 +1,173 @@
...
+			return ic, fmt.Errorf("detected malformed import chain, cannot import remote components from remote components")
+		}
+
+		var pkg types.ZarfPackage
+		name := node.Name
+
+		if isLocal {
+			history = append(history, node.Import.Path)
+			paths := append(history, layout.ZarfYAML)
+			if err := utils.ReadYaml(filepath.Join(paths...), &pkg); err != nil {
```
Racer159, line 77: This would be nice to do with a helper on the linked list instead of managing it in a separate array.

src/pkg/packager/composer/list.go
```diff
@@ -0,0 +1,173 @@
...
+			} else if isRemote {
+				return ic, fmt.Errorf("component %q not found in %q", name, node.Import.URL)
+			}
+		}
+
+		if node.Only.Cluster.Architecture != "" {
+			arch = node.Only.Cluster.Architecture
+		}
+
+		if arch != "" && found.Only.Cluster.Architecture != "" && found.Only.Cluster.Architecture != arch {
```
Racer159, line 111: This isn't quite how this should work - the imported package may have multiple components of the same name with different archs so this logic should be in the found function instead.  See this for an example:

https://github.com/Racer159/zarf-package-k3d/blob/master/install/linux/zarf.yaml

Racer159 (changes requested):

src/pkg/packager/composer/list.go
Noxsios: uhh https://github.com/defenseunicorns/zarf/blob/6a04e7783d1308cb22eda22a1ebca4dcbdcdf2e0/src/internal/packager/validate/validate.go#L56

src/pkg/packager/compose.go
```diff
@@ -5,416 +5,60 @@
...
-			return child, fmt.Errorf("unable to pull skeleton from %s: %w", skelURL, err)
-		}
-		cwd, err := os.Getwd()
-		if err != nil {
-			return child, fmt.Errorf("unable to get current working directory: %w", err)
+	for _, component := range p.cfg.Pkg.Components {
+		arch := p.arch
+		// filter by architecture if set, otherwise use system architecture
+		if component.Only.Cluster.Architecture != "" {
+			arch = component.Only.Cluster.Architecture
```
Racer159, line 24: This check should just be filtered out - if we have a component that doesn't match the arch we are building for there is no reason to compose it.

src/pkg/packager/composer/list.go
```diff
@@ -0,0 +1,420 @@
...
+		return nil
+	}
+
+	tu := archiver.Tar{
+		OverwriteExisting: true,
+		// removes /<name>/ from the paths
+		StripComponents: 1,
+	}
+	return tu.Unarchive(tb, dir)
+}
```
Racer159, line 350: These OCI helpers would likely be better served in a dedicated `oci.go` or `skeleton.go` to keep the package more organized.

src/pkg/packager/composer/list.go
```diff
@@ -0,0 +1,420 @@
...
+		}
+
+		ctx := context.TODO()
+		// ensure the tarball is in the cache
+		exists, err := store.Exists(ctx, componentDesc)
+		if err != nil {
+			return err
+		} else if !exists {
+			copyOpts := remote.CopyOpts
+			// TODO: investigate why the default FindSuccessors in CopyWithProgress is not working
```
Racer159, line 309: todo

src/pkg/packager/composer/list.go
```diff
@@ -0,0 +1,420 @@
...
+			copyOpts.FindSuccessors = content.Successors
+			if err := remote.CopyWithProgress([]ocispec.Descriptor{componentDesc}, store, copyOpts, cache); err != nil {
+				return err
+			}
+			exists, err := store.Exists(ctx, componentDesc)
+			if err != nil {
+				return err
+			} else if !exists {
+				return fmt.Errorf("failed to fetch remote component: %+v", componentDesc)
+			}
```
Racer159, line 319: Does CopyWithProgress not return an error if it failed to pull?

src/pkg/packager/composer/list.go
```diff
@@ -0,0 +1,420 @@
...
+	ic.tail.relativeToHead = rel
+
+	if oci.IsEmptyDescriptor(componentDesc) {
+		// nothing was fetched, nothing to extract
+		return nil
+	}
+
+	tu := archiver.Tar{
+		OverwriteExisting: true,
+		// removes /<name>/ from the paths
```
Racer159, line 346:
```suggestion
		// removes /<component-name>/ from the paths
```

src/pkg/packager/composer/list.go
```diff
@@ -0,0 +1,420 @@
...
+	for node != nil {
+		fixPaths(&node.ZarfComponent, node.relativeToHead)
+
+		// perform overrides here
+		overrideMetadata(&composed, node.ZarfComponent)
+		overrideDeprecated(&composed, node.ZarfComponent)
+		overrideResources(&composed, node.ZarfComponent)
+		overrideActions(&composed, node.ZarfComponent)
+
+		composeExtensions(&composed, node.ZarfComponent, node.relativeToHead)
```
Racer159, line 382:
```suggestion
	// start with an empty component to compose into
	composed = types.ZarfComponent{}

	// start overriding with the tail node
	node := ic.tail
	for node != nil {
		fixPaths(&node.ZarfComponent, node.relativeToHead)

		// perform overrides here
		overrideMetadata(&composed, node.ZarfComponent)
		overrideDeprecated(&composed, node.ZarfComponent)
		overrideResources(&composed, node.ZarfComponent)
		overrideActions(&composed, node.ZarfComponent)

		composeExtensions(&composed, node.ZarfComponent, node.relativeToHead)

		node = node.prev
	}
```
This logic can be simplified (and is wrong for extensions) - this needs a test.

src/pkg/packager/composer/extensions.go
```diff
@@ -0,0 +1,28 @@
...
+
+	// perform any overrides
+	if override.Extensions.BigBang != nil {
+		if override.Extensions.BigBang.ValuesFiles != nil {
+			c.Extensions.BigBang.ValuesFiles = append(c.Extensions.BigBang.ValuesFiles, override.Extensions.BigBang.ValuesFiles...)
+		}
+		if override.Extensions.BigBang.FluxPatchFiles != nil {
+			c.Extensions.BigBang.FluxPatchFiles = append(c.Extensions.BigBang.FluxPatchFiles, override.Extensions.BigBang.FluxPatchFiles...)
+		}
+	}
```
Racer159, line 27:
```suggestion
	// perform any overrides
	if override.Extensions.BigBang != nil {
		if c.Extensions.BigBang == nil {
			c.Extensions.BigBang = override.Extensions.BigBang
		} else {
			if override.Extensions.BigBang.ValuesFiles != nil {
				c.Extensions.BigBang.ValuesFiles = append(c.Extensions.BigBang.ValuesFiles, override.Extensions.BigBang.ValuesFiles...)
				message.Warnf("%s", c.Extensions.BigBang.ValuesFiles)
			}
			if override.Extensions.BigBang.FluxPatchFiles != nil {
				c.Extensions.BigBang.FluxPatchFiles = append(c.Extensions.BigBang.FluxPatchFiles, override.Extensions.BigBang.FluxPatchFiles...)
			}
		}
	}
```

src/pkg/packager/composer/list.go
```diff
@@ -0,0 +1,420 @@
...
+	return composed, nil
+}
+
+// MergeVariables merges variables from the import chain
+func (ic *ImportChain) MergeVariables(existing []types.ZarfPackageVariable) (merged []types.ZarfPackageVariable) {
+	exists := func(v1 types.ZarfPackageVariable, v2 types.ZarfPackageVariable) bool {
+		return v1.Name == v2.Name
+	}
+
+	merged = helpers.MergeSlices(existing, merged, exists)
```
Racer159, line 396: It may be better to initialize the import chain with the variables / constants instead of having to handle the cases here.

src/pkg/utils/helpers/slice.go
```diff
@@ -0,0 +1,30 @@
...
+
+// Package helpers provides generic helper functions with no external imports
+package helpers
+
+// EqualFunc defines a type for a function that determines equality between two elements of type T.
+type EqualFunc[T any] func(a, b T) bool
+
+// MergeSlices merges two slices, s1 and s2, and returns a new slice containing all elements from s1
+// and only those elements from s2 that do not exist in s1 based on the provided equal function.
+func MergeSlices[T any](s1, s2 []T, equals EqualFunc[T]) []T {
```
Racer159, line 12: If we are going to create a new slice.go file all of the slice functions from misc should move here too.

src/test/e2e/50_oci_package_test.go
```diff
@@ -196,7 +196,7 @@ func (suite *RegistryClientTestSuite) Test_5_Copy() {
 	}
 }
 
-func TestRegistryClientTestSuite(t *testing.T) {
+func TestRegistryClientSuite(t *testing.T) {
```
Racer159, line 199: I do think we should map out the OCI tests here and figure out which branches are being covered and which are not - extensions are definitely not covered and would be easy to break in current state.

Racer159 (dismissed):
In addition to these this should have a short ADR on why the linked list and alternatives that were discussed - another thing to mention briefly is that the component find logic will help with expandability to other things like `flavor` (if needed)

src/pkg/packager/composer/list.go
Racer159: I added a comment explaining a bit more on why this was - we could leave that comment in and reevaluate when full caching is implemented if we want (short story is it has to do with multiple pulls and how the filter doesn't pull everything from the skeleton)

src/pkg/packager/composer/list.go
Noxsios: The reason for this is because it maintains an import chains "knowledge" of variables and consts. Since `existing` is going to mutate from chain to chain, saving it into the chain means the chain has knowledge of things outside itself.

lucasrod16 (approved):
lgtm

jeff-mccoy (approved):
way more sane way to manage composition, lgtm

PR #2065, opened 2023-10-11
feat: introduce Agent TLS rolling with `zarf tools update-creds agent`

### LINKED ISSUE #2057: Add the agent TLS to `zarf tools update-creds`
It would be nice to have the Agent's TLS keys in `zarf tools update-creds` as a new `agent` key so that it is easier to update the credentials when they need to be rotated or if they expire.

- **Given** I have an initialized cluster
- **When** I run `zarf tools update-creds agent`
- **Then** Zarf generates new certificates, updates Zarf state, and updates the TLS secret in cluster

### Additional context
This would help with Day 2 operations and expand this command into a gap that it didn't cover before.

Racer159:
## Description

This adds an `agent` option to the `zarf tools update-creds` command to rotate the agent TLS certificates.

## Related Issue

Fixes #2057 

## Type of change

## Checklist before merging


lucasrod16 (commented):
Would probably be good to add this new command to the examples/help output:

https://github.com/defenseunicorns/zarf/blob/77db6d001d644624e1348ca35d79f29fde223a01/src/config/lang/english.go#L504

src/internal/packager/helm/zarf.go
```diff
@@ -72,3 +77,85 @@ func (h *Helm) UpdateZarfGiteaValues() error {
 
 	return nil
 }
+
+// UpdateZarfAgentValues updates the Zarf git server deployment with the new state values
```
lucasrod16, line 81:
```suggestion
// UpdateZarfAgentValues updates the Zarf agent deployment with the new state values
```

src/internal/packager/helm/zarf.go
```diff
@@ -72,3 +77,85 @@ func (h *Helm) UpdateZarfGiteaValues() error {
...
+
+	spinner = message.NewProgressSpinner("Cleaning up Zarf Agent pods after update")
+	defer spinner.Stop()
+
+	// Force pods to be recreated to get the updated secret.
+	pods = h.Cluster.WaitForPodsAndContainers(k8s.PodLookup{
+		Namespace: cluster.ZarfNamespaceName,
+		Selector:  "app=agent-hook",
+	}, nil)
+
```
lucasrod16, line 153: Is there a reason we are making another call to `WaitForPodsAndContainers` here? At this stage we already have the pods in-memory and should be able to iterate over them and delete them without re-fetching.

```suggestion
```

src/pkg/message/credentials.go
```diff
@@ -136,6 +137,14 @@ func PrintCredentialUpdates(oldState *types.ZarfState, newState *types.ZarfState
 			pterm.Printfln("    %s: %s", pterm.Bold.Sprint("URL Address"), compareStrings(oA.Address, nA.Address, false))
 			pterm.Printfln("    %s: %s", pterm.Bold.Sprint("Push Username"), compareStrings(oA.PushUsername, nA.PushUsername, false))
 			pterm.Printfln("    %s: %s", pterm.Bold.Sprint("Push Token"), compareStrings(oA.PushToken, nA.PushToken, true))
+		case AgentKey:
+			oK := oldState.AgentTLS
```
Noxsios, line 141: nit: I don't like calling variables variations of `ok` that are not bool as it is harder to understand behavior at first glance

Noxsios (changes requested):

src/pkg/message/credentials.go
Racer159: https://www.youtube.com/watch?v=yGUNrHWZaVs

src/internal/packager/helm/zarf.go
Racer159: I did this because the update command takes some time and there is a _chance_ that one of the pods was recycled behind the scenes and we no longer hold a reference to it - we could look at doing this better though - will think more on it, maybe adding a delete by selector or something.

Racer159: Just updated to remove this in favor of a DeletePods function.

src/pkg/message/credentials.go
Racer159: changed to oT / nT instead.

lucasrod16 (approved):
lgtm

Noxsios (approved):

PR #2071, opened 2023-10-13
chore: remove the max package size creation prompt and improve docs

### LINKED ISSUE #2060: Remove user prompt for max package size on `zarf package create`
While this is still important for some use-cases, sufficient documentation (there probably should be more than https://docs.zarf.dev/docs/deploy-a-zarf-package/package-sources) should cover this need without forcing every package creator to see this prompt or override the setting with a config.

Racer159:
## Description

This removes the max package size prompt and improves the docs to better allow people who need that feature to find it.

## Related Issue

Fixes #2060 

## Type of change

## Checklist before merging


Noxsios (approved):
finally

PR #2075, opened 2023-10-16
Zarf prepare lint schema

AustinAbro321:
## Description

Intent of this PR is to introduce the command zarf prepare lint, with the ability to validate the zarf schema

## Related Issue

Relates to # 2064 1667

## Type of change

## Checklist before merging

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,6 +42,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.16.0
 	github.com/stretchr/testify v1.8.4
+	github.com/xeipuuv/gojsonschema v1.2.0
 	golang.org/x/crypto v0.13.0
 	golang.org/x/sync v0.3.0
 	golang.org/x/term v0.12.0
@@ -57,6 +58,8 @@ require (
 	sigs.k8s.io/kustomize/kyaml v0.14.3
 	sigs.k8s.io/yaml v1.3.0
 )
+
+require github.com/santhosh-tekuri/jsonschema v1.2.4
 
 require (
 	atomicgo.dev/cursor v0.2.0 // indirect
@@ -367,6 +370,7 @@ require (
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/ryanuber/go-glob v1.0.0 // indirect
 	github.com/sahilm/fuzzy v0.1.0 // indirect
+	github.com/santhosh-tekuri/jsonschema/v5 v5.3.1
 	github.com/sassoftware/go-rpmutils v0.2.0 // indirect
 	github.com/sassoftware/relic v7.2.1+incompatible // indirect
 	github.com/scylladb/go-set v1.0.3-0.20200225121959-cc7b2070d91e // indirect
@@ -411,7 +415,6 @@ require (
 	github.com/xanzy/ssh-agent v0.3.3 // indirect
 	github.com/xeipuuv/gojsonpointer v0.0.0-20190905194746-02993c407bfb // indirect
 	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
-	github.com/xeipuuv/gojsonschema v1.2.0 // indirect
 	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 // indirect
 	github.com/xlab/treeprint v1.2.0 // indirect
 	github.com/xo/terminfo v0.0.0-20220910002029-abceb7e1c41e // indirect
@@ -452,7 +455,7 @@ require (
 	gopkg.in/square/go-jose.v2 v2.6.0 // indirect
 	gopkg.in/warnings.v0 v0.1.2 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
-	gopkg.in/yaml.v3 v3.0.1 // indirect
+	gopkg.in/yaml.v3 v3.0.1
 	inet.af/netaddr v0.0.0-20220811202034-502d2d690317 // indirect
 	k8s.io/apiextensions-apiserver v0.28.2 // indirect
 	k8s.io/apiserver v0.28.2 // indirect
--- a/src/cmd/prepare.go
+++ b/src/cmd/prepare.go
@@ -7,6 +7,7 @@ package cmd
 import (
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"strings"
 
@@ -20,6 +21,8 @@ import (
 	"github.com/defenseunicorns/zarf/src/pkg/utils"
 	"github.com/defenseunicorns/zarf/src/pkg/utils/helpers"
 	"github.com/spf13/cobra"
+	"github.com/xeipuuv/gojsonschema"
+	"sigs.k8s.io/yaml"
 )
 
 var prepareCmd = &cobra.Command{
@@ -157,6 +160,16 @@ var prepareGenerateConfigFile = &cobra.Command{
 	},
 }
 
+var prepareValidateSchema = &cobra.Command{
+	Use:     "lint [ FILENAME ] [ FILENAME ]",
+	Args:    cobra.MaximumNArgs(2),
+	Short:   lang.CmdPrepareGenerateConfigShort,
+	Long:    lang.CmdPrepareGenerateConfigLong,
+	Run: func(cmd *cobra.Command, args []string) {
+
+	},
+}
+
 func init() {
 	v := common.InitViper()
 
@@ -165,6 +178,7 @@ func init() {
 	prepareCmd.AddCommand(prepareComputeFileSha256sum)
 	prepareCmd.AddCommand(prepareFindImages)
 	prepareCmd.AddCommand(prepareGenerateConfigFile)
+	prepareCmd.AddCommand(prepareValidateSchema)
 
 	prepareFindImages.Flags().StringVarP(&pkgConfig.FindImagesOpts.RepoHelmChartPath, "repo-chart-path", "p", "", lang.CmdPrepareFlagRepoChartPath)
 	// use the package create config for this and reset it here to avoid overwriting the config.CreateOptions.SetVariables
--- a/src/cmd/prepare.go
+++ b/src/cmd/prepare.go
@@ -7,22 +7,21 @@ package cmd
 import (
 	"fmt"
 	"io"
-	"log"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/defenseunicorns/zarf/src/cmd/common"
 	"github.com/defenseunicorns/zarf/src/config"
 	"github.com/defenseunicorns/zarf/src/config/lang"
+	"github.com/defenseunicorns/zarf/src/pkg/layout"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 	"github.com/defenseunicorns/zarf/src/pkg/packager"
 	"github.com/defenseunicorns/zarf/src/pkg/transform"
 	"github.com/defenseunicorns/zarf/src/pkg/utils"
 	"github.com/defenseunicorns/zarf/src/pkg/utils/helpers"
 	"github.com/spf13/cobra"
-	"github.com/xeipuuv/gojsonschema"
-	"sigs.k8s.io/yaml"
 )
 
 var prepareCmd = &cobra.Command{
@@ -161,12 +160,16 @@ var prepareGenerateConfigFile = &cobra.Command{
 }
 
 var prepareValidateSchema = &cobra.Command{
-	Use:     "lint [ FILENAME ] [ FILENAME ]",
-	Args:    cobra.MaximumNArgs(2),
-	Short:   lang.CmdPrepareGenerateConfigShort,
-	Long:    lang.CmdPrepareGenerateConfigLong,
+	Use:   "lint [ FILENAME ] [ FILENAME ]",
+	Args:  cobra.MaximumNArgs(2),
+	Short: lang.CmdPrepareGenerateConfigShort,
+	Long:  lang.CmdPrepareGenerateConfigLong,
 	Run: func(cmd *cobra.Command, args []string) {
 
+		pkgClient := packager.NewOrDie(&pkgConfig)
+		defer pkgClient.ClearTempPaths()
+		// Need to find where we load in the file
+		helpers.ValidateZarfSchema()
 	},
 }
 
--- a/go.sum
+++ b/go.sum
@@ -1412,6 +1412,10 @@ github.com/ryanuber/go-glob v1.0.0/go.mod h1:807d1WSdnB0XRJzKNil9Om6lcp/3a0v4qIH
 github.com/sagikazarmark/crypt v0.3.0/go.mod h1:uD/D+6UF4SrIR1uGEv7bBNkNqLGqUr43MRiaGWX1Nig=
 github.com/sahilm/fuzzy v0.1.0 h1:FzWGaw2Opqyu+794ZQ9SYifWv2EIXpwP4q8dY1kDAwI=
 github.com/sahilm/fuzzy v0.1.0/go.mod h1:VFvziUEIMCrT6A6tw2RFIXPXXmzXbOsSHF0DOI8ZK9Y=
+github.com/santhosh-tekuri/jsonschema v1.2.4 h1:hNhW8e7t+H1vgY+1QeEQpveR6D4+OwKPXCfD2aieJis=
+github.com/santhosh-tekuri/jsonschema v1.2.4/go.mod h1:TEAUOeZSmIxTTuHatJzrvARHiuO9LYd+cIxzgEHCQI4=
+github.com/santhosh-tekuri/jsonschema/v5 v5.3.1 h1:lZUw3E0/J3roVtGQ+SCrUrg3ON6NgVqpn3+iol9aGu4=
+github.com/santhosh-tekuri/jsonschema/v5 v5.3.1/go.mod h1:uToXkOrWAZ6/Oc07xWQrPOhJotwFIyu2bBVN41fcDUY=
 github.com/sassoftware/go-rpmutils v0.2.0 h1:pKW0HDYMFWQ5b4JQPiI3WI12hGsVoW0V8+GMoZiI/JE=
 github.com/sassoftware/go-rpmutils v0.2.0/go.mod h1:TJJQYtLe/BeEmEjelI3b7xNZjzAukEkeWKmoakvaOoI=
 github.com/sassoftware/relic v7.2.1+incompatible h1:Pwyh1F3I0r4clFJXkSI8bOyJINGqpgjJU3DYAZeI05A=
--- a/src/pkg/utils/helpers/example-schema.json
+++ b/src/pkg/utils/helpers/example-schema.json
@@ -1,0 +1,31 @@
+{
+  "$id": "https://example.com/product.schema.json",
+  "title": "Product",
+  "description": "A product from Acme's catalog",
+  "type": "object",
+  "properties": {
+    "productId": {
+      "description": "The unique identifier for a product",
+      "type": "integer"
+    },
+    "productName": {
+      "description": "Name of the product",
+      "type": "string"
+    },
+    "price": {
+      "description": "The price of the product",
+      "type": "number",
+      "exclusiveMinimum": 0
+    },
+    "tags": {
+      "description": "Tags for the product",
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "minItems": 1,
+      "uniqueItems": true
+    }
+  },
+  "required": [ "productId", "productName", "price" ]
+}
--- a/src/pkg/utils/helpers/example.yaml
+++ b/src/pkg/utils/helpers/example.yaml
@@ -1,0 +1,6 @@
+productId: 1
+productName: A green door
+price: 12.50
+tags:
+- home
+- green
--- a/src/pkg/utils/helpers/validate_schema.go
+++ b/src/pkg/utils/helpers/validate_schema.go
@@ -3,7 +3,6 @@ package helpers
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"log"
 	"os"
@@ -29,10 +28,6 @@ func ReadSchema(yamlFile, jsonFile string) bool {
 	if err != nil {
 		panic(err)
 	}
-	m, err = toStringKeys(m)
-	if err != nil {
-		panic(err)
-	}
 
 	mJSON, err := json.Marshal(m)
 	if err != nil {
@@ -55,34 +50,4 @@ func ReadSchema(yamlFile, jsonFile string) bool {
 	}
 	fmt.Println("validation successfull")
 	return true
-}
-
-func toStringKeys(val interface{}) (interface{}, error) {
-	var err error
-	switch val := val.(type) {
-	case map[interface{}]interface{}:
-		m := make(map[string]interface{})
-		for k, v := range val {
-			k, ok := k.(string)
-			if !ok {
-				return nil, errors.New("found non-string key")
-			}
-			m[k], err = toStringKeys(v)
-			if err != nil {
-				return nil, err
-			}
-		}
-		return m, nil
-	case []interface{}:
-		var l = make([]interface{}, len(val))
-		for i, v := range val {
-			l[i], err = toStringKeys(v)
-			if err != nil {
-				return nil, err
-			}
-		}
-		return l, nil
-	default:
-		return val, nil
-	}
 }
--- a/src/pkg/utils/helpers/validate_schema.go
+++ b/src/pkg/utils/helpers/validate_schema.go
@@ -3,7 +3,6 @@ package helpers
 import (
 	"bytes"
 	"encoding/json"
-	"fmt"
 	"log"
 	"os"
 
@@ -11,43 +10,41 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-func ReadSchema(yamlFile, jsonFile string) bool {
+func ValidateZarfSchema(yamlFile, jsonFile string) bool {
 
 	yamlBytes, err := os.ReadFile(yamlFile)
 	if err != nil {
-		log.Fatalf("Error reading YAML file: %s", yamlFile)
+		log.Fatalf("Error reading file: %s", err)
 	}
 
-	jsonBytes, err := os.ReadFile(jsonFile)
+	zarfSchema, err := os.ReadFile(jsonFile)
 	if err != nil {
-		log.Fatalf("Error reading YAML file: %s", jsonFile)
+		log.Fatalf("Error reading file: %s", err)
 	}
 
-	var m interface{}
-	err = yaml.Unmarshal(yamlBytes, &m)
-	if err != nil {
-		panic(err)
-	}
-
-	mJSON, err := json.Marshal(m)
+	var unmarshalledYaml interface{}
+	err = yaml.Unmarshal(yamlBytes, &unmarshalledYaml)
 	if err != nil {
 		panic(err)
 	}
-	yamlJsonReader := bytes.NewReader(mJSON)
-	schemaReader := bytes.NewReader(jsonBytes)
+
+	zarfYamlAsJsonBytes, err := json.Marshal(unmarshalledYaml)
+	if err != nil {
+		panic(err)
+	}
 
 	compiler := jsonschema.NewCompiler()
+	inMemoryZarfSchema := "zarf.schema.json"
 
-	if err := compiler.AddResource("schema.json", schemaReader); err != nil {
+	if err := compiler.AddResource(inMemoryZarfSchema, bytes.NewReader(zarfSchema)); err != nil {
 		panic(err)
 	}
-	schema, err := compiler.Compile("schema.json")
+	schema, err := compiler.Compile(inMemoryZarfSchema)
 	if err != nil {
 		panic(err)
 	}
-	if err := schema.Validate(yamlJsonReader); err != nil {
+	if err := schema.Validate(bytes.NewReader(zarfYamlAsJsonBytes)); err != nil {
 		panic(err)
 	}
-	fmt.Println("validation successfull")
 	return true
 }
--- a/src/pkg/utils/helpers/validate_schema.go
+++ b/src/pkg/utils/helpers/validate_schema.go
@@ -7,28 +7,16 @@ import (
 	"os"
 
 	"github.com/santhosh-tekuri/jsonschema"
-	"gopkg.in/yaml.v3"
 )
 
-func ValidateZarfSchema(yamlFile, jsonFile string) bool {
-
-	yamlBytes, err := os.ReadFile(yamlFile)
-	if err != nil {
-		log.Fatalf("Error reading file: %s", err)
-	}
+func ValidateZarfSchema(zarfUnmarshaledYaml interface{}, jsonFile string) bool {
 
 	zarfSchema, err := os.ReadFile(jsonFile)
 	if err != nil {
 		log.Fatalf("Error reading file: %s", err)
 	}
 
-	var unmarshalledYaml interface{}
-	err = yaml.Unmarshal(yamlBytes, &unmarshalledYaml)
-	if err != nil {
-		panic(err)
-	}
-
-	zarfYamlAsJsonBytes, err := json.Marshal(unmarshalledYaml)
+	zarfYamlAsJsonBytes, err := json.Marshal(zarfUnmarshaledYaml)
 	if err != nil {
 		panic(err)
 	}
--- a/src/pkg/utils/helpers/validate_schema.go
+++ b/src/pkg/utils/helpers/validate_schema.go
@@ -1,0 +1,88 @@
+package helpers
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/santhosh-tekuri/jsonschema"
+	"gopkg.in/yaml.v3"
+)
+
+func ReadSchema(yamlFile, jsonFile string) bool {
+
+	yamlBytes, err := os.ReadFile(yamlFile)
+	if err != nil {
+		log.Fatalf("Error reading YAML file: %s", yamlFile)
+	}
+
+	jsonBytes, err := os.ReadFile(jsonFile)
+	if err != nil {
+		log.Fatalf("Error reading YAML file: %s", jsonFile)
+	}
+
+	var m interface{}
+	err = yaml.Unmarshal(yamlBytes, &m)
+	if err != nil {
+		panic(err)
+	}
+	m, err = toStringKeys(m)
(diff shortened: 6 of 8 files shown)


AustinAbro321:
This is ready to review, I have several questions 
- Should my files exist in src/pkg/lint or should I move those into src/pkg/packager ?
- I decided to make ZARF_PKG_TMPL_ in component.import.path to be a warning and check for separately, since we had to take the regex out of the lint. I can switch it to being an error if we'd like.

src/test/packages/11-lint/valid_yaml/zarf.yaml
Racer159, line 1: We could probably just reference an existing file that is valid since the root init package always should be and that would help test our own patterns

src/pkg/lint/validate_schema_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	readAndUnmarshalYaml := func(t *testing.T, path string) interface{} {
+		t.Helper()
+		var unmarshalledYaml interface{}
+		file, _ := os.ReadFile(path)
+		goyaml.Unmarshal(file, &unmarshalledYaml)
+		return unmarshalledYaml
+	}
+
+	readAndUnmarshallZarf := func(t *testing.T, path string) types.ZarfPackage {
```
Racer159, line 30:
```suggestion
	readAndUnmarshallZarfPackage := func(t *testing.T, path string) types.ZarfPackage {
```

Likely better to refer to this as a Package (or Pkg if you prefer) since Zarf can be many things.

src/pkg/lint/validate_schema.go
```diff
@@ -0,0 +1,58 @@
...
+	"strings"
+
+	"github.com/defenseunicorns/zarf/src/types"
+	"github.com/xeipuuv/gojsonschema"
+)
+
+const (
+	zarfInvalidPrefix = "zarf.yaml is not valid:"
+	zarfWarningPrefix = "zarf schema warning:"
+	zarfTemplateVar   = "###ZARF_PKG_TMPL_"
```
Racer159, line 15: It might be better to make this a variable that can be shared since this magic string is also used with `promptAndSetTemplate`

src/pkg/lint/validate_schema.go
```diff
@@ -0,0 +1,58 @@
...
+	errorMessage := zarfWarningPrefix
+	componentWarningStart := " component."
+	for i, component := range zarfYaml.Components {
+		if strings.Contains(component.Import.Path, zarfTemplateVar) {
+			errorMessage = errorMessage + componentWarningStart + strconv.Itoa(i) +
+				".import.path will not resolve ZARF_PKG_TMPL_* variables."
+			valid = false
+		}
+		if strings.Contains(component.Import.URL, zarfTemplateVar) {
+			errorMessage = errorMessage + componentWarningStart + strconv.Itoa(i) +
```
Racer159, line 49: for most messages we use `fmt.Sprintf` instead of string concatenation

src/pkg/lint/unsuccessful_validation/zarf.yaml
```diff
@@ -0,0 +1,21 @@
...
+    not-import:
+      path: packages/distros/k3s
+
+  - name: import-test
+    import:
+      path: "###ZARF_PKG_TMPL_ZEBRA###"
+
+  - name: import-url
+    import:
+      url: "oci://###ZARF_PKG_TMPL_ZEBRA###"
```
Racer159, line 21: We don't usually place test files here in `pkg`.  If we _need_ to we can reuse the `11-lint` test package(s) but should also look for a way to mock this out better if we can.

Racer159 (changes requested):
A few changes, but looking good so far!

src/pkg/lint/unsuccessful_validation/zarf.yaml
AustinAbro321: I know just adding the files in plain text as a variable in go didn't work because our .editerconfig file automatically turns 4 spaces into a tab. 

I will see if any of our existing yaml packages allows for key values additions and see how we like adding schema errors that way

src/cmd/prepare.go
```diff
@@ -205,6 +206,29 @@ var prepareGenerateConfigFile = &cobra.Command{
 	},
 }
 
+var lintCmd = &cobra.Command{
+	Use:     "lint [ DIRECTORY ]",
+	Args:    cobra.MaximumNArgs(1),
+	Aliases: []string{"l"},
+	Short:   lang.CmdLintShort,
+	Run: func(cmd *cobra.Command, args []string) {
+		baseDir := ""
```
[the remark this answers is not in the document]
AustinAbro321: I was initially hesitant to use pkg / creatopts because in order for the lint to work I need a untyped unmarshalled yaml variable. However, I ended up also needing a typed variable when I checked for ###ZARF_PKG_TMPL_ anyway. I will try out using createOpts again

src/pkg/lint/validate_schema.go
AustinAbro321: Exported it for now, maybe a better place to put magic strings like this going forward? english.go didn't feel right

docs/2-the-zarf-cli/100-cli-commands/zarf_prepare.md
```diff
@@ -27,6 +27,7 @@ Tools to help prepare assets for packaging
 * [zarf](zarf.md)	 - DevSecOps for Airgap
 * [zarf prepare find-images](zarf_prepare_find-images.md)	 - Evaluates components in a zarf file to identify images specified in their helm charts and manifests
 * [zarf prepare generate-config](zarf_prepare_generate-config.md)	 - Generates a config file for Zarf
+* [zarf prepare lint](zarf_prepare_lint.md)	 - Tools for validating schema / ensuring best practices
```
bdw617, line 30: verifies the package schema, makes sure there isn't stuff in the yaml zarf will ignore.

go.mod
```diff
@@ -41,6 +41,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.17.0
 	github.com/stretchr/testify v1.8.4
+	github.com/xeipuuv/gojsonschema v1.2.0
```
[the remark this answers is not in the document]
bdw617: I agree, this hasn't seen a release in 3 years

src/config/lang/english.go
```diff
@@ -305,6 +305,9 @@ const (
 	CmdPackageClusterSourceFallback = "%q does not satisfy any current sources, assuming it is a package deployed to a cluster"
 	CmdPackageInvalidSource         = "Unable to identify source from %q: %s"
 
+	//zarf lint
+	CmdLintShort = "Tools for validating schema / ensuring best practices"
```
bdw617, line 309: ensuring best practices, should be like schema is correct, and identifies deprecations and warns the user.

src/pkg/lint/lint.go
bdw617, line 1: pick a style and go with it.

src/pkg/lint/validate_schema.go
```diff
@@ -0,0 +1,58 @@
...
+	"strings"
+
+	"github.com/defenseunicorns/zarf/src/types"
+	"github.com/xeipuuv/gojsonschema"
+)
+
+const (
+	zarfInvalidPrefix = "zarf.yaml is not valid:"
+	zarfWarningPrefix = "zarf schema warning:"
+	ZarfTemplateVar   = "###ZARF_PKG_TMPL_"
```
bdw617, line 15: put in separate block and document

src/pkg/lint/validate_schema.go
```diff
@@ -0,0 +1,58 @@
...
+		for _, desc := range result.Errors() {
+			errorMessage = fmt.Sprintf("%s\n - %s", errorMessage, desc.String())
+		}
+		err = errors.New(errorMessage)
+	}
+
+	return err
+}
+
+func checkForVarInComponentImport(zarfYaml types.ZarfPackage) error {
```
bdw617, line 38: document why we have this here versus the lookahead in the schema.

src/pkg/lint/validate_schema.go
```diff
@@ -0,0 +1,58 @@
...
+	valid := true
+	errorMessage := zarfWarningPrefix
+	componentWarningStart := "component."
+	for i, component := range zarfYaml.Components {
+		if strings.Contains(component.Import.Path, ZarfTemplateVar) {
```
bdw617, line 43: each of these are rules that could not be enforced in the schema due to many reasons.

bdw617 (changes requested):
Looks great! some minor changes on style, better documentation in code on the "why". We now have a pattern for post-schema validation of zarf.yaml as well, and I think this is really important and will be very helpful down the road.

src/cmd/prepare.go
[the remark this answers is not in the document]
AustinAbro321: Ended up using createopts and moved lint code to packager

src/pkg/lint/validate_schema.go
AustinAbro321: Didn't end up exporting as I moved to packager package anyway.

src/pkg/lint/unsuccessful_validation/zarf.yaml
AustinAbro321: Didn't try adding key values, could still do it but I feel it will be kinda ugly. Ended up using the zarf.yaml in the lint package for now, but agree that isn't ideal as someone might change that file and not notice the other test break. My vote would be to put it in plaintext, but then we have to change the .editerconfig to not turn spaces into tabs

bdw617 (approved):
lgtm, one build error with docs.

src/test/e2e/12_lint_test.go
lucasrod16, line 1: Why have an E2E test for schema validation? Can these cases be covered by unit tests instead?

AustinAbro321: Great question, I do have unit tests for src/pkg/packager/lint.go with lint_test.go where most of the schema validation code lives. 

The reason in my mind to have an end to end tests is so we can check that go embed is working properly and the changes to the cli command is working.

go.mod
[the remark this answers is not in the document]
AustinAbro321: We ending up forking the repo and replacing it in the go.mod

src/pkg/lint/unsuccessful_validation/zarf.yaml
AustinAbro321: I ended up adding the file as plaintext in the test file.

src/cmd/prepare.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/defenseunicorns/zarf/src/config/lang"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 	"github.com/defenseunicorns/zarf/src/pkg/packager"
+	validator "github.com/defenseunicorns/zarf/src/pkg/packager/validator"
```
Racer159, line 20:
```suggestion
	"github.com/defenseunicorns/zarf/src/pkg/packager/validator"
```
You don't need to overwrite the name when it is the same name

src/types/runtime.go
```diff
@@ -9,6 +9,11 @@ const (
 	RawVariableType VariableType = "raw"
 	// FileVariableType is a type for a Zarf package variable that loads its contents from a file
 	FileVariableType VariableType = "file"
+
+	ZarfPackageTemplatePrefix = "###ZARF_PKG_TMPL_"
```
Racer159, line 13: lint

src/config/lang/english.go
```diff
@@ -337,6 +337,8 @@ const (
 	CmdPrepareFlagGitAccount    = "User or organization name for the git account that the repos are created under."
 	CmdPrepareFlagKubeVersion   = "Override the default helm template KubeVersion when performing a package chart template"
 
+	CmdPrepareLintShort = "Verifies the package schema"
```
Racer159, line 340: I think we can relatively easily make this do a little more if we refactor packager to have some of its warnings generated from `validator`.  This doesn't _have_ to be in this PR, but if it isn't, a new issue should be created to add it that gets worked on next.

src/pkg/packager/validator/validator.go
```diff
@@ -0,0 +1,92 @@
...
+	var zarfData interface{}
+	if err := utils.ReadYaml(filepath.Join(path, layout.ZarfYAML), &zarfData); err != nil {
+		return err
+	}
+
+	if err = validateSchema(zarfData, zarfSchema); err != nil {
+		return err
+	}
+
+	message.Success("Validation successful")
```
Racer159, line 48: A more complete error message would be nice here - if you run `zarf prepare lint` with a zarf.yaml in the current directory you might not be sure which zarf.yaml it picked up (if you are less Zarf savvy)

Something like:
```suggestion
	message.Success("Schema validation successful for %q", zarfTypedData.Metadata.Name)
```

src/pkg/packager/validator/validater_test.go
```diff
@@ -0,0 +1,125 @@
...
+		var unmarshalledYaml interface{}
+		file := readFileFailFatally(t, path)
+		err := goyaml.Unmarshal(file, &unmarshalledYaml)
+		if err != nil {
+			t.Errorf("error unmarshalling yaml %v", err)
+		}
+		return unmarshalledYaml
+	}
+
+	readAndUnmarshallYamlString := func(t *testing.T, yamlString string) interface{} {
```
Racer159, line 64:
```suggestion
	readAndUnmarshalYamlString := func(t *testing.T, yamlString string) interface{} {
```

src/pkg/packager/validator/validater_test.go
```diff
@@ -0,0 +1,125 @@
...
+		t.Helper()
+		var unmarshalledYaml interface{}
+		err := goyaml.Unmarshal([]byte(yamlString), &unmarshalledYaml)
+		if err != nil {
+			t.Errorf("error unmarshalling yaml string %v", err)
+		}
+		return unmarshalledYaml
```
Racer159, line 71: This is a copy of the function on line 53 and should be desiccated.

src/pkg/packager/validator/validater_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	readAndUnmarshallZarfPackage := func(t *testing.T, path string) types.ZarfPackage {
+		t.Helper()
+		var unmarshalledYaml types.ZarfPackage
+		file := readFileFailFatally(t, path)
+		err := goyaml.Unmarshal(file, &unmarshalledYaml)
+		if err != nil {
+			t.Errorf("error unmarshalling yaml %s", err)
+		}
+		return unmarshalledYaml
```
Racer159, line 82: This could take in a type for the interface to assist with desiccation

src/pkg/packager/validator/validater_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	readAndUnmarshallZarfPackageString := func(t *testing.T, yamlString string) types.ZarfPackage {
+		t.Helper()
+		var unmarshalledYaml types.ZarfPackage
+		err := goyaml.Unmarshal([]byte(yamlString), &unmarshalledYaml)
+		if err != nil {
+			t.Errorf("error unmarshalling yaml %v", err)
+		}
+		return unmarshalledYaml
```
Racer159, line 92: Moar desiccation

src/pkg/packager/validator/validater_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	t.Run("validate schema success", func(t *testing.T) {
+		unmarshalledYaml := readAndUnmarshalYaml(t, "../../../../zarf.yaml")
+		zarfSchema := readSchema(t)
+		err := validateSchema(unmarshalledYaml, zarfSchema)
+		require.NoError(t, err)
+	})
```
Racer159, line 100: I don't think we should have reads from the filesystem in these unit tests.  We could just have a simple "good" Zarf package as a string and use that right?  That would also help dry the above functions.

Racer159 (changes requested):

src/pkg/packager/validator/validator.go
```diff
@@ -0,0 +1,92 @@
...
+	return nil
+}
+
+func checkForVarInComponentImport(zarfYaml types.ZarfPackage) error {
```
bdw617, line 52: this logic is a bit complicated for concatting multiple errors. chatgpt came up with this:

func checkForVarInComponentImport(zarfYaml types.ZarfPackage) error {
	var errorMessages []string

	for i, component := range zarfYaml.Components {
		if strings.Contains(component.Import.Path, types.ZarfPackageTemplatePrefix) {
			errorMessages = append(errorMessages, fmt.Sprintf("component.%d.import.path will not resolve ZARF_PKG_TMPL_* variables", i))
		}
		if strings.Contains(component.Import.URL, types.ZarfPackageTemplatePrefix) {
			errorMessages = append(errorMessages, fmt.Sprintf("component.%d.import.url will not resolve ZARF_PKG_TMPL_* variables", i))
		}
	}

	if len(errorMessages) > 0 {
		return errors.New(zarfWarningPrefix + strings.Join(errorMessages, ", "))
	}
	return nil
}

src/pkg/packager/validator/validator.go
bdw617: returning err and messaging.Warn(err.Error()) could be more consistent :)

bdw617 (changes requested):
very close!

src/pkg/packager/validator/validator.go
AustinAbro321: Yeah agree that it's a bit weird that this function both returns error and uses the message package. Maybe I should have it not return an err and just do all the messaging within the function. Or should it return a custom type, with both warnings and errors. 

Curious what you guys think

src/pkg/packager/validator/validator.go
AustinAbro321: I like this, thanks Barry!

src/config/lang/english.go
AustinAbro321: I'd rather do it in the next PR just created - https://github.com/defenseunicorns/zarf/issues/2149

src/pkg/packager/validator/validator.go
AustinAbro321: Yeah we're going to create a struct that has an array of warnings and an array of errors and return that instead.

AustinAbro321: Ending up returning a validator object with warnings and errors and a regular error. The regular error is not null if something goes wrong in the validation process

bdw617 (dismissed):
AWESOME!!

src/pkg/packager/validator/validator.go
```diff
@@ -0,0 +1,136 @@
...
+func (v Validator) getFormatedError() error {
+	if !v.hasErrors() {
+		return nil
+	}
+	errorMessage := validatorInvalidPrefix
+	for _, errorStr := range v.errors {
+		errorMessage = fmt.Sprintf("%s\n - %s", errorMessage, errorStr.Error())
+	}
+	return errors.New(errorMessage)
+}
```
Noxsios, line 44: Should we implement the `Error` interface here? https://gobyexample.com/errors

src/pkg/packager/validator/validator.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+}
+
+// ValidateZarfSchema validates a zarf file against the zarf schema, returns a validator with warnings or errors if they exist
+func ValidateZarfSchema(path string) (Validator, error) {
+	validator := Validator{}
+	var err error
+	if err := utils.ReadYaml(filepath.Join(path, layout.ZarfYAML), &validator.typedZarfPackage); err != nil {
+		return validator, err
+	}
```
Noxsios, line 89: @Racer159 is it better to return a `*Validator` instead of a `Validator` so that you return `nil` instead of an empty struct?

src/pkg/packager/validator/validator.go
```diff
@@ -0,0 +1,136 @@
...
+
+func checkForVarInComponentImport(validator *Validator) {
+	for i, component := range validator.typedZarfPackage.Components {
+		if strings.Contains(component.Import.Path, types.ZarfPackageTemplatePrefix) {
+			validator.warnings = append(validator.warnings, fmt.Sprintf("component.%d.import.path will not resolve ZARF_PKG_TMPL_* variables", i))
+		}
+		if strings.Contains(component.Import.URL, types.ZarfPackageTemplatePrefix) {
+			validator.warnings = append(validator.warnings, fmt.Sprintf("component.%d.import.url will not resolve ZARF_PKG_TMPL_* variables", i))
+		}
+	}
```
Noxsios, line 116: I would make a `validator.AddWarning | validator.AddError` (tbd on naming) so this code is a bit more readable

src/pkg/packager/validator/validator.go
```diff
@@ -0,0 +1,136 @@
...
+	return errors.New(errorMessage)
+}
+
+func (v Validator) getFormatedWarning() string {
+	if !v.hasWarnings() {
+		return ""
+	}
+	return fmt.Sprintf("%s %s", validatorWarningPrefix, strings.Join(v.warnings, ", "))
+}
```
Noxsios, line 51: @Racer159 What's the UX for this, I think a table would be more readable here than a really long string

src/pkg/packager/validator/validator.go
```diff
@@ -0,0 +1,136 @@
...
+	return validator, nil
+}
+
+func checkForVarInComponentImport(validator *Validator) {
+	for i, component := range validator.typedZarfPackage.Components {
+		if strings.Contains(component.Import.Path, types.ZarfPackageTemplatePrefix) {
+			validator.warnings = append(validator.warnings, fmt.Sprintf("component.%d.import.path will not resolve ZARF_PKG_TMPL_* variables", i))
+		}
+		if strings.Contains(component.Import.URL, types.ZarfPackageTemplatePrefix) {
+			validator.warnings = append(validator.warnings, fmt.Sprintf("component.%d.import.url will not resolve ZARF_PKG_TMPL_* variables", i))
```
Noxsios, line 114:
```suggestion
			validator.warnings = append(validator.warnings, fmt.Sprintf("components.[%d].import.path will not resolve ZARF_PKG_TMPL_* variables", i))
		}
		if strings.Contains(component.Import.URL, types.ZarfPackageTemplatePrefix) {
			validator.warnings = append(validator.warnings, fmt.Sprintf("components.[%d].import.url will not resolve ZARF_PKG_TMPL_* variables", i))
```

I believe this syntax will now more properly resolve in `yq`, giving a better debugging/fixing experience

Noxsios (changes requested):

src/pkg/packager/validator/validator.go
Racer159: yes

Racer159: (it is also helpful if the validator struct ever gets fields that way it is always pointing to the same instantiation

src/pkg/packager/validator/validator.go
AustinAbro321: I'm leaning towards that we shouldn't implement error since I feel it would imply that the validator object is an error while it isn't really. I could see it both ways though, curious what you think @Racer159

src/pkg/packager/validator/validator.go
AustinAbro321: Yeah I was going to add newlines originally at least, but it looks like the current behavior of message.warn gets rid of the newlines and the string a paragraph. Maybe we use a different type of message or change that behavior?

src/pkg/packager/validator/validator.go
AustinAbro321: I had switched validator to implement the error interface, but I ended up deleting this function all together to instead print a table.

src/pkg/packager/validator/validator.go
AustinAbro321: I ended up printing a table here, looks like this
image

src/pkg/packager/lint/lint_test.go
```diff
@@ -96,23 +94,8 @@ func TestValidateSchema(t *testing.T) {
 		unmarshalledYaml := readAndUnmarshalYaml[types.ZarfPackage](t, badZarfPackage)
 		validator := Validator{typedZarfPackage: unmarshalledYaml}
 		checkForVarInComponentImport(&validator)
-		require.Equal(t, validator.warnings[0], "component.2.import.path will not resolve ZARF_PKG_TMPL_* variables")
-		require.Equal(t, validator.warnings[1], "component.3.import.url will not resolve ZARF_PKG_TMPL_* variables")
+		require.Equal(t, validator.warnings[0], "component.[2].import.path will not resolve ZARF_PKG_TMPL_* variables")
```
Noxsios, line 97: To be valid `yq` expression, needs a leading `.` and to be called `components`

Screenshot 2023-11-28 at 4 37 15 PM

src/pkg/packager/lint/validator.go
```diff
@@ -0,0 +1,74 @@
...
+func (v *Validator) addWarning(message string) {
+	v.warnings = append(v.warnings, message)
+}
+
+func (v *Validator) addError(err error) {
+	v.errors = append(v.errors, err)
+}
+
+// IsSuccess returns true if there are not any warnings or errors
+func (v Validator) IsSuccess() bool {
```
Noxsios, line 62: I believe currently success is determined by `hasErrors` _only_, and warnings are fine (per our other usages of error vs warning). When we create more nuanced levels via CLI flags, this function would prob be something like: `IsSuccess(level LintLevel) bool`

@Racer159

src/pkg/packager/lint/validator.go
```diff
@@ -0,0 +1,74 @@
...
+	return !v.hasWarnings() && !v.hasErrors()
+}
+
+// DisplayFormattedMessage Displays the message to the user with proper warnings, failures, or success
+// Will exit if there are errors
+func (v Validator) DisplayFormattedMessage() {
+	if v.IsSuccess() {
+		message.Success(v.getFormatedSuccess())
```
Noxsios, line 70: if `v.getFormattedSuccess` is only used here, I would make delete the function and put the string directly here

src/test/e2e/12_lint_test.go
```diff
@@ -15,9 +15,9 @@ func TestLint(t *testing.T) {
 
 		path := filepath.Join("src", "test", "packages", "12-lint")
 		_, stderr, err := e2e.Zarf("prepare", "lint", path)
-		require.Error(t, err, "Expect error here because the yaml file is not following schema")
-		require.Contains(t, stderr, "- components.0.import: Additional property not-path is not allowed")
-		require.Contains(t, stderr, "component.2.import.path will not resolve ZARF_PKG_TMPL_* variables")
+		require.Error(t, err, "Require an exit code since there was warnings / errors")
+		require.Contains(t, stderr, "components.0.import: Additional property not-path is not allowed")
```
Noxsios, line 19: this does not follow the other error string formats

src/pkg/packager/lint/lint.go
```diff
@@ -0,0 +1,86 @@
...
+	documentLoader := gojsonschema.NewGoLoader(validator.untypedZarfPackage)
+
+	result, err := gojsonschema.Validate(schemaLoader, documentLoader)
+	if err != nil {
+		return err
+	}
+
+	if !result.Valid() {
+		for _, desc := range result.Errors() {
+			validator.addError(errors.New(desc.String()))
```
Noxsios, line 81: I would make an `err` var here so that it is more clear why you need to wrap the schema error in a Go error:

```go
	if !result.Valid() {
		for _, desc := range result.Errors() {
			err := errors.New(desc.String())
			validator.addError(err)
		}
	}
```

Noxsios (changes requested):

src/cmd/prepare.go
```diff
@@ -205,6 +206,33 @@ var prepareGenerateConfigFile = &cobra.Command{
...
+	Short:   lang.CmdPrepareLintShort,
+	Run: func(cmd *cobra.Command, args []string) {
+		baseDir := ""
+		if len(args) > 0 {
+			baseDir = args[0]
+		} else {
+			var err error
+			baseDir, err = os.Getwd()
+			if err != nil {
+				message.Fatalf(err, lang.CmdPrepareFindImagesErr, err.Error())
```
lucasrod16, line 222: We should replace `CmdPrepareFindImagesErr` with an error message that fits this context of linting a Zarf package

src/pkg/packager/lint/lint.go
```diff
@@ -0,0 +1,86 @@
...
+	return ZarfSchema.ReadFile("zarf.schema.json")
+}
+
+// ValidateZarfSchema validates a zarf file against the zarf schema, returns *validator with warnings or errors if they exist
+// along with an error if the validation itself failed
+func ValidateZarfSchema(path string) (*Validator, error) {
+	validator := Validator{}
+	var err error
+	if err := utils.ReadYaml(filepath.Join(path, layout.ZarfYAML), &validator.typedZarfPackage); err != nil {
+		return &validator, err
```
lucasrod16, line 38: We should probably return `nil` instead of returning a likely invalid pointer value of `&validator` in the case of an error

```suggestion
		return nil, err
```

src/pkg/packager/lint/lint.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	checkForVarInComponentImport(&validator)
+
+	if validator.jsonSchema, err = getSchemaFile(); err != nil {
+		return &validator, err
```
lucasrod16, line 44:
```suggestion
		return nil, err
```

src/pkg/packager/lint/lint.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	if err := utils.ReadYaml(filepath.Join(path, layout.ZarfYAML), &validator.untypedZarfPackage); err != nil {
+		return &validator, err
```
lucasrod16, line 48:
```suggestion
		return nil, err
```

src/pkg/packager/lint/lint.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	if err = validateSchema(&validator); err != nil {
+		return &validator, err
```
lucasrod16, line 52:
```suggestion
		return nil, err
```

src/pkg/packager/lint/lint_test.go
```diff
@@ -0,0 +1,101 @@
...
+  - name: baseline
+    required: true
+`
+
+func readAndUnmarshalYaml[T interface{}](t *testing.T, yamlString string) T {
+	t.Helper()
+	var unmarshalledYaml T
+	err := goyaml.Unmarshal([]byte(yamlString), &unmarshalledYaml)
+	if err != nil {
+		t.Errorf("error unmarshalling yaml %v", err)
```
lucasrod16, line 54:
```suggestion
		t.Errorf("error unmarshalling yaml: %v", err)
```

src/pkg/packager/lint/lint_test.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	return unmarshalledYaml
+}
+
+func TestValidateSchema(t *testing.T) {
+	getZarfSchema := func(t *testing.T) []byte {
+		t.Helper()
+		file, err := os.ReadFile("../../../../zarf.schema.json")
+		if err != nil {
+			t.Errorf("error reading file: %s", err)
```
lucasrod16, line 64:
```suggestion
			t.Errorf("error reading file: %v", err)
```

lucasrod16 (changes requested):

src/pkg/packager/lint/validator.go
Racer159: Correct - any errors would result in a non-zero exit code, just warnings would result in a zero exit code, and nothing would result in a zero exit code (for now)

src/pkg/packager/lint/validator.go
Racer159: `make: *** No rule to make target 'delete'.  Stop.`

src/pkg/message/message.go
```diff
@@ -340,3 +340,25 @@ func debugPrinter(offset int, a ...any) {
 func errorPrinter(offset int) *pterm.PrefixPrinter {
 	return pterm.Error.WithShowLineNumber(logLevel > 2).WithLineNumberOffset(offset)
 }
+
+// Table prints a padded table containing the specified header and data
+func Table(header []string, data [][]string) {
```
Racer159, line 345: Usually when ordering the functions in a file the exported functions are above the unexported ones.  (so `Table` would go above `debugPrinter`)

Racer159 (changes requested):

bdw617 (dismissed):
Looks good to me, but would like to see the table have named columns versus [][]string{}, it will make it future proof for adding columns without lots of counting of parameter count :) 

I don't want to hold this up, but that would be a good follow up :)

Racer159 (approved):
lgtm

lucasrod16 (approved):
lgtm

src/pkg/packager/lint/lint.go
```diff
@@ -58,15 +53,28 @@ func ValidateZarfSchema(path string) (*Validator, error) {
 func checkForVarInComponentImport(validator *Validator) {
 	for i, component := range validator.typedZarfPackage.Components {
 		if strings.Contains(component.Import.Path, types.ZarfPackageTemplatePrefix) {
-			validator.addWarning(fmt.Sprintf("component.[%d].import.path will not resolve ZARF_PKG_TMPL_* variables", i))
+			validator.addWarning(fmt.Sprintf(".component.[%d].import.path: Will not resolve ZARF_PKG_TMPL_* variables", i))
```
Noxsios, line 56: one final consistency nit: this needs to be `.components`, not `.component` to be YQ compat

src/pkg/packager/lint/validator.go
```diff
@@ -22,8 +22,21 @@ type Validator struct {
 	untypedZarfPackage interface{}
 }
 
-func (v Validator) printWarningTable() {
-	if !v.IsSuccess() {
+// DisplayFormattedMessage message sent to user based on validator results
+func (v Validator) DisplayFormattedMessage() {
+	if !v.hasWarnings() && !v.hasErrors() {
+		message.Success(fmt.Sprintf("Schema validation successful for %q", v.typedZarfPackage.Metadata.Name))
```
Noxsios, line 28: use `message.Successf` for brevity

Noxsios (changes requested):

src/pkg/packager/lint/lint.go
AustinAbro321: Great catch, thanks

docs/3-create-a-zarf-package/index.md
```diff
@@ -21,6 +21,9 @@ To learn more about creating a Zarf package, you can check out the following res
 The general flow of a Zarf package deployment on an existing initialized cluster is as follows:
 
 ```shell
+# Before creating your package you can lint your zarf.yaml
+$ zarf package lint <directory>
```
Noxsios, line 25:
```suggestion
$ zarf prepare lint <directory>
```

Noxsios (changes requested):

Noxsios (approved):
lgtm, welcome to the party @AustinAbro321

PR #2076, opened 2023-10-16
chore: fix numbers in package templates

### LINKED ISSUE #2074: Better handling of package templates with numeric names
### Describing What Works
If you have the following project structure:
```yaml
# example zarf-config.yaml
package:
  create: 
    set:
      k3d_version: 5.5.2
```
```yaml
# example zarf.yaml
kind: ZarfPackageConfig
metadata:
  name: this-correctly-validates
 
components:
- name: download-assets
  required: true
  files:
    - source: https://github.com/k3d-io/k3d/releases/download/v###ZARF_PKG_K3D_VERSION###/k3d-linux-amd64
      target: k3d
      executable: true
```

Then running the `zarf package create . --confirm` command executes without any issues.

### Describing What Doesn't Work
If you were to remove the `zarf-config.yaml` file and re-run the create command without any changes, I would expect to see a warning stating
```
     ERROR:  Failed to create package: unable to fill values in template: template 'K3D_VERSION' must be '--set'
             when using the '--confirm' flag
```

Currently this doesn't happen and you end up receiving an error because zarf it attempting to download from a nonexistant url 
```
     ERROR:  Failed to create package: unable to add component: failed to download
             https://github.com/k3d-io/k3d/releases/download/v###ZARF_PKG_TMPL_K3D_VERSION###/k3d-linux-amd64:
             bad HTTP status: 404 Not Found
```
I would not expect Zarf to do this because it should have noticed that a `ZARF_PKG_TMPL...` variable value was not provided.

## Describe alternatives you've considered

Removing the digit `3` from the template results in the warning I expected. When I run the `zarf package create . --confirm` command I get the warning message that I'm not supplying a necessary value for a template (note the template name is now `KD_VERSION` instead of `K3D_VERSION`:

```
     ERROR:  Failed to create package: unable to fill values in template: template 'KD_VERSION' must be '--set'
             when using the '--confirm' flag
```

## Desired Outcome
As a package creator, I expect to be warned if I'm forgetting to supply a mandatory template value. I would expect zarf to either
a) Warn me that I my zarf.yaml has an invalid configuration (in this case an invalid character (0-9) in my template name)
b) Tell me that I'm forgetting to provide the value for a mandatory template.

Racer159:
## Description

Fix issues where the template regex does not match that of constants and variables.

## Related Issue

Fixes #2074

## Type of change

## Checklist before merging

### DIFF
--- a/.github/workflows/test-external.yml
+++ b/.github/workflows/test-external.yml
@@ -34,6 +34,9 @@ jobs:
 
       - name: Build binary and zarf packages
         uses: ./.github/actions/packages
+        with:
+          init-package: "false"
+          build-examples: "false"
 
       - name: Setup k3d
         uses: ./.github/actions/k3d
--- a/src/pkg/utils/yaml.go
+++ b/src/pkg/utils/yaml.go
@@ -178,7 +178,7 @@ func FindYamlTemplates(config any, prefix string, suffix string) (map[string]str
 	}
 
 	// Find all strings that are between the given prefix and suffix
-	r := regexp.MustCompile(fmt.Sprintf("%s([A-Z_]+)%s", prefix, suffix))
+	r := regexp.MustCompile(fmt.Sprintf("%s([A-Z0-9_]+)%s", prefix, suffix))
 	matches := r.FindAllStringSubmatch(string(text), -1)
 
 	for _, match := range matches {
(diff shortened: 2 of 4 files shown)


Noxsios (approved):
lgtm ty @YrrepNoj  ❤️

PR #2077, opened 2023-10-16
fix: package publishing an already signed package

Racer159:
## Description

This fixes an issue with attempting to publish an already signed package through the CLI.

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging


lucasrod16 (changes requested):
We need to update the nightly ECR test to require no error when pulling signed Zarf packages without providing a public key since we do not perform signature validation on pull:

https://github.com/defenseunicorns/zarf/blob/09ec9e7f05d75409f8a6bc136d94a112dae098ae/src/test/nightly/ecr_publish_test.go#L68

We should also remove the part where we are pulling the signed package with a provided public key in the ECR test for the same reason that there is no signature validation on pull anymore:

https://github.com/defenseunicorns/zarf/blob/09ec9e7f05d75409f8a6bc136d94a112dae098ae/src/test/nightly/ecr_publish_test.go#L73

src/cmd/package.go
```diff
@@ -410,8 +410,8 @@ func bindRemoveFlags(v *viper.Viper) {
 
 func bindPublishFlags(v *viper.Viper) {
 	publishFlags := packagePublishCmd.Flags()
-	publishFlags.StringVarP(&pkgConfig.PublishOpts.SigningKeyPath, "key", "k", v.GetString(common.VPkgPublishSigningKey), lang.CmdPackagePublishFlagSigningKey)
-	publishFlags.StringVar(&pkgConfig.PublishOpts.SigningKeyPassword, "key-pass", v.GetString(common.VPkgPublishSigningKeyPassword), lang.CmdPackagePublishFlagSigningKeyPassword)
+	publishFlags.StringVar(&pkgConfig.PublishOpts.SigningKeyPath, "signing-key", v.GetString(common.VPkgPublishSigningKey), lang.CmdPackagePublishFlagSigningKey)
+	publishFlags.StringVar(&pkgConfig.PublishOpts.SigningKeyPassword, "signing-key-pass", v.GetString(common.VPkgPublishSigningKeyPassword), lang.CmdPackagePublishFlagSigningKeyPassword)
```
Noxsios, line 414: note for @Racer159 , will need to call out this breaking change in release notes

src/pkg/packager/publish.go
```diff
@@ -26,7 +26,7 @@ import (
 // Publish publishes the package to a registry
 func (p *Packager) Publish() (err error) {
 	_, isOCISource := p.source.(*sources.OCISource)
-	if isOCISource {
+	if isOCISource && p.cfg.PublishOpts.SigningKeyPath == "" {
```
Noxsios, line 29: its a bit more effort, but we _can_ use the filtering behavior of our copy fn to not copy over the existing sig, then pull down the `zarf.yaml`, then re-sign, then push those 2 layers back up, then make a new manifest and push. If we wanted to go for max efficiency that would do the trick, but it does introduce some complexity and needed understanding of the push/pull flow.

Noxsios (approved):

lucasrod16 (approved):
lgtm

src/pkg/packager/publish.go
Racer159: added a comment to the component layers tech debt - it would be more efficient but wanted to fix things faster

PR #2083, opened 2023-10-18
feat: moving cli config to --set for nodeport and storageclass

### LINKED ISSUE #1725: Make non Zarf-generated init values Zarf variables
Some init configuration is passed in on flags like `--nodeport` and `--storage-class` and is templated into our init package globally.

Instead of doing this we could convert these values to Zarf variables and alias the init flags to those variables to provide more flexibility.

- **Given** I have an uninitialized cluster
- **When** I run `zarf init --set REGISTRY_NODEPORT=31998 --set REGISTRY_STORAGE_CLASS=my-storage-class `
- **Then** the `nodeport` and `storageClass` values will be templated

> ⚠️ **NOTE**: We should still maintain backwards compatibility and map `--nodeport` and `--storage-class` to their appropriate Zarf variables

### Additional context
This would also allow these variables to be set through a package deploy of the init package.

bdw617:
## Description

replace -nodeport and --storageclass: `zarf init --nodeport <number> --storage-class <my-storage-class>`
with: `zarf init --set REGISTRY_NODEPORT=31998 --set REGISTRY_STORAGE_CLASS=my-storage-class`

old variables still exist and work, but warnings are given to the user if they use them as deprecated. 

NOTE:  `storageClass to be used for gitea and the registry will by default be the default storageClass of the cluster.` 
 
## Related Issue

Fixes #1725 

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

## Checklist before merging

- [x ] [Contributor Guide Steps](https://github.com/defenseunicorns/zarf/blob/main/CONTRIBUTING.md#developer-workflow) followed


src/cmd/initialize.go
```diff
@@ -73,6 +81,62 @@ var initCmd = &cobra.Command{
...
+	if internalRegistry {
+		if configuredNodePort > 32767 || configuredNodePort < 30000 {
+			configuredNodePort = config.ZarfInClusterContainerRegistryNodePort
+		}
+		pkgConfig.PkgOpts.SetVariables[configVar] = strconv.Itoa(configuredNodePort)
+		pkgConfig.InitOpts.RegistryInfo.Address = fmt.Sprintf("%s:%d", helpers.IPV4Localhost, configuredNodePort)
+		pkgConfig.InitOpts.RegistryInfo.NodePort = configuredNodePort
+	} else {
+		// do not set the nodeport if this is an external registry
+		pkgConfig.PkgOpts.SetVariables[configVar] = ""
```
lucasrod16, line 120: We should be able to remove this else block. If the user specifies an external registry, we shouldn't have to set the `REGISTRY_NODEPORT` var to an empty string

src/cmd/initialize_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"testing"
+)
+
+func TestSetRegistryStorageClass(t *testing.T) {
+	// Test case 1: storageClass is set
+	pkgConfig.PkgOpts.SetVariables = map[string]string{"REGISTRY_STORAGE_CLASS": "test-storage-class"}
+	storageClassArg = ""
+	setRegistryStorageClass()
+	if storageClassArg != "test-storage-class" {
+		t.Errorf("Expected storage class to be set to 'test-storage-class', but got '%s'", storageClassArg)
```
lucasrod16, line 14: We should use `require.Equal()` to be more consistent with the rest of our test suite assertions. When `require.Equal()` fails it prints an "expected but got" message by default for you.

src/cmd/initialize_test.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	// Test case 2: storageClass is not set, use old way
+	pkgConfig.PkgOpts.SetVariables = map[string]string{}
+	storageClassArg = "old-storage-class"
+	setRegistryStorageClass()
+	if pkgConfig.PkgOpts.SetVariables["REGISTRY_STORAGE_CLASS"] != "old-storage-class" {
+		t.Errorf("Expected storage class to be set to old way value, but got '%s'", pkgConfig.PkgOpts.SetVariables["REGISTRY_STORAGE_CLASS"])
```
lucasrod16, line 22: We should use `require.Equal()` to be more consistent with the rest of our test suite assertions. When `require.Equal()` fails it prints an "expected but got" message by default for you.

src/cmd/initialize_test.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	// Test case 3: neither is set, should be empty
+	pkgConfig.PkgOpts.SetVariables = map[string]string{}
+	storageClassArg = ""
+	setRegistryStorageClass()
+	if pkgConfig.PkgOpts.SetVariables["REGISTRY_STORAGE_CLASS"] != "" {
+		t.Errorf("Expected storage class to be set to empty string, but got '%s'", pkgConfig.PkgOpts.SetVariables["REGISTRY_STORAGE_CLASS"])
```
lucasrod16, line 30: We should use `require.Equal()` to be more consistent with the rest of our test suite assertions. When `require.Equal()` fails it prints an "expected but got" message by default for you.

src/pkg/packager/deploy.go
```diff
@@ -216,6 +215,18 @@ func (p *Packager) deployInitComponent(component types.ZarfComponent) (charts []
...
+		// Retrieve the default storage class from the cluster, it will be the default
+		if p.cfg.PkgOpts.SetVariables == nil {
+			p.cfg.PkgOpts.SetVariables = map[string]string{}
+		}
+		if c := p.cfg.PkgOpts.SetVariables["REGISTRY_STORAGE_CLASS"]; c == "" {
+			p.cfg.PkgOpts.SetVariables["REGISTRY_STORAGE_CLASS"], err = p.cluster.GetDefaultStorageClass()
+		}
+
+		if err != nil {
+			message.WarnErr(err, "Unable to get the default storage class from the cluster")
```
lucasrod16, line 228: I think we have an opportunity to fail fast here. If `REGISTRY_PVC_ENABLED` is true (this is the default), the user did not specify a storage class to use, and there was no default storage class found in the cluster, we should fail fast and notify the user that a storage class is needed for the registry to deploy successfully.

RKE2 clusters come to mind since they don't provide a default storage class

lucasrod16:
Closing this PR because we will likely rethink the implementation

Issue #2030, opened 2023-09-25
Update example Pepr webhook package to use new Kubernetes fluent client


lucasrod16:
### Describe what should be investigated or refactored
The example Pepr webhook package should be updated to use the new [Kubernetes fluent client](https://github.com/defenseunicorns/kubernetes-fluent-client) once released and available.

### Links to any relevant code
https://github.com/defenseunicorns/zarf/blob/bdba56599d99642147b5cfbf7637b01126c0cfdc/examples/component-webhooks/capabilities/hook.ts#L89

PR #2090, opened 2023-10-19
chore: update component-webhooks example to use new Pepr k8s fluent client

lucasrod16:
## Description

Update Pepr to `v0.15.0` and use new k8s fluent client

- adds types for parsing zarf package secret data
- adds a `zarf internal gen-types-schema` command to generate a JSON schema from Zarf types (DeployedPackage, ZarfPackage, ZarfState). This command was used to auto-generate the typescript interfaces in `examples/component-webhook/capabilities/zarf-types.ts`

## Related Issue

Fixes https://github.com/defenseunicorns/zarf/issues/2030

## Type of change

## Checklist before merging


examples/component-webhooks/capabilities/hook.ts
```diff
@@ -86,46 +86,60 @@ When(a.Secret)
 async function sleepAndChangeStatus(secretName: string, componentName: string) {
   await sleep(10);
```
bdw617, line 87: I'd highly recommend replacing this, if this needs to wait for something to happen, put it in a retry loop and Log something if the time expires.

examples/component-webhooks/pepr-module-cb5693ef-d13c-5fe1-b5ad-c870fd911b3b.yaml
bdw617, line 1: I highly recommend not checking in pre-built pepr module:
`npx pepr build` should generate this.

bdw617 (changes requested):
Happy to help you work on these tomorrow morning :)

examples/component-webhooks/capabilities/hook.ts
lucasrod16: As discussed synchronously, the sleep here isn't waiting for anything to happen. It is just there to pause execution of a zarf deployment to simulate a background process running (like creating ECR repositories).

The goal for this example isn't to necessarily demonstrate things you can automate with pepr, but rather to demonstrate how pepr can be used to inject custom logic into the deployment lifecycle, similar to zarf actions, with a key differentiator being that pepr modules are defined outside of a `zarf.yaml`. So we shouldn't need any retry logic because the only thing that is happening is zarf waiting for pepr to update the webhook status to something other than `Running` in the package secret. Zarf implements the polling/retry logic while waiting for pepr to update the webhook status here: https://github.com/defenseunicorns/zarf/blob/7cc12011f863d48168202a4a4e2fb60e501cb560/src/internal/cluster/zarf.go#L126

It would be good to set the [`WaitDurationSeconds`](https://github.com/defenseunicorns/zarf/blob/7cc12011f863d48168202a4a4e2fb60e501cb560/src/types/k8s.go#L72C5-L72C5) field to specify how long zarf should poll the webhook status before timing out as a best practice though.

examples/component-webhooks/pepr-module-cb5693ef-d13c-5fe1-b5ad-c870fd911b3b.yaml
lucasrod16: I'm not aware of an e2e workflow that would allow us to not push a locally built pepr module. The flow I'm aware of and have been using is:

1. make changes to pepr typescript code
2. `npx pepr build` (this updates the pepr k8s manifest with the latest build)
3. `zarf package create` (this creates a new example package that we use in CI with the updated manifest)
4. `zarf package deploy` to a zarf initialized cluster

We could maybe add `npx pepr bulid` to the [component webhooks e2e test](https://github.com/defenseunicorns/zarf/blob/7cc12011f863d48168202a4a4e2fb60e501cb560/src/test/e2e/33_component_webhooks_test.go) but we would still need some way to check in the latest changes/build to version control

lucasrod16: And any changes to the typescript code will be visible in the commit diff along with the updated pepr manifest

examples/component-webhooks/capabilities/zarf-types.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface Webhook {
+  name: string;
+  observedGeneration: number;
+  status: string;
+  /**
+   * The number of seconds that Zarf will wait for the webhook to run before timing out and returning an error.
+   */
+  waitDurationSeconds?: number;
+}
```
Racer159, line 23: did we want to generate these similar to how the zarf init aws package works? It may make sense to just add an `internal` command to get types out of zarf more generically and use it in both places (and allow others to use it too)

Racer159 (changes requested):
just a small thing that would be useful longer term

examples/component-webhooks/capabilities/zarf-types.ts
lucasrod16: Added a `zarf internal generate-types-schema` command

src/cmd/internal.go
```diff
@@ -198,7 +218,8 @@ func init() {
 	internalCmd.AddCommand(agentCmd)
 	internalCmd.AddCommand(httpProxyCmd)
 	internalCmd.AddCommand(generateCLIDocs)
```
Racer159, line 220:
```suggestion
	internalCmd.AddCommand(genCLIDocs)
```

src/cmd/internal.go
```diff
@@ -125,6 +125,26 @@ var configSchemaCmd = &cobra.Command{
...
 }
 
+type zarfTypes struct {
+	DeployedPackage types.DeployedPackage
+	ZarfPackage     types.ZarfPackage
+	ZarfState       types.ZarfState
+}
+
+var genTypesSchemaCmd = &cobra.Command{
+	Use:     "generate-types-schema",
```
Racer159, line 135:
```suggestion
	Use:     "gen-types-schema",
```

src/cmd/internal.go
```diff
@@ -111,9 +111,9 @@ var generateCLIDocs = &cobra.Command{
 	},
 }
 
-var configSchemaCmd = &cobra.Command{
-	Use:     "config-schema",
-	Aliases: []string{"c"},
+var genConfigSchemaCmd = &cobra.Command{
+	Use:     "generate-config-schema",
```
Racer159, line 115:
```suggestion
	Use:     "gen-config-schema",
```

To better match `zarf tools gen-pki/gen-key`

Racer159: Same would apply to `generate-cli-docs`

Racer159 (changes requested):
just some consistency changes

Racer159 (approved):
lgtm

bdw617 (approved):

Noxsios (approved):

bdw617 (approved):

PR #2095, opened 2023-10-23
feat: allow passing additional arguments to k9s when invoked by `zarf tools monitor`

### LINKED ISSUE #2094: Allow arguments to be passed to included k9s
Allow the command: zarf tools monitor to pass the remaining commands to k9s. Allows full usage of the k9s command line.

Alternatives are k9s creates a configuration file in ~/.config/k9s/config.yaml. This can be modified instead but still seems useful to pass commands to it as well.

WeaponX314:
## Description
Allow additional arguments to be passed to k9s

Fixes #2094

## Type of change

## Checklist before merging


WeaponX314:
I am unsure how you want to document this if in fact you are ok with this being added.

Noxsios (approved):

Racer159 (commented):
One way that we could document this is with `//go:linkname`.  It is a bit of a hack but if it is limited to `cmd/tools` that is likely ok.  Here is what that would look like:

```go

// Package tools contains the CLI commands for Zarf.
package tools

import (
	"os"

	"github.com/defenseunicorns/zarf/src/config/lang"
	k9s "github.com/derailed/k9s/cmd"
	"github.com/spf13/cobra"

	_ "unsafe"
)

//go:linkname k9sRootCmd github.com/derailed/k9s/cmd.rootCmd
var k9sRootCmd *cobra.Command

func init() {
	k9sCmd := &cobra.Command{
		Use:     "monitor",
		Aliases: []string{"m", "k9s"},
		Short:   lang.CmdToolsMonitorShort,
			// Hack to make k9s think it's all alone
			os.Args = []string{os.Args[0]}
			k9s.Execute()
		},
	}

	k9sCmd.Flags().AddFlagSet(k9sRootCmd.Flags())

	toolsCmd.AddCommand(k9sCmd)
}
```

Then you could run `make docs-and-schema` to auto gen the CLI docs and the flags will show in the help information (and we won't have to `DisableFlagParsing`)

Racer159 (approved):
Approving for now but will update with the comments later to add the docs / help.

PR #2105, opened 2023-10-27
feat: introduce the `only.flavor` key to allow building of package variants

### LINKED ISSUE #2101: Add an `only.flavor` key to Zarf components
**As** Ezra **I want** the ability to specify "flavors" of Zarf packages and select between them on `zarf package create` **so that** I can swap chunks of things like `images` and `valuesFiles` without the need to have an entirely separate copy of a Zarf package definition.

- **Given** I have a Zarf package definition
- **And** that definition has components that specify `only.flavor`
- **When** I run `zarf package create --flavor my-flavor`
- **Then** Zarf will include all components that match that flavor or have an empty flavor string
- **And When** I run `zarf package create` without a `flavor`
- **Then** Zarf will include only components that have an empty flavor string

We could have higher order keys like `imageSets` or have other extensions but this adds complexity without taking advantage of a lot of the base Zarf primitive manipulation that can be done through composability.

### Additional context
This initial ask came from a desire to swap images to have a hardened and an upstream version of the same underlying package to help with package development and testing.

Racer159:
## Description

This implements the `only.flavor` filter as a replacement for component groups to allow the feature to be more declarative when building package variants.

## Related Issue

Fixes #2101

## Type of change

## Checklist before merging


src/pkg/packager/deprecated/common.go
```diff
@@ -81,7 +81,7 @@ func MigrateComponent(build types.ZarfBuildData, component types.ZarfComponent)
 func PrintBreakingChanges(deployedZarfVersion string) {
 	deployedSemver, err := semver.NewVersion(deployedZarfVersion)
 	if err != nil {
-		message.Warnf("Unable to determine init-package version from %s.  There is potential for breaking changes.", deployedZarfVersion)
+		message.Debugf("Unable to check for breaking changes between Zarf versions")
```
jeff-mccoy, line 84: ❤️

src/pkg/packager/compose.go
```diff
@@ -20,12 +20,16 @@ func (p *Packager) composeComponents() error {
 	for _, component := range p.cfg.Pkg.Components {
 		arch := p.arch
 		// filter by architecture
-		if component.Only.Cluster.Architecture != "" && component.Only.Cluster.Architecture != arch {
+		if !composer.CompatibleComponent(component, arch, p.cfg.CreateOpts.Flavor) {
 			continue
+		} else {
+			// if a match was found, strip flavor and architecture to reduce bloat in the package definition
```
jeff-mccoy, line 26: Good add

jeff-mccoy (dismissed):
Straightforward feature add, lgtm. Love the decision to call it `flavor` and support a migration path from the deprecated `group` field.

jeff-mccoy (dismissed):

jeff-mccoy (approved):
🫡

PR #2113, opened 2023-10-31
add pkg templating characters to component schema pattern

### LINKED ISSUE #2112: Update zarf.schema.json to ignore `###ZARF_PKG_TMPL_` variables in component name
I am creating a package that is intended to be the skeleton for many other packages. Because of this, I would like to variables the component name. Right now, the pattern defined in the `zarf.schema.json` does not recognize the variable syntax since the `#` character is unallowed. 

This is a super low importance nice-to-have, but it would be nice for the pattern to be updated to no longer complain.

YrrepNoj:
Closes #2112 

https://regex101.com/r/9jlZfq/1

### DIFF
--- a/src/types/component.go
+++ b/src/types/component.go
@@ -13,7 +13,7 @@ import (
 // ZarfComponent is the primary functional grouping of assets to deploy by Zarf.
 type ZarfComponent struct {
 	// Name is the unique identifier for this component
-	Name string `json:"name" jsonschema:"description=The name of the component,pattern=^[a-z0-9\\-]+$"`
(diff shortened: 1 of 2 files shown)


YrrepNoj:
I think I just realized that I don't actually need this.. I'm not really sure what Jon Perry of 30 minutes ago was thinking...

😪

YrrepNoj:
I was too excited about knowing how to implement a REGEX, I didn't stop to ask myself if I really needed it.

I deeply apologize for the dishonor.

PR #2117, opened 2023-10-31
fix: adds error channel to RenderProgressBarForLocalDirWrite

UncleGedd:
## Description
Adds an error channel to `utils.RenderProgressBarForLocalDirWrite` to handle the case where an operation  (e.g. `oras.Copy`) is not successful and we want to kill the goroutine without displaying a 'success' message


src/internal/packager/images/pull.go
```diff
@@ -188,7 +188,7 @@ func (i *ImageConfig) PullAll() ([]ImgInfo, error) {
 	var progressBarWaitGroup sync.WaitGroup
 	progressBarWaitGroup.Add(1)
 	updateText := fmt.Sprintf("Pulling %d images", imageCount)
-	go utils.RenderProgressBarForLocalDirWrite(i.ImagesPath, totalBytes, &progressBarWaitGroup, doneSaving, updateText, updateText)
+	go utils.RenderProgressBarForLocalDirWrite(i.ImagesPath, totalBytes, &progressBarWaitGroup, doneSaving, nil, updateText, updateText)
```
UncleGedd, line 191: There's some other logic happening with the wait group, see [onLayerWritingError](https://github.com/defenseunicorns/zarf/blob/6a6be5439eec1a1740ab7b8626ae79a0863878e8/src/internal/packager/images/pull.go#L325)

Racer159: onLayerWritingError should be refactored to use an `errorChan` instead of `doneSaving`

Racer159 (changes requested):

Racer159 (approved):
lgtm

PR #2118, opened 2023-10-31
feat: provide the option to allow for Gitea registration, so that SSO can be used

dgershman:
Provide the option to allow for Gitea registration, so that SSO can be used

## Description

We like to use Gitea with Keycloak for SSO.  In order to do that, registration has to be enabled.  This provides the capability as an option, but left as `true` to disable registration by default.

## Related Issue

N/A

## Type of change

## Checklist before merging


Noxsios (approved):

Racer159 (approved):
lgtm

PR #2131, opened 2023-11-07
chore: initial implementation of Helm Chart overrides at deploy time

### LINKED ISSUE #2127: Pass arbitrary helm values files at package deploy time

I want to override helm chart values by providing an arbitrary set of helm values files during zarf package deploy.

- **Given** a Zarf package with a Helm Chart
- **When** a package deployer provides values files
- **Then** Zarf overlays those values files onto the identified package:chart at deploy time

Zarf variables are an attractive option when there are only a handful of settings, but applications with dozens or hundreds of `key:value` pairs cannot be reasonably managed by exposing every setting. Further, maintaining unique Zarf packages with the myriad helm values files is untenable.

Zarf packages contain an unbounded number of components with an unbounded number of helm charts, and the desire is to allow an unbounded number of values files to be written. The `1:many:many:many` relationship from package -> component -> chart -> values file.

Recommend adding a new Viper package deploy key `VPkgDeployComponentChart` that allows referencing helm values file overlays for named charts. An example zarf config file follows.

```yaml
package:
  deploy:
    components: 'postgres-database'
      charts:
        - name: zalando-postgres
          files:
            - values/ha.yaml
            - values/env-b.yaml
```

Racer159:
## Description

This is an initial implementation for Helm Chart overrides on deploy as library code.

## Related Issue

Fixes #2127 

## Type of change

## Checklist before merging


src/internal/packager/helm/utils.go
```diff
@@ -61,7 +62,12 @@ func (h *Helm) parseChartValues() (chartutil.Values, error) {
 	}
 
 	providers := getter.Providers{httpProvider}
-	return valueOpts.MergeValues(providers)
+	chartValues, err := valueOpts.MergeValues(providers)
+	if err != nil {
+		return chartValues, err
+	}
+
+	return helpers.MergeMapRecursive(chartValues, h.ValuesOverrides), nil
```
Noxsios, line 70: what happens if `h.ValuesOverrides` is `nil`?

Noxsios (changes requested):

src/internal/packager/helm/utils.go
Racer159: it will just return

Noxsios (approved):
after sync-up, lgmt

PR #2139, opened 2023-11-13
fix: first step to split up the e2e need a cluster, versus ones that do not

bdw617:
## Description

updated the makefile and the github action to reduce the work for non-cluster e2e tests. 

## Related Issue

Fixes #

Relates to #

## Type of change

- [x ] Other (security config, docs update, etc)

## Checklist before merging

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -43,7 +43,7 @@ jobs:
           path: build/
           retention-days: 1
 
-  validate-e2e-tests-without-cluster:
+  validate-without-cluster:
     runs-on: ubuntu-latest
     needs: build
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/test-e2e.yml
```diff
@@ -43,6 +43,38 @@ jobs:
           path: build/
           retention-days: 1
 
+  validate-e2e-tests-without-cluster:
```
Racer159, line 46:
```suggestion
  validate-without-cluster:
```

Given this is the test-e2e workflow having e2e-tests in the name here stutters

Racer159 (changes requested):
small request to prevent a naming stutter (we usually try to avoid including the name of a thing twice) https://stackoverflow.com/questions/45885985/ways-to-avoid-stuttering-in-go-package-and-struct-names

Racer159 (approved):
lgtm

PR #2140, opened 2023-11-13
feat: add a custom delimited Go template to Zarf templating

### LINKED ISSUE #1483: Conditional Logic with Zarf Variables

When configuring applications with Zarf deploy time variables, there is not always a very clean way to inject the variable as a single string into an application.  For instance, when adding a [Slack endpoint for alerting with BigBang](https://github.com/defenseunicorns/zarf-package-big-bang/pull/50/files#diff-094aae8ebcac39c231aab9b0dc13de1214977bf582dcd3db724cf7703ba1a9c4R122) There's several objects that need to get created to correctly configure the alerts.  I would like to conditional include these manifests when the SLACK_URL variable is non empty.

A more robust templating language (go-templating????) could be one approach that solves this

I could make a helm chart with these manifests inside of them and have them conditioned correctly in the helm chart since the templating logic that I desire is already present in Helm.  This would just create an additional layer between Zarf and the manifests.  Since Zarf already wraps manifests in a helm chart for deployment, it doesn't create any additional complexity on the deployment side, just on the creation side.

Perhaps there's a way to expose helm templating to manifests as they get injected into the wrapper Helm chart during the deploy phase that we'd be able to leverage?

Racer159:
## Description

This is an experimental pull request to add Go templating to Zarf text templates.

Cue and YTT were also evaluated and if this is adopted an ADR should be created.

## Related Issue

Fixes #1483 

## Type of change

## Checklist before merging


bdw617:
Updated example of go templating into a values file:
example below simply populates a value in the podinfo values file with HiHowAreYou. This has been no means tested, and almost certainly NOT the way we're gong to solve this problem in the long term. 

```
 zarf package create examples/helm-charts
 zarf package deploy zarf-package-helm-charts-amd64-0.0.1.tar.zst --set HELLO=HiHowAreYou --components=demo-helm-local-chart --confirm -l trace
```

Noxsios:
Closing for now, mostly due to stale reasons, but also the scope of this effort is still not well-defined.

PR #2145, opened 2023-11-15
fix: updated to not clobber whatever k3d cluster may be running, updated gitea to same version as zarf init

bdw617:
## Description

Very small change to make sure that running the external tests locally do not just clobber a k3d cluster you might have running that uses the default name.

updated the test to use the same version of gitea as the one in the current zarf init
## Related Issue

Fixes #

Relates to #

## Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist before merging


Racer159 (approved):
lgtm

PR #2148, opened 2023-11-16
fix: multiline actions fail faster

UncleGedd:
## Description
Adds `-e` to shells so that multiline actions fail faster. Noting that this only applies to [real shells](https://github.com/defenseunicorns/zarf/blob/356a4da121ae8cd0ffa720ab17ab859c5a2ac07a/src/pkg/utils/exec/exec.go#L173) 🤣 

EDIT: adding capability to Powershell based on comments below

## Related Issue

Fixes https://github.com/defenseunicorns/uds-cli/issues/204

## Type of change

## Checklist before merging

### DIFF
--- a/examples/component-actions/zarf.yaml
+++ b/examples/component-actions/zarf.yaml
@@ -211,3 +211,13 @@ components:
                 kind: configmap
                 name: simple-configmap
                 namespace: zarf
+
+  - name: on-deploy-immediate-failure
+    description: This component will fail on the first error instead of continuing execution
+   # the default for multi-line commands is set -e
+    actions:
+      onDeploy:
+        after:
+          - cmd: |
+              bad_cmd
+              echo "this text shouldn't be printed"
--- a/src/pkg/utils/exec/exec.go
+++ b/src/pkg/utils/exec/exec.go
@@ -154,9 +154,9 @@ func LaunchURL(url string) error {
 }
 
 // GetOSShell returns the shell and shellArgs based on the current OS
-func GetOSShell(shellPref types.ZarfComponentActionShell) (string, string) {
+func GetOSShell(shellPref types.ZarfComponentActionShell) (string, []string) {
 	var shell string
-	var shellArgs string
+	var shellArgs []string
 
 	switch runtime.GOOS {
 	case "windows":
@@ -165,13 +165,13 @@ func GetOSShell(shellPref types.ZarfComponentActionShell) (string, string) {
 			shell = shellPref.Windows
 		}
 
-		shellArgs = "-Command"
+		shellArgs = []string{"-Command", "$ErrorActionPreference = 'Stop';"}
 		if shell == "cmd" {
 			// Change shellArgs to /c if cmd is chosen
-			shellArgs = "/c"
+			shellArgs = []string{"/c"}
 		} else if !IsPowershell(shell) {
 			// Change shellArgs to -c if a real shell is chosen
-			shellArgs = "-c"
+			shellArgs = []string{"-e", "-c"}
 		}
 	case "darwin":
 		shell = "sh"
@@ -179,10 +179,10 @@ func GetOSShell(shellPref types.ZarfComponentActionShell) (string, string) {
 			shell = shellPref.Darwin
 		}
 
-		shellArgs = "-c"
+		shellArgs = []string{"-e", "-c"}
 		if IsPowershell(shell) {
 			// Change shellArgs to -Command if pwsh is chosen
-			shellArgs = "-Command"
+			shellArgs = []string{"-Command", "$ErrorActionPreference = 'Stop';"}
 		}
 	case "linux":
 		shell = "sh"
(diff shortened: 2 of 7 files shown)


UncleGedd:
My 🤖 pair and I did look for Windows equivalents but came up empty
- [Powershell docs](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_exe?view=powershell-5.1)
- [ cmd docs](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/cmd#syntax)

mikevanhemert (approved):

zachariahmiller:
@UncleGedd for powershell, its been a while, but  if i recall correctly, you can either specify the error action on a per cmdlet call basis or with a global preference. https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.3

So if you are running say, [invoke-command](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/invoke-command?view=powershell-7.3) you should be able to provide and error action preference variable as part of the [CommonParameters](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_commonparameters?view=powershell-7.3)

Noxsios:
Additionallly, you can set a "global" ErrorActionPreference

```ps
$ErrorActionPreference = "Stop"
```

mjnagel (dismissed):

src/pkg/utils/exec/exec.go
```diff
@@ -165,39 +165,39 @@ func GetOSShell(shellPref types.ZarfComponentActionShell) (string, string) {
...
-		shellArgs = "-Command"
+		shellArgs = append(shellArgs, "-Command", "$ErrorActionPreference = 'Stop';")
 		if shell == "cmd" {
 			// Change shellArgs to /c if cmd is chosen
-			shellArgs = "/c"
+			shellArgs = append(shellArgs, "/c")
 		} else if !IsPowershell(shell) {
 			// Change shellArgs to -c if a real shell is chosen
-			shellArgs = "-c"
+			shellArgs = append(shellArgs, "-e", "-c")
```
Racer159, line 174: This will cause bad shell args when shell preferences are used:

```
bash -Command $ErrorActionPreference = 'Stop'; -e -c
```

src/pkg/utils/exec/exec.go
```diff
@@ -165,39 +165,39 @@ func GetOSShell(shellPref types.ZarfComponentActionShell) (string, string) {
...
 		if shellPref.Darwin != "" {
 			shell = shellPref.Darwin
 		}
 
-		shellArgs = "-c"
+		shellArgs = append(shellArgs, "-e", "-c")
 		if IsPowershell(shell) {
 			// Change shellArgs to -Command if pwsh is chosen
-			shellArgs = "-Command"
+			shellArgs = append(shellArgs, "-Command")
```
Racer159, line 185: This breaks powershell on macOS

src/pkg/utils/exec/exec.go
```diff
@@ -165,39 +165,39 @@ func GetOSShell(shellPref types.ZarfComponentActionShell) (string, string) {
...
 		if shellPref.Linux != "" {
 			shell = shellPref.Linux
 		}
 
-		shellArgs = "-c"
+		shellArgs = append(shellArgs, "-e", "-c")
 		if IsPowershell(shell) {
 			// Change shellArgs to -Command if pwsh is chosen
-			shellArgs = "-Command"
+			shellArgs = append(shellArgs, "-Command")
```
Racer159, line 196: This breaks powershell on linux

src/pkg/utils/exec/exec.go
```diff
@@ -165,39 +165,39 @@ func GetOSShell(shellPref types.ZarfComponentActionShell) (string, string) {
...
+			shellArgs = append(shellArgs, "-e", "-c")
 		}
 	case "darwin":
 		shell = "sh"
 		if shellPref.Darwin != "" {
 			shell = shellPref.Darwin
 		}
 
-		shellArgs = "-c"
+		shellArgs = append(shellArgs, "-e", "-c")
```
Racer159, line 182:
```suggestion
		shellArgs = "-ce")
```
Shell args can be combined on POSIX (real) shells and then the array complexity might not be needed.

Racer159, line 182: We should document `set -e` as the default behavior in an admonition in the actions documentation under the `cmd` section (https://github.com/defenseunicorns/zarf/blob/356a4da121ae8cd0ffa720ab17ab859c5a2ac07a/docs/3-create-a-zarf-package/7-component-actions.md?plain=1#L54)

examples/component-actions/zarf.yaml
```diff
@@ -211,3 +211,12 @@ components:
                 kind: configmap
                 name: simple-configmap
                 namespace: zarf
+
```
Racer159, line 214: A comment stating that `set -e` is the default for commands could be added here or in the `on-create` component that is intro-ing multi line commands (~ line 54)

src/pkg/utils/exec/exec.go
Racer159: Also the `$ErrorActionPreference = 'Stop'` should be here too.

src/pkg/utils/exec/exec.go
Racer159: Also the `$ErrorActionPreference = 'Stop'` should be here too.

Racer159 (changes requested):

Racer159 (approved):
lgtm

PR #2160, opened 2023-11-27
fix: update error message when the image doesn't exist locally or on a remote

### LINKED ISSUE #2081: zarf's message when you hit the pull rate limit isn't helpful
### Steps to reproduce
1. do not be logged into docker.io
2. run `make test-e2e` too often

workaround is to simply login to docker.io. but still we shouldn't use it as part of our constantly run testing as much.

### Expected result
it should run successfully.

### Actual Result

➜  zarf git:(main) ✗ docker pull docker.io/defenseunicorns/zarf-game:multi-tile-dark
Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

```
  ⠋  Falling back to docker for docker.io/defenseunicorns/zarf-game:multi-tile-dark. This may take some time.
  ⠋  Falling back to docker for docker.io/defenseunicorns/zarf-game:multi-tile-dark. This may take some time.
  ⠋  Falling back to docker for docker.io/defenseunicorns/zarf-game:multi-tile-dark. This may take some time.
     ERROR:  Failed to create package: unable to pull images after 3 attempts: failed to load metadata for all
             images. This may be due to a network error or an invalid image reference: failed to pull image
             docker.io/defenseunicorns/zarf-game:multi-tile-dark: failed to inspect image
             docker.io/defenseunicorns/zarf-game:multi-tile-dark via docker: Error response from daemon: No such
             image: defenseunicorns/zarf-game:multi-tile-dark
```

### Severity/Priority
Low

bdw617:
## Description

This fixes a bug where if zarf (via crane) can't pull an image from a remote repository, and a local tarball doesn't exist, it will output the proper error from crane, versus hiding the issue as "image does not exist"

## Related Issue

Fixes #2081 

Relates to #

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist before merging


bdw617:
will work on a test for this now.

src/internal/packager/images/pull.go
```diff
@@ -434,6 +434,44 @@ func (i *ImageConfig) PullAll() ([]ImgInfo, error) {
...
+	cli.NegotiateAPIVersion(ctx)
+
+	// Inspect the image to get the size.
+	rawImg, _, err := cli.ImageInspectWithRaw(ctx, src)
+	if err != nil {
+		return nil, fmt.Errorf("failed to inspect image %s via docker: %w", src, err)
+	}
+
+	// Warn the user if the image is large.
+	if rawImg.Size > 750*1000*1000 {
```
AustinAbro321, line 460: Would be nice to have a variable to avoid the magic number

bdw617: yes, I'm not fixing that right now.

src/test/e2e/12_remote_pull_fail_test.go
```diff
@@ -0,0 +1,41 @@
+package test // replace with your actual package name
```
Racer159, line 1:
```suggestion
// SPDX-License-Identifier: Apache-2.0
// SPDX-FileCopyrightText: 2021-Present The Zarf Authors

// Package test provides e2e tests for Zarf.
package test
```

src/test/e2e/12_remote_pull_fail_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestBadRemotePackages(t *testing.T) {
+	t.Log("E2E: test bad remote packages")
+
+	zarfYaml := `
```
Racer159, line 14: e2e tests should use a zarf.yaml on disk under `src/test/packages`

src/test/e2e/12_remote_pull_fail_test.go
```diff
@@ -0,0 +1,41 @@
...
+  name: doesnotexist
+components:
+  - name: doesnotexist-docker
+    required: true
+    images:
+      - localhost/doesnotexist:6.13371337
+`
+	t.Run("zarf package create bad images", func(t *testing.T) {
+		// Create a temporary directory
+		tmpDir, err := os.MkdirTemp("", "test")
```
Racer159, line 26: We usually use `t.TempDir()` instead

src/test/e2e/12_remote_pull_fail_test.go
```diff
@@ -0,0 +1,41 @@
...
+		require.NoError(t, err)
+		defer os.RemoveAll(tmpDir) // Clean up
+
+		// Create zarf.yaml in the temporary directory
+		yamlPath := filepath.Join(tmpDir, "zarf.yaml")
+		require.NoError(t, os.WriteFile(yamlPath, []byte(zarfYaml), 0600))
```
Racer159, line 32: Not needed if already on disk

src/internal/packager/images/pull.go
```diff
@@ -448,40 +486,13 @@ func (i *ImageConfig) PullImage(src string, spinner *message.Spinner) (img v1.Im
...
-		if rawImg.Size > 750*1000*1000 {
-			warn := pterm.DefaultParagraph.WithMaxWidth(message.TermWidth).Sprintf("%s is %s and may take a very long time to load via docker. "+
-				"See https://docs.zarf.dev/docs/faq for suggestions on how to improve large local image loading operations.",
-				src, utils.ByteFormat(float64(rawImg.Size), 2))
-			spinner.Warnf(warn)
+		var localErr error
+		img, localErr = i.pullLocalImage(src, spinner)
+		if localErr != nil {
+			// if local failed, return the original remote error
+			return nil, false, err
```
Racer159, line 493: This error could use massaging now that we are showing it to all users if we can - at least to also say that it could not be found by docker since we actually don't know which error the user was expecting to see (it's most likely it would be from a remote but the error may confuse them if it were the opposite)

src/internal/packager/images/pull.go
```diff
@@ -434,6 +434,44 @@ func (i *ImageConfig) PullAll() ([]ImgInfo, error) {
 	return imgInfoList, nil
 }
 
+// pullLocalImage returns a v1.Image will try to return the local tarball, this way we can isolate remote issues from non-existent local images.
```
Racer159, line 437:
```suggestion
// pullLocalImage returns a v1.Image and will try to return the local tarball, this allows us to isolate remote issues from local issues.
```

Racer159 (changes requested):

src/internal/packager/images/pull.go
```diff
@@ -448,39 +486,10 @@ func (i *ImageConfig) PullImage(src string, spinner *message.Spinner) (img v1.Im
...
-		}
-
-		// Use unbuffered opener to avoid OOM Kill issues https://github.com/defenseunicorns/zarf/issues/1214.
-		// This will also take for ever to load large images.
-		if img, err = daemon.Image(reference, daemon.WithUnbufferedOpener()); err != nil {
-			return nil, false, fmt.Errorf("failed to load image %s from docker daemon: %w", src, err)
+		var localErr error
+		img, localErr = i.pullLocalImage(src, spinner)
+		if localErr != nil {
+			return nil, false, fmt.Errorf("remote repository: %w, local filesystem: %w", err, localErr)
```
Racer159, line 492: One displaying both errors I think it would be better to handle this in the fallback logic as two separate messages and follow the git fallback pattern more closely - the retry logic may also be something to look into if we want to show simpler messages in between (debugging the full error) with the last error(s) showing both.

Git Fallback:
image

Current implementation:
image

Racer159 (changes requested):

src/internal/packager/images/pull.go
Racer159: Simplified messaging and broke up lines visually
image

src/internal/packager/images/pull.go
```diff
@@ -446,13 +445,12 @@ func (i *ImageConfig) PullImage(src string, spinner *message.Spinner) (img v1.Im
...
 		// If crane is unable to pull the image, try to load it from the local docker daemon.
-		message.Debugf("crane unable to pull image %s: %s", src, err)
-		spinner.Updatef("Falling back to docker for %s. This may take some time.", src)
+		message.Notef("Falling back to local 'docker' images, failed to find the manifest on a remote: %s", err.Error())
 
 		// Parse the image reference to get the image name.
 		reference, err := name.ParseReference(src)
 		if err != nil {
-			return nil, false, fmt.Errorf("failed to parse image reference %s: %w", src, err)
+			return nil, false, fmt.Errorf("failed to parse image reference: %w", err)
```
AustinAbro321, line 453: Why not give the user the name of the image here?

Racer159: It is given in the parent error `failed to pull <image-ref>` and would stutter unnecessarily here

Noxsios (approved):

PR #2162, opened 2023-11-27
Move cluster library from `internal` to `pkg`

Noxsios:
## Description

Moves the `cluster` library (a wrapper around the k8s Go lib w/ Zarf specific helper functions) from `internal` to `pkg` so it can be leveraged as a public library for other projects such as Zarf UI and UDS CLI.

## Type of change

## Checklist before merging


Racer159 (approved):
lgtm

Issue #2064, opened 2023-10-11
Introduce a `zarf prepare lint` command


Racer159:
### Is your feature request related to a problem? Please describe.
We should introduce a `zarf prepare lint` command that will look at a `zarf.yaml` file and then error or warn when a `zarf.yaml` is out of spec with best practices.

### Describe the solution you'd like

- **Given** I have a `zarf.yaml` that is not following best practices
- **When** I run `zarf prepare lint` on that file
- **Then** I receive a list of notes, warnings or errors according to the following:

* [x] **error:** invalid package schema
* [ ] **warning:** flagged package warning (i.e. from p.warnings / deprecations)
* [x] **warning:** unpinned dependencies (`repos` or `images` without a tagged reference)
* [x] **warning:** `files` without pinned `shasums`
* [ ] **note:** `variables`, packages, `actions` and `components` without `description`s

### Describe alternatives you've considered
We could publish these as docs, but having them coded into a command allows them to be run in CI and updated without people needing to go back and read a docs page.

### Additional context
This will help package development scale and push the Zarf community's view of what makes a "good" Zarf package.

Racer159:
It may be worth it to yamllint downstream manifests and values files that are directly provided to Zarf as well https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1697579246332669?thread_ts=1683149646.315369&cid=C03B6BJAUJ3

AustinAbro321:
I would also like to give the user the line number in the future

AustinAbro321:
For checking unpinned repos I'm leaning towards just checking if "@" or "/-/" exists in the string. Could result in some false negatives since repos can have "@" in the name, but I'm thinking it'll work enough for what we're doing. This should cover github, dev.azure, and gitlab repos. Any source control sites popular among our users that I should take a look at @Racer159

AustinAbro321:
I'm thinking for images we warn when just set to latest / nothing but don't warn if set to a tag with a shasum like 1.0.0

AustinAbro321:
We may want either run every combination of architecture or pass in architectures from lint

AustinAbro321:
We should warn if variables are not being used anywhere

AustinAbro321:
We can also use variable defaults / set values we can get to make sure that the downstream mainfests / helmcharts are valid yaml

AustinAbro321:
[...]

PR #2171, opened 2023-11-30
feat: unpinned resources (`images`/`repos`/`files`) warning on `zarf prepare lint`

AustinAbro321:
## Description
Adding warnings for unpinned images, files and repos

## Related Issue

Relates to #2064 

## Type of change

## Checklist before merging


src/pkg/packager/lint/lint.go
```diff
@@ -50,6 +53,61 @@ func ValidateZarfSchema(path string) (*Validator, error) {
 	return &validator, nil
 }
 
+func isPinnedImage(image string) bool {
+	if strings.HasSuffix(image, ":latest") {
+		return false
+	}
```
Racer159, line 59: I think this is redundant - we should look for digests only (since tags are never _really_ pinned)

src/pkg/packager/lint/lint.go
```diff
@@ -50,6 +53,61 @@ func ValidateZarfSchema(path string) (*Validator, error) {
...
+				validator.addWarning(fmt.Sprintf(".components.[%d].images.[%d]: Unpinned image", i, j))
+			}
+		}
+	}
+}
+
+func checkForUnpinnedFiles(validator *Validator) {
+	for i, component := range validator.typedZarfPackage.Components {
+		for j, file := range component.Files {
+			if file.Shasum == "" {
```
Racer159, line 104: (already talked about this just a reminder to do isURL since we really only want to warn for remote sources)

Racer159 (changes requested):
just a few more notes

lint_all_zarf_packages.sh
lucasrod16, line 1: Are we using this somewhere? Did you intend to check this into git?

src/pkg/packager/composer/list.go
```diff
@@ -51,9 +73,18 @@ type ImportChain struct {
 	remote *oci.OrasRemote
 }
 
-func (ic *ImportChain) append(c types.ZarfComponent, relativeToHead string, vars []types.ZarfPackageVariable, consts []types.ZarfPackageConstant) {
+func (ic *ImportChain) Head() *Node {
```
lucasrod16, line 76: lint

src/pkg/packager/composer/list.go
```diff
@@ -51,9 +73,18 @@ type ImportChain struct {
...
+	return ic.head
+}
+
+func (ic *ImportChain) Tail() *Node {
```
lucasrod16, line 80: lint

src/pkg/packager/composer/list.go
```diff
@@ -72,14 +103,14 @@ func (ic *ImportChain) append(c types.ZarfComponent, relativeToHead string, vars
 }
 
 // NewImportChain creates a new import chain from a component
-func NewImportChain(head types.ZarfComponent, arch, flavor string) (*ImportChain, error) {
+func NewImportChain(head types.ZarfComponent, index int, arch, flavor string) (*ImportChain, error) {
```
lucasrod16, line 106: Anywhere we return an error in this function, we should also return a `nil` *ImportChain pointer

Lines:

- 130
- 135
- 139
- 153
- 160
- 165
- 172
- 189
- 191
- 195
- 197

lucasrod16, line 106: Why are we adding an index to track here? My understanding for part of why a double linked list was chosen for composition was so that we don't have to worry about keeping track of an index. 

cc: @Noxsios @Racer159

lint_all_zarf_packages.sh
AustinAbro321: Not sure exactly where it's used but it changed from direct to indirect when I did a `go mod tidy`

lucasrod16: I was referring to the `lint_all_zarf_packages.sh` script

AustinAbro321: Ah whoops, yeah I put it in on purpose though I can see the argument against putting it in source control. It's a nice way to recursively lint all the zarf packages we have which can be nice

AustinAbro321: Ah yeah I like that. That would make more sense

src/pkg/packager/composer/list.go
lucasrod16: Could we use the component name in the output instead of an index? Might be easier to understand

Noxsios: @lucasrod16 This `index` is not the index in the import chain, but rather the component's index in it's source `zarf.yaml`

lucasrod16: Ah I see, thanks @Noxsios!

AustinAbro321: @lucasrod16 The issue with having the name instead of the index is that two components might have the same name. We see this in our k3s package where we have an arm / amd resulting in two "k3s" components. This also allows the path to be evaluated by yq. Though I could see it being confusing to users when we get into composability. We should do some user testing on this at some point to see how intuitive it is to understand.

lucasrod16 (changes requested):

src/test/packages/12-lint/zarf.yaml
```diff
@@ -13,12 +15,37 @@ components:
 
   - name: import-test
     import:
-      path: 123123
+      path: linted-import
+    files:
+    # Is this a scenario we care about. The downstream overrides with a shasum.
```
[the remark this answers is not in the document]
AustinAbro321: That when I import another zarf package the contents of the component that gets imported is also linted. The linted-imoprt zarf.yaml has an image `registry.com:9001/whatever/image:latest` for example, that we want to warn on

src/pkg/packager/variables.go
```diff
@@ -15,6 +15,28 @@ import (
...
+		return err
+	}
+	return nil
+}
+
+// FindComponentTemplatesAndReload appends ###ZARF_COMPONENT_NAME###  for each component, assigns value, and reloads
+func FindComponentTemplatesAndReload(zarfPackage *types.ZarfPackage) error {
+	// iterate through components to and find all ###ZARF_COMPONENT_NAME, assign to component Name and value
+	for i := range zarfPackage.Components {
+		ReloadComponentTemplate(&zarfPackage.Components[i])
```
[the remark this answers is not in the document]
AustinAbro321: good catch

Noxsios (changes requested):

src/pkg/packager/composer/list.go
```diff
@@ -119,48 +152,55 @@ func NewImportChain(head types.ZarfComponent, arch, flavor string) (*ImportChain
...
-		found := helpers.Filter(pkg.Components, func(c types.ZarfComponent) bool {
-			matchesName := c.Name == name
-			return matchesName && CompatibleComponent(c, arch, flavor)
-		})
+		found := []types.ZarfComponent{}
+		var index int
+		for i, component := range pkg.Components {
+			if component.Name == name && CompatibleComponent(component, arch, flavor) {
+				found = append(found, component)
+				index = i
```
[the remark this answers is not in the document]
AustinAbro321: Well actually I wouldn't be able to verify it in e2e more so than it already is. I could  use mocks, though I'm unsure if we want to start doing that

src/pkg/packager/lint/validator.go
```diff
@@ -7,62 +7,175 @@ package lint
...
+		for _, finding := range v.findings {
+			if finding.packageKey == packageKey {
+				connectData[packageKey] = append(connectData[packageKey],
+					[]string{finding.validationType.String(), finding.getPath(), finding.String()})
+			}
 		}
-		for _, err := range v.errors {
-			connectData = append(connectData, []string{utils.ColorWrap("Error", color.FgRed), err.Error()})
+		//We should probably move this println into info
+		pterm.Println()
```
AustinAbro321, line 110: Thoughts on adding `pterm.Println()` to `message.info()` @Racer159 ? I know you mentioned you wanted to add new lines on messages generally, not sure if we hadn't gotten around to adding it to info or are not adding it for a reason

src/pkg/packager/lint/lint.go
```diff
@@ -50,16 +68,222 @@ func ValidateZarfSchema(path string) (*Validator, error) {
...
+			description: fmt.Sprintf("unable to find components %s", err),
+			packageKey:  packageKey{path: node.GetRelativeToHead(), name: node.GetOriginalPackageName()},
+		})
+	}
+	fillYamlTemplate(validator, node, createOpts, pkgKey)
+}
+
+func fillActiveTemplate(validator *Validator, createOpts types.ZarfCreateOptions) {
+
+	err := packager.FindComponentTemplatesAndReload(&validator.typedZarfPackage)
```
Racer159, line 131:
```suggestion
	err := packager.ReloadComponentTemplatesInPackage(&validator.typedZarfPackage)
```

src/pkg/packager/lint/lint.go
```diff
@@ -50,16 +68,222 @@ func ValidateZarfSchema(path string) (*Validator, error) {
...
+}
+
+func fillYamlTemplate(validator *Validator, yamlObj any, createOpts types.ZarfCreateOptions, pkgKey packageKey) {
+	templateMap := map[string]string{}
+
+	setVarsAndWarn := func(templatePrefix string, deprecated bool) {
+		yamlTemplates, err := utils.FindYamlTemplates(yamlObj, templatePrefix, "###")
+		if err != nil {
+			validator.addWarning(validatorMessage{
+				description: fmt.Sprintf("unable to find variables %s", err),
```
Racer159, line 150: Might not make sense to people if they see this

src/pkg/packager/lint/lint_test.go
```diff
@@ -110,4 +149,48 @@ func TestValidateSchema(t *testing.T) {
...
+				expected: false,
+				err:      nil,
+			},
+			{
+				input:    "ghcr.io/defenseunicorns/pepr/controller:v0.15.0",
+				expected: false,
+				err:      nil,
+			},
+			{
+				input:    "busybox:latest@sha256:3fbc632167424a6d997e74f52b878d7cc478225cffac6bc977eedfe51c7f4e79",
```
Racer159, line 171: We should add this test case to `ParseImageRef`'s test as well

src/pkg/packager/lint/lint_test.go
```diff
@@ -71,31 +66,75 @@ func TestValidateSchema(t *testing.T) {
...
+	})
+
+	t.Run("Unpinnned file warning", func(t *testing.T) {
+		validator := Validator{}
+		filename := "http://example.com/file.zip"
+		zarfFiles := []types.ZarfFile{
+			{
+				Source: filename,
+			},
+		}
```
Racer159, line 133: We should probably have some control options so that we make sure we only have one finding (since a pin and a local shouldn't have findings)

src/pkg/packager/lint/validator.go
Racer159: Could we use note?

src/pkg/packager/lint/validator.go
```diff
@@ -7,62 +7,172 @@ package lint
...
-	if v.hasWarnings() || v.hasErrors() {
-		header := []string{"Type", "Message"}
-		connectData := [][]string{}
-		for _, warning := range v.warnings {
-			connectData = append(connectData, []string{utils.ColorWrap("Warning", color.FgYellow), warning})
+	if !v.hasFindings() {
+		return
+	}
+
+	packageKeys := helpers.Unique(v.getUniquePackageKeys())
```
Racer159, line 95: Unique of unique

src/pkg/packager/lint/validator.go
```diff
@@ -7,62 +7,172 @@ package lint
...
 		}
-		for _, err := range v.errors {
-			connectData = append(connectData, []string{utils.ColorWrap("Error", color.FgRed), err.Error()})
+
+		message.Table(header, connectData[packageKey])
+		message.Info(v.getFormattedFindingCount(packageKey))
+	}
+}
+
+func (v Validator) getUniquePackageKeys() []packageKey {
```
Racer159, line 119: helpers.Unique

src/pkg/utils/random.go
```diff
@@ -42,6 +43,10 @@ func First30last30(s string) string {
 
 // ColorWrap changes a string to an ansi color code and appends the default color to the end
 // preventing future characters from taking on the given color
+// returns string as normal if color is disabled
 func ColorWrap(str string, attr color.Attribute) string {
+	if config.NoColor {
```
Racer159, line 48: First30Last30 and ColorWrap should probably be in `message instead` - `random` isn't junk drawer random - also message could be refactored out of RandomString and `random.go` could be moved to `helpers`

src/test/e2e/12_lint_test.go
```diff
@@ -1,31 +1,49 @@
...
+		// Multiple spaces in lang
+		require.Contains(t, strippedStderr, fmt.Sprintf(lang.PkgValidateTemplateDeprecation, key, key, key))
+		require.Contains(t, strippedStderr, ".components.[2].repos.[0] | Unpinned repository")
+		require.Contains(t, strippedStderr, ".metadata | Additional property description1 is not allowed")
+		require.Contains(t, strippedStderr, ".components.[0].import | Additional property not-path is not allowed")
+		require.Contains(t, strippedStderr, ".components.[2].images.[3] | Image not pinned with digest")
+		require.Contains(t, strippedStderr, ".components.[1].images.[0] | Image not pinned with digest")
+		require.Contains(t, strippedStderr, ".components.[1].images.[2] | Image not pinned with digest")
+		require.Contains(t, strippedStderr, ".components.[3].import.path | Zarf does not evaluate variables at component.x.import.path - ###ZARF_PKG_TMPL_PATH###")
+		require.Contains(t, strippedStderr, ".components.[0].images.[0] | Image not pinned with digest - defenseunicorns/zarf-game:multi-tile-dark")
```
Racer159, line 46: Comments for the test cases would be nice

src/test/packages/12-lint/###ZARF_PKG_TMPL_PATH###/zarf.yaml
```diff
@@ -0,0 +1,10 @@
+kind: ZarfInitConfig
+metadata:
+  name: init
+  description: Testing bad yaml imported
```
Racer159, line 4: init!

src/test/packages/12-lint/linted-import/zarf.yaml
```diff
@@ -0,0 +1,23 @@
+kind: ZarfInitConfig
+metadata:
+  name: init
```
Racer159, line 3: init!

src/test/packages/12-lint/###ZARF_PKG_TMPL_PATH###/zarf.yaml
Racer159, line 1: May be better to just check that import chain doesn't work ###ZARF_PKG_TMPL_*### is a valid dir but unlikely

src/test/packages/12-lint/linted-import/zarf.yaml
```diff
@@ -0,0 +1,23 @@
...
+  description: Testing bad yaml imported
+
+variables:
+  - name: BUSYBOX_IMAGE
+    description: "whatever"
```
Racer159, line 8: This isn't needed

src/test/packages/12-lint/zarf.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
 
 
 variables:
+  - name: BUSYBOX_IMAGE
+    description: "whatever"
```
Racer159, line 9: Not needed

Racer159 (changes requested):

src/pkg/packager/lint/lint_test.go
AustinAbro321: Ahhhh now I get what you mean, I hadn't realized that was a function that we were maintaining

src/cmd/common/common.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"github.com/defenseunicorns/zarf/src/types"
+)
+
+// SetBaseDirectory sets base directory on package config when given in args
+func SetBaseDirectory(args []string, pkgConfig *types.PackagerConfig) {
+	if len(args) > 0 {
+		pkgConfig.CreateOpts.BaseDir = args[0]
+	} else {
+		pkgConfig.CreateOpts.BaseDir = "."
```
AustinAbro321, line 16: The reason I changed it to "." instead of os.Getwd() is because `zarf prepare lint` without a path would print out the full file system path. We could also argue that's something we want, I can see both ways. AFAIK this won't affect package create.

AustinAbro321: Discussed with team, we feel this is fine

src/pkg/packager/lint/lint.go
```diff
@@ -24,22 +31,29 @@ func getSchemaFile() ([]byte, error) {
 	return ZarfSchema.ReadFile("zarf.schema.json")
 }
 
-// ValidateZarfSchema validates a zarf file against the zarf schema, returns *validator with warnings or errors if they exist
+// ValidateZarf validates a zarf file against the zarf schema, returns *validator with warnings or errors if they exist
 // along with an error if the validation itself failed
-func ValidateZarfSchema(path string) (*Validator, error) {
+func ValidateZarf(createOpts types.ZarfCreateOptions) (*Validator, error) {
```
Racer159, line 36:
```suggestion
func Validate(createOpts types.ZarfCreateOptions) (*Validator, error) {
```

Could just be simplified to Validate

src/pkg/packager/lint/lint.go
```diff
@@ -24,22 +31,29 @@ func getSchemaFile() ([]byte, error) {
...
 	}
 
-	if err := utils.ReadYaml(filepath.Join(path, layout.ZarfYAML), &validator.untypedZarfPackage); err != nil {
+	if err := os.Chdir(createOpts.BaseDir); err != nil {
+		return nil, fmt.Errorf("unable to access directory '%s': %w", createOpts.BaseDir, err)
+	}
+
+	validator.baseDir = createOpts.BaseDir
+
+	lintComposableComponents(&validator, &createOpts)
```
Racer159, line 54:
```suggestion
	lintComponents(&validator, &createOpts)
```

This should probably just be components since it does all of them regardless of composition

src/pkg/cluster/state.go
Noxsios, line 1: github PR review is being dumb, so i'lll make a comment on the file as a whole:

In zarf we have taken to not doing `message.Fatal` anywhere except for top level functions in `cmd` because it will do an `os.Exit`

This is undesirable from a library perspective because it does not allow for library users to handle errors as they see fit.

src/pkg/packager/lint/lint.go
```diff
@@ -50,16 +64,194 @@ func ValidateZarfSchema(path string) (*Validator, error) {
...
+func checkForUnpinnedImages(validator *Validator, node *composer.Node) {
+	for j, image := range node.Images {
+		imageYqPath := fmt.Sprintf(".components.[%d].images.[%d]", node.GetIndex(), j)
+		pinnedImage, err := isPinnedImage(image)
+		if err != nil {
+			validator.addError(validatorMessage{
+				yqPath:         imageYqPath,
+				packageRelPath: node.GetRelativeToHead(),
+				packageName:    node.GetOriginalPackageName(),
+				description:    "Invalid image format",
```
Racer159, line 204:
```suggestion
				description:    "Invalid image reference",
```
Format might imply something like its format on disk

src/pkg/packager/composer/list.go
```diff
@@ -73,25 +78,27 @@ type ImportChain struct {
 	remote *oci.OrasRemote
 }
 
-// Head returns the first node in the import chain
+// Head eturns the first node in the import chain
```
Noxsios, line 81:
```suggestion
// Head returns the first node in the import chain
```

src/pkg/packager/lint/lint_test.go
```diff
@@ -71,31 +66,90 @@ func TestValidateSchema(t *testing.T) {
...
+		checkForUnpinnedImages(&validator, &composer.Node{ZarfComponent: component})
+		require.Equal(t, unpinnedImage, validator.findings[0].item)
+		require.Equal(t, badImage, validator.findings[1].item)
+		require.Equal(t, 2, len(validator.findings))
+
+	})
+
+	t.Run("Unpinnned file warning", func(t *testing.T) {
+		validator := Validator{}
+		fileUrl := "http://example.com/file.zip"
```
Racer159, line 135: lint

src/pkg/packager/lint/lint_test.go
```diff
@@ -110,4 +164,67 @@ func TestValidateSchema(t *testing.T) {
 		acutal := makeFieldPathYqCompat(input)
 		require.Equal(t, input, acutal)
 	})
+
+	t.Run("Test comoposable components", func(t *testing.T) {
```
Racer159, line 168:
```suggestion
	t.Run("Test composable components", func(t *testing.T) {
```

src/pkg/packager/composer/list.go
```diff
@@ -129,78 +136,69 @@ func NewImportChain(head types.ZarfComponent, index int, arch, flavor string) (*
...
-				return nil, err
+				return ic, err
 			}
 		}
 
 		name := node.ImportName()
 
 		found := []types.ZarfComponent{}
-		var index int
+		index := []int{}
```
Noxsios, line 187: I would leave a note/comment on the relationship between `found` and `index`

src/pkg/packager/composer/list.go
```diff
@@ -129,78 +136,69 @@ func NewImportChain(head types.ZarfComponent, index int, arch, flavor string) (*
...
-			}
+			return ic, fmt.Errorf("component %q not found in %q", name, relativeToHead)
 		} else if len(found) > 1 {
-			if isLocal {
-				return nil, fmt.Errorf("multiple components named %q found in %q satisfying %q", name, filepath.Join(history...), arch)
-			} else if isRemote {
-				return nil, fmt.Errorf("multiple components named %q found in %q satisfying %q", name, node.Import.URL, arch)
-			}
+			return ic, fmt.Errorf("multiple components named %q found in %q satisfying %q", name, relativeToHead, arch)
 		}
```
Noxsios, line 199: why were these `is*` checks removed?

src/test/e2e/12_lint_test.go
```diff
@@ -1,31 +1,61 @@
...
+	})
+
 	t.Run("zarf test lint fail", func(t *testing.T) {
 		t.Log("E2E: Test lint on schema fail")
 
 		path := filepath.Join("src", "test", "packages", "12-lint")
-		_, stderr, err := e2e.Zarf("prepare", "lint", path)
+		configPath := filepath.Join(path, "zarf-config.toml")
+		os.Setenv("ZARF_CONFIG", configPath)
+		// In this case I'm guessing we should also remove color from the table?
```
Racer159, line 30: Odd guessing comment

src/test/e2e/12_lint_test.go
```diff
@@ -1,31 +1,61 @@
...
+		require.Contains(t, strippedStderr, ".metadata | Additional property description1 is not allowed")
+		require.Contains(t, strippedStderr, ".components.[0].import | Additional property not-path is not allowed")
+		// This is testing the import / compose on lint is working
+		require.Contains(t, strippedStderr, ".components.[1].images.[0] | Image not pinned with digest - registry.com:9001/whatever/image:latest")
+		// This is testing import / compose + variables are working
+		require.Contains(t, strippedStderr, ".components.[2].images.[3] | Image not pinned with digest - busybox:latest")
+		require.Contains(t, strippedStderr, ".components.[3].import.path | Zarf does not evaluate variables at component.x.import.path - ###ZARF_PKG_TMPL_PATH###")
+		// testing OCI imports get linted
+		require.Contains(t, strippedStderr, ".components.[0].images.[0] | Image not pinned with digest - defenseunicorns/zarf-game:multi-tile-dark")
+		// This is
```
Racer159, line 48: This is...

src/test/packages/12-lint/zarf.yaml
```diff
@@ -1,24 +1,60 @@
...
+      name: baseline
+
+  - name: oci-games-url
+    import:
+      path: linted-import
+
+  # - name: import-big-bang
+  #   import:
+  #     url: oci://localhost:555/big-bang-min:2.10.0-skeleton
+
```
Racer159, line 49: Commented out YAML?

Racer159 (changes requested):

src/pkg/packager/lint/lint_test.go
```diff
@@ -117,25 +123,33 @@ func TestValidateSchema(t *testing.T) {
...
+		require.Equal(t, unpinnedImage, validator.findings[0].item)
+		require.Equal(t, badImage, validator.findings[1].item)
+		require.Equal(t, 2, len(validator.findings))
+
 	})
 
 	t.Run("Unpinnned file warning", func(t *testing.T) {
 		validator := Validator{}
-		filename := "http://example.com/file.zip"
+		fileUrl := "http://example.com/file.zip"
```
Noxsios, line 135: lint

src/pkg/packager/lint/validator.go
```diff
@@ -6,98 +6,146 @@ package lint
...
-	yqPath      string
-	filePath    string
-	description string
-	item        string
+	yqPath         string
+	description    string
+	item           string
+	packageRelPath string
+	packageName    string
+	validationType validationType
```
Noxsios, line 30: what about `type validation int`

so you can do

```go
type validatorMessage struct {
   ...
   Type validation
 }
 ```
 
 @Racer159 what's your take on stuttering w/ `type xyzType`

File: src/test/packages/12-lint/zarf.yaml [hunk 36]
```diff
@@ -49,3 +38,23 @@ components:
...
       name: baseline
+
+  - name: oci-games-url
+    import:
+      path: linted-import
+
+  # - name: import-big-bang
+  #   import:
+  #     url: oci://localhost:555/big-bang-min:2.10.0-skeleton
+
```
Noxsios, line 49: remove dead code

Noxsios (changes requested):

src/cmd/common/common.go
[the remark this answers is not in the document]
AustinAbro321: Added it to this file after discussion with @Racer159 because I'm using the function in both package.go and prepare.go. Maybe it would be better if I called the file something like common_functions.go?

src/pkg/cluster/state.go
```diff
@@ -277,10 +283,14 @@ func (c *Cluster) MergeZarfState(oldState *types.ZarfState, initOptions types.Za
...
+			if newState.GitServer.PushPassword, err = helpers.RandomString(config.ZarfGeneratedPasswordLen); err != nil {
+				message.Fatal(err, lang.ErrUnableToGenerateRandomSecret)
+			}
 		}
 		if newState.GitServer.PullPassword == oldState.GitServer.PullPassword && oldState.GitServer.InternalServer {
-			newState.GitServer.PullPassword = utils.RandomString(config.ZarfGeneratedPasswordLen)
+			if newState.GitServer.PullPassword, err = helpers.RandomString(config.ZarfGeneratedPasswordLen); err != nil {
+				message.Fatal(err, lang.ErrUnableToGenerateRandomSecret)
+			}
 		}
```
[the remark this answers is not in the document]
AustinAbro321: Yeah I set it to fatal as I was moving the `message.fatal` out of utils.randomstring. I like your explanation in thinking about it as a library and handing the fatal in the cmd package. Will change it to that

src/pkg/packager/composer/list.go
AustinAbro321: I ended up setting relativeToHead on oci components to be node.Import.URL of the previous node. By doing that this block of code becomes the same in both paths. Additionally it makes the logic in lint simpler as I no longer have to backtrack on the list to get the location of the component.

Though I guess that makes relativeToHead a misnomer. Maybe `relativePathOrUrl`

src/pkg/cluster/state.go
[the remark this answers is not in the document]
AustinAbro321: Made that change. One thing I wasn't sure about was the following function calls in the same file. I could see either copying the initOptions.GitServer to state.GitServer and sending in a pointer or just sending in and receiving a value as I did here. Functionally I believe it's exactly the same not sure if we prefer the style of one over the other. @Racer159 @Noxsios 

```
		if state.GitServer, err = c.fillInEmptyGitServerValues(initOptions.GitServer); err != nil {
			return err
		}
		if state.RegistryInfo, err = c.fillInEmptyContainerRegistryValues(initOptions.RegistryInfo); err != nil {
			return err
		}
```

Makefile
```diff
@@ -76,6 +76,9 @@ docs-and-schema: ## Generate the Zarf Documentation and Schema
 	hack/gen-cli-docs.sh
 	ZARF_CONFIG=hack/empty-config.toml hack/create-zarf-schema.sh
 
+lint-packages-and-examples: build-cli ## Recurisvely lint all zarf.yaml files in the repo except for those dedicated to tests
```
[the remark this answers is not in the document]
AustinAbro321: Added a build-cli-for-system make command, and rely on that. Though it will only every create windows amd, linux amd, mac-intel, mac-apple since those are the only binaries we set zarf_bin atm

src/pkg/packager/lint/validator.go
AustinAbro321: After talking with @Racer159 we ended up changing it to category

src/pkg/cluster/state.go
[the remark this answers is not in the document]
AustinAbro321: After talking with Wayne we decided both styles were reasonable but in this case since we preferred not sending it as a pointer since it's only going to one function and it'll keep it less side-affecty

src/cmd/common/common.go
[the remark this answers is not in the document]
AustinAbro321: Ended up calling it utils.go

src/cmd/common/utils.go
```diff
@@ -0,0 +1,18 @@
...
+
+// Package common handles command configuration across all commands
+package common
+
+import (
+	"github.com/defenseunicorns/zarf/src/types"
+)
+
+// SetBaseDirectory sets base directory on package config when given in args
+func SetBaseDirectory(args []string, pkgConfig *types.PackagerConfig) {
```
Racer159, line 12: This will have to be used in `zarf dev deploy` when `main` is merged into this branch

Racer159 (changes requested):
tiny comment for when merge happens

src/config/lang/english.go
```diff
@@ -365,8 +371,8 @@ $ zarf package publish ./path/to/dir oci://my-registry.com/my-namespace
 	CmdPrepareFlagGitAccount    = "User or organization name for the git account that the repos are created under."
 	CmdPrepareFlagKubeVersion   = "Override the default helm template KubeVersion when performing a package chart template"
 
-	CmdPrepareLintShort = "Verifies the package schema"
-	CmdPrepareLintLong  = "Verifies the package schema and warns the user if they have variables that won't be evaluated"
+	CmdPrepareLintShort = "Lints the given package for valid schema and recommended practices"
+	CmdPrepareLintLong  = "Verifies the package schema, checks if any variables won't be evaluated, and checks for unpinned images/repos/files"
```
Racer159, line 375: CmdPrepare should be CmdDev

src/pkg/packager/composer/list.go
```diff
@@ -22,17 +22,44 @@ import (
 type Node struct {
 	types.ZarfComponent
 
+	index int
+
 	vars   []types.ZarfPackageVariable
 	consts []types.ZarfPackageConstant
 
-	relativeToHead string
+	relativeToHeadOrURL string
```
Racer159, line 30: I think it could be good to remember the OCI url in a separate value just because a filepath.Join (the most common operation run on this value) would throw up if someone tried to use it with the URL.

Racer159: We could also have a simple helper function that gets it for you

Racer159 (changes requested):

src/cmd/common/common.go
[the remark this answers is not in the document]
Racer159: And my 2 pennies - I could go either way on where the function lives - felt common in the CLI was simpler for now, mostly would just be if we wanted library peeps to be able to pass in an empty string to the packager config which might be interesting, but also they should probably explicitly know which path they are running against.

src/pkg/packager/composer/list.go
AustinAbro321: That makes sense, ended up going that route

src/config/lang/english.go
```diff
@@ -371,9 +371,9 @@ $ zarf package publish ./path/to/dir oci://my-registry.com/my-namespace
 	CmdPrepareFlagGitAccount    = "User or organization name for the git account that the repos are created under."
 	CmdPrepareFlagKubeVersion   = "Override the default helm template KubeVersion when performing a package chart template"
```
Racer159: These should also be CmdDev

Makefile
```diff
@@ -85,10 +89,16 @@ build-cli-linux: build-cli-linux-amd build-cli-linux-arm ## Build the Zarf CLI f
...
 build-cli: build-cli-linux-amd build-cli-linux-arm build-cli-mac-intel build-cli-mac-apple build-cli-windows-amd build-cli-windows-arm ## Build the CLI
 
+build-cli-for-system:
+	$(MAKE) $(BUILD_CLI_FOR_SYSTEM)
+
 docs-and-schema: ## Generate the Zarf Documentation and Schema
 	hack/gen-cli-docs.sh
 	ZARF_CONFIG=hack/empty-config.toml hack/create-zarf-schema.sh
 
+lint-packages-and-examples: build-cli-for-system ## Recurisvely lint all zarf.yaml files in the repo except for those dedicated to tests
```
lucasrod16, line 99:
```suggestion
lint-packages-and-examples: build-cli-for-system ## Recursively lint all zarf.yaml files in the repo except for those dedicated to tests
```

src/pkg/packager/composer/list.go
```diff
@@ -44,9 +45,12 @@ func (n *Node) GetOriginalPackageName() string {
 	return n.originalPackageName
 }
 
-// GetRelativeToHeadOrURL gets the path from downstream zarf file to upstream imported zarf file
+// GetRelativeToHeadOrURL gets the path from the base zarf file to the imported zarf file
 func (n *Node) GetRelativeToHeadOrURL() string {
```
Racer159, line 49:
```suggestion
func (n *Node) ImportLocation() string {
```

Racer159 (changes requested):

Racer159 (dismissed):
lgtm

Makefile
```diff
@@ -85,10 +89,16 @@ build-cli-linux: build-cli-linux-amd build-cli-linux-arm ## Build the Zarf CLI f
 
 build-cli: build-cli-linux-amd build-cli-linux-arm build-cli-mac-intel build-cli-mac-apple build-cli-windows-amd build-cli-windows-arm ## Build the CLI
 
+build-cli-for-system:
+	$(MAKE) $(BUILD_CLI_FOR_SYSTEM)
```
Noxsios, line 93: needs `##` help string

lucasrod16 (dismissed):

hack/lint_all_zarf_packages.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+ZARF_BIN=$1
+LINT_SRC_TEST=$2
+SCRIPT=$(realpath "$0")
+SCRIPTPATH=$(dirname "$SCRIPT")
+cd $SCRIPTPATH
```
Noxsios, line 7: suggestions from `shellcheck`

```suggestion
cd "$SCRIPTPATH" || exit
```

hack/lint_all_zarf_packages.sh
```diff
@@ -0,0 +1,17 @@
...
+cd ..
+find "." -type f -name 'zarf.yaml' | while read -r yaml_file; do
+  dir=$(dirname "$yaml_file")
+  if [[ "$dir" == *src/test/* ]] && [ "$LINT_SRC_TEST" != true ]; then
+      continue
+  fi
+  echo "Running 'zarf prepare lint' in directory: $dir"
+  $ZARF_BIN prepare lint $dir
```
Noxsios, line 15: suggestion from `shellcheck`

```suggestion
  $ZARF_BIN prepare lint "$dir"
```

src/pkg/packager/composer/list.go
```diff
@@ -22,17 +22,48 @@ import (
...
+	return n.originalPackageName
+}
+
+// ImportLocation gets the path from the base zarf file to the imported zarf file
+func (n *Node) ImportLocation() string {
+	if n.importURL != "" {
+		return n.importURL
+	}
+	return n.relativeToHead
+}
```
Noxsios, line 54: what added benefit is keeping track of `n.importURL` vs just looking at `n.Import.URL`?

Noxsios (changes requested):

Noxsios (approved):

Issue #2062, opened 2023-10-09
zarf prepare find-images is not consistent between OCI and Helm chart registries


alinalex1392:
### Environment
Device and OS:
- NAME="Ubuntu"
- VERSION="20.04.5 LTS (Focal Fossa)"
App version: v0.30.0
Kubernetes distro being used: None
Other: CLI usage

### Steps to reproduce
1. Execute 

```sh
zarf prepare find-images -l debug --kube-version=1.23.0-0
```

```yaml
kind: ZarfPackageConfig
metadata:
  name: helm-charts
  description: Example showcasing mutiple ways to deploy helm charts
  version: 0.0.1
components:
  - name: demo-helm-oci-chart
    charts:
      - name: asdasd
        version: 6.4.0
        namespace: podinfo-from-oci
        url: oci://ghcr.io/stefanprodan/charts/podinfo
  - name: demo-helm-alt-release-name
    charts:
      - name: podinfo
        version: 6.4.0
        namespace: helm-alt-release-name
        url: https://stefanprodan.github.io/podinfo/podinfo
        releaseName: cool-name
```

The list of images is gathered correctly, even though the name from the demo-helm-oci-chart component is being wrong.

Later Edit:

If I change the component  demo-helm-oci-chart name to asdasd/podinfo then the command fail and the name is taken into consideration:

```yaml
kind: ZarfPackageConfig
metadata:
  name: helm-charts
  description: Example showcasing mutiple ways to deploy helm charts
  version: 0.0.1
components:
  - name: demo-helm-oci-chart
    charts:
      - name: asdasd/podinfo
        version: 6.4.0
        namespace: podinfo-from-oci
        url: oci://ghcr.io/stefanprodan/charts/podinfo
  - name: demo-helm-alt-release-name
    charts:
      - name: podinfo
        version: 6.4.0
        namespace: helm-alt-release-name
        url: https://stefanprodan.github.io/podinfo/podinfo
        releaseName: cool-name
```
Logs:

```text
  DEBUG   2023-10-10T12:39:42Z  -  Reading YAML at zarf.yaml
  DEBUG   2023-10-10T12:39:43Z  -  rename /tmp/zarf-1580237310/components/demo-helm-oci-chart/charts/podinfo-6.4.0.tgz /tmp/zarf-1580237310/components/demo-helm-oci-chart/charts/asdasd/podinfo-6.4.0.tgz: no such file or directory
     ERROR:  Unable to save the chart tarball
```

[...]

Issue #1765, opened 2023-06-01
Helm chart clobbering within components


Noxsios:
### Steps to reproduce

Given the following `zarf.yaml` in a directory side by side w/ the Zarf repo:

```yaml
kind: ZarfPackageConfig
metadata:
  name: scratch
  version: 0.0.1

components:
  - name: helm-charts-clobbering
    required: false
    charts:
      - name: podinfo
        localPath: ../zarf/examples/helm-charts/chart
        namespace: chart-1
        version: 6.3.5
      - name: oci-demo
        version: 6.3.5
        namespace: chart-2
        url: oci://ghcr.io/stefanprodan/charts/podinfo
    images:
      - ghcr.io/stefanprodan/podinfo:6.3.5
```

1. `zarf package create . --confirm`
2. `zarf tools archiver decompress zarf-package-scratch-amd64-0.0.1.tar.zst unpacked --unarchive-all`
3. `tree unpacked/components`

### Expected result

2 charts shown

### Actual Result

```
unpacked/components
└── helm-charts-clobbering
    └── charts
        └── oci-demo-6.3.5.tgz
```

Since the 2nd Helm chart is also `podinfo`, it clobbers the first one and destroys the expected `podinfo-6.3.5.tgz`. This results in a broken package as only the final `podinfo` chart will be saved.

### Severity/Priority

IDK, you _really_ shouldn't be using multiple Helm charts with the same _exact_ name and version within a single component (kinda an antipattern tbh), so not really sure where this falls. Opened the issue for tracking.

### Additional Context
Add any other context or screenshots about the technical debt here.

PR #2174, opened 2023-12-03
fix: resolve helm chart clobbering and confusion between Zarf and Helm `chart` names

Racer159:
## Description

This resolves issues where helm charts would have a potential to clobber each other and also decouples Zarf `charts.Name` from Helm chart names to increase consistency and reduce confusion.

## Related Issue

Fixes https://github.com/defenseunicorns/zarf/issues/2062
Fixes https://github.com/defenseunicorns/zarf/issues/1765

## Type of change

## Checklist before merging


src/internal/packager/helm/repo.go
```diff
@@ -174,8 +182,14 @@ func (h *Helm) DownloadPublishedChart(cosignKeyPath string) error {
 		},
 	}
 
-	// Download the file (we don't control what name helm creates here)
-	saved, _, err := chartDownloader.DownloadTo(chartURL, pull.Version, h.chartPath)
+	// Download the file into a temp directory since we don't control what name helm creates here
+	temp := filepath.Join(h.chartPath, "temp")
+	if err = utils.CreateDirectory(temp, 0700); err != nil {
+		return fmt.Errorf("unable to download the helm chart: %w", err)
```
Noxsios, line 188: this is an improper error message for this operation, `unable to create temp directory: %w` would be a bit better

Noxsios (changes requested):

src/internal/packager/helm/repo.go
Racer159: done

Noxsios (approved):

Issue #1824, opened 2023-06-14
Check for the existence of the correct storage class prior to registry / gitea initialization


Racer159:
### Is your feature request related to a problem? Please describe.
The registry / gitea packages require a storage class / pvc to operate and it would be nice to have them check for the existence of such a storage class before they try to run.

### Describe the solution you'd like

- **Given** I have a cluster configured without a storage class
- **When** I initialize Zarf with the registry or gitea
- **Then** I receive an error telling me that the storage class is misconfigured (instead of a permanent pending pod)

### Additional context
This would be the equivalent of the following action:

```
    actions:
      onDeploy:
        before:
          - description: Check that the cluster has the correct storage class
            maxRetries: 0
            cmd: |
              CLUSTER_STORAGE_CLASSES=$(./zarf tools kubectl get -A StorageClass -o jsonpath={$.items[*].metadata.name})
              if [ -n "${CLUSTER_STORAGE_CLASSES}" && -n "${CLUSTER_STORAGE_CLASSES##*${ZARF_STORAGE_CLASS}*}" ]; then echo "the Zarf init registry package requires a configured storage class to function"; fi
```

Though it may be worth considering a more expressive "assert" syntax that wraps these kinds of checks like `wait` does.

robmcelvenny:
I would like to work on this issue.

lucasrod16:
Reopening this issue. See https://github.com/defenseunicorns/zarf/pull/2311 for more context

PR #2180, opened 2023-12-06
feat: verify storage class exists on init

lucasrod16:
## Description
Verify storage class exists on init

## Related Issue

Fixes https://github.com/defenseunicorns/zarf/issues/1824

## Type of change

## Checklist before merging


packages/gitea/zarf.yaml
```diff
@@ -67,6 +67,17 @@ components:
           - gitea-values.yaml
     actions:
       onDeploy:
+        before:
+          - description: Check that the cluster has a storage class
+            shell:
+              windows: pwsh
```
Racer159, line 73: Pretty sure this won't work in `pwsh`
image

packages/gitea/zarf.yaml
```diff
@@ -67,6 +67,17 @@ components:
...
+            cmd: |
+              STORAGE_CLASSES=$(./zarf tools kubectl get StorageClass -o jsonpath='{.items[*].metadata.name}')
+              if [ -z "${STORAGE_CLASSES}" ]; then
```
Racer159, line 76: Do we want to also check if their selected storage class is available? (it should be available as `${ZARF_STORAGE_CLASS}`)

packages/gitea/zarf.yaml
Racer159: For this stuff we may want to rethink windows support and how we could implement it.  A few ways off the top that might be useful:

1. Mutate if statements in some Powershell friendly way
2. Implement a conditional action kind that can handle this more in Zarf-land

Racer159 (changes requested):
test looks good, just a few notes about compatibility

packages/gitea/zarf.yaml
lucasrod16: What do you think about doing this in the Zarf codebase instead of using shell commands in an action? My thinking is that we wouldn't have to do any tricks to make it work on Windows with that approach

packages/gitea/zarf.yaml
```diff
@@ -67,6 +67,13 @@ components:
           - gitea-values.yaml
     actions:
       onDeploy:
+        before:
+          - description: Check that the cluster has a storage class
+            maxTotalSeconds: 3
+            wait:
+              cluster:
+                kind: storageclass
+                name: "###ZARF_STORAGE_CLASS###" # TODO@lucasrod16: figure out why this doesn't work
```
Racer159, line 76:
```suggestion
                name: "${ZARF_STORAGE_CLASS}"
```
You need to use the actions syntax since in the end this will be pulled from the environment.

Racer159 (changes requested):

packages/zarf-registry/zarf.yaml
```diff
@@ -111,6 +111,15 @@ components:
...
       - "###ZARF_PKG_TMPL_REGISTRY_IMAGE_DOMAIN######ZARF_PKG_TMPL_REGISTRY_IMAGE###:###ZARF_PKG_TMPL_REGISTRY_IMAGE_TAG###"
+    actions:
+      onDeploy:
+        before:
+          - description: Check that the cluster has a storage class
+            maxTotalSeconds: 3
+            wait:
+              cluster:
+                kind: storageclass
+                name: "###ZARF_STORAGE_CLASS###" # TODO@lucasrod16: figure out why this doesn't work
```
Racer159, line 122:
```suggestion
                name: "${ZARF_STORAGE_CLASS}" # TODO@lucasrod16: figure out why this doesn't work
```

Racer159 (changes requested):

src/test/e2e/20_zarf_init_test.go
```diff
@@ -36,26 +38,23 @@ func TestZarfInit(t *testing.T) {
...
-	componentsFlag := ""
-	if e2e.ApplianceMode {
-		// make sure init fails in appliance mode when we try to initialize a k3s cluster
-		// with behavior from the k3s component's actions
-		componentsFlag = "--components=k3s"
+		// Check that `zarf init` returns an error because of the mismatched architectures.
+		// We need to use the --architecture flag here to force zarf to find the package.
+		_, stdErr, err = e2e.Zarf("init", "--architecture", mismatchedArch, "--components=k3s", "--confirm")
+		require.Error(t, err, stdErr)
+		require.Contains(t, stdErr, expectedErrorMessage)
```
Racer159, line 50: can we fix the expected error here?

Racer159 (changes requested):

src/test/e2e/20_zarf_init_test.go
lucasrod16: fixed

Racer159 (approved):
lgtm

Racer159 (approved):
lgtm

src/pkg/utils/wait.go
```diff
@@ -73,7 +75,7 @@ func ExecuteWait(waitTimeout, waitNamespace, condition, kind, identifier string,
 
 	// Setup the spinner messages.
 	conditionMsg := fmt.Sprintf("Waiting for %s%s%s to be %s.", kind, identifierMsg, namespaceMsg, condition)
-	existMsg := fmt.Sprintf("Waiting for %s%s%s to exist.", kind, identifierMsg, namespaceMsg)
+	existMsg := fmt.Sprintf("Waiting for %s%s to exist.", filepath.Join(kind, identifierMsg), namespaceMsg)
```
Noxsios, line 78: on windows this will cause this message to look like: `Waiting for pod\\hello to exist.`, this needs a `filepath.ToSlash` to prevent that

Noxsios (changes requested):

src/pkg/utils/wait.go
lucasrod16: updated

src/pkg/utils/wait.go
```diff
@@ -73,7 +75,7 @@ func ExecuteWait(waitTimeout, waitNamespace, condition, kind, identifier string,
 
 	// Setup the spinner messages.
 	conditionMsg := fmt.Sprintf("Waiting for %s%s%s to be %s.", kind, identifierMsg, namespaceMsg, condition)
-	existMsg := fmt.Sprintf("Waiting for %s%s%s to exist.", kind, identifierMsg, namespaceMsg)
+	existMsg := fmt.Sprintf("Waiting for %s%s to exist.", filepath.ToSlash(filepath.Join(kind, identifierMsg)), namespaceMsg)
```
Racer159, line 78:
```suggestion
	existMsg := fmt.Sprintf("Waiting for %s%s to exist.", path.Join(kind, identifierMsg), namespaceMsg)
```
path.Join (instead of filepath.Join) will always be a slash

src/pkg/utils/wait.go
```diff
@@ -93,16 +95,25 @@ func ExecuteWait(waitTimeout, waitNamespace, condition, kind, identifier string,
...
 				message.Debug(stdout, stderr, err)
 				continue
 			}
 
+			// If there was no identifier provided and no resources of the given kind were found,
+			// return an error. We don't want to return an error if an identifier
+			// was provided because the resource may be in the process of being created.
+			resourceNotFound := strings.Contains(stderr, "No resources found") && identifier == ""
+			if resourceNotFound {
+				return fmt.Errorf("unable to find resource of kind %q in the cluster", kind)
```
Racer159, line 109: this logic should be the same as 100-101 since it could still be in the process of being created and we should wait for it (if the wait times out then we should fail)

Racer159 (changes requested):
just a few small things

src/pkg/utils/wait.go
lucasrod16: updated

Racer159 (approved):
lgtm

Noxsios (approved):

PR #2184, opened 2023-12-11
feat: true multi arch OCI

Noxsios:
## Description

Changes OCI refs from `:<version>-<arch>` to `:<version>` and instead moves that logic more correctly into the `-a/--architecture` flag.

This PR also adds `flavor` to the pkg's build metadata so it can be appended to the ref: `:<version>-<flavor>` if it exists. This is because flavor is a higher level specifier than arch.

- Flavors `vanilla` and `chocolate` can have `amd64/arm64/etc...` arch variants
- No flavor (just `<version>`) can have `amd64/arm64/etc...` arch variants
- Zarf makes OS determinations at runtime, so all OS variants of a package are built and included, regardless of `flavor` or `arch`, therefore for the platform portion of the image index created: `multi` is used to denote that the package is not directly tied to any specific OS.

Example index:

```json
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:42105043580ce0a6f5cb519a57917056597449ffc42b5be12d49f7856f3747ab",
      "size": 2827,
      "platform": {
        "architecture": "arm64",
        "os": "multi"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:556ced6297ff9627b99e07d0e125df1f199be4c20d9caf6024b5c53303becc15",
      "size": 2827,
      "platform": {
        "architecture": "amd64",
        "os": "multi"
      }
    }
  ]
}
```

## Type of change

## Checklist before merging


src/pkg/oci/common.go
```diff
@@ -59,8 +63,26 @@ func NewOrasRemote(url string) (*OrasRemote, error) {
 	copyOpts.PostCopy = o.printLayerCopied
 	o.CopyOpts = copyOpts
 
-	o.WithContext(context.TODO())
-	o.WithInsecureConnection(zarfconfig.CommonOptions.Insecure)
+	// right now --insecure is overloaded to mean both plain HTTP and insecure TLS
+	// putting this here as the "default" for the remote
+	// but can be overridden by a provided modifier
+	if zarfconfig.CommonOptions.Insecure {
```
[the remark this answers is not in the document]
Noxsios: true, removed

src/pkg/oci/push.go
```diff
@@ -153,13 +157,97 @@ func (o *OrasRemote) PublishPackage(pkg *types.ZarfPackage, paths *layout.Packag
...
 	return nil
 }
+
+// UpdateIndex updates the index for the given package.
+func (o *OrasRemote) UpdateIndex(tag string, arch string, publishedDesc ocispec.Descriptor) error {
+	var index ocispec.Index
+
+	o.repo.Reference.Reference = tag
+	// since ref has changed, need to reset root
+	o.root = nil
```
Noxsios, line 180: `o.root` needs a mutex on it

Noxsios:
Needs a PR right after the release to update the `examples/composable-packages` skeleton ref to not have `-skeleton`

Racer159 (changes requested):
partial review

Noxsios:
Example showing the new tabs that show up in GHCR's UI for multi arch

Screenshot 2023-12-15 at 11 55 06 PM

src/pkg/packager/publish.go
```diff
@@ -30,20 +30,24 @@ func (p *Packager) Publish() (err error) {
...
-		srcRemote.WithContext(ctx)
 
 		parts := strings.Split(srcRemote.Repo().Reference.Repository, "/")
 		packageName := parts[len(parts)-1]
 
 		p.cfg.PublishOpts.PackageDestination = p.cfg.PublishOpts.PackageDestination + "/" + packageName
 
-		err = p.setOCIRemote(p.cfg.PublishOpts.PackageDestination)
+		arch := config.GetArch()
+		dstRemote, err := oci.NewOrasRemote(p.cfg.PublishOpts.PackageDestination, oci.WithArch(arch))
```
AustinAbro321, line 40: Check if there is an old zarf package here and check the architecture rather than using the default on the users machine

src/pkg/packager/publish.go
```diff
@@ -57,17 +61,24 @@ func (p *Packager) Publish() (err error) {
 		}
 		expected := content.NewDescriptorFromBytes(ocispec.MediaTypeImageManifest, b)
 
-		// tag the manifest the same as the source
-		if err := p.remote.Repo().Manifests().PushReference(ctx, expected, bytes.NewReader(b), srcRemote.Repo().Reference.Reference); err != nil {
+		srcRoot, err = srcRemote.ResolveRoot()
```
AustinAbro321, line 64: Fetching root twice

src/pkg/packager/sources/oci.go
```diff
@@ -197,7 +197,7 @@ func (s *OCISource) Collect(dir string) (string, error) {
 
 	spinner.Success()
 
-	isSkeleton := strings.HasSuffix(s.Repo().Reference.Reference, oci.SkeletonSuffix)
+	isSkeleton := pkg.Build.Architecture == "skeleton" || strings.HasSuffix(s.Repo().Reference.Reference, oci.SkeletonArch)
```
AustinAbro321, line 200:
```suggestion
	// TODO (@Noxsios) remove the suffix check at v1.0.0
	isSkeleton := pkg.Build.Architecture == "skeleton" || strings.HasSuffix(s.Repo().Reference.Reference, oci.SkeletonArch)
```

src/test/e2e/50_oci_publish_deploy_test.go
AustinAbro321, line 1: Add a flavor test, just publish and inspect is enough

src/pkg/packager/publish.go
```diff
@@ -30,20 +30,34 @@ func (p *Packager) Publish() (err error) {
...
+		}
+
+		pkg, err := srcRemote.FetchZarfYAML()
+		if err != nil {
+			return err
+		}
+
+		// ensure cli arch matches package arch
+		if pkg.Build.Architecture != arch {
+			return fmt.Errorf("architecture mismatch (expected: %q, got %q)", arch, pkg.Build.Architecture)
```
Racer159, line 57:
```suggestion
			return fmt.Errorf("architecture mismatch (specified: %q, found %q)", arch, pkg.Build.Architecture)
```

Racer159 (changes requested):

Racer159 (approved):
lgtm

PR #2196, opened 2023-12-14
 feat: support authenticated Helm repositories from helm repo add

AustinAbro321:
## Description

Support authenticated Helm repositories from helm repo add in zarf. Zarf should be able to package and find images for any repo already on a users system

## Related Issue

Fixes # 2189

## Type of change

## Checklist before merging


AustinAbro321:
Should we add this use case to the example helm chart? I'm leaning towards no to keep things simple to build though I can see either way

AustinAbro321:
Thinking about how we want to pull the chart tarball for the test. We can build our own from an example or we can pull from the internet

docs/3-create-a-zarf-package/2-zarf-components.md
```diff
@@ -68,6 +68,12 @@ Charts using the `url` key can be:
 - A remote URL (oci://) to an OCI registry
 - A remote URL (http/https) to a Helm repository
 
+:::note
+
+To use a private helm repoistory the repo must be added to helm. You can add a repo to helm with the [helm repo add](https://helm.sh/docs/helm/helm_repo_add/) command.
```
Noxsios, line 73:
```suggestion
To use a private Helm repository the repo must be added to Helm. You can add a repo to Helm with the [`helm repo add`](https://helm.sh/docs/helm/helm_repo_add/) command.
```

src/internal/packager/helm/repo.go
```diff
@@ -143,6 +143,14 @@ func (h *Helm) DownloadPublishedChart(cosignKeyPath string) error {
 		chartURL  string
 		err       error
 	)
+	repoFile, err := repo.LoadFile(pull.Settings.RepositoryConfig)
+
+	if err != nil {
+		message.Debug("Unable to load the repo file at %q: %s", pull.Settings.RepositoryConfig, err.Error())
+	}
+
```
Noxsios, line 151: I would leave a comment here denoting why this is not a failure case for this function, despite an `error` occurring, or return this error

Noxsios: this also needs to be `DebugF`, not `Debug` if using `%q` and `%s`

src/internal/packager/helm/repo.go
```diff
@@ -159,8 +167,14 @@ func (h *Helm) DownloadPublishedChart(cosignKeyPath string) error {
...
 		}
 
-		// Perform simple chart download
-		chartURL, err = repo.FindChartInRepoURL(h.chart.URL, chartName, h.chart.Version, pull.CertFile, pull.KeyFile, pull.CaFile, getter.All(pull.Settings))
+		for _, repo := range repoFile.Repositories {
+			if repo.URL == h.chart.URL {
+				username = repo.Username
+				password = repo.Password
+			}
+		}
```
Noxsios, line 175: can `repositories.yaml` contain multiple entries w/ the same URL?

src/test/external/ext_out_cluster_test.go
```diff
@@ -136,6 +146,103 @@ func (suite *ExtOutClusterTestSuite) Test_1_Deploy() {
 	suite.True(success, errorStr)
 }
 
+func (suite *ExtOutClusterTestSuite) Test_2_AuthToPrivateHelmChart() {
+	baseURL := fmt.Sprintf("http://%s", "gitea.localhost:3000")
+
+	createHelmChartInGitea(suite, baseURL, giteaUser, giteaPassword)
+	makeUserPrivate(suite, baseURL, giteaUser, giteaPassword)
+
```
Noxsios, line 154: instead of passing in `suite`, you can make these functions receivers on `suite`, just don't prefix them w/ `Test` and they will not be executed as tests by the suite itself

src/test/external/ext_out_cluster_test.go
```diff
@@ -136,6 +146,103 @@ func (suite *ExtOutClusterTestSuite) Test_1_Deploy() {
...
+	repoFile := repo.NewFile()
+
+	chartURL := fmt.Sprintf("%s/api/packages/%s/helm", baseURL, giteaUser)
+	Entry := repo.Entry{
+		Name:     "temp_entry",
+		Username: giteaUser,
+		Password: giteaPassword,
+		URL:      chartURL,
+	}
+	repoFile.Add(&Entry)
```
Noxsios, line 174:
```suggestion
	entry := &repo.Entry{
		Name:     "temp_entry",
		Username: giteaUser,
		Password: giteaPassword,
		URL:      chartURL,
	}
	repoFile.Add(entry)
```

src/test/external/ext_out_cluster_test.go
```diff
@@ -136,6 +146,103 @@ func (suite *ExtOutClusterTestSuite) Test_1_Deploy() {
...
+
+	req.Header.Set("Content-Type", writer.FormDataContentType())
+	req.SetBasicAuth(username, password)
+
+	client := &http.Client{}
+
+	resp, err := client.Do(req)
+	suite.NoError(err)
+	defer resp.Body.Close()
+}
```
Noxsios, line 218: why use cURL for one task, then Go's HTTP implementation for another?

Noxsios (changes requested):

src/internal/packager/helm/repo.go
Racer159: Yes 
image

src/test/external/ext_out_cluster_test.go
```diff
@@ -5,33 +5,42 @@
...
-	clusterName  = "zarf-external-test"
+	network       = "k3d-k3s-external-test"
+	subnet        = "172.31.0.0/16"
+	gateway       = "172.31.0.1"
+	giteaIP       = "172.31.0.99"
+	giteaHost     = "gitea.localhost"
+	registryHost  = "registry.localhost"
+	clusterName   = "zarf-external-test"
+	giteaUser     = "git-user"
+	giteaPassword = "superSecurePassword"
```
Racer159, line 35:
```suggestion
	commonPassword = "superSecurePassword"
```
Technically this password is used between gitea and the registry

src/test/packages/external/zarf.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: helm-charts
+  description: Example showcasing mutiple ways to deploy helm charts
+  version: 0.0.1
+components:
+- name: demo-helm-charts
+  required: true
+  charts:
+  - name: podinfo
+    version: 6.5.4
```
Racer159, line 11: Can we use 6.4.0 instead to keep our podinfo version consistent (it helps with filesystem space considerations in CI)

Racer159 (changes requested):
a few more changes

src/internal/packager/helm/repo.go
AustinAbro321: Great point, and as far as I can tell you can have different levels of access depending on the helm repo. So someone may have access to version 6.5.0 in repo podinfo but not have access to that version with 6.5.1. Will look into changing that

src/test/packages/external/zarf.yaml
AustinAbro321, line 1: @Racer159 I added this file in packages/external. This is what is built during the external test to make sure we have access to the helm repo. LMK if there is a better place to put this in your opinion.

src/internal/packager/helm/repo.go
AustinAbro321: After discussions with @Racer159 we decided to move forward with this PR for now as it is a pretty unlikely edge case. We added a TODO to the code and plan on handling this case later.

Noxsios (approved):

Racer159:
Adding this here for some of the tech debt :credit_card: that we delayed from being in this PR: https://github.com/defenseunicorns/zarf/issues/2245

PR #2210, opened 2023-12-29
feat: reproducible tarballs for components + sboms

### LINKED ISSUE #2199: Reproducible Tarballs

Component tarballs are not reproducible, needs investigation as to which tar headers need to be stripped similar to:

https://github.com/oras-project/oras-go/commit/93c4cc2ac355f556190fdc0eb89d43e335f965d5

https://github.com/moby/moby/blob/v24.0.7/pkg/archive/archive.go#L464

### Additional context

Reach out to @Noxsios or @Racer159 for more info

Noxsios:
## Description

`archiver@v3` does not expose the functionality needed to create tarballs with file headers containing only deterministic information.  As such, back to back package `create`s against the same data will result in differences in SHAs of `components/*.tar` and `sboms.tar`.

To remedy this, tarballing up these directories manually is the only current path forward in order to guarantee reproducibility. `archiver@v4` contains such functionality, but is still in `alpha`.

## Related Issue

Fixes #2199 

## Type of change

## Checklist before merging


src/pkg/utils/io.go
```diff
@@ -422,3 +424,64 @@ func SHAsMatch(path, expected string) error {
 	}
 	return nil
 }
+
+// CreateReproducibleTarballFromDir creates a tarball from a directory with stripped headers
+func CreateReproducibleTarballFromDir(dirPath, dirPrefix, tarballPath string) error {
```
Racer159, line 429: Can we check this real fast to ensure that this works cross platform (should be easy in a <20 test to have a package create run, then unarchive and check the component / sbom tarballs, then create againa nd check again)

Racer159 (changes requested):

src/test/e2e/05_tarball_test.go
```diff
@@ -0,0 +1,93 @@
...
+	require.NoError(t, err, stdOut, stdErr)
+
+	var pkg2 types.ZarfPackage
+	err = utils.ReadYaml(filepath.Join(unpack2, layout.ZarfYAML), &pkg2)
+	require.NoError(t, err)
+
+	b, err = os.ReadFile(filepath.Join(unpack2, layout.Checksums))
+	require.NoError(t, err)
+	checksums2 := string(b)
+
```
Racer159, line 89:
```suggestion
	checksums2 := string(b)

	e2e.CleanFiles(unpack2, tb)

```

Racer159 (changes requested):
needs a clean files for the 2nd unpack

Racer159 (approved):
lgtm

PR #2211, opened 2024-01-02
Required key --> Optional key (default is `optional: false`)

### LINKED ISSUE #2059: Breaking Change: make components required by default
Currently, all zarf components default to being optional due to the `required` key being optional in the yaml. This leads to behaviors where a package author must ensure they are careful to annotate this key for each component--though the validation doesn't require them to so assumes a sort of "all things are optional" default state.  

When Zarf was first created, we didn't really know how it would evolve and this key was introduced in those very early days. I think at this point it would be better to require all components by default--especially with the introduction of composability and the OCI skeleton work, there is plenty of flexibility in the API to compose bespoke packages assembled from other packages. 

I see a few ways we could handle this:

1. Simply force the `required` key to be a non-optional, so that package authors would be forced to specify it for each component, thereby removing any ambiguity--but also force one more key for every single component every created 🫠

2. Deprecate `required` and introduce an optional `optional` key, which would default to false.  I do think this still feels strange if you did something like `optional: false`.

3. Do something more significant like combine various condition-based things such as `only`, `optional` (instead of `required`), `default`, I would have said `group`, but... 😆

Noxsios:
## Description

> see title for now

## Related Issue

Fixes #2059 

## Type of change

## Checklist before merging

### DIFF
--- a/src/types/component.go
+++ b/src/types/component.go
@@ -21,8 +21,10 @@ type ZarfComponent struct {
 	// Default changes the default option when deploying this component
 	Default bool `json:"default,omitempty" jsonschema:"description=Determines the default Y/N state for installing this component on package deploy"`
 
-	// Required makes this component mandatory for package deployment
-	Required bool `json:"required,omitempty" jsonschema:"description=Do not prompt user to install this component, always install on package deploy"`
+	// DeprecatedRequired makes this component mandatory for package deployment
+	DeprecatedRequired *bool `json:"required,omitempty" jsonschema:"description=[Deprecated] Do not prompt user to install this component, always install on package deploy. This will be removed in Zarf v1.0.0.,deprecated=true"`
+
+	Optional *bool `json:"optional,omitempty" jsonschema:"description=Prompt user to install this component on package deploy, defaults to false"`
 
 	// Only include compatible components during package deployment
 	Only ZarfComponentOnlyTarget `json:"only,omitempty" jsonschema:"description=Filter when this component is included in package creation or deployment"`
@@ -63,6 +65,24 @@ type ZarfComponent struct {
 
 	// Replaces scripts, fine-grained control over commands to run at various stages of a package lifecycle
 	Actions ZarfComponentActions `json:"actions,omitempty" jsonschema:"description=Custom commands to run at various stages of a package lifecycle"`
+}
+
+func (c ZarfComponent) IsRequired() bool {
+	requiredExists := c.DeprecatedRequired != nil
+	optionalExists := c.Optional != nil
+	required := requiredExists && *c.DeprecatedRequired
+	optional := optionalExists && *c.Optional
+
+	if requiredExists && optionalExists {
+		return required && !optional
+	} else if requiredExists {
+		return required
+	} else if optionalExists {
+		return !optional
+	}
+
+	// If neither required nor optional are set, then the component is required
+	return true
 }
 
 // ZarfComponentOnlyTarget filters a component to only show it for a given local OS and cluster.
--- a/docs/3-create-a-zarf-package/4-zarf-schema.md
+++ b/docs/3-create-a-zarf-package/4-zarf-schema.md
@@ -611,12 +611,12 @@ must respect the following conditions
 
 <details>
 <summary>
-<strong> <a name="components_items_required"></a>required</strong>
+<strong> <a name="components_items_optional"></a>optional</strong>
 </summary>
 &nbsp;
 <blockquote>
 
-**Description:** Do not prompt user to install this component
+**Description:** Prompt user to install this component on package deploy
 
 |          |           |
 | -------- | --------- |
--- a/src/internal/packager/validate/validate.go
+++ b/src/internal/packager/validate/validate.go
@@ -129,7 +129,7 @@ func validateComponent(pkg types.ZarfPackage, component types.ZarfComponent) err
 		return fmt.Errorf(lang.PkgValidateErrComponentName, component.Name)
 	}
 
-	if component.Required {
+	if component.IsRequired() {
 		if component.Default {
 			return fmt.Errorf(lang.PkgValidateErrComponentReqDefault, component.Name)
 		}
--- a/src/pkg/oci/pull.go
+++ b/src/pkg/oci/pull.go
@@ -102,7 +102,7 @@ func (o *OrasRemote) LayersFromRequestedComponents(requestedComponents []string)
 	}
 	for _, component := range pkg.Components {
 		// If we requested this component, or it is required, we need to pull its images and tarball
-		if slices.Contains(requestedComponents, component.Name) || component.Required {
+		if slices.Contains(requestedComponents, component.Name) || component.IsRequired() {
 			for _, image := range component.Images {
 				images[image] = true
 			}
--- a/src/pkg/packager/components.go
+++ b/src/pkg/packager/components.go
@@ -61,7 +61,7 @@ func (p *Packager) getSelectedComponents() []types.ZarfComponent {
 
 				selectState, matchedRequest := includedOrExcluded(component, requestedComponents)
 
-				if !component.Required {
+				if !component.IsRequired() {
 					if selectState == excluded {
 						// If the component was explicitly excluded, record the match and continue
 						matchedRequests[matchedRequest] = true
@@ -117,7 +117,7 @@ func (p *Packager) getSelectedComponents() []types.ZarfComponent {
 			} else {
 				component := groupedComponents[groupKey][0]
 
-				if component.Required {
+				if component.IsRequired() {
 					selectedComponents = append(selectedComponents, component)
 				} else if selected := interactive.SelectOptionalComponent(component); selected {
 					selectedComponents = append(selectedComponents, component)
--- a/src/pkg/packager/composer/override.go
+++ b/src/pkg/packager/composer/override.go
@@ -13,7 +13,12 @@ import (
 func overrideMetadata(c *types.ZarfComponent, override types.ZarfComponent) error {
 	c.Name = override.Name
 	c.Default = override.Default
-	c.Required = override.Required
+	if override.DeprecatedRequired != nil {
+		c.DeprecatedRequired = override.DeprecatedRequired
+	}
+	if override.Optional != nil {
+		c.Optional = override.Optional
+	}
 
 	// Override description if it was provided.
 	if override.Description != "" {
--- a/zarf.schema.json
+++ b/zarf.schema.json
@@ -225,7 +225,11 @@
         },
         "required": {
           "type": "boolean",
-          "description": "Do not prompt user to install this component"
+          "description": "[Deprecated] Do not prompt user to install this component"
+        },
+        "optional": {
+          "type": "boolean",
+          "description": "Prompt user to install this component on package deploy"
         },
         "only": {
           "$schema": "http://json-schema.org/draft-04/schema#",


Noxsios:
TODO: 

- [ ] Talk w/ @Racer159 about composition behavior
- [ ] Think about migration behavior
- [ ] Lint addition?
- [ ] `zarf package migrate`?

.github/workflows/test-upgrade.yml
```diff
@@ -105,12 +105,10 @@ jobs:
 
       - name: Run the upgrade tests
         # NOTE: "PATH=$PATH" preserves the default user $PATH. This is needed to maintain the version of zarf installed
-        #       in a previous step. This test run will the current release to create a K3s cluster.
+        #       in a previous step. This test run will use the current release to create a K3s cluster.
         run: |
           sudo env "PATH=$PATH" CI=true zarf tools kubectl describe nodes
 
-          zarf package create src/test/upgrade --set PODINFO_VERSION=6.3.4 --confirm
```
Racer159, line 112: Not sure why this is removed - we should be able to deploy old packages with new Zarf versions (which this tests)

adr/0023-required-optional.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Status
+
+Pending
+
+## Context
+
+> Feature request: <https://github.com/defenseunicorns/zarf/issues/2059>
+
+Currently, all Zarf components default to being optional due to the `required` key being _optional_ in the yaml. This leads to behaviors where a package author must ensure they are careful to annotate this key for each component--though the validation doesn't require them to so assumes a sort of "all things are optional" default state.
```
Racer159, line 13:
```suggestion
Currently, all Zarf components default to being optional due to the `required` key being _optional_ in the yaml. This leads to package authors needing to ensure that they annotate this key for each component, and since nothing in the current validations prompts them about this they may be confused about the "all things are optional" default state.
```

Racer159: could be more concise

adr/0023-required-optional.md
```diff
@@ -0,0 +1,31 @@
...
+
+When Zarf was first created, we didn't really know how it would evolve and this key was introduced in those very early days. At this point it would be better to require all components by default--especially with the introduction of composability and the OCI skeleton work, there is plenty of flexibility in the API to compose bespoke packages assembled from other packages.
+
+A few ways to handle this:
+
+1. Simply force the `required` key to be a non-optional, so that package authors would be forced to specify it for each component, thereby removing any ambiguity--but also force one more key for every single component every created 🫠
```
Racer159, line 19:
```suggestion
1. Simply force the `required` key to be a non-optional, so that package authors would be forced to specify it for each component, thereby removing any ambiguity--but also force one more key for every single component ever created 🫠
```

adr/0023-required-optional.md
```diff
@@ -0,0 +1,31 @@
...
+
+2. Deprecate `required` and introduce an optional `optional` key, which would default to false. I do think this still feels strange if you did something like `optional: false`.
```
Racer159, line 21: Same thing with `required: false` IMO

docs/5-zarf-tutorials/0-creating-a-zarf-package.md
```diff
@@ -50,7 +50,7 @@ components:
   - name: wordpress  # specifies the name of our component and should be unique and unchanging through updates
     description: |   # (optional) a human-readable description of the component you are defining
       "Deploys the Bitnami-packaged WordPress chart into the cluster"
-    required: true   # (optional) sets the component as 'required' so that it is always deployed
+    optional: false   # (optional) sets the component as 'required' so that it is always deployed
```
Racer159, line 53: We should just remove this here since it isn't really needed to be introduces at this point - required is likely to be the desired default in most cases.

Racer159 (changes requested):

src/pkg/packager/sources/oci.go
```diff
@@ -40,11 +40,17 @@ func (s *OCISource) LoadPackage(dst *layout.PackagePaths, unarchiveAll bool) (er
...
-	optionalComponents := helpers.StringToSlice(s.OptionalComponents)
-
 	// pull only needed layers if --confirm is set
 	if config.CommonOptions.Confirm {
-		layersToPull, err = s.LayersFromRequestedComponents(optionalComponents)
+		pkg, err = s.FetchZarfYAML()
+		if err != nil {
+			return err
+		}
+		requested, err := interactive.GetSelectedComponents(s.OptionalComponents, pkg.Components)
```
lucasrod16, line 49: we should use `ForIncludedComponents` instead of `GetSelectedComponents` for remove and mirror operations

src/cmd/dev.go
```diff
@@ -60,6 +67,84 @@ var devDeployCmd = &cobra.Command{
 	},
 }
 
+var devMigrateCmd = &cobra.Command{
+	Use: "migrate",
+	// Short:   lang.CmdDevMigrateShort,
```
lucasrod16, line 72: remember to uncomment these

lucasrod16 (changes requested):

src/pkg/packager/sources/oci.go
Noxsios: remove does not need this performed, as it only calls `s.LoadPackageMetadata`

Noxsios:
Some final thoughts for review:

- Unit testing filters?
- `config.CommonOptions.Confirm` behavior w/ `filters.NewDeploymentFilter`
- `zarf dev migrate` opinionated YAML transformations
- `zarf dev migrate` `RunE` vs `Run`

src/pkg/packager/publish.go
```diff
@@ -98,7 +99,8 @@ func (p *Packager) Publish() (err error) {
 			return err
 		}
 	} else {
-		if err = p.source.LoadPackage(p.layout, false); err != nil {
+		filter := &filters.EmptyFilter{}
+		if err = p.source.LoadPackage(p.layout, filter, false); err != nil {
```
lucasrod16, line 103: What do you think about passing in `nil` as an argument instead of a dummy implementation?

```go
LoadPackage(p.layout, nil, false)
``` 
Would have to check for `nil` before calling any methods on filter in `LoadPackage`, but seems like a clear and simple way to handle calls that don't need a filter.

src/pkg/packager/filters/utils.go
```diff
@@ -0,0 +1,79 @@
...
+	optional := optionalExists && *c.Optional
+
+	if useRequiredLogic {
+		// old required logic does not know about optional
+		if requiredExists {
+			return required
+		}
+		return false
+	} else {
+		// optional "wins" when defined
```
lucasrod16, line 70: lint

src/pkg/packager/yaml.go
```diff
@@ -42,35 +41,10 @@ func (p *Packager) readZarfYAML(path string) error {
...
-
-		// If both the OS and architecture are valid, add the component to the filtered list
-		if validArch && validOS {
-			filteredComponents = append(filteredComponents, component)
-		}
-	}
-	// Update the active package with the filtered components.
-	p.cfg.Pkg.Components = filteredComponents
+	// Currently the arch and os filter will never return an error.
+	p.cfg.Pkg.Components, _ = filters.NewArchAndOSFilter(p.arch, runtime.GOOS).Apply(p.cfg.Pkg)
```
lucasrod16, line 47: We should probably handle the error value returned from:

```go
components, err := filters.NewArchAndOSFilter(p.arch, runtime.GOOS).Apply(p.cfg.Pkg)z
if err != nil {
  return err
}
```

This specific implementation of `Apply()` does not actually return an error at all, but if it's possible for the function signature to return one, it's good practice to handle it in case the implementation changes in the future.

src/pkg/packager/publish.go
Noxsios: yeah, idk about using `nil` as value here, def something to discuss as a team

src/pkg/packager/filters/deploy.go
```diff
@@ -0,0 +1,194 @@
...
+					if d <= 5 {
+						closeEnough = append(closeEnough, c.Name)
+					}
+				}
+				failure := fmt.Errorf(lang.PkgDeployErrNoCompatibleComponentsForSelection, requestedComponent, strings.Join(closeEnough, ", "))
+				if err != nil {
+					err = fmt.Errorf("%w, %w", err, failure)
+				} else {
+					err = failure
+				}
```
Racer159, line 146: This could create a kind of crazy error message if it looped a few times.  It might be best to put everything together outside of the loop (a simple flag error boolean and a list of close enoughs would work to then return an error - it would also be a little simpler logic wise).

src/pkg/packager/filters/deploy.go
```diff
@@ -0,0 +1,194 @@
...
+					}
+				}
+			} else {
+				component := groupedComponents[groupKey][0]
+
+				if isRequired(component, useRequiredLogic) {
+					selectedComponents = append(selectedComponents, component)
+				} else if component.Default && !f.isInteractive {
+					selectedComponents = append(selectedComponents, component)
+				} else {
```
Racer159, line 184: It may be good to pull f.isInteractive() into a higher level if block - you would still fall into the else statement for optional components (it is just that that logic still knows about the non-interactive case) - this block from 155-189 can likely be simplified and feels a bit tacked on.

src/pkg/packager/filters/utils.go
```diff
@@ -0,0 +1,79 @@
...
+			return required
+		}
+		return false
+	}
+
+	// optional "wins" when defined
+	if optionalExists {
+		return !optional
+	} else if requiredExists {
+		return required
```
Racer159, line 75: required should not be considered when optional is defined.

src/pkg/utils/io.go
```diff
@@ -462,7 +462,9 @@ func CreateReproducibleTarballFromDir(dirPath, dirPrefix, tarballPath string) er
 		if err != nil {
 			return fmt.Errorf("error getting relative path: %w", err)
 		}
-		header.Name = filepath.Join(dirPrefix, name)
+		name = filepath.Join(dirPrefix, name)
+		name = filepath.ToSlash(name)
+		header.Name = name
```
Racer159, line 467: Can we make this a path? I would like it to go into a v0.32.3

src/types/component.go
```diff
@@ -21,50 +21,95 @@ type ZarfComponent struct {
...
+	hasDataInjections := len(c.DataInjections) > 0
+
+	if hasImages || hasCharts || hasManifests || hasRepos || hasDataInjections {
+		return true
+	}
+
+	return false
+}
+
+// IsEmpty returns if the components fields (other than the fields we were told to ignore) are empty or set to the types zero-value
```
Racer159, line 86:
```suggestion
// IsEmpty returns true if the components fields (other than the fields we were told to ignore) are empty or set to the types zero-value
```

Racer159 (changes requested):

src/cmd/dev.go
```diff
@@ -60,6 +68,112 @@ var devDeployCmd = &cobra.Command{
...
+	Short: lang.CmdDevMigrateShort,
+	Long:  lang.CmdDevMigrateLong,
+	Args:  cobra.MaximumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		var dir string
+		if len(args) == 0 {
+			dir = "."
+		} else {
+			dir = args[0]
+		}
```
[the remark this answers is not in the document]
Noxsios: that func uses `pkgConfig`: `common.SetBaseDirectory(args, &pkgConfig)`. This operation does not utilize `packager` at all.

src/pkg/utils/io.go
Noxsios: need more context

packages/zarf-registry/zarf.yaml
```diff
@@ -97,6 +95,7 @@ components:
         executable: true
 
   - name: zarf-seed-registry
+    optional: true
```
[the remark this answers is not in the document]
Noxsios: changing as desired, just commenting for clarity

src/pkg/packager/filters/deploy.go
Noxsios: kk, fail on first error here to reduce craziness

src/pkg/packager/migrations/common.go
```diff
@@ -1,80 +1,71 @@
...
+// Migration represents a migration on types satisfying the PackageOrComponent interface.
+type Migration[T PackageOrComponent] interface {
+	fmt.Stringer
+	Run(T) (T, string)
+}
+
+// DeprecatedMigration represents a migration on types satisfying the PackageOrComponent interface.
+type DeprecatedMigration[T PackageOrComponent] interface {
+	Migration[T]
+	Clear(T) T
```
AustinAbro321, line 53: This interface feels a bit weird to me, I don't like that it's easy to clear without running first. For example in the readZarfYaml function we only clear if the migrations are set, whereas on main currently we will migrate. We also don't clear scripts after the actions are migrated, though that behavior is the same as main, but I think it should be changed and tested.  

I can't think of any situations where we would want to clear without running first, I think it'd probably be fine if we only had one migration interface and the deprecated migrations also cleared on run, but I'm open to other ideas.

AustinAbro321 (changes requested):

adr/0025-required-optional.md
```diff
@@ -0,0 +1,127 @@
...
+
+3. Do something more significant like combine various condition-based things such as `only`, `optional` (instead of `required`), or `default`.
+
+4. Introduce `feature` flags to allow for certain schema behavior to be configurable to the user.
+
+## Decision
+
+Option 4. Introduce `.metadata.features` flags to `zarf.yaml` to allow for certain schema behavior to be configurable to the user.
+
+The `features` key will be added to the `metadata` section of the package schema. This key will be an array of strings, where each string is the name of a beta feature that can be enabled. To enable a feature, the user will need to add the name of the feature to this array and edit the package schema accordingly.
```
lucasrod16, line 31: I believe the `.metadata.features` field might not be necessary for this purpose. Have we considered keeping the `required` key optional in the schema and defaulting to `required: true` for all components unless explicitly set to `required: false`? This would achieve the same behavior as the `default-required` flag but without needing the feature flag.

lucasrod16 (commented):
The behavior changes seem reasonable, but I believe we should discuss the addition of the `.metadata.features` field to the package schema further.

AustinAbro321:
This is going to be done in the v1 schema instead

PR #2212, opened 2024-01-02
fix: `ResolveRoot` behavior

Noxsios:
## Description

When using Zarf's OCI client: `OrasRemote`, there is backwards compatibility to support previous package references. Old packages were published under image manifests, while new packages are published under image indexes. Resolving the manifest stored within an image index requires knowing which os/arch (`ocispec.Platform`) to search for.

All internal usages of this client properly use the `oci.WithArch` or `oci.WithPlatform` modifiers that add a `targetPlatform` to the client so that `r.ResolveRoot` is able to properly leverage `oras.ResolveOptions` and locate the correct manifest within the index.

However, if no platform modifier is used, `targetPlatform` remains `nil`. This causes `oras.Resolve` to return the descriptor of the index, not the manifest. To remedy this, simply error after the backwards compat check if no modifier was used to add a `targetPlatform` onto the remote client.

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/oci/fetch.go
+++ b/src/pkg/oci/fetch.go
@@ -30,7 +30,10 @@ func (o *OrasRemote) ResolveRoot() (ocispec.Descriptor, error) {
 		return desc, nil
 	}
 
-	// if target platform is nil, oras.Resolve falls back to a o.repo.Resolve call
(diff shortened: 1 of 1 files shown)


Racer159 (approved):
lgtm

PR #2217, opened 2024-01-08
feat: support yaml extension

AustinAbro321:
## Description

Support yaml extension in lint following these rules https://docs.docker.com/compose/compose-file/11-extension/

## Related Issue

Relates to #2064 

## Type of change

## Checklist before merging


src/pkg/packager/lint/lint.go
Racer159, line 1: You can do patternProperties: `"^x-"` to make this work for jsonschema

https://github.com/alecthomas/jsonschema/blob/9eeeec9d044b28001bc1a5aaf61333d60ac33a64/reflect.go#L74

Racer159 (changes requested):

hack/.templates/jsonschemaforhumans/section_properties_details.md
Racer159, line 1: `is_pattern_property` should be cleaned up where it is no longer used.

zarf.schema.json
```diff
@@ -552,7 +585,7 @@
           "description": "The name of the resource or selector to wait for",
           "examples": [
             "podinfo",
-            "app\u0026#61;podinfo"
+            "app&#61;podinfo"
```
Racer159, line 588: These changes are going to break the website since they will display funny

Racer159 (changes requested):

zarf.schema.json
Racer159: ah ok no you're right - the \u was to get around limitations in the jsonschema library but looks like `jq` did this change which is after that part of the process

Racer159 (changes requested):

src/pkg/packager/lint/lint.go
```diff
@@ -267,7 +267,35 @@ func makeFieldPathYqCompat(field string) string {
...
+		if slice, ok := value.([]interface{}); ok {
+			for _, item := range slice {
+				if nestedMap, ok := item.(map[string]interface{}); ok {
+					removeYamlExtensionFields(nestedMap)
+				}
+			}
+		}
+	}
+}
+
```
Racer159, line 293:
```suggestion
```

src/pkg/packager/lint/lint.go
```diff
@@ -267,7 +267,35 @@ func makeFieldPathYqCompat(field string) string {
...
 func validateSchema(validator *Validator) error {
+	if zarfPackageMap, ok := validator.untypedZarfPackage.(map[string]interface{}); ok {
+		removeYamlExtensionFields(zarfPackageMap)
+	}
+
```
Racer159, line 298:
```suggestion
```

Racer159 (changes requested):
Should have noticed earlier, but this code is not needed now that x- is allowed by schema

Racer159 (approved):
lgtm

PR #2220, opened 2024-01-09
feat: check for available node resources before building injector pod

### LINKED ISSUE #2144: Make sure node has enough resources available for scheduling injector pod
When zarf attempts to run the injector since it's tied to a specific node it should check if the node has sufficient resources available, it does not and can end up in a stuck state where the pod is never scheduled if the node doesn't have enough resources available.

- **Given** a cluster with workloads
- **When** zarf is picking a node for the injector pod
- **Then** zarf checks the resource availability on nodes before picking one

chrishorton:
## Description

Addresses https://github.com/defenseunicorns/zarf/issues/2144 to check if node has enough resources before the injector runs

## Related Issue

Fixes #2144 

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/injector.go
+++ b/src/pkg/cluster/injector.go
@@ -28,6 +28,13 @@ import (
 // The chunk size for the tarball chunks.
 var payloadChunkSize = 1024 * 768
 
+var (
+	requestedCPU    = resource.MustParse(".5")
+	requestedMemory = resource.MustParse("64Mi")
+	limitCPU        = resource.MustParse("1")
+	limitMemory     = resource.MustParse("256Mi")
+)
+
 // StartInjectionMadness initializes a Zarf injection into the cluster.
 func (c *Cluster) StartInjectionMadness(tmpDir string, imagesDir string, injectorSeedSrcs []string) {
 	spinner := message.NewProgressSpinner("Attempting to bootstrap the seed image into the cluster")
@@ -54,7 +61,7 @@ func (c *Cluster) StartInjectionMadness(tmpDir string, imagesDir string, injecto
 	// Get all the images from the cluster
 	timeout := 5 * time.Minute
 	spinner.Updatef("Getting the list of existing cluster images (%s timeout)", timeout.String())
-	if images, err = c.GetAllImages(timeout); err != nil {
+	if images, err = c.GetAllImages(timeout, requestedCPU, requestedMemory); err != nil {
 		spinner.Fatalf(err, "Unable to generate a list of candidate images to perform the registry injection")
 	}
 
@@ -362,12 +369,12 @@ func (c *Cluster) buildInjectionPod(node, image string, payloadConfigmaps []stri
 			// Keep resources as light as possible as we aren't actually running the container's other binaries
 			Resources: corev1.ResourceRequirements{
 				Requests: corev1.ResourceList{
-					corev1.ResourceCPU:    resource.MustParse(".5"),
-					corev1.ResourceMemory: resource.MustParse("64Mi"),
+					corev1.ResourceCPU:    requestedCPU,
+					corev1.ResourceMemory: requestedMemory,
 				},
 				Limits: corev1.ResourceList{
-					corev1.ResourceCPU:    resource.MustParse("1"),
-					corev1.ResourceMemory: resource.MustParse("256Mi"),
+					corev1.ResourceCPU:	   limitCPU,
+					corev1.ResourceMemory: limitMemory,
 				},
 			},
 		},
--- a/src/pkg/cluster/injector.go
+++ b/src/pkg/cluster/injector.go
@@ -29,10 +29,10 @@ import (
 var payloadChunkSize = 1024 * 768
 
 var (
-	requestedCPU    = resource.MustParse(".5")
-	requestedMemory = resource.MustParse("64Mi")
-	limitCPU        = resource.MustParse("1")
-	limitMemory     = resource.MustParse("256Mi")
+	injectorRequestedCPU    = resource.MustParse(".5")
+	injectorRequestedMemory = resource.MustParse("64Mi")
+	injectorLimitCPU        = resource.MustParse("1")
+	injectorLimitMemory     = resource.MustParse("256Mi")
 )
 
 // StartInjectionMadness initializes a Zarf injection into the cluster.
@@ -61,7 +61,7 @@ func (c *Cluster) StartInjectionMadness(tmpDir string, imagesDir string, injecto
 	// Get all the images from the cluster
 	timeout := 5 * time.Minute
 	spinner.Updatef("Getting the list of existing cluster images (%s timeout)", timeout.String())
-	if images, err = c.GetAllImages(timeout, requestedCPU, requestedMemory); err != nil {
+	if images, err = c.GetAllImages(timeout, injectorRequestedCPU, injectorRequestedMemory); err != nil {
 		spinner.Fatalf(err, "Unable to generate a list of candidate images to perform the registry injection")
 	}
 
@@ -369,12 +369,12 @@ func (c *Cluster) buildInjectionPod(node, image string, payloadConfigmaps []stri
 			// Keep resources as light as possible as we aren't actually running the container's other binaries
 			Resources: corev1.ResourceRequirements{
 				Requests: corev1.ResourceList{
-					corev1.ResourceCPU:    requestedCPU,
-					corev1.ResourceMemory: requestedMemory,
+					corev1.ResourceCPU:    injectorRequestedCPU,
+					corev1.ResourceMemory: injectorRequestedMemory,
 				},
 				Limits: corev1.ResourceList{
-					corev1.ResourceCPU:	   limitCPU,
-					corev1.ResourceMemory: limitMemory,
+					corev1.ResourceCPU:	   injectorLimitCPU,
+					corev1.ResourceMemory: injectorLimitMemory,
 				},
 			},
 		},
--- a/src/pkg/k8s/images.go
+++ b/src/pkg/k8s/images.go
@@ -20,7 +20,7 @@ type ImageMap map[string]bool
 type ImageNodeMap map[string][]string
 
 // GetAllImages returns a list of images and their nodes found in pods in the cluster.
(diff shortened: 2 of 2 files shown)


chrishorton:
Description

Addresses #2144 to check if node has enough resources before the injector runs

Related Issue: #2144

Fixes #2144

src/pkg/k8s/images.go
```diff
@@ -68,6 +69,13 @@ findImages:
 			return nil, fmt.Errorf("unable to get the node %s", pod.Spec.NodeName)
 		}
 
+		requestedCPU := resource.MustParse(".5")
+		requestedMemory := resource.MustParse("64MiB")
+		if nodeDetails.Status.Allocatable.Cpu().Cmp(requestedCPU) >= 0 &&
+		   nodeDetails.Status.Allocatable.Memory().Cmp(requestedMemory) >= 0 {
```
Racer159, line 75:
```suggestion
		if nodeDetails.Status.Allocatable.Cpu().Cmp(requestedCPU) < 0 ||
		   nodeDetails.Status.Allocatable.Memory().Cmp(requestedMemory) < 0 {
```
The logic here seems inverted.

Also, I think it is ok if we accept a node if its allocatable resources are equal, there is a chance it could fill up in the time we have remaining (there always could be a chance of that) but it would likely be better to try if we can than fail when things would otherwise have worked.

Racer159 (changes requested):
Thanks for the PR!  Looks like some of the logic needs updating and passing down the resources into `GetImagesWithNodes` from the `cluster` package would help keep the code drier.

src/pkg/k8s/images.go
chrishorton: Thanks for catching that!

Okay thanks for clarifying, that makes sense to me.

src/pkg/k8s/images.go
```diff
@@ -68,6 +69,13 @@ findImages:
 			return nil, fmt.Errorf("unable to get the node %s", pod.Spec.NodeName)
 		}
 
+		requestedCPU := resource.MustParse(".5")
+		requestedMemory := resource.MustParse("64MiB")
```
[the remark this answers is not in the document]
chrishorton: I agree. Made these into `cluster` package variables and passed as arguments into `k8s` package `GetAllImages()` -> `GetImagesWithNodes()`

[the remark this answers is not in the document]
chrishorton: Whoops. Caught this in a different commit but it didn't make it in. Fixed now!

chrishorton:
Resolved all your suggestions in latest commit, thanks!

src/pkg/cluster/injector.go
```diff
@@ -28,6 +28,13 @@ import (
 // The chunk size for the tarball chunks.
 var payloadChunkSize = 1024 * 768
 
+var (
+	requestedCPU    = resource.MustParse(".5")
+	requestedMemory = resource.MustParse("64Mi")
+	limitCPU        = resource.MustParse("1")
+	limitMemory     = resource.MustParse("256Mi")
```
Racer159, line 35: tiny thing on these - since they are specific to the injector pod they should include a reference in the name incase people see them elsewhere when typing `cluster.` with IntelliSense

```suggestion
	injectorRequestedCPU    = resource.MustParse(".5")
	injectorRequestedMemory = resource.MustParse("64Mi")
	injectorLimitCPU        = resource.MustParse("1")
	injectorLimitMemory     = resource.MustParse("256Mi")
```

src/pkg/k8s/images.go
```diff
@@ -19,7 +20,7 @@ type ImageMap map[string]bool
 type ImageNodeMap map[string][]string
 
 // GetAllImages returns a list of images and their nodes found in pods in the cluster.
-func (k *K8s) GetAllImages(timeoutDuration time.Duration) (ImageNodeMap, error) {
+func (k *K8s) GetAllImages(timeoutDuration time.Duration, requestedCPU resource.Quantity, requestedMemory resource.Quantity) (ImageNodeMap, error) {
```
Racer159, line 23:
```suggestion
func (k *K8s) GetAllImages(timeoutDuration time.Duration, minNodeCPU resource.Quantity, minNodeMemory resource.Quantity) (ImageNodeMap, error) {
```
Another name suggestion but minNodeCPU may help folks better understand what these are if they're using them as a library.

src/pkg/k8s/images.go
```diff
@@ -46,7 +47,7 @@ func (k *K8s) GetAllImages(timeoutDuration time.Duration) (ImageNodeMap, error)
 
 // GetImagesWithNodes checks for images on schedulable nodes and returns
 // a map of these images and their nodes in a given namespace.
-func (k *K8s) GetImagesWithNodes(namespace string) (ImageNodeMap, error) {
+func (k *K8s) GetImagesWithNodes(namespace string, requestedCPU resource.Quantity, requestedMemory resource.Quantity) (ImageNodeMap, error) {
```
Racer159, line 50:
```suggestion
func (k *K8s) GetImagesWithNodes(namespace string, minNodeCPU resource.Quantity, minNodeMemory resource.Quantity) (ImageNodeMap, error) {
```

Racer159 (changes requested):
Thanks for the changes! Logic looks good (will let the tests run to make sure) - since this is library code made some suggestions for naming though.

chrishorton:
Pushed those changes!

src/pkg/k8s/images.go
```diff
@@ -68,6 +69,11 @@ findImages:
 			return nil, fmt.Errorf("unable to get the node %s", pod.Spec.NodeName)
 		}
 
+		if nodeDetails.Status.Allocatable.Cpu().Cmp(requestedCPU) < 0 ||
+		   nodeDetails.Status.Allocatable.Memory().Cmp(requestedMemory) < 0 {
```
Racer159, line 73:
```suggestion
		if nodeDetails.Status.Allocatable.Cpu().Cmp(minNodeCPU) < 0 ||
		   nodeDetails.Status.Allocatable.Memory().Cmp(minNodeMemory) < 0 {
```

These would need to be updated as well.

Racer159 (changes requested):

src/pkg/k8s/images.go
chrishorton: Wow. No idea how I missed that. Thanks for the help!!!

chrishorton:
@lucasrod16 I would tend to agree with you on that. Wasn't sure about creating new global variables but not changing the function signature would be more ideal since it's public facing and would be a breaking change for usages.

lucasrod16:
@chrishorton thanks for submitting this PR 🙂

Racer159:
> I think we should try to avoid changing the function signatures of `GetAllImages` and `GetImagesWithNodes` because they are public facing APIs.
> 
> ```go
> var (
> 	injectorRequestedCPU    = resource.MustParse(".5")
> 	injectorRequestedMemory = resource.MustParse("64Mi")
> 	injectorLimitCPU        = resource.MustParse("1")
> 	injectorLimitMemory     = resource.MustParse("256Mi")
> )
> ```
> 
> These variables are static and will likely not change often, so we could move them to the k8s package, make them public, and use them directly in the functions that need them instead of passing them in as parameters.
> 
> @Racer159 @Noxsios thoughts?

I suggested this direction since it would be relatively easier to implement than a larger refactor that is likely warranted for how the injector pod gets built - we only use these functions for this purpose so they could ultimately be tailored more to the injector node selection process and removed from `k8s` if we wanted.  Currently IMO it would make more sense to move things into `cluster` than to `k8s` because no Zarf-specific logic is in `k8s` today (the only "zarf" thing it imports is `helpers` and that also has a rule to not import anything else from Zarf.

The `injector*` variables are global to the `cluster` package so we could mess things up in the future but they are `private` so people importing `cluster` won't be able to mess them up without something like `go:linkname` which is why I was ok with them being there for now (would just be up to the Zarf maintainers to make sure that they aren't modified).

I do agree the function signature is a little odd taking in the node resource quantities though so we could refactor both `GetImagesWithNodes` and `GetAllImages` into a function with a better signature (potentially something like `func  GetImagesFromAvailableNodes(timeoutDuration time.Duration, minResources corev1.ResourceList)  (ImageNodeMap, error)` that would check all resource types and be a bit better named.

Racer159:
For historical context, these functions are also a bit appendix-like - they had more of a purpose back when Zarf performed more node checks on `init` but we removed some of those to reduce the permissions Zarf needs to run in a cluster.

chrishorton:
That would work. When I was working on this, trying for an elegant solution (not changing func signatures) everything produced a circular dependency because of how the logic is intertwined and used in the two packages.

> For historical context, these functions are also a bit appendix-like - they had more of a purpose back when Zarf performed more node checks on init but we removed some of those to reduce the permissions Zarf needs to run in a cluster.

I think this is also a good reason to refactor `getAllImages` and `GetImagesWithNodes` out of the k8s package.

Would be happy to work on this, if a new issue is created to refactor/migrate.

chrishorton:
> @chrishorton thanks for submitting this PR 🙂

Glad I can help a bit! I'm struggling to manually deploy a lot of our software over an air-gap so any tooling that helps is gold to me :D

Noxsios:
I don't have much to add aside from echoing that the relationship between `pkg/k8s` and `pkg/cluster` needs to be re-evaluated. A bit ago we moved `pkg/cluster` from `internal` to `pkg` (https://github.com/defenseunicorns/zarf/pull/2162) to support external usage of `cluster`. IMO `cluster` should be the de-facto library export and `k8s`'s functions should be merged in (as `cluster` is a utility for interacting w/ Zarf specific cluster needs.). Users should be using the Kubernetes K8s lib otherwise.

Doing some cursory checks, I can also see that there are some public receivers on `k8s` that are not used in Zarf, IMO those should be removed during this refactor (yes a breaking change, but 99% of them are bare abstractions on top of the core k8s lib).

Racer159:
Created this issue to capture the above feedback - https://github.com/defenseunicorns/zarf/issues/2222 (feel free to pick it up if you want @chrishorton !)

Going to approve this as is for now though with the intent that the tech debt gets addressed in the future.

Racer159 (approved):
lgtm

lucasrod16 (approved):

PR #2235, opened 2024-01-19
move oci lib out

AustinAbro321:
## Description

Move OCI lib out, so it can be used with other projects

## Type of change

## Checklist before merging


src/pkg/oci/common.go
```diff
@@ -1,224 +0,0 @@
...
-	transport := http.DefaultTransport.(*http.Transport).Clone()
-	o.Transport = utils.NewTransport(transport, nil)
-
-	client := &auth.Client{
-		Cache: auth.DefaultCache,
-		Client: &http.Client{
-			Transport: o.Transport,
-		},
-	}
-	client.SetUserAgent("zarf/" + zarfconfig.CLIVersion)
```
Racer159, line 189: user agent should be configurable in the lib

src/pkg/oci/common.go
```diff
@@ -1,224 +0,0 @@
...
-
-const (
-	// ZarfLayerMediaTypeBlob is the media type for all Zarf layers due to the range of possible content
-	ZarfLayerMediaTypeBlob = "application/vnd.zarf.layer.v1.blob"
-	// ZarfConfigMediaType is the media type for the manifest config
-	ZarfConfigMediaType = "application/vnd.zarf.config.v1+json"
-	// SkeletonArch is the architecture used for skeleton packages
-	SkeletonArch = "skeleton"
-	// MultiOS is the OS used for multi-platform packages
-	MultiOS = "multi"
```
Racer159, line 34: These should stay as consts in zarf

src/pkg/oci/copier.go
```diff
@@ -1,118 +0,0 @@
...
-	"time"
-
-	"github.com/defenseunicorns/zarf/src/pkg/message"
-	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
-	"golang.org/x/sync/errgroup"
-	"golang.org/x/sync/semaphore"
-)
-
-// CopyPackage copies a package from one OCI registry to another
-func CopyPackage(ctx context.Context, src *OrasRemote, dst *OrasRemote, include func(d ocispec.Descriptor) bool, concurrency int) error {
```
Racer159, line 20: CopyPackage could be generalized

src/pkg/oci/manifest.go
```diff
@@ -1,57 +0,0 @@
-// SPDX-License-Identifier: Apache-2.0
-// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
-
-// Package oci contains functions for interacting with Zarf packages stored in OCI registries.
```
Racer159, line 4: This file might make sense to keep as it is what defines a ZarfManifest

src/pkg/oci/pull.go
```diff
@@ -1,311 +0,0 @@
...
-	"github.com/defenseunicorns/zarf/src/types"
-	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
-	"oras.land/oras-go/v2"
-	"oras.land/oras-go/v2/content"
-	"oras.land/oras-go/v2/content/file"
-)
-
-var (
-	// PackageAlwaysPull is a list of paths that will always be pulled from the remote repository.
-	PackageAlwaysPull = []string{layout.ZarfYAML, layout.Checksums, layout.Signature}
```
Racer159, line 30: This should be configurable

src/pkg/oci/pull.go
```diff
@@ -1,311 +0,0 @@
...
-	}
-	return layers, nil
-}
-
-// LayersFromRequestedComponents returns the descriptors for the given components from the root manifest.
-//
-// It also retrieves the descriptors for all image layers that are required by the components.
-//
-// It also respects the `required` flag on components, and will retrieve all necessary layers for required components.
-func (o *OrasRemote) LayersFromRequestedComponents(requestedComponents []string) (layers []ocispec.Descriptor, err error) {
```
Racer159, line 83: This function is very Zarf specific

src/pkg/oci/pull.go
```diff
@@ -1,311 +0,0 @@
...
-
-// PullPackage pulls the package from the remote repository and saves it to the given path.
-//
-// layersToPull is an optional parameter that allows the caller to specify which layers to pull.
-//
-// The following layers will ALWAYS be pulled if they exist:
-//   - zarf.yaml
-//   - checksums.txt
-//   - zarf.yaml.sig
-func (o *OrasRemote) PullPackage(destinationDir string, concurrency int, layersToPull ...ocispec.Descriptor) ([]ocispec.Descriptor, error) {
```
Racer159, line 169: PullPackage should be genericized

src/pkg/oci/push.go
```diff
@@ -1,251 +0,0 @@
...
-	}
-	if err = src.Tag(o.ctx, root, root.Digest.String()); err != nil {
-		return ocispec.Descriptor{}, err
-	}
-
-	return root, nil
-}
-
-// PublishPackage publishes the package to the remote repository.
-func (o *OrasRemote) PublishPackage(pkg *types.ZarfPackage, paths *layout.PackagePaths, concurrency int) error {
```
Racer159, line 106: PublishPackage should be genericized

Racer159 (changes requested):
In addition to these changes we should also move /src/utils/helpers as well since that would be easy and is the core of a lot of our lib code.

Racer159 (changes requested):
After discussing more with @Noxsios it would likely be better to purify this functionality within Zarf (i.e. do the refactoring in Zarf for the OCI library to not rely on Zarf first, then move it out to keep things cleaner)

src/pkg/oci/common.go
AustinAbro321: Do we want multi to stay in OCI?

src/pkg/oci/common.go
```diff
@@ -24,33 +21,59 @@ import (
...
-	// ZarfConfigMediaType is the media type for the manifest config
-	ZarfConfigMediaType = "application/vnd.zarf.config.v1+json"
-	// SkeletonArch is the architecture used for skeleton packages
-	SkeletonArch = "skeleton"
 	// MultiOS is the OS used for multi-platform packages
 	MultiOS = "multi"
 )
 
+// Logger is an interface built to
+type Logger interface {
```
AustinAbro321, line 29: I noticed that we actually only use debug in this package, do we still want a logger interface for the future or should I just accept a single function?

src/pkg/zoci/push.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/defenseunicorns/zarf/src/pkg/layout"
+	"github.com/defenseunicorns/zarf/src/pkg/message"
+	"github.com/defenseunicorns/zarf/src/pkg/utils/helpers"
+	"github.com/defenseunicorns/zarf/src/types"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2/content/file"
+)
+
+// PublishZarfPackage publishes the zarf package to the remote repository.
+func (o *ZarfOrasRemote) PublishZarfPackage(ctx context.Context, pkg *types.ZarfPackage, paths *layout.PackagePaths, concurrency int) error {
```
Noxsios, line 21:
```suggestion
// PublishPackage publishes the zarf package to the remote repository.
func (o *ZarfOrasRemote) PublishPackage(ctx context.Context, pkg *types.ZarfPackage, paths *layout.PackagePaths, concurrency int) error {
```

src/pkg/zoci/push.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		progressBar.Stop()
+		return fmt.Errorf("unable to publish package: %w", err)
+	}
+
+	progressBar.Successf("Published %s [%s]", o.Repo().Reference, ZarfLayerMediaTypeBlob)
+	return nil
+}
+
+func zarfPackageOciAnnotations(metadata *types.ZarfMetadata) map[string]string {
```
Noxsios, line 66:
```suggestion
func annotationsFromMetadata(metadata *types.ZarfMetadata) map[string]string {
```

src/pkg/oci/common.go
AustinAbro321: Agreed, and with that we need to just how our packages will display the logs. Args ..any could be slog style or printf style. Currently I have it as printf

src/pkg/oci/common.go
```diff
@@ -24,33 +21,59 @@ import (
...
+// WithInsecure sets plainHTTP and insecure tls for the remote
+func WithInsecure(insecure bool) Modifier {
 	return func(o *OrasRemote) {
-		o.ctx = ctx
+		plainHTTPMod := WithPlainHTTP(insecure)
+		plainHTTPMod(o)
+		insecureTLSMod := WithInsecureSkipVerify(insecure)
+		insecureTLSMod(o)
 	}
 }
```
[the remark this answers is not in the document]
AustinAbro321: Ah yeah I hadn't thought about that use case

src/pkg/oci/common.go
```diff
@@ -75,31 +98,38 @@ func WithInsecureSkipVerify(insecure bool) Modifier {
...
 
+// WithUserAgent sets the target architecture for the remote
+func WithUserAgent(userAgent string) Modifier {
+	return func(o *OrasRemote) {
+		o.userAgent = userAgent
+	}
+}
+
```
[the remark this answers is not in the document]
AustinAbro321: What do you think a reasonable default user agent should be
Maybe `client.SetUserAgent("zarf/" + zarfconfig.CLIVersion)` for now and `client.SetUserAgent("defenseunicorns/pkg/oci/version")` once we move it?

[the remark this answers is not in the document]
AustinAbro321: Switched the order and changed the functions a bit to so we don't end up with nil pointers / override set modifiers

src/pkg/oci/push.go
```diff
@@ -36,154 +34,88 @@ type ConfigPartial struct {
...
-		desc, err := src.Add(ctx, name, mediaType, path)
-		if err != nil {
-			return err
-		}
-		descs = append(descs, desc)
-	}
-	spinner.Successf("Prepared all layers")
-
+// PublishArtifact publishes the artifact to the remote repository.
+func (o *OrasRemote) PublishArtifact(ctx context.Context, src *file.Store, annotations map[string]string,
```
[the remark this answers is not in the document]
AustinAbro321: Yeah, we haven't used artifact anywhere else, I think publish should be sufficient

src/pkg/oci/push.go
```diff
@@ -36,154 +34,58 @@ type ConfigPartial struct {
...
 
-// pushManifestConfigFromMetadata pushes the manifest config with metadata to the remote repository.
-func (o *OrasRemote) pushManifestConfigFromMetadata(metadata *types.ZarfMetadata, build *types.ZarfBuildData) (ocispec.Descriptor, error) {
-	annotations := map[string]string{
-		ocispec.AnnotationTitle:       metadata.Name,
-		ocispec.AnnotationDescription: metadata.Description,
+// PushManifestConfigFromMetadata pushes the manifest config with metadata to the remote repository.
+func (o *OrasRemote) PushManifestConfigFromMetadata(ctx context.Context, annotations map[string]string, configMediaType string) (*ocispec.Descriptor, error) {
+	if annotations[ocispec.AnnotationTitle] == "" {
+		return nil, fmt.Errorf("invalid annotations: please include value for %q", ocispec.AnnotationTitle)
```
Noxsios, line 45: note: will need to communicate to UDS CLI that they will need to populate this annotation, and show how to make a similar function to the one in Zarf.

Noxsios (changes requested):

src/pkg/oci/manifest.go
```diff
@@ -12,46 +12,42 @@ import (
...
 	ocispec.Manifest
 }
 
-// NewZarfOCIManifest returns a new ZarfOCIManifest.
-func NewZarfOCIManifest(manifest *ocispec.Manifest) *ZarfOCIManifest {
-	return &ZarfOCIManifest{*manifest}
+// NewManifest returns a new Manifest
+func NewManifest(manifest *ocispec.Manifest) *Manifest {
+	return &Manifest{*manifest}
 }
```
AustinAbro321, line 26: @Noxsios Any reason we shouldn't just delete this function? It's not being used in Zarf or UDS CLI. Not sure if there's a specific reason we wanted a constructor though

src/pkg/oci/pull.go
```diff
@@ -197,89 +59,51 @@ func (o *OrasRemote) PullPackage(destinationDir string, concurrency int, layersT
...
-		return fmt.Errorf("invalid media type for file layer: %s", desc.MediaType)
-	}
-	b, err := o.FetchLayer(desc)
+func (o *OrasRemote) PullLayer(ctx context.Context, destinationDir string, desc ocispec.Descriptor) error {
+	b, err := o.FetchLayer(ctx, desc)
 	if err != nil {
 		return err
 	}
 
 	rel := desc.Annotations[ocispec.AnnotationTitle]
```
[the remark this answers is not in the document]
AustinAbro321: If a layer is not a file what is it? In what situation would a layer be intentionally without a title?

src/pkg/oci/common.go
```diff
@@ -159,8 +187,9 @@ func (o *OrasRemote) setRepository(ref registry.Reference) error {
 	if err != nil {
 		return err
 	}
-	repo.Client = client
-	o.repo = repo
+
+	o.repo.Reference = repo.Reference
+	o.repo.Client = client
```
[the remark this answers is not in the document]
AustinAbro321: Because we now call mods before `setRepository` `o.repo.plainHTTP` changes from `withPlainHTTP` would be overwritten if we just set repo.

src/pkg/oci/manifest.go
```diff
@@ -12,46 +12,42 @@ import (
...
+	for _, layer := range m.Layers {
+		if include != nil && include(layer) {
+			layers = append(layers, layer)
+		} else if include == nil {
+			layers = append(layers, layer)
+		}
+	}
+	layers = append(layers, m.Config)
+	return layers
+}
```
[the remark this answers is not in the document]
AustinAbro321: Yeah it's hard for me to tell as we are currently always sending in nil. Going to think about it more

src/pkg/oci/push.go
```diff
@@ -36,154 +34,58 @@ type ConfigPartial struct {
...
 	}
 	manifestConfig := ConfigPartial{
-		Architecture: build.Architecture,
+		Architecture: o.targetPlatform.Architecture,
 		OCIVersion:   "1.0.1",
```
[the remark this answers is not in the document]
AustinAbro321: Yeah they do https://github.com/opencontainers/image-spec/blob/main/specs-go/version.go

src/pkg/utils/helpers/io.go
```diff
@@ -0,0 +1,31 @@
...
+		return fmt.Errorf("unable to write the file at %s contents:%w", path, err)
+	}
+
+	err = f.Close()
+	if err != nil {
+		return fmt.Errorf("error saving file %s: %w", path, err)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
AustinAbro321: Ah yeah, switched them all over. Changed the permissions to 0644 from 0666 in helpers.WriteFile. We may be able to go lower than that I'm not sure at the moment

src/pkg/zoci/copier.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+
+	"github.com/defenseunicorns/zarf/src/pkg/message"
+	"github.com/defenseunicorns/zarf/src/pkg/oci"
+	"github.com/defenseunicorns/zarf/src/pkg/utils/helpers"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// CopyPackage copies a zarf package from one OCI registry to another
+// TODO (@AustinAbro321) we should see if we should also copy the manifests for a complete copy in this function
```
Noxsios, line 18: yes, `CopyPackage` should also copy the manifest from `src.root` --> `dst.root`

Noxsios: `oci.Copy` is a generic mirroring / multithreaded copy operation. Since this is an opinionation on top of that, where we know we _always_ want to copy the manifests, we can do so

Noxsios: much of the code at the top of `src/pkg/packager/publish.go` can now be simplified by bringing it into this func

src/pkg/zoci/pull.go
```diff
@@ -0,0 +1,174 @@
...
+
+// PullPackage pulls the package from the remote repository and saves it to the given path.
+//
+// layersToPull is an optional parameter that allows the caller to specify which layers to pull.
+//
+// The following layers will ALWAYS be pulled if they exist:
+//   - zarf.yaml
+//   - checksums.txt
+//   - zarf.yaml.sig
+func (o *Remote) PullPackage(ctx context.Context, destinationDir string, concurrency int, layersToPull ...ocispec.Descriptor) ([]ocispec.Descriptor, error) {
```
Noxsios, line 43: nit: change from `o *Remote` to `r *Remote` (think rename symbol can do it)

src/pkg/zoci/push.go
```diff
@@ -0,0 +1,113 @@
...
+
+	progressBar := message.NewProgressBar(total, fmt.Sprintf("Publishing %s:%s", o.Repo().Reference.Repository, o.Repo().Reference.Reference))
+	o.Transport.ProgressBar = progressBar
+
+	publishedDesc, err := oras.Copy(ctx, src, root.Digest.String(), o.Repo(), "", copyOpts)
+	if err != nil {
+		return err
+	}
+
+	o.UpdateIndex(ctx, o.Repo().Reference.Reference, publishedDesc)
```
Noxsios, line 78: error is not handled here

Noxsios (changes requested):

src/pkg/oci/common.go
```diff
@@ -111,24 +157,6 @@ func NewOrasRemote(url string, platform ocispec.Platform, mods ...Modifier) (*Or
 	copyOpts.PostCopy = o.printLayerCopied
 	o.CopyOpts = copyOpts
```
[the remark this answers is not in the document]
AustinAbro321: I went ahead and implemented that. LMK what you think. Looks like UDS CLI is actually doing something similar https://github.com/search?q=repo%3Adefenseunicorns%2Fuds-cli+CopyOpts&type=code

src/pkg/oci/copier.go
```diff
@@ -6,50 +6,44 @@ package oci
...
 	start := time.Now()
 
 	for idx, layer := range layers {
-		message.Debug("Copying layer:", message.JSONValue(layer))
+		b, err := json.MarshalIndent(layer, "", "  ")
+		if err != nil {
+			src.log.Debug("ERROR marshalling json: %s", err.Error())
+		}
+		src.log.Debug("Copying layer:", string(b))
 		if err := sem.Acquire(ctx, 1); err != nil {
```
Racer159, line 47: This seems like a lot of logging - it might be better to log the layer digest (the actual contents can be looked up later if needed)

src/pkg/oci/pull.go
```diff
@@ -59,227 +44,57 @@ func (o *OrasRemote) FileDescriptorExists(desc ocispec.Descriptor, destinationDi
...
 		}
 	}
 
-	if copyOpts.FindSuccessors == nil {
-		copyOpts.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, desc ocispec.Descriptor) ([]ocispec.Descriptor, error) {
-			nodes, err := content.Successors(ctx, fetcher, desc)
-			if err != nil {
-				return nil, err
+	preCopy := copyOpts.PreCopy
+	copyOpts.PreCopy = func(ctx context.Context, desc ocispec.Descriptor) error {
```
Racer159, line 57: Woo! This seems a lot better than FindSuccessors.

src/pkg/oci/common.go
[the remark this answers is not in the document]
Noxsios: perfectly fine to remove from the struct, I would rename the receiver `GetDefaultCopyOpts` to denote that it is the default options.

Racer159 (changes requested):
A few little things

src/pkg/zoci/copier.go
```diff
@@ -35,5 +47,27 @@ func CopyPackage(ctx context.Context, src *Remote, dst *Remote, include func(d o
...
+	if err := dst.Repo().Manifests().PushReference(ctx, expected, bytes.NewReader(b), srcRoot.Digest.String()); err != nil {
+		return err
+	}
+
+	tag := src.Repo().Reference.Reference
+	if err := dst.UpdateIndex(ctx, tag, expected); err != nil {
+		return err
+	}
+
+	message.Infof("Published %s to %s", src.Repo().Reference, dst.Repo().Reference)
```
Noxsios, line 71: should `*Remote` also have a `r.log`? food for thought

src/pkg/oci/common.go
```diff
@@ -25,69 +23,18 @@ const (
...
-type ProgressWriter interface {
-	UpdateTitle(string)
-	io.Writer
-}
-
 // OrasRemote is a wrapper around the Oras remote repository that includes a progress bar for interactive feedback.
 type OrasRemote struct {
 	repo           *remote.Repository
 	root           *Manifest
 	Transport      *helpers.Transport
```
Noxsios, line 30: have ideas for `Transport` I wanna run by you

Noxsios (changes requested):

src/pkg/oci/copier.go
Noxsios: i liked the brrr when I was writing the first iteration of this func, + logging the entire descriptor tells you also size and mediatype, really useful for debugging. Plus this is logged _before_ the layer push which allows you to see if your error is on a really large layer.

src/internal/packager/helm/post-render.go
```diff
@@ -61,8 +61,7 @@ func (r *renderer) Run(renderedManifests *bytes.Buffer) (*bytes.Buffer, error) {
 	}
 	path := filepath.Join(tempDir, "chart.yaml")
 
-	// Write the context to a file for processing
-	if err := utils.WriteFile(path, renderedManifests.Bytes()); err != nil {
+	if err := os.WriteFile(path, renderedManifests.Bytes(), 0600); err != nil {
```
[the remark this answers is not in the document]
AustinAbro321: I was initially a bit torn on doing this, as I didn't want to end up making a const for every permutation we use in zarf. I'm going to make one for 0600, 0644, 0700, and 0755 with the export comments being why we generally set files or directories this way. LMK if that's going a bit overboard though

src/pkg/zoci/copier.go
AustinAbro321: I think it's a fair question, zoci will also be used as a library by uds cli. Let's ask the UDS CLI team how they log and if they want their own user experience around logging.

Maybe we could also do some interesting things with unit testing if we passed in our logger, though that's a whole other conversation

src/pkg/packager/publish.go
```diff
@@ -32,55 +30,21 @@ func (p *Packager) Publish() (err error) {
...
-			return err
-		}
+		parts := strings.Split(srcRemote.Repo().Reference.Repository, "/")
+		packageName := parts[len(parts)-1]
 
-		// ensure cli arch matches package arch
-		if pkg.Build.Architecture != arch {
-			return fmt.Errorf("architecture mismatch (specified: %q, found %q)", arch, pkg.Build.Architecture)
-		}
+		p.cfg.PublishOpts.PackageDestination = p.cfg.PublishOpts.PackageDestination + "/" + packageName
```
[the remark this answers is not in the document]
AustinAbro321: I messed up and put that code in the wrong order as I was rearranging the functions. In general, I'm going to read through code more diligently going forward while refactoring so I don't mess up things like this

src/pkg/zoci/copier.go
```diff
@@ -5,26 +5,38 @@
...
 	if err != nil {
 		return err
 	}
-	layers := helpers.Filter(srcRoot.Layers, include)
-	layers = append(layers, srcRoot.Config)
+
+	// ensure cli arch matches package arch
+	if pkg.Build.Architecture != arch {
+		return fmt.Errorf("architecture mismatch (specified: %q, found %q)", arch, pkg.Build.Architecture)
+	}
```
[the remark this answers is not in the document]
AustinAbro321: I believe we can still get rid of the check here, but of note is that this check is the src `pkg.Build.Architecture`. However I still think this doesn't matter as ultimately the source architecture is also getting grabbed by the config.arch since the refactor. We may need to leave this in for backwards compatibility? I will run some tests and see if I can break it without this block

src/pkg/oci/common.go
AustinAbro321: We discussed making transport private. Instead we just took the Base variable off of transport and composed roundtripper normally. I also made this issue for making progress easier to track in the oci package which could make it easier to make transport private in the future https://github.com/defenseunicorns/zarf/issues/2304

src/pkg/oci/copier.go
AustinAbro321: Yeah I don't mind it especially since it's debug

src/pkg/zoci/fetch.go
```diff
@@ -0,0 +1,32 @@
...
+		return pkg, err
+	}
+	return oci.FetchYAMLFile[types.ZarfPackage](ctx, r.FetchLayer, manifest, layout.ZarfYAML)
+}
+
+// FetchImagesIndex fetches the images/index.json file from the remote repository.
+func (r *Remote) FetchImagesIndex(ctx context.Context) (index *ocispec.Index, err error) {
+	manifest, err := r.FetchRoot(ctx)
+	if err != nil {
+		return index, err
```
lucasrod16, line 29: This would return a `nil` pointer index on error which could lead to a runtime panic in calling code that relies on the returned index object

```suggestion
		return nil, err
```

src/pkg/zoci/pull.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/defenseunicorns/zarf/src/types"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2/content/file"
+)
+
+var (
+	// PackageAlwaysPull is a list of paths that will always be pulled from the remote repository.
+	PackageAlwaysPull = []string{layout.ZarfYAML, layout.Checksums, layout.Signature}
+	// ZarfPackageIndexPath is the path to the index.json file in the OCI package.
+	ZarfPackageIndexPath = filepath.Join("images", "index.json")
```
lucasrod16, line 27:
```suggestion
	ZarfPackageIndexPath = filepath.Join(layout.ImagesDir, layout.IndexJSON)
```

src/pkg/zoci/pull.go
```diff
@@ -0,0 +1,171 @@
...
+	// ZarfPackageImagesBlobsDir is the path to the directory containing the image blobs in the OCI package.
+	ZarfPackageImagesBlobsDir = filepath.Join("images", "blobs", "sha256")
```
lucasrod16, line 29:
```suggestion
	ZarfPackageImagesBlobsDir = filepath.Join(layout.ImagesDir, "blobs", "sha256")
```

src/pkg/zoci/pull.go
```diff
@@ -0,0 +1,171 @@
...
+	// ZarfPackageLayoutPath is the path to the oci-layout file in the OCI package.
+	ZarfPackageLayoutPath = filepath.Join("images", "oci-layout")
```
lucasrod16, line 31:
```suggestion
	ZarfPackageLayoutPath = filepath.Join(layout.ImagesDir, layout.OCILayout)
```

src/pkg/zoci/pull.go
AustinAbro321: I decided to just move these to layout as the same filepath.Join was used there for all of these just not as a variable

src/pkg/utils/helpers/io.go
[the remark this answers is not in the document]
AustinAbro321: I think that's fair. I originally was suggested the idea from @Racer159. I do like that we have comments in the code now explaining why we generally use the different permissions though yeah it's way more verbose I would probably lean towards setting it back to the direct numbers. 
Thoughts after seeing it in code @Racer159 @Noxsios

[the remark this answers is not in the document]
Noxsios: imo i think its fine either way. This way is a little more verbose on usage, but a little more readable in the long run when it comes to understanding which file permissions are being set.  Its not a huge change to refactor later, but it does cause a breaking contract change deleting these `helpers.XYZ` public consts.

src/pkg/zoci/pull.go
```diff
@@ -0,0 +1,165 @@
...
+	doneSaving <- err
+	<-doneSaving
+	return layersToPull, err
+}
+
+// LayersFromRequestedComponents returns the descriptors for the given components from the root manifest.
+// It also retrieves the descriptors for all image layers that are required by the components.
+//
+// It also respects the `required` flag on components, and will retrieve all necessary layers for required components.
+func (r *Remote) LayersFromRequestedComponents(ctx context.Context, requestedComponents []string) (layers []ocispec.Descriptor, err error) {
```
Noxsios, line 81: depending upon which one (`filters` or this PR) gets merged first, will need to keep an eye on this func to ensure regression does not happen

Noxsios (commented):
some small linting things can be found w/ `make lint-go`, but otherwise looks pretty good to me

lucasrod16 (approved):
lgtm

lucasrod16 (approved):

src/pkg/zoci/pull.go
AustinAbro321: Gotcha, plan is to get this PR in first

lucasrod16 (dismissed):

lucasrod16 (approved):
lgtm

src/pkg/zoci/pull.go
```diff
@@ -105,12 +97,11 @@ func (r *Remote) LayersFromRequestedComponents(ctx context.Context, requestedCom
...
 		}
 	}
+	var images []string
 	for _, component := range pkg.Components {
 		// If we requested this component, or it is required, we need to pull its images and tarball
 		if slices.Contains(requestedComponents, component.Name) || component.Required {
-			for _, image := range component.Images {
-				images[image] = true
-			}
+			images = append(images, component.Images...)
```
Racer159, line 104: Wouldn't doing this cause multiple of the same image to show in the list?

Racer159 (changes requested):

src/pkg/zoci/pull.go
AustinAbro321: Yes my mistake, good catch

Racer159 (dismissed):
lgtm

Noxsios (dismissed):

Issue #1128, opened 2022-12-14
Ability to have exclusions and globs and multiple flavors on a component definition


Racer159:
### Is your feature request related to a problem? Please describe.
**As** Ezra **I want** a way to filter out combinations of components in a package with multiple flavors on package create **so that** I can more dynamic packages than single flavors alone.

### Describe the solution you'd like

- **Given** I have a package with components that I would like to filter out
- **When** I add a comma-separated, glob and exclude compatible list under `only.flavor` 

i.e. `flavor: "full,*k3s*"`

- **And** run `zarf package create` with `--flavor`
- **Then** Zarf will match the flavors using the same include / exclude logic it uses for component names

### Describe alternatives you've considered
We could add `--components` or `--excludeComponents` to `zarf p c` but since we already have flavors it would likely be best to expand this existing functionality and it gives package creators more control to tailor their packages for given usecases.

### Additional context
This include / exclude logic is already in Zarf and used for the `--components` key and supports excluding names with a leading `-` and glob patterns like `*`.

AustinAbro321:
Another alternative to consider is to introduce an `only.flavors` list key into the zarf schema and deprecating the original `only.flavor`. "I believe the comma-separated list makes sense to use on the command line with components. However, as a user, I would be confused about why I'm adding a comma-separated list to a string key in yaml. 

The big downside of this would be the breaking change. This could be be someone mitigated by `zarf dev migrate` automatically converting only.flavor to a single item list. We could also keep `only.flavor` deprecated for a while and have `only.flavors` take priority. Still this option may not be worth it depending on how often people want to assign multiple flavors to components. If this is a feature people want 1% of the time common separated is likely the way to go, if it's more like 10-20% I think a list may make sense.

Racer159:
I am not particularly in need of it right now - the idea originated from the implementation of flavors and the component wildcard logic and thinking they should probably be combined / similar - it may be worth taking to the next Zarf community meetup and discussing with @zarf-dev/maintainers (ofc if you have specific examples those would help to have here to help shape the story more)

Racer159:
It is worth noting as well that with composable components you can have a common `zarf.yaml` component that is copied into multiple flavors if a package creator desired (even if no changes were made)

AustinAbro321:
[...]

PR #2237, opened 2024-01-21
Adding feature to exclude optional components

dgershman:
## Description

This adds the ability to exclude optional components while build a package.

## Related Issue

Relates to #1128 

## Type of change

## Checklist before merging


Noxsios:
Related to: https://github.com/defenseunicorns/zarf/pull/2211

dgershman:
> Related to: #2211

@Noxsios do you think it makes sense to rebase (and continue rebasing) from your branch?

Racer159:
Depending on timing we could go either way - now that we have a custom gather space it may be good to discuss options for this too - a few things have changed with `--components` since the issue was filed.

As I see it we could:

1. Proceed as is and create a separate components exclusion flag that is disconnected from the --components logic
2. Incorporate the components logic with `-` support and default support except that an empty components list would create with all components instead of just required and default.

dgershman:
After our discussion in the gather space, decided to take a different route leveraging `flavor`.

PR #2239, opened 2024-01-22
Remove workflow for automatically adding issues to the zarf project

### LINKED ISSUE #2238: Switch to GitHub Project workflows
We should remove https://github.com/defenseunicorns/zarf/blob/main/.github/workflows/auto-assign-to-project.yml to switch to project workflows instead

https://github.com/defenseunicorns/zarf/blob/main/.github/workflows/auto-assign-to-project.yml

### Additional context
This will reduce our overall permissions footprint

YrrepNoj:
## Description
Remove the workflow that would automatically add an issue to the [Zarf Project Board](https://github.com/orgs/defenseunicorns/projects/1). This operation is now handled in a workflow at the project level.

This reduces the overall permissions footprint of the Zarf workflows.

## Related Issue
Fixes #2238


Racer159 (approved):
lgtm

Issue #2240, opened 2024-01-22
Add architecture as a required arg to NewOrasRemote


UncleGedd:
### Description
When using Zarf as a library, calling [NewOrasRemote](https://github.com/defenseunicorns/zarf/blob/cf9acb50e5a2240e6bd2af994e5904cd0f73fd55/src/pkg/oci/common.go#L104) requires passing in a [WithArch](https://github.com/defenseunicorns/zarf/blob/cf9acb50e5a2240e6bd2af994e5904cd0f73fd55/src/pkg/oci/common.go#L93) modifer to set the architecture and the underlying `index.json`'s Platform. Since passing in the mod is effectively "required" for interacting with OCI artifacts that have an `index.json` with a Platform (ie. multi-arch OCI artifacts), recommend we make the ocispec's Platform a required arg to `NewOrasRemote`. With the Platform as an arg, Zarf lib users no longer have to remember to pass in the `WithArch` mod because the compiler will catch the error

UncleGedd:
Noting that this may not belong in Zarf but rather in a new common OCI lib

Racer159:
This was completed in #2241

PR #2241, opened 2024-01-22
chore: update NewOrasRemote to take ocispec.Platform as an argument

decleaver:
## Description

makes the ocispec's Platform a required arg to NewOrasRemote. With the Platform as an arg, Zarf lib users no longer have to remember to pass in the WithArch mod because the compiler will catch the error

## Related Issue

Relates to #2240 

## Type of change

## Checklist before merging


src/pkg/oci/common.go
```diff
@@ -75,38 +75,34 @@ func WithInsecureSkipVerify(insecure bool) Modifier {
...
-func WithTargetPlatform(platform *ocispec.Platform) Modifier {
-	return func(o *OrasRemote) {
-		o.targetPlatform = platform
-	}
-}
-
 // WithSkeletonArch sets the target architecture for the remote to skeleton
-func WithSkeletonArch() Modifier {
-	return WithTargetPlatform(&ocispec.Platform{
+func WithSkeletonArch() ocispec.Platform {
```
Racer159, line 79:
```suggestion
func PlatformForSkeleton() ocispec.Platform {
```

src/pkg/oci/common.go
```diff
@@ -75,38 +75,34 @@ func WithInsecureSkipVerify(insecure bool) Modifier {
...
 		OS:           MultiOS,
 		Architecture: SkeletonArch,
-	})
+	}
 }
 
 // WithArch sets the target architecture for the remote
-func WithArch(arch string) Modifier {
-	return WithTargetPlatform(&ocispec.Platform{
+func WithArch(arch string) ocispec.Platform {
```
Racer159, line 87:
```suggestion
func PlatformForArch(arch string) ocispec.Platform {
```
This function naming would help this be distinct from other uses of `With` in this file.

src/pkg/packager/create_stages.go
```diff
@@ -250,7 +251,7 @@ func (p *Packager) output() error {
 		if err != nil {
 			return err
 		}
-		remote, err := oci.NewOrasRemote(ref)
+		remote, err := oci.NewOrasRemote(ref, ocispec.Platform{})
```
Racer159, line 254: This should set an arch and match the convention - it usually isn't good practice to create an empty struct as a filler.

src/pkg/oci/common.go
```diff
@@ -75,38 +75,34 @@ func WithInsecureSkipVerify(insecure bool) Modifier {
...
-func NewOrasRemote(url string, mods ...Modifier) (*OrasRemote, error) {
+func NewOrasRemote(url string, platform ocispec.Platform, mods ...Modifier) (*OrasRemote, error) {
 	ref, err := registry.ParseReference(strings.TrimPrefix(url, helpers.OCIURLPrefix))
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse OCI reference %q: %w", url, err)
 	}
 	o := &OrasRemote{}
+	if platform.OS != "" || platform.Architecture != "" || platform.OSVersion != "" || len(platform.OSFeatures) != 0 || platform.Variant != "" {
+		o.targetPlatform = &platform
+	}
```
Racer159, line 105:
```suggestion
```
I would rather force the platform to be given

src/pkg/packager/create_stages.go
Racer159: (this will be important in the refactor of Publish once it can no longer dig into build information from a package definition once it is genericized)

src/pkg/packager/create_stages.go
```diff
@@ -652,7 +653,7 @@ func (p *Packager) loadDifferentialData() error {
 
 	// Load the package spec of the package we're using as a 'reference' for the differential build
 	if helpers.IsOCIURL(p.cfg.CreateOpts.DifferentialData.DifferentialPackagePath) {
-		remote, err := oci.NewOrasRemote(p.cfg.CreateOpts.DifferentialData.DifferentialPackagePath)
+		remote, err := oci.NewOrasRemote(p.cfg.CreateOpts.DifferentialData.DifferentialPackagePath, ocispec.Platform{})
```
Racer159, line 656: Same here, we should have arch info

src/pkg/packager/publish.go
```diff
@@ -112,7 +112,7 @@ func (p *Packager) Publish() (err error) {
 		return err
 	}
 
-	remote, err := oci.NewOrasRemote(ref)
+	remote, err := oci.NewOrasRemote(ref, ocispec.Platform{})
```
Racer159, line 115: we should have arch info here

Racer159 (changes requested):

src/pkg/oci/common.go
decleaver: updated

src/pkg/oci/common.go
decleaver: updated

src/pkg/packager/create_stages.go
decleaver: updated to pass in `oci.PlatformForArch(config.GetArch())` instead of an empty platform

src/pkg/oci/common.go
decleaver: checks removed, now just setting the platform

src/pkg/packager/publish.go
decleaver: updated, now passing ing `oci.PlatformForArch(config.GetArch())`

Racer159: looks like this one was missed.

Racer159 (changes requested):

src/pkg/packager/publish.go
decleaver: whoops! Thanks again for the review!

Racer159 (approved):
lgtm

PR #2243, opened 2024-01-23
chore: ADR Multi-Node Kubernetes Clusters on init

dgershman:
## Description

ADR for `zarf init` multi-node.

## Related Issue

N/A

## Type of change

## Checklist before merging


adr/0023-multi-node-kubernetes.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Status
+
+Pending
+
+## Context
+
+Currently today, when allowing `zarf init` to handle cluster creation, Zarf doesn't have ability to automatically or semi-automatically provision itself across multiple nodes.  The idea here would be to allow for horizontal scalability across multiple virtual or physical nodes for site reliability and automatic failover.
+
+The first pass would consider scaling horizontally with the Embedded DB model.  There would be minimal changes on the current k3s.service config.  The change required here would be to include a shared token.  By default, if K3S doesn't receive a token it will auto generate one.  You can reset this on an existing cluster by running `k3s token rotate --new-token=foo`.
```
lucasrod16, line 13: I recommend being more explicit about which datastore option this uses and linking to it

```suggestion
The first pass would consider scaling horizontally with the [High Availability Embedded etcd](https://docs.k3s.io/datastore/ha-embedded) model.  There would be minimal changes on the current k3s.service config.  The change required here would be to include a shared token.  By default, if K3S doesn't receive a token it will auto generate one.  You can reset this on an existing cluster by running `k3s token rotate --new-token=foo`.
```

adr/0023-multi-node-kubernetes.md
```diff
@@ -0,0 +1,75 @@
...
+
+For example:
+```shell
+zarf init --components=k3s,git-server --confirm --set K3S_ARGS=\"--disable=traefik --disable=metrics-server --disable=servicelb --tls-san=1.2.3.4 --token=foo\""
+```
+
+This results in a line as such for example:
+
+```ini
+ExecStart=/usr/sbin/k3s server --write-kubeconfig-mode=700 --write-kubeconfig /root/.kube/config --disable=traefik --disable=metrics-server --disable=servicelb --tls-san=1.2.3.4 --token=foo
```
lucasrod16, line 25: I believe we will need to use the `cluster-init` flag to enable clustering in HA embedded etcd mode:

> To get started, first launch a server node with the cluster-init flag to enable clustering and a token that will be used as a shared secret to join additional servers to the cluster.

https://docs.k3s.io/datastore/ha-embedded
https://docs.k3s.io/cli/server#cluster-options

This is what tells k3s to use the embedded etcd datastore rather than the default SQLite datastore

adr/0023-multi-node-kubernetes.md
lucasrod16, line 1: Based on the k3s documentation, I believe the minimum steps to create a highly available control plane would be as follows:

1. Launch a server node with the `--cluster-init` flag to enable clustering and a token that will be used as a shared secret to join additional servers to the cluster.

1. After launching the first server, join the second and third servers to the cluster using the shared secret and pointing to the first launched server node with the --server flag.
    - The docs state:
    
    > An HA K3s cluster with embedded etcd is composed of:
        Three or more server nodes that will serve the Kubernetes API and run other control plane services, as well as host the embedded etcd datastore.

3. Optionally, add agent nodes to the cluster to run workloads on.

Could you update the ADR to reflect this?

https://docs.k3s.io/datastore/ha-embedded

lucasrod16 (changes requested):
Hi @dgershman,

Thanks for working on this! I read through the k3s documentation to better understand the HA embedded etcd deployment model and have left some comments on the ADR based on that.

Could you clarify the workflow for provisioning the servers? Would Zarf assume the infrastructure is already in place, or would it handle provisioning as well? I think it would be helpful to include a section in the ADR that outlines the end-to-end workflow for users, covering infrastructure setup and deployment.

adr/0023-multi-node-kubernetes.md
dgershman: Yes you are right, I wrote this before testing out multiple `master` nodes.

dgershman:
based off some v1.0 discussion today, closing this in favor of maintaining a community init package and adding a docs page and linking back to this blog article and repo:

- https://radiusmethod.com/edge-ready-highly-available-kubernetes-with-zarf/
- https://github.com/radiusmethod/zarf-init-bigbang-k3s-ha

PR #2249, opened 2024-01-24
Update 5-k9s-dashboard.md

KinzP:
## Description
zarf tools k9s is incorrect & replaced by zarf tools monitor
...

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


Racer159:
`zarf tools k9s` does work but `zarf tools monitor` is the default and is what is in help - we mentioned `k9s` in those places just to name the tool if folks were familiar with it though we could consider changing it if there is confusion.  Was there a specific issue you were seeing with `zarf tools k9s`?
image
(Note the aliases section and that `zarf tools m` also works

Racer159:
Closing after talking with @andrewg-xyz

PR #2256, opened 2024-01-24
Fixes for manifest with symlink

### LINKED ISSUE #2255: Manifest w/Symlink Failures
### Environment
Device and OS: N/A
App version: v0.32.1
Other: N/A

### Steps to reproduce
1. Have a manifest (or chart), that has a symlink in it.  For example P1 Big Bang Keycloak.
2. `zarf package create ...`

### Expected result

That the package builds.

### Actual Result

The package build fails.

Screenshot 2024-01-24 at 2 25 30 PM

### Severity/Priority

High (there isn't a known way to work around this).

### Additional Context

dgershman:
## Description

Fixes the issue with symlinks inside of manifests, similar to that seen in the [Big Bang Keycloak Helm Chart](https://repo1.dso.mil/big-bang/product/packages/keycloak/-/blob/main/development/theme-custom/theme/custom-theme/account/resources?ref_type=heads).

## Related Issue

Fixes #2255

## Type of change

## Checklist before merging


Noxsios (changes requested):
The real issue here is not that symlinks are not getting copied, but that symlinks are not being resolved before the tarball is built.

`CreatePathAndCopy` in `io.go` needs the call to `copy.Copy` changed to add in deep copying / resolving of symlinks.

```go
	// Copy all the source data into the destination location
	if err := copy.Copy(source, destination, copy.Options{
		OnSymlink: func(_ string) copy.SymlinkAction {
			return copy.Deep
		},
	}); err != nil {
		return err
	}
```

dgershman:
Looks like the Windows build is failing on the symlink.

Racer159 (changes requested):
Yeah they are all failing - wanted to run them before commenting to make sure but `../img` has to exist:

```
unable to copy manifest: manifests: lstat ../img: no such file or directory
```

dgershman:
> Yeah they are all failing - wanted to run them before commenting to make sure but `../img` has to exist:
> 
> ```
> unable to copy manifest: manifests: lstat ../img: no such file or directory
> ```

It does exist though.  I can't repro on my local windows VM.  Maybe I'm missing something?

src/test/packages/34-manifest-with-symlink/manifests/resources/img
```diff
@@ -0,0 +1 @@
+../img
```
Racer159, line 1: I don't see a file called `img` in the parent directory though in git (which would be `manifests`)

dgershman: I guess because there was no file in it.  I'm pushing up a text file, which is probably better to make sure that recursing works.

dgershman:
> > Yeah they are all failing - wanted to run them before commenting to make sure but `../img` has to exist:
> > ```
> > unable to copy manifest: manifests: lstat ../img: no such file or directory
> > ```
> 
> It does exist though. I can't repro on my local windows VM. Maybe I'm missing something?

It was on my local, my bad.

dgershman:
I'm confused as to why this failed again...

dgershman:
> I'm confused as to why this failed again...

@Racer159 do you have any idea why this might be failing now?

Racer159:
Yep after looking into it more it has to do with how the DeepCopy stuff works:
image
image

The symlink comes back as a relative directory to `../img` when the program isn't chdir'ed into the directory for that to resolve correctly.

That said I don't really think we should be using DeepCopy since it technically is manipulative of the package structure and if there were other things we used `CreatePathAndCopy` for that could be a side effect that we do not want.

The _real_ issue is that we don't actually set the tarball header correctly - `header, err := tar.FileInfoHeader(info, "")` has a second parameter that is supposed to take in a link if this is a symlink file which we simply do not do.  This code should be changed to:

```
return filepath.Walk(dirPath, func(filePath string, info os.FileInfo, err error) error {
		if err != nil {
			return err
		}

		link := ""
		if info.Mode().Type() == os.ModeSymlink {
			link, err = os.Readlink(filePath)
			if err != nil {
				return fmt.Errorf("error reading symlink: %w", err)
			}
		}

		// Create a new header
		header, err := tar.FileInfoHeader(info, link)
		if err != nil {
			return fmt.Errorf("error creating tar header: %w", err)
		}
```

dgershman:
> Yep after looking into it more it has to do with how the DeepCopy stuff works:
> 
> image image
> 
> The symlink comes back as a relative directory to `../img` when the program isn't chdir'ed into the directory for that to resolve correctly.
> 
> That said I don't really think we should be using DeepCopy since it technically is manipulative of the package structure and if there were other things we used `CreatePathAndCopy` for that could be a side effect that we do not want.
> 
> The _real_ issue is that we don't actually set the tarball header correctly - `header, err := tar.FileInfoHeader(info, "")` has a second parameter that is supposed to take in a link if this is a symlink file which we simply do not do. This code should be changed to:
> 
> ```
> 	return filepath.Walk(dirPath, func(filePath string, info os.FileInfo, err error) error {
> 		if err != nil {
> 			return err
> 		}
> 
> 		link := ""
> 		if info.Mode().Type() == os.ModeSymlink {
> 			link, err = os.Readlink(filePath)
> 			if err != nil {
> 				return fmt.Errorf("error reading symlink: %w", err)
> 			}
> 		}
> 
> 		// Create a new header
> 		header, err := tar.FileInfoHeader(info, link)
> 		if err != nil {
> 			return fmt.Errorf("error creating tar header: %w", err)
> 		}
> ```
> 
> (between lines 444 and 450)

This change doesn't fully resolve the issue because the symlink is a directory, not a file, which is the original issue.  This winds up failing on the copy.

```golang
if _, err := io.Copy(tw, file); err != nil {
  return fmt.Errorf("error writing file to tarball: %w", err)
}
```

src/pkg/utils/io.go
```diff
@@ -249,7 +249,11 @@ func CreatePathAndCopy(source string, destination string) error {
 	}
 
 	// Copy all the source data into the destination location
-	if err := copy.Copy(source, destination); err != nil {
+	if err := copy.Copy(source, destination, copy.Options{
+		OnSymlink: func(_ string) copy.SymlinkAction {
+			return copy.Deep
+		},
+	}); err != nil {
```
Racer159, line 256: Did you change this back?

Racer159: It shouldn't enter into that since a symlink isn't a regular file and is header-only in a tarball

dgershman: I'm working on a change now.

Racer159: You will run into this on a deploy, but that is more to do with what Zarf expects manifests are (the test would likely need to be refactored into a local chart)
image

Racer159 (approved):
lgtm

lucasrod16 (approved):
lgtm

Noxsios (approved):

PR #2259, opened 2024-01-25
fix: add the `--key` flag to the init cmd

### LINKED ISSUE #2248: Zarf Init with Signed OCI Package Cannot Be Validated
### Environment
Device and OS: AMD64 / Linux
App version: v0.32.1
Other: N/A

### Steps to reproduce
1. Use a signed init package from an OCI-compliant registry.
2. Run `zarf package pull oci://some-repo/image:tag`
3. Run `zarf init --confirm` 

### Expected result

That you can use a signed init package.

### Actual Result

An error message appears saying to supply a key, except there is no `-k, --key` flag for `zarf init`.  

Screenshot 2024-01-23 at 10 05 10 PM

### Severity/Priority

Low

### Additional Context
We can work around this by validating the signature on `zarf package pull`, and then using the `--insecure` flag on `zarf init`.

You can also work around this by using `zarf package deploy oci://some-repo/image:tag -k`, as long as you don't need any of the flags specific to `zarf init`.

dgershman:
## Description

This adds the `-key` flag to `zarf init` command.

## Related Issue

Fixes #2248 

## Type of change

## Checklist before merging


src/cmd/initialize.go
```diff
@@ -220,5 +220,7 @@ func init() {
```
Racer159, line 220: You'll have to run `make docs-and-schema` to update the CLI docs with the new key flag.

src/test/e2e/35_custom_init_package_test.go
```diff
@@ -0,0 +1,47 @@
...
+	// Test that we don't get an error when we remember to provide the public key
+	stdOut, stdErr, err = e2e.Zarf("package", "inspect", pkgName, publicKeyFlag)
+	require.NoError(t, err, stdOut, stdErr)
+	require.Contains(t, stdErr, "Verified OK")
+
+	/* Test operations during package deploy */
+	// Test that we get an error when trying to deploy a package without providing the public key
+	stdOut, stdErr, err = e2e.Zarf("init", "--confirm")
+	require.Error(t, err, stdOut, stdErr)
+	require.Contains(t, stdErr, "unable to load the package: package is signed but no key was provided - add a key with the --key flag or use the --insecure flag and run the command again")
```
Racer159, line 41: Will run the tests, but this _may_ wrap in the terminal and you might need to use the `StripMessageFormatting` function in `e2e`

Racer159 (changes requested):
One potential issue, and there needs to be a docs update.

src/test/e2e/35_custom_init_package_test.go
Racer159: If the tests do pass then this is fine though

dgershman: The tests passed locally.

src/test/e2e/35_custom_init_package_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+func TestCustomInit(t *testing.T) {
+	t.Log("E2E: Custom Init Package")
+	e2e.SetupWithCluster(t)
+	buildPath := filepath.Join("src", "test", "packages", "35-custom-init-package")
+	pkgName := fmt.Sprintf("build/zarf-init-%s-%s.tar.zst", e2e.Arch, e2e.GetZarfVersion(t))
+	privateKeyFlag := "--signing-key=src/test/packages/zarf-test.prv-key"
+	publicKeyFlag := "--key=src/test/packages/zarf-test.pub"
+
+	stdOut, stdErr, err := e2e.Zarf("package", "create", buildPath, privateKeyFlag, "--confirm", "-o=build")
```
Racer159, line 23: Putting this in the `build` directory will cause a failure with the `k3s` test because it will overwrite the init package that was used to deploy it causing this error: https://github.com/defenseunicorns/zarf/actions/runs/7700470953/job/20984462726?pr=2259#step:7:15068

(this should just be placed in the working directory instead (Zarf will prioritize that file over the one that is next to it's binary)

Racer159 (changes requested):

src/test/e2e/35_custom_init_package_test.go
dgershman: Ok, I think I understand what you are saying.  I pushed an update, let me know if this is what you meant.

Makefile
```diff
@@ -171,7 +171,7 @@ test-e2e: test-e2e-without-cluster test-e2e-with-cluster  ## Run all of the core
 .PHONY: test-e2e-with-cluster
 test-e2e-with-cluster: build-examples ## Run all of the core Zarf CLI E2E tests that DO require a cluster (builds any deps that aren't present)
 	@test -s ./build/zarf-init-$(ARCH)-$(CLI_VERSION).tar.zst || $(MAKE) init-package
-	cd src/test/e2e && go test ./main_test.go ./[2-9]*.go -failfast -v -timeout 35m
+	cd src/test/e2e && go test ./main_test.go ./35*.go -failfast -v -timeout 35m
```
Racer159, line 174: This needs to still be `[2-9]`

Racer159 (changes requested):

Racer159 (approved):
lgtm

Noxsios (approved):

Issue #2261, opened 2024-01-25
Zarf package with no checksums causes runtime panic


mjnagel:
### Environment
Device and OS: MBP, macOS
App version: 0.32.2
Kubernetes distro being used: N/A

### Steps to reproduce
1. Create a zarf.yaml like the below:
    ```yaml
    kind: ZarfPackageConfig
    metadata:
      name: test

    components: []
    ```
2. `zarf p create --skip-sbom --confirm` (note the skip sbom)
3. `zarf p deploy zarf-package-test-*.tar.zst --confirm`
4. Enjoy the panic.

Note that this is reproducible with a component with actions as well such as the below, which is how I actually hit this:
```yaml
kind: ZarfPackageConfig
metadata:
  name: test

components:
  - name: hello
    required: true
    actions:
      onCreate:
        before:
          - cmd: echo hello
```

### Expected result

I would expect no errors and this to effectively do nothing (throwing a warning about no components would be expected with the first package).

### Actual Result

```console
  ⠋  Validating full package checksums                                                                                                                                                                                          
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/defenseunicorns/zarf/src/pkg/packager/sources.ValidatePackageIntegrity.func1({0x0, 0x0})
        /home/runner/work/zarf/zarf/src/pkg/packager/sources/validate.go:88 +0x2f0
github.com/defenseunicorns/zarf/src/pkg/packager/sources.lineByLine({0x14001670dc0?, 0x14001683e60?}, 0x1400155f0d8)
        /home/runner/work/zarf/zarf/src/pkg/packager/sources/validate.go:167 +0xf4
github.com/defenseunicorns/zarf/src/pkg/packager/sources.ValidatePackageIntegrity(0x14001630600, {0x14001670f00, 0x40}, 0x0)
        /home/runner/work/zarf/zarf/src/pkg/packager/sources/validate.go:85 +0x144
github.com/defenseunicorns/zarf/src/pkg/packager/sources.(*TarballSource).LoadPackage(0x14000fecdc0, 0x14001630600, 0x1)
        /home/runner/work/zarf/zarf/src/pkg/packager/sources/tarball.go:98 +0x234
github.com/defenseunicorns/zarf/src/pkg/packager.(*Packager).Deploy(0x14000933680)
        /home/runner/work/zarf/zarf/src/pkg/packager/deploy.go:43 +0x48
[...]

PR #2262, opened 2024-01-26
fix: handle panic during checksum validation

mjnagel:
## Description

Handles a condition where execution would panic when a package contained no checksums.

## Related Issue

Fixes https://github.com/defenseunicorns/zarf/issues/2261

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/packager/sources/validate.go
+++ b/src/pkg/packager/sources/validate.go
@@ -83,10 +83,17 @@ func ValidatePackageIntegrity(loaded *layout.PackagePaths, aggregateChecksum str
 	checkedMap[loaded.Signature] = true
 
 	err = lineByLine(checksumPath, func(line string) error {
+		// If the line is empty (i.e. there is no checksum) simply skip it - this can result from a package with no images/components
 		if line == "" {
 			return nil
 		}
+
 		split := strings.Split(line, " ")
+		// If the line is not splitable into two pieces the file is likely corrupted
+		if len(split) != 2 {
+			return fmt.Errorf("invalid checksum line: %s", line)
+		}
+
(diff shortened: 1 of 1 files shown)


src/pkg/packager/sources/validate.go
```diff
@@ -83,6 +83,9 @@ func ValidatePackageIntegrity(loaded *layout.PackagePaths, aggregateChecksum str
 	checkedMap[loaded.Signature] = true
 
 	err = lineByLine(checksumPath, func(line string) error {
+		if line == "" {
+			return nil
+		}
 		split := strings.Split(line, " ")
 		sha := split[0]
 		rel := split[1]
```
Racer159, line 91:
```suggestion
		// If the line is empty (i.e. there is no checksum) simply skip it - this can result from a package with no images/components
		if line == "" {
			return nil
		}

		split := strings.Split(line, " ")
		// If the line is not splitable into two pieces the file is likely corrupted
		if len(split) != 2 {
			return fmt.Errorf("invalid checksum line: %s", line)
		}

		sha := split[0]
		rel := split[1]
```

Racer159: This would be a better more-total fix in case someone has a bad package (since the string could be non-empy but not properly splitable by spaces)

Racer159 (changes requested):

src/pkg/packager/sources/validate.go
mjnagel: Thanks, this makes sense and is a good check to add 🎉

Racer159 (approved):
lgtm

PR #2264, opened 2024-01-28
fix: use memory friendly file split logic

### LINKED ISSUE #2231: Large memory usage when using `--max-package-size`
Hello! We are creating a zarf package of bigbang which results in roughly 12-13GB of images that are packaged. We were noticing that when we use the `--max-package-size` flag when building the package, when zarf writes the tarball to disk, it uses roughly 12-13GB of memory to do so.

While 12-13GB is doable, trying to build packages as part of a pipeline is rough as building multiple packages from different pipelines can cause memory issues for the host or requires multiple runners, really beefy runners, or a combination of the two

### Steps to reproduce
1. Build a zarf package with a sizable amount of images using `--max-package-size=4500`
2. Wait for zarf to indicate it is writing the tarball to disk
3. Watch memory usage of zarf process as it writes tarball to disk

### Expected result
Zarf builds the package without consuming too much memory

### Actual Result
Zarf consumes roughly 2x the size of the tarball as memory

### Additional Context
[Digging through the code ](https://github.com/defenseunicorns/zarf/blob/main/src/pkg/utils/io.go#L295C9-L295C9)reveals that zarf loads the tarball into memory then starts copying chunks to a new array which would explain why we see a 2x size of the tarball as memory.

I can provide screenshots of `top` showing memory usage of the zarf process while it handles the tarball if requested

daniel-palmer-gu:
## Description

When splitting a tarball using `--max-package-size`, zarf will load the entire tarball into memory and copy it to new slices before writing the new files to disk resulting in memory usage up to 2x of the tarball. This can result in large memory usage when the package tarball is large itself.

This PR updates the logic to read the original tarball in chunks of 16 MiB at a time thereby fixing the memory issue.

## Testing

Will be breaking this down into a few sections:

1. Test setup
1. Test off of main
1. Test off of PR

### Test setup

We will be reusing the `TestMultiPartPackage` test and updating `src\test\packages\05-multi-part\zarf.yaml` to the following to test a larger package size and setting `--max-package-size=500`:

```yaml
kind: ZarfPackageConfig
metadata:
  name: multi-part
components:
  - name: big-ol-file
    required: true
    description: test larger package around 2GB
    images:
      - atlassian/confluence-server:8.6.0-jdk17
      - atlassian/confluence-server:8.5.1-jdk17
```

We will expect the test to fail as the number of expected split files will no longer match the original test, however we will be able to see the memory usage while the test runs.

We will capture memory usage using `top | grep zarf`.

### Test off of main

Before we look at the results from the code changes, lets first look at the results from main for comparison.

[zarf-2024-01-27-23-31-37-3566031216.log](https://github.com/defenseunicorns/zarf/files/14075059/zarf-2024-01-27-23-31-37-3566031216.log)

image

When the package tarball is being split, memory tops out at 2.1g

### Test off of PR

[zarf-2024-01-27-23-42-37-480658840.log](https://github.com/defenseunicorns/zarf/files/14075078/zarf-2024-01-27-23-42-37-480658840.log)

image

When the package tarball is being split, memory remains steady at under 250m

## Related Issue

Fixes #2231 

## Type of change

## Checklist before merging


src/pkg/utils/io.go
```diff
@@ -321,6 +322,127 @@ func SplitFile(path string, chunkSizeBytes int) (chunks [][]byte, sha256sum stri
...
+		bytesRead, err := file.Read(buf)
+
+		if err != nil {
+			if err == io.EOF {
+				// At end of file, break out of loop
+				break
+			} else {
+				return fileNames, "", err
+			}
+		}
```
Noxsios, line 382:
```suggestion
		if err != nil {
			if err == io.EOF {
				// At end of file, break out of loop
				break
			}
			return fileNames, "", err
		}
```

src/pkg/utils/io.go
```diff
@@ -321,6 +322,127 @@ func SplitFile(path string, chunkSizeBytes int) (chunks [][]byte, sha256sum stri
...
+			chunkBytesRemaining = chunkSizeBytes - (bufferSize - chunkBytesRemaining)
+		} else {
+			_, err := chunkFile.Write(buf[0:bytesRead])
+			if err != nil {
+				return fileNames, "", err
+			}
+			chunkBytesRemaining = chunkBytesRemaining - bytesRead
+		}
+	}
+	file.Close()
```
Noxsios, line 420: prob move to a `defer file.Close()` when `file` is initialized so that this function can't be leaky w/ file handles

Noxsios (changes requested):

src/pkg/utils/io.go
daniel-palmer-gu: Do you know if you can remove an open file? Never tried before, but assumed it was required to close the file first.

Noxsios: *nix systems can do it, but Windows has issues.

Only really a problem when you try to access that file again (like in this small change I PRed to `syft` https://github.com/anchore/syft/pull/1668).

Surprisingly static code analysis can't catch this, so I try to always `defer file.Close()` as soon as I open the file. _Then_ later if I want to explicitly close the file I can still do `file.Close` and just not handle the error (as you are doing now).

daniel-palmer-gu: Thanks! Makes sense. I've added that in

src/pkg/packager/common.go
```diff
@@ -342,41 +341,13 @@ func (p *Packager) archivePackage(destinationTarball string) error {
...
-		_ = os.RemoveAll(destinationTarball)
-
-		// Marshal the data into a json file.
-		jsonData, err := json.Marshal(types.ZarfSplitPackageData{
-			Count:     len(chunks),
-			Bytes:     fi.Size(),
-			Sha256Sum: sha256sum,
-		})
+		spinner.Updatef("Package is larger than %dMB, splitting into multiple files", p.cfg.CreateOpts.MaxPackageSizeMB)
+		_, _, err := utils.SplitFile(destinationTarball, chunkSize)
```
Racer159, line 348: Since the return variables are not used the function signature might as well not return them and only return `err`.  Also since this reduces messaging IMO it would be good to implement the progressbar interface for this function where `utils.SplitFile` takes in an `io.Writer` called `progressWriter` and this caller instantiates a `message.NewProgressBar` taking in `fi.Size()` as the total number of bytes (when written to the progress bar will progress forward giving more indication of how long the process is going to take than a simple spinner - before it would show the file it was currently on in the spinner and a progressbar would bring that show of progress back and more).

Racer159: If you have questions on how to implement the progressbar interface let us know and we can help.

For reference you can see how this progressbar is used with io.Copy:

https://github.com/defenseunicorns/zarf/blob/23618d95c49f177d022fb45c206530b6dece54cc/src/pkg/oci/copier.go#L47
https://github.com/defenseunicorns/zarf/blob/23618d95c49f177d022fb45c206530b6dece54cc/src/pkg/oci/copier.go#L100

Racer159 (changes requested):

src/pkg/packager/common.go
daniel-palmer-gu: Sure I can look into adding that in when I get some free time. I'm hoping to go a step further if we are open to using archiver v4. It looks like we could use it to split the file as it gets archived and save some time

daniel-palmer-gu: However, if we go that route we likely won't be able to do a progressbar as we won't know the compressed file size ahead of time.

daniel-palmer-gu: @Racer159, I've cleaned up the returns and have implemented the progressbar. I opted for ending the spinner after the initial archive is written, displaying a note that we are going to split the archive, and then doing a progressbar for the split.

Although this is after it has finished and you won't be able to actually see the progressbar, it will look something like this
image

Racer159: Sorry for the delay - I think that works for the progressbar - it will go quick on smaller packages but looks decent on bigger ones where `--max-package-size` is more likely to be used.

image
image

src/pkg/utils/io.go
```diff
@@ -327,6 +328,140 @@ func SplitFile(path string, chunkSizeBytes int) (chunks [][]byte, sha256sum stri
...
+			}
+			chunkBytesRemaining = chunkBytesRemaining - bytesRead
+		}
+
+		// update progress bar
+		progressBar.Add(bufferSize)
+		title := fmt.Sprintf("[%d/%d] MB bytes written", progressBar.GetCurrent()/1000/1000, fileSize/1000/1000)
+		progressBar.UpdateTitle(title)
+	}
+	progressBar.Successf("Package split across %d files", len(fileNames))
```
Racer159, line 438:
```suggestion
	progressBar.Successf("Package split across %d files", len(fileNames)+1)
```
Tiny change since the number of files will actually be one higher with the metadata file (on larger numbers people aren't likely to notice but on 2 or 3 they might.

It may also be better to move this to after the metadata file being written as well so it calls success once the whole splitting operation is done.

Racer159 (changes requested):
Sorry for the delay, one tiny change on the messaging but should be GTG after that

src/pkg/utils/io.go
daniel-palmer-gu: That's a good point. I've moved the progressBar.Successf down a few lines after the metadata file is written. Since the metadata file gets added to `fileNames`, I chose to keep `len(fileNames)` as it should be an accurate count now

Noxsios (approved):

Racer159 (approved):
lgtm

PR #2270, opened 2024-01-30
chore: updates k9s

UncleGedd:
## Description
Updates k9s to `v0.31.7` to solve a dependency issue with Helm `v3.14.0`. 

More context, I'm attempting to use Zarf lib code `v0.32.2` alongside Helm `v3.14.0`. This results in several compiler errors that complain about types in `popeye`. Upgrading `popeye` alone doesn't solve the issue because that results in type errors with `k9s`. Since `k9s` brings in `popeye` we can just upgrade `k9s` and all is right with the world (see https://github.com/defenseunicorns/uds-cli/pull/327)

## Type of change


Noxsios (changes requested):
if possible, I would prefer this PR _only_ touch deps and things directly related to the upgrade of `k9s`. As it stands this also changes `syft` + creates breaking changes w/ our embedded `syft` cli.

UncleGedd:
> if possible, I would prefer this PR _only_ touch deps and things directly related to the upgrade of `k9s`. As it stands this also changes `syft` + creates breaking changes w/ our embedded `syft` cli.

Agreed! I tried this a couple of different ways. I only touched the line in `go.mod` relating to `k9s`, but running `go mod tidy` updates all of those other deps as well. What am I missing?

UncleGedd:
I went ahead and added a `replace` block to keep the original version of Syft, let know what you guys think

Racer159:
closing this PR - there was a bit more interplay and wanted to get everything in together - thanks for the PR though @UncleGedd !

PR #2271, opened 2024-01-30
chore: `k8s`, and `cluster` partial refactor 

### LINKED ISSUE #2222: Refactor Get Images Functions and Clarify `k8s` vs `cluster`
We should move the `GetImagesWithNodes` and `GetAllImages` functions from the `k8s` package into the `cluster` package and consolidate their function signatures into a single public function (such as `cluster.GetImagesFromAvailableNodes`) that can get the images from nodes with available resources.

We should also remove unused receiver functions from the `k8s` package to slim down it's API surface given those functions are not tested by Zarf.

https://github.com/defenseunicorns/zarf/blob/main/src/pkg/k8s/images.go#L22
https://github.com/defenseunicorns/zarf/blob/main/src/pkg/k8s/images.go#L49

### Additional context
This came out of discussions from https://github.com/defenseunicorns/zarf/pull/2220

chrishorton:
## Description

Refactor `k8s` and `cluster` packages to consolidate `GetImagesWithNodes` and `GetAllImages` into one function `GetAvailableImagesFromNodes`.

Move utility functions from `images.go` to `prepare.go` where they are used.

Delete `images.go`

## Related Issue

Fixes #2222

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/injector.go
+++ b/src/pkg/cluster/injector.go
@@ -202,7 +202,7 @@ func (c *Cluster) createPayloadConfigmaps(seedImagesDir, tarPath string, spinner
 		return configMaps, "", err
 	}
 
-	chunks, sha256sum, err := utils.SplitFile(tarPath, payloadChunkSize)
+	chunks, sha256sum, err := utils.ReadFileByChunks(tarPath, payloadChunkSize)
 	if err != nil {
 		return configMaps, "", err
 	}
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/AlecAivazis/survey/v2 v2.3.7
 	github.com/Masterminds/semver/v3 v3.2.1
 	github.com/alecthomas/jsonschema v0.0.0-20220216202328-9eeeec9d044b
-	github.com/anchore/clio v0.0.0-20240131202212-9eba61247448
+	github.com/anchore/clio v0.0.0-20240202120828-3ef5b3b40ea3
 	github.com/anchore/stereoscope v0.0.1
 	github.com/anchore/syft v0.100.0
 	github.com/derailed/k9s v0.31.7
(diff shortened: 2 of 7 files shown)


chrishorton:
Should `GetImagesFromAvailableNodes` and `ImageNodeMap` be public?

src/pkg/cluster/injector.go
```diff
@@ -426,3 +429,72 @@ func (c *Cluster) buildInjectionPod(node, image string, payloadConfigmaps []stri
 
 	return pod, nil
 }
+
+// GetImagesFromAvailableNodes checks for images on schedulable nodes within a cluster and returns
+func (c *Cluster) GetImagesFromAvailableNodes(timeoutDuration time.Duration) (ImageNodeMap, error) {
```
Racer159, line 434:
```suggestion
func (c *Cluster) getImagesAndNodesForInjection(timeoutDuration time.Duration) (imageNodeMap, error) {
```

src/pkg/cluster/injector.go
```diff
@@ -53,15 +56,15 @@ func (c *Cluster) StartInjectionMadness(tmpDir string, imagesDir string, injecto
 	}
 
 	var err error
-	var images k8s.ImageNodeMap
+	var images ImageNodeMap
```
Racer159, line 59:
```suggestion
	var images imageNodeMap
```

src/pkg/cluster/injector.go
```diff
@@ -35,6 +35,9 @@ var (
 	injectorLimitMemory     = resource.MustParse("256Mi")
 )
 
+// ImageNodeMap is a map of image/node pairs.
+type ImageNodeMap map[string][]string
```
Racer159, line 39:
```suggestion
type imageNodeMap map[string][]string
```

Racer159 (changes requested):
Overall looks good! And yes I think they should be private for now to simplify the public interface.  Thinking more, I also suggested a more descriptive name for the function since it is so injector specific

src/pkg/cluster/injector.go
```diff
@@ -53,15 +56,15 @@ func (c *Cluster) StartInjectionMadness(tmpDir string, imagesDir string, injecto
...
+	var images imageNodeMap
 	var payloadConfigmaps []string
 	var sha256sum string
 	var seedImages []transform.Image
 
 	// Get all the images from the cluster
 	timeout := 5 * time.Minute
 	spinner.Updatef("Getting the list of existing cluster images (%s timeout)", timeout.String())
-	if images, err = c.GetAllImages(timeout, injectorRequestedCPU, injectorRequestedMemory); err != nil {
+	if images, err = c.GetImagesFromAvailableNodes(timeout); err != nil {
```
Racer159, line 67:
```suggestion
	if images, err = c.getImagesAndNodesForInjection(timeout); err != nil {
```
This also has to change

Racer159 (changes requested):

Racer159 (approved):
lgtm

Noxsios (approved):

PR #2282, opened 2024-02-01
feat: set variables on find images

AustinAbro321:
## Description

Currently when you run `zarf prepare find-images` with a package that has variables in it, the variables don't get set, even to their defaults. This can break the helm chart, or make us unable to find images. This PR introduces a `--deploy-set` flag. Additionally we take in the default values to the chart.

## Type of change

## Checklist before merging


src/cmd/dev.go
```diff
@@ -265,9 +268,16 @@ func init() {
...
 	devFindImagesCmd.Flags().StringToStringVar(&pkgConfig.CreateOpts.SetVariables, "set", v.GetStringMapString(common.VPkgCreateSet), lang.CmdDevFlagSet)
+
+	devFindImagesCmd.Flags().MarkDeprecated("set", "This field is replaced by create-set")
+	devFindImagesCmd.Flags().MarkHidden("set")
+	devFindImagesCmd.Flags().StringToStringVar(&pkgConfig.CreateOpts.SetVariables, "create-set", v.GetStringMapString(common.VPkgCreateSet), lang.CmdDevFlagSet)
+	devFindImagesCmd.Flags().StringToStringVar(&pkgConfig.PkgOpts.SetVariables, "deploy-set", v.GetStringMapString(common.VPkgDeploySet), lang.CmdPackageDeployFlagSet)
 	// allow for the override of the default helm KubeVersion
 	devFindImagesCmd.Flags().StringVar(&pkgConfig.FindImagesOpts.KubeVersionOverride, "kube-version", "", lang.CmdDevFlagKubeVersion)
 
+	devFindImagesCmd.Flags().StringVar(&pkgConfig.FindImagesOpts.RegistryOverride, "registry", "127.0.0.1:31999", lang.CmdDevFlagRegistry)
```
Racer159, line 279:
```suggestion
	devFindImagesCmd.Flags().StringVar(&pkgConfig.FindImagesOpts.RegistryOverride, "registry-url", "127.0.0.1:31999", lang.CmdDevFlagRegistry)
```

This would likely be better to align it with the other flags in Zarf.

src/config/lang/english.go
```diff
@@ -618,6 +619,11 @@ const (
 	AgentErrUnableTransform        = "unable to transform the provided request; see zarf http proxy logs for more details"
 )
 
+// src/internal/packager
+const (
+	PkgErrsetVariableMap = "unable to set the active variables: %w"
```
Racer159, line 624:
```suggestion
	PkgErrSetVariableMap = "unable to set the active variables: %w"
```

Racer159: Also will leave it up to @Noxsios as to whether this should be here - the lang pattern we have should be reevaluated.

src/internal/packager/helm/chart.go
```diff
@@ -139,12 +139,20 @@ func (h *Helm) InstallOrUpgradeChart() (types.ConnectStrings, string, error) {
...
 	message.Debugf("helm.TemplateChart()")
 	spinner := message.NewProgressSpinner("Templating helm chart %s", h.chart.Name)
-	defer spinner.Stop()
+	defer func() {
+		if err != nil {
+			spinner.Stop()
+			spinner.Errorf(err, "Problem rendering the helm template for %s: %s", h.chart.Name, err.Error())
+		} else {
+			spinner.Success()
+		}
```
Racer159, line 151: why have this complex logic?

src/internal/packager/helm/common.go
```diff
@@ -148,7 +148,17 @@ func WithKubeVersion(kubeVersion string) Modifier {
...
+		h.cfg = cfg
+	}
+}
+
 // StandardName generates a predictable full path for a helm chart for Zarf.
 func StandardName(destination string, chart types.ZarfChart) string {
 	return filepath.Join(destination, chart.Name+"-"+chart.Version)
 }
+
+func StandardValuesName(destination string, chart types.ZarfChart, idx int) string {
```
Racer159, line 162: lint

src/internal/packager/helm/common.go
```diff
@@ -148,7 +148,17 @@ func WithKubeVersion(kubeVersion string) Modifier {
 	}
 }
 
+func WithPackageConfig(cfg *types.PackagerConfig) Modifier {
```
Racer159, line 151: lint

src/types/runtime.go
```diff
@@ -55,6 +55,7 @@ type ZarfInspectOptions struct {
 type ZarfFindImagesOptions struct {
 	RepoHelmChartPath   string `json:"repoHelmChartPath" jsonschema:"description=Path to the helm chart directory"`
 	KubeVersionOverride string `json:"kubeVersionOverride" jsonschema:"description=Kubernetes version to use for the helm chart"`
+	RegistryOverride    string `json:"registryOverride" jsonschema:"description=Manual override for ###ZARF_REGISTRY###`
```
Racer159, line 58:
```suggestion
	RegistryURL    string `json:"registryURL" jsonschema:"description=Manual override for ###ZARF_REGISTRY###`
```
Registry Override means something else in Zarf (--registry-override)

src/internal/packager/helm/prepare-render.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	values.SetState(&types.ZarfState{})
+
+	return &variableRenderer{
+		Helm:   h,
+		values: *values,
+	}, nil
+}
+
+func (r *variableRenderer) Run(renderedManifests *bytes.Buffer) (*bytes.Buffer, error) {
```
Racer159, line 37: This is a decent amount of copy-pasta - would likely be better to split the normal renderer around line 169 to check for a config boolean as to whether it should interact with the cluster.

Racer159: The cluster logic can be pulled into a helper function and then wrapped in an if relatively easily without the logic needing to be here.

src/test/packages/00-find-images/zarf.yaml
Racer159, line 1: We should probably consider just using the zarf init package for this since it already exists and generally I like testing real things when they would still be decent test cases over contrived examples (since it ensures that our code works on that real thing and it will always stay up to date with how Zarf works for other reasons).

Also this test case should likely have a more descriptive name since it isn't the only 00-find-images test package.

src/test/e2e/00_use_cli_test.go
```diff
@@ -79,6 +79,10 @@ func TestUseCLI(t *testing.T) {
 		stdOut, stdErr, err = e2e.Zarf("prepare", "find-images", "src/test/packages/00-helm-annotations")
 		require.NoError(t, err, stdOut, stdErr)
 		require.Contains(t, stdOut, "registry1.dso.mil/ironbank/opensource/istio/pilot:1.17.2", "The pilot image should be found by Zarf")
+
+		stdOut, stdErr, err = e2e.Zarf("prepare", "find-images", "src/test/packages/00-helm-annotations")
+		require.NoError(t, err, stdOut, stdErr)
+		require.Contains(t, stdOut, "registry1.dso.mil/ironbank/opensource/istio/pilot:1.17.2", "The pilot image should be found by Zarf")
```
Racer159, line 85: duplicated code?

Racer159 (changes requested):

src/test/packages/00-find-images/zarf.yaml
AustinAbro321: The reason I made this test instead of just using the init package is because the zarf init only test the values file case and I wanted to test the manifest and helm cases as well. They all end up as separate paths in code

src/internal/packager/helm/chart.go
AustinAbro321: The spinner wasn't stopping on errors. I could have put spinner.stop on each of the error conditions, but I thought this might be cleaner.

AustinAbro321: I believe I was playing around with the spinner back when I first worked on this. I needed to do a more diligent self review on this one

src/internal/packager/helm/prepare-render.go
AustinAbro321: Separated into two functions as I believe that reads better

src/test/packages/00-find-images/zarf.yaml
Racer159: You could use `examples/manifests` for that second part - that would test the kustomize branch too.

Racer159 (changes requested):

src/test/packages/00-find-images/zarf.yaml
AustinAbro321: I ended up using examples/manifests. I feel better about not testing the helm path now that are on the same code path as the original helm renderer. We still technically aren't testing anything kustomize related since I didn't change the kustomize example to be from a local directory. I don't mind this since we don't evaluate vars in kustomize, only in the manifests after they're built.

src/pkg/packager/prepare.go
```diff
@@ -120,36 +125,55 @@ func (p *Packager) FindImages() (imgMap map[string][]string, err error) {
...
+		values, err := template.Generate(p.cfg)
+		if err != nil {
+			return nil, fmt.Errorf("unable to generate template values")
+		}
 
+		values.SetState(&types.ZarfState{
+			RegistryInfo: types.RegistryInfo{
+				Address: p.cfg.FindImagesOpts.RegistryURL,
+				// Adding a fake secret here so if a chart expects the registry secret, like the init package, it won't fail
+				Secret: "fake-zarf-secret",
```
AustinAbro321, line 137: This feels kinda hacky, but I wanted find-images to work on the init package. I thought about making secret an option on find-images like registryurl but decided against as it doesn't actually pertain to the image like the former

examples/manifests/zarf.yaml
```diff
@@ -3,6 +3,10 @@ metadata:
   name: manifests
   version: 0.0.1
 
+variables:
+  - name: HTTPD_VERSION
+    default: "3.18"
```
Racer159, line 8: This isn't a great example since this version might not be in the package - it would be better to have a variable for a different value (potentially something that enables/disables a dependency in a chart)

src/pkg/packager/prepare.go
Racer159: The `fillInEmpty*` functions in `cluster` could be useful here - they actually do not need the cluster struct and might work as receivers on types.ZarfState.

Racer159 (changes requested):

src/pkg/packager/prepare.go
AustinAbro321: Changed to be receivers on the respective objects

src/internal/packager/template/template.go
```diff
@@ -89,7 +94,7 @@ func (values *Values) GetVariables(component types.ZarfComponent) (templateMap m
 			"STORAGE_CLASS": values.config.State.StorageClass,
 
 			// Registry info
-			"REGISTRY":           values.registry,
+			"REGISTRY":           regInfo.Address,
```
Racer159, line 97: You should drop `registry` from this entirely.

Racer159 (changes requested):
small hanging chad

src/internal/packager/helm/post-render.go
```diff
@@ -42,15 +40,21 @@ func (h *Helm) newRenderer() (*renderer, error) {
 		return nil, err
 	}
 
+	// TODO (@austinabro321) this should be cleaned up after https://github.com/defenseunicorns/zarf/pull/2276 gets merged
```
Noxsios, line 43: linking https://github.com/defenseunicorns/zarf/pull/2276

src/types/k8s.go
```diff
@@ -91,6 +109,43 @@ type GitServerInfo struct {
 	InternalServer bool   `json:"internalServer" jsonschema:"description=Indicates if we are using a git server that Zarf is directly managing"`
 }
 
+// FillInEmptyValues sets every necessary value that's currently empty to a reasonable default
+func (gs GitServerInfo) FillInEmptyValues() (GitServerInfo, error) {
```
Noxsios, line 113: make these pointer receivers and return just an `err` for these `FillInEmptyValues` operations imo

```go
func (gs *GitServerInfo) FillInEmptyValues() error {
```

src/pkg/packager/dev.go
```diff
@@ -62,7 +63,7 @@ func (p *Packager) DevDeploy() error {
 
 	// Set variables and prompt if --confirm is not set
 	if err := p.setVariableMapInConfig(); err != nil {
-		return fmt.Errorf("unable to set the active variables: %w", err)
+		return fmt.Errorf(lang.PkgErrSetVariableMap, err)
```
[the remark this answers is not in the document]
Noxsios: additionally, the added context of `unable to set the active variables: %w` is not affected by its usage in `dev.go` (or anywhere else), so its extraneous sugar. This context should happen inside of `p.setVariableMapInConfig`, where the actual error, and error context occurs

Noxsios (changes requested):

src/types/k8s.go
Noxsios: might be worth researching struct default value population patterns in Go

AustinAbro321: Yeah I was 50/50 on doing that. I can see how the pointer receiver may read more naturally, though I thought since we generally use this on initopts objects that we are going to reassign anyway, the usage of value receivers was a bit nicer

AustinAbro321: Updated to use the pointer receiver method.

Not able to find much for relevant patterns given that for this we're generally starting with an object with some values already filled in by the user. For what it's worth, chatGPT said that the pointer receiver way of doing things was more idiomatic for go.

src/test/common.go
```diff
@@ -72,15 +73,12 @@ func (e2e *ZarfE2ETest) Zarf(args ...string) (string, string, error) {
...
-			// - they use two drives which filepath.Rel cannot cope with.
-			relCacheDir, _ := filepath.Abs(".cache-location")
-			args = append(args, "--zarf-cache", relCacheDir)
-			defer os.RemoveAll(relCacheDir)
-		}
-
+	if !slices.Contains(args, "--zarf-cache") && !slices.Contains(args, "tools") && os.Getenv("CI") == "true" {
+		// We make the cache dir relative to the working directory to make it work on the Windows Runners
+		// - they use two drives which filepath.Rel cannot cope with.
+		relCacheDir, _ := filepath.Abs(fmt.Sprintf(".%d-cache-location", rand.Intn(10000000)))
```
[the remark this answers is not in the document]
eddiezane: Or maybe `e2e.Zarf` can take the `testing.T` and use the built in mktmp utils?

lucasrod16 (dismissed):
lgtm

src/pkg/packager/prepare.go
```diff
@@ -133,19 +134,22 @@ func (p *Packager) FindImages() (imgMap map[string][]string, err error) {
...
-		artifactServer := types.ArtifactServerInfo{}.FillInEmptyValues()
+		artifactServer := &types.ArtifactServerInfo{}
+		artifactServer.FillInEmptyValues()
 		values.SetState(&types.ZarfState{
-			RegistryInfo:   registryInfo,
-			GitServer:      gitServer,
-			ArtifactServer: artifactServer})
+			RegistryInfo:   *registryInfo,
+			GitServer:      *gitServer,
+			ArtifactServer: *artifactServer})
```
Noxsios, line 152: ik im the one who told you to do this, but please change these back to not be pointers 😵‍💫

Noxsios (changes requested):
one final tweak (my fault), then its gtg

src/pkg/packager/prepare.go
Racer159: :point_down: :point_left: :point_right: :point_up:

Racer159 (dismissed):
approving to dismiss req'd changes, agree with @Noxsios on pointers

src/test/e2e/13_find_images_test.go
```diff
@@ -93,9 +57,32 @@ func TestFindImages(t *testing.T) {
...
+	t.Run("zarf test find images --why w/  manifests success", func(t *testing.T) {
+		t.Parallel()
+
+		testPackagePath := filepath.Join("examples", "manifests")
+
+		stdOut, _, err := e2e.Zarf("dev", "find-images", testPackagePath, "--why", "httpd:alpine3.18")
+		require.NoError(t, err)
+		require.Contains(t, stdOut, "component: httpd-local")
+		require.Contains(t, stdOut, "manifest: simple-httpd-deployment")
+		require.Contains(t, stdOut, "image: httpd:alpine3.18")
```
Racer159, line 84: Commenting on this - this is not a particularly great test now - this could be repeated many times or other issues could arise - we can talk about how to handle this, but it is a regression in coverage

Racer159: Might be worth a `tech-debt` issue

Noxsios (approved):

PR #2298, opened 2024-02-09
feat(ci): included dependency review action

naveensrinivasan:
## Description

- Included dependency review action https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review#dependency-review-enforcement

## Type of change

## Checklist before merging


.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,14 @@
+name: 'Dependency Review'
+on: [pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
```
Racer159, line 8:
```suggestion
  validate:
```
Changing the job name to validate will make it match our branch protection rules for the repo - we should also consider if/how we might de-conflict this dep review with `.github/workflows/scan-cves.yml` which does similar things (not license checks) but more manually.  We could leave both running to see which output we like better though over time (`grype` has gotten better but had a lot of false positives when we first started).

Racer159 (changes requested):

Racer159 (approved):
lgtm

Racer159:
Code looks good but some commits are not signed - will need to be resigned to merge
image

Racer159 (changes requested):
There is still this unsigned commit in there: 
image

naveensrinivasan:
> There is still this unsigned commit in there: image

Can you please check now?

Noxsios:
Approving. Will need to circle back in a few weeks/months to evaluate how much of a value add / noise this generates.

.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,14 @@
+name: 'Dependency Review'
+on: [pull_request]
```
Noxsios, line 2:
```suggestion
name: Dependency Review
on: pull_request
```

tiny consistency tweak

.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,14 @@
...
+permissions:
+  contents: read
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+      - name: 'Dependency Review'
```
Noxsios, line 13:
```suggestion
      - name: Checkout
        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
      - name: Dependency Review
```

Noxsios (changes requested):

Racer159 (approved):
lgtm

Noxsios (approved):

PR #2301, opened 2024-02-10
fix: improve cmd failure messaging

### LINKED ISSUE #2299: Actions erroneously return a timeout message when the command did not timeout.
### Environment
Device and OS: System76 drap8
App version: v0.32.3
Kubernetes distro being used: any
Other: n/a

### Steps to reproduce
1. Include an action command (like `exit 1`) that will fail right away without configuring maxTotalSeconds or maxRetries.
2. Run the associated Zarf command (i.e. `zarf package create .` for the below `zarf.yaml`):

```yaml
kind: ZarfPackageConfig
metadata:
  name: actions-error-msg

components:
  - name: actions-error-msg
    actions:
      onCreate:
        before:
          - cmd: exit 1
```

### Expected result

An error message like `Failed to create package: unable to add component "actions-error-msg": unable to run component before action: command "exit 1" failed` displays.

### Actual Result

The error message `Failed to create package: unable to add component "actions-error-msg": unable to run component before action: command "exit 1" timed out after 0 seconds` displays.

image

### Severity/Priority

Medium

### Additional Context
This causes confusion because people may be misled into believing that Zarf has timed out their command when it was the command that failed instead.

docandrew:
…s with no retries

## Description

Add check to see whether action failure was actually due to timeout or not. Currently Zarf reports an error of "timed out after 0 seconds" when a `cmd` within an action fails (with no retries) even if no timeout was set.

## Related Issue

Fixes #2299

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/packager/actions.go
+++ b/src/pkg/packager/actions.go
@@ -161,14 +161,14 @@ retryCmd:
 	select {
 	case <-timeout:
 		// If we reached this point, the timeout was reached or command failed with no retries.
-		if cfg.MaxTotalSeconds < 1 && cfg.MaxRetries < 1 {
-			return fmt.Errorf("command \"%s\" failed (no retries)", cmdEscaped)
+		if cfg.MaxTotalSeconds < 1 {
+			return fmt.Errorf("command %q failed after %d retries", cmdEscaped, cfg.MaxRetries)
 		} else {
-			return fmt.Errorf("command \"%s\" timed out after %d seconds", cmdEscaped, cfg.MaxTotalSeconds)
+			return fmt.Errorf("command %q timed out after %d seconds", cmdEscaped, cfg.MaxTotalSeconds)
 		}
 	default:
 		// If we reached this point, the retry limit was reached.
-		return fmt.Errorf("command \"%s\" failed after %d retries", cmdEscaped, cfg.MaxRetries)
+		return fmt.Errorf("command %q failed after %d retries", cmdEscaped, cfg.MaxRetries)
 	}
 }
 
--- a/src/pkg/packager/actions.go
+++ b/src/pkg/packager/actions.go
@@ -160,9 +160,12 @@ retryCmd:
 
 	select {
 	case <-timeout:
-		// If we reached this point, the timeout was reached.
-		return fmt.Errorf("command \"%s\" timed out after %d seconds", cmdEscaped, cfg.MaxTotalSeconds)
-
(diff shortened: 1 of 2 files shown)


src/pkg/packager/actions.go
```diff
@@ -160,9 +160,12 @@ retryCmd:
 
 	select {
 	case <-timeout:
-		// If we reached this point, the timeout was reached.
-		return fmt.Errorf("command \"%s\" timed out after %d seconds", cmdEscaped, cfg.MaxTotalSeconds)
-
+		// If we reached this point, the timeout was reached or command failed with no retries.
+		if cfg.MaxTotalSeconds < 1 && cfg.MaxRetries < 1 {
+			return fmt.Errorf("command \"%s\" failed (no retries)", cmdEscaped)
```
Racer159, line 165:
```suggestion
		if cfg.MaxTotalSeconds < 1 {
			return fmt.Errorf("command %q failed after %d retries", cmdEscaped, cfg.MaxRetries)
```

src/pkg/packager/actions.go
```diff
@@ -160,9 +160,12 @@ retryCmd:
...
+		} else {
+			return fmt.Errorf("command \"%s\" timed out after %d seconds", cmdEscaped, cfg.MaxTotalSeconds)
```
Racer159, line 167:
```suggestion
			return fmt.Errorf("command %q timed out after %d seconds", cmdEscaped, cfg.MaxTotalSeconds)
```
All of these could be %q to simplify them

src/pkg/packager/actions.go
Racer159: Realized the logic I put in the issue was wrong - you can fall into this if you retried without a timeout being set...

Racer159 (changes requested):

src/pkg/packager/actions.go
docandrew: Made the suggested change, but just want to confirm with you - in that situation would it fall-through to the default branch of the select?

Racer159: Yes because the timeout would be 0 and the timeout channel would still be reached.

Current code with a maxTotalSeconds and maxRetries > 0:
image

Current code with a maxRetries > 0 but a maxTotalSeconds = 0:
image

Your code with a maxRetries > 0 but a maxTotalSeconds = 0:
image

Racer159 (approved):
lgtm

PR #2303, opened 2024-02-12
fix: allow usage of local image archives again

### LINKED ISSUE #2181: Using local image archives no longer works
### Environment
Device and OS: Linux 
App version: v0.31.0
Other:

### Steps to reproduce
1.  Create an image archive locally with `docker save busybox -o busybox.tar`
2.  Create a zarf.yaml that references the archive in the `images:` list for a component.

```
kind: ZarfPackageConfig
metadata:
  name: example
  description: Test component
  version: 1.0.0

components:
  - name: zebra
    description: "This is a zebra and they have stripes"
    images:
      - "/home/rob/busybox.tar"
```

3.  `zarf package create . --confirm`

### Expected result
A package to  be created that includes the referenced image archive.

### Actual Result
Error indicating the package create has failed.  
```
     ERROR:  Failed to create package: failed to create ref for image /home/rob/busybox.tar: invalid reference
             format
```

### Severity/Priority
Not critical

### Additional Context
Originally added in https://github.com/defenseunicorns/zarf/pull/1173,

waveywaves:
## Description

a check in create_stages during package creation tries to parses references for the images provided for components in the zarf.yaml config. By default the reference parsing assumes that the source of the image is that of a container registry. If the image is a tarball instead, the reference parser errors out.

The changes in this commit check if the image reference is that of a tarball and skips the reference check if it is so allowing the image to be loaded by the cli.

## Related Issue

Fixes #2181

## Type of change

## Checklist before merging


Racer159 (changes requested):
This is a good start (and likely all tests will pass currently) but this won't entirely resolve the issue - `zarf package create` will work fine, but `zarf package deploy` will fail.  See the below example (after this patch was applied):

```yaml
kind: ZarfPackageConfig
metadata:
  name: dos-games
  description: Simple example to load classic DOS games into K8s in the airgap
  version: 1.0.0

components:
  - name: baseline
    required: true
    manifests:
      - name: multi-games
        namespace: dos-games
        files:
          - manifests/deployment.yaml
          - manifests/service.yaml
    images:
      - image.tar
    actions:
      onDeploy:
        after:
          - wait:
              cluster:
                kind: deployment
                name: game
                namespace: dos-games
                condition: available
```

(The image was generated with the following:)

```
docker pull defenseunicorns/zarf-game:multi-tile-dark
docker save defenseunicorns/zarf-game:multi-tile-dark > examples/dos-games/image.tar
```

`zarf package create`:
image

`zarf package deploy`:
image

(note that the SBOMs say 0 images / that there is no SBOM - if an image was pulled properly this would not be showing - the image should pull on create _and_ be properly referenced to push into the cluster on deploy)

Racer159:
Would be good to add something like the above as a test case under `src/test/e2e`/`src/test/packages` as well

waveywaves:
@Racer159 I have updated the code to properly pull the image on create, and archives the package as required, upon checking inside the zarf package tar I am also able to see the layers for the images whereas before it was empty

image

On the other hand while I trying to deploy the zarf package the image is referenced as the `zarf-game.tar` I understand that this reference needs to be changed. Below is the package which zarf is trying to deploy. This is the one that contains the local image.

image

I get the following error on attempting a deployment. One error which stands out is the login error to the registry. 

image

I understand that the image is pushed to the Zarf registry in the airgapped cluster before the package is deployed. I am currently trying to figure out the mutation for the image reference so that it is pushed correctly and to the cluster and inspecting the error regarding registry login.

waveywaves:
@Racer159 The PR is updated with a e2e test now, can we enable the CI on this PR ?

waveywaves:
working through the tests

src/test/packages/36-tarball-image-references/zarf.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - name: baseline
+    required: true
+    manifests:
+      - name: multi-games
+        namespace: tarball-dos-games
+        files:
+          - manifests/deployment.yaml
+          - manifests/service.yaml
+    images:
+      - zarf-game.tar
```
AustinAbro321, line 15: I believe we will not want to commit this tar file to git history. One way to get around this is to use the `exec.CmdWithPrint` function within the test to run a docker command to save the image. Make sure the test also cleans up the image. 

@Racer159 Can you validate that this would be a good approach. Any other ideas ?

AustinAbro321 (changes requested):

src/internal/packager/images/push.go
```diff
@@ -31,6 +31,7 @@ func (i *ImageConfig) PushToZarfRegistry() error {
 	var totalSize int64
 	// Build an image list from the references
 	for _, refInfo := range i.ImageList {
+		message.Debug("Loading image", refInfo)
```
Racer159, line 34: This is an uneeded debug statement
```suggestion
```

src/test/packages/36-tarball-image-references/zarf.yaml
Racer159: Correct (it isn't in git history here because we `.gitignore` all tar files) - the best way would actually likely be a `-cmd: ./zarf tools registry pull defenseunicorns/zarf-game:multi-tile-dark zarf-game.tar` in an onCreate before action: https://docs.zarf.dev/docs/create-a-zarf-package/component-actions#additional-action-use-cases that way the package works on it's own outside of tests too.

src/pkg/transform/image.go
```diff
@@ -63,39 +63,53 @@ func ImageTransformHostWithoutChecksum(targetHost, srcReference string) (string,
...
-	if named, ok := ref.(reference.Named); ok {
-		out.Name = named.Name()
-		out.Path = reference.Path(named)
-		out.Host = reference.Domain(named)
-		out.Reference = ref.String()
+	if IsTarball(srcReference) {
+		out.Host = "docker.io"
+		out.Name = srcReference
+		out.Path = fmt.Sprintf("%s/%s/%s", out.Host, "library", strings.TrimSuffix(srcReference, ".tar"))
+		out.Reference = srcReference
```
Racer159, line 71: Editing the reference like this won't result in a proper image reference in the end - while the package will create,  the `org.opencontainers.image.base.name` remains `"zarf-game.tar"` inside `images/index.json` within the zarf package.
image

This is loaded on deploy successfully but then when pushed converts into `docker.io/library/zarf-game` which does not match the `defenseunicorns/zarf-game:multi-tile-dark` the tarball was made from.  This means that the Deployment will never pull the image and will be in an image pull backoff state:
image

Racer159: What needs to happen is that the image reference in the tarball needs to be parsed from it's manifest.json:
image
image
(you can see this if you vim into it, and it is accessible when the tarball is loaded with crane)

Racer159: For this, the `zarf.yaml` image reference may need to be modified to turn `zarf-game.tar` into `defenseunicorns/zarf-game:multi-tile-dark` after pulling it

Racer159 (changes requested):

src/test/packages/36-tarball-image-references/zarf.yaml
Racer159:
```yaml
    actions:
      onCreate:
        before:
          - cmd: ./zarf tools registry pull defenseunicorns/zarf-game:multi-tile-dark zarf-game.tar
```

waveywaves:
ok got it, 
considering that we won't have the .tar image because of .gitignore, we are using https://github.com/defenseunicorns/zarf/pull/2303#discussion_r1496853113 to pull the image before the `package create` command runs so that the image will be available in the CI

Racer159:
> ok got it, considering that we won't have the .tar image because of .gitignore, we are using [#2303 (comment)](https://github.com/defenseunicorns/zarf/pull/2303#discussion_r1496853113) to pull the image before the `package create` command runs so that the image will be available in the CI
> 
> working through the other notes

Yep, the onCreate before action would make the tarball just in time before it was needed for the image pull.

waveywaves:
@Racer159 ready to review

I changed the name and path in the reference because of which the deploy stopped working after pushing the latest commits, now the deploy works well again

waveywaves:
seems like the windows binary test is flaking, can we run that test again ?

Racer159 (changes requested):

src/pkg/transform/image.go
```diff
@@ -63,39 +63,52 @@ func ImageTransformHostWithoutChecksum(targetHost, srcReference string) (string,
...
-	if named, ok := ref.(reference.Named); ok {
-		out.Name = named.Name()
-		out.Path = reference.Path(named)
-		out.Host = reference.Domain(named)
-		out.Reference = ref.String()
+	if IsTarball(srcReference) {
+		out.Host = "docker.io"
+		out.Name = strings.TrimSuffix(srcReference, ".tar")
+		out.Path = fmt.Sprintf("%s/%s/%s", out.Host, "library", out.Name)
+		out.Reference = srcReference
```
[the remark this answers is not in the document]
waveywaves: shall make the changes 
I was able to deploy the package though created from this logic 🤔 not sure how

src/pkg/transform/image.go
```diff
@@ -63,39 +63,52 @@ func ImageTransformHostWithoutChecksum(targetHost, srcReference string) (string,
...
-		out.Reference += ":latest"
+		if !IsTarball(srcReference) {
+			out.Reference += ":latest"
+		}
 	}
 
 	return out, nil
 }
+
+func IsTarball(srcReference string) bool {
```
Racer159, line 112: lint (exported functions required comments) - see the GitHub actions check warning in the `Files` tab.

src/pkg/transform/image.go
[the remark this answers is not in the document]
Racer159: image
image
image
image

the reason is that the cluster caches the image and this same image is used in earlier tests (hence the passes in CI as well).  On a fresh cluster / registry it will fail with an ImagePullBackoff.

Note `127.0.0.1:31999/defenseunicorns/zarf-game:multi-tile-dark-zarf-805192805` differs from `127.0.0.1:31999/docker.io/library/zarf-game:latest-zarf-27383200`

Racer159 (changes requested):

src/pkg/transform/image.go
waveywaves: thanks for pointing it out, i'll check it

waveywaves:
@Racer159 I have updated the logic to get the image information for the image manifest in the tarball

waveywaves:
package create and deploy tests are also passing locally on a newly created and initiated minikube zarf instance, ptal

waveywaves:
@Racer159 I am running this test locally which works well. The test requires the image to be present as a tarball locally which is being done by the onCreate - before action. This works well locally but seems like the image is not accessible to the packager in the CI.

schristoff:
Hey @waveywaves  - amazing work on this pull request. Since it is pretty stale I am going to close it, but if you decide to open a new PR you can still reference the work here :)

PR #2306, opened 2024-02-14
fix: allow host+subpath as the source registry for registry-override

### LINKED ISSUE #2135: Allow host+subpath as the source registry for registry-override
Currently, the --registry-override will respect the subpath when going from a host to a host+subpath but not the other way around.
(so --registry-override docker.io=mydomain.com/subpath works but --registry-override docker.io/subpath=mydomain.com doesn't)

The ask here is to extend the functionality with support for:

- --registry-override docker.io/subpath=mydomain.com
- --registry-override docker.io/subpath=mydomain.com/anothersubpat

### Additional context
Slack thread https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1699559557724489?thread_ts=1687158762.361229&cid=C03B6BJAUJ3

waveywaves:
## Description

Instead of looking for refInfo.Host in the override map loop through the keys and values in i.RegistryOverrides, check if the refInfo.Reference begins with an override key and, if it does, replace that override text with the override value and set it back to actualSrc.

Do not use ImageTransformHostWithoutChecksum since we already have the parsed ref and all the info we need to do the replacement.

## Related Issue

Fixes #2135 

## Type of change

## Checklist before merging


src/internal/packager/images/pull.go
```diff
@@ -83,15 +83,15 @@ func (i *ImageConfig) PullAll() ([]ImgInfo, error) {
...
 			}
 
-			actualSrc := refInfo.Reference
-			if overrideHost, present := i.RegistryOverrides[refInfo.Host]; present {
-				var err error
-				actualSrc, err = transform.ImageTransformHostWithoutChecksum(overrideHost, refInfo.Reference)
-				if err != nil {
-					metadataImageConcurrency.ErrorChan <- fmt.Errorf("failed to swap override host %s for %s: %w", overrideHost, refInfo.Reference, err)
-					return
+			actualSrc := ""
```
AustinAbro321, line 86: We don't know if actualSrc is going to have to change so we should keep it as refInfo.Reference. Otherwise we will also error in the non registry override case.

For this reason we should also remove the `if actualSrc == ""` check

AustinAbro321 (changes requested):

src/internal/packager/images/pull.go
waveywaves: what is the non registry override case ?

AustinAbro321: The non registry override case is the default case. The only time the registry is overridden is when the registry on the image matches one of the registries sent through the --registry-overrides flag. 

For example, if I have an image at ghcr.io/defenseunicorns/zarf/agent unless I sent into my package create --registry-override `ghcr.io=mydomain.com` `actualSrc` would have stayed "" and the code would error.

AustinAbro321:
```suggestion
			actualSrc := refInfo.Reference
```
actualSrc needs to stay as refInfo.Reference as well otherwise we will pull an empty string when there is no override

waveywaves: got it, without the registry override the actualSrc value should be the default reference to the image. When the overriding is happening only then actualSrc should be updated. I have updated the PR. Shall squash and sign the commit.

waveywaves:
@AustinAbro321 signed off and squashed all the commits

AustinAbro321:
This looks good to me. Will need approval from @Racer159 @lucasrod16 @Noxsios

AustinAbro321:
@waveywaves You will need to sign your commits. You can follow this guide https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits. You should see verified next to the commit hash in the UI when it's signed successfully.

waveywaves:
@AustinAbro321 done

Racer159 (dismissed):
lgtm!

(only thing that could be useful would be a test (~line 59 in `src/test/e2e/25_helm_test.go`), though we can backfill that from the Zarf team side if needed.  Will need a second approval from @Noxsios / @lucasrod16 as well.)

lucasrod16 (approved):
lgtm

src/test/e2e/25_helm_test.go
```diff
@@ -55,7 +55,23 @@ func testHelmChartsExample(t *testing.T) {
...
+	// expect to fail as ghcr.io is overriden and the expected final image doesn't exist but the override works well based on the error message in the output
+	stdOut, stdErr, err = e2e.Zarf("package", "create", "examples/helm-charts", "-o", "build", "--registry-override", "ghcr.io=docker.io/noway", "--tmpdir", tmpdir, "--confirm")
+	require.Error(t, err, stdOut, stdErr)
+	require.Contains(t, string(stdErr), "docker.io/noway/stefanprodan/podinfo:6.4.0")
+
+	// Create a test package (with a registry override (host+subpath to host) to test that as well)
+	// works same as the above failing test
+	stdOut, stdErr, err = e2e.Zarf("package", "create", "examples/helm-charts", "-o", "build", "--registry-override", "ghcr.io/stefanprodan=docker.io", "--tmpdir", tmpdir, "--confirm")
+	require.Error(t, err, stdOut, stdErr)
+	require.Contains(t, string(stdErr), "docker.io/podinfo:6.4.0")
```
Racer159, line 72:
```suggestion
	stdOut, stdErr, err = e2e.Zarf("package", "create", "examples/helm-charts", "-o", "build", "--registry-override", "ghcr.io=localhost:555/noway", "--tmpdir", tmpdir, "--confirm")
	require.Error(t, err, stdOut, stdErr)
	require.Contains(t, string(stdErr), "localhost:555/noway/stefanprodan/podinfo:6.4.0")

	// Create a test package (with a registry override (host+subpath to host) to test that as well)
	// works same as the above failing test
	stdOut, stdErr, err = e2e.Zarf("package", "create", "examples/helm-charts", "-o", "build", "--registry-override", "ghcr.io/stefanprodan=localhost:555", "--tmpdir", tmpdir, "--confirm")
	require.Error(t, err, stdOut, stdErr)
	require.Contains(t, string(stdErr), "localhost:555/podinfo:6.4.0")
```

Racer159: Would be better to point these to localhost since theoretically `noway` could be a valid docker org in dockerhub someday (which could make this brittle/_potentially_ insecure) https://hub.docker.com/search?q=noway

Racer159 (changes requested):

src/test/e2e/25_helm_test.go
waveywaves: makes sense, something I will keep in mind
I have updated the PR

waveywaves:
failing tests updated, can we restart the tests ?

Racer159 (approved):
lgtm!

src/test/e2e/25_helm_test.go
```diff
@@ -55,7 +55,23 @@ func testHelmChartsExample(t *testing.T) {
...
-	// Create the package (with a registry override to test that as well)
+	// Create a test package (with a registry override (host+subpath to host+subpath) to test that as well)
+	stdOut, stdErr, err = e2e.Zarf("package", "create", "examples/helm-charts", "-o", "build", "--registry-override", "ghcr.io/stefanprodan=docker.io/stefanprodan", "--tmpdir", tmpdir, "--confirm")
+	require.NoError(t, err, stdOut, stdErr)
+
+	// Create a test package (with a registry override (host to host+subpath) to test that as well)
+	// expect to fail as ghcr.io is overriden and the expected final image doesn't exist but the override works well based on the error message in the output
+	stdOut, stdErr, err = e2e.Zarf("package", "create", "examples/helm-charts", "-o", "build", "--registry-override", "ghcr.io=localhost:555/noway", "--tmpdir", tmpdir, "--confirm")
+	require.Error(t, err, stdOut, stdErr)
+	require.Contains(t, string(stdErr), "localhost:555")
```
AustinAbro321, line 66:
```suggestion
	require.Contains(t, string(stdErr), "localhost:555/noway")
```
This should check for localhost:555/noway to make sure the /noway path is really getting set

AustinAbro321 (changes requested):

AustinAbro321 (approved):

Racer159 (approved):
lgtm

PR #2309, opened 2024-02-16
feat: add `--why` flag for `zarf dev find-images`

### LINKED ISSUE #2272: Add a `--why` to `zarf dev find-images` to show the chart and resource that generated the image reference
**As** Ezra **I want** a way to find out how Zarf came to find an image reference **so that** I can more easily track down where this reference came from in my manifests or helm charts.

- **Given** I have a `zarf.yaml` that returns images from `zarf dev find-images`
- **When** I run `zarf dev find-images <path-to-zarf-yaml> --why <image-ref>`
- **Then** Zarf will not return the full image list
- **And** will instead return the name of the first chart and resource that that reference was found within.

i.e.

```bash
zarf dev find-images examples/dos-games/ --why defenseunicorns/zarf-game:multi-tile-dark
```

should return:

**component**: baseline
**manifest**: multi-games
**resource**:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: game
spec:
  selector:
    matchLabels:
      app: game
  template:
    metadata:
      labels:
        app: game
    spec:
      containers:
        - name: multi-game
          image: "defenseunicorns/zarf-game:multi-tile-dark"
          ports:
            - name: http
              containerPort: 8000
              protocol: TCP
          resources:
            requests:
              memory: "32Mi"
              cpu: "50m"
            limits:
              memory: "128Mi"
              cpu: "250m"
```

We could try to dig into the values files ourselves but this would only work for charts and may not be useful in all cases (it also assumes the images are not hardcoded)

### Additional context
Zarf should do its best to highlight the image reference in the templated manifest by changing its color or bolding it.

waveywaves:
## Description

Add a `--why` flag for `zarf dev find-images`

## Related Issue

Fixes #2272

## Type of change

## Checklist before merging


waveywaves:
Partial (relevant) test output
```
$ build/zarf-mac-apple dev find-images examples/dos-games --why defenseunicorns/zarf-game:multi-tile-dark --tmpdir /var/folders/v7/yqyq_d6x2996hfnnwvgs5f080000gn/T/zarf-1881051164

 NOTE  Using config file /Users/vibhavbobade/go/src/github.com/defenseunicorns/zarf/zarf-config.toml

 NOTE  Saving log file to
       /var/folders/v7/yqyq_d6x2996hfnnwvgs5f080000gn/T/zarf-2024-02-20-20-44-58-2844959790.log

 NOTE  Using build directory examples/dos-games
  ✔  Looking for images in component "baseline" across 3 resources              
  ✔  Looking up cosign artifacts for discovered images (0/1)                    
component: baseline
manifests: multi-games
resources: 

apiVersion: apps/v1
kind: Deployment
metadata:
  name: game
spec:
  selector:
    matchLabels:
      app: game
  template:
    metadata:
      labels:
        app: game
    spec:
      containers:
      - image: defenseunicorns/zarf-game:multi-tile-dark
        name: multi-game
        ports:
        - containerPort: 8000
          name: http
          protocol: TCP
        resources:
          limits:
            cpu: 250m
            memory: 128Mi
          requests:
            cpu: 50m
            memory: 32Mi
```

src/pkg/packager/prepare.go
lucasrod16, line 1: This can be simplified a bit by writing a helper function and calling it when we split the YAML into separate resources when iterating over charts and manifests. For example:

```go
func (p *Packager) findWhyResources(resources []*unstructured.Unstructured, componentName, resourceName string) ([]string, error) {
    // resourceName is either the chart name or manifest name
    // check if the resource contains the --why image
    // print the YAML manifest to the console
}
```

And then call it:

```go
for idx, f := range manifest.Files {
    // existing logic to process manifests...
    
    // call the function
}
```

```go
   // existing logic to process charts...
   
   // call the function
}
```

src/pkg/packager/prepare.go
```diff
@@ -200,9 +215,15 @@ func (p *Packager) FindImages() (imgMap map[string][]string, err error) {
...
-			if matchedImages, maybeImages, err = p.processUnstructuredImages(resource, matchedImages, maybeImages); err != nil {
+			var whyImageResourcesPartial []*unstructured.Unstructured
+			if matchedImages, maybeImages, whyImageResourcesPartial, err = p.processUnstructuredImages(resource, matchedImages, maybeImages, whyImage); err != nil {
 				message.WarnErrf(err, "Problem processing K8s resource %s", resource.GetName())
 			}
+			updatedResources := append(whyImageResources[component.Name].Resources, whyImageResourcesPartial...)
+			whyImageResources[component.Name] = manifestResourceRelation{
+				Manifests: component.Manifests,
+				Resources: updatedResources,
+			}
```
Racer159, line 226: Doing this here will match up references for Zarf `manifests` but not Zarf `charts` (which are also templated in find-images).

Running `zarf dev find-images examples/helm-charts/ --why curlimages/curl:7.69.0` results in many resources being assigned to the same empty string manifest when they really belong to different charts in the component:

image

Racer159: Could have made that clearer in the ACs but the above should result in something like (note the swap from `manifest` to `chart`):

```
component: demo-helm-charts
chart: podinfo-oci
resources: 

apiVersion: v1
kind: Pod
metadata:
  annotations:
    appmesh.k8s.aws/sidecarInjectorWebhook: disabled
    helm.sh/hook: test-success
    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
    linkerd.io/inject: disabled
    sidecar.istio.io/inject: "false"
  labels:
    app.kubernetes.io/name: podinfo-oci
    app.kubernetes.io/version: 6.4.0
    helm.sh/chart: podinfo-6.4.0
  name: podinfo-oci-service-test-xaab9
spec:
  containers:
  - command:
    - sh
    - -c
    - |
      curl -s ${PODINFO_SVC}/api/info | grep version
    env:
    - name: PODINFO_SVC
      value: podinfo-oci.podinfo-from-oci:9898
    image: curlimages/curl:7.69.0
    name: curl
  restartPolicy: Never

component: demo-helm-charts
chart: podinfo-git
resources: 

apiVersion: v1
kind: Pod
metadata:
  annotations:
    appmesh.k8s.aws/sidecarInjectorWebhook: disabled
    helm.sh/hook: test-success
    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
    linkerd.io/inject: disabled
    sidecar.istio.io/inject: "false"
  labels:
    app.kubernetes.io/name: podinfo-git
    app.kubernetes.io/version: 6.4.0
    helm.sh/chart: podinfo-6.4.0
  name: podinfo-git-service-test-ioazw
spec:
  containers:
  - command:
    - sh
    - -c
    - |
      curl -s ${PODINFO_SVC}/api/info | grep version
    env:
    - name: PODINFO_SVC
      value: podinfo-git.podinfo-from-git:9898
    image: curlimages/curl:7.69.0
    name: curl
  restartPolicy: Never

component: demo-helm-charts
chart: podinfo-repo
resources: 

apiVersion: v1
kind: Pod
metadata:
  annotations:
    appmesh.k8s.aws/sidecarInjectorWebhook: disabled
    helm.sh/hook: test-success
    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
    linkerd.io/inject: disabled
    sidecar.istio.io/inject: "false"
  labels:
    app.kubernetes.io/name: cool-release-name-podinfo
    app.kubernetes.io/version: 6.4.0
    helm.sh/chart: podinfo-6.4.0
  name: cool-release-name-podinfo-service-test-h0q8g
spec:
  containers:
  - command:
    - sh
    - -c
    - |
      curl -s ${PODINFO_SVC}/api/info | grep version
    env:
    - name: PODINFO_SVC
      value: cool-release-name-podinfo.podinfo-from-repo:9898
    image: curlimages/curl:7.69.0
    name: curl
  restartPolicy: Never

```

Racer159 (changes requested):

waveywaves:
I am going to make the following changes 
- don't mention other components for `--why` based on https://github.com/defenseunicorns/zarf/pull/2309#issuecomment-1954537274 only show the resources with the relevant image
- simplify the logic with a helper function https://github.com/defenseunicorns/zarf/pull/2309#discussion_r1496740223
- parse manifests _and_ charts https://github.com/defenseunicorns/zarf/pull/2309#discussion_r1496822220

waveywaves:
Updated the PR based on the reviews

src/pkg/packager/prepare.go
```diff
@@ -356,6 +377,28 @@ func (p *Packager) processUnstructuredImages(resource *unstructured.Unstructured
 	return matchedImages, maybeImages, nil
 }
 
+func (p *Packager) findWhyResources(resources []*unstructured.Unstructured, whyImage, componentName, resourceName string, isChart bool) []string {
+	output := []string{}
+	for _, resource := range resources {
+		bytes, _ := resource.MarshalJSON()
```
lucasrod16, line 383: We should handle the error returned from `resource.MarshalJSON()`

src/pkg/packager/prepare.go
```diff
@@ -356,6 +377,28 @@ func (p *Packager) processUnstructuredImages(resource *unstructured.Unstructured
...
+		json := string(bytes)
+		resourceTypeKey := "manifest"
+		if isChart {
+			resourceTypeKey = "chart"
+		}
+
+		var imageFuzzyCheck = regexp.MustCompile(`(?mi)["|=]([a-z0-9\-.\/:]+:[\w.\-]*[a-z\.\-][\w.\-]*)"`)
+		matches := imageFuzzyCheck.FindAllStringSubmatch(json, -1)
+		for _, group := range matches {
+			if whyImage == group[1] {
```
lucasrod16, line 393: Since we already have the k8s resource and the why image as strings, we can simplify this further to use `strings.Contains` to check if the why image is present in the k8s resource

src/pkg/packager/prepare.go
```diff
@@ -356,6 +377,28 @@ func (p *Packager) processUnstructuredImages(resource *unstructured.Unstructured
...
+				yamlResource, _ := yaml.Marshal(resource.Object)
```
lucasrod16, line 394: We should handle the error returned from `yaml.Marshal()`

src/pkg/packager/prepare.go
```diff
@@ -356,6 +377,28 @@ func (p *Packager) processUnstructuredImages(resource *unstructured.Unstructured
 	return matchedImages, maybeImages, nil
 }
 
+func (p *Packager) findWhyResources(resources []*unstructured.Unstructured, whyImage, componentName, resourceName string, isChart bool) []string {
+	output := []string{}
```
lucasrod16, line 381: Rather than return the output as a string slice, we can return a string slice of found `resourceNames`. So if the why image is found in a resource, append the `resourceName` to the slice.

We can print the resource YAML inside of this function as well rather than construct the output outside of this function

src/pkg/packager/prepare.go
```diff
@@ -356,6 +377,28 @@ func (p *Packager) processUnstructuredImages(resource *unstructured.Unstructured
...
+		if isChart {
+			resourceTypeKey = "chart"
+		}
+
+		var imageFuzzyCheck = regexp.MustCompile(`(?mi)["|=]([a-z0-9\-.\/:]+:[\w.\-]*[a-z\.\-][\w.\-]*)"`)
+		matches := imageFuzzyCheck.FindAllStringSubmatch(json, -1)
+		for _, group := range matches {
+			if whyImage == group[1] {
+				yamlResource, _ := yaml.Marshal(resource.Object)
+				output = append(output, fmt.Sprintf("component: %s\n%s: %s\nresource: %s\n\n%s\n", componentName, resourceTypeKey, resourceName, resource.GetName(), string(yamlResource)))
```
lucasrod16, line 395: We shouldn't need `resource.GetName()` here because the name of the YAML manifest will be visible in the console output

lucasrod16 (changes requested):

src/pkg/packager/prepare.go
waveywaves: It's something I did earlier but reverted to constructing the output outside as the componentDefinition or the output for `zarf dev find-images` was also being constructed outside. Will go with your suggestion here. I noticed that the output is more readable when we print the resource yaml from inside of this function.

waveywaves:
> @waveywaves there are some merge conflicts that will need to be resolved

@lucasrod16 I will do the needful

src/pkg/packager/prepare.go
waveywaves: correct, seems like I missed this one

src/pkg/packager/prepare.go
waveywaves: correct, seems like I missed this one as well

waveywaves:
seems like the k3d test is flaking, can we restart that test ?

waveywaves:
the tests failed both times on kiwix data injection test against a minikube cluster, testing the test locally with minikube

src/pkg/packager/prepare.go
```diff
@@ -356,6 +376,31 @@ func (p *Packager) processUnstructuredImages(resource *unstructured.Unstructured
...
+			if err != nil {
+				return foundWhyResources, fmt.Errorf("could not marshal resource: %w", err)
+			}
+			fmt.Printf("component: %s\n%s: %s\nresource: %s\n\n%s\n", componentName, resourceTypeKey, resourceName, resource.GetName(), string(yamlResource))
+			foundWhyResources = append(foundWhyResources, resourceName)
+		}
+	}
+	return foundWhyResources, nil
+}
+
```
lucasrod16, line 403:
```suggestion
func findWhyResources(resources []*unstructured.Unstructured, whyImage, componentName, resourceName string, isChart bool) ([]string, error) {
	foundWhyResources := []string{}
	for _, resource := range resources {
		bytes, err := resource.MarshalJSON()
		if err != nil {
			return nil, err
		}
		json := string(bytes)
		resourceTypeKey := "manifest"
		if isChart {
			resourceTypeKey = "chart"
		}

		if strings.Contains(json, whyImage) {
			yamlResource, err := yaml.Marshal(resource.Object)
			if err != nil {
				return nil, err
			}
			fmt.Printf("component: %s\n%s: %s\nresource:\n\n%s\n", componentName, resourceTypeKey, resourceName, string(yamlResource))
			foundWhyResources = append(foundWhyResources, resourceName)
		}
	}
	return foundWhyResources, nil
}
```

src/pkg/packager/prepare.go
```diff
@@ -268,7 +286,9 @@ func (p *Packager) FindImages() (imgMap map[string][]string, err error) {
 		}
 	}
 
-	fmt.Println(componentDefinition)
+	if whyImage == "" {
+		fmt.Println(componentDefinition)
+	}
```
lucasrod16, line 291:
```suggestion
	if whyImage != "" {
		if len(whyResources) == 0 {
			message.Warnf("image %q not found in any charts or manifests", whyImage)
		}
		return nil, nil
	}

	fmt.Println(componentDefinition)
```

src/pkg/packager/prepare.go
```diff
@@ -158,6 +160,14 @@ func (p *Packager) FindImages() (imgMap map[string][]string, err error) {
...
 				matchedImages[image] = true
 			}
+
+			// Check if the --why flag is set
+			if whyImage != "" {
+				_, err := p.findWhyResources(resources, whyImage, component.Name, chart.Name, true)
+				if err != nil {
+					message.WarnErrf(err, "Error finding why resources for chart %s: %s", chart.Name, err.Error())
+				}
+			}
```
lucasrod16, line 170:
```suggestion
			// Check if the --why flag is set
			if whyImage != "" {
				whyResourcesChart, err := findWhyResources(resources, whyImage, component.Name, chart.Name, true)
				if err != nil {
					message.WarnErrf(err, "Error finding why resources for chart %s: %s", chart.Name, err.Error())
				}
				whyResources = append(whyResources, whyResourcesChart...)
			}
```

src/pkg/packager/prepare.go
```diff
@@ -193,6 +203,14 @@ func (p *Packager) FindImages() (imgMap map[string][]string, err error) {
...
 				yamls, _ := utils.SplitYAML(contents)
 				resources = append(resources, yamls...)
+
+				// Check if the --why flag is set and if it is process the manifests
+				if whyImage != "" {
+					_, err := p.findWhyResources(resources, whyImage, component.Name, manifest.Name, false)
+					if err != nil {
+						message.WarnErrf(err, "Error finding why resources for manifest %s: %s", manifest.Name, err.Error())
+					}
+				}
```
lucasrod16, line 213:
```suggestion
				// Check if the --why flag is set and if it is process the manifests
				if whyImage != "" {
					whyResourcesManifest, err := findWhyResources(resources, whyImage, component.Name, manifest.Name, false)
					if err != nil {
						message.WarnErrf(err, "Error finding why resources for manifest %s: %s", manifest.Name, err.Error())
					}
					whyResources = append(whyResources, whyResourcesManifest...)
				}
```

src/pkg/packager/prepare.go
```diff
@@ -37,6 +38,7 @@ type imageMap map[string]bool
 func (p *Packager) FindImages() (imgMap map[string][]string, err error) {
```
lucasrod16, line 38: Will need to add `whyResources := []string{}` as well

lucasrod16 (changes requested):

waveywaves:
> > the `e2e-test-with-cluster` test failed twice at a single kiwix data injection test against a minikube cluster first time the test failed because zarf could not connect to the cluster and right now it has failed because the test could not find a pod which was successfully deployed as seen in the logs below
> > https://github.com/defenseunicorns/zarf/actions/runs/8013856761/job/21891629494?pr=2309#step:8:5575
> > running the same possibly flakey `TestDataInjection` passes locally with minikube
> > ```
> > --- PASS: TestDataInjection (47.62s)
> > PASS
> > ok      github.com/defenseunicorns/zarf/src/test/e2e    48.707s
> > ```
> 
> Unfortunately that is a known flake - the issue actually isn't with the Kiwix package or Zarf data injections but with how the test expects to be able to lookup the pod afterward to check its contents are there - it assumes it can reuse the same pod but Kubernetes could recycle it in the intervening time and if it does the test will try to reach the wrong pod and fail. We just haven't had time yet to invest in fixing it yet.

If this issue documented ?

src/pkg/packager/prepare.go
waveywaves: done !

src/pkg/packager/prepare.go
```diff
@@ -356,6 +384,31 @@ func (p *Packager) processUnstructuredImages(resource *unstructured.Unstructured
...
+			yamlResource, err := yaml.Marshal(resource.Object)
+			if err != nil {
+				return nil, err
+			}
+			fmt.Printf("component: %s\n%s: %s\nresource:\n\n%s\n", componentName, resourceTypeKey, resourceName, string(yamlResource))
+			foundWhyResources = append(foundWhyResources, resourceName)
+		}
+	}
+	return foundWhyResources, nil
+}
```
lucasrod16, line 410: Looking at this function closer, we can simplify even further by only serializing the resource object once to YAML rather than serializing to JSON and then to YAML. This should be the last optimization we need to make to this logic 🤞 Thank you for your patience

```suggestion
func findWhyResources(resources []*unstructured.Unstructured, whyImage, componentName, resourceName string, isChart bool) ([]string, error) {
	foundWhyResources := []string{}
	for _, resource := range resources {
		bytes, err := yaml.Marshal(resource.Object)
		if err != nil {
			return nil, err
		}
		yaml := string(bytes)
		resourceTypeKey := "manifest"
		if isChart {
			resourceTypeKey = "chart"
		}

		if strings.Contains(yaml, whyImage) {
			fmt.Printf("component: %s\n%s: %s\nresource:\n\n%s\n", componentName, resourceTypeKey, resourceName, yaml)
			foundWhyResources = append(foundWhyResources, resourceName)
		}
	}
	return foundWhyResources, nil
}
```

lucasrod16 (changes requested):

src/pkg/packager/prepare.go
waveywaves: Oh I see what you did there ! No I appreciate the help, I'm learning in the process as well.

src/test/e2e/13_find_images_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	t.Run("zarf test find images success", func(t *testing.T) {
+		t.Log("E2E: Test Find Images")
+
+		testPackagePath := filepath.Join("examples", "dos-games")
+		expectedOutput := []byte{}
+		f, err := os.Open("src/test/packages/13-find-images/dos-games-find-images-expected.txt")
+		defer f.Close()
+
+		_, err = f.Read(expectedOutput)
```
lucasrod16, line 21: `expectedOutput` isn't being populated with data from the txt file, so the test is passing because stdout does contain an empty string. We can change this to use `os.ReadFile` and capture the returned `[]bytes` as `expectedOutput` in each test to ensure the data is read from the files correctly

src/test/packages/13-find-images/helm-charts-find-images-why-expected.txt
```diff
@@ -0,0 +1,197 @@
+component: demo-helm-charts
+chart: podinfo-oci
+resource: podinfo-oci-service-test-wlxqz
```
lucasrod16, line 3: Will need to make sure each expected txt file does not have a resource name since the implementation doesn't output this anymore

so should just be:

```
resource:

// YAML object
```

lucasrod16 (changes requested):

src/test/e2e/13_find_images_test.go
waveywaves: so I was under the wrong impression that the tests were passing as the expectedOutput was being populated 🤦

src/test/packages/13-find-images/helm-charts-find-images-why-expected.txt
waveywaves: I have updated the test also I had to make the names of the pods into regular expressions as the pod names are changing everytime

src/test/e2e/13_find_images_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+		stdout, _, err := e2e.Zarf("dev", "find-images", testPackagePath, "--why", "curlimages/curl:7.69.0")
+		require.NoError(t, err, "Expect no error here")
+		match, err := regexp.MatchString(string(expectedOutput), stdout)
+		require.NoError(t, err, "Expect no error here while matching expected output with actual output"
+		require.True(t, match, "Expected output does not match actual output")
+	})
+
+	t.Run("zarf test find images --why w/  manifests success", func(t *testing.T) {
+		t.Log("E2E: Test Find Images against a helm chart with why flag")
```
lucasrod16, line 41:
```suggestion
		t.Log("E2E: Test Find Images against manifests with why flag")
```

src/test/e2e/13_find_images_test.go
```diff
@@ -0,0 +1,52 @@
...
+		testPackagePath := filepath.Join("examples", "manifests")
+		expectedOutput, err := os.ReadFile("src/test/packages/13-find-images/manifests-find-images-why-expected.txt")
+		require.NoError(t, err, "Expect no error here while reading expectedOutput of the expected output file")
+
+		stdout, _, err := e2e.Zarf("dev", "find-images", testPackagePath, "--why", "httpd:alpine3.18")
+		require.NoError(t, err, "Expect no error here")
+		require.Contains(t, stdout, string(expectedOutput))
+	})
+
+}
```
lucasrod16, line 52: We can remove the extra messages in all `require.NoError` and `require.True` calls in this test. The output from `require` contains this information already, so it is redundant to add extra messaging in these cases.

For example:

instead of:

```go
require.NoError(t, err, "Expect no error here")
```

do:

```go
require.NoError(t, err)
```

src/test/e2e/13_find_images_test.go
```diff
@@ -0,0 +1,52 @@
...
+		t.Log("E2E: Test Find Images against a helm chart with why flag")
+
+		testPackagePath := filepath.Join("examples", "helm-charts")
+		expectedOutput, err := os.ReadFile("src/test/packages/13-find-images/helm-charts-find-images-why-expected.txt")
+		require.NoError(t, err, "Expect no error here while reading expectedOutput of the expected output file")
+
+		stdout, _, err := e2e.Zarf("dev", "find-images", testPackagePath, "--why", "curlimages/curl:7.69.0")
+		require.NoError(t, err, "Expect no error here")
+		match, err := regexp.MatchString(string(expectedOutput), stdout)
+		require.NoError(t, err, "Expect no error here while matching expected output with actual output"
```
lucasrod16, line 36: Missing a closing parentheses

```suggestion
		require.NoError(t, err)
```

src/test/e2e/13_find_images_test.go
lucasrod16, line 1: You can run this test locally using:

```shell
go test ./src/test/e2e -v -count=1 -run TestFindImages
```

https://docs.zarf.dev/docs/contribute-to-zarf/testing#cli-end-to-end-tests

lucasrod16 (changes requested):
Few minor tweaks. Otherwise, this looks good to me!

waveywaves:
@lucasrod16 updated the PR

lucasrod16 (approved):
lgtm

src/pkg/packager/prepare.go
```diff
@@ -193,6 +205,15 @@ func (p *Packager) FindImages() (imgMap map[string][]string, err error) {
 				message.Debugf("%s", contentString)
 				yamls, _ := utils.SplitYAML(contents)
 				resources = append(resources, yamls...)
+
+				// Check if the --why flag is set and if it is process the manifests
+				if whyImage != "" {
+					whyResourcesManifest, err := findWhyResources(resources, whyImage, component.Name, manifest.Name, false)
```
Racer159, line 211:
```suggestion
					whyResourcesManifest, err := findWhyResources(yamls, whyImage, component.Name, manifest.Name, false)
```

src/pkg/packager/prepare.go
```diff
@@ -158,6 +161,15 @@ func (p *Packager) FindImages() (imgMap map[string][]string, err error) {
 			for _, image := range annotatedImages {
 				matchedImages[image] = true
 			}
+
+			// Check if the --why flag is set
+			if whyImage != "" {
+				whyResourcesChart, err := findWhyResources(resources, whyImage, component.Name, chart.Name, true)
```
Racer159, line 167:
```suggestion
				whyResourcesChart, err := findWhyResources(yamls, whyImage, component.Name, chart.Name, true)
```

src/pkg/packager/prepare.go
Racer159: These should use the yamls from the current manfiest set (or chart for the above) - if `resources` is used then it will potentially repeat the resource into places that it actually doesn't exist (this shows in the repeated `podinfo` curlpods in the test and also in the `wordpress` example:
image
...
image

(`connect-services` doesn't have a deployment in it)

Racer159 (changes requested):
One small bug, but otherwise super close!

src/pkg/packager/prepare.go
lucasrod16: @Racer159 good catch!

src/test/e2e/13_find_images_test.go
```diff
@@ -0,0 +1,53 @@
...
+		testPackagePath := filepath.Join("examples", "helm-charts")
+		expectedOutput, err := os.ReadFile("src/test/packages/13-find-images/helm-charts-find-images-why-expected.txt")
+		require.NoError(t, err)
+
+		stdout, _, err := e2e.Zarf("dev", "find-images", testPackagePath, "--why", "curlimages/curl:7.69.0")
+		require.NoError(t, err)
+		match, err := regexp.MatchString(string(expectedOutput), stdout)
+		require.NoError(t, err)
+		require.True(t, match)
+	})
```
lucasrod16, line 39: This test case does not validate the expected output accurately because of the use of `regexp.MatchString`. The pod that uses the curl image in the podinfo chart has a dynamically-generated pod name, leading to non-deterministic output, which makes this example unsuitable for a good test case.

If you run the the command with the --why flag locally for the curl image against examples/helm-charts, you will see that the output correctly only has 3 resources printed to stdout, but the test txt file has 6.

https://github.com/stefanprodan/podinfo/blob/47090ad9e14e60f33c3d562b5ca626742c8d1309/charts/podinfo/templates/tests/service.yaml#L4

Recommend using the `examples/wordpress` helm chart as a test case instead so that we can have deterministic output and more accurately assert against it.

```suggestion
	t.Run("zarf test find images --why  w/ helm chart success", func(t *testing.T) {
		t.Log("E2E: Test Find Images against a helm chart with why flag")

		testPackagePath := filepath.Join("examples", "wordpress")
		expectedOutput, err := os.ReadFile("src/test/packages/13-find-images/helm-charts-find-images-why-expected.txt")
		require.NoError(t, err)

		stdout, _, err := e2e.Zarf("dev", "find-images", testPackagePath, "--why", "docker.io/bitnami/apache-exporter:0.13.3-debian-11-r2")
		require.NoError(t, err)
		require.Contains(t, stdout, string(expectedOutput))
	})
```

src/test/packages/13-find-images/helm-charts-find-images-why-expected.txt
```diff
@@ -0,0 +1,197 @@
...
+    - sh
+    - -c
+    - |
+      curl -s ${PODINFO_SVC}/api/info | grep version
+    env:
+    - name: PODINFO_SVC
+      value: cool-release-name-podinfo.podinfo-from-repo:9898
+    image: curlimages/curl:7.69.0
+    name: curl
+  restartPolicy: Never
```
lucasrod16, line 197:
```suggestion
component: wordpress
chart: wordpress
resource:

apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/instance: wordpress
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: wordpress
    helm.sh/chart: wordpress-16.0.4
  name: wordpress
  namespace: wordpress
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/instance: wordpress
      app.kubernetes.io/name: wordpress
  strategy:
    type: RollingUpdate
  template:
    metadata:
      annotations: null
      labels:
        app.kubernetes.io/instance: wordpress
        app.kubernetes.io/managed-by: Helm
        app.kubernetes.io/name: wordpress
        helm.sh/chart: wordpress-16.0.4
    spec:
      affinity:
        nodeAffinity: null
        podAffinity: null
        podAntiAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
          - podAffinityTerm:
              labelSelector:
                matchLabels:
                  app.kubernetes.io/instance: wordpress
                  app.kubernetes.io/name: wordpress
              topologyKey: kubernetes.io/hostname
            weight: 1
      containers:
      - env:
        - name: BITNAMI_DEBUG
          value: "false"
        - name: ALLOW_EMPTY_PASSWORD
          value: "yes"
        - name: MARIADB_HOST
          value: wordpress-mariadb
        - name: MARIADB_PORT_NUMBER
          value: "3306"
        - name: WORDPRESS_DATABASE_NAME
          value: bitnami_wordpress
        - name: WORDPRESS_DATABASE_USER
          value: bn_wordpress
        - name: WORDPRESS_DATABASE_PASSWORD
          valueFrom:
            secretKeyRef:
              key: mariadb-password
              name: wordpress-mariadb
        - name: WORDPRESS_USERNAME
          value: null
        - name: WORDPRESS_PASSWORD
          valueFrom:
            secretKeyRef:
              key: wordpress-password
              name: wordpress
        - name: WORDPRESS_EMAIL
          value: null
        - name: WORDPRESS_FIRST_NAME
          value: null
        - name: WORDPRESS_LAST_NAME
          value: null
        - name: WORDPRESS_HTACCESS_OVERRIDE_NONE
          value: "no"
        - name: WORDPRESS_ENABLE_HTACCESS_PERSISTENCE
          value: "no"
        - name: WORDPRESS_BLOG_NAME
          value: null
        - name: WORDPRESS_SKIP_BOOTSTRAP
          value: "no"
        - name: WORDPRESS_TABLE_PREFIX
          value: wp_
        - name: WORDPRESS_SCHEME
          value: http
        - name: WORDPRESS_EXTRA_WP_CONFIG_CONTENT
          value: ""
        - name: WORDPRESS_PLUGINS
          value: none
        - name: APACHE_HTTP_PORT_NUMBER
          value: "8080"
        - name: APACHE_HTTPS_PORT_NUMBER
          value: "8443"
        envFrom: null
        image: docker.io/bitnami/wordpress:6.2.0-debian-11-r18
        imagePullPolicy: IfNotPresent
        livenessProbe:
          failureThreshold: 6
          httpGet:
            httpHeaders: []
            path: /wp-admin/install.php
            port: http
            scheme: HTTP
          initialDelaySeconds: 120
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 5
        name: wordpress
        ports:
        - containerPort: 8080
          name: http
        - containerPort: 8443
          name: https
        readinessProbe:
          failureThreshold: 6
          httpGet:
            httpHeaders: []
            path: /wp-login.php
            port: http
            scheme: HTTP
          initialDelaySeconds: 30
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 5
        resources:
          limits: {}
          requests:
            cpu: 300m
            memory: 512Mi
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop:
            - ALL
          runAsNonRoot: true
          runAsUser: 1001
        volumeMounts:
        - mountPath: /bitnami/wordpress
          name: wordpress-data
          subPath: wordpress
      - command:
        - /bin/apache_exporter
        - --scrape_uri
        - http://status.localhost:8080/server-status/?auto
        image: docker.io/bitnami/apache-exporter:0.13.3-debian-11-r2
        imagePullPolicy: IfNotPresent
        livenessProbe:
          failureThreshold: 3
          httpGet:
            path: /metrics
            port: metrics
          initialDelaySeconds: 15
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 5
        name: metrics
        ports:
        - containerPort: 9117
          name: metrics
        readinessProbe:
          failureThreshold: 3
          httpGet:
            path: /metrics
            port: metrics
          initialDelaySeconds: 5
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 3
        resources:
          limits: {}
          requests: {}
      hostAliases:
      - hostnames:
        - status.localhost
        ip: 127.0.0.1
      securityContext:
        fsGroup: 1001
        seccompProfile:
          type: RuntimeDefault
      serviceAccountName: default
      volumes:
      - name: wordpress-data
        persistentVolumeClaim:
          claimName: wordpress

```

lucasrod16 (changes requested):

src/test/packages/13-find-images/helm-charts-find-images-why-expected.txt
waveywaves: ok using regexp.match is not really deterministic testing
hence we changed the test helm chart as well

I considered testing another helm chart for this reason but stuck with the helm-charts example was because it seemed to be the "flagship" helm chart packaging and deployment example, but considering the determinism this makes sense as well

src/test/e2e/13_find_images_test.go
```diff
@@ -0,0 +1,51 @@
+package test
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
```
Racer159, line 6:
```suggestion
```
Looks like this is causing an issue with the tests

Racer159 (changes requested):

lucasrod16 (approved):

Racer159 (approved):
lgtm

PR #2310, opened 2024-02-16
fix: change text detect to check first and last 512 bytes

### LINKED ISSUE #2308: zarf tries to apply template replacements to large files that might be incorrectly assessed as text
### Environment
Device and OS: linux/amd64
App version: 0.32.3
Kubernetes distro being used: k3s (Not applicable for this problem)
Other:

### Steps to reproduce
1. Create a zarf package that pulls a self-executing script like the NVIDIA script:
```yaml
    files:
      - source: https://us.download.nvidia.com/XFree86/Linux-x86_64/535.154.05/NVIDIA-Linux-x86_64-535.154.05.run
        target: /tmp/nvidia-driver.run
        executable: true
```

2. Attempt to deploy it using zarf package deploy

### Expected result

FIle is simply copied to location

### Actual Result

This particular type of file is technically a script with an archive concatenated with it and it is rather large. Zarf detects it as a text file since it has text up front and tries to apply templating (###ZARF_VAR_*) to it.

There should be a way perhaps in the zarf.yaml to force it to not take this behavior. Maybe something like this?
```yaml
    files:
      - source: https://us.download.nvidia.com/XFree86/Linux-x86_64/535.154.05/NVIDIA-Linux-x86_64-535.154.05.run
        target: /tmp/nvidia-driver.run
        executable: true
        noTemplate: true                  // new optional boolean
```

OR

Change the detection to also check the last couple of bytes at the end of the file too, should not be text for these type of installers.

If you opt for the latter, I can probably submit a PR myself. If you want to do the former, I would prefer your team discuss how you want to proceed. Happy to help where I can!

```bash                                                                                                      
  📦 DRIVERS COMPONENT                                                                                
                                                                                                      

  ⠦  Templating /tmp/nvidia-driver.run (9m10s)  
```

### Severity/Priority

I can work around this by using an action to tar.gz it up beforehand but would be nice if I didn't have to do that.

### Additional Context

Looks like this is the area of concern:

https://github.com/defenseunicorns/zarf/blob/7e91d3b9823b52fe6d0f563d692c8af57faa6005/src/pkg/packager/deploy.go#L360

WeaponX314:
## Description

Alters text detection logic to read the first and last 512 bytes.

Tested with 5 files:
- [NVIDIA installer](https://us.download.nvidia.com/XFree86/Linux-x86_64/535.154.05/NVIDIA-Linux-x86_64-535.154.05.run)
    Detected as application type when reading last 512.
- 3 4k size files of junk text with a ZARF_CONST replacement, in straight text, yaml, and json
    All 3 detected as text/plain, ZARF_CONST was replaced.
- 1 small 100 byte file with a ZARF_CONST replacement.
   Was still detected as text and ZARF_CONST was replaced.
  
Existing unit tests succeeded.

## Related Issue

Fixes #2308

Relates to #

## Type of change

## Checklist before merging


src/pkg/utils/io.go
```diff
@@ -471,9 +471,20 @@ func IsTextFile(path string) (bool, error) {
...
-	// Read the first 512 bytes of the file
+	// Get file stat
+	stat, err := f.Stat()
+	if err != nil {
+		return false, err
+	}
+
+	// Create 512 byte buffer
 	data := make([]byte, 512)
-	n, err := f.Read(data)
```
Racer159, line 476: We will still want to read the first bytes as well in case text is added to the end of a binary file (also the end of a file is unlikely to have BOMs or the starter sequences to get an accurate mimetype - this function doesn't _really_ need that but it would be nicer to base the primary decision information on whether this is a text file off of a more accurate mimetype).

Racer159 (changes requested):

src/pkg/utils/io.go
WeaponX314: My only question to that is are you fine with it always reading front and back? Since not matter what, due to this edge case with installers like NVIDIA, it will always need to do that. I guess in order to pass the "is this a text file" check is both front and back must be a text mimetype. Not hard to do at all, I can make this change either today or early next week.

Racer159 (approved):
lgtm

lucasrod16 (approved):

Issue #2312, opened 2024-02-18
Consistent file formatting


Noxsios:
### Describe what should be investigated or refactored

Currently there exists a pre-commit hook to run `go fmt` on changed files.

There exists a `.editorconfig`, and a `.markdownlint.json` config files at root, but these are opt-in formatting configs.

For consistency, all files whose consistent formatting we care about should be formatting using a combination of `go fmt`, `prettier`, `shellcheck`, and `markdownlint-cli2`.

Paths we care about:

- `**/*.md` (`!docs/2-the-zarf-cli/100-cli-commands/`)
- `**/*.go`
- `**/*.ya?ml`
- `hack/*.sh`

? : does `prettier` handle TOML?

related: should we decide upon `yaml` vs `yml`?

related: need to cleanup and re-visit pre-commit hooks, some are deprecated / not maintained anymore

related: should write some small `sh` script to ensure SPDX header is present on all code files.

Noxsios:
A solid baseline may be https://github.com/Azure/aks-periscope/blob/master/.pre-commit-config.yaml

Noxsios:
Can def be split into formatting issues / PRs based upon file type

PR #2314, opened 2024-02-19
fix: multi-part tarballs being mismatched sizes

### LINKED ISSUE #2313: Unevenly sized multi-part packages
### Environment
Device and OS: System76 darp8
App version: v0.32.3
Other: n/a

### Steps to reproduce
1. Create a package with `--max-package-size` set smaller than the resulting package

### Expected result

Even chunks of a package are created
image

### Actual Result

Varying chunks are created
image

### Severity/Priority

Medium

### Additional Context

This effectively breaks this feature on v0.32.3 - workaround for now is to downgrade to v0.32.2 for those who need this

Racer159:
## Description

This fixes multipart tarballs being different sizes with `--max-package-size`

## Related Issue

Fixes #2313

## Type of change

## Checklist before merging


src/pkg/utils/io.go
```diff
@@ -404,10 +404,14 @@ func SplitFile(srcFile string, chunkSizeBytes int) (err error) {
...
+			err = chunkFile.Close()
+			if err != nil {
+				return err
+			}
 
 			// create new file
-			path = fmt.Sprintf("%s.part%03d", srcFile, len(fileNames)+1)
-			chunkFile, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0644)
+			path = fmt.Sprintf("%s.part%03d", srcPath, len(fileNames)+1)
+			chunkFile, err = os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0644)
```
daniel-palmer-gu, line 414: That's unfortunate. I had fixed this at one point. Seemed to have lost this somehow. Thanks for fixing it!

src/test/e2e/05_tarball_test.go
```diff
@@ -36,6 +37,21 @@ func TestMultiPartPackage(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, int64(1000000), part2FileInfo.Size())
+	// Check the package data is correct
+	pkgData := types.ZarfSplitPackageData{}
+	part0File, err := os.ReadFile(parts[0])
+	require.NoError(t, err)
+	err = json.Unmarshal(part0File, &pkgData)
+	require.NoError(t, err)
+	require.Equal(t, pkgData.Count, 6)
+	fmt.Printf("%#v", pkgData)
```
daniel-palmer-gu, line 54: Just so you are aware, the original code would pass this test since the buffer size will dynamically change if it is is below 16MiB. Only when --max-package-size > 17 does the error show up.

Racer159: Good point! updated

Racer159:
(putting this in draft since the s3 file needs to be bigger)

src/pkg/packager/sources/split.go
```diff
@@ -85,6 +85,12 @@ func (s *SplitTarballSource) Collect(dir string) (string, error) {
 		if _, err = io.Copy(pkgFile, f); err != nil {
 			return "", fmt.Errorf("unable to copy file %s: %w", file, err)
 		}
+
+		// Close the file when done copying
+		err = f.Close()
+		if err != nil {
+			return "", fmt.Errorf("unable to close file %s: %w", file, err)
+		}
```
AustinAbro321, line 93:
```suggestion
		if err := f.Close(); err != nil {
			return "", fmt.Errorf("unable to close file %s: %w", file, err)
		}
```
Back to style conversation here https://github.com/defenseunicorns/zarf/issues/2312, do we want to enforce things like this. Something to discuss

src/test/e2e/05_tarball_test.go
```diff
@@ -29,13 +30,28 @@ func TestMultiPartPackage(t *testing.T) {
 	e2e.CleanFiles(deployPath, outputFile)
 
 	// Create the package with a max size of 1MB
```
lucasrod16, line 32: nit: comment wasn't updated to reflect what the code is doing

I'd also be okay with deleting the comment because it's self-explanatory IMO

Other than that, this lgtm

```suggestion
	// Create the package with a max size of 20MB
```

lucasrod16 (approved):

lucasrod16 (approved):
lgtm

Issue #1444, opened 2023-03-16
zarf registry login hangs indefinitely with --password-stdin flag


mikevanhemert:
### Environment
Device and OS:
* On-prem customer server with Ubuntu 22.04.2 LTS
* System76 Darter Pro with Pop!_OS 22.04 LTS

App version: v0.25.0-rc1
Kubernetes distro being used: n/a
Other:

### Steps to reproduce
1. Attempt zarf registry login with valid credentials to `registry1.dso.mil`
    ```bash
    zarf  tools registry login -u [username] --password-stdin registry1.dso.mil -l debug
    # enter your password
    ```

### Expected result
Zarf prompts for password or entering password immediately proceeds to registry login

### Actual Result
Zarf hangs with no indication of success, failure, next steps, etc.

### Visual Proof (screenshots, videos, text, etc)
n/a

### Severity/Priority
Low

### Additional Context
Zarf registry login with valid credentials and the `-p` flag is successful
```bash
zarf tools registry login -u [username] -p [password] registry1.dso.mil -l debug
2023/03/16 14:17:22 logged in via /home/[system username]/.docker/config.json
```

Racer159:
The zarf tools registry login command is from crane and pulls the password from a pipe like:

```
echo $MYPASSWORD | zarf  tools registry login -u [username] --password-stdin registry1.dso.mil
```

The closest issue I could find upstream is this: https://github.com/google/go-containerregistry/issues/718

PR #2319, opened 2024-02-21
fix: add http request header timeout to help stalling image push

Racer159:
## Description

This is a test for fixes to intermittent hanging.

## Related Issue

Relates to #1444

## Type of change

## Checklist before merging


src/cmd/tools/crane.go
AustinAbro321, line 1: Lint

AustinAbro321 (dismissed):

lucasrod16 (approved):
lgtm

lucasrod16 (approved):

src/internal/packager/images/push.go
```diff
@@ -50,6 +51,7 @@ func (i *ImageConfig) PushToZarfRegistry() error {
 
 	httpTransport := http.DefaultTransport.(*http.Transport).Clone()
 	httpTransport.TLSClientConfig.InsecureSkipVerify = i.Insecure
+	httpTransport.ResponseHeaderTimeout = 10 * time.Second
```
Noxsios, line 54: for faster experimentation, should we expose this as a config option?

99% of the time 10s is prob fine, but could be weird cases w/ satcom connections

Noxsios (changes requested):

src/internal/packager/images/push.go
Racer159: After doing a bit more thinking / research I think this would be much better explored in a dedicated issue: https://github.com/defenseunicorns/zarf/issues/2331

It should ideally be something we can apply globally and for sat com connections there are other hard-coded timeouts in the default transport as well (10s is the same as the TLS handshake timeout and should be a reasonably conservative choice for an image registry to return its headers)

Noxsios (approved):

PR #2321, opened 2024-02-21
feat!: filter package components with strategy interface

### LINKED ISSUE #2320: OCI deploy with `default: true` component fails to find image blob
### Environment
Device and OS: MBP
App version: 0.32.3
Other:

### Steps to reproduce
We've seen this consistently on our uds-core package so the steps below reproduce by using that:
```console
zarf p deploy oci://defenseunicorns/uds-k3d:0.4.0 --confirm
zarf init --confirm
zarf p deploy oci://ghcr.io/defenseunicorns/packages/uds/core:0.13.0-registry1 --confirm
```

### Expected result

k3d is deployed with uds-core on top.

### Actual Result

Deployment errors out on deploying metrics-server which is a [`default: true` component](https://github.com/defenseunicorns/uds-core/blob/main/packages/standard/zarf.yaml#L47-L48):
```console
      WARNING  Retrying (1/3): open
 /tmp/zarf-2787421378/images/blobs/sha256/3ef3b57e144bb6cd169eb47bb1e985c863a49d6f15827c8a7467b3b7292e7f07:
 no such file or directory
```

### Severity/Priority

Medium - my understanding is that required/default is being refactored so this might not be an issue for long but it is causing issues with current deployments.

### Additional Context

We noticed that explicitly selecting the `metrics-server` component during deploy seems to resolve the issue, as well as pulling then deploying. This seems to exclusively be an issue with a direct OCI deploy.

Noxsios:
## Description

Consolidate component filtering logic into a `filters` package. Each filter is an implementation of

```go
// ComponentFilterStrategy is a strategy interface for filtering components.
type ComponentFilterStrategy interface {
	Apply(types.ZarfPackage) ([]types.ZarfComponent, error)
}
```

Public construction functions return instances of this interface _not_ instances of their underlying structs. Consumers should be fully cognizant of which filter they are using, and should not be making a common wrapper function (eg. `NewFilter(args...)` to dynamically return a filter.

ex:

```go
func Empty() ComponentFilterStrategy {
	return &emptyFilter{}
}

// emptyFilter is a filter that does nothing.
type emptyFilter struct{}

// Apply returns the components unchanged.
func (f *emptyFilter) Apply(pkg types.ZarfPackage) ([]types.ZarfComponent, error) {
	return pkg.Components, nil
}
```

BREAKING CHANGES: This changes the interface signatures on `sources.PackageSource` to reflect the new behavior whereupon the `zarf.yaml` is loaded into memory within the sources load operations. This allows for `filter`ing to take place during load and for the `zarf.yaml` in memory to reflect these filter operations.

## Related Issue

Fixes #2320 

## Type of change

## Checklist before merging


lucasrod16 (commented):
I don't have a strong opinion on whether filter implementations return the strategy interface or the implementation struct.

My preference would be to return the concrete implementation struct because it makes the code a bit easier to navigate imo. For example, when I click into `Apply()` in vs code in the client code, it takes me to the interface method signature instead of the implementation. It would be nice to be taken directly to the implementation, but not a big enough reason to outweigh other factors.

src/types/component.go
```diff
@@ -21,50 +21,93 @@ type ZarfComponent struct {
 	// Default changes the default option when deploying this component
 	Default bool `json:"default,omitempty" jsonschema:"description=Determines the default Y/N state for installing this component on package deploy"`
 
-	// Required makes this component mandatory for package deployment
-	Required bool `json:"required,omitempty" jsonschema:"description=Do not prompt user to install this component, always install on package deploy"`
+	// DeprecatedRequired makes this component mandatory for package deployment
+	DeprecatedRequired *bool `json:"required,omitempty" jsonschema:"description=[Deprecated] Do not prompt user to install this component, always install on package deploy. This will be removed in Zarf v1.0.0.,deprecated=true"`
```
AustinAbro321, line 25: I'm sure this is just a side effect on the split, but I think we'll want to wait until optional is introduced before we mark required as deprecated

AustinAbro321 (changes requested):

src/pkg/packager/filters/deploy_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"strings"
+	"testing"
+
+	"github.com/defenseunicorns/zarf/src/pkg/utils/helpers"
+	"github.com/defenseunicorns/zarf/src/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func cfq(t *testing.T, q string) types.ZarfComponent {
```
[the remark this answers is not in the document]
Noxsios: Component From Query, since this function is only used in testing, I felt an opaque name was fine. Can def make it more descriptive if that's your jam.

src/pkg/packager/filters/deploy_test.go
```diff
@@ -0,0 +1,128 @@
...
+			},
+		},
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			filter := &deploymentFilter{
+				// we dont currently unit test interactive CLI behavior
+				isInteractive: false,
+			}
```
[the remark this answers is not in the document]
Noxsios: yes, tests should be against public constructors, not private, unless the thing your are testing is purely private in nature.

src/types/component.go
Noxsios: This is why this is marked as a breaking change. The changeover is happening at the same time the deprecation is being marked.

src/pkg/packager/filters/deploy_test.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			// we do not currently support interactive mode in unit tests
+			isInteractive := false
+			filter := ForDeploy(tc.optionalComponents, isInteractive)
+
+			result, err := filter.Apply(tc.pkg)
+			require.NoError(t, err)
```
Racer159, line 163: It would be good to test some of the error states that come out of this.

src/pkg/packager/filters/select_test.go
```diff
@@ -0,0 +1,91 @@
...
+				{Name: "component3"},
+			},
+			expectedResult: []types.ZarfComponent{
+				{Name: "component2"},
+			},
+			expectedError: nil,
+		},
+		{
+			name:                "Test when requestedComponents contains an excluded component name",
+			requestedComponents: "component1, -component2",
```
Racer159, line 52: This would be better if the inclusion included the exclusion:
```suggestion
			requestedComponents: "comp*, -component2",
```

src/pkg/packager/filters/select_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func Test_selectStateFilter_Apply(t *testing.T) {
+	tests := []struct {
+		name                string
+		requestedComponents string
+		components          []types.ZarfComponent
+		expectedResult      []types.ZarfComponent
+		expectedError       error
```
Racer159, line 20: This isn't used (but it likely should)

Racer159 (changes requested):

src/pkg/packager/filters/deploy.go
```diff
@@ -0,0 +1,182 @@
...
+			var groupDefault *types.ZarfComponent
+			var groupSelected *types.ZarfComponent
+
+			for _, component := range groupedComponents[groupKey] {
+				// Ensure we have a local version of the component to point to (otherwise the pointer might change on us)
+				component := component
+
+				selectState, matchedRequest := includedOrExcluded(component.Name, f.requestedComponents)
+
+				if !isRequired(component) {
```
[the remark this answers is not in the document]
Noxsios: I think I will let that happen in `required --> optional` to reduce some of the merge conflicts (for my own sanity)

src/pkg/packager/filters/deploy.go
```diff
@@ -0,0 +1,182 @@
...
+					}
+				}
+			} else {
+				component := groupedComponents[groupKey][0]
+
+				// default takes precedence over required
+				if component.Default {
+					selectedComponents = append(selectedComponents, component)
+					continue
+				}
```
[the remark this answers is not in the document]
Noxsios: fixed

src/pkg/packager/filters/platform.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Apply applies the filter.
+func (f *archAndOSFilter) Apply(pkg types.ZarfPackage) ([]types.ZarfComponent, error) {
+	if !slices.Contains(allowedOs, f.os) {
+		return nil, fmt.Errorf("invalid OS: %s", f.os)
+	}
+
+	if !slices.Contains(allowedArch, f.arch) {
+		return nil, fmt.Errorf("invalid architecture: %s", f.arch)
+	}
```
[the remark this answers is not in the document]
Noxsios: I was attempting to keep the public constructor functions all have the same signature.

src/pkg/packager/filters/select_test.go
Noxsios: ?

src/pkg/packager/deploy.go
```diff
@@ -67,6 +71,11 @@ func (p *Packager) Deploy() (err error) {
 		return fmt.Errorf("deployment cancelled")
 	}
 
+	p.cfg.Pkg.Components, err = filter.Apply(p.cfg.Pkg)
```
AustinAbro321, line 74: Feels a bit weird that we filter again in load then filter again here. I believe the only place that would matter, as the code is currently, is if we are deploying an OCI component without confirm, but since our filters have knowledge of confirm, perhaps we should call filters in OCISources either way and then skip the additional apply here

src/pkg/packager/sources/oci.go
```diff
@@ -35,59 +35,69 @@ type OCISource struct {
...
-
 	// pull only needed layers if --confirm is set
 	if config.CommonOptions.Confirm {
-
-		layersToPull, err = s.LayersFromRequestedComponents(ctx, optionalComponents)
+		pkg, err = s.FetchZarfYAML(ctx)
+		if err != nil {
+			return pkg, nil, err
+		}
+		if filter != nil {
```
AustinAbro321, line 50: I want to better understand why filtering is only done on confirm

AustinAbro321: Filtering should be done on both

src/pkg/interactive/prompt.go
```diff
@@ -17,19 +16,10 @@ import (
...
-			return nil, fmt.Errorf("unable to get password for private key: %w", err)
-		}
-		return []byte(password), nil
+	prompt := &survey.Password{
+		Message: "Private key password (empty for no password): ",
 	}
-
-	// We are returning a nil error here because purposefully avoiding a password input is a valid use condition
-	return nil, nil
+	return []byte(password), survey.AskOne(prompt, &password)
```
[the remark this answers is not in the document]
Noxsios: let me know if this is now resolved!

src/pkg/packager/deploy.go
Noxsios: fixed, was dead code from a previous change

src/pkg/packager/sources/oci.go
Noxsios: fixed

src/pkg/interactive/prompt.go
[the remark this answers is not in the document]
AustinAbro321: Yeah it's fixed now

src/pkg/packager/filters/deploy_test.go
Noxsios: done

AustinAbro321 (approved):
LGTM

lucasrod16 (dismissed):

src/pkg/packager/sources/oci.go
```diff
@@ -35,59 +35,74 @@ type OCISource struct {
...
+			pkg.Components, err = filter.Apply(pkg)
```
Racer159, line 51: setting pkg components like this causes issues (see above)

Racer159 (changes requested):

src/pkg/packager/sources/oci.go
```diff
@@ -35,59 +35,74 @@ type OCISource struct {
...
-	if err := utils.ReadYaml(dst.ZarfYAML, &pkg); err != nil {
-		return err
+	// if --confirm is not set, read the zarf.yaml that was pulled
+	// and apply the filter to the components
+	if !config.CommonOptions.Confirm {
+		pkg, warnings, err = dst.ReadZarfYAML()
+		if err != nil {
+			return pkg, nil, err
+		}
+		pkg.Components, err = filter.Apply(pkg)
```
Racer159, line 84: setting pkg components like this causes issues (see above)

(clicked the wrong one)

Racer159 (changes requested):

AustinAbro321 (approved):

src/pkg/packager/deploy.go
```diff
@@ -44,14 +44,29 @@ func (p *Packager) resetRegistryHPA() {
...
 		filters.ByLocalOS(runtime.GOOS),
-		filters.ForDeploy(p.cfg.PkgOpts.OptionalComponents, !config.CommonOptions.Confirm),
+		filters.ForDeploy(p.cfg.PkgOpts.OptionalComponents, isInteractive),
 	)
 
-	p.cfg.Pkg, p.warnings, err = p.source.LoadPackage(p.layout, filter, true)
-	if err != nil {
-		return fmt.Errorf("unable to load the package: %w", err)
+	if isInteractive {
+		filter := filters.Empty()
```
Racer159, line 55: I am fine with this, but just wanted to note we show k3s now on mac os (still won't prompt which is what matters more IMO)

Racer159 (approved):

src/pkg/packager/deploy.go
Racer159: (_if_ they read it it should be obvious that it is only linux)

PR #2324, opened 2024-02-22
feat: validates signatures against the images if they exist in the images spec

### LINKED ISSUE #2257: Validate cosign signatures if included in `images`
**As** Ezra **I want** to validate cosign signatures on `zarf package create` so that I can have confidence that they will work correctly before the package goes to the air gap.

- **Given** I have a package with cosign signatures defined under `images`
- **When** Zarf pulls the signatures and the images to which they relate
- **Then** Zarf validates the signatures against the images

We could have a separate process for this (and this will slow down `create`) but since it will only run when cosign signatures are defined it should be a fine tradeoff since people can opt into the slowdown if they need / want to and if they do it will save them time in the long run since it would be costly to bring an invalid package to the airgap.

### Additional context
https://github.com/defenseunicorns/zarf/issues/475

waveywaves:
## Description

Validate images using cosign signatures if they exist in the images spec

## Related Issue

Fixes #2257

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/internal/packager/images/pull.go
+++ b/src/internal/packager/images/pull.go
@@ -125,7 +125,7 @@ func (i *ImageConfig) PullAll() ([]ImgInfo, error) {
 			if cosignArtifactsExist {
 				for _, cosignArtifact := range cosignList {
 					if strings.HasSuffix(cosignArtifact, ".sig") {
-						err := utils.CosignVerifyBlob(actualSrc, cosignArtifact, "")
+						err := utils.CosignVerifyBlob(actualSrc, cosignArtifact, "https://zarf.dev/cosign.pub")
 						if err != nil {
 							metadataImageConcurrency.ErrorChan <- fmt.Errorf("failed to get verify image using cosign signature for %s: %w", actualSrc, err)
 							return
--- a/src/internal/packager/images/pull.go
+++ b/src/internal/packager/images/pull.go
@@ -108,24 +108,24 @@ func (i *ImageConfig) PullAll() ([]ImgInfo, error) {
 			}
 
 			cosignArtifactsExist := false
-			cosignList, err := utils.GetCosignArtifacts(actualSrc)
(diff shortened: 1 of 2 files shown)


waveywaves:
Would we have to pull the image down and then use that blob to use for the verification
Also another question about cosign.pub. From what I understand providing a local cosign.pub is deprecated (?). Would a user have to provide another public key for verification using a env var or a flag during `package create` or something else.

Racer159:
> Would we have to pull the image down and then use that blob to use for the verification Also another question about cosign.pub. From what I understand providing a local cosign.pub is deprecated (?). Would a user have to provide another public key for verification using a env var or a flag during `package create` or something else.
> 
> cc @Racer159

The cosign.pub verification to my understanding isn't deprecated (would be interesting to see where you saw that - I don't see anything in cosign running locally), but also isn't the primary method to verify cosign signatures because it doesn't use the more modern keyless infrastructure.  For Zarf's use cases though (working in an airgap/offline) we have largely supported the cosign keys since they are much easier to take into airgaps to perform verification.

This particular issue may require more design though before we can tackle it (it was originally split from an issue that had a few separate issues in it) since there are a few scenarios we should consider.  Many organizations may have a common key they resign things with in a registry mirror like artifactory but there would likely be many more that would want to manage different keys or different verification schemes for their images (i.e. if they had images from multiple sources or teams).  CC @eddiezane for his thoughts on this as well.

waveywaves:
@Racer159 saw that cosignPublicKeyPath was deprecated here

https://github.com/defenseunicorns/zarf/blob/main/src/types/component.go#L34-L35

schristoff:
Hey @waveywaves - thank you so much for this pull request. We appreciate your work on this. I'm going to close this since it is over 6 months old - but if you decide to open a new PR you can still reference it. :)

PR #2327, opened 2024-02-22
update find-images output to contain the images' digest

### LINKED ISSUE #2247: During `zarf dev find-images` automatically add the digest to any tag to help with pinning resources
**As** Ezra **I want** to be able to get the latest digest for a given tag automatically when running `zarf dev find-images` **so that** I do not need to manually do this if I want to pin an image to a specific SHA (as `zarf dev lint` would want me to do)

- **Given** I have a Zarf package with `charts` or `manifests` that need images
- **When** I run `zarf dev find-images` on that package definition
- **Then** Zarf will find the image references that it can
- **And** Zarf will change any non-SHA-pinned images into a SHA pinned reference when shown to the user.

i.e.

```
components:
  - name: baseline
    images:
      - defenseunicorns/zarf-game:multi-tile-dark
```

becomes

```
components:
  - name: baseline
    images:
      - defenseunicorns/zarf-game:multi-tile-dark@sha256:f78e442f0f3eb3e9459b5ae6b1a8fda62f8dfe818112e7d130a4e8ae72b3cbff
```

We could do this automatically on create but that may get annoying for people trying to craft a package at a specific version.

### Additional context
This should base itself off of the `charts` and `manifests` as defined and should not take the existing list of `images` into account.  It should always check the remote registry for the latest manifest SHA at that tag when run. (if the user wishes to hard-pin the image they can edit their values files or manifests to do so)

waveywaves:
## Description

update find-images output to contain the images' digest

## Related Issue

Fixes #2247

## Type of change

## Checklist before merging


src/pkg/packager/prepare.go
```diff
@@ -211,7 +211,14 @@ func (p *Packager) FindImages() (imgMap map[string][]string, err error) {
 			for _, image := range sortedImages {
 				// Use print because we want this dumped to stdout
 				imagesMap[component.Name] = append(imagesMap[component.Name], image)
-				componentDefinition += fmt.Sprintf("      - %s\n", image)
+				imageWithSha, err := utils.GetImageWithSha(image)
+				if err != nil {
+					message.WarnErrf(err, "Problem getting image with sha for %s: %s", image, err.Error())
+					erroredCosignLookups = append(erroredCosignLookups, image)
```
AustinAbro321, line 217: Cosign signatures are different than sha signatures. I recommend creating a new array for erroredShaLookups. You can place a message near erroredCosginLookups in the below block further down in the file

src/pkg/utils/image.go
```diff
@@ -7,15 +7,32 @@ package utils
...
+// GetImageWithSha returns the image reference and the sha256 digest of the image.
+func GetImageWithSha(imgSrc string) (string, error) {
+	regex := regexp.MustCompile(`^.*@sha256:[a-f0-9]{64}$`)
+	if regex.MatchString(imgSrc) {
+		return imgSrc, nil
+	}
+
+	imgDescriptor, err := crane.Head(imgSrc)
+	if err != nil {
+		return "", fmt.Errorf("unable to pull image (%s) to get the sha256 digest: %w", imgSrc, err)
```
AustinAbro321, line 30:
```suggestion
		return "", err
```
IMO it would be better to just return err here, since you're already giving an error message in prepare.go it ends up repeating itself currently
```
Problem getting image with sha for ghcr.io/stefanprodan/podinfo:6.14.0: unable to pull
          image (ghcr.io/stefanprodan/podinfo:6.14.0) to get the sha256 digest: HEAD
          https://ghcr.io/v2/stefanprodan/podinfo/manifests/6.14.0: unexpected status code 404 Not
          Found (HEAD responses have no body, use GET for details)
```

src/pkg/packager/prepare.go
```diff
@@ -228,14 +235,6 @@ func (p *Packager) FindImages() (imgMap map[string][]string, err error) {
...
 				}
 			}
-
-			if len(validImages) > 0 {
-				componentDefinition += fmt.Sprintf("      # Possible images - %s - %s\n", p.cfg.Pkg.Metadata.Name, component.Name)
-				for _, image := range validImages {
-					imagesMap[component.Name] = append(imagesMap[component.Name], image)
-					componentDefinition += fmt.Sprintf("      - %s\n", image)
-				}
-			}
```
AustinAbro321, line 238: I believe we should still keep this. Any reason it was deleted?

AustinAbro321 (changes requested):

src/pkg/packager/prepare.go
waveywaves: reverted this change

src/pkg/packager/prepare.go
waveywaves: added a check with erroredShaLookups

PR #2328, opened 2024-02-22
feat: add `--config` flag, an alternative to `ZARF_CONFIG` env var

### LINKED ISSUE #2178: alternative zarf config path as a CLI option
While you can use the ZARF_CONFIG environment variables it's only findable through reading the docs, which many are not likely to do. 

- **Given** I have a zarf-config.toml at path/to/zarf-config.toml
- **When** I run this command or simlar zarf package create --config path/to/zarf-config.toml
- **Then** my zarf config file is read and used

Keeping it how it is currently. 

### Additional context
I think there's also the potential argument that if you're going to run zarf package create /path/to/zarffile then we should look for the zarf file in that directory

waveywaves:
## Description

add `--config` flag, an alternative to `ZARF_CONFIG` env var

## Related Issue

Fixes #2178

## Type of change

## Checklist before merging


src/cmd/root.go
```diff
@@ -82,4 +82,5 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&config.CommonOptions.CachePath, "zarf-cache", v.GetString(common.VZarfCache), lang.RootCmdFlagCachePath)
 	rootCmd.PersistentFlags().StringVar(&config.CommonOptions.TempDirectory, "tmpdir", v.GetString(common.VTmpDir), lang.RootCmdFlagTempDir)
 	rootCmd.PersistentFlags().BoolVar(&config.CommonOptions.Insecure, "insecure", v.GetBool(common.VInsecure), lang.RootCmdFlagInsecure)
+	rootCmd.PersistentFlags().StringVar(&config.CommonOptions.ConfigPath, "config-path", v.GetString(common.VConfigPath), lang.RootCmdFlagConfigPath)
```
AustinAbro321, line 85:
```suggestion
	rootCmd.PersistentFlags().StringVar(&config.CommonOptions.ConfigPath, "config-path", "", lang.RootCmdFlagConfigPath)
```
the `v.GetString` function gets a value from the configFile, because we are setting the config file here we shouldn't use the `v.GetString`

src/test/e2e/30_config_file_test.go
```diff
@@ -164,3 +165,58 @@ func configFileDefaultTests(t *testing.T) {
 
 	os.Unsetenv("ZARF_CONFIG")
 }
+
+func configFileTestsWithFlag(t *testing.T, dir, path, configPath string) {
```
AustinAbro321, line 169: There is a lot of copying and pasting in this test, it may be better to simply check one variable. This function could then test just the --config-path flag and the other function will test that the config file works as intended.

src/cmd/common/setup.go
AustinAbro321, line 1: I'm not sure exactly where the issue is, but just testing this out locally it seems that if there is a config file in the base directory, that takes precedence over the value set by --config-path

AustinAbro321 (changes requested):

src/test/e2e/30_config_file_test.go
waveywaves: just added this in the latest commit

src/cmd/common/setup.go
waveywaves: which command are you running ?

waveywaves: I ran this command locally it's working well, the other config file in the base directory is not taking precedence in my environment

AustinAbro321: Ah I see it is not that it takes precedence it is that it uses both config files. The config file pointed to by the command flag, and the config file in the base directory. 
To see an example of this in the base directory of the repo you can run `build/zarf package create . --config-path examples/config-file/zarf-config.toml` and you will see that all the values from the configfile in the base directory get propagated through. If you step through the debugger it will show the values from both config files went through.

waveywaves: I know where this is happening
```
	// Don't forget to read config either from cfgFile or from home directory!
	if cfgFile != "" {
		// Use config file from the flag.
		v.SetConfigFile(cfgFile)
	} else {
		// Search config paths in the current directory and $HOME/.zarf.
		v.AddConfigPath(".")
		v.AddConfigPath("$HOME/.zarf")
		v.SetConfigName("zarf-config")
	}

```

waveywaves: I have added some logic which fixes this behavior

AustinAbro321:
Hey @waveywaves sorry for the delay but I would like to get this merged in at some point, I think it's pretty close. Could you merge main into this branch? There are a lot of conflicts but they're all from the new docs site so merging in main and running make docs-and-schema should fix it.

Also could you add the following line to the appropriate block in src/cmd/internal.go
```
addHiddenDummyFlag(toolCmd, "config-path")
```

src/test/e2e/29_config_file_test.go
```diff
@@ -18,17 +18,23 @@ func TestConfigFile(t *testing.T) {
...
+		config     = "zarf-config.toml"
+		configPath = filepath.Join(dir, config)
 	)
+	// defer unsetting the ZARF_CONFIG so that we can test w/ the env var and w/ the flag both
+	defer os.Unsetenv("ZARF_CONFIG")
 
 	e2e.CleanFiles(path)
 
+	// Test the config file flag
+	os.Unsetenv("ZARF_CONFIG")
```
AustinAbro321, line 32: This was originally necessary but can now be deleted. 
IMO each test should expect no environment variables to have been set it other tests as was happening before

src/test/e2e/29_config_file_test.go
```diff
@@ -38,15 +44,17 @@ func TestConfigFile(t *testing.T) {
 	e2e.CleanFiles(path)
 }
 
-func configFileTests(t *testing.T, dir, path string) {
+func configFileTests(t *testing.T, dir, path string, configPath string) {
```
AustinAbro321, line 47: A bit of a nit, but I think it would be simpler if the function signature was
```go
func configFileTests(t *testing.T, dir, path string, extraArgs ...string)
```
Then it could be called like below
```go
configFileTests(t, dir, path, "--config-path", configPath)
```
and the args could simply be appended like this
```go
args := append([]string{"package", "create", dir, "--confirm"}, extraArgs...)
```

AustinAbro321 (changes requested):
Thank you for getting back to this, it is a great feature to have

schristoff:
Hey, since this PR is stale we are going to close it at this time but feel free to reference it in a newer PR :)

PR #2334, opened 2024-02-27
chore: Add PIE Build Mode Artifacts to GoReleaser Configuration

naveensrinivasan:
## Description
Added the GOFLAGS="-buildmode=pie" environment variable to our GoReleaser build configuration.

This change instructs the Go compiler to build the binary as a PIE, a crucial security feature.

PIE binaries are more resistant to specific attacks because they support Address Space Layout Randomization (ASLR), making it significantly harder for attackers to predict the memory location of specific code segments.

This change applies to all our target operating systems (Linux, Darwin/ macOS, and Windows), aligning with best practices for secure software development.

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -10,6 +10,7 @@ universal_binaries:
 builds:
   - env:
       - CGO_ENABLED=0
+      - GOFLAGS="-buildmode=pie" # Building a Position Independent Executable (PIE)
     goos:
       - linux
       - darwin


.goreleaser.yaml
```diff
@@ -10,6 +10,7 @@ universal_binaries:
 builds:
   - env:
       - CGO_ENABLED=0
+      - GOFLAGS="-buildmode=pie" # Building a Position Independent Executable (PIE)
```
Racer159, line 13: Doing this alone will break `zarf` being a statically compiled binary (which will break when used in the `scratch` image used in the agent or for users on alpine, etc):
image

Looking into it a bit there are ways to make a `pie` executable static, but they all seem to require `cgo` which has tradeoffs to consider as a team:

1. Cross-compilation is much more complicated requiring a c compiler for the target platform
2. The binary size is inflated by about 30 MiBs
3. We have to be much more careful that any libraries we use do not link to external libraries - this is possible for `os/user` and `net` with tags but anything else we bring in we would need to check.

Here is the command that builds zarf as a static pie binary for reference (omitting the version `-X` junk):

```
GOOS=linux GOARCH=amd64 go build -ldflags="-linkmode external -s -w -extldflags '--static-pie'" -buildmode=pie -tags 'osusergo,netgo,static_build' -o build/zarf .
```

@jeff-mccoy tagging you in on this as well for any historical input you may have on PIE binary explorations with Zarf.  Given the above limitations I'm leaning against it for now.

Racer159 (changes requested):

.goreleaser.yaml
jeff-mccoy: Are we sure that PIE is actually that valuable when not using dynamically linked libraries? We always need to optimize for static/run anywhere and it also seems as @Racer159 pointed out that this might get messy.

bburky: PIE is a requirement for ASLR. PIE means the executable can be loaded anywhere in memory, which Linux will do if possible through ASLR.

> Are we sure that PIE is actually that valuable when not using dynamically linked libraries?

To get the most value, you want all modules to be loaded with ASLR, so doing it only on one module (no dynamically linked libraries) is the easy case. It's good.

> Doing this alone will break zarf being a statically compiled binary which will break when used in the scratch image used in the agent or for users on alpine, etc)

There's a newer mode called `-static-pie` which is probably what you want. It's for exactly what you're trying to do. I set it up on a much smaller golang project once:
https://repo1.dso.mil/dsop/ironbank-pipelines/pre-receive/-/blob/98dcd09237cc1c729bd5efbcf2c7b22c7605b27d/Dockerfile.static#L18

But, beware that `-static-pie` is less common and some tools may incorrectly show it as "dynamically linked" (usually older ones, and apparently as Alpine's `ldd`). I think technically a dynamic loader puts a static-pie into memory, thus sort-of-dynamically-linked, but it depends on no external dynamically linked libraries.

Alpine on a `-static-pie` binary, with `ldd` being misleading or possibly lying to me:
```
# file pre-receive 
pre-receive: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), static-pie linked, BuildID[sha1]=80ed0b1461d66be4f0c3cba9a39b29ece080edf9, with debug_info, not stripped
# ldd pre-receive 
        /lib/ld-musl-aarch64.so.1 (0xffff88019000)
```

Ubuntu, same binary, clearer output:
```
# file pre-receive 
pre-receive: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), static-pie linked, BuildID[sha1]=80ed0b1461d66be4f0c3cba9a39b29ece080edf9, with debug_info, not stripped
# ldd pre-receive 
        statically linked
```
(Note that `ldd`'s output `statically linked` for static-pie is different than completely static: `not a dynamic executable`)

@Racer159 It looks like your go build flags are already using `-static-pie`. Can you inspect the binary with an up to date `file` and `ldd` command (copy the binary into an `ubuntu:latest` container to test). You may already be done.

That said, it's really just for security hardening. The value is usually that it often forces an attacker to find an info leak to discover a pointer offset in addition to an exploit to gain code execution.

Don't switch to CGO over it, it's probably not that important. But you should be able to get `-static-pie` to work.

Racer159: Yep it is a static binary when using `-static-pie` though from what I was messing with seemed to require CGO and `-linkmode external` to actually build that out.  When removing those it falls back to a dynamically linked binary:
image

Instead of the static one:
image

bburky: Ah, sorry. It looks like did actually build with CGO when I did `static-pie`. So never mind, my golang build experiecne isn't directly applicable. I did manage manually to compile the one external dependency manually and link it statically though. I could see how that would be hard if zarf has a lot of dependencies though.

Also, one note: I used `-tags 'osusergo,netgo,static'` not `-tags 'osusergo,netgo,static_build'`, I'm not quite sure what the difference was but `static_build` didn't work.

lucasrod16: Also a relevant thread https://groups.google.com/g/golang-nuts/c/cXhRsmNsMwo?pli=1

> PIE makes certain kinds of attacks on programs significantly more
difficult. Those attacks must first find a bug in the program,
typically an overflow of a buffer stored on the stack, which lets the
attacker modify the stack in some way. If such a bug exists, using
PIE makes it harder to exploit the bug. In Go it is very hard to
write a program that contains such a bug, much much harder than it is
in C. Therefore, I would say that PIE is much less important for Go
than it is for C. But it is possible with a fair amount of work to
write a Go program that contains such a bug. And Go programs can call
C code, and it might be useful to use PIE to make it harder to exploit
any bugs in that C code. So PIE is not completely useless for Go.

bburky: It's basically impossible I think. I wouldn't be surprised if everything is dynamically linked on Windows and Linux which explains those OSs.
* CGO_ENABLED=0 -buildmode=pie (default -linkmode=internal on linux)
   Creates dynamically linked, depends on libc
* CGO_ENABLED=0 -buildmode=pie -linkmode=external -extldflags '-static-pie'
   error: `-linkmode requires external (cgo) linking, but cgo is not enabled`
* CGO_ENABLED=1 -buildmode=pie -linkmode=external -extldflags '-static-pie'
   Works, creates static-pie, but this is CGO
* just do default golang static like you do today I think.
  Creates static, but isn't PIE
  
I guess just stick with static non-PIE unless CGO is acceptable.

PR #2340, opened 2024-02-28
chore: fix bb test

AustinAbro321:
## Description

Currently when trying to run the BB test outside of CI with
```bash
go test ./src/extensions/bigbang/test -failfast -v -timeout 30m
```
You get
```bash
Error: accepts at most 1 arg(s), received 2
```
We are also getting disk pressure errors in actions. Would like to come to a conclusions as a team on how we handle. Remove default tools or get bigger runners

## Type of change

## Checklist before merging


Racer159 (approved):
lgtm

PR #2344, opened 2024-03-01
chore: fixes potential cross site scripting

naveensrinivasan:
## Description

- fixes  `G203: The used method does not auto-escape HTML. This can potentially lead to 'Cross-site Scripting' vulnerabilities in case the attacker controls the input. (gosec)`

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/internal/packager/sbom/viewer.go
+++ b/src/internal/packager/sbom/viewer.go
@@ -31,8 +31,8 @@ func (b *Builder) createSBOMHTML(filename string, goTemplate string, jsonData []
 
 	defer sbomViewerFile.Close()
 
-	safeJsonData := template.JSEscapeString(string(jsonData))
-	safeJsonList := template.JSEscapeString(string(b.jsonList))
+	safeJSONData := template.JSEscapeString(string(jsonData))
+	safeJSONList := template.JSEscapeString(string(b.jsonList))
 
 	// Create the sbomviewer template data
 	tplData := struct {
@@ -47,8 +47,8 @@ func (b *Builder) createSBOMHTML(filename string, goTemplate string, jsonData []
 	}{
 		ThemeCSS:  b.loadFileCSS("theme.css"),
 		ViewerCSS: b.loadFileCSS("styles.css"),
-		List:      template.JS(safeJsonList),
-		Data:      template.JS(safeJsonData),
+		List:      template.JS(safeJSONList),
+		Data:      template.JS(safeJSONData),
 		LibraryJS: b.loadFileJS("library.js"),
 		CommonJS:  b.loadFileJS("common.js"),
 		ViewerJS:  b.loadFileJS("viewer.js"),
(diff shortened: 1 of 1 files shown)


src/internal/packager/sbom/viewer.go
```diff
@@ -44,8 +47,8 @@ func (b *Builder) createSBOMHTML(filename string, goTemplate string, jsonData []
 	}{
 		ThemeCSS:  b.loadFileCSS("theme.css"),
 		ViewerCSS: b.loadFileCSS("styles.css"),
-		List:      template.JS(b.jsonList),
-		Data:      template.JS(jsonData),
+		List:      template.JS(safeJsonList),
+		Data:      template.JS(safeJsonData),
```
Racer159, line 51:
```suggestion
		List:      template.JS(safeJSONList),
		Data:      template.JS(safeJSONData),
```
linting

Racer159 (changes requested):

src/internal/packager/sbom/viewer.go
```diff
@@ -31,6 +31,9 @@ func (b *Builder) createSBOMHTML(filename string, goTemplate string, jsonData []
 
 	defer sbomViewerFile.Close()
 
+	safeJsonData := template.JSEscapeString(string(jsonData))
+	safeJsonList := template.JSEscapeString(string(b.jsonList))
```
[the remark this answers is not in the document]
naveensrinivasan: Thanks. I took care of this.

src/internal/packager/sbom/viewer.go
naveensrinivasan: Thanks. I took care of this.

src/internal/packager/sbom/viewer.go
```diff
@@ -44,8 +47,8 @@ func (b *Builder) createSBOMHTML(filename string, goTemplate string, jsonData []
 	}{
 		ThemeCSS:  b.loadFileCSS("theme.css"),
 		ViewerCSS: b.loadFileCSS("styles.css"),
-		List:      template.JS(b.jsonList),
-		Data:      template.JS(jsonData),
+		List:      template.JS(safeJSONList),
+		Data:      template.JS(safeJSONData),
```
Racer159, line 51: Should have checked earlier, but this actually breaks the way we are including these values into the sbom html files:

image

(note the escaped values)

image

simple repro:

```
zarf package create examples/dos-games/
zarf package inspect zarf-package-dos-games-amd64-1.0.0.tar.zst -s
```

Racer159 (changes requested):

Racer159:
After more thought and discussion, closing this since the threat surface is lower (would require modifying syft json data and there aren't any creds to steal on the site given it is a static HTML page).  Fixing this would likely require more effort than it is worth at this juncture.

PR #2376, opened 2024-03-11
fix(release): hotfix `publish` not respecting source package architecture

Noxsios:
## Description

`zarf package publish <src>` would publish to CLI/runtime arch only and not respect source package's architecture, leading to packages existing at the wrong platform in the published index.

## Type of change

## Checklist before merging


Racer159 (approved):
lgtm

PR #2223, opened 2024-01-11
fix: refactor create stages into separate lib

lucasrod16:
## Description
refactor create stages into separate lib

## Related Issue

Fixes https://github.com/defenseunicorns/zarf/issues/2188

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Other (security config, docs update, etc)

## Checklist before merging

- [x] Test, docs, adr added or updated as needed
- [x] [Contributor Guide Steps](https://github.com/defenseunicorns/zarf/blob/main/CONTRIBUTING.md#developer-workflow) followed

PR #2386, opened 2024-03-19
fix: populate `p.sbomViewFiles` on `deploy` and `mirror`

lucasrod16:
## Description
Refactoring that took place in https://github.com/defenseunicorns/zarf/pull/2223 introduced a bug on `deploy` where the user is not prompted to view SBOMs. This occurs because `p.sbomViewFiles` is not being populated correctly. This PR fixes this issue by populating `p.sbomViewFiles`.

## Related Issue
N/A

## Type of change

## Checklist before merging

### DIFF
--- a/src/pkg/layout/sbom.go
+++ b/src/pkg/layout/sbom.go
@@ -69,15 +69,16 @@ func (s *SBOMs) Archive() (err error) {
 }
 
 // StageSBOMViewFiles copies SBOM viewer HTML files to the Zarf SBOM directory.
-func (s *SBOMs) StageSBOMViewFiles() (warnings []string, err error) {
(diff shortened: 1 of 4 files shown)


AustinAbro321 (dismissed):
LGTM

Noxsios (changes requested):
let's throw a `require.NotContains` in the `init` package test (or in a test we know for sure must contains SBOMs) to ensure it always has SBOMs. Not a 100% great test, but should prevent this regression in the future.

lucasrod16:
> let's throw a `require.NotContains` in the `init` package test (or in a test we know for sure must contains SBOMs) to ensure it always has SBOMs. Not a 100% great test, but should prevent this regression in the future.

Good call. Adding that now

Noxsios (approved):

Issue #2365, opened 2024-03-07
refactor: Plumb through context


eddiezane:
### Describe what should be investigated or refactored

As of opening this issue there are ~50 uses of `context.TODO` throughout the codebase. We should refactor these library functions to accept a context.

### Links to any relevant code

https://github.com/search?q=repo%3Adefenseunicorns%2Fzarf+NOT+path%3A*test*+context.TODO&type=code&p=1

### Additional context

While we could create additional `GetPodsWithContext` functions it might be better to bite the bullet and make this as part of a breaking change.

phillebaba:
Relates to #2592 and #2594

PR #2405, opened 2024-03-26
refactor: context usage in k8s code

lucasrod16:
## Description
refactor context usage in k8s code

## Related Issue
Relates to https://github.com/defenseunicorns/zarf/issues/2365

## Type of change

## Checklist before merging


src/cmd/destroy.go
```diff
@@ -29,15 +31,21 @@ var destroyCmd = &cobra.Command{
 	Short:   lang.CmdDestroyShort,
 	Long:    lang.CmdDestroyLong,
 	Run: func(_ *cobra.Command, _ []string) {
-		c, err := cluster.NewClusterWithWait(cluster.DefaultTimeout)
+		ctxLong, cancelLong := context.WithTimeout(context.Background(), 5*time.Minute)
```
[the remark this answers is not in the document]
lucasrod16: fixed

src/cmd/tools/crane.go
```diff
@@ -217,20 +224,23 @@ func pruneImages(_ *cobra.Command, _ []string) error {
...
 	// Load the currently deployed packages
-	zarfPackages, errs := c.GetDeployedZarfPackages()
+	zarfPackages, errs := c.GetDeployedZarfPackages(ctx)
 	if len(errs) > 0 {
 		return lang.ErrUnableToGetPackages
 	}
 
 	// Set up a tunnel to the registry if applicable
-	registryEndpoint, tunnel, err := c.ConnectToZarfRegistryEndpoint(zarfState.RegistryInfo)
+	registryEndpoint, tunnel, err := c.ConnectToZarfRegistryEndpoint(ctx, zarfState.RegistryInfo)
```
AustinAbro321, line 243: I feel like it's very unlikely a few secrets + a svc times this out at 30s but it still gives me a little pause since people might be in uber slow environments. Especially since there's no override atm and we're combining multiple actions under the same timeout.

Are we able to mimic a situation locally where a timeout would be beneficial? I'm assuming if we can't connect to the cluster we will error pretty quick, but I wonder if there's a state our users could get stuck in. That would convince me we should add the timeouts.

AustinAbro321 (changes requested):
Looks like you're still making changes, so I'll come back to this later and leave these comments for now

src/pkg/cluster/common.go
```diff
@@ -27,9 +28,9 @@ var labels = k8s.Labels{
 	config.ZarfManagedByLabel: "zarf",
 }
 
-// NewClusterOrDie creates a new Cluster instance and waits up to 30 seconds for the cluster to be ready or throws a fatal error.
-func NewClusterOrDie() *Cluster {
-	c, err := NewClusterWithWait(DefaultTimeout)
+// NewClusterOrDie creates a new Cluster instance and waits for the cluster to be ready or throws a fatal error.
+func NewClusterOrDie(ctx context.Context) *Cluster {
```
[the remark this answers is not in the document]
lucasrod16: fixed

src/cmd/tools/crane.go
lucasrod16: fixed

AustinAbro321 (changes requested):
Looking mostly good. The main thing is a bunch of uses of `timeout.After` which are causing extra waits on the first attempt of for loops. I just put a comment on one of them, but there's a similar block in several places. I think that's whats making the tests take longer as well

src/pkg/k8s/pods.go
```diff
@@ -72,111 +74,120 @@ func (k *K8s) DeletePods(target PodLookup) error {
...
 }
 
 // WaitForPodsAndContainers attempts to find pods matching the given selector and optional inclusion filter
-// It will wait up to 90 seconds for the pods to be found and will return a list of matching pod names
+// It will wait up to 30 seconds for the pods to be found and will return a list of matching pod names
 // If the timeout is reached, an empty list will be returned.
-func (k *K8s) WaitForPodsAndContainers(target PodLookup, include PodFilter) []corev1.Pod {
-	for count := 0; count < waitLimit; count++ {
+func (k *K8s) WaitForPodsAndContainers(ctx context.Context, target PodLookup, include PodFilter) []corev1.Pod {
+	waitCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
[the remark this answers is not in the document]
lucasrod16: The `waitLimit` timeout is set to 30 seconds, so there is no change here. The code comment that said 90 seconds was not accurate

https://github.com/defenseunicorns/zarf/blob/b4201718aa96f34f37dac7afa5ccfe59affedbd0/src/pkg/k8s/pods.go#L17

src/pkg/k8s/namespace.go
```diff
@@ -45,19 +45,20 @@ func (k *K8s) DeleteNamespace(ctx context.Context, name string) error {
...
-	// Indefinitely wait for the namespace to be deleted, use context.WithTimeout to limit this
 	for {
-		// Keep checking for the namespace to be deleted
-		_, err := k.Clientset.CoreV1().Namespaces().Get(ctx, name, metav1.GetOptions{})
-		if errors.IsNotFound(err) {
-			return nil
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(1 * time.Second):
```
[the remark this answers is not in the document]
lucasrod16: Yeah good point about the initial wait. We should be using `time.After()` rather than `time.Sleep()` though because it is non-blocking. We can change the order of operations so that there is no pause before executing the logic we want:

```go
for {
      _, err := k.Clientset.CoreV1().Namespaces().Get(ctx, name, metav1.GetOptions{})
      if errors.IsNotFound(err) {
          return nil
      }
      select {
      case <-ctx.Done():
          return ctx.Err()
      case <-time.After(1 * time.Second):
      }
}
```

src/pkg/k8s/pods.go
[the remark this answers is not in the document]
lucasrod16: Scratch that. It should be 90 seconds actually. I wasn't accounting for the 3 second sleep between each iteration. I'll set it to 90 seconds

src/pkg/k8s/namespace.go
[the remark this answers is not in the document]
AustinAbro321: A relatively rare but possible case is the context getting cancelled by the parent function instead of by timeout. In that case, ctx could be done before the first loop and the caller would expect ctx.Err() returned rather than a successful attempt. I think I've found the pattern we're looking for actually

```
func main() {
	start := time.Now()
	instantButCancel()
	fmt.Println(time.Since(start))
	instantCancel()
}

func instantButCancel() {
	ctx, _ := context.WithTimeout(context.TODO(), 500 * time.Millisecond)
	timer := time.NewTimer(0) 
	for {
		select {
		case <-ctx.Done():
			return 
		case <-timer.C:
			fmt.Println("called by instantNoWait")
			timer.Reset(1 * time.Second)
		}
	}
}

func instantCancel() {
	ctx, cancel := context.WithCancel(context.TODO())
	cancel()
	timer := time.NewTimer(0) 
	for {
		select {
		case <-ctx.Done():
			fmt.Println("in done block instant cancel")
			return 
		case <-timer.C:
			fmt.Println("not called by instantCancel")
			timer.Reset(1 * time.Second)
		}
	}
}
```

Running this yields which I believe is everything we want
```
called by instantNoWait
500.650383ms
in done block instant cancel
```

src/pkg/k8s/pods.go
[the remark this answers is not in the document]
lucasrod16: fixed

src/pkg/k8s/namespace.go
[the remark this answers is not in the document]
lucasrod16: That seems reasonable. Where would the actual logic get executed with this pattern?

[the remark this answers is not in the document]
AustinAbro321: Under timer.C like this, goes in instantly first loop, all loops after that have a 1 second delay and are nonblocking on ctx
```
	timer := time.NewTimer(0) 
	for {
		select {
		case <-ctx.Done():
			fmt.Println("in done block instant cancel")
			return 
		case <-timer.C:
			fmt.Println("do the logic here")
			timer.Reset(1 * time.Second)
		}
	}
```

src/pkg/k8s/pods.go
```diff
@@ -72,111 +79,125 @@ func (k *K8s) DeletePods(target PodLookup) error {
...
+	timer := time.NewTimer(0)
+	defer timer.Stop()
 
-		k.Log("Found %d pods for target %#v", len(pods.Items), target)
-
-		var readyPods = []corev1.Pod{}
+	for {
+		select {
+		case <-waitCtx.Done():
+			k.Log("Pod lookup timeout or context cancelled: %w", ctx.Err())
```
AustinAbro321, line 110:
```suggestion
			k.Log("Pod lookup failed: %w", ctx.Err())
```
timeout or context cancelled should already be clear from the ctx.Err()

src/pkg/k8s/sa.go
```diff
@@ -15,48 +15,50 @@ import (
...
-
-		default:
-			sa, err := k.Clientset.CoreV1().ServiceAccounts(ns).Get(context.TODO(), name, metav1.GetOptions{})
-			if err != nil {
-				if errors.IsNotFound(err) {
-					time.Sleep(1 * time.Second)
-					continue
-				}
+		case <-ctx.Done():
+			return nil, fmt.Errorf("timed out waiting for service account %s/%s to exist: %w", ns, name, ctx.Err())
```
AustinAbro321, line 48:
```suggestion
			return nil, fmt.Errorf("failed to get service account %s/%s: %w", ns, name, ctx.Err())
```

AustinAbro321 (changes requested):
Pretty close I think, also looks like the e2e test time is back down

src/pkg/k8s/common.go
```diff
@@ -42,68 +43,54 @@ func New(logger Log, defaultLabels Labels) (*K8s, error) {
...
 	for {
 		select {
-		// on timeout abort
-		case <-expired:
-			return fmt.Errorf("timed out waiting for cluster to report healthy")
-
-		// after delay, try running
-		default:
+		case <-ctx.Done():
+			return fmt.Errorf("timed out waiting for cluster to report healthy: %w", ctx.Err())
```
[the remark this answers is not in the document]
lucasrod16: fixed

src/pkg/k8s/common.go
```diff
@@ -42,68 +43,54 @@ func New(logger Log, defaultLabels Labels) (*K8s, error) {
...
+				k.Log("No nodes reporting healthy yet: %v\n", err)
+				timer.Reset(1 * time.Second)
 			}
 
 			// Get the cluster pod list
-			if pods, err = k.GetAllPods(); err != nil {
+			if pods, err = k.GetAllPods(ctx); err != nil {
 				k.Log("Could not get the pod list: %w", err)
-				continue
+				timer.Reset(1 * time.Second)
```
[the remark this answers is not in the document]
lucasrod16: > I recommend keeping similar continue statements and a timer.Reset outside of the select loop

unreachable-code

Updated to use `timer.Reset()` and `continue` together

AustinAbro321 (dismissed):

AustinAbro321 (dismissed):

src/pkg/packager/common_test.go
```diff
@@ -132,7 +133,7 @@ func TestValidatePackageArchitecture(t *testing.T) {
 				return true, nodeList, nil
 			})
 
-			err := p.validatePackageArchitecture()
+			err := p.validatePackageArchitecture(context.TODO())
```
schristoff-du, line 136: Should this be switched to a normal context?

src/test/e2e/21_connect_creds_test.go
```diff
@@ -27,7 +28,9 @@ func TestConnectAndCreds(t *testing.T) {
 	prevAgentSecretData, _, err := e2e.Kubectl("get", "secret", "agent-hook-tls", "-n", "zarf", "-o", "jsonpath={.data}")
 	require.NoError(t, err)
 
-	connectToZarfServices(t)
+	ctx := context.TODO()
```
schristoff-du, line 31: Same TODO question as above

src/test/e2e/21_connect_creds_test.go
```diff
@@ -68,7 +71,7 @@ func TestMetrics(t *testing.T) {
 
 	tunnel, err := c.NewTunnel("zarf", "svc", "agent-hook", "", 8888, 8443)
 	require.NoError(t, err)
-	_, err = tunnel.Connect()
+	_, err = tunnel.Connect(context.TODO())
```
schristoff-du, line 74: Same TODO question as above (I know these ones are tests and the others weren't, but maybe still valid)

src/test/e2e/23_data_injection_test.go
```diff
@@ -42,7 +42,7 @@ func TestDataInjection(t *testing.T) {
 	// need target to equal svc that we are trying to connect to call checkForZarfConnectLabel
 	c, err := cluster.NewCluster()
 	require.NoError(t, err)
-	tunnel, err := c.Connect("kiwix")
+	tunnel, err := c.Connect(context.TODO(), "kiwix")
```
schristoff-du, line 45: Same TODO question as above

src/test/e2e/26_simple_packages_test.go
```diff
@@ -26,7 +27,7 @@ func TestDosGames(t *testing.T) {
 
 	c, err := cluster.NewCluster()
 	require.NoError(t, err)
-	tunnel, err := c.Connect("doom")
+	tunnel, err := c.Connect(context.TODO(), "doom")
```
schristoff-du, line 30: Same TODO question as above

src/test/e2e/30_config_file_test.go
```diff
@@ -137,6 +140,7 @@ func configFileDefaultTests(t *testing.T) {
 
 	// Test remaining default initializers
 	os.Setenv("ZARF_CONFIG", filepath.Join("src", "test", "zarf-config-test.toml"))
```
schristoff-du, line 142: Should be able to set these with `t.Setenv` , not a priority though

src/cmd/common/utils.go
```diff
@@ -35,3 +37,14 @@ func ExitOnInterrupt() {
 		}
 	}()
 }
+
+// NewClusterOrDie creates a new Cluster instance and waits for the cluster to be ready or throws a fatal error.
+func NewClusterOrDie() *cluster.Cluster {
```
schristoff-du, line 42: Is there a reason why we wouldn't want to pass the `ctx` we already have from `cmd` in here?

schristoff-du (changes requested):
This looks great! Apologies on the delay of the review. I think we could clean up a bunch of those `context.Todo()`, and I had a couple questions on implementation.

src/cmd/tools/helm/root.go
```diff
@@ -156,7 +156,7 @@ func NewRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
 	// This call is required to gather configuration information prior to
 	// execution.
 	flags.ParseErrorsWhitelist.UnknownFlags = true
-	flags.Parse(args)
+	_ = flags.Parse(args)
```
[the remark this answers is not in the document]
lucasrod16: Added this to resolve local linting error. Will remove as it is not consequential and is obviously confusing as to what the value is.

lint-error

src/extensions/bigbang/test/bigbang_test.go
```diff
@@ -119,18 +119,18 @@ func TestReleases(t *testing.T) {
 	require.NoError(t, err, stdOut, stdErr)
 
 	// Test connectivity to Twistlock
-	testConnection(t)
+	testConnection(context.TODO(), t)
```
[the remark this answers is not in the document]
lucasrod16: fixed

src/test/e2e/21_connect_creds_test.go
lucasrod16: fixed

src/test/e2e/21_connect_creds_test.go
lucasrod16: fixed

src/test/e2e/23_data_injection_test.go
lucasrod16: fixed

src/test/e2e/26_simple_packages_test.go
lucasrod16: fixed

src/test/e2e/30_config_file_test.go
lucasrod16: fixed

src/pkg/k8s/namespace.go
```diff
@@ -45,19 +45,25 @@ func (k *K8s) DeleteNamespace(ctx context.Context, name string) error {
...
+	timer := time.NewTimer(0)
+	defer timer.Stop()
+
 	for {
-		// Keep checking for the namespace to be deleted
-		_, err := k.Clientset.CoreV1().Namespaces().Get(ctx, name, metav1.GetOptions{})
-		if errors.IsNotFound(err) {
-			return nil
+		select {
+		case <-ctx.Done():
```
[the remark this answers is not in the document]
lucasrod16: No, there is no deadline set for this context. It's using the global command context `context.Background()`.

There is no timeout/deadline currently in Zarf for this, but we can add one here if you think one needs to be

src/pkg/k8s/pods.go
```diff
@@ -72,111 +79,125 @@ func (k *K8s) DeletePods(target PodLookup) error {
...
-		for _, pod := range pods.Items {
-			k.Log("Testing pod %q", pod.Name)
+			var readyPods = []corev1.Pod{}
 
-			// If an include function is provided, only keep pods that return true
-			if include != nil && !include(pod) {
-				continue
-			}
+			// Sort the pods from newest to oldest
+			sort.Slice(pods.Items, func(i, j int) bool {
```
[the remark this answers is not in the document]
lucasrod16: I think we should keep this. ChatGPT explains it better than me:

"In many Kubernetes deployments, newer pods are typically those that have just been created due to a scale-up action, a rolling update, or recovery from a failure. These newer pods are more likely to be in the desired state (e.g., running and ready) compared to older pods, which might be pending termination or in the process of being replaced during an update.

By sorting the pods from newest to oldest, the function starts checking the pods that are most likely to be ready first. This means there is a higher probability of quickly finding a pod that meets all the necessary conditions, thus allowing the function to return sooner without needing to check all pods."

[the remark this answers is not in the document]
lucasrod16: And from my own anecdotal local testing with the `TestDataInjection` E2E test, removing the sort logic slows this function down by a noticeable amount (5+ seconds on average).

src/pkg/k8s/pods.go
```diff
@@ -72,111 +79,125 @@ func (k *K8s) DeletePods(target PodLookup) error {
...
-					if isRunning && container.Name == target.Container {
+				// Handle container targeting
+				if target.Container != "" {
+					k.Log("Testing pod %q for container %q", pod.Name, target.Container)
+					var matchedContainer bool
+
+					// Check the status of initContainers for a running match
+					for _, initContainer := range pod.Status.InitContainerStatuses {
+						isRunning := initContainer.State.Running != nil
+						if isRunning && initContainer.Name == target.Container {
```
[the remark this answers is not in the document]
lucasrod16: Not sure there is a good reason, it is existing logic. I swapped them to check the name first

src/pkg/k8s/pods.go
```diff
@@ -72,111 +79,125 @@ func (k *K8s) DeletePods(target PodLookup) error {
...
+							// On running match in initContainer break this loop
+							matchedContainer = true
+							readyPods = append(readyPods, pod)
+							break
+						}
+					}
+					if matchedContainer {
```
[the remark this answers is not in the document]
lucasrod16: Your comment made me realize the `matchedContainer` boolean is not needed so I removed it. Should be more simple and straightforward now

src/pkg/packager/common_test.go
lucasrod16: Changed to use `context.Background()`

src/cmd/common/utils.go
lucasrod16: Good catch. Updated to pass in a `ctx`

src/cmd/dev.go
```diff
@@ -50,12 +50,12 @@ var devDeployCmd = &cobra.Command{
 		pkgConfig.PkgOpts.SetVariables = helpers.TransformAndMergeMap(
 			v.GetStringMapString(common.VPkgDeploySet), pkgConfig.PkgOpts.SetVariables, strings.ToUpper)
 
-		// Configure the packager
 		pkgClient := packager.NewOrDie(&pkgConfig)
 		defer pkgClient.ClearTempPaths()
 
-		// Create the package
-		if err := pkgClient.DevDeploy(); err != nil {
+		ctx := cmd.Context()
```
phillebaba, line 56: I think it is a bit verbose to define a ctx variable instead of just in lining the function call.

src/cmd/tools/helm/repo_add.go
```diff
@@ -128,11 +130,15 @@ func (o *repoAddOptions) run(out io.Writer) error {
 		lockPath = o.repoFile + ".lock"
 	}
 	fileLock := flock.New(lockPath)
-	lockCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	lockCtx, cancel := context.WithTimeout(context.Background(), cluster.DefaultTimeout)
```
phillebaba, line 133: Add context to run and pass cobra Command context to function.

src/extensions/bigbang/test/bigbang_test.go
```diff
@@ -60,77 +60,79 @@ func TestReleases(t *testing.T) {
 		zarfCache = fmt.Sprintf("--zarf-cache=%s", CIMount)
 	}
 
+	ctx := context.Background()
```
phillebaba, line 63: Use context.TODO() in tests.

src/pkg/cluster/injector.go
```diff
@@ -432,66 +436,61 @@ func (c *Cluster) buildInjectionPod(node, image string, payloadConfigmaps []stri
...
 }
 
-// GetImagesFromAvailableNodes checks for images on schedulable nodes within a cluster and returns
-func (c *Cluster) getImagesAndNodesForInjection(timeoutDuration time.Duration) (imageNodeMap, error) {
-	timeout := time.After(timeoutDuration)
+// getImagesAndNodesForInjection checks for images on schedulable nodes within a cluster.
+func (c *Cluster) getImagesAndNodesForInjection(ctx context.Context) (imageNodeMap, error) {
 	result := make(imageNodeMap)
 
+	timer := time.NewTimer(0)
```
phillebaba, line 443: This is an interesting thing todo. Is the purpose of this and the reset further down to make sure that the timer is immediate?

src/pkg/packager/common_test.go
```diff
@@ -132,7 +133,7 @@ func TestValidatePackageArchitecture(t *testing.T) {
 				return true, nodeList, nil
 			})
 
-			err := p.validatePackageArchitecture()
+			err := p.validatePackageArchitecture(context.Background())
```
phillebaba, line 136: Use context.TODO() in tests.

src/pkg/cluster/injector.go
lucasrod16: Yes, exactly. Using `time.NewTimer(0)` lets the timer start right away. Originally, we used `time.Sleep()`, but it's a blocking operation, which isn't ideal. I then switched to `time.After()` for better responsiveness since it doesn't block other operations. @AustinAbro321 suggested switching to `time.NewTimer` and it seems to work well for this use case

src/cmd/dev.go
lucasrod16: updated

src/cmd/tools/helm/repo_add.go
lucasrod16: updated

src/extensions/bigbang/test/bigbang_test.go
lucasrod16: I originally was using `context.TODO()` in the tests, but @schristoff-du asked if we should use a real context in the tests, and I wasn't sure of when you should or shouldn't use TODO, so switched back to using `Background()`.

Will switch back to using TODO in tests.

src/pkg/cluster/injector.go
schristoff: TODO is supposed to be a placeholder, in tests we can make a context and pass it through the functions to ensure we are tests the functions as close as they'll be used normally. In tests I have seen it more common to make a context, or pass a background.

src/extensions/bigbang/test/bigbang_test.go
schristoff: TODO is supposed to be a placeholder, in tests we can make a context and pass it through the functions to ensure we are tests the functions as close as they'll be used normally. In tests I have seen it more common to make a context, or pass a background.

schristoff: [Reference to source with notes on when to use Background vs TODO](https://github.com/golang/go/blob/master/src/context/context.go#L207)

lucasrod16:
> @lucasrod16 I had a deeper look at the k8s package and am now wondering why we even have it. Ignoring some helper functions, it for the most part just wraps a single call to the client set. In a sense it just increases complexity. My suggestion is to remove it and just use the client set directly.

Removing the k8s package could be something we discuss as a team, but it is out of scope for this PR. This PR is more strictly focused on how context flows through the program to the client set.

src/extensions/bigbang/test/bigbang_test.go
lucasrod16: @schristoff @phillebaba updated to use `context.Background()` in tests again. I don't think it really matters what which one we use in the tests. The docs for `Background` explicitly states that it's typically used in tests so that is likely the best option of the two.

phillebaba (approved):
I think this looks fine. I still have some thoughts about using `timer.Reset` but we can revisit that at a later PR. It is a lot more valuable to get this merge to unlock a lot of other fixes.

AustinAbro321 (approved):

schristoff-du (approved):

PR #2415, opened 2024-04-04
fix: filter on running pods when finding an image for injector pod

### LINKED ISSUE #2356: Zarf injector should filter on "running" pods when initializing
Recently, I had a k3s cluster get corrupted by the master of master nodes being deleted.  Even the zarf-docker-registry was corrupt, and I lost it and the images it contained.  I tried a zarf init, and the injector continued to clone pods that were in ImagePullBackoff of similar sates.  The timeout before moving on to another pod to try to clone was pretty long, so it took an enormous amount of time to start up the injector pod correctly and some manual finagling with taints to get on a node that had the fewest pods in error states.

- **Given** an existing cluster with many pods running, but also many that are in error states (like ImagePullBackoff)
- **When** running `zarf init`
- **Then** the injector filters out all pods but those that are healthy, in the "running" state to clone

An alternative might be to set the pod and/or node which to clone via an environment variable or --set.

### Additional context
The timeout for an injector pod to get to the "running" state could also be lowered or made overrideable by an env var or --set

### LINKED ISSUE #2410: Bug: Zarf fails to init successfully if there's a pod that's currently stuck in terminating state and the node on that terminating pod doesn't exist anymore
Zarf fails to init successfully if there's a pod that's currently stuck in terminating state and the node on that terminating pod doesn't exist anymore.

I think this chunk of code is related - it doesn't seem to check if a pod is in a bad state, i.e. terminating and possibly the node that it was supposed to live on is now gone 😬 

https://github.com/defenseunicorns/zarf/blob/08c92e12a4a2b05d0ea5abe055c7c01ba9964051/src/pkg/cluster/injector.go#L453-L466

I know this is a very specific scenario, but it happened -
- We had an EKS cluster and were upgrading it from 1.27 -> 1.29
- Rolled all node and eks cluster versions to 1.29
- Pod stuck in terminating
- This terminating `pod.Spec.NodeName` has a node on it that doesn't exist anymore because nodes were rolled to get new eks k8s version
- Zarf tries to run `GetNode` on this node from k8s api, that errors out
- Zarf prints error: `Unable to generate a list of candidate images to perform the registry injection`

lucasrod16:
## Description
filter on running pods when finding an image for injector pod

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase

Description of the `Running` pod phase:

> The Pod has been bound to a node, and all of the containers have been created. At least one container is still running, or is in the process of starting or restarting.

## Related Issue
Fixes #2356
Fixes #2410

## Type of change

## Checklist before merging


AustinAbro321 (dismissed):

src/pkg/k8s/pods.go
```diff
@@ -79,13 +79,15 @@ func (k *K8s) CreatePod(pod *corev1.Pod) (*corev1.Pod, error) {
...
 
 // GetPods returns a list of pods from the cluster by namespace.
-func (k *K8s) GetPods(namespace string) (*corev1.PodList, error) {
-	metaOptions := metav1.ListOptions{}
-	return k.Clientset.CoreV1().Pods(namespace).List(context.TODO(), metaOptions)
+func (k *K8s) GetPods(namespace string, listOpts *metav1.ListOptions) (*corev1.PodList, error) {
+	if listOpts == nil {
+		listOpts = &metav1.ListOptions{}
+	}
+	return k.Clientset.CoreV1().Pods(namespace).List(context.TODO(), *listOpts)
```
Noxsios, line 90: Should not mutate types from non-pointer to pointer when not needed: `List(ctx context.Context, opts metav1.ListOptions)`

Change this from `listOpts *metav1.ListOptions` to `listOpts metav1.ListOptions`, and instead of passing in `nil` when blank, use `metav1.ListOptions{}`

Noxsios (changes requested):

src/pkg/k8s/pods.go
lucasrod16: updated

Noxsios (approved):

PR #2419, opened 2024-04-05
refactor: move validate to expose it as receivers

Noxsios:
## Description

Moves `validate` behavior out of `internal` and onto receivers to their given types. Since validation requires no external logic, this fits better and has a cleaner experience. Additionally, validation can be called against each type itself individually instead of only living at the package level.

## Type of change

## Checklist before merging


src/types/validate.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+
+	return check(as.Before) || check(as.After) || check(as.OnSuccess) || check(as.OnFailure)
+}
+
+// Validate runs all validation checks on component action sets.
+func (as ZarfComponentActionSet) Validate() error {
+	validate := func(actions []ZarfComponentAction) error {
+		for _, action := range actions {
+			return action.Validate()
```
Noxsios, line 234: this should be `if err := action.Validate(); err != nil`, otherwise this function exits before it has validated all `actions`

lucasrod16 (approved):
lgtm

Issue #2425, opened 2024-04-08
Image Index Sha 


AustinAbro321:
## Introduction

Currently Zarf does not handle pulling index sha's for container images correctly. An index sha is a sha256 sum which is generated from an image index json file. View the definition of an image index here: https://github.com/opencontainers/image-spec/blob/main/image-index.md. Below is the image index for the zarf agent at version v0.32.6. There are three distinct digests that have been created at this tag. There is the arm64 digest, the amd64 digest, and the index digest. The index digest value is sha256 of the index.json. If the index sha is used without any additional options during a docker pull or crane pull the tool will pull down the correct image for your architecture. If an image index sha is used in Zarf it will break on deploy. 

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:b3fabdc7d4ecd0f396016ef78da19002c39e3ace352ea0ae4baa2ce9d5958376",
      "size": 673,
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:454bf871b5d826b6a31ab14c983583ae9d9e30c2036606b500368c5b552d8fdf",
      "size": 673,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    }
  ]
}
```

## Background

We create an image index in Zarf but it's used to specify all the images in a Zarf package rather than all the images at a specific tag. The image index within the Zarf package differs from image indexes registries store at tags (like the agent example above) in that it stores entirely different images rather than different platforms of images. It has all the images in the Zarf package. Because of this we annotate each entry in the manifest list with the name of the image we pull. This is what the current generated image index for a Zarf package looks like with one image at an index sha. The actual digest is different between the annotated digest. The actual digest, the one in the digest field begins with e81. The digest in the annotation field begins with ob6. This is because Zarf intended to pull down ob6, but in reality pulled down e81. Zarf uses crane to pull down images, crane will pull down the image according to the architecture specified by the user (or default user arch). This is why there is a mismatch. This is a bug

```json
  "manifests": [
    {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
[...]

Issue #2394, opened 2024-03-22
NginX Ingress Controller - provided digest did not match uploaded content


FrankWigger:
### Environment
Device and OS: Debian bookworm
zarf version: v0.32.5
Kubernetes distro being used: from zarf
Other:

### Steps to reproduce
1. Create a zarf.yaml withn ginx ingress controller.
2. Call zarf prepare find-images
3. zarf package create.
4. On target maschine call zarf package deplpy

### Expected result
Nginx ingress is up and running

### Actual Result
Zarf reports
 Retrying (1/3) in 5s: PUT
http://127.0.0.1:34819/v2/ingress-nginx/controller/manifests/sha256:42b3f0e5d0846876b1791cd3afeb5f1cbbe4259d6f35651dcc1b5c980925379c:
          DIGEST_INVALID: provided digest did not match uploaded content

### Zarf.yaml
My zarf.yaml looks like:
```
kind: ZarfPackageConfig # ZarfPackageConfig is the package kind for most normal zarf packages
metadata:
  name: ingress-nginx      
  version: v0.1.2  
  description: |        
    "A Zarf Package that deploys the nginx ingress controller"
components:
  - name: ingress-nginx
    description: |   
      "Deploys nginx ingress controller"
    required: true   
    charts:
      - name: ingress-nginx
        url: https://kubernetes.github.io/ingress-nginx
        version: 4.10.0
        namespace: ingress-nginx
    images:
      - registry.k8s.io/ingress-nginx/controller:v1.10.0@sha256:42b3f0e5d0846876b1791cd3afeb5f1cbbe4259d6f35651dcc1b5c980925379c
      - registry.k8s.io/ingress-nginx/kube-webhook-certgen:v1.4.0@sha256:44d1d0e9f19c63f58b380c5fddaca7cf22c7cee564adeff365225a5df5ef3334
      # Cosign artifacts for images - ingress-nginx - ingress-nginx
      - registry.k8s.io/ingress-nginx/controller:sha256-42b3f0e5d0846876b1791cd3afeb5f1cbbe4259d6f35651dcc1b5c980925379c.sig
      - registry.k8s.io/ingress-nginx/kube-webhook-certgen:sha256-44d1d0e9f19c63f58b380c5fddaca7cf22c7cee564adeff365225a5df5ef3334.sig
```

Any ideas?

AustinAbro321:
[...]

PR #2429, opened 2024-04-09
fix: error on create if an index sha is used

AustinAbro321:
## Description

Errors on create if an index sha is used and gives users options of all the other platforms they can use

## Related Issue

Relates to #2425, #2394 

## Checklist before merging


src/internal/packager/images/pull.go
```diff
@@ -308,6 +311,20 @@ func (i *ImageConfig) PullImage(src string, spinner *message.Spinner) (img v1.Im
 			return nil, false, fmt.Errorf("failed to load image from docker daemon: %w", err)
 		}
 	} else {
+		var idx v1.IndexManifest
+		if err := json.Unmarshal(manifest, &idx); err != nil {
```
schristoff, line 315: Is there a reason to not return the err from `json.Unmarshal` here? - It could be appended to the "lang.Err*" so we don't lose any context 🤔

src/test/e2e/14_create_sha_index_test.go
```diff
@@ -0,0 +1,24 @@
...
+// Package test provides e2e tests for Zarf.
+package test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCreateIndexShaErrors(t *testing.T) {
```
schristoff, line 13: For tests like these I see it more commonly to make them [table tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) - in this case I think it would improve readability and allow us to see what cases we are missing, but it isn't a big deal either. :)

src/pkg/packager/creator/normal.go
```diff
@@ -186,11 +187,15 @@ func (pc *PackageCreator) Assemble(dst *layout.PackagePaths, components []types.
 			}
 
 			pulled, err = imgConfig.PullAll()
+			if errors.Is(err, lang.ErrUnsupportedImageType) {
```
schristoff, line 190: I think it would be good to provide some context by providing this error

src/internal/packager/images/pull.go
```diff
@@ -308,6 +311,20 @@ func (i *ImageConfig) PullImage(src string, spinner *message.Spinner) (img v1.Im
...
+			return nil, false, lang.ErrUnsupportedImageType
+		}
+
+		if strings.Contains(src, "@") && (idx.MediaType == ctypes.OCIImageIndex || idx.MediaType == ctypes.DockerManifestList) {
+			imageOptions := "please select one of the images below based on your platform"
```
schristoff, line 320: Two things: 
I think we should be testing the output of `imageOptions` in our tests, but also I think this may read a little awkward without any line breaks.

schristoff (commented):
Added some pre-coffee thoughts. :)

src/internal/packager/images/pull.go
AustinAbro321: Yeah kinda the middle ground I've gone for is checking that one of the listed images in the image index is in the stderr in the e2e test. So for one of the tests I check that ghcr.io/defenseunicorns/zarf/agent:v0.32.6@sha256:b3fabdc7d4ecd0f396016ef78da19002c39e3ace352ea0ae4baa2ce9d5958376 is in stderr which is the arm64 image

Agree it's not perfect though and we see things like this in a lot of our e2e tests. Trying to balance readability / not locking tests to messages. Open to checking against more of the output or other suggestions.

src/pkg/packager/creator/normal.go
AustinAbro321: Sorry a bit confused by this suggestion. By context do you mean comments in code / logging to user / or adding values to the actual go context object?

src/test/e2e/14_create_sha_index_test.go
AustinAbro321: Yeah we don't usually do them in e2e tests, but this seems like a reasonable case that improves clarity. I like it

src/internal/packager/images/pull.go
```diff
@@ -308,6 +311,23 @@ func (i *ImageConfig) PullImage(src string, spinner *message.Spinner) (img v1.Im
...
+			}
+			var idx v1.IndexManifest
+			if err := json.Unmarshal(manifest, &idx); err != nil {
+				return nil, false, fmt.Errorf("%w: %w", lang.ErrUnsupportedImageType, err)
+			}
+			for _, manifest := range idx.Manifests {
+				imageOptions = fmt.Sprintf("%s\n %s@%s for platform %s", imageOptions, imageBaseName, manifest.Digest, manifest.Platform)
+			}
+			return nil, false, fmt.Errorf("%w: %s", lang.ErrUnsupportedImageType, imageOptions)
+		}
```
lucasrod16, line 329: A couple comments on this block of code:

  - I think it would be more clean and less error prone to use a library to parse the image reference
  - we should be able to use the `desc.Manifest` directly rather than fetch the manifest from remote again
  
     Example:

    ```go
    reference, err := name.ParseReference(src)
    if err != nil {
	    return nil, false, err
    }
    
    if _, ok := reference.(name.Digest); ok && (desc.MediaType == ctypes.OCIImageIndex || desc.MediaType == ctypes.DockerManifestList) {
	      var idx v1.IndexManifest
	      if err := json.Unmarshal(desc.Manifest, &idx); err != nil {
	          return nil, false, fmt.Errorf("%w: %w", lang.ErrUnsupportedImageType, err)
	      }
	      
	      imageBaseName := path.Join(reference.Context().Registry.RegistryStr(), reference.Context().RepositoryStr())
    
    }
    ```
    
    if we went this route, we would have to update the test to not require `stderr` to contain an image reference with the tag in it because images that contain a tag + sha are technically considered a digest rather than a tag, so the tag is essentially ignored

lucasrod16 (changes requested):

src/pkg/packager/creator/normal.go
schristoff: I was unclear, I would say "logging to user", but I'll never tell you not to add comments ;)

src/test/e2e/14_create_sha_index_test.go
schristoff: update: love this :)

src/pkg/packager/creator/normal.go
AustinAbro321: Ah yeah, the reason I don't log it here is because the error gets logged to the user during the retry function

src/internal/packager/images/pull.go
lucasrod16: I agree that `imageOptions` should be checked for in the test. There is currently a bug because `"please select one of the images below based on your platform"` isn't being printed to stderr, and testing for this output would have caught it quickly.

lucasrod16: >  There is currently a bug because "please select one of the images below based on your platform" isn't being printed to stderr, and testing for this output would have caught it quickly.

Scratch this part. I see it is showing up in the output, but testing for it with our current setup might be weird. But I do think it would be good to test for the output

src/pkg/packager/creator/normal.go
schristoff: After typing it, I went "this probably gets outputted somewhere not here", and to be transparent I just didn't look. 🙃

schristoff (dismissed):
Other than the CVE checker being mad, lgtm

src/internal/packager/images/pull.go
```diff
@@ -308,6 +311,27 @@ func (i *ImageConfig) PullImage(src string, spinner *message.Spinner) (img v1.Im
 			return nil, false, fmt.Errorf("failed to load image from docker daemon: %w", err)
 		}
 	} else {
+		refInfo, err := transform.ParseImageRef(src)
```
schristoff, line 314: Would really love comments on this logic here because it's a lot, but isn't necessary, and is looking a bit cleaner.

src/test/e2e/00_use_cli_test.go
```diff
@@ -216,21 +217,31 @@ func TestUseCLI(t *testing.T) {
...
+		originalPath := "src/test/packages/00-yq-checks"
+
+		originalFile := filepath.Join(originalPath, "file1.yaml")
+		originalOtherFile := filepath.Join(originalPath, "file2.yaml")
+
+		file := filepath.Join(tmpdir, "file1.yaml")
+		otherFile := filepath.Join(tmpdir, "file2.yaml")
+
+		copy.Copy(originalFile, file)
+		copy.Copy(originalOtherFile, otherFile)
```
lucasrod16, line 230: We should handle the error returned from `copy.Copy()`

error

lucasrod16 (changes requested):

src/test/e2e/00_use_cli_test.go
lucasrod16: Also need to run `go mod tidy`

lucasrod16 (commented):
I think we will need to wait for the [`pkg/helpers` PR](https://github.com/defenseunicorns/pkg/pull/63) to be merged and the CVE workflow to be fixed before merging this in. Otherwise, all of the changes in this PR lgtm

lucasrod16 (approved):

PR #2431, opened 2024-04-10
chore: cleanup root level files and add SPDX check for Go files

Noxsios:
## Description

...

## Related Issue

Fixes #

Relates to #

## Checklist before merging


.pre-commit-config.yaml
Noxsios, line 1: update `.github/CONTRIBUTING.md` w/ how to install `revive` through `go install`

.github/workflows/test-bigbang.yml
```diff
@@ -101,8 +108,6 @@ jobs:
         env:
           IRON_BANK_ROBOT_USERNAME: ${{ secrets.IRON_BANK_ROBOT_USERNAME }}
         run: |
-          sudo mkdir /mnt/zarf-tmp
-          sudo chown -R runner:runner /mnt/zarf-tmp
           CI=true go test ./src/extensions/bigbang/test -failfast -v -timeout 30m
```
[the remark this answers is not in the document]
AustinAbro321: the init package gets deleted during `make delete-package` now so that step has to be run before cleanup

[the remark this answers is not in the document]
Noxsios: hmm, ok I didn't realize the side effect of that change to `delete-packages`, if you would be so kind, please revert that change to the `Makefile` and this change to CI.

jeff-mccoy (approved):

Makefile
```diff
@@ -219,11 +218,12 @@ test-docs-and-schema:
...
 
 cve-report: ## Create a CVE report for the current project (must `brew install grype` first)
 	@test -d ./build || mkdir ./build
-	go run main.go tools sbom scan . -o json --exclude './docs-website' --exclude './examples' | grype -o template -t hack/.templates/grype.tmpl > build/zarf-known-cves.csv
+	go run main.go tools sbom scan . -o json --exclude './site' --exclude './examples' | grype -o template -t hack/grype.tmpl > build/zarf-known-cves.csv
 
 lint-go: ## Run revive to lint the go code (must `brew install revive` first)
-	revive -config revive.toml -exclude src/cmd/viper.go -formatter stylish ./src/...
+	revive -config hack/revive.toml -exclude src/cmd/viper.go -formatter stylish ./src/...
+	hack/check-spdx-go.sh src >/dev/null || (echo "SPDX check for go failed, please run 'hack/check-spdx-go.sh src' to see the errors" && exit 1)
```
lucasrod16, line 229: suggest silencing this output when `make lint-go` is ran
```suggestion
	@hack/check-spdx-go.sh src >/dev/null || (echo "SPDX check for go failed, please run 'hack/check-spdx-go.sh src' to see the errors" && exit 1)
```

hack/lint-all-zarf-packages.sh
```diff
@@ -1,4 +1,6 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
+set -euo pipefail
 
 ZARF_BIN=$1
 LINT_SRC_TEST=$2
```
lucasrod16, line 6: unbound-var

When this script is called in `make lint-packages-and-examples`, it fails because we do not provide a value for the `LINT_SRC_TEST` variable

lucasrod16 (changes requested):

PR #2438, opened 2024-04-11
docs: add redirect for examples

lucasrod16:
## Description
 add redirect for examples
 
 https://docs.zarf.dev/examples/

## Checklist before merging

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -2,7 +2,27 @@
   base = "site/"
   command = "npm run build"
   publish = "dist/"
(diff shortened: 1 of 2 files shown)


AustinAbro321 (dismissed):

Noxsios (changes requested):
Use this guy instead, it will handle the dynamic redirects as well. Astro when built as a static site cannot do dynamic route redirects.

https://github.com/defenseunicorns/zarf/blob/0d44b76e5bed7f7d258cb11f197f925151c394de/netlify.toml

Leave the root `/docs/zarf-overview --> /` redirect, but use this guy for the others

Noxsios (approved):

PR #2440, opened 2024-04-12
feat: config to enable resilient registry

Michael-Kruggel:
## Description

- Provided fields for affinity and toleration of the registry pod
- Added s3 irsa configurability for the docker registry helm chart if using a compatible image

## Related Issue

Relates to #375

## Checklist before merging


corang:
https://repo1.dso.mil/dsop/opensource/docker/registry-v2/-/issues/112

zack-is-cool:
thinking about how to automate the release of this registry1 flavored zarf-init package

1. We have a `zarf-config-registry1.toml` under `hack/` dir which will be utilized for building the `-registry1.alpha` flavored zarf init package
2. renovate is watching IB to see if version needs to be updated
	1. Naturally this will lag behind zarf tags
3. renovate picks up changes && creates PR w/ changes to this file (container versions)
4. changes get merged into main on that file
5. pipeline triggers to publish a new release for the `init:$zarf_ver-registry1.alpha` zarf init package

@Noxsios Do we want to put in testing for this package before cutting the release or just ignore it for now? I think most of your guys tests have the [docker registry image version hardcoded](https://github.com/defenseunicorns/zarf/blob/e00cbdd20879ab17ba8e66c80b89882e2f1e3e01/src/test/common.go#L129), we'd just need that to be able to be an input when running the go tests or get it from environment variables or something.

AustinAbro321:
> I think most of your guys tests have the [docker registry image version hardcoded](https://github.com/defenseunicorns/zarf/blob/e00cbdd20879ab17ba8e66c80b89882e2f1e3e01/src/test/common.go#L129), we'd just need that to be able to be an input when running the go tests or get it from environment variables or something.

That function is setting up a docker registry to have a separate registry from the cluster so we can test OCI pulling and pushing of Zarf packages to external registries. The cluster will stand up whichever registry is specified by init package, so we don't have to worry about that.

One way to test the custom init package, would be to add logic to the [init test](https://github.com/defenseunicorns/zarf/blob/main/src/test/e2e/20_zarf_init_test.go#L69) to parametrize the init package that's deployed. In this scenario you'd add an option to make command so you can run something like `make test-e2e-with-cluster init-package=registry_1_package`, and it'd run the end to end test suite with your init package. If we go with this approach you could run the entire e2e suite on the registry1 init package before you publish. We could run it on main as well but I'd be concerned with increasing the registry1 footprint in our tests as it tends to introduce flakes.

packages/zarf-registry/zarf.yaml
```diff
@@ -58,6 +58,40 @@ variables:
...
+  - name: REGISTRY_CREATE_SERVICE_ACCOUNT
+    description: Toggle the ability to use an IRSA for the registry
+    default: "false"
+
+  - name: REGISTRY_SERVICE_ACCOUNT_NAME
+    description: The name of the service account to use. If not set and create is true, a name is generated using fullname template
+    default: ""
+
+  - name: REGISTRY_SERVICE_ACCOUNT_ANNOTATIONS
+    description: The ARN of the IAM Role to assume
```
AustinAbro321, line 70: Recommend taking out explicit mention of AWS things here since it should be general

AustinAbro321 (changes requested):

packages/zarf-registry/zarf.yaml
corang: New descriptions are more generic/appropriate

Makefile
```diff
@@ -131,19 +131,21 @@ init-package: ## Create the zarf init package (must `brew install coreutils` on
...
 publish-init-package:
 	$(ZARF_BIN) package publish build/zarf-init-$(ARCH)-$(CLI_VERSION).tar.zst oci://$(REPOSITORY_URL)
 	$(ZARF_BIN) package publish . oci://$(REPOSITORY_URL)
 
+# INTERNAL: used to build an iron bank version of the init package and output to a local directory for testing
+build-registry1-init-package:
+	@test -s $(ZARF_BIN) || $(MAKE) build-cli
+	ZARF_CONFIG=$(ZARF_CONFIG) $(ZARF_BIN) package create --output build --confirm .
+
+# INTERNAL: used to build a release version of the ib init package with a specific agent image
```
lucasrod16, line 144: The changes to the registry chart and image in this PR are valuable as it relates to #375.

On the other hand, I do not think the Zarf team should be publishing and maintaining a registry1 flavored init package because that is a specific use case that can be solved with users building custom init packages:

https://docs.zarf.dev/tutorials/8-custom-init-packages/

We currently only publish and maintain the core Zarf init package, and I do not see a good reason for the Zarf team to maintain any custom init packages for registry1. I would like to see any registry1 specific additions in this PR removed.

lucasrod16 (changes requested):

Makefile
corang: Will go ahead and remove the registry1 stuff and move to our own repo for publishing custom init package

lucasrod16 (changes requested):
We should add a note in the `init` docs that explains this feature and a simple example of how to configure the registry to achieve HA.

https://github.com/defenseunicorns/zarf/blob/main/site/src/content/docs/ref/init-package.mdx

Between `What Makes the Init Package Special` and `The zarf init Lifecycle` would probably be a good place to add a section. Doesn't have to be super detailed or in depth.

Please reference our style guide to see how we prefer our documentation to be written:

https://github.com/defenseunicorns/zarf/blob/main/site/src/content/docs/contribute/style-guide.mdx

site/src/content/docs/ref/init-package.mdx
```diff
@@ -104,6 +104,35 @@ The `registry:2` image and the Zarf Agent image can be configured with a custom
 
 :::
 
+## Making the Registry Highly-Availabile
+
+By default the registry included in the init package creates a ReadWriteOnce PVC and only schedules onto one node at a time. For smaller and simpler deployments this is usually enough. However, for larger deployments or deployments where nodes will be restarted or updated somewhat frequently, you may want to make the registry highly-available.
```
lucasrod16, line 109:
```suggestion
By default, the registry included in the init package creates a ReadWriteOnce PVC and is only scheduled to run on one node at a time. This setup is usually enough for smaller and simpler deployments. However, for larger deployments or those where nodes are frequently restarted or updated, you may want to make the registry highly-available.
```

site/src/content/docs/ref/init-package.mdx
```diff
@@ -104,6 +104,35 @@ The `registry:2` image and the Zarf Agent image can be configured with a custom
...
+
+This can have some prerequisites, such as having a storage class that supports ReadWriteMany, or being in an environment where you can configure the registry to use an s3 compatible backend. Beyond the backing storage, though, you will need to pass some custom configuration to the registry so that it spreads out across all nodes and has the correct amount of replicas. Below is an example configuration using a ReadWriteMany storage class:
```
lucasrod16, line 111:
```suggestion
This approach requires certain prerequisites, such as a storage class that supports ReadWriteMany, or being in an environment that allows you to configure the registry to use an S3-compatible backend. Additionally, you must provide custom configuration to the registry to ensure it is distributed across all nodes and has the appropriate number of replicas. Below is an example configuration using a ReadWriteMany storage class:
```

site/src/content/docs/ref/init-package.mdx
```diff
@@ -104,6 +104,35 @@ The `registry:2` image and the Zarf Agent image can be configured with a custom
...
+                labelSelector:
+                  matchExpressions:
+                    - key: app
+                      operator: In
+                      values:
+                        - docker-registry
+                topologyKey: kubernetes.io/hostname
+```
+
+Of note is the `REGISTRY_AFFINITY_CUSTOM` variable which overrides the default pod anti-affinity and `REGISTRY_HPA_AUTO_SIZE` which will automatically set the min and max replicas for the registry based on the number of nodes in the cluster. If you would prefer to set the min and max replicas manually you can set `REGISTRY_HPA_MIN` and `REGISTRY_HPA_MAX` to the desired values instead.
```
lucasrod16, line 134:
```suggestion
Notably, the `REGISTRY_AFFINITY_CUSTOM` variable overrides the default pod anti-affinity, and `REGISTRY_HPA_AUTO_SIZE` automatically adjusts the minimum and maximum replicas for the registry based on the number of nodes in the cluster. If you prefer to manually set the minimum and maximum replicas, you can use `REGISTRY_HPA_MIN` and `REGISTRY_HPA_MAX` to specify the desired values.
```

lucasrod16 (changes requested):
Just a few final, minor doc/wording suggestions from me

site/src/content/docs/ref/init-package.mdx
```diff
@@ -104,6 +104,35 @@ The `registry:2` image and the Zarf Agent image can be configured with a custom
 
 :::
 
+## Making the Registry Highly-Availabile
```
lucasrod16, line 107:
```suggestion
## Making the Registry Highly-Available
```

lucasrod16 (changes requested):

lucasrod16 (dismissed):

AustinAbro321 (dismissed):

packages/zarf-registry/zarf.yaml
```diff
@@ -58,6 +58,40 @@ variables:
...
+    description: Toggle the creation of a new service account for the registry
+    default: "false"
+
+  - name: REGISTRY_SERVICE_ACCOUNT_NAME
+    description: The name of the service account to use. If not set and create is true, a name is generated using fullname template
+    default: ""
+
+  - name: REGISTRY_SERVICE_ACCOUNT_ANNOTATIONS
+    description: Map of annotations to add to the created service account
+    default: ""
```
Noxsios, line 71: should this have `autoIndent: true`?

Noxsios (changes requested):

packages/zarf-registry/zarf.yaml
Michael-Kruggel: Yes

site/src/content/docs/ref/init-package.mdx
```diff
@@ -104,6 +104,35 @@ The `registry:2` image and the Zarf Agent image can be configured with a custom
...
 
+## Making the Registry Highly-Available
+
+By default, the registry included in the init package creates a ReadWriteOnce PVC and is only scheduled to run on one node at a time. This setup is usually enough for smaller and simpler deployments. However, for larger deployments or those where nodes are frequently restarted or updated, you may want to make the registry highly-available.
+
+This approach requires certain prerequisites, such as a storage class that supports ReadWriteMany, or being in an environment that allows you to configure the registry to use an S3-compatible backend. Additionally, you must provide custom configuration to the registry to ensure it is distributed across all nodes and has the appropriate number of replicas. Below is an example configuration using a ReadWriteMany storage class:
+
+```yaml
+package:
+  deploy:
```
Noxsios, line 115:
```suggestion
This approach requires certain prerequisites, such as a storage class that supports ReadWriteMany, or being in an environment that allows you to configure the registry to use an S3-compatible backend. Additionally, you must provide custom configuration to the registry to ensure it is distributed across all nodes and has the appropriate number of replicas. Below is an example [configuration file](/ref/config-files/) using a ReadWriteMany storage class:

```yaml
# zarf-config.yaml
package:
```

Noxsios (changes requested):
small docs tweak, then lgtm

site/src/content/docs/ref/init-package.mdx
```diff
@@ -104,6 +104,35 @@ The `registry:2` image and the Zarf Agent image can be configured with a custom
...
+## Making the Registry Highly-Available
+
+By default, the registry included in the init package creates a ReadWriteOnce PVC and is only scheduled to run on one node at a time. This setup is usually enough for smaller and simpler deployments. However, for larger deployments or those where nodes are frequently restarted or updated, you may want to make the registry highly-available.
+
+This approach requires certain prerequisites, such as a storage class that supports ReadWriteMany, or being in an environment that allows you to configure the registry to use an S3-compatible backend. Additionally, you must provide custom configuration to the registry to ensure it is distributed across all nodes and has the appropriate number of replicas. Below is an example [configuration file](/ref/config-files/) using a ReadWriteMany storage class:
+
+```yaml
+# zarf-config.yaml
+package:
+    set:
```
lucasrod16, line 116: I think we will need `deploy` in here

https://github.com/defenseunicorns/zarf/blob/b4201718aa96f34f37dac7afa5ccfe59affedbd0/examples/config-file/zarf-config.yaml#L10-L13

lucasrod16 (changes requested):

site/src/content/docs/ref/init-package.mdx
Noxsios: thats mb, looks like my `suggest` diff didnt sync properly

lucasrod16 (approved):

Noxsios (approved):

PR #2443, opened 2024-04-17
ci: minor conviences and checks

AustinAbro321:
## Description

- rennovate check grype
- update references to docs-website
- lint on pre-commit
- nicer link for openssf viewing

## Checklist before merging

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -85,7 +85,7 @@
     },
     {
       "fileMatch": [
-        ".github/actions/install-tools/action.yaml"
+        ".github/**"
       ],
       "matchStrings": [
         "# renovate: datasource=github-tags depName=anchore/grype versioning=semver\n\\s*VERSION: (?<currentValue>v[\\d.]+)"
--- a/renovate.json
+++ b/renovate.json
@@ -82,6 +82,16 @@
         "https:\\/\\/github.com\\/(?<depName>[\\w\\/\\-\\.\\+\\%]+?)\\/releases\\/download\\/(?<currentValue>[\\w\\/\\-\\.\\+\\%]+?)\\/"
       ],
       "datasourceTemplate": "github-releases"
+    },
+    {
+      "fileMatch": [
+        ".github/actions/install-tools/action.yaml"
+      ],
+      "matchStrings": [
+        "# renovate: datasource=github-tags depName=anchore/grype versioning=semver\n\\s*VERSION: (?<currentValue>v[\\d.]+)"
+      ],
+      "datasourceTemplate": "github-tags",
+      "depNameTemplate": "anchore/grype"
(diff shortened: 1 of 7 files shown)


renovate.json
```diff
@@ -82,6 +82,16 @@
...
+    {
+      "fileMatch": [
+        ".github/actions/install-tools/action.yaml"
+      ],
+      "matchStrings": [
+        "# renovate: datasource=github-tags depName=anchore/grype versioning=semver\n\\s*VERSION: (?<currentValue>v[\\d.]+)"
+      ],
+      "datasourceTemplate": "github-tags",
+      "depNameTemplate": "anchore/grype"
     }
```
lucasrod16, line 95: Should we should make this `regexManager` more flexible to match against all `.github/**` files and use a pattern that would work for other dependencies other than grype? For example, we may want to pin our k3d action and configure renovate to watch for updates as well. Compared all of the other `regexManager`s we have, this one is much more restrictive as it will only work for a single dependency in a single file.

lucasrod16 (commented):
All of the changes look good to me. Curious to get your thoughts on the renovate config

lucasrod16 (approved):

renovate.json
AustinAbro321: Yeah the reason I did it like that is because  it's unlikely we'll have this regex anywhere else so I thought it was cleaner to be specific.  But I suppose if we moved the file or renamed it to .yml it would break. I'm going to take your suggestion because I don't like touching renovate and it'll make it less likely we'll have to edit it in the future lol

lucasrod16: I used `.github/**` as an example, but I don't think renovate actually support glob patterns. I believe we will have to use a regex pattern.

Something like `"(^|/).github/.*\\.(yaml|yml)$"` should work

https://docs.renovatebot.com/modules/manager/#file-matching

lucasrod16: Actually, the built-in github actions manager already matches against the right files for us:

https://docs.renovatebot.com/modules/manager/github-actions/#file-matching

default config:
```json
{
  "fileMatch": [
    "(^|/)(workflow-templates|\\.(?:github|gitea|forgejo)/(?:workflows|actions))/.+\\.ya?ml$",
    "(^|/)action\\.ya?ml$"
  ]
}
```

so we should be able to simplify even more:

```json
{
  "matchStrings": [
    "# renovate: datasource=github-tags depName={{dependency}} versioning=semver\n\\s*VERSION: (?<currentValue>v[\\d.]+)"
  ],
  "datasourceTemplate": "github-tags",
  "depNameTemplate": "{{dependency}}"
}
```

AustinAbro321:
Closing to combine with #2431

PR #2444, opened 2024-04-18
fix: differential package create with non local sources

AustinAbro321:
## Description

Currently there is a bug wherein differential will only work if the source is a local file

## Checklist before merging


lucasrod16 (dismissed):
lgtm

lucasrod16 (approved):

PR #2448, opened 2024-04-22
ci: compare cves to main

AustinAbro321:
## Description

This would add a compare CVEs to main workflow to replace the "analyze cves" which currently runs on PRs. This new workflow would only fail on PRs is new CVE's were added rather than always failing if CVE's exist in the PR. 

Assuming we want to move forward with this we would need to add some extra signal for ourselves to notify us when we do have CVEs on PRs. Maybe a status icon on the repo? Maybe emails? Maybe auto opened issues ? Github security also opens up alerts though those don't take into account the .grype.yaml so there is more noise


AustinAbro321:
Here is a run of this, added `push:` temporarily on the workflow so we could see it run https://github.com/defenseunicorns/zarf/actions/runs/8806198963/job/24170475617?pr=2448#step:6:341

lucasrod16 (approved):

PR #2456, opened 2024-04-23
fix!: do not uninstall helm chart after failed install or upgrade

### LINKED ISSUE #2455: Zarf's cleanup-on-failure logic can cause unintended deletion of applications during failed upgrades
When Zarf fails to install a component it attempts to purge the helm install as shown below:
```
  :package: KEYCLOAK COMPONENT

  ✔  Pushed 2 images to the zarf registry
  ⠹  preparing upgrade for keycloak
 WARNING  Retrying (1/3) in 5s: unable to complete the helm chart install/upgrade: another operation
          (install/upgrade/rollback) is in progress
  ⠏  preparing upgrade for keycloak (6s)
 WARNING  Retrying (2/3) in 10s: unable to complete the helm chart install/upgrade: another
          operation (install/upgrade/rollback) is in progress
  ⠙  preparing upgrade for keycloak (16s)
 WARNING  Retrying (3/3) in 20s: unable to complete the helm chart install/upgrade: another
          operation (install/upgrade/rollback) is in progress
  ⠧  purge requested for keycloak
     ERROR:  Failed to deploy bundle: unable to deploy component "keycloak": unable to install helm chart(s):
             unable to install chart after 3 attempts
```

On initial install this makes some amount of sense, and there is logic that attempts to prevent removal on a failed upgrade. However, in some edge cases (as shown above), Zarf may mistake an upgrade as an initial install and delete something that should instead be rolled back.

These rough steps caused an existing (system critical) deployment to be removed from the cluster:
1. Successfully deploy a Zarf package including Keycloak
1. Attempt an upgrade some time later
1. Realize the upgrade is doomed
1. ctrl-c out of the deploy (a timeout would have resulted in a rollback, but would have taken ~45mins)
1. Make a change and re-redeploy
1. Laugh/cry a little as Keycloak is removed from the cluster

The deletion is almost certainly explained by the aborted deployment leaving the helm release in an incomplete state, but the existence of the behavior in the first place is very concerning for long-lived production deployments.

- **Given** a Zarf package
- **When** when there is an error during deployment
- **Then** Zarf attempts a rollback (if available), and does nothing if not 

In this particular case, a rollback would not have been possible either due to the invalid helm state -- but in 9/10 situations, I would rather have a broken deployment than no deployment..

Potentially, Zarf could catch the abort signal from ctrl-c and attempt a rollback immediately on exit rather than leaving the helm release in an incomplete state. (half baked idea.. but maybe there's something there..)

lucasrod16:
## Description
Do not uninstall helm chart after failed install or upgrade

## Related Issue

Fixes #2455

## Checklist before merging


AustinAbro321 (dismissed):

src/internal/packager/helm/chart.go
```diff
@@ -115,12 +115,6 @@ func (h *Helm) InstallOrUpgradeChart() (types.ConnectStrings, string, error) {
 			return nil, "", fmt.Errorf("unable to upgrade chart after %d attempts", h.retries)
 		}
 
-		spinner.Updatef("Performing chart uninstall")
```
Racer159, line 118: May be useful to warn the user that we couldn't uninstall and that if they want things gone they should do a `zarf package remove`

lucasrod16: included a message to recommend using `zarf package remove` in the returned error

site/src/content/docs/ref/deploy.mdx
```diff
@@ -120,3 +120,43 @@ $ zarf connect [service name]
...
+Use the `--retries` flag with `zarf init` and `zarf package deploy` to change the number of retry attempts.
+
+### Rollback Process
+
+If attempts to upgrade a chart fail, Zarf tries to roll the chart back to its last successful release.
+
+Key features of Zarf's `helm rollback` include:
+
+   - Deleting newly created resources if a rollback fails (`helm rollback --cleanup-on-fail`)
+   - Forcing updates by deleting and recreating resources as necessary (`helm rollback --force`)
```
AustinAbro321, line 162:
```suggestion
If attempts to upgrade a chart fail, Zarf tries to roll the chart back to its last successful release. During this rollback process:

  - Any resources created during the failed upgrade attempt are deleted (`helm rollback --cleanup-on-fail`)
  - Resource updates are forced through delete and recreate if needed (`helm rollback --force`)
```

This reads a bit to me like Zarf has an included helm rollback feature, rather than it being an internal mechanism. What do you think of this suggestion?

lucasrod16: I can see how that can be confusing. Sounds good to me

AustinAbro321 (approved):

Noxsios (approved):

PR #2460, opened 2024-04-25
refactor: cleaner image pulls

Noxsios:
## Description

Image pulls needed some love.

## Checklist before merging


src/internal/packager/images/pull.go
```diff
@@ -27,329 +32,332 @@ import (
...
 	}
 
-	for refInfo, img := range refInfoToImage {
-		imgDigest, err := img.Digest()
-		if err != nil {
-			return nil, fmt.Errorf("unable to get digest for image %s: %w", refInfo.Reference, err)
-		}
-		referenceToDigest[refInfo.Reference] = imgDigest.String()
-	}
+	clear(shas)
```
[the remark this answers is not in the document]
Noxsios: was wondering if there were any memory implications of clearing the shas map after is is no longer needed, still need to verify

src/internal/packager/images/pull.go
```diff
@@ -27,329 +32,332 @@ import (
...
 
 	logs.Warn.SetOutput(&message.DebugWriter{})
 	logs.Progress.SetOutput(&message.DebugWriter{})
 
-	metadataImageConcurrency := helpers.NewConcurrencyTools[ImgInfo, error](len(i.ImageList))
+	eg, _ := errgroup.WithContext(ctx)
+	eg.SetLimit(10)
+
+	var fetchedLock, shaLock sync.Mutex
+	totalBytes := int64(0)
```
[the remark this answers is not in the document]
Noxsios: good catch, will do

src/internal/packager/images/pull.go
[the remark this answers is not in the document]
Noxsios: trying to get better at understanding go garbage collection and more efficient memory management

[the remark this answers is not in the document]
AustinAbro321: Agreed, it would be easy to read as well

src/internal/packager/images/pull.go
```diff
@@ -27,329 +32,329 @@ import (
...
 	logs.Warn.SetOutput(&message.DebugWriter{})
 	logs.Progress.SetOutput(&message.DebugWriter{})
 
-	metadataImageConcurrency := helpers.NewConcurrencyTools[ImgInfo, error](len(i.ImageList))
+	eg, _ := errgroup.WithContext(ctx)
+	eg.SetLimit(10)
 
-	defer metadataImageConcurrency.Cancel()
+	var fetchedLock, shaLock sync.Mutex
+	shas := make(map[string]bool)
```
phillebaba, line 81: NIT: Keep declaration syntax the same.

```suggestion
	shas := map[string]bool{}
```

src/internal/packager/images/pull.go
```diff
@@ -27,329 +32,329 @@ import (
...
+	opts := append(CommonOpts(cfg.Arch), crane.WithContext(ctx))
 
-	spinner.Updatef("Fetching image metadata (0 of %d)", len(i.ImageList))
+	var fetched = map[transform.Image]v1.Image{}
```
phillebaba, line 84:
```suggestion
	fetched := map[transform.Image]v1.Image{}
```

src/internal/packager/images/pull.go
```diff
@@ -27,329 +32,329 @@ import (
...
-		imgInfoList       []ImgInfo
-	)
-
-	type digestInfo struct {
-		refInfo transform.Image
-		digest  string
-	}
-
+// Pull pulls all of the images from the given config.
+func Pull(ctx context.Context, cancel context.CancelFunc, cfg PullConfig) (map[transform.Image]v1.Image, error) {
```
phillebaba, line 49: It would be more idiomatic to only pass the context as an argument and then define a new sub context with cancel in the beginning of this function.

src/internal/packager/images/pull.go
```diff
@@ -27,329 +32,329 @@ import (
...
+				}
+			} else {
+				reference, err := name.ParseReference(ref)
+				if err != nil {
+					return fmt.Errorf("failed to parse reference: %w", err)
+				}
+				desc, err = crane.Get(ref, opts...)
+				if err != nil {
+					if strings.Contains(err.Error(), "unexpected status code 429 Too Many Requests") {
+						cancel()
```
phillebaba, line 118: Why is cancel being called here? Any error returned from an error group will cancel the context used to create it.

src/internal/packager/images/pull.go
```diff
@@ -27,329 +32,329 @@ import (
...
-			if _, ok := processedLayers[layerDigest.Hex]; !ok {
-				size, err := layer.Size()
+			for _, layer := range layers {
+				digest, err := layer.Digest()
 				if err != nil {
-					return nil, fmt.Errorf("unable to get size of layer: %w", err)
+					return fmt.Errorf("unable to get digest for image layer: %w", err)
+				}
+
+				shaLock.Lock()
```
phillebaba, line 201: Lock is not released on defer here. If an error is returned due to the layer size the lock will never be released. Any other pull waiting for the critical section will block forever.

src/internal/packager/images/pull.go
```diff
@@ -27,329 +32,329 @@ import (
...
+	}
+
+	spinner := message.NewProgressSpinner("Fetching info for %d images. %s", imageCount, longer)
 	defer spinner.Stop()
 
 	logs.Warn.SetOutput(&message.DebugWriter{})
 	logs.Progress.SetOutput(&message.DebugWriter{})
 
-	metadataImageConcurrency := helpers.NewConcurrencyTools[ImgInfo, error](len(i.ImageList))
+	eg, _ := errgroup.WithContext(ctx)
```
phillebaba, line 77: The context returned here should be used.

> The derived Context is canceled the first time a function passed to Go returns a non-nil error or the first time Wait returns, whichever occurs first.

```suggestion
	eg, gCtx := errgroup.WithContext(ctx)
```

You want to specifically use this context inside of the parallel code blocks to make sure that they cancel on first error.

src/pkg/packager/creator/normal.go
```diff
@@ -174,37 +178,32 @@ func (pc *PackageCreator) Assemble(dst *layout.PackagePaths, components []types.
...
-			return fmt.Errorf("unable to pull images: %w", err)
+		pulled, err := images.Pull(ctx, cancel, pullCfg)
+		if err != nil {
+			return err
 		}
 
-		for _, imgInfo := range pulled {
-			if err := dst.Images.AddV1Image(imgInfo.Img); err != nil {
+		for info, img := range pulled {
+			info, img := info, img
```
phillebaba, line 197: Re-declaring loop variables (in Go 1.21) is only required if you intend to keep reference to them past the loop iteration. It only becomes an issue during concurrent processing, like running go routines in a for loop.

FYI this has been fixed in Go 1.22.

src/internal/packager/images/pull.go
```diff
@@ -27,329 +32,329 @@ import (
...
+					message.WarnErr(err, "failed to clean up in-progress layers, please remove them manually")
+				}
+				return err
+			}
 
-	hasImageLayers, err = utils.HasImageLayers(img)
-	if err != nil {
-		return nil, false, fmt.Errorf("failed to check image layer mediatype: %w", err)
-	}
+			mu.Lock()
```
phillebaba, line 344: Ditto with the deadlock issue.

phillebaba (commented):
It's a lot of refactoring and it looks like it is needed.

I have some thoughts on the saving logic which as I have heard fails on concurrent saves due to layers overwriting each other. This should be fixed in Crane in a perfect world. Until that is done I have an alternative idea. Create a map of all the layer digests with a mutex for each. Before writing wait for all the layers locks for the current image. Write  locks for the new image that will be written. That way you will avoid conflicts caused by two images writing the same layer at the same time.

src/internal/packager/images/pull.go
Noxsios: I should have tested it, but I was under the impression context cancellation was a one way data flow, where parent cancellations would kill dependent contexts, but not the other way around.

src/internal/packager/images/pull.go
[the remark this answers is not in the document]
Noxsios: Yeah, was playing around w/ some stuff trying to visualize in my head _if_ this would net any benefit

src/internal/packager/images/pull.go
```diff
@@ -27,329 +32,329 @@ import (
...
-		// If crane is unable to pull the image, try to load it from the local docker daemon.
-		message.Notef("Falling back to local 'docker' images, failed to find the manifest on a remote: %s", err.Error())
+// CleanupInProgressLayers removes incomplete layers from the cache.
+func CleanupInProgressLayers(ctx context.Context, img v1.Image) error {
+	layers, err := img.Layers()
+	if err != nil {
+		return err
+	}
+	eg, _ := errgroup.WithContext(ctx)
+	eg.SetLimit(-1)
```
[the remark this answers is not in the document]
Noxsios: agreed, I went w/ this to be explicit about there being no limit, but default behavior should be generally understood and consistent when it comes to errgroups

src/pkg/packager/creator/normal.go
Noxsios: linting was being mean to me, so I re-declared, I'll check if I can remove / validate my linting rules

src/internal/packager/images/pull.go
Noxsios: Refactored to not need `cancel` at all as context usage is is now more properly scoped

src/internal/packager/images/pull.go
Noxsios: Is there a better way to cancel within an `eg.Go` aside from `return err`?

Noxsios: I was unsure how this variable is supposed to be used as there is no CancelFunc associated w/ it, so the only way I know of to cancel it would be returning an err inside one of its callbacks

src/internal/packager/images/push.go
```diff
@@ -46,85 +41,95 @@ func (i *ImageConfig) PushToZarfRegistry() error {
...
 
-	for refInfo, img := range refInfoToImage {
-		refTruncated := helpers.Truncate(refInfo.Reference, 55, true)
-		progressBar.UpdateTitle(fmt.Sprintf("Pushing %s", refTruncated))
+		pushed := []transform.Image{}
+		defer func() {
+			for _, refInfo := range pushed {
+				delete(toPush, refInfo)
+			}
+		}()
```
lucasrod16, line 85: Seems like we don't need to manually cleanup resources here. The garbage collector will clean this up for us

lucasrod16 (commented):
Looking pretty good overall. Left a few minor comments

src/internal/packager/images/push.go
AustinAbro321: If I understand the code correctly, this is necessary because the code here is run in a helpers.retry loop. If there's a failed attempt we only want to retry on images we haven't pushed yet

src/internal/packager/images/common.go
```diff
@@ -5,23 +5,119 @@
...
+	RegistryOverrides map[string]string
+
+	CacheDirectory string
+}
+
+// PushConfig is the configuration for pushing images.
+type PushConfig struct {
+	SourceDirectory string
 
 	ImageList []transform.Image
```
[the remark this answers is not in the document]
Noxsios: yeah, i meant to go back and make them consistent, mb

src/internal/packager/images/push.go
Noxsios: correct, no matter what we want to retry incrementally, as failure can happen partway through, there is no need to re-attempt images that have already succeeded

lucasrod16: That makes sense. Thanks for the explanation @AustinAbro321 @Noxsios

src/internal/packager/images/pull.go
```diff
@@ -27,329 +31,314 @@ import (
...
+		})
+	}
+	return nil
+}
+
+// SaveSequential saves images sequentially.
+func SaveSequential(ctx context.Context, cl clayout.Path, m map[transform.Image]v1.Image) (map[transform.Image]v1.Image, error) {
+	saved := map[transform.Image]v1.Image{}
+	for info, img := range m {
+		info, img := info, img
```
lucasrod16, line 289: I don't think we need to shadow the `info` and `img` variables in this `for` loop, because the logic within the `for` loop operates sequentially within a single goroutine and the variables are reassigned on each iteration of the loop, rather than being shared across multiple goroutines.

```suggestion
```

src/internal/packager/images/pull.go
```diff
@@ -27,329 +31,314 @@ import (
...
-			return nil, false, fmt.Errorf("%w: %s", lang.ErrUnsupportedImageType, imageOptions)
-		}
 
-		// Manifest was found, so use crane to pull the image.
-		if img, err = crane.Pull(src, config.GetCraneOptions(i.Insecure, i.Architectures...)...); err != nil {
-			return nil, false, fmt.Errorf("failed to pull image: %w", err)
-		}
-		cacheImage = true
-	}
+			if err := cl.WriteImage(img); err != nil {
```
lucasrod16, line 321: Are we able to use `AppendImage` here like we do in `SaveSequential`? As opposed to using `WriteImage` + `AppendDescriptor`

src/internal/packager/images/pull.go
```diff
@@ -27,329 +31,314 @@ import (
...
+			}
+			if info.Size() != size {
+				if err := os.Remove(location); err != nil {
+					return fmt.Errorf("failed to remove incomplete layer %s: %w", digest.Hex, err)
+				}
+			}
+			return nil
+		})
+	}
+	return nil
```
lucasrod16, line 282: We will need to use `eg.Wait()` in this function to ensure that all goroutines complete before returning and to propagate any errors encountered by the goroutines

```suggestion
	return eg.Wait()
```

lucasrod16 (changes requested):
After another review, I realized I missed a few things the first go round

src/internal/packager/images/pull.go
Noxsios: it was being mean to me when run concurrently, so I swapped to using a mutex on the `AppendDescriptor` because I believe that function is not go-routine safe

lucasrod16 (dismissed):

src/internal/packager/images/pull.go
```diff
@@ -27,329 +31,313 @@ import (
...
+			digest, err := layer.Digest()
+			if err != nil {
+				return err
+			}
+			size, err := layer.Size()
+			if err != nil {
+				return err
+			}
+			cacheDir := filepath.Join(config.GetAbsCachePath(), layout.ImagesDir)
+			location := filepath.Join(cacheDir, digest.Hex)
```
AustinAbro321, line 269: The file path ends up looking like `.zarf-cache/images/sha256:779edd1984138d80000e2e7ca5deb8ee927712e1edc1426ec165bd156ff08c96` so we need the `sha256:` on top of the digest

AustinAbro321 (changes requested):

src/internal/packager/images/pull.go
Noxsios: whoops, I used `.Hex` instead of `.String`, good catch

AustinAbro321 (approved):

PR #2468, opened 2024-04-30
docs: injector and init package reference material

Noxsios:
## Description

- Finally start documenting the `injector`
- Update `init` package docs
- Consolidate `init` package reference material into single page, bring it out of FAQ

## Checklist before merging


Noxsios:
Commenting the preview URL because netlify overwrites its comment when it cancels the build (due to no changes)

<https://deploy-preview-2468--zarf-docs.netlify.app/>

site/src/content/docs/ref/init-package.mdx
```diff
@@ -5,110 +5,117 @@ sidebar:
...
 
-:::note
+Shoving random data into a cluster is generally a bad idea, and an antipattern overall to containerization. However in the case of Zarf, and air-gapped environments, certain liberties must be taken.
 
-The Zarf Registry is initially injected as a series of config maps that bootstraps itself into the cluster and then binds to a NodePort to allow the kubelet to pull images and setup the final registry deployment.  Doing this keeps Zarf cluster agnostic however does require that the kubelet be able to reach out to a cluster NodePort service which may require changes to firewall configurations like allowing UDP traffic between nodes if using something like VXLAN tunneling.
+While there is no distro-agnostic to inject images into a cluster, every cluster has support for `configmaps`. However, the size of a `configmap` is limited to 1MB (technically only limited by whatever is configured in `etcd`, the default is 1MB), and the `registry:2` image is around 10MB (as of this writing). So we split the `registry:2` into chunks and inject them into the cluster as `configmaps`.
```
lucasrod16, line 50: Seems like this should read:

> While there is no distro-agnostic <way|method|pattern|etc> to inject images into a cluster...

lucasrod16, line 50:
```suggestion
While there is no distro-agnostic to inject images into a cluster, every cluster has support for `configmaps`. However, the size of a `configmap` is limited to 1MB (technically only limited by whatever is configured in `etcd`, the default is 1MB), and the `registry:2` image is around 10MB (as of this writing). So we split the `registry:2` image into chunks and inject them into the cluster as `configmaps`.
```

site/src/content/docs/ref/init-package.mdx
```diff
@@ -5,110 +5,117 @@ sidebar:
...
 
-| Components   | Description                                                                                                                                                       |
-| ------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| k3s          | REQUIRES ROOT (not sudo). Installs a lightweight Kubernetes Cluster on the local host&mdash;[K3s](https://k3s.io/)&mdash;and configures it to start up on boot.   |
-| logging      | Adds a log monitoring stack&mdash;[promtail/loki/grafana (aka PLG)](https://github.com/grafana/loki)&mdash;into the cluster.                                      |
-| git-server   | Adds a [GitOps](https://about.gitlab.com/topics/gitops/)-compatible source control service&mdash;[Gitea](https://gitea.io/en-us/)&mdash;into the cluster. |
+> Doing this keeps Zarf cluster agnostic however does require that the kubelet be able to reach out to a cluster NodePort service which may require changes to firewall configurations like allowing UDP traffic between nodes if using something like VXLAN tunneling.
```
lucasrod16, line 66: I think this would be a good place to use a `note` to draw more attention to it:

```suggestion
:::note

Doing this keeps Zarf cluster agnostic, however does require that the kubelet be able to reach out to a cluster NodePort service, which may require changes to firewall configurations like allowing UDP traffic between nodes if using something like VXLAN tunneling.

:::
```

site/src/content/docs/ref/init-package.mdx
```diff
@@ -134,58 +141,166 @@ package:
...
+
+{/* TODO: document and flesh out how the mutations operate for the agent */}
+
+The `zarf-agent` is a [Kubernetes Mutating Webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) that intercepts requests to create resources and uses the `zarf-state` secret to mutate them to point to their air-gapped equivalents.
+
+The `zarf-agent` is responsible for modifying [Kubernetes PodSpec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#PodSpec) objects [Image](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#Container.Image) fields to point to the Zarf Registry.
+
+This allows the cluster to pull images from the Zarf Registry instead of the internet without having to modify the original image references.
+
+The `zarf-agent` also modifies [Flux GitRepository](https://fluxcd.io/docs/components/source/gitrepositories/) objects to point to the local Git Server.
```
lucasrod16, line 154: Might be good to include that we the agent also mutates ArgoCD `Application` and `Repository` objects

lucasrod16 (commented):
Looking good! Few minor comments

site/src/content/docs/ref/init-package.mdx
Noxsios: I left a TODO to really flesh out how the mutations work, but I will add that info

lucasrod16 (approved):
lgtm

site/src/content/docs/ref/init-package.mdx
```diff
@@ -134,58 +145,166 @@ package:
...
+
+> Support for mutating `Application` and `Repository` objects in ArgoCD is in [`beta`](/roadmap#beta) and can be tested on non-production clusters.
```
AustinAbro321, line 160: IMO given that the argo code has been in main for about 9 months I think it would be fair to say it's no longer in beta, but I'm not sure how we draw that line. Also do we want to suggest that users shouldn't use beta features in prod? I feel that a user of Zarf reading [the roadmap](https://docs.zarf.dev/roadmap/#beta) would assume that beta is safe to use in prod as long as they're ready to migrate to workarounds

site/src/content/docs/ref/init-package.mdx
```diff
@@ -134,58 +145,166 @@ package:
...
+You can learn more about creating a custom init package in the [Creating a Custom 'init' Package Tutorial](/tutorials/8-custom-init-packages).
 
 :::
+
+A minimal `zarf.yaml` for the 'init' package looks something like:
+
+```yaml
+# zarf.yaml
+kind: ZarfInitConfig
+metadata:
```
AustinAbro321, line 286: metadata shouldn't be empty

site/src/content/docs/ref/init-package.mdx
```diff
@@ -134,58 +145,166 @@ package:
...
+```bash
+git clone https://github.com/defenseunicorns/zarf.git
+cd zarf
+mv zarf.yaml zarf.yaml.bak
+cat <<EOF > zarf.yaml
+...
+EOF
+```
+
+You can learn more about creating a custom init package in the [Creating a Custom 'init' Package Tutorial](/tutorials/8-custom-init-packages).
```
AustinAbro321, line 277: IMO this note is a bit confusing. I think the average user will be unsure if this section is about how Zarf interacts with init packages or if it's about how to create your own init packages. I also don't think they'll know what to put in the ellipsis.

AustinAbro321 (changes requested):

site/src/content/docs/ref/init-package.mdx
Noxsios: Time in `beta` status currently is no indication of stability. There are plenty of features that are `beta`, have been around for a long time and are fully broken (like skeleton).

Personally, when I am using `beta` features of an application, I have a general understanding that they are not fully production ready / can have unknown edges that could bring down a deployment.

lucasrod16 (approved):

PR #2476, opened 2024-05-03
fix: update docker media type in registry

AustinAbro321:
## Description

Looks like the media type of the registry:2.8.3 has been changed from `application/vnd.docker.distribution.manifest.v2+json` to  `application/vnd.oci.image.manifest.v1+json`. This breaks minikube, hence the current pipeline fails, as it will do a check to make sure the mediatypes match, which they don't since they're hardcoded 

## Checklist before merging


schristoff-du (approved):
🥹 beautiful, thank you

lucasrod16 (approved):

PR #2494, opened 2024-05-10
fix: adopt namespace metadata

### LINKED ISSUE #2489: ZARF removes namespace labels pertaining to pod security when using --adopt-existing-resources
### Environment
Device and OS:
App version:  0.33.1
Kubernetes distro being used: rke2 1.28.9 
Other:

### Steps to reproduce
1. Create a namespace with the labels required to allow a pod to have privileges when restricted pod security context is enabled per CIS benchmark
2. Deploy an app (in our case mattermost-team-edition) via helm to that namespace
3. Create a zarf package for the app and deploy it with --adopt-existing-resources
4. The pods will not be scheduled due to security context violations and the labels will have been removed from the namespace   

### Expected result
The existing labels that do not interfere with ZARF remain 
### Actual Result
Existing labels pertaining to pod security are removed
### Additional Context

AustinAbro321:
## Description

Changes behavior to keep existing metadata on namespaces rather than steamrolling them

## Related Issue

Fixes #2489 

## Checklist before merging


src/internal/packager/helm/post-render.go
```diff
@@ -167,22 +178,16 @@ func (r *renderer) editHelmResources(resources []releaseutil.Manifest, finalMani
...
 				message.WarnErrf(err, "could not parse namespace %s", rawData.GetName())
 			} else {
 				message.Debugf("Matched helm namespace %s for zarf annotation", namespace.Name)
-				if namespace.Labels == nil {
-					// Ensure label map exists to avoid nil panic
-					namespace.Labels = make(map[string]string)
-				}
-				// Now track this namespace by zarf
-				namespace.Labels[config.ZarfManagedByLabel] = "zarf"
-				namespace.Labels["zarf-helm-release"] = r.chart.ReleaseName
```
AustinAbro321, line 182: Just noticed this label while I was looking through the code. It comes into play when there are manifests or helm charts that create a namespace separate from the declared namespace. Maybe it helps when people have helm charts that release to multiple namespaces (though that's an antipattern)? But it only goes into extra namespaces that aren't declared in the Zarf chart. The label also isn't stripped on Zarf destroy. 

If we keep it I think we should start adding it to Zarf declared namespaces as well, which I'd be fine with.

src/internal/packager/helm/post-render.go
```diff
@@ -35,7 +37,16 @@ func (h *Helm) newRenderer() (*renderer, error) {
 
 	namespaces := make(map[string]*corev1.Namespace)
 	if h.cluster != nil {
-		namespaces[h.chart.Namespace] = h.cluster.NewZarfManagedNamespace(h.chart.Namespace)
+		namespace, err := h.cluster.K8s.GetNamespace(h.chart.Namespace)
+		if err != nil && !errors.IsNotFound(err) {
+			return nil, fmt.Errorf("unable to check for existing namespace %q in cluster: %s", h.chart.Namespace, err)
```
lucasrod16, line 42:
```suggestion
			return nil, fmt.Errorf("unable to check for existing namespace %q in cluster: %w", h.chart.Namespace, err)
```

src/pkg/k8s/common.go
```diff
@@ -20,11 +20,15 @@ import (
...
 )
 
-// cannot import config.ZarfManagedByLabel due to import cycle
-const zarfManagedByLabel = "app.kubernetes.io/managed-by"
+const (
+	//ZarfManagedByLabel is used to denote Zarf manages the lifecycle of a resource
+	ZarfManagedByLabel = "app.kubernetes.io/managed-by"
+	//AgentLabel is used to give instructions to the Zarf agent
+	AgentLabel = "zarf.dev/agent"
+)
```
lucasrod16, line 28:
```suggestion
const (
	// ZarfManagedByLabel is used to denote Zarf manages the lifecycle of a resource
	ZarfManagedByLabel = "app.kubernetes.io/managed-by"
	// AgentLabel is used to give instructions to the Zarf agent
	AgentLabel = "zarf.dev/agent"
)
```

src/pkg/cluster/injector.go
```diff
@@ -321,7 +321,7 @@ func (c *Cluster) buildInjectionPod(node, image string, payloadConfigmaps []stri
 	pod.Labels["app"] = "zarf-injector"
 
 	// Ensure zarf agent doesn't break the injector on future runs
-	pod.Labels[agentLabel] = "ignore"
+	pod.Labels[k8s.ZarfManagedByLabel] = "ignore"
```
lucasrod16, line 324:
```suggestion
	pod.Labels[k8s.AgentLabel] = "ignore"
```

lucasrod16 (changes requested):
Few small things on initial review. I will re-review on Monday and do a more thorough review

src/internal/packager/helm/post-render.go
```diff
@@ -35,7 +37,16 @@ func (h *Helm) newRenderer() (*renderer, error) {
 
 	namespaces := make(map[string]*corev1.Namespace)
 	if h.cluster != nil {
-		namespaces[h.chart.Namespace] = h.cluster.NewZarfManagedNamespace(h.chart.Namespace)
+		namespace, err := h.cluster.K8s.GetNamespace(h.chart.Namespace)
+		if err != nil && !errors.IsNotFound(err) {
+			return nil, fmt.Errorf("unable to check for existing namespace %q in cluster: %w", h.chart.Namespace, err)
+		}
+		if h.cfg.DeployOpts.AdoptExistingResources && !errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
phillebaba: I now get what you are doing. The error could be not found which is what you are checking for. Disregard my last comment.

src/internal/packager/helm/post-render.go
```diff
@@ -35,7 +37,16 @@ func (h *Helm) newRenderer() (*renderer, error) {
...
+			namespaces[h.chart.Namespace] = k8s.UpdateNamespaceToBeZarfManaged(namespace)
+		} else {
+			namespaces[h.chart.Namespace] = k8s.NewZarfManagedNamespace(h.chart.Namespace)
+		}
+
```
phillebaba, line 49: A suggestion to refactor the function to make it simpler to read.

```go
func (h *Helm) newRenderer() (*renderer, error) {
	message.Debugf("helm.NewRenderer()")

	rend := &renderer{
		Helm:           h,
		connectStrings: types.ConnectStrings{},
		namespaces:     map[string]*corev1.Namespace{},
	}
	if h.cluster == nil {
		return rend, nil
	}
	namespace, err := h.cluster.K8s.GetNamespace(h.chart.Namespace)
	if err != nil && !errors.IsNotFound(err) {
		return nil, fmt.Errorf("unable to check for existing namespace %q in cluster: %w", h.chart.Namespace, err)
	}
	if h.cfg.DeployOpts.AdoptExistingResources && !errors.IsNotFound(err) {
		rend.namespaces[h.chart.Namespace] = k8s.UpdateNamespaceToBeZarfManaged(namespace)
	} else {
		rend.namespaces[h.chart.Namespace] = k8s.NewZarfManagedNamespace(h.chart.Namespace)
	}
	return rend, nil
}
```

src/internal/packager/helm/post-render.go
```diff
@@ -167,22 +177,16 @@ func (r *renderer) editHelmResources(resources []releaseutil.Manifest, finalMani
...
-					namespace.Labels = make(map[string]string)
-				}
-				// Now track this namespace by zarf
-				namespace.Labels[config.ZarfManagedByLabel] = "zarf"
-				namespace.Labels["zarf-helm-release"] = r.chart.ReleaseName
+				namespace = k8s.UpdateNamespaceToBeZarfManaged(namespace)
 
 				// Add it to the stack
-				r.namespaces[namespace.Name] = &namespace
+				r.namespaces[namespace.Name] = namespace
```
lucasrod16, line 189: Suggest making this a one-liner like we do currently on line `214`:

```go
r.namespaces[namespace.Name] = k8s.UpdateNamespaceToBeZarfManaged(namespace)
```

src/internal/packager/helm/post-render.go
```diff
@@ -33,16 +35,24 @@ type renderer struct {
...
-		connectStrings: make(types.ConnectStrings),
-		namespaces:     namespaces,
-	}, nil
+		connectStrings: types.ConnectStrings{},
+		namespaces:     map[string]*corev1.Namespace{},
+	}
+	if h.cluster == nil {
+		return rend, nil
+	}
+	namespace, err := h.cluster.K8s.GetNamespace(h.chart.Namespace)
```
lucasrod16, line 46: We don't have to explicitly call `K8s` here because it is composed into `Cluster`:

```suggestion
	namespace, err := h.cluster.GetNamespace(h.chart.Namespace)
```

src/internal/packager/helm/post-render.go
```diff
@@ -33,16 +35,24 @@ type renderer struct {
...
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, fmt.Errorf("unable to check for existing namespace %q in cluster: %w", h.chart.Namespace, err)
+	}
+	if h.cfg.DeployOpts.AdoptExistingResources && !errors.IsNotFound(err) {
```
schristoff-du, line 50: Do we need the `!errors.IsNotFound` again here?
Also I would switch the statement to be like
```
if error.IsNotFound {
 NewZarfManagedNamespace
else {
UpdateNamespaceToBeZarfManaged
}
```

AustinAbro321: AFAIK there's no way around it. First check is checking if there is a true err instead of the namespace not being found. Second check if to see if the namespace doesn't already exist. 

Thanks for the rec, reads nicer without the double negative

.pre-commit-config.yaml
```diff
@@ -33,7 +33,7 @@ repos:
         entry: golangci-lint run --enable-only goimports --fix
         types: [go]
         language: golang
-        pass_filenames: true
+        pass_filenames: false
```
AustinAbro321, line 36: without this I was getting the following errors. This part of the lint is super fast anyway
```
ERRO [linters_context] typechecking error: named files must all be in one directory; have src/pkg/cluster and src/pkg/k8s 
ERRO [linters_context] typechecking error: named files must all be in one directory; have src/pkg/cluster and src/pkg/packager 
ERRO [linters_context] typechecking error: named files must all be in one directory; have src/pkg/cluster and src/config
```

lucasrod16 (commented):
Since we plan on removing the `k8s` package, do you think we should still add `GetNamespace()` as a `k8s` receiver?

We could use client set directly instead. Would require passing context to helm functions, but they should all be getting called from within a function that already has context passed to it

Example of using client set directly instead of `GetNamespace()`:

```go
func (h *Helm) newRenderer(ctx context.Context) (*renderer, error) {
	message.Debugf("helm.NewRenderer()")

	rend := &renderer{
		Helm:           h,
		connectStrings: types.ConnectStrings{},
		namespaces:     map[string]*corev1.Namespace{},
	}
	if h.cluster == nil {
		return rend, nil
	}
	namespace, err := h.cluster.Clientset.CoreV1().Namespaces().Get(ctx, h.chart.Namespace, metav1.GetOptions{})
	if err != nil && !errors.IsNotFound(err) {
		return nil, fmt.Errorf("unable to check for existing namespace %q in cluster: %w", h.chart.Namespace, err)
	}
	if errors.IsNotFound(err) {
		rend.namespaces[h.chart.Namespace] = k8s.NewZarfManagedNamespace(h.chart.Namespace)
	} else if h.cfg.DeployOpts.AdoptExistingResources {
		namespace.Labels = k8s.AdoptZarfManagedLabels(namespace.Labels)
		rend.namespaces[h.chart.Namespace] = namespace
	}

	return rend, nil
}
```

AustinAbro321:
> We could use client set directly instead.
I'm 50/50 on this. On one hand it's simpler, on the other hand I don't mind abstracting the calls away to a separate library that way. It was nice when context was implemented for example since it was a common way to change all of our k8s calls in the same package. Gonna call it direct for now, and we'll see if anyone disagrees.

src/internal/packager/helm/post-render.go
```diff
@@ -29,21 +32,34 @@ type renderer struct {
...
 	message.Debugf("helm.NewRenderer()")
 
-	namespaces := make(map[string]*corev1.Namespace)
-	if h.cluster != nil {
-		namespaces[h.chart.Namespace] = h.cluster.NewZarfManagedNamespace(h.chart.Namespace)
+	rend := &renderer{
+		Helm:           h,
+		connectStrings: types.ConnectStrings{},
+		namespaces:     map[string]*corev1.Namespace{},
+		ctx:            context.Background(),
```
lucasrod16, line 45: The `context` package docs recommend to not store contexts inside a struct type and to pass it as the first argument to the function:

> Programs that use Contexts should follow these rules to keep interfaces consistent across packages and enable static analysis tools to check context propagation:

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

```go
func DoSomething(ctx context.Context, arg Arg) error {
	// ... use ctx ...
}
```
https://pkg.go.dev/context#pkg-overview

lucasrod16 (changes requested):

src/pkg/k8s/namespace.go
```diff
@@ -20,6 +20,12 @@ func (k *K8s) GetNamespaces(ctx context.Context) (*corev1.NamespaceList, error)
 	return k.Clientset.CoreV1().Namespaces().List(ctx, metaOptions)
 }
 
+// GetNamespace gets a namespace by name
+func (k *K8s) GetNamespace(name string) (*corev1.Namespace, error) {
```
schristoff-du, line 24: Can you pass the context in `editHelmResources()` into here, it seems to be the only place this is being called

AustinAbro321: Whoops I meant to delete this function (it's actually not getting called at all anymore `editHelmResources()` calls `getNamespaces()` )

src/internal/packager/helm/post-render.go
AustinAbro321: Thanks for linking those docs, I was trying to pass the context through to the `Run()` function which satisfies a helm interface so it can't take context, but yeah we're not doing anything with the context anyway and it's better to follow go convention. I'm just going to continue creating the context in those spots

src/internal/packager/helm/post-render.go
```diff
@@ -34,16 +37,27 @@ type renderer struct {
...
+	if h.cluster == nil {
+		return rend, nil
 	}
 
-	return &renderer{
-		Helm:           h,
-		connectStrings: make(types.ConnectStrings),
-		namespaces:     namespaces,
-	}, nil
+	namespace, err := h.cluster.Clientset.CoreV1().Namespaces().Get(context.TODO(), h.chart.Namespace, metav1.GetOptions{})
```
lucasrod16, line 49: I think we are trying to avoid using `context.TODO()` if possible

```suggestion
	namespace, err := h.cluster.Clientset.CoreV1().Namespaces().Get(context.Background(), h.chart.Namespace, metav1.GetOptions{})
```

lucasrod16 (changes requested):
Small suggestion on context usage

src/internal/packager/helm/post-render.go
AustinAbro321: IMO it fits here, since we want a context to be passed through instead at some point in the future, assuming we intend to add it context to the create and dev flows.

phillebaba: I think this is the right place to use it as it is a function missing a context in its signature, but we want to add a context in the future.

src/internal/packager/helm/post-render.go
```diff
@@ -14,13 +14,16 @@ import (
...
 	"github.com/defenseunicorns/zarf/src/config"
+	"github.com/defenseunicorns/zarf/src/pkg/cluster"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 	"github.com/defenseunicorns/zarf/src/pkg/utils"
 	"github.com/defenseunicorns/zarf/src/types"
 	"helm.sh/helm/v3/pkg/releaseutil"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/yaml"
 
+	"k8s.io/apimachinery/pkg/api/errors"
```
phillebaba, line 25: NIT: Rename this import kerrors to make it the same across all packages which may import errors from std lib.

phillebaba (approved):
LGTM

lucasrod16 (approved):
lgtm

Issue #2203, opened 2023-12-19
Refactor action wait code


lucasrod16:
### Describe what should be investigated or refactored
The code for executing action waits should be refactored to make it easier to extend

### Links to any relevant code
[`src/pkg/utils/wait.go`](https://github.com/defenseunicorns/zarf/blob/main/src/pkg/utils/wait.go)

### Additional context
@Noxsios @AustinAbro321

Looked into adding a `--silent` flag to control the output of `zarf tools wait-for` in https://github.com/defenseunicorns/zarf/pull/2180 and found that the function signature for `ExecuteWait` would grow to 7 input parameters

Racer159:
This may be a good opportunity to look at making this a common library for things like Lula as well

phillebaba:
I second the point of adopting kstatus and adding a healthChecks list additionally. We had success in Flux doing it this way as it would also support health checking CRDs as long as they conformed to kstatus. Over time other projects added support for kstatus as people requested compatibility with Flux.

Noxsios:
I'm always down for removing shell abstractions, send it

phillebaba:
I have done some exploration of the existing code now to understand any potential challenges. The main one is that we need a standard method of fetching Kubernetes configuration. Right now the CLI depends on the host machine being properly configured. We may want to add a `--kubeconfig` flag to allow explicitly setting this.

phillebaba:
My suggestion for addressing this is to add health checking as a feature now that we have a shared health checking pkg. This would mimic the behavior that Flux has pretty closely. My suggestion is that we add a new field to the components that allow the users to specify their health checks.

```yaml
kind: ZarfPackageConfig
metadata:
  name: argocd
  description: Example showcasing installing ArgoCD
components:
  - name: argocd-helm-chart
    required: true
    charts:
      - name: argo-cd
        version: 5.54.0
        namespace: argocd
        url: https://argoproj.github.io/argo-helm
        releaseName: argocd-baseline
        valuesFiles:
          - baseline/values.yaml
    images:
      - docker.io/library/redis:7.0.15-alpine
      - quay.io/argoproj/argocd:v2.9.6
      # Cosign artifacts for images - argocd - argocd-helm-chart
      - quay.io/argoproj/argocd:sha256-2dafd800fb617ba5b16ae429e388ca140f66f88171463d23d158b372bb2fae08.sig
[...]

PR #2497, opened 2024-05-13
feature: new wait for resource command that uses kstatus

phillebaba:
## Description

This change adds a new `wait-for-resource` command which is meant to replace the original `wait-for` command in the future. The main difference with the new command is that it uses kstatus to determine the readiness of a resource rather than requiring users to specify a condition. This functionality is implemented using a library rather than shelling out to kubectl. The goal is to in a later PR replace the Zarf wait shorthand to use this code instead of shelling out to the tool.

## Related Issue

Relates to #2203

## Checklist before merging

### DIFF
--- a/src/cmd/tools/wait.go
+++ b/src/cmd/tools/wait.go
@@ -5,6 +5,7 @@
 package tools
 
 import (
+	"context"
 	"time"
 
 	"github.com/defenseunicorns/zarf/src/config/lang"
@@ -13,12 +14,17 @@ import (
 	"github.com/spf13/cobra"
 
 	// Import to initialize client auth plugins.
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
 var (
 	waitTimeout   string
 	waitNamespace string
+
+	waitResourceTimeout    string
+	waitResourceNamespace  string
+	waitResourceKubeconfig string
 )
 
 var waitForCmd = &cobra.Command{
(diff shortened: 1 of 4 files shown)


src/pkg/utils/wait.go
```diff
@@ -28,6 +36,49 @@ func isJSONPathWaitType(condition string) bool {
...
+	if err != nil {
+		return err
+	}
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{})
+	if err != nil {
+		return err
+	}
+	identifiers := []object.ObjMetadata{
+		{
+			GroupKind: gk,
```
phillebaba, line 50: Group has to be set for Kinds that have one. I nice thing to do would be to resolve the group for the user.

src/pkg/utils/wait.go
```diff
@@ -28,6 +36,49 @@ func isJSONPathWaitType(condition string) bool {
 	return true
 }
 
+func ExecuteWaitResource(ctx context.Context, gk schema.GroupKind, namespace, name string) error {
+	cfg, err := config.GetConfig()
```
phillebaba, line 40: Need to verify that flag `--kubeconfig` is picked up properly. Not the greatest solution, as passing it explicitly would be a lot nicer.

src/cmd/tools/wait.go
```diff
@@ -56,9 +62,40 @@ var waitForCmd = &cobra.Command{
...
+
 func init() {
 	toolsCmd.AddCommand(waitForCmd)
 	waitForCmd.Flags().StringVar(&waitTimeout, "timeout", "5m", lang.CmdToolsWaitForFlagTimeout)
 	waitForCmd.Flags().StringVarP(&waitNamespace, "namespace", "n", "", lang.CmdToolsWaitForFlagNamespace)
 	waitForCmd.Flags().BoolVar(&message.NoProgress, "no-progress", false, lang.RootCmdFlagNoProgress)
+
+	toolsCmd.AddCommand(waitForResourceCmd)
+	waitForResourceCmd.Flags().StringVar(&waitResourceTimeout, "timeout", "5m", lang.CmdToolsWaitForFlagTimeout)
+	waitForResourceCmd.Flags().StringVarP(&waitResourceKubeconfig, "kubeconfig", "f", "", lang.CmdToolsWaitForFlagNamespace)
```
phillebaba, line 98: It might be a lot better to place this flag in the root command so that it applies to all sub commands. This would simplify future credential configuration.

src/pkg/utils/wait.go
phillebaba: The flag should be global in the CLI so lets ignore this for now.

src/cmd/tools/wait.go
phillebaba: Flag should exist in the root command so lets leave that for a later PR.

src/pkg/utils/wait.go
phillebaba: Resolve this at a later time.

lucasrod16 (changes requested):
Nice work on this @phillebaba!

@brandtkeller and @Racer159 noted that they would like the functionality of `ExecuteWaitResource` available as a library function in [pkg](https://github.com/defenseunicorns/pkg) so that it's available for reuse in [Lula](https://github.com/defenseunicorns/lula) and [Maru](https://github.com/defenseunicorns/maru-runner) without having to import Zarf. This means we would move `ExecuteWaitResource` to the `pkg` repo and consume it as a library in Zarf. This would give be a good opportunity to add some unit tests to cover the behavior as well

Thoughts?

phillebaba:
I had a look at this PR again as I need the wait functionality for another feature in Zarf. My question is if this is enough for it's own package. I realize that other projects also need this functionality. Would it not be better to document how to use kstatus instead? Most of the code do use kstatus is for the most part to get the required config objects.

phillebaba:
I am closing this in favor of adding this as a feature in the package spec instead.

PR #2499, opened 2024-05-14
ci: run revive using golang-lint-ci

phillebaba:
## Description

This replaces revive with golang-ci-lint, and instead runs revive through the aforementioned tool. Additionally it replaces the header check script with the goheader linter. The benefit of this change is that we can enable more linters along with revive in the future.

## Related Issue

Part of #2503 

## Checklist before merging


.golangci.yaml
```diff
@@ -0,0 +1,20 @@
...
+    - goheader
+    - goimports
+linters-settings:
+  goheader:
+    template: |-
+      SPDX-License-Identifier: Apache-2.0
+      SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+issues:
+  exclude-dirs:
+    - "src/cmd/tools/helm"
```
phillebaba, line 20: I disabled linting the Helm tools package as it is a copy of Helm code.

.golangci.yaml
```diff
@@ -0,0 +1,20 @@
+run:
+  timeout: 5m
+linters:
+  disable-all: true
+  enable:
+    - gosimple
+    - govet
+    - staticcheck
+    - unused
+    - revive
```
AustinAbro321, line 10: I believe revive doesn't lint anything by default. I believe you're going to have to copy the rules from the revive.toml into a linter-settings format unfortunately https://golangci-lint.run/usage/linters/#revive

AustinAbro321 (changes requested):

.golangci.yaml
phillebaba: I checked the revive docs and assumed the opposite.

> Please notice that if no particular configuration is provided, revive will behave as go-lint does, i.e. all go-lint rules are enabled (the [Available Rules table](https://github.com/mgechev/revive?tab=readme-ov-file#available-rules) details what are the go-lint rules). When a configuration is provided, only rules in the configuration are enabled.

https://github.com/mgechev/revive?tab=readme-ov-file#golangci-lint

phillebaba: Fixed this by enabling rules that are by default ignored by golangci-lint.

https://golangci-lint.run/usage/false-positives/#exc0012

.pre-commit-config.yaml
```diff
@@ -37,13 +37,10 @@ repos:
...
-        name: revive go lint
-        entry: revive
+        name: golangci-lint go lint
+        entry: golangci-lint
         args:
-          - "-config"
-          - "hack/revive.toml"
-          - "-exclude"
-          - "src/cmd/viper.go"
+          - "run"
```
AustinAbro321, line 43: Looks like when we run this on pre-commit it only takes into account the single file as it will run something like `golangci-lint run src/pkg/packager/deploy.go` which can result in false positives like this.
```
src/pkg/packager/deploy.go:35:10: undefined: Packager (typecheck)
func (p *Packager) resetRegistryHPA() {
         ^
src/pkg/packager/deploy.go:44:10: undefined: Packager (typecheck)
func (p *Packager) Deploy() (err error) {
         ^
src/pkg/packager/deploy.go:123:10: undefined: Packager (typecheck)
func (p *Packager) deployComponents() (deployedComponents []types.DeployedComponent, err error) {
```

We should turn off pass_filesiles and use the --new flag imo

AustinAbro321 (changes requested):

.github/workflows/scan-lint.yml
```diff
@@ -10,12 +10,7 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-
-      - name: Run Revive Action by pulling pre-built image
-        uses: docker://morphy/revive-action:v2.5.7@sha256:087d4e61077087755711ab7e9fae3cc899b7bb07ff8f6a30c3dfb240b1620ae8
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@v6
```
Noxsios, line 14:
```suggestion
        uses: golangci/golangci-lint-action@a4f60bb28d35aeee14e6880718e0c85ff1882e64 # v6.0.1
```

Noxsios: pins so that dependabot / revive updates on sha change

Noxsios (changes requested):

.pre-commit-config.yaml
phillebaba: Took inspiration from another repo and fixed the issue.

https://github.com/golangci/golangci-lint/blob/master/.pre-commit-hooks.yaml

phillebaba:
Just FYI I realized we could remove the requirement to install goimports by just relying on golangci-lint to fix any errors for that specific linter.

.pre-commit-config.yaml
```diff
@@ -29,24 +29,16 @@ repos:
...
-        args:
-          - "-config"
-          - "hack/revive.toml"
-          - "-exclude"
-          - "src/cmd/viper.go"
-        files: .go$
-        language: system
-        pass_filenames: true
+        name: golangci-lint go lint
+        entry: golangci-lint run
```
AustinAbro321, line 38: Ready to approve this PR, but thoughts on using the `--new` flag here to speed up lint time? I know it they say here that linters like unused won't work with `--new` but I wasn't able to reproduce that https://github.com/golangci/golangci-lint/blob/master/.pre-commit-hooks.yaml#L3

phillebaba: I would rather have all linter rules run, as that is the way it is going to run in the CI. You are not going to save any time if unused would have failed if run, and you will have to wait for the CI to run before you figure that out.

Noxsios: @AustinAbro321 for your own local dev, you can always edit your `.pre-commit-config.yaml` to add the `--new` and just not commit it.  Not the best solution, but it should satisfy all parties for now, thoughts?

AustinAbro321: Yeah I'm fine with it running without the lint flag just wanted to get Phillips thoughts. Approving now.

AustinAbro321 (approved):

Noxsios (approved):

PR #2501, opened 2024-05-14
refactor: enable errcheck linter

phillebaba:
## Description

This change enables the errcheck linter which requires all returned errors to be checked.

## Related Issue

Part of #2503
Depends on #2499 

## Checklist before merging

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -3,6 +3,7 @@ run:
 linters:
   disable-all: true
   enable:
+    - errcheck
     - gosimple
     - govet
     - ineffassign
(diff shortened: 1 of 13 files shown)


.golangci.yaml
```diff
@@ -54,6 +55,12 @@ linters-settings:
       - name: redefines-builtin-id
   testifylint:
     enable-all: true
+  errcheck:
+    exclude-functions:
+      - (*github.com/spf13/cobra.Command).Help
+      - (*github.com/spf13/cobra.Command).MarkFlagRequired
+      - (*github.com/spf13/pflag.FlagSet).MarkHidden
+      - (*github.com/spf13/pflag.FlagSet).MarkDeprecated
```
AustinAbro321, line 63: Should we add the response writer and pterm function we're ignoring here as well?

phillebaba: These are things I do want to handle in the future but it is just not ideal right now. Either because it would greatly change function signatures or because the logger isn't present yet.

AustinAbro321 (approved):

PR #2513, opened 2024-05-16
refactor: remove k8s package in favor of using k8s client directly

### LINKED ISSUE #2507: Remove k8s package

The k8s package with exception for a couple of functions simply wraps a Kubernetes clientset with very little value add. The functions return structs from k8s api package so the current implementation does not decouple the dependency towards these packages. This just increases bloat and complexity without any functionality. I suggest that we remove the k8s package and instead export a clientset in the Cluster struct. Allowing the calling code to use the clientset directly.

https://github.com/defenseunicorns/zarf/blob/48c11d99547b59ff2f10fa60078741376059f821/src/pkg/k8s/pods.go#L81
https://github.com/defenseunicorns/zarf/blob/48c11d99547b59ff2f10fa60078741376059f821/src/pkg/k8s/secrets.go#L18

### Additional context

N/A

phillebaba:
## Description

This change removes the k8s package completely. I have attempted to make minimal changes on purpose due to the lack of unit tests. Most of these changes are either moving existing code to a new package or replacing k8s package function calls with calls to the clientset. After this is done we can hopefully start writing unit tests and refactoring the code.

## Related Issue

Fixes #2507

## Checklist before merging


phillebaba:
After working on this for a while I realize that it is going to be very difficult to get a mega change merged without introducing bugs. This is partly due to lack of tests, but also unexplained behavior by some code.

I will leave this PR for reference, but will split it into multiple PRs that will leave the k8s package until no dependencies towards it exist anymore.

phillebaba:
Closing as most of these things have been split into their own PRs.

Issue #2512, opened 2024-05-16
Zarf Test Week MegaIssue


schristoff-du:
Goal: This issue will lead test week work

Single Issues:
- [x]  #2514 
- [ ] #2477 
- [ ] #2364 
- [ ] #2357 
- [x] #1615
- [ ] #1409 
- [ ] #1750 

Unit Tests (focus areas):
- [ ] src/cmd
- [x] src/internal/agent
- [ ] pkg/packager
    - [ ]  pkg/packager/sources
     - [ ]  pkg/packager/filters
     - [ ]  pkg/packager/composer
     - [ ]  pkg/packager/creator
- [ ] src/pkg/cluster
    
E2E:
- [ ]  Identify if E2E is running sequentially (if yes, resolve. If no, remove numbering system)
- [ ] Remove retries and waits where possible (this is to reduce the amount of time e2e takes to run)

PR #2516, opened 2024-05-16
test: add unit tests for mutating webhooks

lucasrod16:
## Description
Adds unit tests for the agent's mutating webhooks

## Related Issue

Relates to #2512

## Checklist before merging


phillebaba (approved):
I think this looks good, a simple way of testing the admission hook.

My only nitpick is the package structure. Why are we creating yet another sub package in the http package? It seems like all of the admission related code should be contained in the same package instead of being spread out across different pacakges.

lucasrod16:
> I think this looks good, a simple way of testing the admission hook.
> 
> My only nitpick is the package structure. Why are we creating yet another sub package in the http package? It seems like all of the admission related code should be contained in the same package instead of being spread out across different packages.

@phillebaba I structured it this way to avoid import cycle error. Package `http` imports `hooks`, so when I was trying to use `http.NewHandler()`, I got an import cycle error, so I moved it to a sub package `admission` to resolve. Open to other structures

PR #2518, opened 2024-05-17
fix(agent): missing path for pod without labels

### LINKED ISSUE #2517: [bug] Pod without labels will fail mutation
### Environment
Device and OS: AMD64 linux Server (ubuntu)
App version: v0.33.2
Kubernetes distro being used: Kind

### Steps to reproduce
1. Deploy zarf init package
2. Create a pod that contains no labels

### Expected result
Pod should be created and mutated by the zarf agent

### Actual Result
internal server error
```
Internal error occurred: replace operation does not apply: doc is missing path: /metadata/labels/zarf-agent: missing value
```

### Additional Context
Likely an edge case that does not occur often due to many clients adding default labels.

brandtkeller:
## Description

identify when a pod has no labels (and thus has no `/metadata/labels` for jsonpatch replace operations. use the `add` jsonpath operation to add the labels path with the map[string]string value of `{"zarf-agent": "patched"}`

## Related Issue

Fixes #2517 

## Testing
Deploy a pod without labels to test:
```
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: null
  name: test
  namespace: test
spec:
  containers:
  - image: nginx
    name: test
    resources: {}
  dnsPolicy: ClusterFirst
  restartPolicy: Always
status: {}
```

## Checklist before merging

### DIFF
--- a/src/internal/agent/hooks/pods.go
+++ b/src/internal/agent/hooks/pods.go
@@ -100,7 +100,7 @@ func mutatePod(r *v1.AdmissionRequest) (*operations.Result, error) {
 
 	// Add a label noting the zarf mutation
 	if pod.Labels == nil {
-		// If the labels path does not exist - create with nap[string]string value
+		// If the labels path does not exist - create with map[string]string value
 		patchOperations = append(patchOperations, operations.AddPatchOperation("/metadata/labels",
 			map[string]string{
 				"zarf-agent": "patched",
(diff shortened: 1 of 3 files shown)


justinthelaw (dismissed):

lucasrod16 (changes requested):
Thanks for the PR! Would you mind adding a test case to help prevent the regression from reoccurring?

lucasrod16 (approved):
lgtm! Thanks for the PR

PR #2519, opened 2024-05-17
chore: add unit tests to variables pkg

Racer159:
## Description

This moves the unit tests from pkg (incorporating feedback from over there)

## Related Issue

Fixes #N/A

## Checklist before merging

- [X] [Contributor Guide Steps](https://github.com/defenseunicorns/zarf/blob/main/.github/CONTRIBUTING.md#developer-workflow) followed


AustinAbro321 (dismissed):

src/pkg/variables/templates_test.go
```diff
@@ -0,0 +1,157 @@
...
+				t.Fatalf("wanted err: %s, got err: %s", tc.wantErr, gotErr)
+			}
+		} else if gotErr != nil {
+			t.Fatalf("got unexpected err: %s", gotErr)
+		} else {
+			gotContents, _ := os.ReadFile(tc.path)
+			if string(gotContents) != tc.wantContents {
+				t.Fatalf("wanted contents: %s, got contents: %s", tc.wantContents, string(gotContents))
+			}
+		}
```
Noxsios, line 154: this can most likely be simplified w/ something similar to:

```go
			if tc.expectedErr != nil {
				require.ErrorIs(t, err, tc.expectedErr)
			}
			require.NoError(t, err)
			...
```

src/pkg/variables/variables.go
```diff
@@ -76,9 +76,13 @@ func (vc *VariableConfig) SetVariable(name, value string, sensitive bool, autoIn
 
 // CheckVariablePattern checks to see if a current variable is set to a value that matches its pattern
 func (vc *VariableConfig) CheckVariablePattern(name, pattern string) error {
-	if regexp.MustCompile(pattern).MatchString(vc.setVariableMap[name].Value) {
-		return nil
+	if variable, ok := vc.setVariableMap[name]; ok {
+		if regexp.MustCompile(pattern).MatchString(variable.Value) {
```
Noxsios, line 80: in this case, should use `Compile`, `MustCompile` will panic if a bad pattern is used

src/pkg/variables/variables_test.go
Noxsios, line 1: should refactor from `t.Fatal` to `require` calls for consistency and clarity across tests

Noxsios (changes requested):

src/pkg/variables/variables.go
```diff
@@ -76,9 +76,18 @@ func (vc *VariableConfig) SetVariable(name, value string, sensitive bool, autoIn
...
 // CheckVariablePattern checks to see if a current variable is set to a value that matches its pattern
 func (vc *VariableConfig) CheckVariablePattern(name, pattern string) error {
-	if regexp.MustCompile(pattern).MatchString(vc.setVariableMap[name].Value) {
+	if variable, ok := vc.setVariableMap[name]; ok {
+		r, err := regexp.Compile(pattern)
+		if err != nil {
+			return err
+		}
+
+		if !r.MatchString(variable.Value) {
```
schristoff-du, line 85:
```suggestion
		if r.MatchString(variable.Value) {
		return nil 
		}
		return fmt.Errorf("thaterrorhere")
```

src/pkg/variables/variables_test.go
```diff
@@ -0,0 +1,167 @@
...
+				"NAME": {Variable: Variable{Name: "NAME", Sensitive: true, AutoIndent: true, Type: FileVariableType}, Value: "Set"},
+			},
+		},
+		{
+			vc: VariableConfig{setVariableMap: SetVariableMap{}, prompt: prompt},
+			vars: []InteractiveVariable{
+				{Variable: Variable{Name: "NAME"}, Prompt: true},
+			},
+			presets:    map[string]string{},
+			wantErrMsg: "",
```
schristoff-du, line 81: nit again w/ bool instead of string for wantErrMsg

schristoff-du (approved):

lucasrod16 (approved):

Noxsios (approved):

PR #2521, opened 2024-05-20
test: add unit tests for detect distro

phillebaba:
## Description

Adds tests to detect distro logic.

## Related Issue

Relates to #2512

## Checklist before merging


src/pkg/cluster/distro_test.go
```diff
@@ -0,0 +1,181 @@
...
+					},
+				},
+			},
+		},
+		{
+			distro: DistroIsK3s,
+			node: corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"": "node.kubernetes.io/instance-type=k3s",
```
phillebaba, line 53: This passes but I think the logic is actually wrong.

src/pkg/cluster/distro_test.go
```diff
@@ -0,0 +1,181 @@
...
+					ProviderID: "kind://docker/kind/kind-control-plane",
+				},
+			},
+		},
+		{
+			distro: DistroIsMicroK8s,
+			node: corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"": "microk8s.io/cluster=true",
```
phillebaba, line 79: This passes but I think the logic is actually wrong.

src/pkg/cluster/distro.go
```diff
@@ -0,0 +1,101 @@
...
+		return DistroIsGKE
+	}
+
+	// https://github.com/kubernetes/kubernetes/blob/v1.23.4/staging/src/k8s.io/legacy-cloud-providers/azure/azure_wrap.go#L46
+	if aksNodeRegex.MatchString(node.Spec.ProviderID) {
+		return DistroIsAKS
+	}
+
+	labels := node.GetLabels()
+	for _, label := range labels {
```
phillebaba, line 66: This looks wrong. It assumes the label is in selector format. In reality is should look at key and value instead.

Noxsios (commented):
LGTM, just needs some merge nonsense cleaned up from (some stuff referenced no longer exists in the `k8s` package), running `make build` should show the offending imports.

AustinAbro321 (approved):

PR #2525, opened 2024-05-21
test: add test for packager source

phillebaba:
## Description

This change adds tests for packager source which tests all implementations.

## Related Issue

Relates to #2512 

## Checklist before merging


src/pkg/packager/sources/new_test.go
```diff
@@ -5,49 +5,168 @@
...
+			ps, err := New(opts)
+			require.NoError(t, err)
+			packageDir := t.TempDir()
+			pkgLayout := layout.New(packageDir)
+			pkg, warnings, err := ps.LoadPackage(pkgLayout, filters.Empty(), false)
+			require.NoError(t, err)
+			require.Empty(t, warnings)
+			require.Equal(t, expectedPkg, pkg)
+
+			ps, err = New(opts)
```
phillebaba, line 155: I dont really understand why the package source isn't idempotent. This is forcing me to create a new instance for every check. This should be explored in the future.

src/pkg/packager/sources/tarball.go
```diff
@@ -204,5 +204,31 @@ func (s *TarballSource) LoadPackageMetadata(dst *layout.PackagePaths, wantSBOM b
...
+	dstFile, err := os.Create(linkErr.New)
+	if err != nil {
+		return "", err
+	}
+	defer dstFile.Close()
+	_, err = io.Copy(dstFile, srcFile)
+	if err != nil {
+		return "", err
+	}
+	return dst, nil
```
AustinAbro321, line 232: I believe we want to delete the file from the original partition here. 

Also what are your thoughts on the "github.com/otiai10/copy".Copy() method? Is this simple enough to not be library worthy ?

src/pkg/packager/sources/testdata/expected-pkg.json
AustinAbro321, line 1: Tests are currently expecting a zarf package, which is likely git ignored. I would be okay with committing it since it's just plaintext but will leave that you to you / the team

`open ./testdata/zarf-package-wordpress-amd64-16.0.4.tar.zst: no such file or directory`

AustinAbro321 (changes requested):

src/pkg/packager/sources/new_test.go
Noxsios: At the time I wrote sources, idempotence was not a consideration I was thinking about.

https://github.com/defenseunicorns/zarf/blob/f2d360261da70e656b090aa2f3b9bf036cf1dc19/src/pkg/packager/sources/new.go#L22-L30

In retrospect, I should have, as it allows for muuch better testing + repeatability + fuzzing.

afaik, every `zarf package deploy | inspect | remove | pull | publish` still only call 1 of these load functions, with some being destructive.

src/pkg/packager/sources/tarball.go
phillebaba: You are completely right.

I think that it could be worth discussing in a different PR, we probably need some use case and benchmarks to explain why we would pull in another dependency for this.

src/pkg/packager/sources/new_test.go
phillebaba: I will let the tests remain as they are, then we can revisit this in the future. The docs helped be understand what was happening. I just feel like this could cause problems in the future if used improperly.

lucasrod16 (approved):

PR #2526, opened 2024-05-21
test: pod agent unit tests

AustinAbro321:
## Description

This also changes how we the agent gets secrets from the cluster. Instead of mounting secrets in the deployment and reading from a file the agent will instead read the secret directly from the cluster. This is done for simplicity, and easier testing. Additionally, the agent will need to read from the cluster in #1974 as well so this helps prepare for that change. 

This only changes and tests pods to keep the scope of the PR slim. This means temporarily the agent will both read from mounted file secrets and regular kubernetes secrets depending on the hook, but this will be changed shortly in future PRs. 

## Related Issue

Relates to #2512 

## Checklist before merging


src/internal/agent/hooks/pods_test.go
```diff
@@ -0,0 +1,149 @@
...
+				),
+			},
+			code: http.StatusOK,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
```
schristoff-du, line 136: [Interesting, I've never seen this but I like it!](https://engineering.mercari.com/en/blog/entry/20220408-how_to_use_t_parallel/)

AustinAbro321: credit to @lucasrod16!

src/internal/agent/hooks/pods_test.go
```diff
@@ -0,0 +1,149 @@
...
+			resp := sendAdmissionRequest(t, tt.admissionReq, handler, tt.code)
+			if tt.expectedPatch != nil {
```
schristoff-du, line 138: I would swap if/else to not `!` if can

schristoff-du (dismissed):
🫡

packages/zarf-agent/manifests/rolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: zarf-agent-binding
```
phillebaba, line 4: Why do we set the namespace in the subjects but not here?

phillebaba (commented):
Small comment then I can approve.

packages/zarf-agent/manifests/rolebinding.yaml
AustinAbro321: Good catch, technically doesn't matter since Zarf will set them anyway because the zarf namespace is declared in the manifest block, but I added it for clarity

lucasrod16 (approved):

PR #2528, opened 2024-05-21
test: agent flux unit test

AustinAbro321:
## Description

Relates to #2512 

## Checklist before merging


src/config/lang/english.go
```diff
@@ -665,10 +665,12 @@ const (
 	AgentErrBindHandler            = "Unable to bind the webhook handler"
 	AgentErrCouldNotDeserializeReq = "could not deserialize request: %s"
 	AgentErrGetState               = "failed to load zarf state: %w"
+	AgentErrParsePod               = "failed to parse pod: %w"
 	AgentErrHostnameMatch          = "failed to complete hostname matching: %w"
 	AgentErrImageSwap              = "Unable to swap the host for (%s)"
 	AgentErrInvalidMethod          = "invalid method only POST requests are allowed"
 	AgentErrInvalidOp              = "invalid operation: %s"
+	AgentErrTransformGitURL        = "unable to transform the git url"
```
AustinAbro321, line 673: Formatted this a bit differently (without %w) than the other errors so I could use this constant in my tests, lmk if you disagree with this approach

src/internal/agent/hooks/flux.go
```diff
@@ -98,17 +90,12 @@ func mutateGitRepo(r *v1.AdmissionRequest) (result *operations.Result, err error
...
 
-	// If a prior secret exists, replace it
-	if secretName != "" {
-		patches = append(patches, operations.ReplacePatchOperation("/spec/secretRef/name", config.ZarfGitServerSecretName))
-	} else {
-		// Otherwise, add the new secret
-		patches = append(patches, operations.AddPatchOperation("/spec/secretRef", SecretRef{Name: config.ZarfGitServerSecretName}))
-	}
+	newSecretRef := fluxmeta.LocalObjectReference{Name: config.ZarfGitServerSecretName}
+	patches = append(patches, operations.AddPatchOperation("/spec/secretRef", newSecretRef))
```
AustinAbro321, line 98: The way mutating webhooks work if you have a patch operation it will just override the entire object so we don't need to conditionally replace the secretname since addpatch has the same effect. There are no fields on secretref besides name

lucasrod16 (dismissed):
lgtm

lucasrod16 (approved):

src/config/lang/english.go
Noxsios: although since this is only used internally, i'm fine w/ keeping it in `english`, its an easy fix later

feel free to resolve if you feel strongly enough about keeping it

src/internal/agent/http/admission/handler.go
```diff
@@ -68,27 +68,41 @@ func (h *Handler) Serve(hook operations.Hook) http.HandlerFunc {
 		}
 
 		result, err := hook.Execute(review.Request)
+		admissionMeta := metav1.TypeMeta{
+			APIVersion: corev1.SchemeGroupVersion.String(),
+			Kind:       "AdmissionReview",
```
Noxsios, line 73: is there a const we can use here?

Noxsios (changes requested):

go.mod
```diff
@@ -227,7 +227,7 @@ require (
 	github.com/felixge/httpsnoop v1.0.4 // indirect
 	github.com/fluxcd/pkg/apis/acl v0.1.0 // indirect
 	github.com/fluxcd/pkg/apis/kustomize v1.3.0 // indirect
-	github.com/fluxcd/pkg/apis/meta v1.3.0 // indirect
+	github.com/fluxcd/pkg/apis/meta v1.3.0
```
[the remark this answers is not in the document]
AustinAbro321: Huh weird, go mod tidy doesn't fix that automatically

src/internal/agent/http/admission/handler.go
AustinAbro321: I can't find any

src/config/lang/english.go
AustinAbro321: Gonna move just this one to the package for now, leaving a TODO to move the rest of the agent errors in a future refactor (there's other errors I want to move as well)

Noxsios (approved):

PR #2529, opened 2024-05-21
chore: add codecov

schristoff-du:


.github/workflows/test-unit.yml
```diff
@@ -31,3 +31,8 @@ jobs:
 
       - name: Run unit tests
         run: make test-unit
```
lucasrod16, line 33: Do we want to change our `test-unit` make target so that it generates a code coverage report to upload?

image

https://github.com/defenseunicorns/zarf/actions/runs/9178093280/job/25237190251?pr=2529#step:5:44

https://github.com/defenseunicorns/zarf/blob/29d3ee39dfcbe173405f6e0010cd0dfb9468cd6e/Makefile#L209-L213

.github/workflows/test-unit.yml
```diff
@@ -31,3 +31,8 @@ jobs:
...
+
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v4.0.1
```
lucasrod16, line 36: Can we pin by hash? latest version is `v4.4.1`

https://github.com/codecov/codecov-action/commit/125fc84a9a348dbcf27191600683ec096ec9021c

```suggestion
        uses: codecov/codecov-action@125fc84a9a348dbcf27191600683ec096ec9021c # v4.4.1
```

.github/workflows/test-unit.yml
schristoff-du: I don't think it needs to be in the make target, but I did add the line we talked about earlier.

lucasrod16: Nice! We should be able to remove make test-unit from this workflow, because now this workflow runs the unit tests twice

Noxsios (approved):

PR #2531, opened 2024-05-21
chore: add unit tests for creator.LoadPackageDefinition

lucasrod16:
Relates to #2512


src/pkg/packager/creator/normal_test.go
```diff
@@ -56,3 +58,44 @@ func TestDifferentialPackagePathSetCorrectly(t *testing.T) {
...
+
+	tests := []struct {
+		name      string
+		testDir   string
+		expectErr bool
+	}{
+		{
+			name:      "valid package definition",
+			testDir:   "valid",
+			expectErr: false,
```
phillebaba, line 73: Would be nice to change this to expectedErr that is a string. That way you can do require.EqualError to see if the errors match. Alternatively do a require.ErrorContains.

src/pkg/packager/creator/normal_test.go
```diff
@@ -56,3 +58,44 @@ func TestDifferentialPackagePathSetCorrectly(t *testing.T) {
...
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			src := layout.New(filepath.Join("testdata", tt.testDir))
+			pc := NewPackageCreator(types.ZarfCreateOptions{}, "")
+			pkg, _, err := pc.LoadPackageDefinition(src)
+
+			switch {
+			case tt.expectErr:
```
phillebaba, line 92: Change this to a if statement which checks if expected error and the returns after check, becomes much cleaner.

src/pkg/packager/creator/normal.go
```diff
@@ -119,6 +119,11 @@ func (pc *PackageCreator) LoadPackageDefinition(dst *layout.PackagePaths) (pkg t
 		}
 	}
 
+	// Perform early package validation.
+	if err := pkg.Validate(); err != nil {
+		return types.ZarfPackage{}, nil, fmt.Errorf("unable to validate package: %w", err)
+	}
+
 	return pkg, warnings, nil
```
Noxsios, line 127: if the error returned from `pkg.Validate` is descriptive enough / does not require additional context, you can always:

```go
return pkg, warnings, pkg.Validate()
```

Noxsios (changes requested):

src/pkg/packager/creator/normal_test.go
lucasrod16: updated to check error string value

src/pkg/packager/creator/normal.go
lucasrod16: Good point. I don't think the error returned from `Validate` needs extra context here so I changed to just return the `err` directly instead of wrapping.

I chose to leave the call to `pkg.Validate` the same rather than in a return statement because I want to ensure we always return an empty pkg on error because it's easier to test

src/pkg/packager/creator/testdata/invalid/zarf.yaml
```diff
@@ -0,0 +1,5 @@
+kind: ZarfPackageConfig
+metadata:
+  name: minimal-invalid
+  # must have at least 1 component
+  description: Minimal invalid package
```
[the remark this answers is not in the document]
lucasrod16: updated

src/types/validate.go
```diff
@@ -47,6 +48,10 @@ func (pkg ZarfPackage) Validate() error {
 		return fmt.Errorf(lang.PkgValidateErrPkgName, pkg.Metadata.Name)
 	}
 
+	if pkg.Components == nil {
```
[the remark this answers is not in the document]
lucasrod16: good catch! updated

src/pkg/packager/creator/normal.go
Noxsios: sounds good to me, feel free to resolve

Noxsios (approved):

PR #2535, opened 2024-05-22
test: add tunnel tests

phillebaba:
## Description

Adds tests to cluster tunnel.

## Related Issue

Relates to #2512 

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/tunnel_test.go
+++ b/src/pkg/cluster/tunnel_test.go
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+package cluster
+
+import (
+	"context"
+	"testing"
+
+	"github.com/defenseunicorns/zarf/src/config"
+	"github.com/defenseunicorns/zarf/src/pkg/k8s"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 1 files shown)


phillebaba:
The tunnel connect function is currently too difficult to test. It will require some refactoring to split out the logic from the functionality to increase code coverage.

src/pkg/cluster/tunnel_test.go
```diff
@@ -0,0 +1,76 @@
...
+				Name:      "missing-label",
+				Namespace: "",
+			},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "wrong-label",
+				Namespace: "",
+				Labels: map[string]string{
+					config.ZarfConnectLabelName: "wrong",
```
schristoff-du, line 40: Shouldn't these error?

src/pkg/cluster/tunnel_test.go
```diff
@@ -0,0 +1,76 @@
...
+		require.NoError(t, err)
+	}
+
+	ti, err := c.checkForZarfConnectLabel(ctx, "good")
+	require.NoError(t, err)
+	require.Equal(t, k8s.SvcResource, ti.resourceType)
+	require.Equal(t, "good-service", ti.resourceName)
+	require.Equal(t, "good-namespace", ti.namespace)
+	require.Equal(t, 9090, ti.remotePort)
+	require.Equal(t, "foobar", ti.urlSuffix)
```
lucasrod16, line 75: It looks like this test primarily covers the happy path. I noticed the missing-label and wrong-label test cases, and I assume you would like to test those error paths as well.

I think we can iterate over the `svcs` and pass in `svc.Name` as an argument to `checkForZarfConnectLabel`

lucasrod16 (changes requested):

PR #2538, opened 2024-05-22
chore: fix codecov

schristoff:


.github/workflows/test-unit.yml
```diff
@@ -25,17 +25,17 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Setup golang
+        uses: ./.github/actions/golang
       
       - name: Run coverage
-        run: go test -race -coverprofile=coverage.out -covermode=atomic
+        run: go test $(go list ./... | grep -v '^github.com/defenseunicorns/zarf/src/test' | grep -v 'github.com/defenseunicorns/zarf/src/extensions/bigbang/test') -race -coverprofile=coverage.out -covermode=atomic
```
AustinAbro321, line 33: IMO we should change make test-unit to this command and add coverage.out to the gitignore

AustinAbro321 (changes requested):

.github/workflows/test-unit.yml
schristoff-du: That was @lucasrod16 initial thing, but I didn't know if people would want this everytime they ran things locally. Seemed like it would add a bunch of time to local stuff (which already takes forever), but I am not that strongly opposed to it

AustinAbro321: Locally for me since we're only running unit tests this only takes about 11 seconds and stayed pretty much the same with / without coverage. Using -count 1 to avoid caching on the time test

lucasrod16: > IMO we should change make test-unit to this command and add coverage.out to the gitignore

I agree

lucasrod16 (approved):

PR #2541, opened 2024-05-23
refactor: remove use of k8s configmap

phillebaba:
## Description

This change removes the k8s configmap code and replaces it with direct use of the client set instead. This increases legibility of the code. Technically this also helps with code coverage as it removes code that has no tests.

## Related Issue

Relates to #2512 

## Checklist before merging


schristoff-du (dismissed):
love this

schristoff-du (approved):

PR #2543, opened 2024-05-23
refactor: remove use of k8s service

phillebaba:
## Description

This change removes the k8s service code and replaces it with direct use of the client set instead.

## Related Issue

Relates to #2512 

## Checklist before merging


src/pkg/packager/deploy.go
```diff
@@ -511,13 +513,11 @@ func (p *Packager) pushReposToRepository(ctx context.Context, reposPath string,
...
-			svcInfo, _ := k8s.ServiceInfoFromServiceURL(gitClient.Server.Address)
-
-			var err error
-			var tunnel *k8s.Tunnel
+			namespace, name, port, err := serviceInfoFromServiceURL(gitClient.Server.Address)
 
 			// If this is a service (svcInfo is not nil), create a port-forward tunnel to that resource
-			if svcInfo != nil {
+			// TODO: Find a better way as ignoring the error is not a good solution to decide to port forward.
+			if err == nil {
```
Noxsios, line 520: we generally have shied away from using `err == nil` for control flow, but not sure what the best way forward is here

Noxsios (approved):

src/pkg/packager/deploy.go
phillebaba: Yeah I get why it was done, but I almost broke this on mistake when I at first flipped the err check. I added a TODO so that future Unicorns can fix this.

PR #2545, opened 2024-05-23
refactor: allow callers to directly set logfile location

### LINKED ISSUE #2524: lib: configure log file prefix
UDS CLI uses Zarf as a lib to deploy packages, and we'd like to be able to [set the log prefix](https://github.com/defenseunicorns/zarf/blob/29d3ee39dfcbe173405f6e0010cd0dfb9468cd6e/src/pkg/message/message.go#L85) when configuring logs

- **Given** a user is using UDS CLI to deploy a Zarf pkg
- **When** the user deploys the app
- **Then** the generated log file start with `uds` as opposed to `zarf`

Noxsios:
Changes `message.UseLogFile` to accept an `f *os.File` and wrap it in a `*message.PausableWriter`. Adds unit tests for `PausableWriter`.

## Related Issue

Fixes #2524 

## Checklist before merging


schristoff-du (approved):
I feel like there is another way to approach not printing credentials to logfiles, but that is out of scope of this PR.

lucasrod16 (approved):

PR #2550, opened 2024-05-28
refactor: remove use of k8s namespace

phillebaba:
## Description

This change removes the k8s namespace code and replaces it with direct use of the client set instead.

## Related Issue

Relates to #2507

## Checklist before merging


src/pkg/packager/deploy.go
```diff
@@ -451,7 +454,9 @@ func (p *Packager) setupState(ctx context.Context) (err error) {
 		// Try to create the zarf namespace
 		spinner.Updatef("Creating the Zarf namespace")
 		zarfNamespace := cluster.NewZarfManagedNamespace(cluster.ZarfNamespaceName)
-		if _, err := p.cluster.CreateNamespace(ctx, zarfNamespace); err != nil {
+		// NOTE: Should this ignore if already exists?
+		_, err := p.cluster.Clientset.CoreV1().Namespaces().Create(ctx, zarfNamespace, metav1.CreateOptions{})
```
phillebaba, line 458: Original implementation skips create if namespace already exists. Is this required here?

src/pkg/cluster/state.go
```diff
@@ -107,7 +108,8 @@ func (c *Cluster) InitZarfState(ctx context.Context, initOptions types.ZarfInitO
 		// Try to create the zarf namespace.
 		spinner.Updatef("Creating the Zarf namespace")
 		zarfNamespace := NewZarfManagedNamespace(ZarfNamespaceName)
-		if _, err := c.CreateNamespace(ctx, zarfNamespace); err != nil {
+		// NOTE: Should this ignore if namespace exists?
+		_, err := c.Clientset.CoreV1().Namespaces().Create(ctx,zarfNamespace, metav1.CreateOptions{})
```
phillebaba, line 112: Original implementation skips create if namespace already exists. Is this required here?

src/internal/packager/helm/post-render.go
```diff
@@ -117,29 +118,35 @@ func (r *renderer) Run(renderedManifests *bytes.Buffer) (*bytes.Buffer, error) {
...
+		for _, serverNamespace := range namespaceList.Items {
 			if serverNamespace.Name == name {
 				existingNamespace = true
 			}
 		}
 
 		if !existingNamespace {
 			// This is a new namespace, add it
-			if _, err := c.CreateNamespace(ctx, namespace); err != nil {
+			_, err := c.Clientset.CoreV1().Namespaces().Create(ctx, namespace, metav1.CreateOptions{})
```
phillebaba, line 137: I assume we can skip checking before create here as it is already done.

src/pkg/cluster/namespace.go
```diff
@@ -18,7 +20,28 @@ func (c *Cluster) DeleteZarfNamespace(ctx context.Context) error {
...
+			_, err := c.Clientset.CoreV1().Namespaces().Get(ctx, ZarfNamespaceName, metav1.GetOptions{})
+			if kerrors.IsNotFound(err) {
+				return nil
+			}
+			if err != nil {
+				return err
+			}
+			timer.Reset(1 * time.Second)
+		}
+	}
```
phillebaba, line 44: Is there a purpose for waiting for the namespace to be completely removed?

src/pkg/packager/deploy.go
lucasrod16: It's not necessary to check if the namespace exists before attempting to create it. However, we should handle the case where the `zarf` namespace already exists.

We can use the [`k8s.io/apimachinery/pkg/api/errors`](https://pkg.go.dev/k8s.io/apimachinery@v0.29.1/pkg/api/errors#pkg-overview) package to check if the error indicates that the namespace already exists:

```go
_, err := p.cluster.Clientset.CoreV1().Namespaces().Create(ctx, zarfNamespace, metav1.CreateOptions{})
if err != nil && !kerrors.IsAlreadyExists(err) {
	spinner.Fatalf(err, "Unable to create the zarf namespace")
}
```

This will fix the current failure in the `DevDeploy` test

lucasrod16 (changes requested):
Looks like this PR needs `go mod tidy` ran

src/pkg/packager/deploy.go
phillebaba: Ok that is good to know that it is actually required. In that case this should be a update or create function to make sure that any metadata changes are applied.

src/pkg/cluster/namespace.go
phillebaba: Lets keep this for now.

src/pkg/cluster/state.go
phillebaba: It should not be required as this only runs if state does not already exist.

phillebaba:
I want to get #2556 before this as it adds some tests to Zarf init state which resolves some questions that I have had before.

AustinAbro321 (dismissed):

src/internal/packager/helm/post-render.go
```diff
@@ -117,29 +118,35 @@ func (r *renderer) Run(renderedManifests *bytes.Buffer) (*bytes.Buffer, error) {
...
+			if err != nil {
 				return fmt.Errorf("unable to create the missing namespace %s", name)
 			}
 		} else if r.cfg.DeployOpts.AdoptExistingResources {
-			if r.cluster.IsInitialNamespace(name) {
+			// IsInitialNamespace returns true if the given namespace name is an initial k8s namespace: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#initial-namespaces
```
lucasrod16, line 142: Can we update this comment? The `IsInitialNamespace` function is deleted in this PR

lucasrod16 (changes requested):

src/internal/packager/helm/post-render.go
phillebaba: I also realized that the logic was wrong so I fixed it. Checking for the prefix would give false positives for namespaces that begin with `kube-`.

lucasrod16: Yeah I think it's fine to be more precise with the default namespace checks. The k8s docs recommend to avoid creating namespaces with `kube-` prefix, but we don't necessarily have to enforce that in our logic.

https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#working-with-namespaces

> Avoid creating namespaces with the prefix `kube-`, since it is reserved for Kubernetes system namespaces.

lucasrod16 (approved):

PR #2553, opened 2024-05-28
refactor: remove use of k8s pods

phillebaba:
## Description

Removes use of k8s pod functions.

## Related Issue

Relates to #2507

## Checklist before merging


src/internal/packager/helm/zarf.go
```diff
@@ -61,18 +62,21 @@ func (h *Helm) UpdateZarfAgentValues(ctx context.Context) error {
...
 
 	// Get the current agent image from one of its pods.
-	pods := h.cluster.WaitForPodsAndContainers(
-		ctx,
-		k8s.PodLookup{
-			Namespace: cluster.ZarfNamespaceName,
-			Selector:  "app=agent-hook",
-		},
-		nil,
-	)
```
phillebaba, line 71: I do not see any reason why this should wait for pods that match the selector.

src/internal/packager/helm/zarf.go
```diff
@@ -124,13 +128,20 @@ func (h *Helm) UpdateZarfAgentValues(ctx context.Context) error {
...
 
 	// Force pods to be recreated to get the updated secret.
-	err = h.cluster.DeletePods(
-		ctx,
-		k8s.PodLookup{
-			Namespace: cluster.ZarfNamespaceName,
-			Selector:  "app=agent-hook",
-		},
-	)
+	deleteGracePeriod := int64(0)
```
phillebaba, line 131: We should explain why no grace period is given to the pod.

src/pkg/k8s/tunnel.go
```diff
@@ -254,19 +255,20 @@ func (tunnel *Tunnel) getAttachablePodForService(ctx context.Context) (string, e
...
-	selectorLabelsOfPods := MakeLabels(service.Spec.Selector)
-
-	servicePods := tunnel.kube.WaitForPodsAndContainers(
-		ctx,
-		PodLookup{
-			Namespace: tunnel.namespace,
-			Selector:  selectorLabelsOfPods,
-		},
-		nil,
-	)
```
phillebaba, line 266: Function is only used to list ready pods, probably not to wait for pods to run. Better instead to remove the wait logic.

src/internal/packager/helm/zarf.go
AustinAbro321: Likely since the WaitForPodsAndContainers function was the only one that returned pods it was used for convenience

AustinAbro321 (changes requested):

src/internal/packager/helm/zarf.go
AustinAbro321: I can't think of a reason, seems especially weird for the agent which could be in the middle of a request

src/pkg/cluster/injector.go
```diff
@@ -108,7 +108,13 @@ func (c *Cluster) StartInjectionMadness(ctx context.Context, tmpDir string, imag
 		spinner.Updatef("Attempting to bootstrap with the %s/%s", node, image)
 
 		// Make sure the pod is not there first
-		err = c.DeletePod(ctx, ZarfNamespaceName, "injector")
+		deleteGracePeriod := int64(0)
```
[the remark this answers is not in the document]
phillebaba: I will track this in some way. I think its fine to do as long as we are aware of why we are doing it.

src/internal/packager/helm/zarf.go
phillebaba: Most likely the reason. In that case we can just ignore this.

src/internal/packager/helm/zarf.go
phillebaba: I have added TODOs to both the delete grace periods, lets not change behavior here and instead resolve it in a separate PR.

AustinAbro321 (approved):

PR #2556, opened 2024-05-29
test: zarf init state

phillebaba:
## Description

This change adds tests to Zarf init state. It also fixes bugs that were found during testing. 

* Existing Zarf namespace would receive agent ignore label
* Existing Zarf namespace would not receive managed by label
* Clusters with zero nodes would cause panic

## Related Issue

Relates to #2550

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/state_test.go
+++ b/src/pkg/cluster/state_test.go
@@ -5,15 +5,178 @@
 package cluster
 
 import (
+	"context"
 	"fmt"
 	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
 
 	"github.com/defenseunicorns/pkg/helpers"
+
+	"github.com/defenseunicorns/zarf/src/pkg/k8s"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 	"github.com/defenseunicorns/zarf/src/pkg/pki"
 	"github.com/defenseunicorns/zarf/src/types"
-	"github.com/stretchr/testify/require"
 )
+
+func TestInitZarfState(t *testing.T) {
+	tests := []struct {
+		name        string
+		initOpts    types.ZarfInitOptions
+		nodes       []corev1.Node
+		namespaces  []corev1.Namespace
+		secrets     []corev1.Secret
+		expectedErr string
+	}{
+		{
+			name:        "no nodes in cluster",
+			expectedErr: "cannot init Zarf state in empty cluster",
+		},
+		{
+			name:     "no namespaces exist",
+			initOpts: types.ZarfInitOptions{},
+			nodes: []corev1.Node{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "node",
+					},
+				},
+			},
+		},
+		{
+			name: "namespaces exists",
+			nodes: []corev1.Node{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "node",
+					},
+				},
+			},
+			namespaces: []corev1.Namespace{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "kube-system",
+					},
+				},
+				{
+					ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


AustinAbro321 (changes requested):
Generally looks good, but there is a flake atm from the global spinner getting set to nil in parallel tests

src/pkg/cluster/state_test.go
```diff
@@ -5,16 +5,183 @@
...
+
+	"github.com/defenseunicorns/zarf/src/pkg/k8s"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 	"github.com/defenseunicorns/zarf/src/pkg/pki"
 	"github.com/defenseunicorns/zarf/src/types"
-	"github.com/stretchr/testify/require"
 )
 
+func TestZarfInitState(t *testing.T) {
+	t.Parallel()
```
lucasrod16, line 27: Echoing what @AustinAbro321 already stated about a race condition causing a nil panic due to the spinner not being goroutine safe

```bash
go test ./src/pkg/cluster -v -count=1 -failfast -race -run TestZarfInitState
```

data-race

It seems there may be other race condition bugs in our unit tests related to the spinner. To avoid this issue for now, I'm okay with _not_ running the test cases for `InitZarfState` in parallel. It would be helpful to leave a comment explaining why these tests cannot be run in parallel and to create a tech debt issue to address the spinner race condition. We should also consider adding the `-race` flag to our `go test` command in our `test-unit` make target to catch these problems in the future.

src/pkg/cluster/state_test.go
```diff
@@ -5,16 +5,183 @@
...
 	"github.com/defenseunicorns/pkg/helpers"
+
+	"github.com/defenseunicorns/zarf/src/pkg/k8s"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 	"github.com/defenseunicorns/zarf/src/pkg/pki"
 	"github.com/defenseunicorns/zarf/src/types"
-	"github.com/stretchr/testify/require"
 )
 
+func TestZarfInitState(t *testing.T) {
```
lucasrod16, line 26: To better reflect its purpose of testing the `InitZarfState` function, I recommend renaming this test to `TestInitZarfState`.

src/pkg/cluster/state_test.go
```diff
@@ -5,16 +5,183 @@
...
+					if err != nil {
+						continue
+					}
+					sa := &corev1.ServiceAccount{
+						ObjectMeta: metav1.ObjectMeta{
+							Namespace: ns.Name,
+							Name:      "default",
+						},
+					}
+					cs.CoreV1().ServiceAccounts(ns.Name).Create(ctx, sa, metav1.CreateOptions{})
```
lucasrod16, line 157: We should ensure that we handle the error returned from this:

```go
_, err = cs.CoreV1().ServiceAccounts(ns.Name).Create(ctx, sa, metav1.CreateOptions{})
if err != nil {
    t.Errorf("err msg: %v", err)
}
```

note: the linter on my local machine didn't seem happy about using `require.NoError` or `t.Failf` in a goroutine that is not the main test function goroutine

require-lint


tfatal-lint

lucasrod16 (changes requested):
Looking good so far! I have a few comments, but overall it's coming along well

src/pkg/cluster/state_test.go
```diff
@@ -5,16 +5,183 @@
...
+		name        string
+		initOpts    types.ZarfInitOptions
+		nodes       []corev1.Node
+		namespaces  []corev1.Namespace
+		secrets     []corev1.Secret
+		expectedErr string
+	}{
+		{
+			name:        "no nodes in cluster",
+			expectedErr: "cannot init Zarf state in empty cluster",
```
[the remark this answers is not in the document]
phillebaba: I do agree but lets do it during a larger error refactor.

src/pkg/cluster/state_test.go
phillebaba: t cannot be used in a go routine, we would need to come up with some other solution for this in that case.

phillebaba:
I have removed the parallel instruction from the test to move the PR along. We will have to solve the race conditions with the Spinner in a later PR.

AustinAbro321 (approved):

lucasrod16 (approved):
lgtm

Issue #2562, opened 2024-05-30
test: reduce duration of E2E tests


lucasrod16:
### Describe what should be investigated or refactored
Our E2E tests currently take about 20 minutes to run. We should aim to speed up this feedback loop by removing low-value tests and parallelizing where possible.

lucasrod16:
- [ ] Remove `TestMismatchedArchitectures`
    - `TestValidatePackageArchitecture` unit test covers the same scenarios and more

PR #2564, opened 2024-05-30
test: remove TestMismatchedVersions e2e test

lucasrod16:
## Description
Removes `TestMismatchedVersions` and adds `TestPrintBreakingChanges` unit test

## Related Issue

Relates to #2562

## Checklist before merging


lucasrod16:
Sharing screenshots to compare the output of `PrintBreakingChanges` before and after the refactoring in this PR

Before:

before


After:

after

phillebaba:
Might be out of scope for this but I would much rather have PrintBreakingChanges return a string and then have the calling function output it to the correct writer. Makes things cleaner and easier to test. Other than that this looks good.

AustinAbro321 (approved):
Agree with Phillip, that it should be refactored but it's fine to merge now. Just make an issue for the refactoring if you merge now

lucasrod16:
> Might be out of scope for this but I would much rather have PrintBreakingChanges return a string and then have the calling function output it to the correct writer. Makes things cleaner and easier to test. Other than that this looks good.

@phillebaba @AustinAbro321  https://github.com/defenseunicorns/zarf/issues/2570

PR #2566, opened 2024-05-31
refactor: move k8s tunnel to cluster package

phillebaba:
## Description

Moves the tunnel code to cluster pacakge.

This is far from a perfect solution but I do not want to make a enormous change. The tunnel code needs some refactoring in the future but before that can be done we need to understand how dependencies on Cluster looks. 

## Related Issue

Relates to #2507 

## Checklist before merging


phillebaba:
Tests are failing because they depend on the old tunnel logic. The old logic would list pods according to a service selector, order them by creation timestamp and wait for them to be ready. The issue is that when the registry credentials are updated the command does not block until the registry has restarted. So the tunnel is created to the old registry pod.

I have no idea if this was the intended design or just a fluke which allowed the test to pass. I will solve this issue by making sure that update credentials blocks until all pods have restarted instead.

phillebaba:
Blocked until https://github.com/defenseunicorns/pkg/pull/98 is merged.

src/pkg/cluster/tunnel.go
```diff
@@ -273,3 +280,255 @@ func serviceInfoFromNodePortURL(services []corev1.Service, nodePortURL string) (
...
+		return tunnel.resourceName, nil
+	case SvcResource:
+		return tunnel.getAttachablePodForService(ctx)
+	default:
+		return "", fmt.Errorf("unknown resource type: %s", tunnel.resourceType)
+	}
+}
+
+// getAttachablePodForService will find an active pod associated with the Service and return the pod name.
+func (tunnel *Tunnel) getAttachablePodForService(ctx context.Context) (string, error) {
```
AustinAbro321, line 513: Taking the wait away from this function is interesting. I like that it's faster feedback to the user if the pod isn't there, however I would like to know the behavior if we return an err'd pod or a pod that isn't ready. I'm guessing the port forward command would just error?

phillebaba: It would not return a failing Pod because we filter pods based on status phase. This behavior is a lot more inline with kubectl portforward. While it is nice to have some sort of waiting functionality in the tunnel I do not think it is the right place for it.

AustinAbro321: Makes sense, agree being similar to kubectl port forward is better

AustinAbro321 (approved):

Issue #2520, opened 2024-05-17
Handling SIGINT and sequential deployments


andrewg-xyz:
### Environment
Device and OS: M2 Max, OSX Sonoma 14.4.1
App version: v0.33.0
Kubernetes distro being used: k3d (uds deploy k3d-core-slim-dev:0.21.1)
Other:

### Steps to reproduce
1. `zarf package deploy ...` My app has an error that fails deployment "0 out of 1 expected pods are ready"
2. SIGINT (ctrl+c) to stop the `zarf package deploy`
3. (make updates, rebuild zarf package) and run `zarf package deploy`

### Expected result
The interruption is handled and future `zarf package deploy` function.

### Actual Result
My cluster has a helm CR with a status of `pending-install` Which prevents sequential `zarf package deploy`-ments of my package (once I've fixed the error)

Sequential deploys fail/retry with `WARNING  Retrying (1/3) in 5s: unable to complete the helm chart install/upgrade: another operation                                                               (install/upgrade/rollback) is in progress`  

### Visual Proof (screenshots, videos, text, etc)
![image](https://github.com/defenseunicorns/zarf/assets/46836330/c9563610-1259-47a8-96fa-bce917d6b1ce)
![image](https://github.com/defenseunicorns/zarf/assets/46836330/64b92feb-8d00-483c-9bc1-b8e0103aefd6)
![image](https://github.com/defenseunicorns/zarf/assets/46836330/07c54d1c-9fb8-4f82-aaf9-1944a8502921)

### Additional Context
workaround is to manually remove the helm CR and clean up the failed deployment.

phillebaba:
This should be covered by #2505. It is good however to have a specific example of why this is needed.

lucasrod16:
@andrewg-xyz I am not sure there is a way Zarf can reliably guarantee cleanup of resources when a deployment is canceled with a SIGINT. Zarf uses the Helm SDK to cleanup resources on a best effort basis for failed upgrades and rollbacks, but it is risky for Zarf to start removing resources from the cluster outside of the Helm SDK on a failed deployment in my opinion.

I would recommend running `zarf package remove` to cleanup the failed deployment before attempting re-deployment

andrewg-xyz:
Thanks - I'm good to close with the context. I'll keep in mind as I interact with similar situations.

PR #2567, opened 2024-05-31
fix: cancel Cobra parent context on interrupt

### LINKED ISSUE #2505: Add signal handling to context

Context usage is being introduced in #2405 which is great. When this is done it is going to be possible to cancel contexts also. In the Zarf CLI currently if a user presses Ctrl+C the process will be killed immediately without having the opportunity to cancel the context. Which leaves in flight requests abandoned.

- **Given** A running CLI process
- **When** An interrupt signal occurs
- **Then** All contexts should be cancelled

I am not sure if this is a thing people do when building CLIs using Cobra. The alternative solution is to not handle any interrupt signals at all.

### Additional context
N/A

phillebaba:
## Description

Right now different functions handle interrupts in different ways. Some will register their own signal listeners while others will use the global signal handler that exits the program. This means that contexts are never cancelled giving the process time to shut down and clean up. This change adds a signal handler to the Cobra parent context and removes the other signal handlers. Now all commands will use the same signal handler.

## Related Issue

Fixes #2505 
Relates to #2520 

## Checklist before merging


lucasrod16 (approved):
lgtm

src/internal/agent/start.go
```diff
@@ -27,35 +28,40 @@ const (
 )
 
 // StartWebhook launches the Zarf agent mutating webhook in the cluster.
-func StartWebhook() {
+func StartWebhook(ctx context.Context) error {
 	message.Debug("agent.StartWebhook()")
-
-	startServer(agentHttp.NewAdmissionServer(httpPort))
+	return startServer(ctx, agentHttp.NewAdmissionServer(httpPort))
```
AustinAbro321, line 33: This context should be used in the agent, but this is fine to merge so I'll just spin up another PR to do that

AustinAbro321 (approved):

PR #2568, opened 2024-05-31
test: delete agent e2e label test

AustinAbro321:
## Description

I believe this test is no longer necessary since we've added unit tests to the agent 

Relates to #2562

## Checklist before merging


phillebaba:
@AustinAbro321 could you link to the unit tests which now covers this e2e test?

AustinAbro321:
Yes @phillebaba. In these locations we test that pods with no labels will not error and pods with labels will bring not lose those labels
- https://github.com/defenseunicorns/zarf/blob/main/src/internal/agent/hooks/pods_test.go#L105-L109
- https://github.com/defenseunicorns/zarf/blob/main/src/internal/agent/hooks/pods_test.go#L46-L50

phillebaba (dismissed):

lucasrod16 (approved):

PR #2569, opened 2024-05-31
test: validate package

AustinAbro321:
## Description

This PR adds tests to validate package

## Checklist before merging


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/defenseunicorns/zarf
 
-go 1.21.8
```
phillebaba, line 3: Bump Go version in a separate PR.

src/types/validate_test.go
```diff
@@ -0,0 +1,494 @@
...
+	"github.com/defenseunicorns/zarf/src/config/lang"
+	"github.com/defenseunicorns/zarf/src/pkg/variables"
+	"github.com/stretchr/testify/require"
+)
+
+func TestZarfPackageValidate(t *testing.T) {
+	tests := []struct {
+		name     string
+		pkg      ZarfPackage
+		wantErrs []string
```
phillebaba, line 23: Rename this to expectedErrs

phillebaba (changes requested):

src/types/validate_test.go
lucasrod16, line 1: We should be able to run all of these test functions and test cases in parallel with `t.Parallel()`

lucasrod16 (changes requested):

src/types/validate_test.go
lucasrod16, line 1: Could you also ensure that the top-level test functions are run in parallel?

lucasrod16 (changes requested):

lucasrod16 (dismissed):
lgtm

src/types/validate.go
```diff
@@ -40,27 +40,28 @@ func SupportedOS() []string {
 
 // Validate runs all validation checks on the package.
 func (pkg ZarfPackage) Validate() error {
+	errs := []error{}
 	if pkg.Kind == ZarfInitConfig && pkg.Metadata.YOLO {
-		return fmt.Errorf(lang.PkgValidateErrInitNoYOLO)
+		errs = append(errs, fmt.Errorf(lang.PkgValidateErrInitNoYOLO))
```
schristoff-du, line 45: Could we use `errors.Join` as an append instead of making a `errs := []error{}`
I would say either do a `var err error` and then `errors.Join(error, fmtErrorf(...)`

zarf.schema.json
```diff
@@ -1003,7 +1003,7 @@
       "properties": {
         "name": {
           "type": "string",
-          "pattern": "^[a-z0-9\\-]*[a-z0-9]$",
+          "pattern": "^[a-z0-9][a-z0-9\\-]*$",
```
schristoff-du, line 1006: What drove this change?

src/types/validate_test.go
```diff
@@ -0,0 +1,506 @@
...
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			err := tt.component.Validate()
+			if tt.expectedErrs == nil {
```
schristoff-du, line 498: I nitted Wayne about it may be better to make expectedErrs a bool, but it's up to you.

schristoff-du (commented):
Couple quick questoins

src/types/validate_test.go
phillebaba: I think this is better to avoid situations where expectedErrs is not empty but the bool is false.

zarf.schema.json
AustinAbro321: It's the wrong regex, we currently do a check here https://github.com/defenseunicorns/zarf/blob/main/src/types/validate.go#L26 at create time. I'm going to refactor to use the schema instead of a separate test, but wanted to get the schema in line and create tests first

schristoff: Dope :)

schristoff-du (approved):

lucasrod16 (approved):

phillebaba (approved):

PR #2573, opened 2024-05-31
test: remove mismatched arch checks from init test

lucasrod16:
## Related Issue

Relates to #2562


AustinAbro321:
@lucasrod16 Could you link the unit tests we have covering this flow?

lucasrod16:
> @lucasrod16 Could you link the unit tests we have covering this flow?

https://github.com/defenseunicorns/zarf/blob/0c02643b9f5631a7bdd98650cc87c74901190771/src/pkg/packager/common_test.go#L22

lucasrod16:
There is value in keeping this logic in the init test, so I am going to keep it as is and close this PR. The part of the test this PR deletes covers arch validation when creating a k3s cluster, which is not covered by any other tests at the moment since that is handled in an onDeploy action.

At first glance, I assumed this was duplicate test logic that `TestValidatePackageArchitecture` covers, but that is not the case.

PR #2574, opened 2024-05-31
feat: add webhooks for crossplane resources to zarf-agent

### LINKED ISSUE #2572: Add support for Crossplane CRDs
Crossplane reads some image references in its Provider, Function, and Configuration manifests before creating a Deployment.
Since Zarf-Agent does not yet cover these CRDs, crossplane attempts to pull these Images from the Internet.

- **Given** a Crossplane Provider, Configuration, or Function resource
- **When** creating or updating
- **Then** mutate spec.package and add spec.packagePullSecret

An alternative would be to support arbitrary CRDs but this would be far more complex to implement.

### Additional context
This is a typical Provider resource (Configurations, and Functions look similar):
```yaml
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: provider-kubernetes
spec:
  package: xpkg.upbound.io/crossplane-contrib/provider-kubernetes:v0.13.0
```
The Crossplane Pod will pull the image referenced in spec.package read some metadata, and create a Deployment for it.

I have already implemented a working prototype and if there is interest from your side I would be happy to finalize it with your feedback.

Avarei:
## Description

This Implements the Crossplane resources
* Configuration
* Provider
* Function
to rewrite the spec.package and spec.packagePullSecrets field in a similar way as for Pod.

This allows the Crossplane Pod to pull the images from the local registry.

Well... not quite yet. This also implements an extension of the private-registry secret, which is created for namespaces managed by zarf. additionally to the localhost reference to the registry this also adds the same credentials for the route over the kubernetes service.

This implementation rewrites
```yaml
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  name: provider-terraform
spec:
  package: xpkg.upbound.io/upbound/provider-terraform:v0.16.0
```
into
```yaml
apiVersion: pkg.crossplane.io/v1
kind: Provider
metadata:
  # Removed for brevity
  labels:
    zarf-agent: patched
  name: provider-terraform
spec:
  ignoreCrossplaneConstraints: false
  package: zarf-docker-registry.zarf.svc.cluster.local:5000/upbound/provider-terraform:v0.16.0-zarf-255403924 # patched by zarf-agent
  packagePullPolicy: IfNotPresent
  packagePullSecrets:
  - name: private-registry # added by zarf-agent
  revisionActivationPolicy: Automatic
  revisionHistoryLimit: 1
  runtimeConfigRef:
    apiVersion: pkg.crossplane.io/v1beta1
    kind: DeploymentRuntimeConfig
    name: default
  skipDependencyResolution: false
```

which will by default create a pod like this:
```yaml
apiVersion: v1
kind: Pod
metadata:
  # Removed for brevity
spec:
  containers:
    - image: 127.0.0.1:31999/upbound/provider-terraform:v0.16.0-zarf-255403924-zarf-928246402
       name: package-runtime
  # Removed for brevity
```
The resulting Pod is a bit of a problem. Since when
https://github.com/defenseunicorns/zarf/blob/0c02643b9f5631a7bdd98650cc87c74901190771/src/internal/agent/hooks/pods.go#L76
is called the function does just replace the hostname.
I could not yet find a nice way to implement this, as the const i would need is in
https://github.com/Avarei/zarf/blob/13a7957b46a0dcacb9baa63d7826f42ea44915d7/src/types/k8s.go#L51
and the package surrounding
https://github.com/defenseunicorns/zarf/blob/0c02643b9f5631a7bdd98650cc87c74901190771/src/pkg/transform/image.go#L27
seems to avoid this kind of external dependency. and i would like to avoid negatively impacting your code quality.

Currently I sidestep this issue by creating a DeploymentRuntimeConfig in Crossplane which lets me overwrite the used image back to the internet version, which will be replaced back to the localhost version when the Pod is created.

## Related Issue

Fixes  #2572

## Checklist before merging


Avarei:
I reconsidered and implemented the additional logic to the function in the hope that it makes review easier, than thinking about hypotheticals.
The resulting pod now looks like it should. I also added a test to assert this.

lucasrod16:
I am closing this PR because the Zarf team has decided to not support mutating Crossplane resources at this time

https://github.com/defenseunicorns/zarf/pull/2628#issuecomment-2186990599

PR #2575, opened 2024-06-03
fix: add custom error printing for Zarf commands

phillebaba:
## Description

This change adds some custom logic for printing returned errors from Cobra commands. This will allow embedded tools like Helm to behave the same as the original tools. While the Zarf commands will print errors in the same expected error format.

## Related Issue

Relates to #2567

## Checklist before merging

### DIFF
--- a/src/cmd/root.go
+++ b/src/cmd/root.go
@@ -17,6 +17,7 @@ import (
 	"github.com/defenseunicorns/zarf/src/pkg/layout"
 	"github.com/defenseunicorns/zarf/src/pkg/message"
 	"github.com/defenseunicorns/zarf/src/types"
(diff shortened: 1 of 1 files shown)


AustinAbro321:
This doesn't affect Zarf commands because we already either `message.Fatal` or return no err correct?

phillebaba:
Yep none of the commands today return an error. Something that we want to change in the future. An issue currently is that message fatal takes both a message and an error, but only outputs the message to stderr. It is going to be a breaking change if we want to wrap everything into a single error.

AustinAbro321 (approved):

lucasrod16 (approved):

PR #2578, opened 2024-06-03
refactor: use root ctx in agent

AustinAbro321:
## Description
Updates the agent to use ctx from the root

## Checklist before merging


schristoff-du (approved):

schristoff-du:
It's smol, I'm happy with it if e2e is as well

PR #2580, opened 2024-06-04
fix: add wait for deployments when updating credentials

phillebaba:
## Description

This change waits for deployments to update using kstatus when updating credentials. The reason this is necessary is to make sure that all pods have received the updated configuration before the update command returns. Otherwise it would be possible to assume that the configuration has been applied when it has not. Sadly Helm is not able to do this itself as it will not wait for old pods to delete before returning. 

## Related Issue

Relates to #2566 

## Checklist before merging


phillebaba:
I realize that this change does some duplicate work of #2497 while not introducing new features that require a bit more consensus. It would be for the best to come up with a standard method of using kstatus in the PR as it is also required in other projects like https://github.com/defenseunicorns/lula/issues/229.

Issue #2485, opened 2024-05-08
Run `zarf dev lint` on create and deploy


AustinAbro321:
### Is your feature request related to a problem? Please describe.
Packages should be linted on create and on deploy. If there is an error on create it should stop the package from being built, warning should be printed to the user. If there is an error on deploy the package should still deploy, but it should be printed to the user

### Describe the solution you'd like

- **Given** I have a zarf.yaml with schema errors
- **When** I run zarf package create
- **Then** my package create fails and I get the lint output
---
- **Given** I have a package with schema errors
- **When** I run zarf package deploy
- **Then** my package deploy succeeds (or tries it's best) and I get the lint output

PR #2585, opened 2024-06-05
feat: run schema validation on create

AustinAbro321:
## Description

This removes the duplicated checks between the schema and validate, and runs error linting on create. This introduces a new type `packageError` to track errors and warnings through packager. This type replaces `packager.warnings`, so this PR touches many files, however besides lint and create behavior is the same. 

## Related Issue

Relates to #2485 

## Checklist before merging


AustinAbro321:
This will change how the errors are presented to users to look like what happens during `zarf dev lint` rather than a normal error message
image

src/pkg/packager/composer/list.go
```diff
@@ -257,19 +257,30 @@ func (ic *ImportChain) String() string {
...
+				YqPath:              fmt.Sprintf("components.[%d]", node.index),
+				Category:            types.SevWarn,
+			})
+		}
 		node = node.next
 	}
 	if len(warnings) > 0 {
-		final := fmt.Sprintf("Migrations were performed on the import chain of: %q", ic.head.Name)
-		warnings = append(warnings, final)
+		final := types.PackageError{
```
schristoff-du, line 277: a more descriptive variable name would be nice

src/pkg/packager/creator/creator_test.go
```diff
@@ -0,0 +1,91 @@
...
+			creator:     NewSkeletonCreator(types.ZarfCreateOptions{}, types.ZarfPublishOptions{}),
+		},
+		{
+			name:        "invalid package definition",
+			testDir:     "invalid",
+			expectedErr: "errors during lint",
+			creator:     NewSkeletonCreator(types.ZarfCreateOptions{}, types.ZarfPublishOptions{}),
+		},
+	}
+	b, err := os.ReadFile("../../../../zarf.schema.json")
```
schristoff-du, line 61: I would suggest a changeDir, but then we would have to move back. 🤔  We could copy? 
This isn't a big deal, but it does look ugly, and it we move this file it'll cause problems. (low priority nit)

src/pkg/packager/creator/creator_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+type mockSchemaLoader struct {
+	b []byte
+}
+
+func (m *mockSchemaLoader) ReadFile(_ string) ([]byte, error) {
+	return m.b, nil
+}
+
+func TestLoadPackageDefinition(t *testing.T) {
```
schristoff-du, line 27: `func TestLoadPackageDefinitionWithValidate`

src/pkg/packager/creator/normal.go
```diff
@@ -61,7 +61,8 @@ func NewPackageCreator(createOpts types.ZarfCreateOptions, cwd string) *PackageC
 }
 
 // LoadPackageDefinition loads and configures a zarf.yaml file during package create.
-func (pc *PackageCreator) LoadPackageDefinition(ctx context.Context, src *layout.PackagePaths) (pkg types.ZarfPackage, warnings []string, err error) {
+// Use LoadPackageDefinitionWithValidate unless there is a specific reason to skip validation or run validation separately
+func (pc *PackageCreator) LoadPackageDefinition(ctx context.Context, src *layout.PackagePaths) (pkg types.ZarfPackage, warnings []types.PackageError, err error) {
```
schristoff-du, line 65: Should LoadPackageDefinition now just call LoadPackageDefinitionWithValidate, or where (should?) this still be used

src/pkg/packager/creator/utils.go
```diff
@@ -5,15 +5,49 @@
...
+	"github.com/defenseunicorns/zarf/src/pkg/packager/lint"
 	"github.com/defenseunicorns/zarf/src/types"
 )
 
+func validate(createOpts types.ZarfCreateOptions, pkg types.ZarfPackage) error {
+	if err := pkg.Validate(); err != nil {
+		return fmt.Errorf("package validation failed: %w", err)
+	}
+
+	findings, err := lint.Validate(context.TODO(), createOpts)
```
schristoff-du, line 26: context.TODO() -> context.Background

src/pkg/packager/creator/utils.go
```diff
@@ -5,15 +5,49 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to lint package: %w", err)
+	}
+
+	lint.PrintFindings(findings, types.SevErr, createOpts.BaseDir, pkg.Metadata.Name)
+	if lint.HasErrors(findings) {
```
schristoff-du, line 32: unsure of this is needed if print findings prints (also, should be severr or sevwarn?)

src/pkg/packager/dev.go
```diff
@@ -105,3 +107,46 @@ func (p *Packager) DevDeploy(ctx context.Context) error {
...
+		return err
+	}
+
+	lintFindings, err := lint.Validate(ctx, p.cfg.CreateOpts)
+	if err != nil {
+		return fmt.Errorf("linting failed: %w", err)
+	}
+	p.warnings = append(p.warnings, lintFindings...)
+
+	if len(p.warnings) == 0 {
```
schristoff-du, line 140: 🤔  Why print this - I think it could be left out

src/pkg/packager/dev.go
```diff
@@ -105,3 +107,46 @@ func (p *Packager) DevDeploy(ctx context.Context) error {
...
+		message.Successf("0 findings for %q", p.cfg.Pkg.Metadata.Name)
+		return nil
+	}
+
+	lint.PrintFindings(p.warnings, types.SevWarn, p.cfg.CreateOpts.BaseDir, p.cfg.Pkg.Metadata.Name)
+
+	if lint.HasErrors(p.warnings) {
```
schristoff-du, line 147: Same comments as above

src/pkg/packager/lint/lint.go
```diff
@@ -6,154 +6,96 @@ package lint
...
 )
 
-// ZarfSchema is exported so main.go can embed the schema file
-var ZarfSchema embed.FS
-
-func getSchemaFile() ([]byte, error) {
-	return ZarfSchema.ReadFile("zarf.schema.json")
+// FileLoader is an interface for reading files, it decouples the lint package from the go embed package
+type FileLoader interface {
+	ReadFile(path string) ([]byte, error)
```
schristoff-du, line 25: 🤔  Do we ever have to worry about multiple reads on the files? (i.e. do we need a locking mechanism?)

schristoff-du:
I'll pass through this again in the morning, but lots of questions ✨

zarf.schema.json
```diff
@@ -382,8 +382,7 @@
       "additionalProperties": false,
       "type": "object",
       "required": [
-        "name",
-        "namespace"
```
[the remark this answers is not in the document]
AustinAbro321: While namespace is required for charts, since having the namespace set through an imported component is possible namespace isn't truly required in the schema if you have import. There is no way for jsonschema to check values across objects (charts cannot check values on the component object) so we do this check with runtime validation instead, which happens after compose.

src/pkg/packager/creator/creator_test.go
AustinAbro321: Yeah I guess no matter what we still have to reference the file, kinda a pick your poison situation. This seems simplest.

src/pkg/packager/creator/normal.go
AustinAbro321: LoadPackageDefinition should be used where we don't want to validate on a package yet, because we are going to change more in it, such as it's use in `zarf dev deploy` or when we want to do additional validations on the package even if it's invalid, such as in `zarf dev lint`. 

In the regular case where a caller is just doing a package create, they can use LoadPackageDefinitionWithValidate.

src/pkg/packager/creator/utils.go
AustinAbro321: I've kept the print as severr intentionally for now, as many of the warnings from lint can be pretty noisy. In the future, I'd like to change some of the lint results to "notes" then print out the lint warnings. 

The table will be printed out, but I want to return an error still that will cause the program to fail if a validate schema is not used. This is what it will look like when someone has a schema error. 

image

src/pkg/packager/lint/lint.go
AustinAbro321: We don't have to worry about multiple reads as lint is not called in a goroutine, however unsynchronized simultaneous reads would be safe anyway

src/types/package.go
```diff
@@ -72,3 +74,29 @@ type ZarfBuildData struct {
 	LastNonBreakingVersion     string            `json:"lastNonBreakingVersion,omitempty" jsonschema:"description=The minimum version of Zarf that does not have breaking package structure changes"`
 	Flavor                     string            `json:"flavor,omitempty" jsonschema:"description=The flavor of Zarf used to build this package"`
 }
+
+// PackageError is a struct that contains a finding about something wrong with a package
+type PackageError struct {
+	// YqPath is the path to the key where the error originated from, this is sometimes empty in the case of a general error
+	YqPath      string
```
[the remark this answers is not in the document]
AustinAbro321: Wanted to differentiate between the packagePath where the error occurred. for example "packages/gitea" vs the yqPath of the error in yaml ".metadata.name"

src/types/package.go
```diff
@@ -21,7 +21,7 @@ type ZarfPackage struct {
 	Kind       ZarfPackageKind                 `json:"kind" jsonschema:"description=The kind of Zarf package,enum=ZarfInitConfig,enum=ZarfPackageConfig,default=ZarfPackageConfig"`
 	Metadata   ZarfMetadata                    `json:"metadata,omitempty" jsonschema:"description=Package metadata"`
 	Build      ZarfBuildData                   `json:"build,omitempty" jsonschema:"description=Zarf-generated package build data"`
-	Components []ZarfComponent                 `json:"components" jsonschema:"description=List of components to deploy in this package"`
+	Components []ZarfComponent                 `json:"components" jsonschema:"description=List of components to deploy in this package,minItems=1"`
```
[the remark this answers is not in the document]
AustinAbro321: Yeah it's very nice. There's more we can add to the schema as well, but wanted to keep it simple in this PR. I think we can also do it in code rather than in json comments which will probably be easier to read for anything complex.

src/pkg/packager/dev.go
AustinAbro321: It just feels like we should give feedback to the user so they know the lint actually happened. Screenshot of without vs with. 
image

src/pkg/packager/dev.go
AustinAbro321: Yeah I don't love this, but I still want to give an error to cobra so that it exits with 1. Here is it with the error message vs without. I'm fine with either way. 
image

src/pkg/packager/creator/utils.go
AustinAbro321: As I think about "errors in package" is probably better for the create case

src/cmd/package.go
```diff
@@ -344,6 +344,11 @@ func bindCreateFlags(v *viper.Viper) {
 	createFlags.BoolVarP(&pkgConfig.CreateOpts.ViewSBOM, "sbom", "s", v.GetBool(common.VPkgCreateSbom), lang.CmdPackageCreateFlagSbom)
 	createFlags.StringVar(&pkgConfig.CreateOpts.SBOMOutputDir, "sbom-out", v.GetString(common.VPkgCreateSbomOutput), lang.CmdPackageCreateFlagSbomOut)
 	createFlags.BoolVar(&pkgConfig.CreateOpts.SkipSBOM, "skip-sbom", v.GetBool(common.VPkgCreateSkipSbom), lang.CmdPackageCreateFlagSkipSbom)
+	validateSchema := true
+	if v.Get(common.VPkgCreateValidateSchema) != nil {
+		validateSchema = v.GetBool(common.VPkgCreateValidateSchema)
+	}
+	createFlags.BoolVar(&pkgConfig.CreateOpts.ValidateSchema, "validate-schema", validateSchema, lang.CmdPackageCreateFlagValidateSchema)
```
AustinAbro321, line 351: Went back and forth on whether I should add this flag, want to hear teams thoughts. In theory we have yaml extension covered so we shouldn't need an escape hatch. Not sure if we want this just in case tho.

src/pkg/packager/creator/creator.go
```diff
@@ -13,7 +13,9 @@ import (
 
 // Creator is an interface for creating Zarf packages.
 type Creator interface {
-	LoadPackageDefinition(ctx context.Context, src *layout.PackagePaths) (pkg types.ZarfPackage, warnings []string, err error)
+	LoadPackageDefinition(ctx context.Context, src *layout.PackagePaths) (pkg types.ZarfPackage, warnings []types.PackageError, err error)
+	LoadPackageDefinitionWithValidate(ctx context.Context, src *layout.PackagePaths) (pkg types.ZarfPackage, warnings []types.PackageError, err error)
```
lucasrod16, line 17: I don't think we should add any methods to the Creator interface for schema validation. It clutters the purpose of the interface. Schema validation is an implementation detail that should be hidden from the interface's API. We could probably use the functional options pattern to allow configuring strict schema validation, but we should not add specific validation methods to the interface.

lucasrod16 (changes requested):
Can you split this PR into multiple PRs so that it's easier to review? There is a good amount of refactoring and changes to keep track of in this PR that would be easier to review in smaller chunks

src/pkg/packager/creator/creator.go
AustinAbro321: IMO it makes sense to expose Validate because a caller may want to make changes to the package before assembling then validate the package as we do in `dev deploy` or they may want to load a package even if it's incorrect so additional data can be extracted from the package as we do in `dev lint`.

What do you think of instead having a function `LoadPackageDefinitionWithOutValidate` so it's more clear to the caller that `LoadPackageDefinition` is the default? After talking with @phillebaba I'm planning to remove the option to not validate the schema. Even if we did allow users to not validate their schema I think we should still keep the other validations (`pkg.Validate()`) we already have in place so I'd rather not do it through a create option.

lucasrod16: I agree that we should make schema validation the default behavior in `LoadPackageDefinition`. We can add a boolean field in `CreateOptions` to allow callers to disable it for cases where strict validation isn't wanted or needed. Since we already use `pkg.Validate` in `LoadPackageDefinition`, we just need to add a call to `lint.Validate` without changing the interface.

AustinAbro321: I think there are two questions here. 
1. Should we add the `Validate()` to the creator interface
2. Should we add the `LoadPackageDefinitionWithOutValidate()` to the interface. 

IMO we should add `Validate()` to the interface. It will be useful to have a common exposed way to ensure the package is valid if changes are made after the package is loaded. We do this already in `dev deploy`, and it'd be reasonable for a library user to want to do something similar. 

As far as not adding the `LoadPackageDefinitionWithOutValidate()` and instead using a createOpt, I think I'm fine with that. Though I'm unsure if it should be passed to the user as that would give them more options for bad packages than they have currently.

lucasrod16: My answer is no to both questions. I do not think we need to make any changes to the Creator interface to run schema validation during package create.

I am suggesting a more simple approach by adding a single boolean field to CreateOptions and a single `lint.Validate` function call right after `pkg.Validate` is currently called in `LoadPackageDefinition`. The validate logic would always get called by default, unless the CreateOptions boolean field for disabling it is set to true. It achieves the same goals and behavior with less code and less abstractions/complexity. Callers are already required to pass a `ZarfCreateOptions` struct to instantiate a Creator object, so it would be very simple to simple set a boolean field to true in that sruct to disable schema validation if desired, rather than calling a separate `Validate` method at the right time.

AustinAbro321:
Broke this out into #2658 and #2660

src/pkg/lint/findings.go
```diff
@@ -0,0 +1,115 @@
...
+
+func filterLowerSeverity(findings []PackageFinding, severity Severity) []PackageFinding {
+	findings = helpers.RemoveMatches(findings, func(finding PackageFinding) bool {
+		return finding.Severity > severity
+	})
+	return findings
+}
+
+// PrintFindings prints the findings of the given severity in a table
+func PrintFindings(findings []PackageFinding, severity Severity, baseDir string, packageName string) {
```
phillebaba, line 65: I know that it is tricky but this would be nice to defer to the cli calling. If it is very difficult to do right now you can ignore this.

src/pkg/lint/findings_test.go
```diff
@@ -0,0 +1,107 @@
...
+			},
+			packageName: "testPackage",
+			want: map[string][]PackageFinding{
+				"path": {{Severity: SevWarn, PackageNameOverride: "import", PackagePathOverride: "path"}},
+				".":    {{Severity: SevErr, PackageNameOverride: "testPackage", PackagePathOverride: "."}},
+			},
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
```
phillebaba, line 50: This is no longer needed in Go 1.22

src/pkg/lint/findings_test.go
```diff
@@ -0,0 +1,107 @@
...
+				{
+					Severity: SevErr,
+				},
+			},
+			severity: SevErr,
+			expected: true,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
```
phillebaba, line 101: Ditto

src/pkg/lint/lint.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	PrintFindings(findings, SevWarn, createOpts.BaseDir, pkg.Metadata.Name)
+
+	if HasSevOrHigher(findings, SevErr) {
+		return errors.New("errors during lint")
+	}
+
+	return nil
+}
```
phillebaba, line 58: NIT but I would remove the white space in between lines.

src/pkg/lint/findings.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+func (f PackageFinding) itemizedDescription() string {
+	if f.Item == "" {
+		return f.Description
+	}
+	return fmt.Sprintf("%s - %s", f.Description, f.Item)
+}
+
+func colorWrapSev(s Severity) string {
```
phillebaba, line 48: We could change this rater but UI stuff should not be here.

phillebaba (commented):
Some small comments.

src/pkg/packager/creator/creator_test.go
```diff
@@ -6,16 +6,33 @@ package creator
...
+func (m *mockSchemaLoader) ReadFile(_ string) ([]byte, error) {
+	return m.b, nil
+}
+
+// Satisfy fs.ReadFileFS interface
+func (m *mockSchemaLoader) Open(_ string) (fs.File, error) {
+	return nil, nil
+}
+
+func TestLoadPackageDefinitionWithValidate(t *testing.T) {
```
lucasrod16, line 33: Can we rename this test back to `TestLoadPackageDefinition` for clarity?

src/pkg/packager/creator/utils.go
```diff
@@ -5,15 +5,36 @@
...
 
 	"github.com/defenseunicorns/zarf/src/config"
+	"github.com/defenseunicorns/zarf/src/pkg/lint"
 	"github.com/defenseunicorns/zarf/src/pkg/packager/deprecated"
 	"github.com/defenseunicorns/zarf/src/types"
 )
 
+// Validate errors if a package violates the schema or any runtime validations
+// This must be run while in the parent directory of the zarf.yaml being validated
+func Validate(pkg types.ZarfPackage) ([]lint.PackageFinding, error) {
```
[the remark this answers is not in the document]
AustinAbro321: No this is create time validation. Deploy will have to allow for breakages in the schema as there could be fields that previously existed which were removed or new fields added when deploying with a different version of Zarf.

src/pkg/packager/creator/creator_test.go
```diff
@@ -43,17 +60,28 @@ func TestLoadPackageDefinition(t *testing.T) {
...
+			defer func() {
+				err = os.Chdir(cwd)
+				require.NoError(t, err)
+			}()
+			path := filepath.Join("testdata", tt.testDir)
+			err = os.Chdir(path)
+			require.NoError(t, err)
 
-			src := layout.New(filepath.Join("testdata", tt.testDir))
+			src := layout.New(".")
```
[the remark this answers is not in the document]
AustinAbro321: Since `ValidateSchema` is doing a read on the `zarf.yaml` we need to to be in the correct directory. The Creator interface already has that assumption, otherwise compose won't work, but we aren't composing in this test. #2683 will change this hopefully, but it'll be pretty involved.

src/pkg/lint/findings.go
AustinAbro321: Yeah it's difficult to do right now as we'd have to change create & dev to return findings to the cli so it could then be printed. I think I'm leaning towards doing that. Would love to chat synchronously on the implications of implementing that.

src/pkg/lint/findings.go
AustinAbro321: Hmm pretty much the entirety of this file is UI stuff. Do you feel we should move this entire file minus the data types to message like `printCredentialsTable`?

AustinAbro321: Keeping this here until we have a better solution as moving to message would require keeping the lint types in the types package which we want to avoid

zarf.schema.json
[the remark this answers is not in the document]
AustinAbro321: resolving as this happened in another pr

phillebaba (changes requested):

Issue #2576, opened 2024-06-03
Refactoring of logging


phillebaba:
### Describe what should be investigated or refactored

Test week proved that certain functions are difficult because logging logic is tightly coupled with other logic. For example there are a couple of functions which will print formatted tables to a global logger instead of returning structured data and allowing the caller to deal with the output. Additionally there are a couple of places where errors are logged and exit is called instead of returning errors and allowing the called deal with them.  We have also discovered that Pterm is not thread safe as documented in #2558 which causes more issues for unit testing.

For these reasons it has become apparent that some sort of refactoring of logging is required. A lot has happened since the initial implementation, like structured logging in standard lib. We need to define what logging is specific to the CLI and what is logging which is useful for those importing Zarf as a CLI. While it may complicate things we have to decide whether or not spinners and other "UI" elements should be used in any exported packages.

### Links to any relevant code

https://github.com/defenseunicorns/zarf/blob/0c02643b9f5631a7bdd98650cc87c74901190771/src/pkg/utils/wait.go#L53
https://github.com/defenseunicorns/zarf/blob/0c02643b9f5631a7bdd98650cc87c74901190771/src/pkg/packager/common.go#L132

### Additional context

We should split the work into a couple of steps to avoid creating a mega PR. 

### Tasks
- [x] Add logger pkg
- [x] Introduce `--log-format` flag and disable message when it's enabled.

#### Add logger lines alongside existing message lines
(Note, not all of these commands will have messages but they're here to ensure we check them all)

pkgr
- [x] zarf package create
- [x] zarf package deploy
- [x] zarf package inspect
- [x] zarf package list
- [ ] zarf package publish
- [x] zarf package pull
- [x] zarf package mirror-resources
- [x] zarf package remove
 
init
- [x] zarf init

connect
- [x] zarf connect
- [x] zarf connect list

destroy
- [x] zarf destroy

dev
- [x] zarf dev deploy 
- [x] zarf dev generate
- [x] zarf dev find-images
- [x] zarf dev lint
- [x] zarf dev sha256sum
- [x] zarf dev patch-git

Not needed
- zarf version
- zarf tools
- zarf completion
- zarf internal
- zarf dev generate-config

phillebaba:
I realize that message fatal prints the debug stack when an error occurs which exits the program. Additionally some logs will differentiate between the message and the error. The error is only written to the debug logger which the message is written to the error writer.

Before removing this logic we need to determine if this is a feature required by end users.

[...]

Issue #2558, opened 2024-05-29
test: enable -race flag for running unit tests in CI


lucasrod16:
### Steps to reproduce
1. On the `main` branch of the repository, run all of the unit tests with the `-race` flag:

```bash
go test -failfast -v -race -coverprofile=coverage.out -covermode=atomic $(go list ./... | grep -v '^github.com/defenseunicorns/zarf/src/test' | grep -v 'github.com/defenseunicorns/zarf/src/extensions/bigbang/test')
```

### Expected result
All unit tests pass

### Actual Result
Unit tests fail due to data races

### Visual Proof (screenshots, videos, text, etc)
![data-race](https://github.com/defenseunicorns/zarf/assets/87675701/263e01a5-54b7-4d67-98ca-e5af2c51435c)

lucasrod16:
pterm's `SpinnerPrinter` and `ProgressbarPrinter` are not thread safe:

[Better concurrency pattern for LivePrinters #439](https://github.com/pterm/pterm/issues/439)
[Data race on pterm.SpinnerPrinter #482](https://github.com/pterm/pterm/issues/482)

There is an open draft PR that is supposed to fix most of these issues, but it appears to be stale:

[Fix most data race issues & enable in CI tests #447](https://github.com/pterm/pterm/pull/447)

This means our tests will fail with the `-race` flag, whether run sequentially or concurrently, due to the bugs mentioned above.

I will keep this issue open to track any progress on fixing the concurrency issues in pterm.

PR #2588, opened 2024-06-06
chore: logging ADR

phillebaba:
## Description

Adds an ADR for refactoring logging.

## Related Issue

Relates to #2576 

## Checklist before merging


phillebaba:
There are still some open questions that need to be decided.

* How should the logger be retrieved?  Passing through context vs. singleton.
* How do we support structured output like tables.
* How will the spinner work together with the logger.

schristoff-du:
What does this change look like from the user perspective?
How will this work be implemented (can we break it out into segments)
What other types of loggers were considered and why did we not go with them?

phillebaba:
I have extended the ADR a bit, my plan is to add some code examples to make it a bit easier to understand.

adr/0025-logging.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Decision
+
+I am proposing to completely refactor the logging functionality to follow a more standardized format using the newish slog interfaces. On top of that we would refactor the current internationalization by converting them to standardized errors.
+
+* Replace the message package with a slog interface.
+* Replace the lang package with static errors.
+* Remove any use of `message.Fatal` and instead return errors.
+* Refactor use of `message.WarnErr` to either return error or to log the error.
+* Refactor existing functions that print formatted outputs to defer the printing to the CLI.
```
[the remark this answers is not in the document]
phillebaba: You ask and I deliver.

adr/0025-logging.md
```diff
@@ -0,0 +1,235 @@
...
+
+Spinners and progress bars however are a bit more challenging. They need to be refactored so that they are no long instantiated outside of the CLI code. They should instead be passed as a function parameter to each function that needs them. A good example of a project that solves the problem in a similar manner is Kind. In Kind they create a [status object from the logger](https://github.com/kubernetes-sigs/kind/blob/7799e72306db315ea4f4b1cac90ff68404da4f28/pkg/internal/cli/status.go#L39) which they then [pass to where it is needed](https://github.com/kubernetes-sigs/kind/blob/7799e72306db315ea4f4b1cac90ff68404da4f28/pkg/cluster/internal/create/create.go#L133). Doing so results in a single status object created which is reused where ever it is needed. A lot of inspiration can be take from how Kind delas with CLI output. While they use klog instead of slog there are a lot of similarities. They have for example a check if the output is in a terminal or not, and will disable the spinner accordingly. 
+
+Here is a suggestion for how a thread safe spinner could be implemented with a shared logger. This also allows for paralell spinners and progress bars.
```
lucasrod16, line 67: Have you verified that this is thread-safe? Running `go run -race main.go` shows many data races in the pterm printers, likely related to issue #2558. It seems upstream changes to the pterm printers might be needed for thread safety

adr/0025-logging.md
```diff
@@ -0,0 +1,235 @@
...
+## Decision
+
+I am proposing to completely refactor the logging functionality to follow a more standardized format using the newish slog interfaces. On top of that we would refactor the current internationalization by converting them to standardized errors.
+
+* Replace the message package with a slog interface.
+* Replace the lang package with static errors.
+* Remove any use of `message.Fatal` and instead return errors.
+* Refactor use of `message.WarnErr` to either return error or to log the error.
+* Refactor existing functions that print formatted outputs to defer the printing to the CLI.
+* Define a new interface for spinner and progress bar that is passed in as a function parameter.
```
lucasrod16, line 22: I like these suggestions. Could we also aim to reduce the overall log output in Zarf as part of this refactor? What do you think?

phillebaba: I dont think that is the explicit plan but it will probably happen on its own. We should revisit the logs after this change is implemented.

adr/0025-logging.md
phillebaba: I initially did not understand all of the race issues in pterm. After looking through things it is obvious that there are issues in pterm that we cannot fix outside of the project. We have two options. Either we wait for pterm to fix the issue internally, or we find and alternative to pterm that offer the same features. It is slightly difficult to find a project that supports multi writer ontop of spinners and progress bars that is not from charm.

adr/0025-logging.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Decision
+
+I am proposing to completely refactor the logging functionality to follow a more standardized format using the newish slog interfaces. On top of that we would refactor the current internationalization by converting them to standardized errors.
+
+* Replace the message package with a slog interface.
+* Replace the lang package with static errors.
+* Remove any use of `message.Fatal` and instead return errors.
+* Refactor use of `message.WarnErr` to either return error or to log the error.
+* Refactor existing functions that print formatted outputs to defer the printing to the CLI.
```
Noxsios, line 21: Anything that does fancy UI things should prob happen in CMD, all lib code should use the `slog` interface

adr/0025-logging.md
```diff
@@ -0,0 +1,235 @@
...
+* Define a new interface for spinner and progress bar that is passed in as a function parameter.
+
+The message package currently exports the following functions which should be replaced by its logr counterpart.
+
+| Function | Replacement | Comment |
+| --- | --- | --- |
+| ZarfCommand | Info | Just formats a slice as a command. |
+| Command | Info | Outputs a command with a prefix and style. |
+| Debug | Degub | |
```
Noxsios, line 30:
```suggestion
| Debug | Debug | |
```

adr/0025-logging.md
```diff
@@ -0,0 +1,235 @@
...
+
+Here is a suggestion for how a thread safe spinner could be implemented with a shared logger. This also allows for parallel spinners and progress bars.
+
+```golang
+package main
+
+import (
+	"context"
```
Noxsios, line 73: I'm a big fan of the `FromContext`/`WithContext` pattern https://pkg.go.dev/github.com/charmbracelet/log#FromContext https://pkg.go.dev/github.com/charmbracelet/log#WithContext

Noxsios: should make it a little easier / standarized on how the logger is passed around library code: its always in `ctx`, `FromContext` also nicely handles if no logger has been passed, thereby preventing nil pointer errors / erroneous printing.

adr/0025-logging.md
phillebaba: While this is true it is not currently possible right now. The spinner is too heavily integrated into the all of the code base. The best we can do initially is to remove the need for a global spinner that has threading issues. Then in the future we can look at how we can move UI like behavior into the cmd package.

adr/0025-logging.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Context
+
+Zarf is currently using an in-house built logging solution which in turn depends on pterm. This solution is used to output information to the end user who is using the Zarf CLI. This output is for the most part formatted with the purpose of look good for users of the CLI. Logging function calls are done both in the CLI specific code as well as in the packages. The logging is implemented in such a way that different levels exist with different impacts and log destinations. A common pattern that is used is to call `message.Fatal` whenever an error occurs. This will output the message to STDERR while writing the actual error message to a debug log and exit the program. Exiting the program in this manner makes unit testing difficult while also skipping proper context handling skipping any clean-up that was intended to run before exiting the program. Some logging components like the progress bar and spinner are accessed through a shared global state that is not thread safe. The coupling to logging becomes complicated as disabling the progress bar is a challenge while multi threading these functions which access the global state results in complicated access patterns.
```
Racer159, line 11:
```suggestion
Zarf is currently using an in-house built logging solution which in turn depends on pterm. This solution is used to output information to the end user who is using the Zarf CLI. This output is for the most part formatted with the purpose of CLI user expierience. Logging function calls are done both in the CLI specific code as well as in the packages. The logging is implemented in such a way that different levels exist with different impacts and log destinations. A common pattern that is used is to call `message.Fatal` whenever an error occurs. This will output the message to STDERR while writing the actual error message to a debug log and exiting the program. Exiting the program in this manner makes unit testing difficult while also skipping proper context handling and skipping any clean-up that was intended to run before exiting the program. Some logging components like the progress bar and spinner are accessed through a shared global state that is not thread safe. The coupling to logging becomes complicated as disabling the progress bar is a challenge while multi threading these functions which access the global state resulting in complicated access patterns.
```

adr/0025-logging.md
Racer159: It may not be the best pattern but an implementation to look into could be what was done with the ProgressWriter and Maru - https://github.com/defenseunicorns/maru-runner/blob/main/src/message/progress.go#L29 - that makes the spinner and progressbar a bit more generic and a bit less decoupled.

The main thing I don't love about that implementation though it that spinner's `Write()` and progressbar's `Write()` mean slightly different things, but it is probably a decent starting point for `* Define a new interface for spinner and progress bar that is passed in as a function parameter.`

adr/0025-logging.md
```diff
@@ -0,0 +1,235 @@
...
+
+The message package currently exports the following functions which should be replaced by its logr counterpart.
+
+| Function | Replacement | Comment |
+| --- | --- | --- |
+| ZarfCommand | Info | Just formats a slice as a command. |
+| Command | Info | Outputs a command with a prefix and style. |
+| Debug | Degub | |
+| Debugf | Debug | |
+| ErrorWebf | N/A | Not used anywhere. |
```
Racer159, line 32: (historical context - this is vestigial from when Zarf had a web UI)

adr/0025-logging.md
```diff
@@ -0,0 +1,235 @@
...
+
+Spinners and progress bars however are a bit more challenging. They need to be refactored so that they are no long instantiated outside of the CLI code. They should instead be passed as a function parameter to each function that needs them. A good example of a project that solves the problem in a similar manner is Kind. In Kind they create a [status object from the logger](https://github.com/kubernetes-sigs/kind/blob/7799e72306db315ea4f4b1cac90ff68404da4f28/pkg/internal/cli/status.go#L39) which they then [pass to where it is needed](https://github.com/kubernetes-sigs/kind/blob/7799e72306db315ea4f4b1cac90ff68404da4f28/pkg/cluster/internal/create/create.go#L133). Doing so results in a single status object created which is reused where ever it is needed. A lot of inspiration can be take from how Kind deals with CLI output. While they use klog instead of slog there are a lot of similarities. They have for example a check if the output is in a terminal or not, and will disable the spinner accordingly.
```
Racer159, line 65:
```suggestion
Spinners and progress bars however are a bit more challenging. They need to be refactored so that they are no longer instantiated outside of the CLI code. They should instead be passed as a function parameter to each function that needs them. A good example of a project that solves the problem in a similar manner is Kind. In Kind they create a [status object from the logger](https://github.com/kubernetes-sigs/kind/blob/7799e72306db315ea4f4b1cac90ff68404da4f28/pkg/internal/cli/status.go#L39) which they then [pass to where it is needed](https://github.com/kubernetes-sigs/kind/blob/7799e72306db315ea4f4b1cac90ff68404da4f28/pkg/cluster/internal/create/create.go#L133). Doing so results in a single status object created which is reused where ever it is needed. A lot of inspiration can be take from how Kind deals with CLI output. While they use klog instead of slog there are a lot of similarities. They have for example a check if the output is in a terminal or not, and will disable the spinner accordingly. 
```

adr/0025-logging.md
```diff
@@ -0,0 +1,235 @@
...
+| Infof | Info | |
+| Success | Info | Seems to be a info log with a checkmark prefix. |
+| Successf | Info | Seems to be a info log with a checkmark prefix. |
+| Question | ? | Open question how to resolve this. |
+| Questionf | ? | Open question how to resolve this. |
+| Note | Info | Should just be an info or maybe a debug log. |
+| Notef | Info | Should just be an info or maybe a debug log. |
+| Title | ? | Not sure how this should be replaced as it formats with separator. |
+| HeaderInfof | ? | |
+| HorizontalRule | ? | |
```
Racer159, line 49: For these CLI UXy pieces it may be good to take a look at the Figma files @Madeline-UX put together to see how they fit / are organized there - there may be ways to build some of this in an "interactive" package that intentionally prints differently to the log vs the CLI - Horizontal Rules, ColorWrapped YAML, and Questions are really only there to be used in the CLI UI and could/should probably be represented differently in the log

schristoff-du (approved):

adr/0025-logging.md
schristoff: +1

schristoff (approved):

AustinAbro321 (approved):

Issue #1136, opened 2022-12-16
SPIKE: Zarf Plugin System implementation


runyontr:
# Plugin System

There has been discussion of a "plugin system" that would all users of zarf to build custom additions to the Zarf ecosystem that would allow new `components` in Zarf.yaml files to be recognized. 

## Questions

* What does the plugin need to implement?  Just a `package` function and a `deploy` function?  Do we require the plugin to generate SBOMs?
 * How does Zarf discover the plugins available?
 * How does Zarf bring that plugin to deployment?  How do we deal with different architectures of the computer running the plugin?  How will the Zarf controller need to be adjusted to leverage plugins?
 * How does Zarf run the plugin (e.g. within a container) and how does this impact the needed capabilities in both the package environment and the deploy environment?
 * How does someone making a plugin get started?  Are we providing a scaffold/sample like Kubebuilder does with controllers?

## Action Items
- [ ] Create ADR for plugin system
- [ ] Implement Plugin System hooks within Zarf
- [ ] Build toy plugin for testing/demo
- [ ] Add plugin system to docs
- [ ] Add tests for plugin system

Related:

- https://github.com/defenseunicorns/zarf/issues/939

jeff-mccoy:
How I see this working (and what I plan on experimenting with over the break): 

- a component can specify a number of `plugins` (name to tbd), with each one having some set of the following:
  - the image to use
  - the command to use on `package create`
  - the command to use on `package deploy`
  - maybe env variables for each?
- on `package create`, [Dagger SDK](https://dagger.io/blog/go-sdk) would be used to perform the operations, there may be an option to fallback to the host binaries if not available
- there would need to be some expected output format or file outputs for zarf to bundle for the other side
- on `package deploy`, push the image to the registry and schedule a K8s Job, perform data injected and use the specified command to create some effect
- probably have the ability to receive outputs from this job to provide to Zarf as #1137
- clean up after the job runs and move on with life

AustinAbro321:
Closing this as stale as we have no intention of creating a plugin system in the near future

PR #2600, opened 2024-06-07
chore: schema adr

AustinAbro321:
## Description

This ADR talks about the schema changes that will happen as we release Zarf v1 

## Checklist before merging


adr/0026-schema.md
```diff
@@ -0,0 +1,48 @@
...
+
+- `setVariable` deprecated in favor of `setVariables`
+- `scripts` deprecated in favor of `actions`
+- `required` soon to be deprecated in favor of `optional`.
+- `group` deprecated in favor of `flavor`, however these work functionally different and cannot be automatically migrated
+- `cosignKeyPath` deprecated in favor of specifying the key path at create time
+
+Zarf has not disabled any deprecated keys thus far. On create the user is always warned when using a deprecated field, however the field still exists in the schema and functions properly. Some of the deprecated keys can be migrated automatically as the old item is a subset or directly related to it's replacement. For example, setVariable is automatically migrated to a single item list of setVariables. This migration occurs on the zarf.yaml fed into the package during package create, however the original field is not deleted from the packaged zarf.yaml because the Zarf binary used in the airgap or delivery environment is not assumed to have the new schema fields that the deprecated key was migrated to.
+
+There are two problems this ADR aims to solve
```
lucasrod16, line 21:
```suggestion
There are two problems this ADR aims to solve

```

adr/0026-schema.md
```diff
@@ -0,0 +1,48 @@
...
+- Supporting deprecated features forever is complex and time consuming
+- Dropping feature support can frustrate or confuse users
+
+# Decisions
+
+Zarf will introduce proper schema versions. A top level key, `apiVersion`, will be introduced to allow users to specify the schema. If `apiVersion` is not specified it will be assumed to be the 1.0 schema. In v1.0, deprecated features will be entirely deleted from Zarf. Any key that exists at the introduction of 1.0 will last the entirety of that schema lifetime. The features may be deprecated, but will not be removed until the next schema version.
+
+The `zarf dev update-schema` command will be introduced to automatically update any deprecated fields in the users `zarf.yaml`, it will also add the apiVersion key and value. `zarf package create` will fail if the user has any deprecated keys and they will have a call to action to run `zarf dev update-schema`
+
+
```
lucasrod16, line 37:
```suggestion
```

adr/0026-schema.md
```diff
@@ -0,0 +1,48 @@
...
+## BDD scenarios
+### 1.0 create with deprecated keys
```
lucasrod16:
```suggestion
#### 1.0 create with deprecated keys

```

adr/0026-schema.md
```diff
@@ -0,0 +1,48 @@
...
+- *Given*: A user has a `zarf.yaml` with keys deprecated pre v1.0
+- *when* the user runs `zarf package create`
+- *then* they will receive an error and be told to run `zarf dev update-schema` or how to migrate off cosign key paths or how to use flavors over groups depending on the error
+
+### create with pre 1.0 -> deploy with 1.0
```
lucasrod16:
```suggestion
#### create with pre 1.0 -> deploy with 1.0

```

adr/0026-schema.md
```diff
@@ -0,0 +1,48 @@
...
+- *Given*: A package is created with Zarf pre v1.0
+- *and* that package has deprecated keys
+- *when* the package is deployed with Zarf v1.0
+- *then* the package will be read without error, but the deprecated keys will be ignored.
+
+### create with post 1.0 -> deploy with pre 1.0
```
lucasrod16:
```suggestion
#### create with post 1.0 -> deploy with pre 1.0

```

adr/0026-schema.md
lucasrod16, line 1: Could you please add a `Consequences` section to this ADR? It would help keep it consistent with the format of our existing ADRs by briefly outlining the pros and cons of the proposal

https://github.com/defenseunicorns/zarf/blob/main/adr/template.md#consequences

adr/0026-schema.md
```diff
@@ -0,0 +1,67 @@
...
+
+The existing types which comprise the Zarf schema will be moved to types/alpha and be renamed to alpha. Once v1 is released this package will not be touched again. A copy of these types will be created in a package called v1
+
+Any key that exists at the introduction of v1 will last the entirety of that schema lifetime. The features may be deprecated, but will not be removed until the next schema version.
+
+The `zarf dev update-schema` command will be introduced to automatically update deprecated fields in the users `zarf.yaml` where possible. It will also add the apiVersion key and set it to v1. `zarf package create` will fail if the user has deprecated keys with a list of the deprecated keys and a call to action to run `zarf dev update-schema`.
+
+
+## BDD scenarios
+### 1 create with deprecated keys
```
lucasrod16, line 39:
```suggestion
#### 1 create with deprecated keys

```

adr/0026-schema.md
```diff
@@ -0,0 +1,67 @@
...
+- *Given*: A user has a `zarf.yaml` with keys deprecated pre v1
+- *when* the user runs `zarf package create`
+- *then* they will receive an error and be told to run `zarf dev update-schema` or how to migrate off cosign key paths or how to use flavors over groups depending on the error
+
+### pre 1 auto migrate create -> 1 deploy
```
lucasrod16, line 44:
```suggestion
#### pre 1 auto migrate create -> 1 deploy

```

adr/0026-schema.md
```diff
@@ -0,0 +1,67 @@
...
+- *Given*: A package is created with Zarf pre v1
+- *and* that package has deprecated keys that can be automatically migrated (required, scripts, & set variables)
+- *when* the package is deployed with Zarf v1
+- *then* the keys will be automatically migrated & the package will be deployed without error.
+
+### pre 1 create feature removal -> 1 deploy
```
lucasrod16, line 50:
```suggestion
#### pre 1 create feature removal -> 1 deploy

```

adr/0026-schema.md
```diff
@@ -0,0 +1,67 @@
...
+- *Given*: A package is created with Zarf pre v1
+- *and* that package has deprecated keys that cannot be automatically migrated (groups, cosignKeyPath)
+- *when* the package is deployed with Zarf v1
+- *then* then deploy of that package will fail and the user will be instructed to update their package
+
+### 1 create -> pre 1 deploy
```
lucasrod16, line 56:
```suggestion
#### 1 create -> pre 1 deploy

```

adr/0026-schema.md
lucasrod16, line 39: I might be misunderstanding something here, but it seems like the `create with deprecated keys` scenario contradicts the `pre 1 auto migrate create -> 1 deploy` scenario.

The former says this about running `zarf p create` w/ deprecated keys pre v1:

>
- *Given*: A user has a `zarf.yaml` with keys deprecated pre v1
- *when* the user runs `zarf package create`
- *then* they will receive an error and be told to run `zarf dev update-schema` or how to migrate off cosign key paths or how to use flavors over groups depending on the error

The latter says this about running `zarf p create` w/ deprecated keys pre v1:

>  
- *Given*: A package is created with Zarf pre v1
- *and* that package has deprecated keys that can be automatically migrated (required, scripts, & set variables)
- *when* the package is deployed with Zarf v1
- *then* the keys will be automatically migrated & the package will be deployed without error.

Having `zarf package create` error out on deprecated keys might be frustrating for users on pre v1 version. It’s better to give a warning with suggestions so they can update without being completely blocked from creating packages.

Also, I believe the automatic migrations would happen during `package create` rather than `package deploy`. This is what we currently do in `ComposeComponents`

https://github.com/defenseunicorns/zarf/blob/fd452917584382f7ed588ffc5eaf428ecc7bb868/src/pkg/packager/creator/compose.go#L43

lucasrod16 (changes requested):

adr/0026-schema.md
AustinAbro321: It looks the same either way since the next line is a list

adr/0026-schema.md
AustinAbro321: They weren't worded well, but they don't contradict as the first situation involves creating a package on Zarf v1 whereas the latter is a packaged created pre v1 that is deployed on v1. We want the existing packages to still work on deploy, but we don't want new packages to be made using the deprecated fields.  Zarf will have a `zarf dev update-schema` command that is going to do everything automatically for users except for features that are removed (groups and cosignKeyPath). In addition, we want to force users to add an apiVersion. This allows us to assume that any package without an apiVersion is alpha (the current schema).

You're correct in that we currently do the automatic migration at create time, I just wanted to be explicit that packages using deprecated keys with automatic migration paths will still be supported in v1. The only situation this can happen is if the package was built pre v1.

adr/0026-schema.md
```diff
@@ -0,0 +1,569 @@
+# 25. Zarf Schema for 1
+
+Date: 2024-06-07
+
+## Status
+
+Proposed
+
+## Context
```
schristoff-du, line 9: I would love to see the context mention how we are basically creating the version schema to set a standard for what we are going to support long term/maintain backwards compat from v1.0 + 
It basically makes the schema a "contract" we have with our users

adr/0026-schema.md
```diff
@@ -0,0 +1,569 @@
...
+
+There are two problems this ADR aims to solve:
+- Supporting deprecated features forever is complex and time consuming
+- Dropping feature support can frustrate or confuse users
+
+## Decision
+
+Zarf will begin having proper schema versions. A top level key, `apiVersion`, will be introduced to allow users to specify the schema. At the release of v1 the only valid user input for `apiVersion` will be v1. Zarf will not allow users to build at the pre v1 version. `zarf package create` will fail if the user has deprecated keys or if `apiVersion` is missing and the user will be instructed to run the new `zarf dev update-schema` command. `zarf dev update-schema` will automatically migrate deprecated fields in the users `zarf.yaml` where possible. It will also add the apiVersion key and set it to v1.
```
schristoff-du, line 27: Why `apiVersion` instead of `schemaVersion` or just `version`?

schristoff-du: What decisions led to us supporting schema migration?

adr/0026-schema.md
```diff
@@ -0,0 +1,569 @@
...
+
+The existing go types which comprise the Zarf schema will be moved to types/alpha and will never change. An updated copy of these types without the deprecated fields will be created in a package types/v1 and all schema changes will happen on these objects. Any place in the code that is using a struct included in the Zarf schema will change from `types.structName` to `v1.structName`.
```
schristoff-du, line 29: Should we continue having an `alpha` if it is.. more GA? or just core to the project now?

adr/0026-schema.md
```diff
@@ -0,0 +1,569 @@
...
+
+All deprecated features will cause an error on create. Deprecated features with a direct migration path will still be deployed if the package was created v1, as migrations will add the non deprecated fields. If a feature does not have a direct automatic migration path (cosignKeyPath & groups) the package will fail on deploy. This will happen until the alpha schema is entirely removed from Zarf, which will happen one year after v1 is released.
+
+Any key that exists at the introduction of v1 will last the entirety of that schema lifetime. The features may be deprecated, but will not be removed until the next schema version.
```
schristoff-du, line 33: I would love to see some expansion on this and maybe a heading for deprecation process going forward, it seems like you're alluding that it will always take two schema versions (major, minor?) to deprecate a user facing feature

schristoff-du (commented):
This is amazing first pass and I know there will be lots to talk about on this

adr/0026-schema.md
phillebaba: apiVersion is used to keep similarity with the KRM.

https://github.com/kubernetes/design-proposals-archive/blob/main/architecture/resource-management.md

adr/0026-schema.md
```diff
@@ -0,0 +1,569 @@
...
+          autoIndent: true
+          pattern: ".+"
+          type: raw
+        description: action-description
+        wait:
+          cluster: # Only one of cluster / network can be used
+            kind: pod
+            name: my-pod
+            namespace: pod-ns
+            condition: ready
```
phillebaba, line 206: This may be the right time to change this feature to use kstatus instead?

adr/0026-schema.md
```diff
@@ -0,0 +1,569 @@
...
+- name: a-component
+  description: Zarf description
+  default: false # Austin to check if we remove this
+  only:
+    localOS: darwin
+    cluster:
+      architecture: amd64
+      distros:
+      - ubuntu
+    flavor: a-flavor # this will only be used when there are multiple components
```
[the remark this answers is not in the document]
AustinAbro321: Flavor is used pretty consistently across all of the UDS codebases. Generally there is an ironbank flavor (maybe chainguard now instead), and an upstream flavor. I haven't heard any complaints or confusion around the name from the UDS teams or the community. Though I could see the argument that it's not super discoverable for community members. Given that it's a feature that was built with the UDS use case in mind, and likely most community members don't have a need for this feature, I'd be in favor of keeping it as flavor.

adr/0026-schema.md
```diff
@@ -0,0 +1,569 @@
...
+    url: oci://
+  manifests:
+  - name: manifest
+    namespace: manifest-ns
+    files:
+    - a-file.yaml
+    kustomizeAllowAnyDirectory: false
+    kustomizations:
+    - a-kustomization.yaml
+    noWait: false
```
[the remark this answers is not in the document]
AustinAbro321: Yeah I think it'd in favor of changing it to `wait` that defaults to true. That's what the flag is called in helm as well https://helm.sh/docs/helm/helm_install/

adr/0026-schema.md
```diff
@@ -0,0 +1,569 @@
...
+          cluster: # Only one of cluster / network can be used
+            kind: pod
+            name: my-pod
+            namespace: pod-ns
+            condition: ready
+          network:
+            protocol: http
+            address: github.com
+            code: 200
+constants:
```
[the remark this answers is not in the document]
AustinAbro321: Yeah the idea is that they're unable to be changed after create whereas variables can be set on deploy

adr/0026-schema.md
AustinAbro321: alpha will only be used in the codebase and will not be an explicit version the user can use. The reason to have `alpha` is so we can delete the deprecated fields from the new struct, while still having the ability to read these fields into the old Zarf struct and transform it into the new struct. I should be more explicit that we're going to do a transformation.

adr/0026-schema.md
AustinAbro321: I was thinking it would just be one major version. So that contract would be anything in V1 at the start of v1 will be there until v2. Even if it's deprecated it will still be "supported" until v2 where it would be removed

adr/0026-schema.md
```diff
@@ -0,0 +1,569 @@
...
+  name: everything-zarf-package
+  description: A zarf package with a non empty value for every
+  version: v1.0.0
+  url: https://my-package-website.com
+  image: https://my-image-url-to-use-in-deprecated-zarf-ui # TODO This field should be deprecated
+  uncompressed: true
+  architecture: amd64
+  yolo: false
+  authors: cool-kidz
+  documentation: my-package-documentation
```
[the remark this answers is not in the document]
AustinAbro321: The description in jsonschema is "Link to package documentation when online".  I believe the idea is to be similar to `home` field in helm. Not sure if anyone actually uses this field though

adr/0026-schema.md
```diff
@@ -0,0 +1,569 @@
...
+
+## Consequences
+- As long as the only deprecated features in a package have migration path, and the package was built after the feature was deprecated so migrations were run, Zarf will be successful both creating a package with v1 and deploying with prev1, and creating a package with prev1 and deploying with v1.
+- Users of deprecated group or cosignKeyPath might be frustrated if their packages, created prev1, error out on Zarf v1, however this is likely preferable to unexpected behavior occurring in the cluster.
+- Users may be frustrated that they have to run `zarf dev update-schema` to edit their `zarf.yaml` to remove the deprecated fields and add `apiVersion`.
+- We will have to have two different schema types which will be mostly be duplicate code. However the original type should never change, which mitigates much of the issue.
+
+Below is an example v1 zarf.yaml with, somewhat, reasonable & nonempty values for every key
+```yaml
```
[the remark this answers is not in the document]
AustinAbro321: We auto gen a schema already from our go structs here https://github.com/defenseunicorns/zarf/blob/main/zarf.schema.json

adr/0026-schema.md
```diff
@@ -0,0 +1,569 @@
...
+            protocol: http
+            address: github.com
+            code: 200
+constants:
+- name: CONSTANT
+  value: constant-value
+  description: constant-value
+  autoIndent: false
+  pattern: ".+"
+variables:
```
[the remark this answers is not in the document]
AustinAbro321: Credentials are stored in zarf state and are cluster wide whereas variables are package specific. Sensitive variables operator as normal variables within a Zarf package the only difference is they are sanitized from the logs.

adr/0026-schema.md
AustinAbro321: Since we're going to have code in the backend anyway to transform deprecated fields into their new counterparts I feel it makes sense to expose that to the user to give them easier upgrade paths.

Additionally, Razzle already started on the code here https://github.com/defenseunicorns/zarf/pull/2211/files#diff-048b26f18d5c4dcd3484e355fb29fce16f91b49940754f2d5096f7c9494994c3.

adr/0026-schema.md
```diff
@@ -0,0 +1,569 @@
...
+
+The existing go types which comprise the Zarf schema will be moved to types/alpha and will never change. An updated copy of these types without the deprecated fields will be created in a package types/v1 and all schema changes will happen on these objects. Any place in the code that is using a struct included in the Zarf schema will change from `types.structName` to `v1.structName`.
+
+All deprecated features will cause an error on create. Deprecated features with a direct migration path will still be deployed if the package was created v1, as migrations will add the non deprecated fields. If a feature does not have a direct automatic migration path (cosignKeyPath & groups) the package will fail on deploy. This will happen until the alpha schema is entirely removed from Zarf, which will happen one year after v1 is released.
+
+Any key that exists at the introduction of v1 will last the entirety of that schema lifetime. The features may be deprecated, but will not be removed until the next schema version.
+
+### BDD scenarios
```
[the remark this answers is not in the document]
AustinAbro321: Added

adr/0026-schema.md
```diff
@@ -0,0 +1,578 @@
...
+    shasum: shasum
+    executable: false
+    symlinks:
+    - /path/to/symlink
+    extractPath: /path/to/extract
+  images:
+  - podinfo@v1
+  repos:
+  - https://github.com/defenseunicorns/zarf
+  extensions:
```
jeff-mccoy, line 169: I think we need to pull extensions out of zarf before 1.0. I know we've discussed before, but I cant seem to find an issue on it now.

jeff-mccoy (commented):
left some comments, great work everyone, can't wait to reach this milestone 🦄

adr/0026-schema.md
phillebaba: Yes it would mean removing the condition field and adding a apiVersion field.

adr/0026-schema.md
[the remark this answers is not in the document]
phillebaba: The reason this is needed is backwards compatibility. We need to write both versions of the schema to the package if we want to allow packages built with post v1 Zarf to deploy with pre v1 Zarf.

adr/0026-schema.md
```diff
@@ -0,0 +1,578 @@
...
+Below is an example v1 zarf.yaml with, somewhat, reasonable & nonempty values for every key
+```yaml
+kind: ZarfPackageConfig
+apiVersion: v1
+metadata:
+  name: everything-zarf-package
+  description: A zarf package with a non empty value for every
+  version: v1.0.0
+  url: https://my-package-website.com
+  image: https://my-image-url-to-use-in-deprecated-zarf-ui # TODO This field should be deprecated
```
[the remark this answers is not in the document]
AustinAbro321: Gotcha the json description mentioned it was for the Zarf UI which is why I added that note, but that's good to know. We will keep it in.

adr/0026-schema.md
AustinAbro321: I don't think we have an issue specifically for bigbang, but you may be referring to this issue https://github.com/defenseunicorns/zarf/issues/1136

adr/0026-schema.md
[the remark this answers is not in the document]
AustinAbro321: @jeff-mccoy Can you give me details or an example of what it'll be used for? I want to update the json description to be more clear

phillebaba:
As discussed during the meeting today, we should just add a new healthChecks field to the Zarf spec rather than modify the existing actions. This would create a single health check that would run after a deploy.

adr/0026-schema.md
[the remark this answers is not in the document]
schristoff: Is there an information on how this field is helpful to Zarf + UDS Marketplace, not just UDS side of things?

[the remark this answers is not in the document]
jeff-mccoy: Yeah I don't think it's just marketplace or zarf ui. It's helpful data to package in airgap that someone could use to show info about the package in a common way. Having an image for a package is very common for any tool listing application in a catalog, it could be a logo or screenshot. 

Leaving that in a standard place and having zarf load that on package create is something I still think we should do for the ux for anyone who builds any kind of package catalog.

adr/0026-schema.md
jeff-mccoy: Yeah that never happened obviously so this is one of those big things we should move to the SDK model prior to 1.0.

adr/0026-schema.md
```diff
@@ -106,6 +132,7 @@ build: # Everything here is created by Zarf not be users
     - missing-component
   flavor: cool-flavor
   lastNonBreakingVersion: "v0.99.9"
+  aggregateChecksum: shasum # this is moved from .metadata
```
schristoff-du, line 135: still cool with this being deprecated if it doesnt effect anything

AustinAbro321: I believe we want to keep it, just move it off metadata. Basically Zarf creates a checksum.txt file with all the sha's of every file that compromises a Zarf package. We then take the sha of the contents of that checksum.txt and that becomes the aggregate checksum. It allows us to ensure at deploy time the package is exactly what we wanted it to be immediately after create.

adr/0026-schema.md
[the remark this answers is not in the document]
phillebaba: One thought that was brought up would be a labels or annotations field. This would cover most of these use cases where users want to add arbitrary metadata. This is a common practice in other projects and would allow for more features than just images. Does that sound like a good idea @jeff-mccoy?

adr/0026-schema.md
```diff
@@ -0,0 +1,613 @@
...
+```yaml
+kind: ZarfPackageConfig
+apiVersion: v1
+metadata:
+  name: everything-zarf-package
+  description: A zarf package with a non empty value for every
+  version: v1.0.0
+  uncompressed: true
+  architecture: amd64
+  airgap: true # changed from yolo
```
lucasrod16, line 111: I assume the `airgap` key will default to `true`? So to pull from the internet on deploy users would need to use `airgap: false`?

schristoff: that is my understanding, correct.

adr/0026-schema.md
Noxsios: Its supposed to be on `build`, its a bit of an oopsie that its on a user-facing configuration

Noxsios: It actually _cannot_ be removed as it is used for package integrity

signature --> zarf.yaml --> aggregateChecksum --> checksums.txt

adr/0026-schema.md
```diff
@@ -0,0 +1,614 @@
...
+
+The release of v1 will provide an opportunity to delete deprecated features that Zarf has warned will be dropped in v1.
+
+Creating a v1 schema will allow Zarf to establish a contract with it's user base that features will be supported long term. When a feature is deprecated in the v1 schema, it will remain usable in the schema for the lifetime of v1.
+
+## Decision
+
+Zarf will begin having proper schema versions. A top level key, `apiVersion`, will be introduced to allow users to specify the schema. At the release of v1 the only valid user input for `apiVersion` will be v1. Zarf will not allow users to build using the v0 schema. `zarf package create` will fail if the user has deprecated keys or if `apiVersion` is missing and the user will be instructed to run the new `zarf dev update-schema` command. `zarf dev update-schema` will automatically migrate deprecated fields in the users `zarf.yaml` where possible. It will also add the apiVersion key and set it to v1.
```
Noxsios, line 30: Might be good to take a look at `zarf dev migrate` from https://github.com/defenseunicorns/zarf/pull/2211, some overlap in ideas

adr/0026-schema.md
```diff
@@ -0,0 +1,614 @@
...
+
+All deprecated features will cause an error on create. Deprecated features with a direct migration path will still be deployed if the package was created v1, as migrations will add the non deprecated fields. If a feature does not have a direct automatic migration path (cosignKeyPath & groups) the package will fail on deploy. This will happen until the alpha schema is entirely removed from Zarf, which will happen one year after v1 is released.
+
+At create time Zarf will package both a `zarf.yaml` and a `zarfv1.yaml`. If a `zarfv1.yaml` exists Zarf will use that. If a `zarfv1.yaml` does not exist, then Zarf will know that the package was created prior to v1 and use the regular `zarf.yaml`. If the package is deployed with v0 it will read the `zarf.yaml` as normal even if the package has a `zarfv1.yaml`. This will make it simpler to drop deprecated items with migration paths from the v1 schema while remaining backwards compatible as those deprecated items will exist in the `zarf.yaml`
```
Noxsios, line 36: Curious if this is needed.

Would going from v0 --> v1 of the schema happen before 1.0?

If it happens at the same time I don't think this deprecation logic needs to happen, as it can be understood that v1 `zarf.yaml`s will be unable to be deployed by v0.+ `zarf`, and vice-versa.  Any further deprecation changes can happen based upon `apiVersion`. Thoughts?

adr/0026-schema.md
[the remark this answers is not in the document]
Noxsios: Personally I would be fine w/ that breakage.

adr/0026-schema.md
```diff
@@ -0,0 +1,614 @@
...
+kind: ZarfPackageConfig
+apiVersion: v1
+metadata:
+  name: everything-zarf-package
+  description: A zarf package with a non empty value for every
+  version: v1.0.0
+  uncompressed: true
+  architecture: amd64
+  airgap: true # changed from yolo
+  labels: # All of these are v0 fields that will be deprecated in favor of a choose your own adventure label map
```
Noxsios, line 113: in the case of making it a map, it might be better to change this to `annotations` and let the keys conform to the OCI annotation spec (since that's what most of these fields get mapped to anyways)

adr/0026-schema.md
```diff
@@ -0,0 +1,614 @@
...
+            name: my-pod
+            namespace: pod-ns
+            condition: ready
+          network:
+            protocol: http
+            address: github.com
+            code: 200
+      onSuccess:
+      - mute: false
+        maxTotalSeconds: 0
```
Noxsios, line 287: now that context is used better, this could get changed to `timeout` that accepts a `time.Duration`

adr/0026-schema.md
```diff
@@ -0,0 +1,614 @@
...
+        maxRetries: 0
```
Noxsios, line 288: I think changing to `retries` would be better than `maxRetries`

adr/0026-schema.md
AustinAbro321: Yeah essentially the idea is copied from that PR. Internally we'll probably focus on changing that PR and getting it in relatively soon

adr/0026-schema.md
AustinAbro321: Ah yeah, I do like that, I hadn't realized most of these were the pre-definied annotation keys https://github.com/opencontainers/image-spec/blob/main/annotations.md. I think we'll still avoid the reverse domain notation and just maps the keys to the annotations in indexes or manifests if they exist on publish.

adr/0026-schema.md
AustinAbro321: Agree

adr/0026-schema.md
AustinAbro321: Are you picturing a format like what flux uses -> `1h5m0s`? I think that would be good.

adr/0026-schema.md
AustinAbro321: It would happen at the same time. 
Given that:
- most migrations will be automated
- users don't have a guarantee that their create version is the same as their deploy version
- Having a migration function should be relatively simple.
IMO it makes sense to keep the packages compatible. If we do end up making a change, like has been floated with actions, where  a large portion of packages can't be auto migrated then having v1 packages break on v0 deploys and vice versa may make sense.

adr/0026-schema.md
```diff
@@ -0,0 +1,615 @@
...
+In summary, a `wait` block will be auto migrated if, and only if:
+  - it is `onDeploy.after`
+  - the condition is `ready`, `available` or equivalent
+  - The kind is not a custom resource.
+- There may be further, more holistic changes to how actions works within Zarf that would have a significant affect on the schema. This will be covered in a separate ADR.
+
+### BDD scenarios
+The following are (behavior driven development)[https://en.wikipedia.org/wiki/Behavior-driven_development] scenarios to provide context of what Zarf will do in specific situations given the above decisions.
```
daveworth, line 70: Since we do not (to my knowledge - happy to be wrong) practice BDD explicitly and/or learn on gherkin style testing I suggest we rewrite/replace the scenarios as a slightly distilled bit of peers for each of these scenarios.  I think what the scenarios describe is quite useful - I'm not certain BDD/gherkin language adds much and they *may* distract from the content.

adr/0026-schema.md
```diff
@@ -0,0 +1,615 @@
...
+- *then* then deploy of that package will fail and the user will be instructed to update their package
+
+#### v1 create -> v0 deploy
+- *Given*: A package is created with Zarf v1
+- *and* that package uses keys that don't exist in v0
+- *when* the package is deployed with Zarf v0
+- *then* Zarf v0 will deploy the package without issues. If there are fields unrecognized by the v0 schema, then the user will be warned they are deploying a package that has features that do not exist in the current version of Zarf.
+
+## Consequences
+- As long as the only deprecated features in a package have migration path, and the package was built after the feature was deprecated so migrations were run, Zarf will be successful both creating a package with v1 and deploying with v0, and creating a package with v0 and deploying with v1.
```
daveworth, line 97:
```suggestion
- As long as the only deprecated features in a package have migration path, and the package was built after the feature was deprecated so migrations were run, Zarf will be successful in both creating a package with v1 and deploying with v0, and creating a package with v0 and deploying with v1.
```

adr/0026-schema.md
```diff
@@ -0,0 +1,615 @@
...
+- Users of deprecated group, cosignKeyPath, and action.Wait outside of onDeploy might be frustrated if their packages, created v0, error out on Zarf v1, however this is preferable to unexpected behavior occurring in the cluster.
```
daveworth, line 98:
```suggestion
- Users of deprecated `group`, `cosignKeyPath`, and `action.Wait` keys outside of `onDeploy` might be frustrated if their packages, created v0, error out on Zarf v1, however this is preferable to unexpected behavior occurring in the cluster.
```

adr/0026-schema.md
```diff
@@ -0,0 +1,615 @@
...
+- Users may be frustrated that they have to run `zarf dev update-schema` to edit their `zarf.yaml` to remove the deprecated fields and add `apiVersion`.
+- We will have to have two different schema types which will be mostly be duplicate code. However the original type should never change, which mitigates much of the issue.
```
daveworth, line 100:
```suggestion
- We will have to have two different schema types which will be mostly duplicated. However the original type should never change, which mitigates much of the issue.
```

In this case do we mean versions of schemas or types, as in data-types, of values?

adr/0026-schema.md
```diff
@@ -0,0 +1,615 @@
...
+- By having a zarf.yaml and a zarfv1.yaml it will be easy to read and write from objects to yamls directly without having to include deprecated v0 fields in the v1 schema. However, this will also mean any new keys in v1 won't exist in the `zarf.yaml` so a v0 deploy of a v1 package will not be able to warn users of unrecognized keys, we will have to use some other method.
```
daveworth, line 101:
```suggestion
- By having a `zarf.yaml` and a `zarfv1.yaml` it will be easy to read and write from objects to yaml files directly without having to include deprecated v0 fields in the v1 schema. However, this will also mean that any new keys in v1 won't exist in the `zarf.yaml` so a v0 deploy of a v1 package will not be able to warn users of unrecognized keys, we will have to use some other method.
```

adr/0026-schema.md
```diff
@@ -0,0 +1,615 @@
...
+
+Below is an example v1 zarf.yaml with, somewhat, reasonable & nonempty values for every key
```
daveworth, line 103: I would suggest this large YAML blob live outside of this ADR file and be referenced from here.

adr/0026-schema.md
AustinAbro321: I meant schema types, but that sentence was confusing. Clarified the sentence and added the examples we based this idea on, hopefully that helps

adr/0026-schema.md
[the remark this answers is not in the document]
AustinAbro321: Ended up making this an annotation

adr/0026-schema.md
AustinAbro321: Yeah I like that

adr/0026-schema.md
AustinAbro321: We encourage BDD in the [Zarf feature request template](https://github.com/defenseunicorns/zarf/issues/new?assignees=&labels=enhancement+%E2%9C%A8&projects=&template=feature_request.md&title=), but I recognize that not everyone is familiar with it. The substance of the scenarios is covered in the decisions tab, but I wanted to outline them more clearly as I believe they are what users will be most concerned with. Any suggestions on alternative formats?

daveworth: TIL!

For alternative formats I would say you could rewrite the first scenario as:

> For Zarf version v1 when the `zarf.yaml` has no apiVersion or deprecated keys when the user runs `zarf package create` they will receive an error and be told to run `zarf dev update-schema` or how to migrate off cosign key paths or how to use flavors over groups depending on the error.

daveworth: That said, my suggestion is not a *MUST* but more of a "maybe"

adr/0026-schema.md
```diff
@@ -0,0 +1,102 @@
...
+- *Given*: A package is created with Zarf v0
+- *and* that package has deprecated keys that cannot be automatically migrated (groups, cosignKeyPath)
+- *when* the package is deployed with Zarf v1
+- *then* then deploy of that package will fail and the user will be instructed to update their package
+
+#### v1 create -> v0 deploy
+- *Given*: A package is created with Zarf v1
+- *and* that package uses keys that don't exist in v0
+- *when* the package is deployed with Zarf v0
+- *then* Zarf v0 will deploy the package without issues. If there are fields unrecognized by the v0 schema, then the user will be warned they are deploying a package that has features that do not exist in the current version of Zarf.
```
Racer159, line 94: is this warning practically possible (it would need to be introduced - IIRC Zarf doesn't warn about extra keys today)

adr/0026-schema.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Consequences
+- As long as the only deprecated features in a package have migration path, and the package was built after the feature was deprecated so migrations were run, Zarf will be successful in both creating a package with v1 and deploying with v0, and creating a package with v0 and deploying with v1.
```
Racer159, line 97:
```suggestion
- As long as the only deprecated features in a package have a migration path, and the package was built after the feature was deprecated so migrations were run, Zarf will be successful in both creating a package with v1 and deploying with v0, and creating a package with v0 and deploying with v1.
```

adr/0026-schema.md
Racer159: I think timeout would be more Go/K8s friendly IMO (and not too hard to calculate / auto migrate)

adr/0026-schema.md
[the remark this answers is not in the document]
Racer159: One small note if image stays as is for the airgap use case - it would need to be baked into the package somehow for it to show to a user there.

adr/0026-schema.md
Racer159: Yeah I was tracking it against #1136 and agree it should get smushed out of the base schema

adr/0026-schema.md
AustinAbro321: Yeah it's planned to be introduced as a feature somewhat soon, but it will only work after it's introduced. I should note that in the ADR

adr/0026-schema.md
```diff
@@ -0,0 +1,102 @@
...
+In summary, a `wait` block will be auto migrated if, and only if:
+  - it is `onDeploy.after`
+  - the condition is `ready`, `available` or equivalent
+  - The kind is not a custom resource.
+- There may be further, more holistic changes to how actions works within Zarf that would have a significant affect on the schema. This will be covered in a separate ADR.
```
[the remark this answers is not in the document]
AustinAbro321: The idea of taking it away would be to make waits simpler to use, and therefore get more widespread adoption. Further detail is on this issue #2718. Could you link us situations where you use `onDeploy.Before` waits, or waits anywhere else. We want to better understand the use cases.

adr/0026-schema.md
```diff
@@ -0,0 +1,102 @@
...
+    url: https://stefanprodan.github.io/podinfo
+    path: charts/podinfo
+
+  oci:
+    url: oci://ghcr.io/stefanprodan/charts/podinfo
+
+  local:
+   path: chart
+```
+- `actions.wait` will be deprecated in favor of a wait list per component that is run on deploy. `action.wait` blocks that are `onDeploy.After` can be auto migrated and will work if deployed on v1. `wait` blocks that are not on `onDeploy.After` will not be auto migrated and the package will error out if deployed on v1. Zarf v1 will use kstatus for wait functionality instead of `kubectl wait`. `apiVersion` will be added as a required field on `wait`, and the `apiVersion` will be automatically set to the most recent `apiVersion` if the waited for resource is not a custom resource. Users can still wait on custom resources, but the resources must implement kStatus. `wait.condition` will be removed from the schema and Zarf will assume it is waiting for `ready`. A `wait` block will not be auto migrated if `condition` is not `set` to  `ready`, `available` or equivalent.
```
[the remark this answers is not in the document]
AustinAbro321: 1. Existence of resources should be [possible through kstatus](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md#statuses), but we will need to make sure we allow for more than just current. Could you comment on #2718 asking about this use case. 
3. I have not tested it but I assume [kstatus](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md#kstatus) has a current status for stateful sets which will cover that situation
4. Currently undecided from the team, but yes it should be explicitly stated in the ADR. Could you give us situations you see `http` waits used that would be hard to replace in a Kubernetes native way? One worry the team has about http waits is that we're encouraging users to use Zarf features rather than k8s features, expanding the scope for the tool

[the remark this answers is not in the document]
AustinAbro321: FYI there is a kstatus implementation for all Kubernetes native resources, the only question is if CRDs implement the necessary fields for kstatus

adr/0026-schema.md
AustinAbro321: Yeah it being more k8s friendly makes a lot sense, updated the adr

adr/0026-schema.md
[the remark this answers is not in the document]
AustinAbro321: Change of plans, we are going to keep the waits on onAny

schristoff (dismissed):

schristoff (approved):

PR #2606, opened 2024-06-10
feat: remove .metadata.image from schema

AustinAbro321:
## Description
This was a field reserved for the, now deprecated, Zarf UI 

## Checklist before merging


schristoff-du:
Can you also remove from [here](https://github.com/defenseunicorns/zarf/blob/main/examples/component-webhooks/capabilities/zarf-types.ts#L777)

schristoff-du (approved):

AustinAbro321:
@schristoff-du We agreed last week we would deprecate component webhooks. I just created [an issue for it.](https://github.com/defenseunicorns/zarf/issues/2610) The example isn't up to date with the schema anyway, I would rather just delete it.

PR #2614, opened 2024-06-11
fix: only build a single binary in the init-package make target

lucasrod16:
## Description
This PR updates the `init-package` make target to ensure it builds only the binary appropriate for the local system


schristoff-du (approved):
I didn't even know it ran this

PR #2615, opened 2024-06-11
fix: delete injector pod immediately

lucasrod16:
## Description
This PR adds back the behavior of deleting the injector pod immediately in `StopInjectionMadness` with no grace period before deletion. This was removed in [2553](https://github.com/defenseunicorns/zarf/pull/2553/files#diff-ce8c8d39489a071f6e62ecb731b08598f9a180bf773d2c216f67cb55a01497abL149) but it has lead to failures in our upgrade test because there are times when the injector pod from the initial `zarf init` in the test still exists in a `terminating` state, and it fails when we try to create the injector pod on the second "upgrade" `zarf init` because the pod already exists.

https://github.com/defenseunicorns/zarf/actions/runs/9470798079/job/26092606648?pr=2612#step:10:2776

### DIFF
--- a/src/pkg/cluster/injector.go
+++ b/src/pkg/cluster/injector.go
@@ -153,7 +153,13 @@ func (c *Cluster) StartInjectionMadness(ctx context.Context, tmpDir string, imag
 // StopInjectionMadness handles cleanup once the seed registry is up.
 func (c *Cluster) StopInjectionMadness(ctx context.Context) error {
 	// Try to kill the injector pod now
-	err := c.Clientset.CoreV1().Pods(ZarfNamespaceName).Delete(ctx, "injector", metav1.DeleteOptions{})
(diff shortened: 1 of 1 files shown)


lucasrod16:
The error is still occurring in CI. Need to look into this more

https://github.com/defenseunicorns/zarf/actions/runs/9473387226/job/26100983827?pr=2615#step:10:2786

phillebaba:
So there was a place where grace period actually matters.... :disappointed:  A bit annoying that the fix was applied to all pod deletes forcing us to guess which one needed it.

The other issue looks like it is related to the health checking that was done when port forwarding was done. This e2e test seems to also rely on the fact that it will wait for Pods to be ready.....

lucasrod16:
Closing in favor of #2620

PR #2618, opened 2024-06-12
feat: revert "feat: remove .metadata.image from schema (#2606)"

AustinAbro321:
This reverts commit ed2f876dc0ff2fed647b64d62f5a0843d41eed75.

## Description

We received feedback that people do have uses for the .metadata.images field. Adding it back

## Checklist before merging


AustinAbro321:
Closing for now, current proposed solution is an annotation field

Noxsios:
Target use case: populating an Android UI w/ thumbnail images of packages from Base64 encoded strings.

Eventually this would be better lived as an OCI annotation, `.metadata.image` is the best place until the annotation spec is implemented.

schristoff (approved):

PR #2620, opened 2024-06-12
fix: avoid injector pod name collisions

lucasrod16:
## Description
This PR changes how the injector pod is named. It creates a sha256 hash of the image name used to run the injector, and appends the hash to the pod name as a suffix to help avoid collisions where an injector pod is still in a `terminating` status from a previous `zarf init`.


src/pkg/packager/deploy.go
```diff
@@ -253,7 +253,10 @@ func (p *Packager) deployInitComponent(ctx context.Context, component types.Zarf
 
 	// Before deploying the seed registry, start the injector
 	if isSeedRegistry {
-		p.cluster.StartInjectionMadness(ctx, p.layout.Base, p.layout.Images.Base, component.Images)
+		err := p.cluster.StartInjectionMadness(ctx, p.layout.Base, p.layout.Images.Base, component.Images)
+		if err != nil {
+			return charts, err
```
phillebaba, line 258: charts should be nil here

phillebaba:
One small comment should be good to merge after that.

src/pkg/cluster/injector.go
```diff
@@ -384,13 +407,22 @@ func (c *Cluster) createService(ctx context.Context) (*corev1.Service, error) {
...
 func (c *Cluster) buildInjectionPod(node, image string, payloadConfigmaps []string, payloadShasum string) (*corev1.Pod, error) {
 	executeMode := int32(0777)
+
+	// Create a SHA-256 hash of the image name to allow unique injector pod names.
+	// This prevents collisions where `zarf init` is ran back to back and a previous injector pod still exists.
+	hasher := sha256.New()
+	if _, err := hasher.Write([]byte(image)); err != nil {
+		return nil, err
+	}
+	hash := hex.EncodeToString(hasher.Sum(nil))[:8]
```
AustinAbro321, line 417: If we select the same image as the first Zarf init, wouldn't we end up with the same sha making the collision still possible?

phillebaba: The Pod should eventually be deleted so this should not be an issue right? Unless you run them immediatly after each other.

AustinAbro321: Yes although running Zarf init right after each other is what is breaking the upgrade test. The bigger problem seems to be that the 0 second grace period isn't taking effect. Running locally the injector pod is terminating for around 20 seconds now.

I also don't think it's an unreasonable use case for a real user to run into. 
zarf init -> realize they forgot to select git server -> zarf init --components=git-server

src/pkg/packager/deploy.go
lucasrod16: updated

src/pkg/cluster/injector.go
AustinAbro321: Grace period of 0 is fine actually, likely we are going to refactor this anyway

lucasrod16: Setting grace period to 0 does not fix the error in CI

https://github.com/defenseunicorns/zarf/pull/2615#issuecomment-2161701286

I have yet to see the error occur in this PR after running multiple pipelines back to back

src/pkg/cluster/injector.go
```diff
@@ -115,59 +113,82 @@ func (c *Cluster) StartInjectionMadness(ctx context.Context, tmpDir string, imag
...
+		err = pkgkubernetes.WaitForReady(waitCtx, c.Watcher, objs)
+		if err != nil {
+			return err
+		}
+		spinner.Success()
 		// Otherwise just continue to try next image
 	}
-
-	// All images were exhausted and still no happiness
-	spinner.Fatalf(nil, "Unable to perform the injection")
```
AustinAbro321, line 150: We should return an error here, before the program was entirely stopping as spinner.Fatalf calls os.exit(1)

src/pkg/cluster/injector.go
```diff
@@ -56,43 +59,38 @@ func (c *Cluster) StartInjectionMadness(ctx context.Context, tmpDir string, imag
...
 	}
 
 	var err error
 	var images imageNodeMap
 	var payloadConfigmaps []string
 	var sha256sum string
-	var seedImages []transform.Image
 
-	// Get all the images from the cluster
-	spinner.Updatef("Getting the list of existing cluster images")
```
AustinAbro321, line 69: Why did we get rid of all the spinner updates?

AustinAbro321 (changes requested):

src/pkg/cluster/injector.go
lucasrod16: Because they are spammy and low value to the user IMO

lucasrod16: Also I didn't remove all of them. For example I kept the ones that update the user when long-running operations are happening, like creating the payload configmaps

AustinAbro321: Yeah good point, I like this better

src/pkg/cluster/injector.go
lucasrod16: The `spinner.Fatalf` was necessary before because we were logging errors in the for loop instead of returning them, so it used to be possible for the loop to complete even when errors occurred, but that is no longer possible because we return any errors we encounter immediately, which means if any iteration of the loop successfully creates the pod, the function will immediately return nil. The error path will never reach this branch if the function encounters any error. That was acting as a catch-all for errors

AustinAbro321: Actually we should probably only loop to get an image and after that do everything outside of a for loop. Since we switched the continue statements to returns. The only thing that worries me a bit is I don't have a good sense for how flaky the injector process is so I don't know if those continue statements were actually getting used

AustinAbro321: In that case we should only loop over the images until we have an acceptable image, then run everything else outside of the loop. 

The only worry I have is that I don't have a good sense if the continue blocks were getting used. I'd lean towards saying it's a good simplification though

lucasrod16: The loop over images may not be necessary. We plan to simplify this logic in #2621. Since the scope of this PR has already expanded, let’s handle that refactor separately

phillebaba (approved):
LGTM

src/pkg/cluster/injector.go
AustinAbro321: There was some confusion cleared up in convo, basically just simplified the code a bit. We will still need to do https://github.com/defenseunicorns/zarf/issues/2621 separately

src/pkg/cluster/injector.go
AustinAbro321: Added UUID instead of sha for better collision resistance

AustinAbro321 (approved):

phillebaba (approved):

PR #2623, opened 2024-06-13
fix: no longer remove the agent ignore label from namespaces

Racer159:
## Description

Fixes namespaces being able to be excluded from Zarf's mutation (deleting the label forces everything through the Zarf agent).

## Related Issue

Fixes #N/A

## Checklist before merging

- [X] [Contributor Guide Steps](https://github.com/defenseunicorns/zarf/blob/main/.github/CONTRIBUTING.md#developer-workflow) followed


Racer159:
Tested that the test got stuck at the right spot: https://github.com/defenseunicorns/zarf/actions/runs/9505355587/job/26200144190?pr=2623

AustinAbro321:
Great catch @Racer159. The intent of this deletion is to allow namespaces to be adopted that were set to be ignored by zarf init. Though it only actually should needs to happen in [post render here](https://github.com/defenseunicorns/zarf/blob/3932a1394bb63674de9cf8652040f40003dbff94/src/internal/packager/helm/post-render.go#L59).

AustinAbro321 (approved):

lucasrod16 (approved):

PR #2653, opened 2024-06-24
chore: run unit tests on main

AustinAbro321:
## Checklist before merging


schristoff-du (approved):
i would say we should add a badge for this in the readme but we have so many badges

PR #2656, opened 2024-06-24
test: site and links

### LINKED ISSUE #2649: Add link checker to docs in CI
We should have a github workflow that ensure that all workflows within Zarf and functional. It is too easy to change or remove the header of a page and break an existing link. This happened in #2643

AustinAbro321:
## Description

Fixes #2649 

## Checklist before merging

### DIFF
--- a/.github/workflows/test-site.yml
+++ b/.github/workflows/test-site.yml
@@ -1,0 +1,33 @@
+name: Test Site
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+# Abort prior jobs in the same workflow / PR
+concurrency:
+  group: site-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./site
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Setup Node.js
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+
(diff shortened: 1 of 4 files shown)


site/linkinator.config.json
```diff
@@ -1,4 +1,4 @@
 {
-  "skip": "^(https://.*github.com)",
+  "skip": "^(https://.*)",
```
AustinAbro321, line 2: Probably better to just scan internal links (there are still 236 of them) as this will avoid flakes and some sites might block return 403 from github runners

.github/workflows/test-site.yml
```diff
@@ -0,0 +1,33 @@
...
+        working-directory: ./site
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Setup Node.js
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+
+      - name: npm install
+        run: npm install
```
lucasrod16, line 27: We should use `npm ci` rather than `npm install` in CI environments

https://docs.npmjs.com/cli/v10/commands/npm-ci

.github/workflows/test-site.yml
```diff
@@ -0,0 +1,33 @@
+name: Test site
```
lucasrod16, line 1: To stay consistent with our other workflow names, we should capitalize each word in the workflow name

`Test Site`

lucasrod16 (changes requested):

schristoff-du (approved):
🤔  curious to see if the site thing will work as intended

lucasrod16 (approved):

PR #2660, opened 2024-06-24
refactor: make lint use more accessible data type

AustinAbro321:
## Description

This PR refactors lint to use a data type that's more accessible to other fields. This will eventually culminate in other packages also returning types.PackageError. We also add tests and some features to the schema. This will eventually allow us to get rid of some of the checks in `ZarfPackage.Validate()` in favor of native schema checks

Relates to #2485 

## Checklist before merging


src/pkg/packager/lint/lint.go
```diff
@@ -25,120 +22,109 @@ import (
...
 
-// ZarfSchema is exported so main.go can embed the schema file
-var ZarfSchema embed.FS
-
-func getSchemaFile() ([]byte, error) {
-	return ZarfSchema.ReadFile("zarf.schema.json")
+// FileLoader is an interface for reading files, it decouples the lint package from the go embed package
+type FileLoader interface {
+	ReadFile(path string) ([]byte, error)
 }
```
lucasrod16, line 28: Seems like a simple enough change, but I am curious why defining our own `FileLoader` interface is preferable over using the standard library `embed` package?

src/types/package.go
```diff
@@ -72,3 +72,29 @@ type ZarfBuildData struct {
...
+
+// Severity is the type of package error
+// Either Err or Warning
+type Severity int
+
+// different severities of package errors
+const (
+	SevErr Severity = iota + 1
+	SevWarn
+)
```
lucasrod16, line 100: These types and constants should probably live in the `src/pkg/packager/lint` package because they are used for linting and schema validation. It would move the data closer to the behavior and logic as well

src/types/component.go
```diff
@@ -121,7 +122,7 @@ type ZarfChart struct {
 	RepoName    string              `json:"repoName,omitempty" jsonschema:"description=The name of a chart within a Helm repository (defaults to the Zarf name of the chart)"`
 	GitPath     string              `json:"gitPath,omitempty" jsonschema:"description=(git repo only) The sub directory to the chart within a git repo,example=charts/your-chart"`
 	LocalPath   string              `json:"localPath,omitempty" jsonschema:"description=The path to a local chart's folder or .tgz archive"`
-	Namespace   string              `json:"namespace" jsonschema:"description=The namespace to deploy the chart to"`
+	Namespace   string              `json:"namespace,omitempty" jsonschema:"description=The namespace to deploy the chart to"`
```
lucasrod16, line 125: Are there any known side effects to making the `.charts.namespace` field optional rather than required?

src/types/package.go
```diff
@@ -21,7 +21,7 @@ type ZarfPackage struct {
 	Kind       ZarfPackageKind                 `json:"kind" jsonschema:"description=The kind of Zarf package,enum=ZarfInitConfig,enum=ZarfPackageConfig,default=ZarfPackageConfig"`
 	Metadata   ZarfMetadata                    `json:"metadata,omitempty" jsonschema:"description=Package metadata"`
 	Build      ZarfBuildData                   `json:"build,omitempty" jsonschema:"description=Zarf-generated package build data"`
-	Components []ZarfComponent                 `json:"components" jsonschema:"description=List of components to deploy in this package"`
+	Components []ZarfComponent                 `json:"components" jsonschema:"description=List of components to deploy in this package,minItems=1"`
```
lucasrod16, line 24: TIL about `minItems` for list validation in json schemas. Nice! This means we can delete [this code](https://github.com/defenseunicorns/zarf/blob/b72d4c8a5ed34b10b481cd3fe6c20337507ac064/src/types/validate.go#L52-L54)

https://json-schema.org/understanding-json-schema/reference/array#length

src/types/component.go
```diff
@@ -240,3 +241,16 @@ type ZarfComponentImport struct {
...
+	path, _ := schema.Properties.Get("path")
+	url, _ := schema.Properties.Get("url")
+
+	notSchema := &jsonschema.Schema{
+		Pattern: ZarfPackageTemplatePrefix,
+	}
+
+	path.Not = notSchema
+	url.Not = notSchema
+}
```
lucasrod16, line 256: Can we use the builtin `pattern=` struct tag to restrict URL and Path naming using regular expressions? Seems a bit redundant to have a pattern and a not pattern for the URL field, which also means we wouldn't need this `JSONSchemaExtend` function

lucasrod16 (changes requested):

src/pkg/packager/lint/lint.go
AustinAbro321: You cannot use simulate go embed in test AFAIK so this interface makes it possible to unit test the validate function

src/types/package.go
AustinAbro321: The plan is to have `packageError` used across Zarf as a way to store warnings or errors, as is done in #2585

src/types/component.go
AustinAbro321: We still have runtime validation after compose to ensure this is set https://github.com/defenseunicorns/zarf/blob/main/src/types/validate.go#L294-L296.

The issue with having it required is that it can be set by an import so vscode and `zarf dev lint` will incorrectly tell users their Zarf file is breaking the schema.

src/types/package.go
AustinAbro321: We can't delete it yet as this PR does not introduce schema validation during create. But it is deleted in #2585

src/types/component.go
AustinAbro321: I could be wrong here, but from what I've seen this is the only way as doing as a normal pattern would require a negative lookahead which is not allowed in go regex.

src/pkg/packager/lint/lint.go
lucasrod16: Would you mind adding that as a comment for context?

src/types/package.go
lucasrod16: Which fields are used for linting/validation versus generic warnings/errors? Can we achieve better separation of concerns and clarity by having one struct for validation errors and another for generic warnings/errors? Rather than using the same struct for all errors and cases where the validation specific fields are not needed

Something like this:

```go
type PackageError struct {
	Description string
	Category    Severity
}

type ValidationErr struct {
	PackageError
	// Fields specific to ValidationErr
	YqPath               string
	Item                 string
	PackageNameOverride  string
	PackagePathOverride  string
}
```

src/pkg/packager/lint/lint.go
```diff
@@ -25,120 +22,109 @@ import (
...
-// along with an error if the validation itself failed
-func Validate(ctx context.Context, createOpts types.ZarfCreateOptions) (*Validator, error) {
-	validator := Validator{}
-	var err error
+// ZarfSchema is exported so main.go can embed the schema file
+var ZarfSchema FileLoader
 
-	if err := utils.ReadYaml(filepath.Join(createOpts.BaseDir, layout.ZarfYAML), &validator.typedZarfPackage); err != nil {
+// Validate validates a zarf file
+func Validate(ctx context.Context, createOpts types.ZarfCreateOptions) ([]types.PackageError, error) {
```
lucasrod16, line 34: Instead of returning a custom error struct directly in our functions, how about we make our custom error types implement the `error` interface? Here’s why I think this would be beneficial:

1. This lets us add more context to our errors while still returning them as standard error types.

1. It ensures our errors play nicely with Go's built-in error handling functions like `errors.Is`, `errors.As`, and error wrapping with `%w` in `fmt.Errorf`.

1. Handling all errors as the error interface would allow us to manage both third-party and custom errors uniformly, which simplifies the code.

src/types/package.go
AustinAbro321: So the idea behind these fields is that packages outside of lint could create PackageErrors. Maybe finding would be a better name. I think in order to grab the context we want from errors we need creators of packageErrors to have access to these all of these fields even if they don't have something to put in all fields. 

`PackageNameOverride` & `packagePathOverride` are used when a error is found within an imported component so the user can be given the component that the error came from. 
'yqPath' is used when the error has a path in the code like `.metadata.name`
'Item' is used for additional context to the user, for example stating the image during an unpinned image warning

src/pkg/packager/lint/validator_test.go
```diff
@@ -0,0 +1,118 @@
...
+					Category: types.SevWarn,
+				},
+				{
+					Category: types.SevErr,
+				},
+			},
+			severity: types.SevErr,
+			expected: true,
+		},
+	}
```
lucasrod16, line 110: It would be good to add a test case to cover the `false` code path

src/pkg/packager/lint/validator_test.go
```diff
@@ -0,0 +1,118 @@
...
+	for _, tc := range tests {
+		tc := tc
+		t.Run("test has severity", func(t *testing.T) {
```
lucasrod16, line 113: Can we use a `name` field in the test case struct to capture the name of the individual test cases? When a test fails, it makes it easier to know which exact test case failed in the output

lucasrod16 (changes requested):
Few small things, otherwise looks good to me

lucasrod16 (approved):

src/pkg/packager/dev.go
```diff
@@ -105,3 +107,30 @@ func (p *Packager) DevDeploy(ctx context.Context) error {
 	// cd back
 	return os.Chdir(cwd)
 }
+
+// Lint ensures a package is valid & follows suggested conventions
+func (p *Packager) Lint(ctx context.Context) (err error) {
```
phillebaba, line 112: Dont use named returns unless it is actually required.

src/pkg/packager/dev.go
```diff
@@ -105,3 +107,30 @@ func (p *Packager) DevDeploy(ctx context.Context) error {
...
+	}
+
+	if len(lintFindings) == 0 {
+		message.Successf("0 findings for %q", p.cfg.Pkg.Metadata.Name)
+		return nil
+	}
+
+	lint.PrintFindings(lintFindings, types.SevWarn, p.cfg.CreateOpts.BaseDir, p.cfg.Pkg.Metadata.Name)
+
+	if lint.HasErrors(lintFindings) {
```
phillebaba, line 131: I understand what you are doing with this. Could we rename this to lint.HasFindings instead to not confuse people in the future?

src/pkg/packager/lint/lint.go
```diff
@@ -25,120 +22,109 @@ import (
...
+	ReadFile(path string) ([]byte, error)
 }
 
-// Validate validates a zarf file against the zarf schema, returns *validator with warnings or errors if they exist
-// along with an error if the validation itself failed
-func Validate(ctx context.Context, createOpts types.ZarfCreateOptions) (*Validator, error) {
-	validator := Validator{}
-	var err error
+// ZarfSchema is exported so main.go can embed the schema file
+var ZarfSchema FileLoader
```
phillebaba, line 31: Why are we not passing the schema as a function argument? I would like to avoid the use of global variables as much as possible.

src/pkg/packager/lint/lint.go
```diff
@@ -25,120 +22,109 @@ import (
...
 )
 
-// ZarfSchema is exported so main.go can embed the schema file
-var ZarfSchema embed.FS
-
-func getSchemaFile() ([]byte, error) {
-	return ZarfSchema.ReadFile("zarf.schema.json")
+// FileLoader is an interface for reading files, it decouples the lint package from the go embed package which enables testing
+type FileLoader interface {
+	ReadFile(path string) ([]byte, error)
```
phillebaba, line 27: This function is part of the fs.ReadFileFS interface.

https://pkg.go.dev/io/fs#ReadFileFS

src/pkg/packager/dev.go
```diff
@@ -105,3 +107,30 @@ func (p *Packager) DevDeploy(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("linting failed: %w", err)
+	}
+
+	if len(lintFindings) == 0 {
+		message.Successf("0 findings for %q", p.cfg.Pkg.Metadata.Name)
+		return nil
+	}
+
+	lint.PrintFindings(lintFindings, types.SevWarn, p.cfg.CreateOpts.BaseDir, p.cfg.Pkg.Metadata.Name)
```
phillebaba, line 129: I would prefer that the printing is done by caller of this function.

src/pkg/packager/lint/lint.go
```diff
@@ -25,120 +22,109 @@ import (
...
 		}
 	}
+	return findings, nil
 }
 
-func fillComponentTemplate(validator *Validator, node *composer.Node, createOpts *types.ZarfCreateOptions) {
-	err := creator.ReloadComponentTemplate(&node.ZarfComponent)
+func fillComponentTemplate(c *types.ZarfComponent, createOpts *types.ZarfCreateOptions) []types.PackageFinding {
+	err := creator.ReloadComponentTemplate(c)
+	var findings []types.PackageFinding
```
phillebaba, line 98: Flip this and the previous line to show that we are immediately checking the error.

phillebaba (changes requested):

src/pkg/packager/lint/lint.go
AustinAbro321: Thanks, I should have looked up and seen if there were any interfaces that met that.

src/pkg/packager/lint/lint.go
AustinAbro321: It's global because the value is set in main.go through go embed. Go embed has to happen in main because we can't use file backtrackers in go embed, for example, I can't do "../../../../zarf.schema.json". I think the only other way of passing it through to the commands would be adding it as a value to the context.

src/pkg/packager/dev.go
AustinAbro321: That might be misnomer since we don't want this function to return true if there are findings with a non error severity, maybe hasSevErrorFindings would be better? The goal is to return an error, and ultimately exit 1 the program, only if there are error findings

AustinAbro321: Or simply `HasSeverity(findings, types.SevErr)`

src/pkg/packager/dev.go
AustinAbro321: Yeah I don't really like print logic at the moment either. Do you think it would be worth creating a separate package outside of lint to store the `packageFinding` type and to do the printing to the user there? Most of what is in `src/pkg/package/lint/validate.go` feels like it should be it's own or a different package, but I'm not sure where that code should live.

Following the current codebase it would probably be in `message` beside things like `printCredentialTable`, though I'm not sure we want to keep following that pattern.

AustinAbro321: I changed it to do the printing in that function

src/pkg/packager/dev.go
```diff
@@ -105,3 +109,67 @@ func (p *Packager) DevDeploy(ctx context.Context) error {
...
+
+	findings, err := lint.Validate(ctx, p.cfg.CreateOpts)
+	if err != nil {
+		return fmt.Errorf("linting failed: %w", err)
+	}
+
+	if len(findings) == 0 {
+		message.Successf("0 findings for %q", p.cfg.Pkg.Metadata.Name)
+		return nil
+	}
```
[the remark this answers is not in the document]
AustinAbro321: Great catch thanks

src/pkg/packager/dev.go
```diff
@@ -112,12 +113,14 @@ func (p *Packager) DevDeploy(ctx context.Context) error {
...
 func (p *Packager) Lint(ctx context.Context) error {
-	config.CommonOptions.Confirm = true
-
 	if err := os.Chdir(p.cfg.CreateOpts.BaseDir); err != nil {
 		return fmt.Errorf("unable to access directory %q: %w", p.cfg.CreateOpts.BaseDir, err)
 	}
 
+	if err := utils.ReadYaml(layout.ZarfYAML, &p.cfg.Pkg); err != nil {
+		return err
+	}
```
lucasrod16, line 122: What do you think about changing `Validate` to accept a `ZarfPackage` struct as an argument, like we do with `lintComponents`?

This way, we can read the zarf.yaml file once and pass the pkg object to `Validate`:

```golang
Validate(ctx, p.cfg.Pkg,  p.cfg.CreateOpts)
```
Seems odd and inefficient to read the zarf.yaml twice here

It would avoid reading the file twice and make `Validate` easier to test

lucasrod16 (changes requested):

src/pkg/packager/dev.go
AustinAbro321: Yeah the only thing is that validate will have to assume the package it gets has not already been composed, but I think a comment on the function explaining that will be sufficient

AustinAbro321: Updated

src/pkg/packager/lint/lint.go
```diff
@@ -26,119 +24,99 @@ import (
...
-
-	validator.baseDir = createOpts.BaseDir
-
-	lintComponents(ctx, &validator, &createOpts)
-
-	if validator.jsonSchema, err = getSchemaFile(); err != nil {
+	var untypedZarfPackage interface{}
+	if err := utils.ReadYaml(layout.ZarfYAML, &untypedZarfPackage); err != nil {
 		return nil, err
 	}
```
lucasrod16, line 46: Why do we read the zarf.yaml into an interface here? Can we just pass the `pkg` argument to `validateSchema` instead?

lucasrod16 (changes requested):

src/pkg/packager/lint/lint.go
AustinAbro321: Good question, but it needs to be untyped because otherwise we would lose certain data when the yaml values are put into a struct and given zero values. The TestZarfSchema function in this PR has examples of where this is applicable (github is being weird about giving me a link)

lucasrod16: I deleted this code and `TestZarfSchema` still passes because `runSchema` accepts an interface, so we can still pass an arbitrary, untyped object in the test without having to read the file into memory again in `Validate`. So our actual code is strongly typed, but untyped for testing invalid schema fields and what not

src/pkg/packager/lint/lint_test.go
```diff
@@ -9,136 +9,249 @@ import (
...
 func readAndUnmarshalYaml[T interface{}](t *testing.T, yamlString string) T {
 	t.Helper()
 	var unmarshalledYaml T
 	err := goyaml.Unmarshal([]byte(yamlString), &unmarshalledYaml)
-	if err != nil {
-		t.Errorf("error unmarshalling yaml: %v", err)
-	}
+	require.NoError(t, err)
 	return unmarshalledYaml
 }
```
lucasrod16, line 26: I think this function isn't necessary. It's only used once and using generics here complicates things. It would be simpler to remove it and handle it directly where it's called.

```golang
var untypedPkg T
err := goyaml.Unmarshal([]byte(badZarfPackage), &untypedPkg)
require.NoError(t, err)
```

src/pkg/packager/lint/lint_test.go
```diff
@@ -9,136 +9,249 @@ import (
...
+	t.Parallel()
 	getZarfSchema := func(t *testing.T) []byte {
 		t.Helper()
 		file, err := os.ReadFile("../../../../zarf.schema.json")
-		if err != nil {
-			t.Errorf("error reading file: %v", err)
-		}
+		require.NoError(t, err)
 		return file
 	}
```
lucasrod16, line 35: Let's read the schema file once at the start of this test function and reuse the schema byte slice. This way, we avoid reading the file from disk multiple times.

lucasrod16 (changes requested):

src/pkg/packager/lint/lint.go
AustinAbro321: You can still pass in the package but when a yaml file is read into a Go struct it will drop fields that are not on the yaml object and Go will give zero values to every non pointer field in the struct. For example, you cannot replicate the below errors if you pass in a `ZarfPackage` struct. The extra field will be dropped, you can't make a string an int, and the `name` field will pass the required check since it exists with a `""` value . You can also see this by running the e2e test `TestLint`
image

src/pkg/packager/lint/lint_test.go
AustinAbro321: Yup, good catches

src/pkg/packager/lint/lint.go
lucasrod16: Oh I understand what you're saying now after looking at the `TestLint` e2e test. Makes sense thanks for the explanation

lucasrod16 (approved):

phillebaba (approved):
LGTM

PR #2664, opened 2024-06-25
refactor: remove use message.Fatal in cmd package

phillebaba:
## Description

This change removes use of message.Fatal in all Cobra commands. Instead all errors will be returned and logged at a single place.

## Related Issue

Relates to #2588

## Checklist before merging


src/cmd/package.go
```diff
@@ -258,23 +283,24 @@ func choosePackage(args []string) string {
...
+		return "", fmt.Errorf("package path selection canceled: %w", err)
 	}
 
-	return path
+	return path, nil
 }
 
-func identifyAndFallbackToClusterSource() (src sources.PackageSource) {
-	var err error
+// TODO: This code does not seem to do what it was intended.
```
AustinAbro321, line 292: how so?

src/cmd/connect.go
```diff
@@ -50,7 +50,7 @@ var (
 				tunnel, err = c.Connect(ctx, target)
 			}
 			if err != nil {
-				spinner.Fatalf(err, lang.CmdConnectErrService, err.Error())
+				return fmt.Errorf("unable to connect to the service: %w", err)
```
AustinAbro321, line 53:
```suggestion
				return fmt.Errorf("lost connection to the service: %w", err)
```
IK it was already like this but IMO this makes more sense

AustinAbro321 (changes requested):
small things

src/cmd/dev.go
```diff
@@ -249,19 +255,21 @@ var devLintCmd = &cobra.Command{
...
 		validator, err := lint.Validate(cmd.Context(), pkgConfig.CreateOpts)
 		if err != nil {
-			message.Fatal(err, err.Error())
+			return err
 		}
 		validator.DisplayFormattedMessage()
 		if !validator.IsSuccess() {
+			// TODO: Replace exit status with an error.
 			os.Exit(1)
 		}
```
lucasrod16, line 271: Noticed the TODO here. What do you think about replacing the exit status with an error?

src/cmd/tools/zarf.go
```diff
@@ -52,12 +54,22 @@ var getCredsCmd = &cobra.Command{
...
-	Run: func(cmd *cobra.Command, args []string) {
+	RunE: func(cmd *cobra.Command, args []string) error {
+		timeoutCtx, cancel := context.WithTimeout(cmd.Context(), cluster.DefaultTimeout)
+		defer cancel()
+		c, err := cluster.NewClusterWithWait(timeoutCtx)
+		if err != nil {
+			return err
+		}
+
 		ctx := cmd.Context()
```
lucasrod16, line 65: Would it make sense to either move this `ctx` variable assignment to the top of this `RunE` function and reuse it, or remove it and use `cmd.Context()` directly?

src/cmd/tools/zarf.go
```diff
@@ -76,27 +89,36 @@ var updateCredsCmd = &cobra.Command{
...
 		}
 
+		timeoutCtx, cancel := context.WithTimeout(cmd.Context(), cluster.DefaultTimeout)
+		defer cancel()
+		c, err := cluster.NewClusterWithWait(timeoutCtx)
+		if err != nil {
+			return err
+		}
+
 		ctx := cmd.Context()
```
lucasrod16, line 110: Would it make sense to either move this `ctx` variable assignment to the top of this `RunE` function and reuse it, or remove it and use `cmd.Context()` directly?

src/cmd/dev.go
phillebaba: I think @AustinAbro321 has fixed this in another PR already.

AustinAbro321: Yeah it's done here https://github.com/defenseunicorns/zarf/pull/2660

src/cmd/connect.go
phillebaba: I agree, my plan was to avoid changing error messages for now but this one is small.

AustinAbro321 (approved):

PR #2667, opened 2024-06-26
feat: diff images during dev deploy

Noxsios:
## Description

Back to back `zarf dev deploy <dir> --no-yolo` should diff the images in a running registry and only pull + push images that do not exist.

This will make `no-yolo` back to back `dev-deploy`s muuuch faster, getting `dev deploy` closer to `helm repo install`.

curent edge cases (wip solving):

- image tag changes but name stays the same between `dev deploy` runs

This does however move `dev deploy --no-yolo` further away from a true full mirror of `zarf package create && zarf package deploy` since it will be mutating the `zarf.yaml` stored in the cluster, removing images that are not pulled + pushed.

This means that commands like `zarf tools registry prune` and other stuff that relies upon the consistency of the `zarf.yaml`s stored in the cluster is messed up. _buuuuut_ this is `dev deploy`, its not supposed to be run in ANY prod env or persistent setting.


src/pkg/packager/dev.go
```diff
@@ -67,6 +72,41 @@ func (p *Packager) DevDeploy(ctx context.Context) error {
...
+
+			return diff
+		}
+
+		normalizeImageName := func(s string) string {
+			imgInfo, _ := transform.ParseImageRef(s)
+			return imgInfo.Path
+		}
+
+		if c, err := cluster.NewCluster(); err == nil { // if NO error
```
phillebaba, line 95: We should stop creating the cluster inside of packages.

phillebaba (commented):
Could we get a test for this use case?

Noxsios:
> Could we get a test for this use case?

ya, still doing experimentation on how I think it should work

this patch is mainly just for some dash days speed increases, after this week i'll get it actually fleshed out

schristoff:
Hey, thank you so much for this PR. I'm closing this due to it being stale but feel free to reopen it in the future and reference this PR :)

PR #2678, opened 2024-06-28
feat: add health checks

### LINKED ISSUE #2203: Refactor action wait code

phillebaba:
## Description

This change adds a new `healthChecks` field which will verify the health of any resource after deployment. This PR does not affect the current wait logic in actions, both exist simultaneously.

## Related Issue

Relates to #2203 

## Checklist before merging

### DIFF
--- a/src/config/lang/english.go
+++ b/src/config/lang/english.go
@@ -276,7 +276,7 @@ $ zarf package mirror-resources <your-package.tar.zst> \
 	CmdPackageDeployFlagShasum                         = "Shasum of the package to deploy. Required if deploying a remote package and \"--insecure\" is not provided"
 	CmdPackageDeployFlagSget                           = "[Deprecated] Path to public sget key file for remote packages signed via cosign. This flag will be removed in v1.0.0 please use the --key flag instead."
 	CmdPackageDeployFlagSkipWebhooks                   = "[alpha] Skip waiting for external webhooks to execute as each package component is deployed"
-	CmdPackageDeployFlagTimeout                        = "Timeout for Helm operations such as installs and rollbacks"
+	CmdPackageDeployFlagTimeout                        = "Timeout for health checks and Helm operations such as installs and rollbacks"
 	CmdPackageDeployValidateArchitectureErr            = "this package architecture is %s, but the target cluster only has the %s architecture(s). These architectures must be compatible when \"images\" are present"
 	CmdPackageDeployValidateLastNonBreakingVersionWarn = "The version of this Zarf binary '%s' is less than the LastNonBreakingVersion of '%s'. You may need to upgrade your Zarf version to at least '%s' to deploy this package"
 	CmdPackageDeployInvalidCLIVersionWarn              = "CLIVersion is set to '%s' which can cause issues with package creation and deployment. To avoid such issues, please set the value to the valid semantic version for this version of Zarf."
--- a/src/pkg/packager/deploy.go
+++ b/src/pkg/packager/deploy.go
@@ -401,7 +401,7 @@ func (p *Packager) deployComponent(ctx context.Context, component v1alpha1.ZarfC
 	if len(component.HealthChecks) > 0 {
 		healthCheckContext, cancel := context.WithTimeout(ctx, p.cfg.DeployOpts.Timeout)
 		defer cancel()
-		spinner := message.NewProgressSpinner("Running Health checks for %s", component.Name)
+		spinner := message.NewProgressSpinner("Running health checks")
 		defer spinner.Stop()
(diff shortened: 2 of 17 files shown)


src/config/lang/english.go
```diff
@@ -276,7 +276,7 @@ $ zarf package mirror-resources <your-package.tar.zst> \
 	CmdPackageDeployFlagShasum                         = "Shasum of the package to deploy. Required if deploying a remote package and \"--insecure\" is not provided"
 	CmdPackageDeployFlagSget                           = "[Deprecated] Path to public sget key file for remote packages signed via cosign. This flag will be removed in v1.0.0 please use the --key flag instead."
 	CmdPackageDeployFlagSkipWebhooks                   = "[alpha] Skip waiting for external webhooks to execute as each package component is deployed"
-	CmdPackageDeployFlagTimeout                        = "Timeout for Helm operations such as installs and rollbacks"
+	CmdPackageDeployFlagTimeout                        = "Timeout for health checks and Helm operations such as installs and rollbacks"
```
AustinAbro321, line 279: Do we want the existing `--timeout` flag to apply to health checks or should we implement a new flag. Alternatively, we could add timeout to the schema, though I'm not a big fan of that, especially if we want to add a wait flag that works with /health checks as well.

src/pkg/packager/deploy_test.go
```diff
@@ -272,3 +282,81 @@ func TestServiceInfoFromServiceURL(t *testing.T) {
...
+
+var podYaml = `
+apiVersion: v1
+kind: Pod
+metadata:
+  name: in-progress-pod
+  namespace: ns
+`
+
+func yamlToUnstructured(t *testing.T, yml string) *unstructured.Unstructured {
```
phillebaba, line 307: Add `t.Helper`

src/pkg/packager/deploy_test.go
```diff
@@ -272,3 +282,81 @@ func TestServiceInfoFromServiceURL(t *testing.T) {
...
+					Namespace:  pod.GetNamespace(),
+					Name:       pod.GetName(),
+				},
+			}
+			err := runHealthChecks(ctx, statusWatcher, objs)
+			if tt.expectErr != nil {
+				require.ErrorIs(t, err, tt.expectErr)
+			} else {
+				require.NoError(t, err)
+			}
```
phillebaba, line 359:
```suggestion
			if tt.expectErr != nil {
				require.ErrorIs(t, err, tt.expectErr)
				return
			} 
		    require.NoError(t, err)
```

src/pkg/packager/deploy.go
```diff
@@ -370,6 +398,17 @@ func (p *Packager) deployComponent(ctx context.Context, component v1alpha1.ZarfC
 		return nil, fmt.Errorf("unable to run component after action: %w", err)
 	}
 
+	if len(component.HealthChecks) > 0 {
+		healthCheckContext, cancel := context.WithTimeout(ctx, p.cfg.DeployOpts.Timeout)
+		defer cancel()
+		spinner := message.NewProgressSpinner("Running Health checks for %s", component.Name)
```
phillebaba, line 404: Lets not add more spinners tot things before we figure out how we want to deal with this long term.

AustinAbro321: I see your point, but I do think we should give some feedback to the users that health checks are running. I'm fine with switching to a message.Infof() though it looks a bit strange since every other part of the deployment uses a spinner which leaves a different text color and has a check mark.

image

src/pkg/packager/deploy.go
```diff
@@ -370,6 +398,17 @@ func (p *Packager) deployComponent(ctx context.Context, component v1alpha1.ZarfC
 		return nil, fmt.Errorf("unable to run component after action: %w", err)
 	}
 
+	if len(component.HealthChecks) > 0 {
+		healthCheckContext, cancel := context.WithTimeout(ctx, p.cfg.DeployOpts.Timeout)
+		defer cancel()
+		spinner := message.NewProgressSpinner("Running health checks")
```
schristoff, line 404: Do we need this?

schristoff (approved):

PR #2681, opened 2024-06-29
fix: remove unpinned image warning in lint for cosign signatures

### LINKED ISSUE #2577: Zarf dev lint warns that cosign signatures are not pinned with a digest
Currently running `zarf dev lint` against a `zarf.yaml` with a cosign signature will give a warning like below. It's redundant as cosign signatures are unique to an image sha, and you will not have two cosign signatures at the same image
```
Image not pinned with digest - ghcr.io/stefanprodan/podinfo:sha256-57a654ace69ec02ba8973093b6a786faa15640575fbf0dbb603db55aca2ccec8.sig
```

If an image has ".sig" at the end, it should simply be ignored from lint

jasonwashburn:
## Description
Removes unpinned image warning generated by `zarf dev lint` for cosign signatures.
...

## Related Issue

Fixes #2577 

## Checklist before merging

- [ x ] [Contributor Guide Steps](https://github.com/defenseunicorns/zarf/blob/main/.github/CONTRIBUTING.md#developer-workflow) followed

### DIFF
--- a/src/pkg/packager/lint/lint.go
+++ b/src/pkg/packager/lint/lint.go
@@ -64,7 +64,6 @@ func lintComponents(ctx context.Context, pkg types.ZarfPackage, createOpts types
 		}
 
 		chain, err := composer.NewImportChain(ctx, component, i, pkg.Metadata.Name, arch, createOpts.Flavor)
-
 		if err != nil {
 			return nil, err
 		}
@@ -144,7 +143,7 @@ func isPinnedImage(image string) (bool, error) {
 		}
 		return false, err
 	}
-	if isCosignSignature(transformedImage.Tag) {
+	if isCosignSignature(transformedImage.Tag) || isCosignAttestation(transformedImage.Tag) {
 		return true, nil
 	}
 	return (transformedImage.Digest != ""), err
@@ -152,6 +151,10 @@ func isPinnedImage(image string) (bool, error) {
 
 func isCosignSignature(image string) bool {
 	return (strings.HasSuffix(image, ".sig"))
+}
+
+func isCosignAttestation(image string) bool {
+	return (strings.HasSuffix(image, ".att"))
 }
 
 func isPinnedRepo(repo string) bool {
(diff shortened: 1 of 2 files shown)


jasonwashburn:
@AustinAbro321 Would the preference for this to treat signatures effectively as pinned (`isPinnedImage`)  or rather to add separate handling i.e. `isSignedImage` ?

AustinAbro321:
Thanks @jasonwashburn, looks mostly good. Could you also have the linter ignore ".att" images ending in .att, as these images are also already sha'd and used by cosign for provenance and integrity.

jasonwashburn:
> Thanks @jasonwashburn, looks mostly good. Could you also have the linter ignore ".att" images ending in .att, as these images are also already sha'd and used by cosign for provenance and integrity.

Sure thing!

src/pkg/packager/lint/lint.go
```diff
@@ -144,9 +143,20 @@ func isPinnedImage(image string) (bool, error) {
...
 		return false, err
 	}
+	if isCosignSignature(transformedImage.Tag) || isCosignAttestation(transformedImage.Tag) {
+		return true, nil
+	}
 	return (transformedImage.Digest != ""), err
 }
 
+func isCosignSignature(image string) bool {
+	return (strings.HasSuffix(image, ".sig"))
```
phillebaba, line 153: Remove the excessive parentheses wrapping the function.

src/pkg/packager/lint/lint.go
```diff
@@ -144,9 +143,20 @@ func isPinnedImage(image string) (bool, error) {
...
+}
+
+func isCosignAttestation(image string) bool {
+	return (strings.HasSuffix(image, ".att"))
```
phillebaba, line 157: Remove the excessive parentheses wrapping the function.

phillebaba (changes requested):
Small NIT otherwise looks good, thank you for expanding the tests to cover this.

src/pkg/packager/lint/lint.go
jasonwashburn: 🤦‍♂️ good catch, fixed!

phillebaba (approved):

PR #2683, opened 2024-07-01
refactor: remove use of os.Chdir

### LINKED ISSUE #2682: Remove use of os.Chdir

There are a couple of places in the code base where os.Chdir is used to access files in specific paths. Change of working directory is done temporarily in a function. Usually in the function the working directory is changed back in a deferred function. This is not a thread safe operation as setting the working directory is a per process system call. Packager and other components should support being used in multiple goroutines. For this reason we need to remove any use of os.Chdir. 

https://github.com/defenseunicorns/zarf/blob/601a282c78701bb039924c6c72802690656395df/src/pkg/packager/publish.go#L51
https://github.com/defenseunicorns/zarf/blob/601a282c78701bb039924c6c72802690656395df/src/pkg/packager/dev.go#L37

### Additional context

N/A

phillebaba:
## Description

This change removes use of os.Chdir from exported code.

## Related Issue

Fixes #2682 

## Checklist before merging

### DIFF
--- a/src/pkg/packager/create.go
+++ b/src/pkg/packager/create.go
@@ -7,7 +7,7 @@ package packager
 import (
 	"context"
 	"fmt"
-	"os"
+	"path/filepath"
 
 	"github.com/defenseunicorns/pkg/helpers/v2"
 	"github.com/defenseunicorns/zarf/src/config"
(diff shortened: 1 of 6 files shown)


phillebaba:
I realize that packager needs to be refactored before we can do this. There are too many struct properties being used as function parameters which makes tracing down paths more difficult than it should be.

schristoff-du:
This is probably my favorite PR of yours so far

PR #2690, opened 2024-07-02
ADR: Practical Component Composition

marshall007:
## Description

Composition within Zarf packages feels clunky and difficult right now. Specifically, there is no way to declare a self-contained "optional" component that overrides Helm chart values (or otherwise modifies the configuration of component(s) it is intended to be used with).

## Checklist before merging

### DIFF
--- a/adr/0025-composition.md
+++ b/adr/0025-composition.md
@@ -140,6 +140,7 @@ components:
   - name: mattermost
     description: "Deploy Mattermost"
     required: true
+
     ## The `mattermost` component is now a self-contained, deployable unit.
     ## Thus, all the configuration from `common/zarf.yaml` has been inlined
     ## into this example. We no longer import a `common` "skeleton package".
@@ -162,7 +163,12 @@ components:
         version: 2.6.55
         valuesFiles:
           - values/common-values.yaml
-          - values/upstream-values.yaml
+
+          ## No longer needed because this was only used for replacing
+          ## image tags, see new `images` API usage below.
+          #
+          # - values/upstream-values.yaml
+
     ## Kustomize-style image replacements facilitate proper composition.
     ## This makes it easy for downstream components to override image tags
     ## without knowing anything about the Helm chart(s) being referenced
@@ -189,6 +195,7 @@ components:
                 condition: "'{.status.phase}'=Ready"
 
   - name: mattermost-registry1
+
     ## Previously a `required` component with the "registry1" `flavor`,
     ## `mattermost-registry1` is now defined simply as a component that directly
     ## extends the default `mattermost` component above.
@@ -237,6 +244,7 @@ components:
   ## (i.e. `mattermostApp.extraInitContainers`).
   - name: mattermost-plugins
     required: false
+
     ## Note that the semantics of `requires` is slightly different from `extends`.
     ## The idea is that `extends` signals intent to _replace_ the original component
     ## whereas `requires` signals that the originaly component(s) are being overlaid


adr/0025-composition.md
```diff
@@ -0,0 +1,282 @@
...
+
+Date: 2024-07-02
+
+## Status
+
+Draft
+
+## Context
+
+Presently "composition" within Zarf is only possible at the package level. This can only be done with a
```
Racer159, line 11: It can also be done locally with path: ../ but agree the skeleton is the more obtuse use case (and has been acknowledged in the past)

Racer159: (skeleton is only used in the OCI case in Zarf)

adr/0025-composition.md
```diff
@@ -0,0 +1,282 @@
...
+          - description: Validate Mattermost Package
+            maxTotalSeconds: 300
+            wait:
+              cluster:
+                kind: Packages
+                name: mattermost
+                namespace: mattermost
+                condition: "'{.status.phase}'=Ready"
+
+  - name: mattermost-registry1
```
Racer159, line 197: how would you build on component vs another?

adr/0025-composition.md
```diff
@@ -0,0 +1,282 @@
...
+    #       #     tag: 9.9.0
+    #       # initContainerImage:
+    #       #   repository: registry1.dso.mil/ironbank/redhat/ubi/ubi9-minimal
+    #       #   tag: 9.4
+    # images:
+    #   - registry1.dso.mil/ironbank/redhat/ubi/ubi9-minimal:9.4
+    #   - registry1.dso.mil/ironbank/opensource/mattermost/mattermost:9.9.0
+    images:
+      - name: appropriate/curl
+        newName: registry1.dso.mil/ironbank/redhat/ubi/ubi9-minimal
```
Racer159, line 234: feels verbose / complex to me (you'd have to match strings in a Helm post render which might not be reliable - these references could be anywhere including a command arg passed into an operator)

adr/0025-composition.md
```diff
@@ -0,0 +1,282 @@
...
+          #           mountPath: /mattermost/plugins/
+    images:
+      - name: uds-package-mattermost/mattermost-extra-plugins
+        newTag: latest
+    actions:
+      onCreate:
+        before:
+          - dir: plugins
+            cmd: |
+              docker buildx . -t uds-package-mattermost/mattermost-extra-plugins:latest
```
Racer159, line 275: The other way to solve this could be at the UDS CLI level to allow an additional image to be included alongside a Helm override (may still need an API change to the Zarf SDK to support but may be cleaner)

adr/0025-composition.md
```diff
@@ -0,0 +1,282 @@
...
+    images:
+      - registry1.dso.mil/ironbank/redhat/ubi/ubi9-minimal:9.4
+      - registry1.dso.mil/ironbank/opensource/mattermost/mattermost:9.9.0
+```
+
+### Proposed Solutions
+
+1. `components[].extends` (`string`): delcares this component as an extension (overlay) of another component.
+   Similar to `flavor`, the resulting component is considered the "deployable unit" and cannot be deployed alongside the component it extends.
+2. `components[].requires` (`[]string`): similar to `extends`, declares this component as an extension (overlay) of another component.
```
Racer159, line 108: Zarf has discussed boing from `required` -> `optional` but this may conflict with `required` semantics currently.

adr/0025-composition.md
Racer159: (we could probably also get away from a docker build here ourselves and publish this image instead (and require others to do the same for theirs))

adr/0025-composition.md
```diff
@@ -0,0 +1,282 @@
...
+
+    ## Note that the semantics of `requires` is slightly different from `extends`.
+    ## The idea is that `extends` signals intent to _replace_ the original component
+    ## whereas `requires` signals that the originaly component(s) are being overlaid
+    ## with additional configuration.
+    requires: [ mattermost ]
+    charts:
+      - name: mattermost-enterprise-edition
+        valuesFiles:
+          - values/extra-plugins-values.yaml
```
Racer159, line 256: This part _could_ be interesting and is probably the main practical issue - optional components can't influence other components when they are selected - if a component could set a `chartOverride` map to be applied to later components that could be interesting - you would also need component name here as well to map those in correctly

adr/0025-composition.md
marshall007: I thought both local path and OCI imports were considered "skeleton packages". They are the same in all respects, correct?

adr/0025-composition.md
marshall007: Verbose relative to what? The proposed API allows you to avoid specifying Helm chart value overrides entirely while still declaring which images to include in the package.

> you'd have to match strings in a Helm post render which might not be reliable

This should be performed as structural edits on the YAML, not by string matching. Here are the field specs that `kustomize` looks at by default: https://github.com/kubernetes-sigs/kustomize/blob/c1de0301f5e71ffbecf77f5bc8687e7693878eb5/api/internal/konfig/builtinpluginconsts/images.go#L9-L16

adr/0025-composition.md
marshall007: For the purposes of this proposal, it doesn't matter whether we are doing a local docker build or not. This was just a self-contained way to demonstrate the need for an optional component to bundle both an additional image and related Helm chart value overrides.

adr/0025-composition.md
marshall007: The [merge strategies](https://docs.zarf.dev/ref/components/#merge-strategies) for "named primitive arrays" (like `charts` in this example) is already well-defined. Why would further extending the API with something like `chartOverride` be preferable?

> you would also need component name here as well to map those in correctly

The component name that you are extending in this example is defined by `requires: [ mattermost ]`.

adr/0025-composition.md
marshall007: It does conflict, I think any components that use `components[].requires` to declare dependencies on other named components within the package MUST be considered optional. Same with `components[].extends`.

adr/0025-composition.md
marshall007: I don't understand. Can you rephrase?

adr/0025-composition.md
phillebaba: How far are we planning on taking this? This basically means that we need to implement a DAG for the components to make sure that we don't end up with cyclical references.

marshall007: @phillebaba is that bad? What prevents cyclical path `imports` right now?

phillebaba: Nothing, but that does not mean that we should add to the existing complexity. My feeling is that we are trying to  partially implement Kustomize features in Zarf. Before we do that we should probably look at Kustomize and the challenges it has faced. Mixing declarative YAML with imperative functions is difficult and tends to become messy in complex situations. Which is why Helm is so popular even with all of its flaws. I am no fan of mixing Go templating with YAML but there is a reason people prefer that over Kustomize.

Noxsios: https://github.com/defenseunicorns/zarf/blob/014d1d22544c1b85f2444b9cce7584eac6a70ab2/src/pkg/packager/composer/list.go#L166-L174

adr/0025-composition.md
marshall007: I think I answered this below in https://github.com/defenseunicorns/zarf/pull/2690/files/27b8b7d9640e31c00b6f51904e2aba58fe682120#r1679778222

tl;dr: `zarf package create . --include [...<optional component>]`

schristoff:
Hey! Thank you so much for this context. It will be used to fuel the design in the future :)

Issue #2694, opened 2024-07-03
Refactor packager struct properties


phillebaba:
### Describe what should be investigated or refactored

While working on #2682 I realized that packager makes use of a lot of struct properties in place of local variables or function arguments. This made the work very difficult. So before we can clean up packager the use of all struct properties that are not thread safe has to be removed.

### Links to any relevant code
[(optional) i.e. - https://github.com/defenseunicorns/zarf/blob/main/README.md?plain=1#L1](https://github.com/defenseunicorns/zarf/blob/43e50bb0c34ed891eccf41ca379bd58c17a73f3e/src/pkg/packager/common.go#L38-L40)

### Additional context
N/A

brandtkeller:
Closing as duplicate

PR #2697, opened 2024-07-03
refactor: move lint logic from pacakger to command

phillebaba:
## Description

This moves the CLI specific lint logic out of the packager and into the command instead. This means that you do not need a packager instance to do the linting. Additionally this removes the use of os.Chdir for the linter.

Doing this reduces the complexity as the linter logic does not need to know about all other packager related logic which currently exists.

## Related Issue

Relates to #2694

## Checklist before merging


AustinAbro321:
@phillebaba This gets covered by #2585, albeit in a different way. The code moved into cmd here is instead moved to the lint package. I believe this should be closed, however if you want to get this merged in first or prefer having the logic in cmd rather than lint lmk.

phillebaba:
You are right, closing in favor of your current work.

PR #2698, opened 2024-07-03
refactor: sources and move pull logic out of packager

phillebaba:
## Description

This change refactors the use of package options in sources and moves the package pull logic out of the packager.

This change also fixes so that load package is an idempotent function which can be called multiple times.  

## Related Issue

Relates to #2694

## Checklist before merging


src/pkg/packager/sources/oci.go
```diff
@@ -29,15 +29,16 @@ var (
 
 // OCISource is a package source for OCI registries.
 type OCISource struct {
-	*types.ZarfPackageOptions
-	*zoci.Remote
+	Remote        *zoci.Remote
```
schristoff, line 32: Do all of these need to be exported?

src/pkg/packager/sources/oci.go
```diff
@@ -29,15 +29,16 @@ var (
...
+	Src           string
```
schristoff, line 33: Was `OCISource.Source` unappealing? Seems different than what you were doing above.

src/pkg/packager/sources/sources.go
```diff
@@ -41,51 +41,64 @@ type PackageSource interface {
 }
 
 // Identify returns the type of package source based on the provided package source string.
-func Identify(pkgSrc string) string {
-	if helpers.IsURL(pkgSrc) {
-		parsed, _ := url.Parse(pkgSrc)
+func Identify(src string) string {
+	if helpers.IsURL(src) {
+		parsed, _ := url.Parse(src)
```
schristoff, line 46: Can we 
```
func Identify(src string) (string, error) {
	if helpers.IsURL(src) {
		parsed, err := url.Parse(src)

```
to keep the error? It doesn't look like it's in too many places so it should be an easy change

src/pkg/packager/sources/sources.go
```diff
@@ -41,51 +41,64 @@ type PackageSource interface {
...
 	var source PackageSource
-
-	pkgSrc := pkgOpts.PackageSource
-
-	switch Identify(pkgSrc) {
+	switch Identify(src) {
 	case "oci":
-		if pkgOpts.Shasum != "" {
-			pkgSrc = fmt.Sprintf("%s@sha256:%s", pkgSrc, pkgOpts.Shasum)
+		ociSrc := src
```
schristoff, line 66: This seems like this is done for clarity but does not follow Go idioms - I would wager to remove.

src/pkg/packager/sources/split.go
```diff
@@ -28,12 +28,14 @@ var (
 
 // SplitTarballSource is a package source for split tarballs.
 type SplitTarballSource struct {
-	*types.ZarfPackageOptions
+	Src           string
```
schristoff, line 31: Same comment as above, do all of these need to be exported?

schristoff (changes requested):
Great job, just a few comments. Thanks for your work on this.

schristoff (approved):
Updating based off internal conversations.

PR #2703, opened 2024-07-09
refactor: connect command list printing

phillebaba:
## Description

This change moves the printing of connection list to the command. Additionally it adds test for listing connections.

## Related Issue

Part of #2576 

## Checklist before merging


src/cmd/connect.go
```diff
@@ -22,83 +22,80 @@ var (
...
+		}
 
-			defer tunnel.Close()
-			url := tunnel.FullURL()
+		ctx := cmd.Context()
 
-			// Dump the tunnel URL to the console for other tools to use.
-			fmt.Print(url)
+		var tunnel *cluster.Tunnel
+		if connectResourceName != "" {
```
schristoff-du, line 49: would swap from a `!= ""` to a `= ""` 
Thought there would be an `isEmpty()` in the strings lib, but there isn't :(

schristoff-du (approved):
Thanks for adding tests.

AustinAbro321 (approved):

PR #2720, opened 2024-07-12
fix: data injection to return errors

phillebaba:
## Description

This changes the data injection signature to defer parallelization to the caller, enabling it to return errors. It also replaces the wait group with a error group to handle the errors. 

## Related Issue

Relates to #2719
Relates to #2576

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/data.go
+++ b/src/pkg/cluster/data.go
@@ -13,7 +13,6 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -32,12 +31,10 @@ import (
 
 // HandleDataInjection waits for the target pod(s) to come up and inject the data into them
 // todo:  this currently requires kubectl but we should have enough k8s work to make this native now.
-func (c *Cluster) HandleDataInjection(ctx context.Context, wg *sync.WaitGroup, data types.ZarfDataInjection, componentPath *layout.ComponentPaths, dataIdx int) {
-	defer wg.Done()
+func (c *Cluster) HandleDataInjection(ctx context.Context, data types.ZarfDataInjection, componentPath *layout.ComponentPaths, dataIdx int) error {
 	injectionCompletionMarker := filepath.Join(componentPath.DataInjections, config.GetDataInjectionMarker())
 	if err := os.WriteFile(injectionCompletionMarker, []byte("🦄"), helpers.ReadWriteUser); err != nil {
-		message.WarnErrf(err, "Unable to create the data injection completion marker")
-		return
+		return fmt.Errorf("unable to create the data injection completion marker: %w", err)
 	}
 
 	tarCompressFlag := ""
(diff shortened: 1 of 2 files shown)


src/pkg/cluster/data.go
```diff
@@ -59,103 +56,110 @@ func (c *Cluster) HandleDataInjection(ctx context.Context, wg *sync.WaitGroup, d
...
-				kubectlCmd,
-				untarCmd,
-			)
-
-			if err := exec.CmdWithPrint(shell, append(shellArgs, cpPodCmd)...); err != nil {
-				message.Warnf("Error saving the zarf sync completion file after injection into pod %#v\n", pod.Name)
-				continue iterator
+				// Do the actual data injection
+				if err := exec.CmdWithPrint(shell, append(shellArgs, cpPodCmd)...); err != nil {
+					return fmt.Errorf("could not coy data ino the pod %s: %w", pod.Name, err)
```
lucasrod16, line 123: coy -> copy
ino -> into

src/pkg/cluster/data.go
```diff
@@ -59,103 +56,110 @@ func (c *Cluster) HandleDataInjection(ctx context.Context, wg *sync.WaitGroup, d
...
-		// Block one final time to make sure at least one pod has come up and injected the data
-		// Using only the pod as the final selector because we don't know what the container name will be
-		// Still using the init container filter to make sure we have the right running pod
-		_ = waitForPodsAndContainers(ctx, c.Clientset, podOnlyTarget, podFilterByInitContainer)
+			// Block one final time to make sure at least one pod has come up and injected the data
+			// Using only the pod as the final selector because we don't know what the container name will be
+			// Still using the init container filter to make sure we have the right running pod
+			_, err = waitForPodsAndContainers(ctx, c.Clientset, podOnlyTarget, podFilterByInitContainer)
+			if err != nil {
+				return nil
```
lucasrod16, line 151: Is there a reason we are returning `nil` here rather than the `err`?

src/pkg/packager/deploy.go
```diff
@@ -359,6 +356,10 @@ func (p *Packager) deployComponent(ctx context.Context, component types.ZarfComp
 		return charts, fmt.Errorf("unable to run component after action: %w", err)
 	}
 
+	err = g.Wait()
+	if err != nil {
+		return nil, err
+	}
```
lucasrod16, line 362: Is there a reason we are calling `g.Wait()` here rather than immediately after looping over DataInjections and calling `HandleDataInjection`?

src/pkg/cluster/data.go
phillebaba: I think my initial logic was that errors were not handled before so lets not do it now. That was stupid the more I think about it though.

src/pkg/packager/deploy.go
phillebaba: Mostly to reflect the logic which was present previously. The impact of doing it this way is that the actions may run during data injection is under way. Maybe we don't want this behavior at all?

AustinAbro321 (approved):

PR #2722, opened 2024-07-12
refactor: move setup CLI to only run once in root command

phillebaba:
## Description

This change refactors the setup CLI to only run once in the root command. Additionally it adds error returns and removes use of global configuration variables.

## Related Issue

Relates to #2576

## Checklist before merging

### DIFF
--- a/src/cmd/root.go
+++ b/src/cmd/root.go
@@ -26,20 +26,43 @@ import (
 var (
 	// Default global config for the packager
 	pkgConfig = types.PackagerConfig{}
+	// LogLevelCLI holds the log level as input from a command
+	LogLevelCLI string
+	// SkipLogFile is a flag to skip logging to a file
+	SkipLogFile bool
+	// NoColor is a flag to disable colors in output
+	NoColor bool
 )
 
 var rootCmd = &cobra.Command{
 	Use: "zarf COMMAND",
-	PersistentPreRun: func(cmd *cobra.Command, _ []string) {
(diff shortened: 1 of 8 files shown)


AustinAbro321 (changes requested):
Not sure where the error is but when I built this branch and run zarf package create there is no color

phillebaba:
Weird I tested it before and it worked, I may have changed something since then so will have a look.

phillebaba:
Turns out I had reverted a bool check that needed to be inverted.

src/cmd/root.go
```diff
@@ -26,20 +26,40 @@ import (
...
-			return
+			return nil
 		}
-		// Don't log the help command
+
+		skipLogFile := SkipLogFile
+
+		// Dont write tool commands to file.
+		comps := strings.Split(cmd.CommandPath(), " ")
+		if len(comps) > 1 && comps[1] == "tools" {
```
AustinAbro321, line 49: We should also check if the command is version, since we deleted the check PersistentPreRun from there

AustinAbro321 (changes requested):
One small fix then approved

src/cmd/root.go
phillebaba: I can add it but in reality it has no effect at all. Just setting it to false is not enough because use of global variables.

AustinAbro321 (approved):

PR #2724, opened 2024-07-12
chore: remove component webhooks

### LINKED ISSUE #2610: Remove component webhooks
Component web hooks are complex to use and the Zarf team is not aware any use case for them. We have advertised the feature as alpha https://docs.zarf.dev/ref/examples/component-webhooks/#_top. It should be removed before GA.

lucasrod16:
## Description
remove component webhooks

## Related Issue

Fixes #2610
Relates to #2562


phillebaba:
Code wise this looks good. Removing this cleans up  code if it is not used.

PR #2737, opened 2024-07-18
chore: add id token

AustinAbro321:
## Description

add id token which is necessary for aws creds to work now

## Checklist before merging


phillebaba (approved):

AustinAbro321:
fixed by #2739

PR #2739, opened 2024-07-19
chore: move public test repo

AustinAbro321:
## Description

old link was archived, this should be owned by the zarf-dev org anyway.

The numbers changing in this pr are crc hashes that are changing because the repo name changed. 

## Checklist before merging

### DIFF
--- a/src/test/packages/22-git-data/zarf.yaml
+++ b/src/test/packages/22-git-data/zarf.yaml
@@ -9,7 +9,7 @@ components:
     required: true
     repos:
       # Do a full Git Repo Mirror
-      - https://github.com/defenseunicorns/zarf-public-test.git
+      - https://github.com/zarf-dev/zarf-public-test.git
       # The following performs a full Git Repo Mirror forcing a fallback to host `git`
       - https://dev.azure.com/defenseunicorns/zarf-public-test/_git/zarf-public-test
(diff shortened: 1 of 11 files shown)


src/test/packages/22-git-data/zarf.yaml
```diff
@@ -9,19 +9,17 @@ components:
     required: true
     repos:
       # Do a full Git Repo Mirror
-      - https://github.com/defenseunicorns/zarf-public-test.git
+      - https://github.com/zarf-dev/zarf-public-test.git
       # The following performs a full Git Repo Mirror forcing a fallback to host `git`
       - https://dev.azure.com/defenseunicorns/zarf-public-test/_git/zarf-public-test
-      # Perform a full repo mirror of a simple repository with a single branch - (this causes an "already up to date" error in go-git)
-      - https://github.com/defenseunicorns/golang-tekton-hello-world.git
```
AustinAbro321, line 16: I'm guessing this was a bug we once ran into and added this to check? Feels like a weird thing to include in an e2e test. There is nothing tekton specific in the test or relating to this repo. I'm hesitant to add another repo here, as it doesn't look like in the code we are doing any special error handling of an, "already up to date", error and any repo we put here could change

AustinAbro321: Had to change it cause the repo was moved to private

phillebaba (approved):

PR #2746, opened 2024-07-23
chore: move context.TODO to context.Background() (2)

schristoff:
Removing `Context.TODO()` from `CmdWithCtx(...)` by creating a `CmdWithTesting(t *testing.T ....)` that ties the context to the test. Plumbed through a couple of testing helpers to also take a `testing.T` as the first argument in the function signature, as this practice reduces the chance of someone mistaking these testing helpers as core functionality.


phillebaba (commented):
For the most part this looks good. My only comment is going to be annoying. It looks like for exec we always expect there to be no error. If that is the case would it not be simpler to assert this in the actual function instead of doing the same check 100s of times?

schristoff:
@phillebaba  I'm not 100% solid on removing the ability to assess errors from the actual test (caller) itself, mainly because there may be times we are expecting a certain error in the long term.

phillebaba (approved):
Yeah I am fine with not checking the errors, we can always revisit this at a later time.

PR #2749, opened 2024-07-23
chore: move context.TODO to context.Background() (4)

schristoff:
Relies on #2746 & #2747 
⏰  `CosignVerifyBlob` is being imported by UDS-CLI - a PR to their codebase will need to be implemented during release.


schristoff:
PR title check is no fun

src/pkg/packager/actions/actions.go
```diff
@@ -130,7 +130,7 @@ retryCmd:
 		// If no timeout is set, run the command and return or continue retrying.
 		if actionDefaults.MaxTotalSeconds < 1 {
 			spinner.Updatef("Waiting for \"%s\" (no timeout)", cmdEscaped)
-			if err := tryCmd(context.TODO()); err != nil {
+			if err := tryCmd(context.Background()); err != nil {
```
schristoff, line 133: Passing in the main `ctx` I think is causing flakes because it's being cancelled?

schristoff: Oh, it needs cancelled.

phillebaba (approved):

Issue #2748, opened 2024-07-23
test releases against UDS-CLI


schristoff:
Releases should test integration with UDS-CLI (more to be assessed later) - this should be implemented by next release.

phillebaba:
Does this not flip the dependency on its head? Might just be me but I have not built a project where the implementer tests downstream dependents. Could this not also be achieved through getting to a point where the exported SDK is stable and what we export is a conscious decision.

PR #2750, opened 2024-07-23
chore: move context.TODO to context.Background() (5)

schristoff:
Last one! 
Relies on #2746, #2747, #2748 

Implements a global "testing util" under test/helper - open to naming ideas or different implementations. This allows for `TestContext(t *testing.T) context{}` to be used by any test, and overtime we can place more things in here.

### DIFF
--- a/src/test/helpers/helpers.go
+++ b/src/test/helpers/helpers.go
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2021-Present The Zarf Authors
 
-// Package provides global testing helper functions
+// Package helpers provides global testing helper functions
 package helpers
 
 import (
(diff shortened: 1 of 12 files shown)


src/test/helpers/helpers.go
```diff
@@ -0,0 +1,18 @@
...
+package helpers
+
+import (
+	"context"
+	"testing"
+)
+
+// TestContext takes a testing.T and returns a context that is
+// attached to the test by t.Cleanup()
+func TestContext(t *testing.T) context.Context {
```
phillebaba, line 14: Good idea, this will be a great place to add out test logger in the future which only output logs for failed tests.

schristoff: Can I'll change pkg name to testutil

phillebaba (approved):
LGTM

PR #2753, opened 2024-07-23
refactor: add error to packager set paths

phillebaba:
## Description

This change removes the debug log and replaces it with an error  when a package path is of unknown format.

## Related Issue

Relates to #2576 

## Checklist before merging


phillebaba:
@AustinAbro321 I did this PR to see what you think first. Personally I do not like to silently error. If we are going to do so we should probably make this more apparent and explain why we fail silently.

phillebaba:
I was wrong about this change. The unmatched cases are ignored for example when the path is a manifest. I still feel like we should test for that and do nothing to catch any future problem however.

PR #2755, opened 2024-07-23
feat: annotate image mutation

### LINKED ISSUE #2754: Add an annotation / label to Pods with their old image reference when mutating
In some cases, like controllers/operators etc, we might not know all the dependencies as far as images, even after using the provided tools like `zarf dev find-images`. In our case, we found images that were dependencies but what what repo we should be point to. The initial attempts were pointing at a repo that didnt match what the pod expected, but we couldnt see that due to the mutating webhook effect on the pods spec.

Adding a label or annotation to the pods which would state the original image repo, version, and name that the pods would be looking for before the mutating webhook changed it to point to the internal image repo would be useful. This would make it extremely easy to find what images are missing and add those missing images to their zarf.yaml. 

Discussed in your discord, I originally thought there might be a command to reverse the UID, but its a one way hash so options are limited. Alternate option which I will add here for historical value, was given in the discord which is helpful:
"You can check if an image name you have matches the hash through the command `zarf internal crc32 my_image` at least"

AustinAbro321:
## Description
We have received many questions from the community on how the CRC hash works, usually they ask this because they are unsure which image was supposed to be deployed before Zarf mutated the pod. This adds an annotation so it can be easily checked. 

## Related Issue

Fixes #2754 

## Checklist before merging

### DIFF
--- a/src/internal/agent/hooks/pods.go
+++ b/src/internal/agent/hooks/pods.go
@@ -79,7 +79,7 @@ func mutatePod(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster.Clu
 			message.Warnf(lang.AgentErrImageSwap, container.Image)
 			continue // Continue, because we might as well attempt to mutate the other containers for this pod
 		}
-		updatedAnnotations[fmt.Sprintf("zarf.dev/original-init-image[%d]", idx)] = container.Image
+		updatedAnnotations[fmt.Sprintf("zarf.dev/original-init-image-%d", idx)] = container.Image
 		patches = append(patches, operations.ReplacePatchOperation(path, replacement))
 	}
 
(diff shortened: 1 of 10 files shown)


phillebaba:
Elegant and simple solution. I have looked around trying to find an example of some other project doing something similar but have had no luck so far. If I can't find any good example soon I think we should be fine to merge.

src/internal/agent/hooks/pods.go
```diff
@@ -74,6 +79,7 @@ func mutatePod(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster.Clu
 			message.Warnf(lang.AgentErrImageSwap, container.Image)
 			continue // Continue, because we might as well attempt to mutate the other containers for this pod
 		}
+		updatedAnnotations[fmt.Sprintf("zarf.dev/original-init-image-%d", idx)] = container.Image
```
phillebaba, line 82: Lets make the annotation prefix a constant, maybe even wrap it in a function. Could we also change it from the index to the name of the container?

phillebaba (approved):

PR #2760, opened 2024-07-25
feat: expose root cmd

Noxsios:
## Description

Closes https://github.com/zarf-dev/zarf/issues/2765

Expose `rootCmd` so that Zarf can be properly embedded in other CLIs.

## Checklist before merging


schristoff:
We aren't sure this is the path forward we want to take at this time. We are going to close this and if we decide this is the action we want to take this will be reopened and merged.

PR #2763, opened 2024-07-25
refactor: load state to return error if loading fails

phillebaba:
## Description

This change refactors places where we are ignoring the load Zarf state error or just logging it instead of returning.

## Related Issue

Relates to #2576 

## Checklist before merging

### DIFF
--- a/src/cmd/destroy.go
+++ b/src/cmd/destroy.go
@@ -44,7 +44,7 @@ var destroyCmd = &cobra.Command{
 		//       the scripts to remove k3s, we will still try to remove a locally installed k3s cluster
 		state, err := c.LoadZarfState(ctx)
 		if err != nil {
-			message.WarnErr(err, lang.ErrLoadState)
+			message.WarnErr(err, err.Error())
 		}
 
 		// If Zarf deployed the cluster, burn it all down
--- a/src/pkg/cluster/state.go
+++ b/src/pkg/cluster/state.go
@@ -7,11 +7,11 @@ package cluster
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"slices"
 	"time"
 
-	"github.com/fatih/color"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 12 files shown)


src/pkg/cluster/state.go
```diff
@@ -210,13 +210,14 @@ func (c *Cluster) InitZarfState(ctx context.Context, initOptions types.ZarfInitO
 
 // LoadZarfState returns the current zarf/zarf-state secret data or an empty ZarfState.
 func (c *Cluster) LoadZarfState(ctx context.Context) (state *types.ZarfState, err error) {
+	stateErr := errors.New("failed to load the Zarf State from the cluster, has Zarf been initiated?")
 	secret, err := c.Clientset.CoreV1().Secrets(ZarfNamespaceName).Get(ctx, ZarfStateSecretName, metav1.GetOptions{})
 	if err != nil {
-		return nil, fmt.Errorf("%w. %s", err, message.ColorWrap("Did you remember to zarf init?", color.Bold))
+		return nil, errors.Join(stateErr, err)
```
AustinAbro321, line 216: nit but I would still use fmt.Errorf. Newlines look a bit awkward when we print to the user and I think it reads better for the calls to action to be at the end of the error. 
image

src/pkg/packager/deploy.go
```diff
@@ -453,10 +455,11 @@ func (p *Packager) setupState(ctx context.Context) (err error) {
 	defer spinner.Stop()
 
 	state, err := p.cluster.LoadZarfState(ctx)
-	// Return on error if we are not in YOLO mode
+	// We ignore the error if in YOLO mode because Zarf should not be initiated.
 	if err != nil && !p.cfg.Pkg.Metadata.YOLO {
```
AustinAbro321, line 459: IMO we should only ignore the error in yolo mode if the error is `kerrors.IsNotFound(err)`

AustinAbro321 (changes requested):

src/pkg/cluster/state.go
phillebaba: Will do that for now and look back at formatting errors in the future.

AustinAbro321 (approved):

src/pkg/packager/deploy.go
```diff
@@ -714,21 +721,26 @@ func (p *Packager) installChartAndManifests(ctx context.Context, componentPaths
...
+		return nil
+	}
+	// Grab a fresh copy of the state to print the most up-to-date version of the creds
+	latestState, err := p.cluster.LoadZarfState(ctx)
+	if err != nil {
+		return err
 	}
+	if latestState == nil {
+		latestState = p.state
+	}
```
AustinAbro321, line 741: I think we can delete this. We only had it because we kept going even if we errored.

AustinAbro321 (changes requested):

src/pkg/packager/deploy.go
AustinAbro321: Yeah here we set the latestState if it's nil, but it's only going to be nil if it error'd and now we return on error.

src/cmd/destroy.go
```diff
@@ -44,7 +44,7 @@ var destroyCmd = &cobra.Command{
 		//       the scripts to remove k3s, we will still try to remove a locally installed k3s cluster
 		state, err := c.LoadZarfState(ctx)
 		if err != nil {
-			message.WarnErr(err, lang.ErrLoadState)
```
schristoff, line 47: Why removing `lang.ErrorLoadState`,

phillebaba: Because I have moved this to a error message that is returned from the LoadZarfState function. That way we dont have to declare it everywhere.

AustinAbro321 (approved):

PR #2778, opened 2024-07-29
fix: stop mutating zarf state to reflect a git server when there is none

AustinAbro321:
## Description

...

## Related Issue

Fixes #

Relates to #

## Checklist before merging


AustinAbro321:
packager is being entirely refactored, this change should happen there

PR #2781, opened 2024-07-29
refactor: utilize invopop features better

AustinAbro321:
## Description

Relates to #

## Checklist before merging

### DIFF
--- a/src/cmd/internal.go
+++ b/src/cmd/internal.go
@@ -162,6 +162,10 @@ var genConfigSchemaCmd = &cobra.Command{
 	Short:   lang.CmdInternalConfigSchemaShort,
 	RunE: func(_ *cobra.Command, _ []string) error {
 		reflector := jsonschema.Reflector(jsonschema.Reflector{ExpandedStruct: true})
+		typePackagePath := filepath.Join("src", "types")
+		if err := reflector.AddGoComments("github.com/zarf-dev/zarf", typePackagePath); err != nil {
+			return fmt.Errorf("this command must be called from the root of the Zarf repo: %w", err)
+		}
 		schema := reflector.Reflect(&types.ZarfPackage{})
 		output, err := json.MarshalIndent(schema, "", "  ")
 		if err != nil {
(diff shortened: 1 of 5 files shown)


zarf.schema.json
```diff
@@ -1060,18 +1084,19 @@
         },
         "vendor": {
           "type": "string",
-          "description": "Name of the distributing entity, organization or individual."
```
phillebaba, line 1063: Missing punctuation

zarf.schema.json
```diff
@@ -1060,18 +1084,19 @@
...
+          "description": "Name of the distributing entity, organization or individual"
         },
         "aggregateChecksum": {
           "type": "string",
-          "description": "Checksum of a checksums.txt file that contains checksums all the layers within the package."
```
phillebaba, line 1067: Ditto

phillebaba (changes requested):

schristoff (dismissed):

zarf.schema.json
AustinAbro321: I just went ahead and added punctuation to all the json schema descriptions

src/cmd/internal.go
```diff
@@ -156,12 +157,30 @@ tableOfContents: false
...
+	addCommentErr := errors.New("this command must be called from the root of the Zarf repo")
+
+	typePackagePath := filepath.Join("src", "types")
+	if err := reflector.AddGoComments("github.com/zarf-dev/zarf", typePackagePath); err != nil {
+		return fmt.Errorf("%w: %w", addCommentErr, err)
+	}
+	varPackagePath := filepath.Join("src", "pkg", "variables")
+	if err := reflector.AddGoComments("github.com/zarf-dev/zarf", varPackagePath); err != nil {
+		return fmt.Errorf("%w: %w", addCommentErr, err)
+	}
```
phillebaba, line 170: I am always a bit worried when I see use of reflections. Could we get an explanation what this is actually doing?

AustinAbro321: I'm pretty sure it's literally parsing the file rather than using reflections. https://github.com/invopop/jsonschema?tab=readme-ov-file#using-go-comments

phillebaba (approved):

PR #2784, opened 2024-07-29
fix: detect invalid helm release names

jamestexas:
## Description

This pull request adds validation for Helm release names in the Zarf project. A malformed `releaseName` in a `zarf.yaml` file would previously pass checks such as `zarf package create` and `zarf dev find-images .`. This change introduces validation logic to ensure that Helm release names comply with [DNS-1035 standards](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names), preventing issues during deployment.

Note: The `zarf dev lint .` command was not addressed in this PR as the validation logic didn't seem to get applied there.  I am also new to contributing here so please let me know if I missed anything!

## Related Issue

Fixes https://github.com/zarf-dev/zarf/issues/2776
image

## Checklist before merging


src/config/lang/english.go
```diff
@@ -645,6 +645,8 @@ const (
 	PkgValidateErrManifestNameLength      = "manifest %q exceed the maximum length of %d characters"
 	PkgValidateErrManifestNameNotUnique   = "manifest name %q is not unique"
 	PkgValidateErrPkgConstantPattern      = "provided value for constant %q does not match pattern %q"
+	PkgValidateErrChartReleaseNameEmpty   = "release name empty, unable to fallback to chart name"
+	PkgValidateErrChartReleaseNameInvalid = "invalid release name '%s': a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character"
```
phillebaba, line 649: We are trying to phase out use of the language package so just put the error message in the code. Also don't use quotes in your error messages if you do not really have to.

src/types/validate.go
```diff
@@ -252,6 +254,19 @@ func (action ZarfComponentAction) Validate() error {
 	return err
 }
 
+// ValidateReleaseName validates against DNS 1035 kubernetes label spec
+// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names
+// https://github.com/zarf-dev/zarf/issues/2776
```
phillebaba, line 259: Don't need to link the issue as that will be tracked in git anyways.

src/types/validate.go
```diff
@@ -277,6 +292,19 @@ func (chart ZarfChart) Validate() error {
...
+	// Validate releaseName if provided, else fallback to chart.Name
+	// This also ensures that the release name is a valid Helm release name
+	// Lastly this same fallback mechanism happens in src/internal/packager/helm/chart.go:InstallOrUpgradeChart
+	releaseName := chart.ReleaseName
+	if releaseName == "" {
+		releaseName = chart.Name
+	}
+	// Use Helm's ValidateReleaseName function for release name validation
+	if nameErr := ValidateReleaseName(releaseName); nameErr != nil {
+		log.Printf("Error validating ZarfChart ReleaseName: %s", releaseName)
```
phillebaba, line 304: Remove the log line.

phillebaba (changes requested):
Some small NIT comments otherwise it looks good. Thank you for adding tests as part of the change :rose:

src/config/lang/english.go
jamestexas: Noted! I assumed this would be fine as a private constant since it's not exposed outside the package. I've updated the error messages to avoid using quotes unless necessary, and moved them directly into the code as recommended. Thanks for the guidance!

src/types/validate.go
```diff
@@ -252,6 +255,18 @@ func (action ZarfComponentAction) Validate() error {
 	return err
 }
 
+// ValidateReleaseName validates against DNS 1035 kubernetes label spec
+// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names
+func ValidateReleaseName(releaseName string) error {
```
schristoff, line 260: Does this need to be exported?

jamestexas: It probably doesn't. I was halfway thinking it didn't need to be a method at all be erred on the side of being more explicit. I'll make it private (unless you have opinions about it being just part of the `Validate` method instead since it's simple conditionals.

src/types/validate.go
```diff
@@ -277,6 +292,18 @@ func (chart ZarfChart) Validate() error {
 		err = errors.Join(err, fmt.Errorf(lang.PkgValidateErrChartVersion, chart.Name))
 	}
 
+	// Validate releaseName if provided, else fallback to chart.Name
+	// This also ensures that the release name is a valid Helm release name
+	// Lastly this same fallback mechanism happens in src/internal/packager/helm/chart.go:InstallOrUpgradeChart
+	releaseName := chart.ReleaseName
```
schristoff, line 298: Is it possible to move this logic into the `validateReleaseName` function?

schristoff:
This is a great first contribution to the Zarf repository! Thank you so much :)

src/types/validate.go
AustinAbro321: IMO just include this in the `func (chart ZarfChart) Validate()` method

AustinAbro321: better yet just use the helm [ValidateReleaseName function](https://github.com/helm/helm/blob/a958f3a9b65df7c809c88c180cc89886fee7768c/pkg/chartutil/validate_name.go#L79)

src/types/validate.go
jamestexas: I assumed there wasn't any qualms with using named returns here, but let me know if you'd rather have the `err` var explicitly declared. https://github.com/zarf-dev/zarf/commit/2658ff43b0a040fca86e34dbe2e91388ffaa2a8c

src/types/validate.go
jamestexas: @AustinAbro321 Thanks for your suggestion! I actually decided not to use the Helm validation method due to how Kubernetes utilizes the releaseName later on. The distinction is subtle but important:

1. Helm's regex allows for periods `.` in the release name.
2. However, Kubernetes applies stricter rules for certain resources (like Services) using the DNS-1035 label standard.
3. This stricter standard doesn't allow periods and requires names to start with a letter. 

That being said, after considering this further, we could use what's in the [Kubernetes API](https://github.com/kubernetes/kubernetes/blob/974f3d3d8ff6bfb33a375f7207c34c69b3e8b932/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L235-L244) if it's preferred?

To summarize, my understanding is that the constraints in Helm are more lenient than in Kubernetes. Because Zarf abstracts away some of the configuration (a service receiving the releaseName in this case), I think it would make sense for us to be more strict in the Zarf package to prevent potential issues downstream.

What do you think about this approach? I'm open to discussing further if you have any concerns or alternative suggestions.

jamestexas: It was really easy to add this, so I did. But in order to maintain the original way tests functioned, while also accounting for the error string that the k8s api yields (which is rather length) it added a small amount of conditionals to the existing tests. If that is a problem let me know, and I'll be happy to tweak further.

phillebaba (approved):

PR #2787, opened 2024-07-30
refactor: remove magic strings in struct tags

### LINKED ISSUE #2786: Refactor the types to clean up json / jsonschema tags
Currently the types package is messy to read because of all the jsonschema comments. We can utilize features in the invopop/jsonschema to project to auto add go comments as descriptions to the schema, cleaning up the types package. Additionally we can use constants for things like enums, defaults, and patterns which leaves less magic strings in the code.

AustinAbro321:
## Description

Fixes #2786

This should be merged after #2781 

## Checklist before merging


src/pkg/variables/types.go
```diff
@@ -21,24 +22,31 @@ const (
...
 	// An optional regex pattern that a variable value must match before a package deployment can continue.
 	Pattern string `json:"pattern,omitempty"`
 	// Changes the handling of a variable to load contents differently (i.e. from a file rather than as a raw variable - templated files should be kept below 1 MiB)
-	Type VariableType `json:"type,omitempty" jsonschema:"enum=raw,enum=file"`
+	Type VariableType `json:"type,omitempty"`
+}
+
+// JSONSchemaExtend extends the generated json schema during `zarf internal gen-config-schema`
+func (Variable) JSONSchemaExtend(schema *jsonschema.Schema) {
+	kind, _ := schema.Properties.Get("type")
```
phillebaba, line 45: Is this ignoring an error :thinking:

AustinAbro321: It's ignoring the present flag but I'd rather have the code panic if it doesn't find something since this is only run on `zarf internal gen-config-schema`

AustinAbro321: I can't change the signature to return an error, since this signature is necessary for this function to get called by the jsonschema package during schema creation

schristoff: We cannot be a production grade tool that panics.

AustinAbro321: This function is only called during schema generation, `zarf internal gen-config-schema`, and this would only panic if, during development, someone changed the json tag to not equal what is in the `schema.Properties.Get` call. 

Given that we don't have control over this function signature I think there's two options: 
1. return when the key isn't found - IMO this is worse because there won't be any signal from the CLI that something went wrong during schema generation. I'm fine with this though, it would happen rarely & it should be relatively obvious from the git diff that something went wrong. 
2. Keep panicing, maybe explicitly panic with a clear error message.

AustinAbro321:
Closing this as we decided we don't like the fact that it's possible for jsonschemextend to panic

Issue #2709, opened 2024-07-10
Untangle package imports


phillebaba:
### Describe what should be investigated or refactored

The current package structure causes a lot of cyclical import errors when the smallest change is made. From my perspective this is caused by keeping data types and logic in separate packages while also creating very large packages that do multiple things. Ownership of data types is not clear causing logic to be spit into separate packages.

The main culprit is the types package which contains a lot of data types related to `ZarfState` but also option parameters for the packager. There is value in defining structs that should have a stable API but these should not be mixed with other data types. We need to better understand which types can be updated because they are only used in memory and which need to be static because they are encoded and stored in files.

At the same time we need to understand which packages we are exporting for use as a library and which packages are for internal use only. To resolve future issues we will need to keep the exported packages stable. Right now we have not documented what the plan is at all.

A few suggestions for what we should do.

```[tasklist]
### Tasks
- [ ] Move user facing structs to a versioned api package.
- [ ] Move command line options structs out of types.
- [ ] Move data types used in single packages to their specific packages.
- [ ] Split large packages into smaller single purpose packages to reduce risk of circular imports.
```

### Links to any relevant code

https://github.com/defenseunicorns/zarf/blob/main/src/types/packager.go
https://github.com/defenseunicorns/zarf/blob/main/src/types/runtime.go

### Additional context

N/A

AustinAbro321:
Closing this. We've moved types closer to where they're being used / in a versioned package. The only struct that remains in types is global config ZarfCommonOptions struct https://github.com/zarf-dev/zarf/blob/main/src/types/runtime.go

PR #2835, opened 2024-08-05
refactor: move package generation to a local variable

phillebaba:
## Description

This change moves generation from a struct property in packager to the only place where it is needed.

## Related Issue

Relates to #2709 

## Checklist before merging

### DIFF
--- a/src/pkg/packager/deploy.go
+++ b/src/pkg/packager/deploy.go
@@ -138,20 +138,10 @@ func (p *Packager) Deploy(ctx context.Context) error {
 
 // deployComponents loops through a list of ZarfComponents and deploys them.
 func (p *Packager) deployComponents(ctx context.Context) (deployedComponents []types.DeployedComponent, err error) {
-	// Check if this package has been deployed before and grab relevant information about already deployed components
-	if p.generation == 0 {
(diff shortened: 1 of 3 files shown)


src/pkg/packager/deploy.go
```diff
@@ -138,20 +138,10 @@ func (p *Packager) Deploy(ctx context.Context) error {
...
-		deployedComponent := types.DeployedComponent{
-			Name:               component.Name,
-			Status:             types.ComponentStatusDeploying,
-			ObservedGeneration: p.generation,
-		}
-
-		// If this component requires a cluster, connect to one
+		// Connect to cluster if a component requires it.
+		packageGeneration := 1
 		if component.RequiresCluster() {
```
phillebaba, line 145: Took me some time to realize but we a have a fun little chicken and egg problem here. As we use an init package to setup a k3s cluster in some cases we cant just assume that the cluster we need to connect to exists if one of the components requires cluster access.

Instead we have to execute all of the component actions until we get to a point where the cluster is up and running and we can connect to it.

AustinAbro321 (approved):
We are really untangling a web here, nice work

PR #2838, opened 2024-08-05
fix: ignore config file not found errors

AustinAbro321:
## Description

Currently file not found errors are not being ignored because viper returns value errors and we're checking for pointer errors

## Checklist before merging


src/cmd/common/viper.go
```diff
@@ -168,8 +168,11 @@ func printViperConfigUsed() {
 		return
 	}
 	// Optional, so ignore file not found errors
```
schristoff, line 170: Remove comment

schristoff (approved):

PR #2843, opened 2024-08-06
refactor: proxy and add tests

phillebaba:
## Description

This change cleans up logging and adds tests to the proxy.

## Related Issue

Relates to #2576 

## Checklist before merging


AustinAbro321 (approved):

src/cmd/internal.go
```diff
@@ -51,7 +55,11 @@ var httpProxyCmd = &cobra.Command{
 	Short: lang.CmdInternalProxyShort,
 	Long:  lang.CmdInternalProxyLong,
 	RunE: func(cmd *cobra.Command, _ []string) error {
-		return agent.StartHTTPProxy(cmd.Context())
+		cluster, err := cluster.NewCluster()
```
schristoff, line 58: This path didn't previously have a `cluster.NewCluster`, so beforehand if there was none it would error?

src/internal/agent/start.go
```diff
@@ -28,20 +33,48 @@ const (
 )
 
 // StartWebhook launches the Zarf agent mutating webhook in the cluster.
-func StartWebhook(ctx context.Context) error {
-	srv, err := agentHttp.NewAdmissionServer(ctx, httpPort)
-	if err != nil {
-		return err
-	}
-	return startServer(ctx, srv)
+func StartWebhook(ctx context.Context, cluster *cluster.Cluster) error {
```
schristoff, line 36: It could be pretty easy to add tests for this with an test http client.

schristoff (approved):
Great work :)

src/cmd/internal.go
phillebaba: The cluster was just created further up the  chain. This makes things more clear and easier to test.

src/internal/agent/start.go
phillebaba: I agree, I am going to revisit this package in the future to get more tests in.

Issue #2827, opened 2024-08-02
Replace custom retry logic with a single retry function


phillebaba:
### Describe what should be investigated or refactored
We have a couple of ways to implement retry logic in Zarf. The most common are use of timers or the retry function form the helpers package. The former respects contexts while the latter does not. It would be better if we could decide on a single retry package to use in this function to simplify things.

A suggestion is to switch to using retry-go. This project has been around for a while and is pretty stable.
https://github.com/avast/retry-go

### Links to any relevant code
https://github.com/zarf-dev/zarf/blob/5e77e3816051973c3b5ec3d48be30865b40db5e1/src/pkg/cluster/cluster.go#L82-L120
https://github.com/zarf-dev/zarf/blob/5e77e3816051973c3b5ec3d48be30865b40db5e1/src/internal/packager/images/push.go#L57
https://github.com/zarf-dev/zarf/blob/5e77e3816051973c3b5ec3d48be30865b40db5e1/src/pkg/cluster/namespace.go#L30-L46
https://github.com/zarf-dev/zarf/blob/5e77e3816051973c3b5ec3d48be30865b40db5e1/src/pkg/cluster/state.go#L132-L154

### Additional context
N/A

AustinAbro321:
Small note that there is a `helpers.RetryWithContext` function we use, though from a glance at the retry-go package it seems like a reasonable substitute for both. 
https://github.com/zarf-dev/zarf/blob/5e77e3816051973c3b5ec3d48be30865b40db5e1/src/internal/packager/images/pull.go#L243

phillebaba:
Was not aware of that function, that would also be an option. We just need to look at the pros and cons of bringing in a dependency to do this.

salaxander:
#2027

PR #2856, opened 2024-08-08
fix: replace helpers.Retry with go-retry and adjust delay

phillebaba:
## Description

This change replaces the different retry implementations with only using go-retry.

## Related Issue

Part of #2827  
Relates to #2576 

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/tunnel.go
+++ b/src/pkg/cluster/tunnel.go
@@ -13,7 +13,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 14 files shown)


src/pkg/cluster/tunnel.go
```diff
@@ -352,38 +351,16 @@ func (tunnel *Tunnel) Wrap(function func() error) error {
...
-			}
-
-			timer.Reset(time.Duration(delay) * time.Second)
+	url, err := retry.DoWithData(func() (string, error) {
+		url, err := tunnel.establish(ctx)
+		if err != nil {
+			return "", err
 		}
+		return url, nil
+	}, retry.Context(ctx), retry.Attempts(3))
```
AustinAbro321, line 360: Not sure if it's important, but the amount of delay here is changing significantly. There was about 30 seconds total delay before, now there is about 300 milliseconds total delay.

phillebaba: Will increase the amount of attempts instead to fix this.

schristoff (approved):

PR #2867, opened 2024-08-12
refactor: remove unnecessary retry logic from data injection

phillebaba:
## Description

This change removes the retrying in data injections, as it does not have any effect. The current implementation will only retry at a single condition which is when the pods returned from the wait function is empty.

https://github.com/zarf-dev/zarf/blob/0943d76b25464ab7ad948b4c93957b22d966c02b/src/pkg/cluster/data.go#L88-L90

This event wont actually happen because the wait logic will run until the context cancelled if the pod list is empty. This means that returning an empty list is impossible.

Removing the nested retry implementation increases legibility. 

## Related Issue

Depends on #2856

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/data.go
+++ b/src/pkg/cluster/data.go
@@ -60,109 +60,100 @@ func (c *Cluster) HandleDataInjection(ctx context.Context, data v1alpha1.ZarfDat
 		return fmt.Errorf("unable to execute tar, ensure it is installed in the $PATH: %w", err)
 	}
 
-	// TODO: Refactor to use retry.
-	for {
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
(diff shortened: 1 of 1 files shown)


AustinAbro321 (approved):

schristoff (changes requested):
tests please

phillebaba:
The data injection can not be tested in its current state as it is shelling out to both tar and kubectl. What exactly should be tested here maybe other than `waitForPodsAndContainers` ?

schristoff:
Yeah, upon re-review it will be too burdensome to try and write functional tests for `HandleDataInjection` without a major refactor of the functionality itself.

schristoff (approved):

PR #2868, opened 2024-08-12
refactor: action retry to use go-retry

phillebaba:
## Description

This change replaces the custom retry logic with using go-retry. It simplifies legibility a lot by removing use of labels to break out of the outer loop.

## Related Issue

Depends on #2856 

## Checklist before merging

### DIFF
--- a/src/pkg/packager/actions/actions.go
+++ b/src/pkg/packager/actions/actions.go
@@ -93,6 +93,7 @@ func runAction(ctx context.Context, defaultCfg v1alpha1.ZarfComponentActionDefau
 	timeout := time.After(duration)
 
 	// Keep trying until the max retries is reached.
+	// TODO: Refactor using go-retry
 retryCmd:
 	for remaining := actionDefaults.MaxRetries + 1; remaining > 0; remaining-- {
 		// Perform the action run.
--- a/src/pkg/packager/actions/actions.go
+++ b/src/pkg/packager/actions/actions.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/avast/retry-go/v4"
 	"github.com/defenseunicorns/pkg/helpers/v2"
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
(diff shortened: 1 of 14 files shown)


src/pkg/packager/actions/actions.go
```diff
@@ -89,82 +90,55 @@ func runAction(ctx context.Context, defaultCfg v1alpha1.ZarfComponentActionDefau
...
+			if actionDefaults.MaxTotalSeconds > 0 {
+				timeoutStr = fmt.Sprintf("(timeout: %ds)", actionDefaults.MaxTotalSeconds)
 			}
-
-			return nil
-		}
+			spinner.Updatef("Waiting for \"%s\" %s", cmdEscaped, timeoutStr)
+		}),
+	)
+	if actionCtx.Err() != nil {
```
phillebaba, line 135: This will also happen on context cancel so need another way of checking for timeout. An option is to use with cause.

src/pkg/packager/actions/actions.go
```diff
@@ -89,82 +90,55 @@ func runAction(ctx context.Context, defaultCfg v1alpha1.ZarfComponentActionDefau
...
 
-	// Keep trying until the max retries is reached.
-retryCmd:
-	for remaining := actionDefaults.MaxRetries + 1; remaining > 0; remaining-- {
-		// Perform the action run.
-		tryCmd := func(ctx context.Context) error {
-			// Try running the command and continue the retry loop if it fails.
-			if out, err = actionRun(ctx, actionDefaults, cmd, actionDefaults.Shell, spinner); err != nil {
+		// If an output variable is defined, set it.
+		for _, v := range action.SetVariables {
```
phillebaba, line 107: Should this be retried? It it seems like this was done because implementing it would have been difficult in the old code. It seems a lot smarter to only retry the action on errors and not the setting on variables as I assume the output variables will be the same every time the actions run.

phillebaba:
Before merging we should add tests to the Run function to validate the error output for the different failure modes.

PR #2869, opened 2024-08-12
test: docker exec to in memory registry

### LINKED ISSUE #2837: test flake: failed to publish package during nightly eks
We have seen a flake in the [Test EKS cluster workflow](https://github.com/zarf-dev/zarf/blob/main/.github/workflows/nightly-eks.yml) runs we get a failure that looks like this. It is strange that we have only seen this error in the test eks workflow, as it is happening during [50_oci_publish_deploy_test.go](https://github.com/zarf-dev/zarf/blob/main/src/test/e2e/50_oci_publish_deploy_test.go) which doesn't test anything in the cluster

```
failed to publish package: Post "http://localhost:555/v2/helm-charts/blobs/uploads/": read tcp [::1]:34830->[::1]:555: read: connection reset by peer
```

### Full workflow logs
https://github.com/zarf-dev/zarf/actions/runs/10234319361/job/28313642984

AustinAbro321:
## Description

This switches to an in-memory registry for testing rather than a docker container which we have seen flake in actions

Fixes #2837

## Checklist before merging


go.mod
```diff
@@ -35,6 +36,7 @@ require (
 	github.com/mholt/archiver/v3 v3.5.1
 	github.com/moby/moby v24.0.9+incompatible
 	github.com/opencontainers/image-spec v1.1.0
+	github.com/phayes/freeport v0.0.0-20220201140144-74d24b5ae9f5
```
schristoff, line 39: Not super in love with adding this repo that hasn't been maintained, is there anything else out there that we could use?

schristoff (dismissed):
See if you can swap out that dep, otherwise g2g

go.mod
AustinAbro321: I just switched to hardcoding the ports. It's unlikely we'll see conflicts

schristoff (approved):

PR #2886, opened 2024-08-14
chore: separate schema generation

### LINKED ISSUE #2788: Move JSON schema generation
Currently schema generation is done as part of an internal function. This adds the dependency to generate the schema as part of the Zarf code base. As schema generation is not needed during runtime we should move this code to be a separate component instead.

#2781 

### Additional context
N/A

AustinAbro321:
## Description

This moves schema generation to it's own folder rather than being a command in Zarf

This doesn't remove the invopop/jsonschema from the Zarf CLI, however there is only one place left it is used. Likely this can be removed in a future PR and replaced with runtime validation.  
https://github.com/zarf-dev/zarf/blob/8100c698d92dd1e3fed450d548b2fea1d08cf32e/src/api/v1beta1/component.go#L318

Relates to #2788

## Checklist before merging


src/cmd/internal.go
```diff
@@ -166,66 +161,6 @@ tableOfContents: false
...
-}
-
-type zarfTypes struct {
-	DeployedPackage types.DeployedPackage
-	ZarfPackage     v1alpha1.ZarfPackage
-	ZarfState       types.ZarfState
-}
-
-var genTypesSchemaCmd = &cobra.Command{
-	Use:     "gen-types-schema",
```
AustinAbro321, line 210: I didn't build a replacement for this as I don't think it's needed. I'm not sure what the original use case was for adding this.

phillebaba:
Good jobs getting this done. A couple of comments.

Is the docs changes related to these changes or are we fixing this because we found out it was wrong during this work?

We wont remove any dependency in the Zarf CLI in this change. Just making sure because the description makes it seem like we are reducing the amount of dependencies.

The last thing is that I think this code should go in a `tools` directory located in the root of the project. It would be good if the shell script that goes along with it would also be located with the Go code instead of in the `hack` directory. Here is an example of how other projects solves this.
https://github.com/openshift/origin/tree/master/tools

I would like @schristoff input on this type of project structure.

AustinAbro321:
> Is the docs changes related to these changes or are we fixing this because we found out it was wrong during this work?

It's related to this work. Essentially we're copying the alpha schema to a different folder. I'm not deleting the original zarf.schema.json since many users rely on it. However once we're in v1 and users no longer need that original zarf.schema.json I'm thinking we can move all the schemas to the tools folder

> We wont remove any dependency in the Zarf CLI in this change. Just making sure because the description makes it seem like we are reducing the amount of dependencies.

Yeah my description was confusing, fixed it. 

> The last thing is that I think this code should go in a tools directory located in the root of the project. It would be good if the shell script that goes along with it would also be located with the Go code instead of in the hack directory. Here is an example of how other projects solves this.

I would agree

AustinAbro321:
Removing the needs design flag, as I simplified this PR to only change where the schema is generated, rather than generating a schema for the apiVersion

phillebaba (approved):

phillebaba (approved):

PR #2904, opened 2024-08-19
fix: update creds not breaking when internal git server not deployed

### LINKED ISSUE #2903: Zarf tools update-creds errors when updating internal git or artifact server when it doesn't exist
### Environment
App version: v0.38.2 (though the issue may have been introduced earlier)

### Steps to reproduce
1. zarf init --confirm (no git server deployed)
2. zarf tools update-creds

### Expected result
Update creds successfully

### Actual Result
Fails because Zarf tries to login to the git server to generate a new artifact token, but the git server doesn't exist. 

### Temporary workaround
Update creds one at a time. For example if you only need to update the registry creds and the state in the agent run
```bash
zarf tools update-creds registry
zarf tools update-creds agent
```

### Note
The underlying issue here is that we save state for git servers and artifact servers even when they don't exist, but we should create a fix for this in the short term. This is related to #2740

AustinAbro321:
## Description

This PR fixes a bug where Zarf will error when the git server or artifact server doesn't exist and `zarf tools update-creds` is run

## Related Issue

Fixes #2903 

## Checklist before merging

### DIFF
--- a/src/cmd/tools/zarf.go
+++ b/src/cmd/tools/zarf.go
@@ -21,7 +21,6 @@ import (
 	"github.com/zarf-dev/zarf/src/cmd/common"
 	"github.com/zarf-dev/zarf/src/config"
 	"github.com/zarf-dev/zarf/src/config/lang"
-	"github.com/zarf-dev/zarf/src/internal/gitea"
 	"github.com/zarf-dev/zarf/src/internal/packager/helm"
 	"github.com/zarf-dev/zarf/src/internal/packager/template"
 	"github.com/zarf-dev/zarf/src/pkg/cluster"
@@ -30,6 +29,7 @@ import (
 	"github.com/zarf-dev/zarf/src/pkg/pki"
 	"github.com/zarf-dev/zarf/src/pkg/zoci"
 	"github.com/zarf-dev/zarf/src/types"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 var subAltNames []string
@@ -97,7 +97,7 @@ var updateCredsCmd = &cobra.Command{
 		} else {
 			if !slices.Contains(validKeys, args[0]) {
 				cmd.Help()
-				return fmt.Errorf("invalid service key specified, valid keys are: %s, %s, and %s", message.RegistryKey, message.GitKey, message.ArtifactKey)
+				return fmt.Errorf("invalid service key specified, valid key choices are: %v", validKeys)
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


src/cmd/tools/zarf.go
```diff
@@ -155,31 +155,17 @@ var updateCredsCmd = &cobra.Command{
...
-						return err
-					}
-					newState.ArtifactServer.PushToken = tokenSha1
-					return nil
-				})
-				if err != nil {
-					// Warn if we couldn't actually update the git server (it might not be installed and we should try to continue)
-					message.Warnf(lang.CmdToolsUpdateCredsUnableCreateToken, err.Error())
+				if kerrors.IsNotFound(err) {
+					message.Warn(updateErr.Error())
```
schristoff, line 166: can we return errs as errs plz :)

src/cmd/tools/zarf.go
```diff
@@ -155,31 +155,17 @@ var updateCredsCmd = &cobra.Command{
...
-					return err
-				}
-				_, err = tunnel.Connect(cmd.Context())
-				if err != nil {
-					return err
+				pushToken, err := c.UpdateInternalArtifactServerToken(ctx, oldState.GitServer)
+				updateErr := fmt.Errorf("unable to create the new Gitea artifact token: %w", err)
+				// TODO once Zarf is changed so the default state is empty for a service when it is not deployed
+				// and sufficient time has passed for users state to get updated we should error on isNotFound
+				if err != nil && !kerrors.IsNotFound(err) {
```
schristoff, line 162: > Check if err 
If err 
>>  Check if is not found 
return  not found 
> return `newState.ArtifactServer.PushToken = pushToken` , no else

AustinAbro321: I think the formatting got messed up here

src/cmd/tools/zarf.go
AustinAbro321: The point of this PR is to log a warning instead of an error when `kerrors.IsNotFound` is true since otherwise `zarf tools update-creds` fails every time.

An alternative I considered was to create a function `internalGitServerExists()` to use here instead of `isInternal()` which would avoid this block all together if it's not found. I ultimately went with this way, as I thought two functions might be confusing (though maybe that's okay since the reality is that it's confusing and we need to change how we create state)

schristoff: okay, that makes sense. :)

src/cmd/tools/zarf.go
schristoff: fixed (hopefully that makes more sense)

AustinAbro321: I wanted to not return the errors on `kerrors.IsNotFound` since some users might not have a git server deployed and in that case this function would always error. 

I just changed the code to not get to this block if the git server doesn't exist which hopefully clarifies intent

schristoff (approved):

PR #2911, opened 2024-08-20
fix: evaulate templates on schema check

### LINKED ISSUE #2908: Linter running before template vars are applied
### Environment
Device and OS: macOS 14.5 & ubuntu 22
App version: 0.38.2
Kubernetes distro being used: k3s

### Steps to reproduce
1. create zarf.yaml:
```
kind: ZarfPackageConfig
metadata:
  name: "###ZARF_PKG_TMPL_PACKAGE_NAME###"
components:
  - name: files
    files:
      - source: ./zarf.yaml
        target: zarf.yaml
```
2. run `zarf package create --confirm --set PACKAGE_NAME="beer" `

### Expected result
```
> zarf package create --confirm --set PACKAGE_NAME="beer"   
...
✔  Package saved to "zarf-package-beer-arm64.tar.zst" 
```
### Actual Result
```
> zarf package create --confirm --set PACKAGE_NAME="beer"                                                                                                                                   Tue Aug 20 14:43:29 2024

 NOTE  Saving log file to
       /var/folders/bx/hywhy9714_g5st9x3lgd3zjh0000gn/T/zarf-2024-08-20-14-44-43-3045695344.log

 NOTE  Using build directory .

 NOTE  Linting package "beer" at .

     Type  | Path           | Message
     Error | .metadata.name | Does not match pattern '^[a-z0-9][a-z0-9\-]*$'

     ERROR:  failed to create package: found errors in schema
```

### Additional Context
the files component is just so the package builds

AustinAbro321:
## Description
We are currently not evaluating templates on schema creation, this can lead to issues if templates are used where there are regex patterns in the schema. This fixes it to evaluate templates on schema creation

## Issue
Fixes #2908

## Checklist before merging

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,16 +40,3 @@ repos:
         types: [go]
         language: golang
         pass_filenames: false
-  - repo: https://github.com/python-jsonschema/check-jsonschema
-    rev: 0.14.0
-    hooks:
-      - id: check-jsonschema
-        name: "Validate Zarf Configs Against Schema"
-        files: "zarf.yaml"
-        types: [yaml]
-        args: ["--schemafile", "zarf.schema.json"]
-        exclude: |
-          (?x)^(
-            src/test/packages/12-lint/.*|
-            src/pkg/lint/testdata/package-with-templates/zarf.yaml
-          )$
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,5 +50,6 @@ repos:
         args: ["--schemafile", "zarf.schema.json"]
         exclude: |
           (?x)^(
-            src/test/packages/12-lint/.*
+            src/test/packages/12-lint/.*|
+            src/pkg/lint/testdata/package-with-templates/zarf.yaml
           )$
--- a/src/pkg/lint/schema_test.go
+++ b/src/pkg/lint/schema_test.go
@@ -7,11 +7,13 @@ package lint
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"testing"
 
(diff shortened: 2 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -40,15 +40,3 @@ repos:
         types: [go]
         language: golang
         pass_filenames: false
-  - repo: https://github.com/python-jsonschema/check-jsonschema
```
AustinAbro321, line 43: Removing this as now that that we check the schema on create so the only time it will be invalid is if it's in a test

src/pkg/lint/schema_test.go
```diff
@@ -190,6 +192,24 @@ components:
...
+	ZarfSchema = testutil.LoadSchema(t, "../../../zarf.schema.json")
+	setVariables := map[string]string{
+		"PACKAGE_NAME": "test-package",
+		"MY_COMP_NAME": "test-comp",
+	}
+	cwd, err := os.Getwd()
+	require.NoError(t, err)
+	err = os.Chdir(filepath.Join("testdata", "package-with-templates"))
+	require.NoError(t, err)
+	defer func() {
```
schristoff, line 205: Why are we doing this?

schristoff (approved):
I would remove that defer but otherwise okay

src/pkg/lint/schema_test.go
AustinAbro321: Need to change directory because the function schema validation along with the entire create flow relies on being in the same directory as the zarf.yaml file. 

Doing a `defer func()` so the linter doesn't complain to me about not checking the error

PR #2915, opened 2024-08-21
chore: update injector make

AustinAbro321:
## Description

This simplifies the makefile by using [cross](https://github.com/cross-rs/cross) and gets rid of external dependencies on S3 for building the object. 

It does also eliminate the ability to create the injector on an arm64 machine. The arm64 binary can still be created through cross on an amd64 machine. 

## Checklist before merging

### DIFF
--- a/packages/zarf-registry/zarf.yaml
+++ b/packages/zarf-registry/zarf.yaml
@@ -111,9 +111,9 @@ components:
         architecture: amd64
     files:
       # Rust Injector Binary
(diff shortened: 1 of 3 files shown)


packages/zarf-registry/zarf.yaml
```diff
@@ -111,9 +111,9 @@ components:
         architecture: amd64
     files:
       # Rust Injector Binary
-      - source: https://zarf-init.s3.us-east-2.amazonaws.com/injector/###ZARF_PKG_TMPL_INJECTOR_VERSION###/zarf-injector-amd64
+      - source: ../../src/injector/target/x86_64-unknown-linux-musl/release/zarf-injector
```
AustinAbro321, line 114: This is just changed for testing, will revert this file before merge

PR #2933, opened 2024-08-27
refactor: provide usage when payload sha is missing

joonas:
## Description

Prior to this change, `zarf-injector` will panic if it's not provided a payload as an argument:
```shell
$ ./zarf-injector 
thread 'main' panicked at src/main.rs:261:35:
index out of bounds: the len is 1 but the index is 1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
zsh: abort      ./zarf-injector
```

With this change, `zarf-injector` will return usage as expected and exit gracefully:
```shell
$ ./zarf-injector 
Usage: ./zarf-injector <sha256sum>
```

## Related Issue

I don't believe there's a related issue for this, but I figured it was worth submitting a fix for it since I came across this.

## Checklist before merging

### DIFF
--- a/src/injector/src/main.rs
+++ b/src/injector/src/main.rs
@@ -8,6 +8,7 @@ use std::io;
 use std::io::Read;
 use std::io::Write;
 use std::path::PathBuf;
+use std::process::exit;
 
 use axum::{
     body::Body,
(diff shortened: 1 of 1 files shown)


schristoff:
This is a great PR, thank you so much! I'm going to check that importing `std` doesn't put us under our 1MiB size limit.

PR #2934, opened 2024-08-27
ci: stop codeql on merge queue

### LINKED ISSUE #2954: codeql flakes during merge queues
Codeql can break during merge queues so we are going to stop running it in merge queues. It is exceedingly unlikely, maybe impossible, for a codeql error to not occur in two different commits separately, but then occur when the commits are merged, so I believe there is little value in codeql on merge queues anyway. 

I have a suspicion this happens when multiple commits are in the merge queue, but I don't have hard evidence. Below is a screenshot of the error
image

AustinAbro321:
## Description

matrix jobs add to the job name which breaks the shim. Additionally, codeql can break during merge queues so we are going to stop running it in merge queues for the time being. It should be impossible for a change from a previous branch to cause a new security error in the merged branch anyway. 
image

## Relating issue
Fixes #2954

## Checklist before merging

### DIFF
--- a/.github/workflows/scan-codeql.yml
+++ b/.github/workflows/scan-codeql.yml
@@ -16,16 +16,6 @@ on:
       - "adr/**"
       - "docs/**"
       - "CODEOWNERS"
-  merge_group:
-    paths-ignore:
-      - "**.md"
-      - "**.jpg"
-      - "**.png"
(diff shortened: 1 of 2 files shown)


mkcp (approved):
Discussed in person. LGTM

PR #2941, opened 2024-08-29
refactor: remove connect strings from packager property

phillebaba:
## Description

This change moves the connect strings map to the installed charts struct. This removes potential race conditions that could be caused by multiple threads writing to the same struct property.

## Related Issue

Depends on #2940
Relates to #2576

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/zarf.go
+++ b/src/pkg/cluster/zarf.go
@@ -134,8 +134,8 @@ func (c *Cluster) PackageSecretNeedsWait(deployedPackage *types.DeployedPackage,
 }
 
 // RecordPackageDeploymentAndWait records the deployment of a package to the cluster and waits for any webhooks to complete.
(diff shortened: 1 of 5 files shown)


src/pkg/cluster/zarf.go
```diff
@@ -187,6 +187,16 @@ func (c *Cluster) RecordPackageDeployment(ctx context.Context, pkg v1alpha1.Zarf
 		componentWebhooks = existingPackageSecret.ComponentWebhooks
 	}
 
+	// TODO: This is done for backwards compartibility and could be removed in the future.
+	connectStrings := types.ConnectStrings{}
```
phillebaba, line 191: Do we want to keep this? I cant find any code that is reading this from the Zarf state.

AustinAbro321: Yeah the one situation where I thought we might need it `zarf connect list` pulls the info direct from cluster. If I was to guess the reason it's on here was for the deprecated Zarf UI, though that may be incorrect. IMO we should remove it from deployed package.

phillebaba: Lets leave it and come back at a later time and remove it in its own PR.

AustinAbro321 (approved):

PR #2942, opened 2024-08-29
refactor: move table printing out of pkg

phillebaba:
## Description

This change moves table printing out of pkg and into cmd. Any table printing should be specific to the CLI and should not be a concern for the exported library.

## Related Issue

Depends on #2941
Relates to #2576

## Checklist before merging


phillebaba:
I need to figure out how to pause log output for credentials in a proper way before this can be merged.

PR #2946, opened 2024-08-29
fix(ci): test-imports workflow breaks when called from a fork

### LINKED ISSUE #2952: Test import workflow breaks when called from a fork
### Description
We can see an example of the test import workflow breaking on a fork here - https://github.com/zarf-dev/zarf/actions/runs/10581228147/job/29404470354?pr=2933. This happens because the workflow currently tries to pull the current commit of the PR from the Zarf repo. If the commit is on a fork, then it will not exist on the Zarf repo. Instead we should check if we can import the codebase from a local directory. 
image

### Severity/Priority
High - we don't want to block outside contributors or have them think their contributions are breaking in some way.

AustinAbro321:
## Description
The test-imports workflows break when called from a fork

## Related issue
fixes #2952

## Checklist before merging

### DIFF
--- a/.github/workflows/test-import.yaml
+++ b/.github/workflows/test-import.yaml
@@ -30,8 +30,7 @@ jobs:
           cd test-import
           echo "$GO_MAIN" > main.go
           go mod init github.com/zarf-dev/test-import
-          # go get github.com/${{ github.event.pull_request.head.repo.full_name }}@${{ github.event.pull_request.head.sha }}
-          go get github.com/${{ github.event.pull_request.head.repo.full_name }}@9f442cea1f989c62b0ec29ed6d8aded0f747f94e
+          go get github.com/${{ github.event.pull_request.head.repo.full_name }}@$COMMIT_SHA
           go mod tidy
           cat go.mod
           cat go.mod | grep -q ${COMMIT_SHA:0:12}
--- a/.github/workflows/test-import.yaml
+++ b/.github/workflows/test-import.yaml
@@ -28,7 +28,7 @@ jobs:
           cd $(mktemp -d)
           echo "$GO_MAIN" > main.go
           go mod init github.com/zarf-dev/test-import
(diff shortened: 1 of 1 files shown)


schristoff:
Needs associated issue and prio please

.github/workflows/test-import.yaml
```diff
@@ -25,15 +25,15 @@ jobs:
...
-          cd $(mktemp -d)
+          cd ..
+          mkdir test-import
+          cd test-import
           echo "$GO_MAIN" > main.go
           go mod init github.com/zarf-dev/test-import
-          go mod edit -replace github.com/zarf-dev/zarf=github.com/${{ github.repository }}@${COMMIT_SHA:0:12}
+          go mod edit -replace github.com/zarf-dev/zarf=../zarf
           go mod tidy
-          cat go.mod | grep -q ${COMMIT_SHA:0:12}
```
phillebaba, line 33: We want to keep this test to make sure that the correct commit is being used in the go.mod file.

AustinAbro321:
@schristoff created #2952

.github/workflows/test-import.yaml
AustinAbro321: Made sure 9f442cea1f989c62b0ec29ed6d8aded0f747f94e failed which is the commit before we fixed the issue last time.

.github/workflows/test-import.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
           cd $(mktemp -d)
           echo "$GO_MAIN" > main.go
           go mod init github.com/zarf-dev/test-import
-          go mod edit -replace github.com/zarf-dev/zarf=github.com/${{ github.repository }}@${COMMIT_SHA:0:12}
+          go get github.com/${{ github.event.pull_request.head.repo.full_name }}@$COMMIT_SHA
```
phillebaba, line 31: This should in my head still use `go mod edit` to replace the original Zarf package with the fork.

phillebaba (approved):

PR #2960, opened 2024-09-03
refactor: move finding table printing to CLI

phillebaba:
## Description

This change moves printing of the findings table from linting to the CLI rather than in the pkg code.

## Related Issue

Relates to #2576 

## Checklist before merging


phillebaba:
I have removed the severity filter because I do not think it is being used. When calling `lint.ValidatePackageSchema` all findings will have the severity error. 

Additionally we should start returning an exit code 1 for linter warnings. It should be up to the user to explicitly ignore this if they do not want to.

src/cmd/dev.go
```diff
@@ -282,7 +292,19 @@ var devLintCmd = &cobra.Command{
...
+			common.PrintFindings(lintErr)
+			// Do not return an error if the findings are all warnings.
+			if lintErr.OnlyWarnings() {
+				return nil
+			}
+		}
+		if err != nil {
+			return err
+		}
+		return nil
```
AustinAbro321, line 307: We should add something here to tell the user the operation was successful

AustinAbro321 (changes requested):
Small comment, otherwise looks great

src/cmd/dev.go
phillebaba: I do not agree. We log if something went wrong and exit 1. If all is good we should do nothing.

AustinAbro321: I believe providing assurances to users that their action was completed successfully provides a better UX so they feel confident what they asked for really occurred. That being said I see the argument that an exit 0 is that feedback and is expected with CLI commands and the UNIX style. 

Looking at some examples across different tools `helm lint` and `pylint` both give output. They do both provide additional information, Helm gives the amount of charts linted while pylint gives a score out of 10 and the improvement over the last run. 

`shellcheck`, `yamllint` and `tflint` all give no output on success.

I am going to approve this, after looking at different tools I think no output is reasonable given our case.

AustinAbro321 (approved):

PR #2964, opened 2024-09-04
feat: switch to health checks for init package waits

### LINKED ISSUE #2855: Image Race Condition Between Zarf Seed Registry and Zarf Permanent Registry With Multiple Registry Replicas
### Environment
Device and OS: Nutanix VM RHEL8
App version: 0.36.1
Kubernetes distro being used: RKE2

### Steps to reproduce
1. Run a zarf init with its registry deployment replica set to many, or have HPA enabled.

### Expected result
Zarf to successfully initialize in the cluster.

### Actual Result
Image pull backoff on some of the permanent registry pods because of what is explained in the below additional context.

Need to produce... standby

### Additional Context
We don't have a reproducible test case outside a Nutanix deployment of ours right now. We believe that when using multiple replicas of the registry and persistent volume provisioning is slow, this can happen.

We have the below custom zarf init package for our Nutanix CSI driver that imports the seed registry and permanent registry from the zarf project. 

What we believe happens is the [wait on the permanent registry](https://github.com/zarf-dev/zarf/blob/02f29325acfbb07081fbf8ee1476582e3ff5e7bc/packages/zarf-registry/zarf.yaml#L181) immediately succeeds and moves on because the seed registry is using the same name as the permanent registry. This can cause the permanent registry to not actually be ready with all the images pushed to it in its persistent volume.

When the permanent registry replicas start to come up, some of the initial pods succeed because they pulled images from the temporary seed registry. Once that seed registry is gone, because the wait didn't actually wait for the permanent registry, and if the permanent registries persistent volumes were not quite ready or fully filled with all the images the seed registry had, any remaining registry pods can't pull images and remain in an image pull backoff. We can scale down to 1 registry pod and that allows the zarf init to continue and finish successfully. 

Custom zarf init:
```yaml
kind: ZarfInitConfig
metadata:
  name: init
  description: "Nutanix CSI Driver Custom Zarf Init Package"
  architecture: amd64
  version: "0.0.1" # This version is not used by zarf, but is used for tracking with the published versions

variables:
  - name: DYNAMIC_FILE_STORE_NAME
    description: "Name of Nutanix File Server to use for Dynamic File storageclass. Should match the name value for the file server in Prism."
  - name: PRISM_ENDPOINT
    description: "IP or hostname of Prism Element."
  - name: PRISM_USERNAME
    description: "Username of prism user to use for Nutanix CSI driver."
  - name: PRISM_PASSWORD
    description: "Password for prism user to use for Nutanix CSI driver."
  - name: STORAGE_CONTAINER
    description: "Name of Nutanix Storage Container for CSI driver to create volumes in."

components:
  # (Optional) Deploys a k3s cluster
  - name: k3s
    import:
      url: oci://ghcr.io/defenseunicorns/packages/init:v0.36.1

  - name: zarf-injector
    required: true
    import:
      url: oci://ghcr.io/defenseunicorns/packages/init:v0.36.1

  # Creates the temporary seed-registry
  - name: zarf-seed-registry
    required: true
    import:
      url: oci://ghcr.io/defenseunicorns/packages/init:v0.36.1
    charts:
      - name: docker-registry
        valuesFiles:
          - values/registry-values.yaml
    # On upgrades ensure we retain the existing PV
    actions:
      onDeploy:
        before:
          - description: Set persistence for upgrade seed registry
            cmd: ./zarf tools kubectl get pvc zarf-docker-registry -n zarf >/dev/null 2>&1 && echo true || echo false
            mute: true
            setVariables:
              - name: UPGRADE_PERSISTENCE
          - description: Set env vars for upgrade seed registry
            mute: true
            cmd: |
              ./zarf tools kubectl get pvc zarf-docker-registry -n zarf >/dev/null 2>&1 && \
              echo "" || \
              echo "- name: REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY
                value: \"/var/lib/registry\""
            setVariables:
              - name: UPGRADE_ENV_VARS
                autoIndent: true

  # Push nutanix csi images to seed-registry
  - name: nutanix-csi-images-initial
    required: true
    description: Push nutanix images to the zarf registry
    images:
      - registry.k8s.io/sig-storage/snapshot-controller:v8.0.1
      - registry.k8s.io/sig-storage/snapshot-validation-webhook:v8.0.1
      - quay.io/karbon/ntnx-csi:v2.6.10
      - registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.11.1
      - registry.k8s.io/sig-storage/csi-provisioner:v5.0.1
      - registry.k8s.io/sig-storage/csi-snapshotter:v8.0.1
      - registry.k8s.io/sig-storage/csi-resizer:v1.11.2
      - registry.k8s.io/sig-storage/livenessprobe:v2.13.1
      - registry1.dso.mil/ironbank/opensource/velero/velero-plugin-for-csi:v0.7.1
      - registry1.dso.mil/ironbank/opensource/velero/velero-plugin-for-aws:v1.10.0

  - name: nutanix-csi-storage
    required: true
    charts:
      # renovate: datasource=helm
      - name: nutanix-csi-storage
        url: https://github.com/defenseunicorns/nutanix-helm.git # fork containing fix for imagepullsecrets needed for pods to pull images from zarf registry
        version: v2.6.10-modified
        gitPath: charts/nutanix-csi-storage
        namespace: ntnx-system
        valuesFiles:
          - values/nutanix-storage-values.yaml
    actions:
      onDeploy:
        before:
          - description: Delete Storage Classes
            cmd: ./zarf tools kubectl delete sc nutanix-volume --ignore-not-found=true

  - name: nutanix-dynamicfile-manifests
    required: true
    manifests:
      - name: nutanix-dynamicfile-manifests
        namespace: ntnx-system
        files:
          - nutanix-dynamicfile.yaml
    actions:
      onDeploy:
        before:
          - description: Delete Storage Classes
            cmd: ./zarf tools kubectl delete sc nutanix-dynamicfile --ignore-not-found=true

  - name: zarf-registry
    required: true
    import:
      url: oci://ghcr.io/defenseunicorns/packages/init:v0.36.1

  # Push nutanix csi (and registry) images to permanent registry
  - name: nutanix-csi-images
    required: true
    description: Push nutanix csi images to the zarf registry
    images:
      - registry.k8s.io/sig-storage/snapshot-controller:v8.0.1
      - registry.k8s.io/sig-storage/snapshot-validation-webhook:v8.0.1
      - quay.io/karbon/ntnx-csi:v2.6.10
      - registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.11.1
      - registry.k8s.io/sig-storage/csi-provisioner:v5.0.1
      - registry.k8s.io/sig-storage/csi-snapshotter:v8.0.1
      - registry.k8s.io/sig-storage/csi-resizer:v1.11.2
      - registry.k8s.io/sig-storage/livenessprobe:v2.13.1
      - registry1.dso.mil/ironbank/opensource/velero/velero-plugin-for-csi:v0.7.1
      - registry1.dso.mil/ironbank/opensource/velero/velero-plugin-for-aws:v1.10.0

  - name: zarf-agent
    required: true
    import:
      url: oci://ghcr.io/defenseunicorns/packages/init:v0.36.1

  # (Optional) Adds a git server to the cluster
  - name: git-server
    import:
      url: oci://ghcr.io/defenseunicorns/packages/init:v0.36.1

```

AustinAbro321:
## Description

By using health checks we can ensure that resources are fully reconciled before continuing to the next component. This is especially important for the registry component. Since the registry is a re-deploy of the same helm chart as the seed-registry the deployment is already available with extra pods and the current wait condition passes immediately. 

## Related Issue

Fixes #2855

## Checklist before merging

### DIFF
--- a/packages/zarf-registry/zarf.yaml
+++ b/packages/zarf-registry/zarf.yaml
@@ -171,12 +171,8 @@ components:
     images:
       # This image (or images) must match that used for injection (see zarf-config.toml)
       - "###ZARF_PKG_TMPL_REGISTRY_IMAGE_DOMAIN######ZARF_PKG_TMPL_REGISTRY_IMAGE###:###ZARF_PKG_TMPL_REGISTRY_IMAGE_TAG###"
-    actions:
-      onDeploy:
-        after:
-          - wait:
-              cluster:
-                kind: deployment
-                namespace: zarf
-                name: app=docker-registry
-                condition: Available
+    healthChecks:
+      - name: zarf-docker-registry
+        namespace: zarf
+        kind: Deployment
+        apiVersion: apps/v1
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,6 @@ require (
 	github.com/opencontainers/image-spec v1.1.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.18.0
-	github.com/prometheus/prometheus v0.47.2
 	github.com/pterm/pterm v0.12.79
 	github.com/sergi/go-diff v1.3.2-0.20230802210424-5b0b94c5c0d3
 	github.com/sigstore/cosign/v2 v2.2.3
@@ -78,12 +77,9 @@ require (
 	github.com/go-logr/logr v1.4.2 // indirect
 	github.com/gofrs/uuid v4.2.0+incompatible // indirect
 	github.com/gorilla/handlers v1.5.1 // indirect
-	github.com/grafana/regexp v0.0.0-20221122212121-6b5c0a4cb7fd // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.20.0 // indirect
 	github.com/hashicorp/golang-lru/arc/v2 v2.0.5 // indirect
 	github.com/hashicorp/golang-lru/v2 v2.0.5 // indirect
-	github.com/jpillora/backoff v1.0.0 // indirect
-	github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f // indirect
 	github.com/onsi/gomega v1.32.0 // indirect
 	github.com/redis/go-redis/extra/rediscmd/v9 v9.0.5 // indirect
 	github.com/redis/go-redis/extra/redisotel/v9 v9.0.5 // indirect
@@ -434,7 +430,7 @@ require (
 	github.com/pkg/profile v1.7.0 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/prometheus/client_model v0.5.0 // indirect
-	github.com/prometheus/common v0.45.0
+	github.com/prometheus/common v0.45.0 // indirect
 	github.com/prometheus/procfs v0.12.0 // indirect
 	github.com/protocolbuffers/txtpbfmt v0.0.0-20231025115547-084445ff1adf // indirect
 	github.com/rakyll/hey v0.1.4 // indirect
@@ -532,7 +528,7 @@ require (
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/op/go-logging.v1 v1.0.0-20160211212156-b2cb9fa56473 // indirect
 	gopkg.in/warnings.v0 v0.1.2 // indirect
-	gopkg.in/yaml.v2 v2.4.0
+	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	gorm.io/gorm v1.25.5 // indirect
 	k8s.io/apiextensions-apiserver v0.30.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -39,6 +39,7 @@ require (
 	github.com/opencontainers/image-spec v1.1.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.18.0
+	github.com/prometheus/prometheus v0.47.2
 	github.com/pterm/pterm v0.12.79
 	github.com/sergi/go-diff v1.3.2-0.20230802210424-5b0b94c5c0d3
 	github.com/sigstore/cosign/v2 v2.2.3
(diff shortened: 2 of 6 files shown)


packages/zarf-registry/zarf.yaml
```diff
@@ -171,12 +171,8 @@ components:
...
-              cluster:
-                kind: deployment
-                namespace: zarf
-                name: app=docker-registry
-                condition: Available
+    healthChecks:
+      - name: zarf-docker-registry
+        namespace: zarf
+        kind: Deployment
+        apiVersion: apps/v1
```
AustinAbro321, line 178: One interesting thing I'm seeing now that health checks are added is that sometimes we see the following error on image push for the agent (the component right after the registry). The failure happens nearly instantaneously and we retry right away so there is no effect on the user besides seeing the text. It seems that the service is trying to forward to the pod right as it's dying. This didn't happen before because we didn't wait for the seed registry to terminate. This still happens if I add a health check on the service.
```bash
    Pushing ghcr.io/zarf-dev/zarf/agent:local  0sE0904 09:28:20.759620 3387486 portforward.go:413] an error occurred forwarding 39239 -> 5000: error forwarding port 5000 to pod 8b4ba41141648cc01c39f674e94dfd83f36755ee5416d118cd012a88d0b46476, uid : failed to execute portforward in network namespace "/var/run/netns/cni-41dd6540-8191-9e97-19ed-a1ca0d90316c": failed to connect to localhost:5000 inside namespace "8b4ba41141648cc01c39f674e94dfd83f36755ee5416d118cd012a88d0b46476", IPv4: dial tcp4 127.0.0.1:5000: connect: connection refused IPv6 dial tcp6 [::1]:5000: connect: connection refused 
  ✔  Pushed 1 images
```
I don't think this should stop us from merging, but wanted to take a note of it.

phillebaba: Is this because we are now too fast to do a port-forward?

AustinAbro321: I'm thinking that it's trying to do a port-forward on the old pod after it died, but the timing is very tight. It always works on the second try, if it fails at all. The UID in the error doesn't match the new pod, so I'm guessing it would match the old one, not sure if there's an easy way to get deleted pod UIDs

AustinAbro321: This is strange, I did re-runs and watched the UID of the registry pods. The UID in the error message is neither the old registry pod or the new registry pod.

AustinAbro321: Current theory is that the deployment and service can be ready but the old endpoint slice for the seed registry still exists. This would be consistent with what's written in the [kstatus docs](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md#status-for-a-resource-depends-on-the-status-of-other-resources) about resources that create other resources

AustinAbro321: Leaving this here, this is what the error message looks like. It originates from this line in the Kubernetes port-forward code. I think it gets put on the progress bar somehow, since we don't return errors when the retry doesn't fail. https://github.com/kubernetes/client-go/blob/4fca7081f8d6f62e92271052035e50fe1c6e9ac7/tools/portforward/portforward.go#L365
image

AustinAbro321: I changed the code to get the pods instead of using the service. As far as I can tell from running a script that inits on my local PC 10 times, we don't end up with the above error message, but it can still error once, it just works on the next try and since the retry function returns nil we don't ever see the error message.

AustinAbro321:
Closing this as #3043 is going to add this functionality to all deployments by default

PR #2965, opened 2024-09-04
fix: progress bar image name flashes on push

### LINKED ISSUE #2651: Improved image push progress bar support for long image names
When pushing images with long paths, the Zarf image push output tends to strobe between `Pushing <x> images` and a truncated message with the name of the image (shown in the video below).

I believe this may be a result of the [default max line width (100c)](https://github.com/defenseunicorns/zarf/blob/main/src/pkg/message/progress.go#L33) in the progress bar.

Ironbank images (as a common example) tend to be quite long due to the repository folder structure:
```
registry1.dso.mil/ironbank/tetrate/istio/proxyv2:1.22.1-tetratefips-v0-zarf-3754112098
```

At 87 characters^, this image name leaves little room on the progress bar line.

### Proposed solution(s)
* Increase the default width to 120 characters (or some other arbitrary length to support longer format image names)
* Separate the progress bar output from the name of the image being pushed (more lines)

### Additional context
Example:
https://github.com/defenseunicorns/zarf/assets/9499393/7d27784e-0bcc-4889-8e6f-d6eef6d0a8a0

AustinAbro321:
## Description

This PR fixes the CLI flashing when pushing images. This can be seen in the linked gif in the issue #2651

## Related Issue

Fixes #2651

## Checklist before merging

### DIFF
--- a/src/internal/packager/images/push.go
+++ b/src/internal/packager/images/push.go
@@ -87,11 +87,6 @@ func Push(ctx context.Context, cfg PushConfig) error {
 			refTruncated := helpers.Truncate(refInfo.Reference, 55, true)
 			progress.Updatef(fmt.Sprintf("Pushing %s", refTruncated))
 
-			size, err := calcImgSize(img)
-			if err != nil {
-				return err
-			}
-
 			// If this is not a no checksum image push it for use with the Zarf agent
 			if !cfg.NoChecksum {
 				offlineNameCRC, err := transform.ImageTransformHost(registryURL, refInfo.Reference)
@@ -102,8 +97,6 @@ func Push(ctx context.Context, cfg PushConfig) error {
 				if err = pushImage(img, offlineNameCRC); err != nil {
 					return err
 				}
-
-				totalSize -= size
 			}
 
 			// To allow for other non-zarf workloads to easily see the images upload a non-checksum version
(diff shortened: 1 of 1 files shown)


phillebaba:
Does the total progress even matter? We set it when we define the progress bar but we never update it as far as I can see?  I want to make sure that reusing the progress bar wont result in starting at the progress state where we ended.

AustinAbro321:
Great question @phillebaba. The total progress does matter. We send crane a custom implementation of the `http.RoundTripper`. This can be found in the [helpers](https://github.com/defenseunicorns/pkg/blob/main/helpers/transport.go#L89-L94) pkg package. This implementation updates the progress bar based on the amount of bytes received from the request. 

The current code creates a new progress bar before the program starts and on each retry without closing the old one. This always results in at least two progress bars active at the same time which causes the flashing.

phillebaba:
Are you sure the issue is not that we are not deffering stopping the progress bar before a retry is run? Looking at the current code if the retry is run the old progress bar is never stopped.

AustinAbro321:
Not stopping the progress bar is part of the issue with the current code, but either way the flashing will happen because we have two progress bars going at the same time. IMO we should just have one progress bar at top level. If there is an error in the request the progress bar won't get updated on that attempt, but will get updated on subsequent attempts. 

Here is a video from the current state of this branch where I artificially make an error happen. When an error happens the progress bar still completes. It leaves behind a progress bar on error but that happens no matter what. 
[Screencast from 09-05-2024 09:10:29 AM.webm](https://github.com/user-attachments/assets/b31ca93a-2ca1-4c84-9604-90a13f120e52)

Here is a video from the latest release where I artificially make an error happen. Yes it leaves behind progress bars, but it's also flashing from the start. 
[Screencast from 09-05-2024 09:06:00 AM.webm](https://github.com/user-attachments/assets/c094f0ed-1e59-40b9-b353-66e3a493a9bb)

phillebaba (approved):
LGTM

Issue #2953, opened 2024-08-30
Fix error handing


phillebaba:
In an effort to standardize error handling in Zarf we need to refactor some of the existing error handling. This includes fixing any place where we ignore errors or chose actively to log them instead of returning them.

- [x] https://github.com/zarf-dev/zarf/issues/2950
- [x] https://github.com/zarf-dev/zarf/issues/2951
- [x] https://github.com/zarf-dev/zarf/issues/2949
- [x] https://github.com/zarf-dev/zarf/issues/2906
- [x] https://github.com/zarf-dev/zarf/issues/2872
- [x] https://github.com/zarf-dev/zarf/issues/2700
- [ ] https://github.com/zarf-dev/zarf/issues/2756

mkcp:
Related https://github.com/zarf-dev/zarf/issues/2978

mkcp:
Marking #2950 done w/ #2979 and #2987

AustinAbro321:
Closing this as every other issue is closed and https://github.com/zarf-dev/zarf/issues/2756 is blocked, but will easily be fixed during the move to Helm 4

Issue #2951, opened 2024-08-30
Return nil values when returning an error


phillebaba:
A good practice in Go when returning errors is to return nil or empty data for all other return values. It makes sure there is no other potential confusion on the calling end that the data returned is valid and could be used. 

Here are some examples where we return data and an error.

https://github.com/zarf-dev/zarf/blob/f60b2f48727bbe4be3ebbe90366ecdd4ef8b3f39/src/pkg/utils/yaml.go#L200
https://github.com/zarf-dev/zarf/blob/f60b2f48727bbe4be3ebbe90366ecdd4ef8b3f39/src/pkg/utils/io.go#L43
https://github.com/zarf-dev/zarf/blob/f60b2f48727bbe4be3ebbe90366ecdd4ef8b3f39/src/pkg/zoci/fetch.go#L20
https://github.com/zarf-dev/zarf/blob/f60b2f48727bbe4be3ebbe90366ecdd4ef8b3f39/src/pkg/cluster/zarf.go#L281

This is slightly related to #2950 and could be done in conjunction with it.

mkcp:
I'm intending on skipping packager and extensions (e.g. bigbang) for this to minimize churn with other active refactorings.

Issue #2950, opened 2024-08-30
Remove use of named returns


phillebaba:
While use of named returns is not directly wrong, it is a matter of taste whether to use them or not. From experience it has a tendency to cause confusion both when understanding code and immediately seeing what data is actually returned. There are a few cases where named returns are needed but we should be explicit when using them in these cases.

Checking for named returns can be done with the linter using `nonamedreturns`.
https://golangci-lint.run/usage/linters/#nonamedreturns

mkcp:
Need to run another PR on stuff outside of pkg before we can mark this done!

PR #2972, opened 2024-09-06
WIP: Fix error handling

mkcp:
## Description

Currently this WIP PR adds some error handling within `utils/yaml.go`. The intent is to address one or part of the concerns in #2950 & #2951, after understanding the scope of the changes more. I just wanted to address some low hanging fruit that I found in the meantime.

## Related Issue
Relates to #2953 

## Checklist before merging

### DIFF
--- a/src/pkg/packager/interactive.go
+++ b/src/pkg/packager/interactive.go
@@ -18,10 +18,13 @@ import (
 	"github.com/zarf-dev/zarf/src/pkg/utils"
 )
 
-func (p *Packager) confirmAction(stage string, warnings []string, sbomViewFiles []string) bool {
+func (p *Packager) confirmAction(stage string, warnings []string, sbomViewFiles []string) (bool, error) {
 	pterm.Println()
 	message.HeaderInfof("📦 PACKAGE DEFINITION")
-	utils.ColorPrintYAML(p.cfg.Pkg, p.getPackageYAMLHints(stage), true)
+	err := utils.ColorPrintYAML(p.cfg.Pkg, p.getPackageYAMLHints(stage), true)
+	if err != nil {
+		return false, err
+	}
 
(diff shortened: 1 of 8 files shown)


src/pkg/utils/yaml.go
```diff
@@ -36,8 +36,11 @@ func yamlFormat(attr color.Attribute) string {
 }
 
 // ColorPrintYAML pretty prints a yaml file to the console.
-func ColorPrintYAML(data any, hints map[string]string, spaceRootLists bool) {
-	text, _ := goyaml.Marshal(data)
+func ColorPrintYAML(data any, hints map[string]string, spaceRootLists bool) error {
```
mkcp, line 39: Add error handling around input

src/pkg/packager/interactive.go
```diff
@@ -18,10 +18,13 @@ import (
 	"github.com/zarf-dev/zarf/src/pkg/utils"
 )
 
-func (p *Packager) confirmAction(stage string, warnings []string, sbomViewFiles []string) bool {
+func (p *Packager) confirmAction(stage string, warnings []string, sbomViewFiles []string) (bool, error) {
```
mkcp, line 21: Needed to be updated to handle errors from ColorPrintYAML. Alternatively we might be able to just quietly error or debug log these if Packager has access to a logger and they're low impact. Very open to feedback here!

AustinAbro321: Great question! This is low impact, but we're trying to get rid of all the quiet debugs, unless there is a good reason for it. This is a benign example, but there were plenty of places in the code where we continued because it might not cause a problem and that made implementations confusing and harder to read. If `goyaml.Marshall` errors there probably is something pretty bad going on, though 99.999% of the time it will be fine unless there's an actual issue in our code.

mkcp:
Closing this one out to avoid changes to packager for now.

Issue #2969, opened 2024-09-05
reimplement packager functions parent issue


schristoff:
**Purpose:**
As part of schema work it has been identified that `src/packager` intertwines "deploy time" and "package create time" logic. This makes it burdensome, or possibly impossible, to implement versioned schemas without passing a versioned structs throughout the codebase. 

We will begin by refactoring the following pieces of `src/packager` into `src/packager2` - this will allow us to re-implement this functionality without making breaking changes to Zarf in between releases. We will then swap over the calls to `packager.()` functions to `packager2.()` and deprecate `src/packager` functionality for `src/packager2` (while renaming packager2 to packager)

**Decisions Made:**
- To refactor "Opts" out of the larger Packager struct, `Opts` will be moved into their respective "primary file". All exported functions will take its "related" `Opts` struct as a function argument. 
E.g. - [InspectOpts](https://github.com/zarf-dev/zarf/blob/cf4e9891d531b792849394f80830363411242fd9/src/types/packager.go#L27) is moved into the [inspect.go](https://github.com/zarf-dev/zarf/pull/2990/files#diff-ce1d4a01e228f255d9bda4038e0ba0c7dc1d8f798d3629503d4e137076e6f90aR14) file within `packager2`

**To Be Worked:** 
- [x] [func New(cfg *types.PackagerConfig, mods ...Modifier) (*Packager, error)](https://pkg.go.dev/github.com/defenseunicorns/zarf/src/pkg/packager#New)
- [x] [func (p *Packager) ClearTempPaths()](https://pkg.go.dev/github.com/defenseunicorns/zarf/src/pkg/packager#Packager.ClearTempPaths)
- [x] [func (p *Packager) Create(ctx context.Context) error](https://pkg.go.dev/github.com/defenseunicorns/zarf/src/pkg/packager#Packager.Create)
- [ ] [func (p *Packager) Deploy(ctx context.Context) error](https://pkg.go.dev/github.com/defenseunicorns/zarf/src/pkg/packager#Packager.Deploy)
- [ ] [func (p *Packager) DevDeploy(ctx context.Context) error](https://pkg.go.dev/github.com/defenseunicorns/zarf/src/pkg/packager#Packager.DevDeploy)
- [ ] [func (p *Packager) FindImages(ctx context.Context) (map[string][]string, error)](https://pkg.go.dev/github.com/defenseunicorns/zarf/src/pkg/packager#Packager.FindImages)
[...]

PR #2975, opened 2024-09-06
refactor: mirror-resources

phillebaba:
## Description

Re implement mirror.

## Related Issue

Relates to #2969 

## Checklist before merging


phillebaba:
The tests are failing because the same registry args are used for mirror-resoruces and init. Because I changed the argument to in mirror to be similar to the git server argument it no longer works with init.

src/internal/packager2/load.go
```diff
@@ -0,0 +1,212 @@
...
+		if err := sources.ValidatePackageIntegrity(pkgPaths, pkg.Metadata.AggregateChecksum, isPartial); err != nil {
+			return nil, err
+		}
+		if opt.SkipSignatureValidation {
+			if err := sources.ValidatePackageSignature(ctx, pkgPaths, opt.PublicKeyPath); err != nil {
+				return nil, err
+			}
+		}
+	}
+	for _, component := range pkg.Components {
```
schristoff, line 143: Will this return errors now without `unarchiveAll`

src/internal/packager2/load.go
```diff
@@ -0,0 +1,212 @@
...
+		_, err = io.Copy(tarFile, f)
+		if err != nil {
+			return fmt.Errorf("unable to copy file %s: %w", splitFile, err)
+		}
+		err = f.Close()
+		if err != nil {
+			return fmt.Errorf("unable to close file %s: %w", splitFile, err)
+		}
+	}
+	if err := helpers.SHAsMatch(tarPath, pkgData.Sha256Sum); err != nil {
```
schristoff, line 208: How does `pkgData.Sha245Sum` get set in the split tar  - I'm not seeing where we are unmarshaling the package metadata anymore

src/internal/packager2/load.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	tarDir, err := utils.MakeTempDir(config.CommonOptions.TempDirectory)
+	if err != nil {
+		return nil, err
+	}
+	defer os.RemoveAll(tarDir)
+	tarPath := filepath.Join(tarDir, "packager.tar")
```
schristoff, line 50: Should the `tar` specific stuff here be moved into the tar case statement?

src/internal/packager2/load.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/zarf-dev/zarf/src/config"
+	"github.com/zarf-dev/zarf/src/pkg/layout"
+	"github.com/zarf-dev/zarf/src/pkg/packager/filters"
+	"github.com/zarf-dev/zarf/src/pkg/packager/sources"
+	"github.com/zarf-dev/zarf/src/pkg/utils"
+	"github.com/zarf-dev/zarf/src/types"
+)
+
+// LoadOptions are the options for LoadPackage.
+type LoadOptions struct {
```
schristoff, line 30: I lightly think it may be good to add something like isPartial into here since it is state that is getting modified at runtime across different areas of execution, but it may make things messy so I don't have a strong stance.

src/internal/packager2/mirror.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/zarf-dev/zarf/src/pkg/layout"
+	"github.com/zarf-dev/zarf/src/pkg/message"
+	"github.com/zarf-dev/zarf/src/pkg/packager/filters"
+	"github.com/zarf-dev/zarf/src/pkg/transform"
+	"github.com/zarf-dev/zarf/src/pkg/utils"
+	"github.com/zarf-dev/zarf/src/types"
+)
+
+// MirrorOptions are the options for Mirror.
+type MirrorOptions struct {
```
schristoff, line 33: Should we add stuff like `New/GetMirrorOptions() *MirrorOptions { ` while we are here?

src/internal/packager2/load.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, fmt.Errorf("unknown source type: %s", opt.Source)
+	}
+	if srcType != "oci" && opt.Shasum != "" {
+		err := helpers.SHAsMatch(tarPath, opt.Shasum)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Extract the package
```
schristoff, line 80: Could `extractPackage()` be it's own subfunc out of here?

schristoff (commented):
good work. consolidating sources is not fun. 
it may be valuable to break this up into multiple PRs (dns, load and mirror) 
but up to you.

src/internal/dns/dns.go
```diff
@@ -0,0 +1,45 @@
...
+	"regexp"
+	"strconv"
+)
+
+var (
+	// localClusterServiceRegex is used to match the local cluster service format:
+	localClusterServiceRegex = regexp.MustCompile(`^(?P<name>[^\.]+)\.(?P<namespace>[^\.]+)\.svc\.cluster\.local$`)
+)
+
+// IsServiceURL returns true if the give url complies with the service url format.
```
schristoff, line 20:
```suggestion
// IsServiceURL returns if the give url complies with the service url format.
```

src/internal/packager2/mirror.go
```diff
@@ -0,0 +1,236 @@
...
+	Cluster         *cluster.Cluster
+	PackagePaths    layout.PackagePaths
+	Filter          filters.ComponentFilterStrategy
+	RegistryInfo    types.RegistryInfo
+	GitInfo         types.GitServerInfo
+	NoImageChecksum bool
+	Retries         int
+}
+
+// Mirror mirrors the package contents to the give registry and git server.
```
schristoff, line 43:
```suggestion
// Mirror mirrors the package contents to the given registry and git server.
```

src/internal/dns/dns.go
phillebaba: Why remove true? It returns either way but only true if the pattern matches?

src/internal/packager2/load.go
phillebaba: No because each case will in the end produce a tar of the package. I do think it would be good to make it its own function though.

src/internal/packager2/load.go
phillebaba: The bool `unarchiveAll` is only false in the publish logic. I done really understand why this is. Personally I want to get away from all the different states that a package path can exist in by removing options.

src/internal/packager2/load.go
phillebaba: The shasum that exists in the package is the same as the tar package that is assembled at the end. In the current implementation we are not actually checking the shasum of the assembled tar, only what the metadata says. This means that the resulting tar package could be something completely different. Shasum checking is done after the tar is assembled.

This comment did however made me realize that like the pull command we dont have a flag to set the shasum for mirror. So I have added that.

src/internal/packager2/mirror.go
phillebaba: I dont think we should as it would create the same situation with function signatures that we did not like in the first place.

src/cmd/package.go
```diff
@@ -128,18 +129,47 @@ var packageMirrorCmd = &cobra.Command{
 		}
 	},
 	RunE: func(cmd *cobra.Command, args []string) error {
-		packageSource, err := choosePackage(args)
+		var c *cluster.Cluster
+		if dns.IsServiceURL(pkgConfig.InitOpts.RegistryInfo.Address) || dns.IsServiceURL(pkgConfig.InitOpts.GitServer.Address) {
```
AustinAbro321, line 133: The way this code is right now, if someone has an internal git registry, and an external git server or vice versa mirror resources will break. Maybe it would be better to split apart this command into two sub commands that do one or the other. 

We are breaking most of the existing use cases by not accepting localhost for the in cluster registry service. If we're committed to doing that, it might be a clearer signal to users to just change the command in general.

AustinAbro321 (changes requested):

src/cmd/package.go
phillebaba: We can add an additional check before if that makes things better. I dont know if we have such a use case but it would be simple to support.

In regards to the breaking change I am bringing things inline with what the git server configuration is. Something that should have been done from the beginning. In the current solution we have two different logic for the git server and registry.  Currently any localhost address for the registry is assumed to be an internal registry, which is a large assumption to make. I have updated the docs and we can add a note in the Changelog, but we have to be able to make breaking changes that adjust behavior to be more standardized, especially before a v1 release.

src/internal/packager2/mirror.go
```diff
@@ -0,0 +1,239 @@
...
+
+	for _, component := range pkg.Components {
+		for _, repoURL := range component.Repos {
+			repository, err := git.Open(pkgPaths.Components.Dirs[component.Name].Repos, repoURL)
+			if err != nil {
+				return err
+			}
+			err = retry.Do(func() error {
+				if c == nil {
+					message.Infof("Pushing Git repository %s", gitInfo.Address)
```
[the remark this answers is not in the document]
phillebaba: We are not pushing to the repoURL, I will change the log message to make this more clear.

src/cmd/package.go
AustinAbro321: Gotcha, yeah an additional check to support would be good. I believe software factory might have that use case, though I'm not 100% sure. They use the Zarf registry, but instead of the Zarf gitea instance they spin up their own gitlab instance.

src/internal/dns/dns.go
schristoff: This is not typically how I see the comments for those that return bools, but if it's what you are used to feel free to ignore.

src/internal/packager2/mirror.go
schristoff: I'm not sure I agree, but we can continue this discussion throughout the refactor.

AustinAbro321 (approved):
This simplifies the code a lot, nice work

schristoff (approved):

PR #2985, opened 2024-09-10
chore: only show config file if there is one

catsby:
## Description
 
Don't output the `NOTE` on config file usage if there is no config file set.

Currently `zarf version` [will not go through the process of setting the configuration file](https://github.com/zarf-dev/zarf/blob/efa306fdcbe605ec6d17f4cf72c707f7227af52a/src/cmd/common/viper.go#L122-L124), but the root cmd code path still has a part where it calls `printViperConfigUsed()` and tries to print it. As a result you get empty output, even if you have a valid config available:

no_config

In this PR we simply omit the config file note if we have no config file set. The [printViperConfigUsed()](https://github.com/zarf-dev/zarf/blob/efa306fdcbe605ec6d17f4cf72c707f7227af52a/src/cmd/common/viper.go#L164) function doesn't accept any parameters and doesn't seem to have access to anything that would tell it what specific command is being run, so I didn't see an easy way to only omit the output if the command is the `version` command. 

## Checklist before merging

### DIFF
--- a/src/cmd/common/viper.go
+++ b/src/cmd/common/viper.go
@@ -175,7 +175,9 @@ func printViperConfigUsed() {
 		message.WarnErrf(vConfigError, lang.CmdViperErrLoadingConfigFile, vConfigError.Error())
 		return
 	}
-	message.Notef(lang.CmdViperInfoUsingConfigFile, v.ConfigFileUsed())
+	if cfgFile := v.ConfigFileUsed(); cfgFile != "" {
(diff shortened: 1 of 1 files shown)


AustinAbro321:
@catsby This is a good change thanks. Could you follow the [dco instructions](https://github.com/zarf-dev/zarf/pull/2985/checks?check_run_id=29939263344) to sign your commit

AustinAbro321 (approved):

PR #2988, opened 2024-09-11
chore: ensure we return zeroed value when returning errors

### LINKED ISSUE #2951: Return nil values when returning an error

mkcp:
## Description
This PR goes through everything in src (ignoring packager and extensions)  and ensures we don't return an error as well as some kind of partial or dirty state from our return vars. Most of these weren't actual vectors for bugs or system instability, it's just good practice. There's some minor cleanups along the way too. 

It's worth noting that I used a specific ag command to find error returns (discussed in #2951) so any error variables not starting with `err` will have been missed in this refactor. I expect that there will be very few of these given how err is the convention.

I left some review notes which I'll highlight in comments for whoever gets eyes on.

## Related Issue

Fixes #2951 

## Checklist before merging

### DIFF
--- a/src/cmd/tools/helm/load_plugins.go
+++ b/src/cmd/tools/helm/load_plugins.go
@@ -318,12 +318,12 @@ func loadFile(path string) (*pluginCommand, error) {
 	cmds := new(pluginCommand)
 	b, err := os.ReadFile(path)
 	if err != nil {
-		return &pluginCommand{}, fmt.Errorf("file (%s) not provided by plugin. No plugin auto-completion possible", path)
+		return nil, fmt.Errorf("file (%s) not provided by plugin. No plugin auto-completion possible", path)
 	}
 
 	err = yaml.Unmarshal(b, cmds)
 	if err != nil {
-		return &pluginCommand{}, err
+		return nil, err
 	}
 	return cmds, nil
 }
--- a/src/cmd/tools/helm/load_plugins.go
+++ b/src/cmd/tools/helm/load_plugins.go
@@ -318,11 +318,14 @@ func loadFile(path string) (*pluginCommand, error) {
 	cmds := new(pluginCommand)
 	b, err := os.ReadFile(path)
 	if err != nil {
-		return cmds, fmt.Errorf("file (%s) not provided by plugin. No plugin auto-completion possible", path)
+		return &pluginCommand{}, fmt.Errorf("file (%s) not provided by plugin. No plugin auto-completion possible", path)
 	}
 
 	err = yaml.Unmarshal(b, cmds)
-	return cmds, err
+	if err != nil {
+		return &pluginCommand{}, err
+	}
+	return cmds, nil
 }
 
 // pluginDynamicComp call the plugin.complete script of the plugin (if available)
--- a/src/pkg/cluster/pvc.go
+++ b/src/pkg/cluster/pvc.go
@@ -10,9 +10,8 @@ import (
 )
 
 // UpdateGiteaPVC updates the existing Gitea persistent volume claim and tells Gitea whether to create or not.
-// REVIEW(mkcp): It looks like we've got some error+signal coming back from these functions where we return a both a
-// string true/false downstream but sometimes with errors. So I'm not going to make these `return "", err` but we may
-// want to consider if returning the error is necessary and not, for example, better served with a new type.
+// TODO(mkcp): We return both string true/false and errors here so our callers get a string. This should be returning an
+// empty val if we error, but we'll have to refactor upstream beforehand.
 func (c *Cluster) UpdateGiteaPVC(ctx context.Context, pvcName string, shouldRollBack bool) (string, error) {
 	if shouldRollBack {
 		pvc, err := c.Clientset.CoreV1().PersistentVolumeClaims(ZarfNamespaceName).Get(ctx, pvcName, metav1.GetOptions{})
--- a/src/pkg/cluster/pvc.go
+++ b/src/pkg/cluster/pvc.go
@@ -10,6 +10,9 @@ import (
 )
 
 // UpdateGiteaPVC updates the existing Gitea persistent volume claim and tells Gitea whether to create or not.
+// REVIEW(mkcp): It looks like we've got some error+signal coming back from these functions where we return a both a
+// string true/false downstream but sometimes with errors. So I'm not going to make these `return "", err` but we may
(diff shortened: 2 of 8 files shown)


src/pkg/cluster/pvc.go
```diff
@@ -10,6 +10,9 @@ import (
 )
 
 // UpdateGiteaPVC updates the existing Gitea persistent volume claim and tells Gitea whether to create or not.
+// REVIEW(mkcp): It looks like we've got some error+signal coming back from these functions where we return a both a
```
mkcp, line 13: Review request (opt @phillebaba)

src/pkg/cluster/zarf.go
```diff
@@ -52,7 +52,11 @@ func (c *Cluster) GetDeployedZarfPackages(ctx context.Context) ([]types.Deployed
 		deployedPackages = append(deployedPackages, deployedPackage)
 	}
 
-	return deployedPackages, errors.Join(errs...)
+	// REVIEW(mkcp): This seems like it could be a breaking behavior change. Can I get a second set of eyes here?
```
mkcp, line 55: Review request

src/cmd/tools/helm/load_plugins.go
```diff
@@ -318,11 +318,14 @@ func loadFile(path string) (*pluginCommand, error) {
 	cmds := new(pluginCommand)
 	b, err := os.ReadFile(path)
 	if err != nil {
-		return cmds, fmt.Errorf("file (%s) not provided by plugin. No plugin auto-completion possible", path)
+		return &pluginCommand{}, fmt.Errorf("file (%s) not provided by plugin. No plugin auto-completion possible", path)
```
phillebaba, line 321: I think it is better to return nil instead of a pointer to an empty struct.

src/pkg/cluster/pvc.go
phillebaba: Your analysis is correct. The problem is that the caller of this function expects to have a string to print to STDOUT which is then  consumed by something else that is calling the CLI. Until we can fix the underlying implementation keeping the TODO is all we can do.

src/pkg/cluster/zarf.go
```diff
@@ -33,11 +33,11 @@ func (c *Cluster) GetDeployedZarfPackages(ctx context.Context) ([]types.Deployed
 	listOpts := metav1.ListOptions{LabelSelector: ZarfPackageInfoLabel}
 	secrets, err := c.Clientset.CoreV1().Secrets(ZarfNamespaceName).List(ctx, listOpts)
 	if err != nil {
-		return nil, err
+		return []types.DeployedPackage{}, err
```
phillebaba, line 36: Slices can be returned as nil.

src/pkg/cluster/zarf.go
```diff
@@ -33,11 +33,11 @@ func (c *Cluster) GetDeployedZarfPackages(ctx context.Context) ([]types.Deployed
...
 	}
 
-	errs := []error{}
-	deployedPackages := []types.DeployedPackage{}
+	errs := make([]error, 0)
+	deployedPackages := make([]types.DeployedPackage, 0)
```
phillebaba, line 40: I think it is fine with the original implementation. I am generally not a fan of make to create a struct or slice unless it is for the explicit reason to pre allocate capacity.

src/pkg/cluster/zarf.go
```diff
@@ -52,7 +52,11 @@ func (c *Cluster) GetDeployedZarfPackages(ctx context.Context) ([]types.Deployed
 		deployedPackages = append(deployedPackages, deployedPackage)
 	}
 
-	return deployedPackages, errors.Join(errs...)
+	// REVIEW(mkcp): This seems like it could be a breaking behavior change. Can I get a second set of eyes here?
+	if 0 < len(errs) {
```
phillebaba, line 56: Flip around, len(errs) should go first.

src/pkg/cluster/zarf.go
phillebaba: This is not breaking as far as I can see. When calling `errors.Join` it will return nil if the slice given is empty. Here is a suggestion for how I would implement this.

```golang
err = errors.Join(errs)
if err != nil {
  return nil, err
}
return deployedPackages, nil
```

phillebaba (changes requested):

src/pkg/cluster/pvc.go
mkcp: Alright i'm going to swap this over to a TODO that summarizes

src/pkg/cluster/zarf.go
mkcp: Ok i have been convinced

src/pkg/cluster/zarf.go
mkcp: Yeah I agree with this

src/cmd/tools/helm/load_plugins.go
mkcp: Yeah I agree with this, I usually do that but I'd seen some empty declarations in various places.

src/pkg/cluster/zarf.go
mkcp: I can get behind this as long as we're consistent. I've learned to allocate with make and goland was flagging it, but afaik it compiles exactly the same.

src/pkg/cluster/zarf.go
mkcp: huh, TIL! resolved in recent commit

phillebaba (approved):

PR #2989, opened 2024-09-11
refactor: pull command

phillebaba:
## Description

This change refactors the pull command in packager.

## Related Issue

Relates to #2969 

## Checklist before merging


src/internal/packager2/packager2.go
```diff
@@ -0,0 +1,225 @@
...
+func Fetch(ctx context.Context, src, dir, shasum string, filter filters.ComponentFilterStrategy) error {
+	u, err := url.Parse(src)
+	if err != nil {
+		return err
+	}
+	if u.Scheme == "" {
+		return errors.New("scheme cannot be empty")
+	}
+	if u.Host == "" {
+		return errors.New("scheme cannot be empty")
```
schristoff, line 42: host

src/internal/packager2/packager2.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return err
+	}
+	desc, err := remote.ResolveRoot(ctx)
+	if err != nil {
+		return fmt.Errorf("could not fetch images index: %w", err)
+	}
+	layersToPull := []ocispec.Descriptor{}
+	skeletonPlatform := zoci.PlatformForSkeleton()
+	if desc.Platform == nil || !(desc.Platform.Architecture == skeletonPlatform.Architecture && desc.Platform.OS == skeletonPlatform.OS) {
```
schristoff, line 113: Should we abstract this logic into a helper function or can it be made more explicit to enhance readability?

src/internal/packager2/packager2.go
```diff
@@ -0,0 +1,225 @@
...
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+
+	"github.com/zarf-dev/zarf/src/api/v1alpha1"
+	"github.com/zarf-dev/zarf/src/config"
+	"github.com/zarf-dev/zarf/src/pkg/layout"
+	"github.com/zarf-dev/zarf/src/pkg/packager/filters"
+	"github.com/zarf-dev/zarf/src/pkg/utils"
+	"github.com/zarf-dev/zarf/src/pkg/zoci"
+)
+
```
schristoff, line 31: Should we move `PullOpts` into here following the guidelines we are setting there?
 #2969

src/cmd/package.go
```diff
@@ -308,15 +311,18 @@ var packagePullCmd = &cobra.Command{
...
-		pkgClient, err := packager.New(&pkgConfig)
+		outputDir := pkgConfig.PullOpts.OutputDirectory
+		if outputDir == "" {
+			wd, err := os.Getwd()
+			if err != nil {
+				return err
+			}
+			outputDir = wd
+		}
+		err := packager2.Fetch(cmd.Context(), args[0], outputDir, pkgConfig.PkgOpts.Shasum, filters.Empty())
```
schristoff, line 322: Very supportive of this change :)
However, I'm unsure if it's worth it to pull in the filters pkg to just pass in a `filters.Empty()` into this function. Could you expand on why this approach was chosen?

src/internal/packager2/packager2.go
```diff
@@ -0,0 +1,225 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+// Package packager2 is the new implementation for packager.
+package packager2
```
schristoff, line 5: Is the file name intended to be packager2?

src/internal/packager2/packager2.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	err = archiver.Archive(allTheLayers, tarPath)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func fetchHTTP(ctx context.Context, src, tarPath, shasum string) error {
+	if !config.CommonOptions.Insecure && shasum == "" {
```
schristoff, line 143: Should we do this in `cmd`?

schristoff (changes requested):

src/cmd/package.go
phillebaba: I do this because mirror will need filter to work.

src/internal/packager2/packager2.go
phillebaba: Yes.

src/internal/packager2/packager2.go
phillebaba: This only applies for HTTP fetch which means that we need to parse source twice. Which is better?

src/internal/packager2/pull.go
```diff
@@ -0,0 +1,234 @@
...
+		return err
+	}
+	// Check checksum if src included one.
+	if shasum != "" {
+		received, err := helpers.GetSHA256OfFile(tarPath)
+		if err != nil {
+			return err
+		}
+		if received != shasum {
+			return fmt.Errorf("shasum mismatch for file %s, expected %s bu got %s ", tarPath, shasum, received)
```
schristoff, line 176:
```suggestion
			return fmt.Errorf("shasum mismatch for file %s, expected %s  got %s ", tarPath, shasum, received)
```

src/internal/packager2/pull.go
```diff
@@ -0,0 +1,234 @@
...
+	})
+	if err != nil {
+		return "", err
+	}
+	if pkg.Metadata.Name == "" {
+		return "", fmt.Errorf("%s does not contain a zarf.yaml", path)
+	}
+
+	arch := config.GetArch(pkg.Metadata.Architecture, pkg.Build.Architecture)
+	if pkg.Build.Architecture == zoci.SkeletonArch {
```
schristoff, line 204: Should we validate that theses are not different instead of making them the same?

phillebaba: This code came from the old code. After reading it I am a bit confused of the purpose? Should the architecture not be whatever the `pkg.Build.Architecture` states?

schristoff (approved):

src/internal/packager2/pull.go
```diff
@@ -0,0 +1,228 @@
...
+	defer os.Remove(tmpDir)
+	tmpPath := filepath.Join(tmpDir, "data.tar.zst")
+
+	switch u.Scheme {
+	case "oci":
+		err := pullOCI(ctx, src, tmpPath, shasum, filter)
+		if err != nil {
+			return err
+		}
+	case "http", "https":
```
AustinAbro321, line 57: Current code supports sget. We are deprecating it in v1. Do we want to deprecate now?

src/internal/packager2/pull.go
```diff
@@ -0,0 +1,228 @@
...
+	_, err = io.Copy(f, resp.Body)
+	if err != nil {
+		return err
+	}
+	received, err := helpers.GetSHA256OfFile(tarPath)
+	if err != nil {
+		return err
+	}
+	if received != shasum {
+		return fmt.Errorf("shasum mismatch for file %s, expected %s but got %s", tarPath, shasum, received)
```
AustinAbro321, line 171: I think this error message looks confusing. Could we instead give the user the url rather than the file. Additionally, I think we should error at the start of this function if shasum == ""

```
shasum mismatch for file /tmp/zarf-3623879537/data.tar.zst, expected but got
             690799dbe8414238e11d4488754eee52ec264c1584cd0265e3b91e3e251e8b1a
```

src/internal/packager2/pull.go
AustinAbro321: The `config.GetArch` function returns the highest priority. First it gets the global cli arg `--architecture` then whichever parameters are called in first, then it gets the default runtime. In this case we call `config.GetArch` with no args in `pullOCI` then we call it again here. 

Given this I think we should instead just use pkg.Build.Architecture without `config.GetArch` here, as we will always want what is actually in the package.

AustinAbro321 (changes requested):

AustinAbro321 (approved):

PR #2993, opened 2024-09-12
chore: directly handle ignored errs and nolint intentionally ignored errs

### LINKED ISSUE #2949: Enable linting ignored
We should explicitly check for all errors. This means never using blank identifiers for returned errors nor not capturing any errors.

The check should be enabled in the linter and all warnings should be fixed.

```yaml
linters-settings:
    errcheck:
        check-blank: true
issues:
    include:
        - EXC0001

```

mkcp:
## Description

This PR relies on enabling additional errcheck rules on golangci-lint and revive to find errors that we've failed to handle or propagate. The most common cases are defers, which we handle by joining them with a err named return (a bit too much magic imo, but is the most airtight in case of a panic). Ideally we handle a specific error that wasn't passed up, but sometimes we have to change the function signature. Lastly, there's times where propagating the error isn't viable or useful and we'll have to log out.

Unfortunately the linter will flag when there's unused nolint checks, so we can't mark intentionally ignored errors with `//nolint:errcheck` until the actual PR that introduces the linter checks into main.

Rules changes are:
```
diff --git a/.golangci.yaml b/.golangci.yaml
index 6423df77..18b3f0f1 100644
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -59,6 +59,7 @@ linters-settings:
   testifylint:
     enable-all: true
   errcheck:
+    check-blank: true
     check-type-assertions: true
     exclude-functions:
       - (*github.com/spf13/cobra.Command).Help
@@ -68,6 +69,7 @@ linters-settings:
 issues:
   # Revive rules that are disabled by default.
   include:
+    - EXC0001
     - EXC0012
     - EXC0013
     - EXC0014
```

## Related Issue

Blocks #3053  
Relates to #2953 

## Checklist before merging


src/pkg/packager/deploy.go
```diff
@@ -467,9 +470,9 @@ func (p *Packager) processComponentFiles(component v1alpha1.ZarfComponent, pkgLo
 
 		// Copy the file to the destination
 		spinner.Updatef("Saving %s", file.Target)
-		err := helpers.CreatePathAndCopy(fileLocation, file.Target)
-		if err != nil {
-			return fmt.Errorf("unable to copy file %s to %s: %w", fileLocation, file.Target, err)
+		err2 := helpers.CreatePathAndCopy(fileLocation, file.Target)
```
AustinAbro321, line 473: why call this err2?

src/pkg/utils/auth.go
```diff
@@ -60,7 +63,6 @@ func credentialParser(path string) ([]Credential, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer file.Close()
```
AustinAbro321, line 63: I believe we still want this to be a defer

src/pkg/utils/auth.go
```diff
@@ -91,7 +97,6 @@ func netrcParser(path string) ([]Credential, error) {
...
```
AustinAbro321, line 94: same as above

src/pkg/utils/network.go
```diff
@@ -96,7 +101,6 @@ func httpGetFile(url string, destinationFile *os.File) error {
 	if err != nil {
 		return fmt.Errorf("unable to download the file %s", url)
 	}
-	defer resp.Body.Close()
```
AustinAbro321, line 99: Should we change this to use the same named parameter strategy above

src/pkg/variables/templates.go
```diff
@@ -130,5 +129,11 @@ func (vc *VariableConfig) ReplaceTextTemplate(path string) error {
 		}
 	}
 
+	// We're done scanning, close our file
+	err = textFile.Close()
```
AustinAbro321, line 133: We should do named parameter defer return to ensure we always get to the close block

AustinAbro321 (changes requested):

src/pkg/packager/deploy.go
mkcp: A general aversion to var shadowing 👀 i don't mind sticking to err though.

src/cmd/destroy.go
```diff
@@ -71,8 +76,11 @@ var destroyCmd = &cobra.Command{
 					return fmt.Errorf("received an error when executing the script %s: %w", script, err)
 				}
 
-				// Try to remove the script, but ignore any errors
-				_ = os.Remove(script)
+				// Try to remove the script, but ignore any errors and debug log them
+				err = os.Remove(script)
+				if err != nil {
+					slog.Debug("Unable to remove script", "script", script, "error", err)
```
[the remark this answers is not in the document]
mkcp: I'm good w/ that, resolved in latest.

src/pkg/packager/deploy.go
AustinAbro321: I think this syntax should accomplish that, though it's not possible for functions with multiple returns
```go
if err := helpers.CreatePathAndCopy(fileLocation, file.Target); err != nil {
	return fmt.Errorf("unable to copy file %s to %s: %w", fileLocation, file.Target, err)
}
```
IMO best to just use err when we are going to return right away, then use different names when we want to use old errors. That way it's clear to the reader when we're keeping around old errors / avoiding shadowing on purpose.

src/internal/gitea/gitea.go
```diff
@@ -60,10 +61,14 @@ func (g *Client) DoRequest(ctx context.Context, method string, path string, body
 	req.Header.Add("accept", "application/json")
 	req.Header.Add("content-type", "application/json")
 	resp, err := g.httpClient.Do(req)
+	// Ensure we close the body of the http client and capture the error
+	defer func() {
+		errClose := resp.Body.Close()
+		err = errors.Join(err, errClose)
+	}()
```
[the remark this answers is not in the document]
mkcp: That's a good catch -- pushed up a fix

src/pkg/variables/templates.go
mkcp: Yeah, that's fair -- these don't have any returns in the way so they should be handled inline, but there's no harm in guarding against panics.

src/pkg/utils/auth.go
mkcp: Mentioned this below, but, while this should get handled linearly without getting ejected from an error there's no harm in guarding against panics w/ a defer

src/pkg/packager/deploy.go
mkcp: yeah i vibe w/ that

src/cmd/package.go
```diff
@@ -477,9 +498,18 @@ func bindCreateFlags(v *viper.Viper) {
 
 	createFlags.IntVar(&pkgConfig.PkgOpts.Retries, "retries", v.GetInt(common.VPkgRetries), lang.CmdPackageFlagRetries)
 
-	createFlags.MarkHidden("output-directory")
-	createFlags.MarkHidden("key")
-	createFlags.MarkHidden("key-pass")
+	errOD := createFlags.MarkHidden("output-directory")
```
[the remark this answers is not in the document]
mkcp: If I understand it right, the cobra stuff has a lack of error handling in a framework we call out -- we could skip these, but it strikes me as a place we'd want to handle if we could and we can at least debug log them.

src/pkg/layout/split.go
```diff
@@ -18,30 +18,37 @@ import (
 )
 
 // splitFile will split the file into chunks and remove the original file.
-func splitFile(srcPath string, chunkSize int) error {
+func splitFile(srcPath string, chunkSize int) (err error) {
 	srcFile, err := os.Open(srcPath)
 	if err != nil {
 		return err
 	}
-	defer srcFile.Close()
```
AustinAbro321, line 26: I know the existing code closes it later on, but I believe we should keep this one and delete the other

src/pkg/layout/split.go
```diff
@@ -18,30 +18,37 @@ import (
...
 		path := fmt.Sprintf("%s.part%03d", srcPath, fileCount+1)
 		dstFile, err := os.OpenFile(path, os.O_CREATE|os.O_RDWR, helpers.ReadAllWriteUser)
 		if err != nil {
 			return err
 		}
-		defer dstFile.Close()
+		defer func(dstFile *os.File) {
+			err2 := dstFile.Close()
+			err = errors.Join(err, err2)
+		}(dstFile)
```
AustinAbro321, line 51: This defer is good, but we close the file again below. I know that's the existing code, but could you delete that other close

src/cmd/destroy.go
```diff
@@ -71,8 +75,11 @@ var destroyCmd = &cobra.Command{
 					return fmt.Errorf("received an error when executing the script %s: %w", script, err)
 				}
 
-				// Try to remove the script, but ignore any errors
-				_ = os.Remove(script)
+				// Try to remove the script, but ignore any errors and debug log them
+				err = os.Remove(script)
+				if err != nil {
+					message.Debug("Unable to remove script", "script", script, "error", err)
```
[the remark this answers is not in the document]
mkcp: Resolved in latest

src/cmd/dev.go
```diff
@@ -208,14 +213,17 @@ var devSha256SumCmd = &cobra.Command{
...
+			err = errors.Join(err, errClose)
+		}(data)
 
 		hash, err := helpers.GetSHA256Hash(data)
 		if err != nil {
 			return errors.Join(hashErr, err)
 		}
 		fmt.Println(hash)
-		return nil
+		return err // Must return err for defer errors.Join
```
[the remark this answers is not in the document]
mkcp: Agreed, also need to remove these because this comment is wrong and returning nil directly still allows you to return deferred error joins.

src/cmd/initialize.go
```diff
@@ -130,6 +138,7 @@ func downloadInitPackage(ctx context.Context, cacheDirectory string) (string, er
 	message.Note(lang.CmdInitPullNote)
 
 	// Prompt the user if --confirm not specified
+	// REVIEW(mkcp): It looks like this condition can't be met - maybe --confirm was removed at some point?
```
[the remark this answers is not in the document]
mkcp: Screenshot 2024-10-01 at 10 07 02 AM
OK, I had a silly mistake here misreading the warning I was getting from goland. It was saying that the if block is unnecessary because it always runs at that point.

src/pkg/layout/split.go
mkcp: Fixed in latest

src/pkg/layout/split.go
mkcp: Yup -- missed the inline close. Deleted in latest

mkcp: This actually caused some issues on windows. We need to close the file before removing it, and that can lead to some resource conflicts between a happy path close-and-delete, and a deferred close that we want to clean up any error paths. I ensured that any deferred close won't return its error if it's `os.ErrClosed` to get around this.

src/pkg/packager/interactive.go
```diff
@@ -21,7 +22,10 @@ import (
 func (p *Packager) confirmAction(stage string, warnings []string, sbomViewFiles []string) bool {
 	pterm.Println()
 	message.HeaderInfof("📦 PACKAGE DEFINITION")
-	utils.ColorPrintYAML(p.cfg.Pkg, p.getPackageYAMLHints(stage), true)
+	err := utils.ColorPrintYAML(p.cfg.Pkg, p.getPackageYAMLHints(stage), true)
+	if err != nil {
+		slog.Error("unable to print yaml", "error", err)
```
AustinAbro321, line 27: We don't use slog yet

src/pkg/utils/network.go
```diff
@@ -87,16 +92,19 @@ func DownloadToFile(ctx context.Context, src, dst, cosignKeyPath string) error {
 		}
 	}
 
-	return nil
+	return err
```
AustinAbro321, line 95: leftover err-> nil I think.

src/pkg/variables/templates_test.go
```diff
@@ -133,10 +133,12 @@ func TestReplaceTextTemplate(t *testing.T) {
...
 
-			f, _ := os.Create(tc.path)
-			defer f.Close()
+			f, err := os.Create(tc.path)
+			require.NoError(t, err)
 
-			_, err := f.WriteString(start)
+			_, err = f.WriteString(start)
+			require.NoError(t, err)
+			err = f.Close()
```
AustinAbro321, line 141: I believe we want to check the err here within the defer block

AustinAbro321 (approved):

PR #3008, opened 2024-09-19
refactor: remove

phillebaba:
## Description

Refactors remove command.

## Related Issue

Relates to #2969 

## Checklist before merging

### DIFF
--- a/src/internal/packager2/remove.go
+++ b/src/internal/packager2/remove.go
@@ -1,0 +1,163 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+package packager2
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"slices"
+
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/cli"
+	"helm.sh/helm/v3/pkg/storage/driver"
+
+	"github.com/zarf-dev/zarf/src/api/v1alpha1"
+	"github.com/zarf-dev/zarf/src/config"
+	"github.com/zarf-dev/zarf/src/pkg/cluster"
+	"github.com/zarf-dev/zarf/src/pkg/message"
+	"github.com/zarf-dev/zarf/src/pkg/packager/actions"
+	"github.com/zarf-dev/zarf/src/pkg/packager/filters"
+	"github.com/zarf-dev/zarf/src/types"
+)
+
+// RemoveOptions are the options for Remove.
+type RemoveOptions struct {
+	Source                  string
+	Cluster                 *cluster.Cluster
+	Filter                  filters.ComponentFilterStrategy
+	SkipSignatureValidation bool
+}
+
+// Remove removes a package that was already deployed onto a cluster, uninstalling all installed helm charts.
+func Remove(ctx context.Context, opt RemoveOptions) error {
+	pkg, err := packageFromSourceOrCluster(ctx, opt.Cluster, opt.Source, opt.SkipSignatureValidation)
+	if err != nil {
+		return err
+	}
+
+	// If components were provided; just remove the things we were asked to remove
+	components, err := opt.Filter.Apply(pkg)
+	if err != nil {
+		return err
+	}
+	// Check that cluster is configured if required.
+	requiresCluster := false
+	componentIdx := map[string]v1alpha1.ZarfComponent{}
+	for _, component := range components {
+		componentIdx[component.Name] = component
+		if component.RequiresCluster() {
+			if opt.Cluster == nil {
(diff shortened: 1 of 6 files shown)


src/internal/packager2/remove.go
```diff
@@ -0,0 +1,160 @@
...
+					message.Warnf("Unable to update the secret for package %s, this may be normal if the cluster was removed: %s", depPkg.Name, err.Error())
+				}
+			}
+			return nil
+		}()
+		if err != nil {
+			err := actions.Run(ctx, comp.Actions.OnRemove.Defaults, comp.Actions.OnRemove.OnFailure, nil)
+			if err != nil {
+				// TODO: I think we should return an error if failure action fails. Does this break existing use cases?
+				return fmt.Errorf("unable to run the failure action: %w", err)
```
phillebaba, line 145: I want to start returning an error if on failure action fails. Before it was a log message. Does this break existing use cases?

schristoff (dismissed):

src/internal/packager2/remove.go
AustinAbro321: It is a breaking change, but I would agree it makes sense to start erroring as we do with the other actions. We should make note of it in the release announcement.

src/pkg/cluster/zarf.go
```diff
@@ -74,6 +74,65 @@ func (c *Cluster) GetDeployedPackage(ctx context.Context, packageName string) (*
...
+	}
+	err = func() error {
+		_, err := c.Clientset.CoreV1().Secrets(packageSecret.Namespace).Get(ctx, packageSecret.Name, metav1.GetOptions{})
+		if err != nil && !kerrors.IsNotFound(err) {
+			return err
+		}
+		if kerrors.IsNotFound(err) {
+			_, err = c.Clientset.CoreV1().Secrets(packageSecret.Namespace).Create(ctx, packageSecret, metav1.CreateOptions{})
+			if err != nil {
+				return fmt.Errorf("unable to create the zarf state secret: %w", err)
```
AustinAbro321, line 110: This secret isn't Zarf state, but the existing remove also uses this text, so up to you if you'd like to change it here or in a separate PR.

src/internal/packager2/remove.go
phillebaba: This will affect people running Zarf in a pipeline as an exit code 1 would most likely stop the pipeline from continuing. I think it is worth causing some problems if it makes it clear to end users that an action actually failed instead of failing silently. I assume that most people are not reading their CI logs meticulously every time to catch the error log.

src/pkg/cluster/zarf.go
phillebaba: No reason not to fix this now.

src/internal/packager2/remove.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+			return err
+		}
+	}
+
+	// All the installed components were deleted, therefore this package is no longer actually deployed
+	if opt.Cluster != nil && len(depPkg.DeployedComponents) == 0 {
+		err := opt.Cluster.DeleteDeployedPackage(ctx, depPkg.Name)
+		if err != nil {
+			message.Warnf("Unable to delete the secret for package %s, tis may be normal if the cluster was removed: %s", depPkg.Name, err.Error())
```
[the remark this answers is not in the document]
phillebaba: Sadly we cant because if the component we are deleting is actually deleting the cluster, we have removed the cluster we want to remove the secret from. So goes the life of chicken and the egg.

src/internal/packager2/remove.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	// Get or build the secret for the deployed package
+	depPkg := &types.DeployedPackage{}
+	if requiresCluster {
+		depPkg, err = opt.Cluster.GetDeployedPackage(ctx, pkg.Metadata.Name)
+		if err != nil {
+			return fmt.Errorf("unable to load the secret for the package we are attempting to remove: %s", err.Error())
+		}
+	} else {
```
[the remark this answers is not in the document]
phillebaba: Yes otherwise we would overwrite the deployed package fetched from the cluster.

src/internal/packager2/remove.go
```diff
@@ -0,0 +1,168 @@
...
+
+// Remove removes a package that was already deployed onto a cluster, uninstalling all installed helm charts.
+func Remove(ctx context.Context, opt RemoveOptions) error {
+	pkg, err := packageFromSourceOrCluster(ctx, opt.Cluster, opt.Source, opt.SkipSignatureValidation)
+	if err != nil {
+		return err
+	}
+
+	// If components were provided; just remove the things we were asked to remove
+	filter := filters.Combine(
```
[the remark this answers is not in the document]
phillebaba: I will move it to cmd to align with the others. I have a plan to refactor filter in the future.

PR #3020, opened 2024-09-23
fix: modify the wait condition according to return values from earlier method call

soltysh:
## Description
Discussed on slack, tl;dr is: currently no matter what `PackageSecretNeedsWait` method returns we just return the package, we need to invert the condition to wait, ie. return an error, to give the `retry.DoWithData` to do its job. 

## Related Issue

None

## Checklist before merging


AustinAbro321 (approved):
Great find

PR #3022, opened 2024-09-23
refactor: layout

phillebaba:
## Description

The change refactors packager layout.

## Related Issue

Relates to #2969 

## Checklist before merging


src/internal/packager2/inspect.go
```diff
@@ -86,23 +89,25 @@ func handleSBOMOptions(ctx context.Context, pkg v1alpha1.ZarfPackage, opt ZarfIn
 	if err != nil {
 		return err
 	}
-	if opt.SBOMOutputDir != "" {
-		out, err := layout.SBOMs.OutputSBOMFiles(opt.SBOMOutputDir, pkg.Metadata.Name)
+
+	sbomDirPath := opt.SBOMOutputDir
+	if sbomDirPath == "" {
```
schristoff, line 94: Why did you choose this path and not to evaluate `opt.SBOMOutputDir` here and then pass through tmpDir in your if below

src/internal/packager2/layout/package.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	err = validatePackageSignature(ctx, pkgLayout, opt.PublicKeyPath, opt.SkipSignatureValidation)
+	if err != nil {
+		return nil, err
+	}
+	return pkgLayout, nil
+}
+
+// Cleanup removes any temporary directories created.
+func (p *PackageLayout) Cleanup() error {
```
schristoff, line 122: Does this increase readability?

src/internal/packager2/layout/package.go
```diff
@@ -0,0 +1,306 @@
...
+	_, err := os.Stat(sourcePath)
+	if err != nil {
+		return "", fmt.Errorf("component %s does not exist in package: %w", componentName, err)
+	}
+	tmpDir, err := utils.MakeTempDir(config.CommonOptions.TempDirectory)
+	if err != nil {
+		return "", err
+	}
+	defer os.RemoveAll(tmpDir)
+	// TODO (phillebaba): We are not using archiver.Extract here because there is a bug in Windows where the files will not be extracted properly from nested directories.
```
schristoff, line 152: Could you add an issue number so this can be tracked?

src/internal/packager2/layout/package.go
```diff
@@ -0,0 +1,306 @@
...
+	err = archiver.Unarchive(sourcePath, tmpDir)
+	if err != nil {
+		return "", err
+	}
+	compPath := filepath.Join(tmpDir, componentName, string(ct))
+	_, err = os.Stat(compPath)
+	if err != nil {
+		return "", fmt.Errorf("component %s does not have a %s directory: %w", componentName, ct, err)
```
schristoff, line 160:
```suggestion
		return "", fmt.Errorf("component %s could not access %s directory: %w", componentName, ct, err)
```
os.Stat could return isnotfound or permissions error, want add clarity here

schristoff (approved):

src/internal/packager2/layout/package.go
phillebaba: Compared to having callers call os.RemoveAll? I think it does, plus it enables us to do more clean up in the future if we need to.

src/internal/packager2/inspect.go
phillebaba: I think this solution remvoes code duplication as we only need a tmpdir when sbom output diir is not set. The previous solution did not require a tmpdir either way which is why i flipped the logic.

schristoff (approved):

Issue #2647, opened 2024-06-20
Several functions in Zarf need to be more context aware and return an error after context is cancelled


AustinAbro321:
With the recent change in #2567 which changed Zarf to cancel context instead of running `os.exit(1)` after the first SIGINT, we need to change functions to use context and return an err up all the way up the stack when context is cancelled.

Note: A second sigint will always `os.exit(1)`

The following is a list of functionality that needs context, but there are likely several more:
- [ ] File downloads on create
- [ ] During data injections on deploy
- [x] Image pull on create - this cannot be properly fixed until we switch from using crane, perhaps to ORAS which implements context, or crane implements context
- [x] helm install / upgrade

AustinAbro321:
Relates to #3434

AustinAbro321:
SBOM creation doesn't respect context cancellation, but we do pass context through to Syft so it's up to them to implement it, they have a recent issue https://github.com/anchore/syft/issues/3705. 

We are not planning to update data injections as it will be replaced by https://github.com/kubernetes/enhancements/issues/4639 once it is stable. 

Once file downloads respect context we can close this issue. If there are additional places users want context cancelled they should make a new issue

PR #3031, opened 2024-09-25
feat: pass context to helm install & upgrade

AustinAbro321:
## Description

Helm has an easy way to pass context through during install and upgrade.  We've since started passing context throughout Zarf, now that it's available in these functions we should pass it through

## Related Issue

Relates to #2647

## Checklist before merging


mkcp (approved):
LGTM

phillebaba (approved):

PR #3039, opened 2024-09-26
test: fix external git flake

### LINKED ISSUE #3024: Test flake - test external fails during port forward
There is a test flake we see during the "test-external" workflow due to docker port forwards. We saw a similar flake in #2837, though the solution there was to avoid port forwarding and replace it with an in memory registry. Finding and using an in memory git server to completely avoid the port forward is an option, though ideally we can find a solution to the flaking without having to re-design the test. 

image

This is the job where this occurred. It happening during the port forward to the mirror command https://github.com/zarf-dev/zarf/actions/runs/11018290255/job/30598211261

AustinAbro321:
### Description

Hard to prove definitively that this will fix the flake, but it does seem like there is at least an issue with the logic of wait check currently. 

## Related Issue

Fixes #3024

^ I figure if it doesn't fix it we can re-open the issue

## Checklist before merging

### DIFF
--- a/.github/workflows/build-rust-injector.yml
+++ b/.github/workflows/build-rust-injector.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: "Checkout Repo"
-        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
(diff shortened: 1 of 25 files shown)


mkcp (approved):
LGTM -- i also ran into some flakiness in the ext_out and ext_in tests, but with the filesystem. In the future, improving test isolation will save us some headaches.

PR #3040, opened 2024-09-26
chore: cleanup errchecking in tests

### LINKED ISSUE #2949: Enable linting ignored

mkcp:
## Description

This PR is split out from #2993 to only fix ignored errors in tests.

## Related Issue

Fixes #2949 
Relates to #2993, #2953 

## Checklist before merging

### DIFF
--- a/src/test/common.go
+++ b/src/test/common.go
@@ -6,7 +6,9 @@ package test
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
+	"log/slog"
 	"os"
 	"regexp"
 	"runtime"
@@ -53,13 +55,16 @@ func GetCLIName() string {
 }
 
 // Zarf executes a Zarf command.
-func (e2e *ZarfE2ETest) Zarf(t *testing.T, args ...string) (string, string, error) {
+func (e2e *ZarfE2ETest) Zarf(t *testing.T, args ...string) (_ string, _ string, err error) {
 	if !slices.Contains(args, "--tmpdir") && !slices.Contains(args, "tools") {
 		tmpdir, err := os.MkdirTemp("", "zarf-")
 		if err != nil {
 			return "", "", err
 		}
-		defer os.RemoveAll(tmpdir)
+		defer func(path string) {
+			errRemove := os.RemoveAll(path)
+			err = errors.Join(err, errRemove)
+		}(tmpdir)
 		args = append(args, "--tmpdir", tmpdir)
 	}
 	if !slices.Contains(args, "--zarf-cache") && !slices.Contains(args, "tools") && os.Getenv("CI") == "true" {
@@ -74,7 +79,10 @@ func (e2e *ZarfE2ETest) Zarf(t *testing.T, args ...string) (string, string, erro
 			return "", "", err
 		}
 		args = append(args, "--zarf-cache", cacheDir)
-		defer os.RemoveAll(cacheDir)
+		defer func(path string) {
+			errRemove := os.RemoveAll(path)
+			err = errors.Join(err, errRemove)
+		}(cacheDir)
 	}
 	return exec.CmdWithTesting(t, exec.PrintCfg(), e2e.ZarfBinPath, args...)
 }
@@ -89,7 +97,10 @@ func (e2e *ZarfE2ETest) Kubectl(t *testing.T, args ...string) (string, string, e
 // CleanFiles removes files and directories that have been created during the test.
 func (e2e *ZarfE2ETest) CleanFiles(files ...string) {
 	for _, file := range files {
-		_ = os.RemoveAll(file)
+		err := os.RemoveAll(file)
+		if err != nil {
+			slog.Debug("Unable to cleanup files", "files", files, "error", err)
+		}
 	}
 }
 
--- a/src/test/common.go
+++ b/src/test/common.go
@@ -8,7 +8,6 @@ import (
 	"bufio"
 	"errors"
 	"fmt"
-	"log/slog"
 	"os"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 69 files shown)


src/test/e2e/14_oci_compose_test.go
```diff
@@ -176,6 +174,12 @@ func (suite *PublishCopySkeletonSuite) Test_2_FilePaths() {
 			isSkeleton = true
 		}
 		suite.verifyComponentPaths(unpacked, components, isSkeleton)
+
+		// Cleanup resources
+		err = os.RemoveAll(unpacked)
```
AustinAbro321, line 179: IMO we should keep these as defers so that if there is a failure during the test the resources still get cleaned up. We can do this to check the error
```go 
		defer func(){
			suite.NoError(os.RemoveAll(unpacked))
		}()
```

AustinAbro321 (changes requested):
Only one comment, but it applies to most of the defers that got deleted. I think we should keep the `defers`, and check errors within them otherwise during when a test fails before the cleanup step there will be no cleanup.

src/test/e2e/14_oci_compose_test.go
mkcp: Thank you for catching that, I don't remember what my justification was for moving those cleanups out of defers.

mkcp:
@AustinAbro321 Defers should be fixed in latest, also added some todos for when we swap over to the linter in main

src/test/e2e/29_config_file_test.go
```diff
@@ -139,30 +143,37 @@ func configFileDefaultTests(t *testing.T) {
 	}
 
 	// Test remaining default initializers
-	t.Setenv("ZARF_CONFIG", filepath.Join("src", "test", "zarf-config-test.toml"))
-	defer os.Unsetenv("ZARF_CONFIG")
+	envKey := "ZARF_CONFIG"
```
schristoff, line 146: nit: Could just set as `const` mayb?

schristoff (dismissed):

src/test/common.go
```diff
@@ -89,7 +97,10 @@ func (e2e *ZarfE2ETest) Kubectl(t *testing.T, args ...string) (string, string, e
 // CleanFiles removes files and directories that have been created during the test.
 func (e2e *ZarfE2ETest) CleanFiles(files ...string) {
 	for _, file := range files {
-		_ = os.RemoveAll(file)
+		err := os.RemoveAll(file)
```
AustinAbro321, line 100: Would we run into issues if we did a `require.NoError` here instead? Given that [removeAll](https://pkg.go.dev/os#RemoveAll) returns nil when a file doesn't exist I would think we wouldn't, but I have not tried it out.

src/test/e2e/21_connect_creds_test.go
```diff
@@ -71,10 +71,12 @@ func TestMetrics(t *testing.T) {
 	client := &http.Client{Transport: tr}
 	httpsEndpoint := strings.ReplaceAll(tunnel.HTTPEndpoint(), "http", "https")
 	resp, err := client.Get(httpsEndpoint + "/metrics")
+	defer func() {
+		require.NoError(t, resp.Body.Close())
```
AustinAbro321, line 75: This should be after the error, as if it's before we can get a nil pointer exception on `resp.Body`

src/test/e2e/14_oci_compose_test.go
```diff
@@ -137,45 +137,54 @@ func (suite *PublishCopySkeletonSuite) Test_2_FilePaths() {
...
-			}
-			manifestDir := filepath.Join(unpacked, "components", "test-compose-package", "manifests")
-			for _, manifest := range kustomizeGeneratedManifests {
-				manifestPath := filepath.Join(manifestDir, manifest)
-				suite.FileExists(manifestPath, "expected to find kustomize-generated manifest: %q", manifestPath)
-				var configMap corev1.ConfigMap
-				err := utils.ReadYaml(manifestPath, &configMap)
-				suite.NoError(err)
-				suite.Equal("ConfigMap", configMap.Kind, "expected manifest %q to be of kind ConfigMap", manifestPath)
+		// Wrap in a fn to ensure our defers cleanup resources on each iteration
```
AustinAbro321, line 140: Don't feel super strongly about this, but I would prefer we don't create the anonymous function here to keep things simple. Either way a defer for each iteration will be called at the end of the function. This kinda reads like something may go wrong if the file isn't deleted at the end of each iteration.

AustinAbro321 (changes requested):
A few more comments, but it's very close. Nice work! I've learned a lot about defer while doing this reviews

src/test/e2e/29_config_file_test.go
mkcp: def, opted to keep it consistent but const was preferred

src/test/common.go
mkcp: That would be ideal, but there wasn't testing.T state here. I'll look into passing it through

src/test/e2e/21_connect_creds_test.go
mkcp: Agreed, it's unlikely but best for this to be watertight.

src/test/e2e/14_oci_compose_test.go
mkcp: The main thing this guards against is large inputs, like files or responses that can overrun memory. It's not needed in every case but good practice for defers in loops.

mkcp:
> A few more comments, but it's very close. Nice work! I've learned a lot about defer while doing this reviews

Should all be reserved (cc @schristoff) and yeah 100% here too. I ended up trying a bunch of different error handling patterns, from manual orders to the named err + join, and definitely added a few tricks to the toolbox. named err + join is probably the closest thing we have in Go to a `Finally` block.

AustinAbro321 (approved):

PR #3042, opened 2024-09-27
refactor: creator

phillebaba:
## Description

Refactors skeleton creator.

A major feature of the refactor is that the implementation no longer requires sitting in the correct directory to build.

## Related Issue

Relates to #2969 

## Checklist before merging


src/internal/packager2/layout/create.go
```diff
@@ -0,0 +1,467 @@
...
+
+	pkg.Build.Architecture = pkg.Metadata.Architecture
+
+	// Record the Zarf Version the CLI was built with.
+	pkg.Build.Version = config.CLIVersion
+
+	// Record the time of package creation.
+	pkg.Build.Timestamp = now.Format(time.RFC1123Z)
+
+	// Record the migrations that will be ran on the package.
```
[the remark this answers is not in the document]
phillebaba: Yes this stops migrations from being run.

src/internal/packager2/layout/create.go
```diff
@@ -0,0 +1,467 @@
...
+		// Strip non-deterministic header data
+		header.ModTime = time.Time{}
+		header.AccessTime = time.Time{}
+		header.ChangeTime = time.Time{}
+		header.Uid = 0
+		header.Gid = 0
+		header.Uname = ""
+		header.Gname = ""
+
+		header.Mode = header.Mode &^ 0o077
```
schristoff, line 436: `&^ 0o077` -> should put a comment on why this looks weird

phillebaba, line 436: Explain why this works.

src/internal/packager2/layout/create.go
```diff
@@ -0,0 +1,467 @@
...
+				return fmt.Errorf("unable to build kustomization %s: %w", path, err)
+			}
+		}
+
+		// remove kustomizations
+		component.Manifests[manifestIdx].Kustomizations = nil
+	}
+
+	// Write the tar component.
+	size, err := helpers.GetDirSize(compBuildPath)
```
schristoff, line 267: Confused on why this is occurring, and can this just be replaced with an `os.Stat`?

src/internal/packager2/layout/create.go
```diff
@@ -0,0 +1,467 @@
...
+	tw := tar.NewWriter(tb)
+	defer tw.Close()
+
+	// Walk through the directory and process each file
+	return filepath.Walk(dirPath, func(filePath string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		link := ""
```
schristoff, line 413:
```suggestion
		var link string 
```

src/pkg/packager/publish.go
```diff
@@ -48,26 +50,36 @@ func (p *Packager) Publish(ctx context.Context) (err error) {
 	}
 
 	if p.cfg.CreateOpts.IsSkeleton {
```
schristoff, line 52: Does this have to be changed?

src/internal/packager2/layout/import.go
```diff
@@ -0,0 +1,408 @@
...
+	// Deprecated
+	// Override cosign key path if it was provided.
+	if override.DeprecatedCosignKeyPath != "" {
+		c.DeprecatedCosignKeyPath = override.DeprecatedCosignKeyPath
+	}
+
+	c.DeprecatedGroup = override.DeprecatedGroup
+
+	// Merge deprecated scripts for backwards compatibility with older zarf binaries.
+	c.DeprecatedScripts.Before = append(c.DeprecatedScripts.Before, override.DeprecatedScripts.Before...)
```
schristoff, line 249: small nit: i think `overrideDeprecated` and `overrideActions`  was easier to consume as two functions

src/internal/packager2/layout/create.go
phillebaba: Interesting matter of taste, I dont mind changing if this is what you prefer.

src/pkg/packager/publish.go
phillebaba: No we can leave changing this until after publish has been refactored if you prefer.

src/internal/packager2/layout/create.go
phillebaba: This could be replaced by `os.ReadDir` as we dont really care about the size, just that the directory is not empty.

src/internal/packager2/layout/create.go
```diff
@@ -0,0 +1,467 @@
...
+			return "", err
+		}
+	}
+
+	checksumContent, checksumSha, err := getChecksum(buildPath)
+	if err != nil {
+		return "", err
+	}
+	checksumPath := filepath.Join(buildPath, Checksums)
+	err = os.WriteFile(checksumPath, []byte(checksumContent), helpers.ReadWriteUser)
```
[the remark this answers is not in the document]
phillebaba: Set all to ReadWriteUser.

src/pkg/packager/publish.go
schristoff: I think that would be best

src/internal/packager2/layout/create.go
```diff
@@ -0,0 +1,475 @@
...
+			return fmt.Errorf("unable to copy data injection %s: %s", data.Source, err.Error())
+		}
+
+		component.DataInjections[dataIdx].Source = rel
+	}
+
+	// Iterate over all manifests.
+	for manifestIdx, manifest := range component.Manifests {
+		for fileIdx, path := range manifest.Files {
+			rel := filepath.Join(string(ManifestsComponentDir), fmt.Sprintf("%s-%d.yaml", manifest.Name, fileIdx))
```
AustinAbro321, line 240: This is an issue with the current skeleton, but packages that have remote manifests will break here. I believe a `helpers.IsUrl()` check will solve this

phillebaba: If that is a current issue with Skeleton packages I think it is fine if the new implementation also has it. We need to set the DoD somewhere.

schristoff (dismissed):

src/internal/packager2/layout/import.go
```diff
@@ -0,0 +1,414 @@
...
+	rel, err := filepath.Rel(abs, dir)
+	if err != nil {
+		return "", err
+	}
+	return rel, nil
+}
+
+func overrideMetadata(comp v1alpha1.ZarfComponent, override v1alpha1.ZarfComponent) (v1alpha1.ZarfComponent, error) {
+	// Metadata
+	comp.Name = override.Name
```
AustinAbro321, line 227: nit: name will always be the same here so no reason to override

src/internal/packager2/layout/create.go
AustinAbro321: Fair enough, I'll make an issue

src/internal/packager2/layout/import.go
AustinAbro321: eh current code is like this too

src/internal/packager2/layout/create.go
```diff
@@ -0,0 +1,475 @@
...
+			if err := helpers.CreatePathAndCopy(filepath.Join(packagePath, path), dst); err != nil {
+				return fmt.Errorf("unable to copy manifest %s: %w", path, err)
+			}
+
+			component.Manifests[manifestIdx].Files[fileIdx] = rel
+		}
+
+		for kustomizeIdx, path := range manifest.Kustomizations {
+			// Generate manifests from kustomizations and place in the package.
+			kname := fmt.Sprintf("kustomization-%s-%d.yaml", manifest.Name, kustomizeIdx)
```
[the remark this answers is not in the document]
AustinAbro321: I'll make an issue for this

src/internal/packager2/layout/create.go
AustinAbro321: #3082

schristoff (approved):

PR #3043, opened 2024-09-27
feat: wait using kstatus

### LINKED ISSUE #2855: Image Race Condition Between Zarf Seed Registry and Zarf Permanent Registry With Multiple Registry Replicas

### LINKED ISSUE #2973: Introduce `wait` key to run health checks for all deployed resources
In #2678 health checks were introduced to allow users to wait for their resources to be reconciled.

When wait is true we would wait for all the resources applied to the cluster. This is the preferred method that we would want users to use.
```yaml
kind: ZarfPackageConfig
metadata:
  name: argocd
  description: Example showcasing installing ArgoCD
components:
  - name: argocd-helm-chart
    required: true
    charts:
      - name: argo-cd
        version: 5.54.0
        namespace: argocd
        url: https://argoproj.github.io/argo-helm
        releaseName: argocd-baseline
        valuesFiles:
          - baseline/values.yaml
    images:
      - docker.io/library/redis:7.0.15-alpine
      - quay.io/argoproj/argocd:v2.9.6
      # Cosign artifacts for images - argocd - argocd-helm-chart
      - quay.io/argoproj/argocd:sha256-2dafd800fb617ba5b16ae429e388ca140f66f88171463d23d158b372bb2fae08.sig
      - quay.io/argoproj/argocd:sha256-2dafd800fb617ba5b16ae429e388ca140f66f88171463d23d158b372bb2fae08.att
    wait: true
```

 In the v1alpha1 schema `wait` will default to false and helm / manifests `noWait` will default to false. However in the v1beta1 schema `wait` will default to true while helm / manifests `wait` default  to false

AustinAbro321:
## Description

This PR changes the existing wait functionality to have use kstatus after helm wait

## Related Issue

Fixes #2973
Fixes #2855

## Checklist before merging


phillebaba:
I think it would be good to have new packages be in internal instead of pkg initially if there isn't an explicit interest of exporting the functionality for others to use outside of Zarf.

src/internal/packager/helm/chart.go
```diff
@@ -118,6 +121,33 @@ func (h *Helm) InstallOrUpgradeChart(ctx context.Context) (types.ConnectStrings,
...
+			APIVersion: apiVersion,
+			Kind:       kind,
+			Name:       resource.Name,
+			Namespace:  resource.Namespace,
+		})
+	}
+	if !h.chart.NoWait {
+		// This re-uses the timeout from helm. This will increase the total amount of time a timeout can take
+		// However it is unlikely this step will take long
+		healthChecksCtx, cancel := context.WithTimeout(ctx, h.timeout)
```
AustinAbro321, line 142: Want to get the teams thoughts here. This way is much simpler, but may result in unexpected behavior. 

If we want to use the same context as the successful helm install, I believe we would have to use `retry.DoWithData` and make a custom object to send back both the context and the cancel function

phillebaba (commented):
Another small comment I would move the health check package out of packager.

src/internal/packager/helm/chart.go
```diff
@@ -118,6 +123,33 @@ func (h *Helm) InstallOrUpgradeChart(ctx context.Context) (types.ConnectStrings,
...
+		healthChecks = append(healthChecks, v1alpha1.NamespacedObjectKindReference{
+			APIVersion: apiVersion,
+			Kind:       kind,
+			Name:       resource.Name,
+			Namespace:  resource.Namespace,
+		})
+	}
+	if !h.chart.NoWait {
+		// Ensure we don't go past the timeout by getting the time since the helm operation started
+		healthCheckTimeout := h.timeout - time.Since(helmOpStart)
```
phillebaba, line 143: I understand the idea but it seems like it makes things complicated. Could we create a single context with timeout before install chart, that is used for both chart installing and health checking?

AustinAbro321: I went back and forth on this. The retry loop makes context a bit complicated here. We want to initialize the timeout during each retry, but not cancel until the end of the function. I changed it the code to use the context, lmk what you think.

src/internal/packager/helm/chart.go
```diff
@@ -58,28 +61,35 @@ func (h *Helm) InstallOrUpgradeChart(ctx context.Context) (types.ConnectStrings,
...
 
 		releases, histErr := histClient.Run(h.chart.ReleaseName)
 
 		spinner.Updatef("Checking for existing helm deployment")
-
+		// cancel context from previous retry if it exists
+		if helmCtxCancel != nil {
+			helmCtxCancel()
+		}
+		helmCtx, helmCtxCancel = context.WithTimeout(ctx, h.timeout)
```
phillebaba, line 79: I think everything would be simpler if we did not reset the timeout every time. Is there a reason we  do not create a single context with timeout in the beginning of the function call?

src/internal/packager/helm/chart.go
```diff
@@ -90,6 +96,7 @@ func (h *Helm) InstallOrUpgradeChart(ctx context.Context) (types.ConnectStrings,
 		spinner.Success()
 		return nil
 	}, retry.Context(ctx), retry.Attempts(uint(h.retries)), retry.Delay(500*time.Millisecond))
+	defer helmCtxCancel()
```
phillebaba, line 99: Why cancel twice?

src/internal/packager/helm/chart.go
AustinAbro321: I did not originally do this because it affects the timeout on retries. Given how successful retries seemingly never happen in this situation however, I agree it's better to go the simpler route

src/internal/packager/helm/chart.go
AustinAbro321: Good catch thanks

phillebaba (approved):

PR #3044, opened 2024-09-27
feat: add `metadata.annotations` to `v1alpha1` schema

marshall007:
## Description

This adds `metadata.annotations` from `v1beta1` to the existing `v1alpha1` schema. It also adds them to the OCI config descriptor and image manifests.

## Related Issue

Relates to https://github.com/zarf-dev/zarf/issues/2976.

## Checklist before merging

### DIFF
--- a/src/pkg/zoci/push.go
+++ b/src/pkg/zoci/push.go
@@ -7,6 +7,7 @@ package zoci
 import (
 	"context"
 	"fmt"
+	"maps"
 
 	"github.com/defenseunicorns/pkg/helpers/v2"
 	"github.com/defenseunicorns/pkg/oci"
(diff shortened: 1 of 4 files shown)


src/pkg/zoci/push.go
```diff
@@ -110,5 +111,8 @@ func annotationsFromMetadata(metadata *v1alpha1.ZarfMetadata) map[string]string
 		annotations[ocispec.AnnotationVendor] = vendor
 	}
 
+	// annotations explicitly defined in `metadata.annotations` take precedence over legacy fields
+	maps.Copy(annotations, metadata.Annotations)
+
```
marshall007, line 116: I could not find any existing tests for populating annotations on the OCI manifest and config. Happy to add some if there is an obvious place to do that.

AustinAbro321: Yeah there aren't currently any tests that verify that annotations end up on the manifest or manifest config

AustinAbro321 (approved):

phillebaba (changes requested):
We need to discuss if this should be added to a new API Version or not before merging this.

schristoff:
Thank you for this contribution, but due to the work we are planning on v1.0 this may conflict with our own implementations.

PR #3064, opened 2024-10-03
fix: support registry override for dev deploy (yolo)

mjnagel:
## Description

Adds a new hidden flag `--registry-url` which can be used on dev deploy (yolo only) to set the value to use for `ZARF_REGISTRY`.

## Related Issue

Fixes https://github.com/zarf-dev/zarf/issues/2713

## Checklist before merging


src/types/runtime.go
```diff
@@ -80,6 +80,8 @@ type ZarfDeployOptions struct {
 	Timeout time.Duration
 	// [Library Only] A map of component names to chart names containing Helm Chart values to override values on deploy
 	ValuesOverridesMap map[string]map[string]map[string]interface{}
+	// [Dev Deploy Only] Manual override for ###ZARF_REGISTRY###
+	RegistryURL string
```
mjnagel, line 84: Wanted to avoid touching too many other spots, but this might make sense to pull into a separate `ZarfDevDeployOptions`, along with `NoYolo` (currently under create). Open to thoughts on this and could make that change if desired.

src/test/e2e/99_yolo_test.go
```diff
@@ -53,9 +53,17 @@ func TestDevDeploy(t *testing.T) {
 		return
 	}
 
+	// Generic test of dev deploy
 	stdOut, stdErr, err := e2e.Zarf(t, "dev", "deploy", "examples/dos-games")
 	require.NoError(t, err, stdOut, stdErr)
 
-	stdOut, stdErr, err = e2e.Zarf(t, "package", "remove", "dos-games", "--confirm")
+	stdOut, stdErr, err = e2e.Zarf(t, "tools", "kubectl", "delete", "namespace", "dos-games")
+	require.NoError(t, err, stdOut, stdErr)
```
mjnagel, line 61: I might just be crazy but I don't think `package remove` works on something that was `dev deploy`ed, so I switched this to just delete the namespace instead...

schristoff (approved):

PR #3067, opened 2024-10-03
(feat): Remove progress bar

schristoff:
This could help with #2104

### DIFF
--- a/src/internal/packager/images/common.go
+++ b/src/internal/packager/images/common.go
@@ -13,7 +13,6 @@ import (
 	"github.com/google/go-containerregistry/pkg/crane"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/zarf-dev/zarf/src/config"
-	"github.com/zarf-dev/zarf/src/pkg/message"
 	"github.com/zarf-dev/zarf/src/pkg/transform"
 	"github.com/zarf-dev/zarf/src/types"
 )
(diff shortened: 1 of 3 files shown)


src/internal/packager/images/common.go
```diff
@@ -107,9 +107,9 @@ func createPushOpts(cfg PushConfig, pb *message.ProgressBar) []crane.Option {
 	// TODO (@WSTARR) This is set to match the TLSHandshakeTimeout to potentially mitigate effects of https://github.com/zarf-dev/zarf/issues/1444
 	transport.ResponseHeaderTimeout = 10 * time.Second
 
-	transportWithProgressBar := helpers.NewTransport(transport, pb)
+	transportWithoutProgressBar := helpers.NewTransport(transport, nil)
```
AustinAbro321, line 110: The helpers.NewTransport is just [oras.NewTransport](https://github.com/oras-project/oras-go/blob/main/registry/remote/retry/client.go#L45) with a progressbar. Should we switch to that instead?

schristoff: I'm trying to keep the change minimal rn to see if it helps with the issue, however if this stays longterm we should do that

AustinAbro321:
How would you feel about adding a message.info like we do in the new mirror? I know some users have many large images / have 100s of images. IMO we'll want to give some indication that the system is progressing
https://github.com/zarf-dev/zarf/blob/e1a99f67636ac40f8f2d1a260f5a4692bcfeebd8/src/internal/packager2/mirror.go#L126

schristoff:
closing due to internal discussion

PR #3075, opened 2024-10-07
refactor: server side apply over create / update

### LINKED ISSUE #2626: Use Server Side Apply when creating Kubernetes resources

We have a couple of places in Zarf where we need to do a create or update on a resource. This is handled in a couple of different ways currently. Either the resource is deleted first and then created. Alternatively we use a verbose logic to try creating and falling back to an update if the resource already exists.

A much better way is to switch over to SSA instead. This moves any responsibility to the API server instead, removing any need for us to deal with checking if a resource exists or not first

https://github.com/defenseunicorns/zarf/blob/7415967772410ceeb77aae7b61367dad85b1516f/src/pkg/cluster/state.go#L114-L127

### Additional context

https://kubernetes.io/docs/reference/using-api/server-side-apply/
https://kubernetes.io/blog/2022/10/20/advanced-server-side-apply/

AustinAbro321:
## Description

There are several situations where we create, check if the resource already existed, and if it does then we update. It is simpler in this situation to just use server side apply

## Related Issue

Fixes #2626

## Checklist before merging


src/pkg/cluster/testdata/expected-injection-pod.json
AustinAbro321, line 1: expanded this so it'd be easier to read / edit. The only differences is that creationTimestamp and status no longer show up in the json (because the apply objects use pointers instead of values)

src/pkg/cluster/testdata/expected-injection-service.json
AustinAbro321, line 1: Expanded for clarity, but content is the same

src/internal/packager/helm/post-render.go
```diff
@@ -54,8 +54,17 @@ func (h *Helm) newRenderer(ctx context.Context) (*renderer, error) {
 	if kerrors.IsNotFound(err) {
 		rend.namespaces[h.chart.Namespace] = cluster.NewZarfManagedNamespace(h.chart.Namespace)
 	} else if h.cfg.DeployOpts.AdoptExistingResources {
-		namespace.Labels = cluster.AdoptZarfManagedLabels(namespace.Labels)
-		rend.namespaces[h.chart.Namespace] = namespace
+		b, err := json.Marshal(namespace)
```
AustinAbro321, line 57: This feels a bit strange, but it seems to be the best way to get then apply a resource with ssa, since the get is a different type

The library has `extractresource` functions, however these will only get resources from the managed fields. Because we're just switching to server side apply now, existing resources will not have managed fields. Additionally, in the case of adoptexistingresources, we can't assume we previously owned the field

https://kubernetes.io/blog/2021/08/06/server-side-apply-ga/#using-server-side-apply-in-a-controller

AustinAbro321: Ended up avoiding the get->apply flows

src/pkg/cluster/state.go
```diff
@@ -89,6 +90,9 @@ func (c *Cluster) InitZarfState(ctx context.Context, initOptions types.ZarfInitO
 		}
 		// Mark existing namespaces as ignored for the zarf agent to prevent mutating resources we don't own.
 		for _, namespace := range namespaceList.Items {
+			if namespace.Name == "zarf" {
```
phillebaba, line 93: Why is this being added?

phillebaba (commented):
Small comment. Additionally have we looked into if it is possible to set a global field manager?

AustinAbro321:
@phillebaba As far as I can tell it's not possible to set a global field manager. I thought about making a function like `cluster.DefaultApplyOpts()`, lmk if you'd prefer that

src/pkg/cluster/state.go
AustinAbro321: Apply doesn't delete existing labels, but adds to them. Whereas update fully replaces the object. Previously if the Zarf namespace already existed on init, we would add the `zarf-agent: ignore` to it, but it would get replaced in the update a few lines down. Now that we apply, the `zarf` ns would keep the `zarf-agent: ignore` label, so I excluded it from receiving the label

phillebaba:
For now just make a constant called something like `FieldManagerName` so that we can track where it is being used.

src/pkg/cluster/secrets_test.go
```diff
@@ -192,7 +192,7 @@ func TestUpdateZarfManagedSecrets(t *testing.T) {
 					},
 				},
 				Type: corev1.SecretTypeOpaque,
-				Data: map[string][]byte{},
+				Data: nil,
```
phillebaba, line 195: No Need to set nil if this is supposed to be nil.

src/internal/packager/helm/post-render.go
```diff
@@ -164,49 +163,17 @@ func (r *renderer) adoptAndUpdateNamespaces(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
-		// TODO: Refactor as error is not checked instead of checking for not found error.
-		currentRegistrySecret, _ := c.Clientset.CoreV1().Secrets(name).Get(ctx, config.ZarfImagePullSecretName, metav1.GetOptions{})
-		sameSecretData := maps.EqualFunc(currentRegistrySecret.Data, validRegistrySecret.Data, func(v1, v2 []byte) bool { return bytes.Equal(v1, v2) })
-		if currentRegistrySecret.Name != config.ZarfImagePullSecretName || !sameSecretData {
```
phillebaba, line 170: What is the effect of removing this logic?

AustinAbro321: The current code only updates the secret if the name or the data is different. Maybe it was there to preserve labels and annotations? I'm not sure, but that's no longer a worry with apply at least. 

Additionally, the current code updates the git server secret if the registry secret has a different name or different data, this seems like it was a bug with indentation, but I'm not 100% sure. I don't see any reason for that logic. 

Still, I should have made a unit test to verify the behavior, I will do that now.

src/pkg/cluster/secrets.go
```diff
@@ -138,11 +117,8 @@ func (c *Cluster) UpdateZarfManagedImageSecrets(ctx context.Context, state *type
 		if err != nil {
 			return err
 		}
-		if maps.EqualFunc(currentRegistrySecret.Data, newRegistrySecret.Data, func(v1, v2 []byte) bool { return bytes.Equal(v1, v2) }) {
```
AustinAbro321, line 141: Given that apply operations do not cost much, even if the data is the same, IMO it's better to simplify the code and let the k8s server figure out what to change / keep the same.

src/internal/packager/helm/post-render.go
AustinAbro321: Actually, I am realizing the existing logic is flawed. I don't think there's much value in testing it. I could separate the new logic and write a test for it, but since it's just two apply calls, I'd pretty much just be testing that the fake client works. 

Additionally, after evaluating further, there is no reason to dog sled errors here. We always want both of these secrets created

phillebaba (approved):

PR #3085, opened 2024-10-09
POC: attribute redactions with slog

AustinAbro321:
## Description

This is a POC of how attribute redactions might look with slog. 

This changes two places where Zarf prints. The first is in the function `debugPrintZarfState`. Here is an example of what that function would print now. 
```bash
state.ZarfAppliance=true state.distro=k3s state.architecture=amd64 state.storageClass=standard state.agentTLS.ca=REDACTED state.agentTLS.cert=REDACTED state.agentTLS.key=REDACTED state.gitServer.pushUsername=friend state.gitServer.pushPassword=**sanitized** state.gitServer.pullUsername=other-friend state.gitServer.pullPassword=**sanitized** state.gitServer.address="" state.registryInfo.pushUsername=friend state.registryInfo.pushPassword=**sanitized** state.registryInfo.pullUsername=other-friend state.registryInfo.pullPassword=**sanitized** state.registryInfo.address="" state.registryInfo.nodePort=0 state.registryInfo.secret=**sanitized** state.artifactServer.pushUsername="" state.artifactServer.pushToken=**sanitized** state.artifactServer.address=whatever
```
The second is in the function `PrintCredentialUpdates` which now may look like
```bash
msg="Artifact info" "URL Address.old"=whatever "URL Address.new"=cooler-address "Push Username"="(unchanged) cool-guy" "Push Token.old"=REDACTED "Push Token.new"=REDACTED
```

While this is clean in that any time any of the objects are printed we don't have to remember to redact, it does force us to implement LogValue functions on every object that includes a ZarfPassword sub object. This will also require us to update the object functions when we want something new on them

There is a lot of noise in the PR since I changed types. Relevant files are: 
- src/pkg/message/credentials.go
- src/pkg/cluster/state.go
- src/types/k8s.go

## Checklist before merging


mkcp:
Nice work investing this! I like that this offers another concrete benefit for elevating random values into their own types.

AustinAbro321:
Decided we're not going to use this strategy for now. Zarf already has sanitation logic in place, this doesn't offer enough benefit to switch that out.

PR #3094, opened 2024-10-12
fix: issue where kustomization files are empty

### LINKED ISSUE #3065: The `kustomizations` does not seem to render out correctly for discoverying images
### Environment
Device and OS: fedora
App version: fluxcd `2.4.0`
Kubernetes distro being used: rke2, 1.31.1
Other: zarf `0.40.1`

### Steps to reproduce
0. Install [`fluxcd`](https://github.com/fluxcd/flux2/releases/tag/v2.4.0)
1. Create `zarf.yaml` with the following contents:

-   ```yaml
    kind: ZarfPackageConfig
    metadata:
      name: fluxcd-install
    components:
      - name: flux-install
        required: true
        manifests:
        - name: flux-install
          namespace: flux-system
          noWait: true
          kustomizeAllowAnyDirectory: true
          kustomizations:
          - ./.
        images: []
    ```

2. Create `kustomization.yaml` with the following contents:

-   ```yaml
    kind: Kustomization
    resources:
    - gotk-components.yml
    ```

3. run: `flux install --export > gotk-components.yml`
4. run: `zarf dev find-images`

### Expected result

That `zarf` discovers the images for fluxcd and its signatures.

- ghcr.io/fluxcd/source-controller:v1.4.1
- ghcr.io/fluxcd/kustomize-controller:v1.4.0
- ghcr.io/fluxcd/helm-controller:v1.1.0
- ghcr.io/fluxcd/notification-controller:v1.4.0

### Actual Result

```shell
 NOTE  Using build directory .
  ✔  Looking for images in component "flux-install" across 0 resources

components:
```

a1994sc:
## Description

This PR addresses a where `zarf` is unable to properly discover the images from a kustomization file.

The culprit from my testing seems to be cause by `f` and `newDestination` being the same values when calling `helpers.CreatePathAndCopy` [here](https://github.com/zarf-dev/zarf/blob/main/src/pkg/packager/prepare.go#L230)

```go
	...
	if err := helpers.CreatePathAndCopy(f, newDestination); err != nil {
	...
```

That function, found [here](https://github.com/defenseunicorns/pkg/blob/main/helpers/files.go#L108), basically copies a file from one path to another, however there is no check if the `source` and `destination` are same values. This causes the resulting file to be empty.

The change in my PR uses the `temp` folder already created by `zarf` to make sure that source and destination files are different, preventing an empty file from being scanned.

## Related Issue

Fixes #3065

## Checklist before merging


src/pkg/packager/prepare.go
```diff
@@ -209,7 +209,8 @@ func (p *Packager) findImages(ctx context.Context) (map[string][]string, error)
 			for idx, k := range manifest.Kustomizations {
 				// Generate manifests from kustomizations and place in the package
 				kname := fmt.Sprintf("kustomization-%s-%d.yaml", manifest.Name, idx)
-				destination := filepath.Join(componentPaths.Manifests, kname)
+				// Use the temp folder because "helpers.CreatePathAndCopy" will with the same path will result in the file being empty
```
a1994sc, line 212: Just realized grammatical error in my comment, I will update it as soon as I can

phillebaba:
@a1994sc how much effort do you think it would be to add a test for this fix?

a1994sc:
I would have look into how the existing testing is done, so I'm unsure

a1994sc:
@phillebaba  I updated the PR with a basic unit test.

AustinAbro321 (approved):

PR #3099, opened 2024-10-15
Test checksums from package create

phillebaba:
## Description

This change adds a new workflow which verifies the checksums of all the examples. The purpose of this is to avoid regression related to package creation.

## Related Issue

Relates to #2969

## Checklist before merging


phillebaba:
I have removed checking checksums for packages with Git repositories as they will always produce components with different shasums due to Git modifying the metadata. I have also removed checking the checksum of the SBOM as it will also produce new results on each build.

Even if we don't check everything in this test it is a good step forward to making sure a PR does not change how a package is built in unexpected ways.

AustinAbro321 (approved):

PR #3106, opened 2024-10-16
feat: move pkg oci to Zarf and log during health checks

AustinAbro321:
## Description
This moves the [pkg kubernetes functions](https://github.com/defenseunicorns/pkg/tree/main/kubernetes) into Zarf to increase our flexibility in changes, and make the release lifecycle quicker. 

Additionally this adds a logger to health checks, so when something goes wrong, it's clear to the user which resources were not ready

## Checklist before merging


src/internal/healthchecks/healthchecks.go
```diff
@@ -32,9 +38,97 @@ func Run(ctx context.Context, watcher watcher.StatusWatcher, healthChecks []v1al
...
+
+	for _, id := range objs {
+		rs := statusCollector.ResourceStatuses[id]
+		switch rs.Status {
+		case status.CurrentStatus:
+			message.Debugf("%s: %s ready", rs.Identifier.Name, strings.ToLower(rs.Identifier.GroupKind.Kind))
+		case status.NotFoundStatus:
+			message.Warnf("%s: %s not found", rs.Identifier.Name, strings.ToLower(rs.Identifier.GroupKind.Kind))
+		default:
+			message.Warnf("%s: %s not ready", rs.Identifier.Name, strings.ToLower(rs.Identifier.GroupKind.Kind))
```
phillebaba, line 94: I think all of these should be Debug to not pollute the output even more.

AustinAbro321: I disagree, when either of these situations are encountered a generic error will be returned and the deployment will fail. In that case I think we should communicate what went wrong.

AustinAbro321: Maybe instead of these logs we should craft a custom error message with the not found / not ready resources?

AustinAbro321: Changed to return the not found and not ready resources within the error

src/internal/healthchecks/healthchecks.go
```diff
@@ -32,9 +39,100 @@ func Run(ctx context.Context, watcher watcher.StatusWatcher, healthChecks []v1al
...
+	}
+
+	// Only check parent context error, otherwise we would error when desired status is achieved.
+	if ctx.Err() != nil {
+		errs := []error{}
+		for _, id := range objs {
+			rs := statusCollector.ResourceStatuses[id]
+			switch rs.Status {
+			case status.CurrentStatus:
+				message.Debugf("%s: %s ready", rs.Identifier.Name, rs.Identifier.GroupKind.Kind)
```
phillebaba, line 98: Could we remove this?

phillebaba (dismissed):

src/internal/healthchecks/healthchecks.go
AustinAbro321: yes

phillebaba (approved):

PR #3108, opened 2024-10-16
feat: add logger package and tests

mkcp:
## Description
This PR adds a logger package using `log/slog` and tests. It's extracted from the proof of concept in #3101.

## Related Issue
Relates to #2576 

## Checklist before merging


src/pkg/logger/logger.go
```diff
@@ -0,0 +1,147 @@
...
+
+import (
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"strings"
+)
+
+// Level declares each supported log level. These are 1:1 what log/slog supports by default. Info is the default level.
```
AustinAbro321, line 15: If we are 1:1 with the slog levels should we just use `slog.Level`? IMO we can always introduce our own levels if needed in the future, but better to stay simple for now

mkcp: logger.Level is part of the config and external API, so if we want to change it in the future then backwards compatibility becomes an issue. It's simpler in the long run to add the type now and always have the option to add more or change its implementation later.

src/pkg/logger/logger.go
```diff
@@ -0,0 +1,147 @@
...
+type Level int
+
+var (
```
AustinAbro321, line 18: FYI if you add a comment at the top of a constant / var block revive will be happy

src/pkg/logger/logger.go
```diff
@@ -0,0 +1,147 @@
...
+// ToLower takes a Format string and converts it to lowercase for case-agnostic validation. Users shouldn't have to care
+// about "json" vs. "JSON" for example - they should both work.
+func (f Format) ToLower() Format {
+	return Format(strings.ToLower(string(f)))
+}
+
+// TODO(mkcp): Add dev format
+var (
+	// FormatEmpty means no format was supplied. This is equivalent to a default, or "Text".
+	FormatEmpty Format = "" //nolint:revive
```
AustinAbro321, line 63: unneeded nolint

src/pkg/logger/logger.go
AustinAbro321: Yeah that's a fair point

src/pkg/logger/logger.go
mkcp: revive was barking at this for storing an empty value. thinking thru it more, it's fine if we just check for empty string in the switch case.

src/pkg/logger/logger.go
mkcp: Yeah that is way better

src/pkg/logger/logger.go
```diff
@@ -0,0 +1,147 @@
...
+// ToLower takes a Format string and converts it to lowercase for case-agnostic validation. Users shouldn't have to care
+// about "json" vs. "JSON" for example - they should both work.
+func (f Format) ToLower() Format {
+	return Format(strings.ToLower(string(f)))
+}
+
+// TODO(mkcp): Add dev format
+var (
+	// FormatText uses the standard slog TextHandler
+	FormatText Format = "text"
```
phillebaba, line 64: What is the difference between text and dev?

src/pkg/logger/logger.go
```diff
@@ -0,0 +1,147 @@
...
+	default:
+		return nil, fmt.Errorf("unsupported log format: %s", cfg.Format)
+	}
+
+	log := slog.New(handler)
+	return log, nil
+}
+
+// Default gets a logger from the atomic slog default.
+func Default() *slog.Logger {
```
phillebaba, line 140: Do we still want a global logger when we are going to pass the logger through a context?

src/pkg/logger/logger.go
mkcp: It would point towards different handlers, the default text output does not have line numbers or pretty-printed structs, but a dev readout would provide more info. Think of it like "debug" format

schristoff (dismissed):

mkcp:
I reintroduced the package level default logger here rather than passing into slog's default, and added more clarity on how it's intended to be used as a development aid.

When adding a `Default().Info()` call to init(),
Screenshot 2024-10-17 at 12 08 29

we get a usable default logger:
Screenshot 2024-10-17 at 12 09 47

schristoff (approved):
beautiful

PR #3114, opened 2024-10-17
refactor: normal creator

phillebaba:
## Description

This change refactors the package creation.
## Related Issue

Relates to #2969

## Checklist before merging


src/internal/packager2/layout/layout.go
```diff
@@ -47,10 +47,9 @@ func ParseZarfPackage(b []byte) (v1alpha1.ZarfPackage, error) {
 	if err != nil {
 		return v1alpha1.ZarfPackage{}, err
 	}
-	if len(pkg.Build.Migrations) > 0 {
-		for idx, component := range pkg.Components {
-			pkg.Components[idx], _ = deprecated.MigrateComponent(pkg.Build, component)
-		}
+	// TODO (phillebaba): Figure out when migrations should actually be run.
```
phillebaba, line 50: Address this.

src/internal/packager2/layout/create_test.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/zarf-dev/zarf/src/test/testutil"
 )
 
-func TestCreateSkeleton(t *testing.T) {
+func TestCreate(t *testing.T) {
```
AustinAbro321, line 19: This is a copy of the `TestCreateSkeleton` function

src/internal/packager2/layout/sbom_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	ctx := testutil.TestContext(t)
+
+	outputPath := t.TempDir()
+	img := empty.Image
+	b, err := createImageSBOM(ctx, t.TempDir(), outputPath, img, "docker.io/foo/bar:latest")
+	require.NoError(t, err)
+	require.NotEmpty(t, b)
+
+	fileContnent, err := os.ReadFile(filepath.Join(outputPath, "docker.io_foo_bar_latest.json"))
```
AustinAbro321, line 24: nit: `fileContnent` -> `fileContent`

src/cmd/package.go
```diff
@@ -64,24 +63,21 @@ var packageCreateCmd = &cobra.Command{
...
+			MaxPackageSizeMB:   pkgConfig.CreateOpts.MaxPackageSizeMB,
+			SBOMOut:            pkgConfig.CreateOpts.SBOMOutputDir,
+			SkipSBOM:           pkgConfig.CreateOpts.SkipSBOM,
+			Output:             pkgConfig.CreateOpts.Output,
+		}
+		err := packager2.Create(cmd.Context(), pkgConfig.CreateOpts.BaseDir, opt)
 		if err != nil {
 			return err
 		}
-		defer pkgClient.ClearTempPaths()
```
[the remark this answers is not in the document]
phillebaba: We print linting errors in a lot of places, and this creates a lot of noise and complexity. If you look at a lot of other tools they will not lint during build, instead it is a separate command. We have a separate command today which users should use.

src/internal/packager2/layout/create.go
```diff
@@ -131,7 +259,204 @@ func validate(pkg v1alpha1.ZarfPackage, packagePath string, setVariables map[str
...
+				// get the compressedFileName from the source
+				compressedFileName, err := helpers.ExtractBasePathFromURL(file.Source)
+				if err != nil {
+					return fmt.Errorf(lang.ErrFileNameExtract, file.Source, err.Error())
+				}
+				tmpDir, err := utils.MakeTempDir(config.CommonOptions.TempDirectory)
+				if err != nil {
+					return err
+				}
+				defer os.RemoveAll(tmpBuildPath)
```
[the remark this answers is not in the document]
phillebaba: tmpDir is already used in this function.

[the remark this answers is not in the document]
AustinAbro321: Yeah but in this function we `defer os.RemoveAll(tmpBuildPath)` twice, but we never `defer os.RemoveAll(tmpDir)`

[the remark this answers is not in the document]
phillebaba: Now I get what you mean.

AustinAbro321:
Generally looks good. I do request that we keep the `Package saved to "zarf-init-amd64-v0.43.0.tar.zst"` message we have currently. 

I know we're trying to limit output for non errors, but I think the package tar path is worth logging since the filename isn't intuitive. It's a combination of whether it's an init or package config, amd or arm64, and the version. As a user I've found it useful to copy + paste the name from terminal when my next step is to publish or deploy, especially when I have multiple packages in the same directory.

phillebaba:
I have now added back the logging for the path the package is saved to.

src/internal/packager2/actions/actions.go
AustinAbro321, line 1: This is missing the recent logging changes. Aside from re-adding the `basepath` paramater to `run` and `runAction` you should be able to copy paste `src/pkg/packager/actions.go` here without issue.

phillebaba: Are you sure? I just copied it today?

AustinAbro321: you're right, I must've been on an old copy

src/internal/packager2/layout/import.go
```diff
@@ -46,10 +47,12 @@ func resolveImports(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath,
...
 		if component.Import.Path != "" {
 			importPath := filepath.Join(packagePath, component.Import.Path)
-			if _, ok := seenImports[importPath]; ok {
-				return v1alpha1.ZarfPackage{}, fmt.Errorf("package %s imported in cycle by %s", filepath.ToSlash(importPath), filepath.ToSlash(packagePath))
+			importKey := fmt.Sprintf("%s-%s", component.Name, importPath)
+			if _, ok := seenImports[importKey]; ok {
+				return v1alpha1.ZarfPackage{}, fmt.Errorf("package %s imported in cycle by %s in component %s", filepath.ToSlash(importPath), filepath.ToSlash(packagePath), component.Name)
 			}
-			seenImports[importPath] = nil
+			seenImports = maps.Clone(seenImports)
```
AustinAbro321, line 54: Ah so by using the maps.clone it creates a shallow copy so when the next component in the for loop comes by it doesn't have the sibling branches on it. Very neat!

AustinAbro321 (approved):

PR #3120, opened 2024-10-18
feat: introduce logger to ctx, refactor cmd, and migrate logo to Zarf Say

mkcp:
## Description
There's a bunch of changes in this one but they all intersect on the critical path of cmd.Root, so it made sense to batch them together. s/o to @eddiezane for `logger.From()` - I was going a little insane looking at the typecast pattern everywhere, this is much more elegant. I'm not 100% sold on adding the say Command to root in `Execute()`, ideally these would all be declared in the same space as the others, but I wanted to avoid introducing another init() and instead begin us on a better layout for cmd.

Also this adds `--log-format` and `json` is now supported as well as text.

Next steps are diving through commands one at a time and replacing the message logs with the new logger.

## Related Issue
Relates to #2576 

## Checklist before merging


src/config/lang/english.go
```diff
@@ -46,6 +46,7 @@ const (
 		"using a declarative packaging strategy to support DevSecOps in offline and semi-connected environments."
 
 	RootCmdFlagLogLevel              = "Log level when running Zarf. Valid options are: warn, info, debug, trace"
+	RootCmdFlagLogFormat             = "Select a logging format. Defaults to 'text'. Valid options are: 'text', 'json'"
```
phillebaba, line 49: We are working to remove the lang package so we shouldn't add more to it.

src/pkg/logger/logger.go
```diff
@@ -145,6 +154,36 @@ func New(cfg Config) (*slog.Logger, error) {
...
+
+// From takes a context and reads out a "logger" value, optionally taking a key string. If multiple keys are provided,
+// any after the first will be ignored. Note that if From does not find a value, or that value is not a *slog.Logger,
+// it will return return nil.
+//
+// Usage:
+//
+//	l := From(ctx)
+//	l := From(ctx, "logger2")
+func From(ctx context.Context, key ...CtxKey) *slog.Logger {
```
phillebaba, line 168: Most other logger implementations either as a from context or discard, or they return an error when the logger is not found. I think we should return a discard logger if it is not found in the context.

mkcp: I agree abt the discard logger - updated in latest

src/config/lang/english.go
mkcp: My intent was to keep using the existing structure until we moved it all at once, but there's not rly a downside to adding the string directly in root.go. Moved it over it latest.

src/cmd/root.go
```diff
@@ -122,15 +138,94 @@ func init() {
...
+
+	// Security
 	rootCmd.PersistentFlags().BoolVar(&config.CommonOptions.Insecure, "insecure", v.GetBool(common.VInsecure), lang.RootCmdFlagInsecure)
 	rootCmd.PersistentFlags().MarkDeprecated("insecure", "please use --plain-http, --insecure-skip-tls-verify, or --skip-signature-validation instead.")
 	rootCmd.PersistentFlags().BoolVar(&config.CommonOptions.PlainHTTP, "plain-http", v.GetBool(common.VPlainHTTP), lang.RootCmdFlagPlainHTTP)
 	rootCmd.PersistentFlags().BoolVar(&config.CommonOptions.InsecureSkipTLSVerify, "insecure-skip-tls-verify", v.GetBool(common.VInsecureSkipTLSVerify), lang.RootCmdFlagInsecureSkipTLSVerify)
+
+	// HACK(mkcp): This is a workaround for us testing that help output matches to the byte. Undo this and update tests
+	// before release.
+	rootCmd.PersistentFlags().MarkHidden("log-format")
```
schristoff, line 160: Can we keep the error here?

mkcp: discussed out of band

PR #3132, opened 2024-10-23
feat: migrate package create to slog

mkcp:
## Description
This PR:
- Adds contributor QOL improvements:
  - `sloglint`
  -  New formats: `devslog` (via `--log-format="dev"`) and `console-slog` (via `--log-format="console"`)  
- adds WithContext to `packager.Packager` and stores a context on it so we can log in its New() which has some load bearing logging. This is intended as a workaround.
- Makes an empty log format, the default, `--log-format=""`  == "none. So slog must be directly enabled.
- Adds setup functions for both message and logger in `cmd/root` and feature flagging for disabling message and using logger.
  - When a log-format is provided, `setupMessage()` ensures PTerm output is discarded and sets the values for `--no-progress`, and `--confirm`.
- Adds structured logging outputs throughout `zarf package create`
- Marks TODOS on message calls in Create stating they're deprecated

## Related Issue
Relates to #2576

## Checklist before merging

### DIFF
--- a/src/internal/packager/images/pull.go
+++ b/src/internal/packager/images/pull.go
@@ -93,10 +93,8 @@ func Pull(ctx context.Context, cfg PullConfig) (map[transform.Image]v1.Image, er
 	switch c := len(cfg.ImageList); {
 	case c > 15:
 		l.Info("fetching info for images. This step may take a couple of minutes to complete", "count", c, "destination", cfg.DestinationDirectory)
-		break
 	case c > 5:
 		l.Info("fetching info for images. This step may take several seconds to complete", "count", c, "destination", cfg.DestinationDirectory)
-		break
 	default:
 		l.Info("fetching info for images", "count", c, "destination", cfg.DestinationDirectory)
 	}
--- a/src/internal/packager/images/pull.go
+++ b/src/internal/packager/images/pull.go
@@ -17,6 +17,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/avast/retry-go/v4"
 	"github.com/defenseunicorns/pkg/helpers/v2"
@@ -65,6 +66,8 @@ func checkForIndex(refInfo transform.Image, desc *remote.Descriptor) error {
 func Pull(ctx context.Context, cfg PullConfig) (map[transform.Image]v1.Image, error) {
 	l := logger.From(ctx)
 	var longer string
+	pullStart := time.Now()
+
 	imageCount := len(cfg.ImageList)
 	// Give some additional user feedback on larger image sets
 	if imageCount > 15 {
@@ -83,18 +86,19 @@ func Pull(ctx context.Context, cfg PullConfig) (map[transform.Image]v1.Image, er
 	}
 
 	// Give some additional user feedback on larger image sets
+	imageFetchStart := time.Now()
 	// TODO(mkcp): Remove message on logger release
 	spinner := message.NewProgressSpinner("Fetching info for %d images. %s", imageCount, longer)
 	defer spinner.Stop()
 	switch c := len(cfg.ImageList); {
 	case c > 15:
-		l.Info("fetching info for images. This step may take a couple of minutes to complete", "count", c)
+		l.Info("fetching info for images. This step may take a couple of minutes to complete", "count", c, "destination", cfg.DestinationDirectory)
 		break
 	case c > 5:
-		l.Info("fetching info for images. This step may take several seconds to complete", "count", c)
+		l.Info("fetching info for images. This step may take several seconds to complete", "count", c, "destination", cfg.DestinationDirectory)
 		break
 	default:
-		l.Info("fetching info for images", "count", c)
+		l.Info("fetching info for images", "count", c, "destination", cfg.DestinationDirectory)
 	}
 
 	logs.Warn.SetOutput(&message.DebugWriter{})
@@ -240,7 +244,7 @@ func Pull(ctx context.Context, cfg PullConfig) (map[transform.Image]v1.Image, er
 
 	// TODO(mkcp): Remove message on logger release
 	spinner.Successf("Fetched info for %d images", imageCount)
-	l.Debug("done fetching info for images", "count", len(cfg.ImageList))
+	l.Debug("done fetching info for images", "count", len(cfg.ImageList), "duration", time.Since(imageFetchStart))
 
 	doneSaving := make(chan error)
 	updateText := fmt.Sprintf("Pulling %d images", imageCount)
@@ -302,6 +306,8 @@ func Pull(ctx context.Context, cfg PullConfig) (map[transform.Image]v1.Image, er
 	if err != nil {
 		return nil, err
 	}
+
+	l.Debug("done pulling images", "count", len(cfg.ImageList), "duration", time.Since(pullStart))
 
 	return fetched, nil
 }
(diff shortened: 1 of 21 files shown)


src/internal/packager/images/pull.go
```diff
@@ -80,8 +85,21 @@ func Pull(ctx context.Context, cfg PullConfig) (map[transform.Image]v1.Image, er
...
 
+	// Give some additional user feedback on larger image sets
+	imageFetchStart := time.Now()
+	// TODO(mkcp): Remove message on logger release
 	spinner := message.NewProgressSpinner("Fetching info for %d images. %s", imageCount, longer)
 	defer spinner.Stop()
+	switch c := len(cfg.ImageList); {
+	case c > 15:
+		l.Info("fetching info for images. This step may take a couple of minutes to complete", "count", c, "destination", cfg.DestinationDirectory)
+		break
```
AustinAbro321, line 96: `break` isn't necessary for go switch blocks

src/internal/packager/sbom/catalog.go
```diff
@@ -52,51 +54,70 @@ var componentPrefix = "zarf-component-"
...
 	}
 	builder := Builder{
+		// TODO(mkcp): Remove message on logger release
 		spinner:    message.NewProgressSpinner("Creating SBOMs for %d images and %d components with files.", imageCount, componentCount),
 		cachePath:  cachePath,
 		imagesPath: paths.Images.Base,
 		outputDir:  paths.SBOMs.Path,
 	}
 	defer builder.spinner.Stop()
+	l.Info("creating SBOMs for images and components with files", "images", imageCount, "components", componentCount)
```
AustinAbro321, line 72: Thoughts on deleting this message since it may be redundant alongside the other "creating SBOMs for images" and "creating SBOMs for components" messages

src/internal/packager/sbom/catalog.go
```diff
@@ -52,51 +54,70 @@ var componentPrefix = "zarf-component-"
...
+		// TODO(mkcp): Remove message on logger release
 		builder.spinner.Errorf(err, "Unable to generate the SBOM image list")
+		l.Error("unable to generate the SBOM image list", "error", err.Error())
 		return err
 	}
 	builder.jsonList = json
 
 	// Generate SBOM for each image
 	currImage := 1
+	l.Info("creating SBOMs for images", "images", imageCount)
```
AustinAbro321, line 89: Should we change this key to `"count"`

src/internal/packager/sbom/catalog.go
```diff
@@ -106,22 +127,31 @@ func Catalog(ctx context.Context, componentSBOMs map[string]*layout.ComponentSBO
 	currComponent := 1
 
 	// Generate SBOM for each component
+	l.Info("creating SBOMs for components", "components", componentCount)
```
AustinAbro321, line 130: Should we change this key to `"count"`

src/internal/packager/sbom/catalog.go
```diff
@@ -106,22 +127,31 @@ func Catalog(ctx context.Context, componentSBOMs map[string]*layout.ComponentSBO
...
 			message.Debugf("Component %s has invalid SBOM, skipping", component)
+			l.Debug("component has invalid SBOM, skipping", "component", component)
 			continue
 		}
 
 		jsonData, err := builder.createFileSBOM(ctx, *componentSBOMs[component], component)
 		if err != nil {
+			// TODO(mkcp): Remove message on logger release
 			builder.spinner.Errorf(err, "Unable to create SBOM for component %s", component)
+			l.Error("unable to create SBOM for component", "component", component, "error", err.Error())
```
AustinAbro321, line 147: While the current code does print the error on the spinner, IMO it makes more sense to craft and return an error here, otherwise we will end up printing it out twice. I'm guessing it was sent to `spinner.Errorf` so the spinner would have the icon showing it failed.

AustinAbro321: same comment for the other `builder.spinner.Errorf` calls

src/pkg/packager/create.go
```diff
@@ -8,51 +8,84 @@ import (
...
 		return fmt.Errorf("package creation canceled")
 	}
 
-	if err := pc.Assemble(ctx, p.layout, p.cfg.Pkg.Components, p.cfg.Pkg.Metadata.Architecture); err != nil {
+	aStart := time.Now()
+	l.Debug("starting package assembly", "kind", pkg.Kind)
+	if err := pc.Assemble(ctx, p.layout, pkg.Components, pkg.Metadata.Architecture); err != nil {
 		return err
 	}
+	l.Debug("done package assembly", "kind", pkg.Kind, "duration", time.Since(aStart))
```
AustinAbro321, line 78: Maybe "finished package assembly" or "done assembling package"

src/pkg/packager/creator/normal.go
```diff
@@ -490,12 +536,19 @@ func (pc *PackageCreator) addComponent(ctx context.Context, component v1alpha1.Z
 			}
 		}
 		spinner.Success()
+		l.Debug("done building k8s manifests",
```
AustinAbro321, line 539: Thoughts on changing to "done processing k8s manifests", we only build manifests if kustomizations are used.

src/pkg/packager/creator/normal.go
```diff
@@ -443,9 +480,12 @@ func (pc *PackageCreator) addComponent(ctx context.Context, component v1alpha1.Z
 			}
 		}
 		spinner.Success()
+		l.Debug("done loading data injections", "duration", injectStart)
```
AustinAbro321, line 483: missing `time.since`

src/pkg/utils/network.go
```diff
@@ -111,16 +117,18 @@ func httpGetFile(url string, destinationFile *os.File) (err error) {
...
+	// TODO(mkcp): Remove message on logger release
 	title := fmt.Sprintf("Downloading %s", filepath.Base(url))
 	progressBar := message.NewProgressBar(resp.ContentLength, title)
-
-	if _, err = io.Copy(destinationFile, io.TeeReader(resp.Body, progressBar)); err != nil {
+	reader := io.TeeReader(resp.Body, progressBar)
+	// Copy response body to file
+	if _, err = io.Copy(destinationFile, reader); err != nil {
 		progressBar.Failf("Unable to save the file %s: %s", destinationFile.Name(), err.Error())
+		l.Error("unable to save the file %s: %s", destinationFile.Name(), err.Error())
```
AustinAbro321, line 128: I recommend returning this error straight to the user

AustinAbro321:
You will need to edit [cmd.Execute](https://github.com/zarf-dev/zarf/blob/4aeeb7569588fd13cfc19b42d52f31b30b10867c/src/cmd/root.go#L124-L125) to slog an error message

src/pkg/packager/creator/normal.go
mkcp: sure is - fixed in latest

src/internal/packager/sbom/catalog.go
mkcp: Yeah that's fine - it did feel a bit redundant when I was porting over the spinner messages.

src/pkg/packager/create.go
```diff
@@ -8,51 +8,84 @@ import (
...
+	// Load package def
+	pkg, warnings, err := pc.LoadPackageDefinition(p.ctx, lo)
 	if err != nil {
 		return err
 	}
+	//  Store on packager config
 	p.cfg.Pkg = pkg
+	if len(warnings) > 0 {
+		l.Warn("warnings found when loading package definition", "warnings", warnings)
+	}
```
AustinAbro321, line 61:
```suggestion
	for _, warning := range warnings {
		l.Warn(warning)
	}
```
Thoughts on changing it to this? IMO it will read better

mkcp: Yeah definitely better with more lines

mkcp:
Revisions applied - should be good to go. Just waiting on ci checks.

src/internal/packager/sbom/catalog.go
mkcp: fixed in latest

AustinAbro321 (approved):

PR #3149, opened 2024-10-25
chore: improve the Info and Debug logging for image pulls

mkcp:
## Description
@AustinAbro321 suggested this improvement to the image pull logging while we reviewed #3132. This gives users better feedback into the parts of pull that we expect will take a while. Also added durations to the debug "done" logs so we get a bit more perf data.

## Related Issue
Relates to #3132 

## Checklist before merging

### DIFF
--- a/src/internal/packager/images/pull.go
+++ b/src/internal/packager/images/pull.go
@@ -90,14 +90,7 @@ func Pull(ctx context.Context, cfg PullConfig) (map[transform.Image]v1.Image, er
 	// TODO(mkcp): Remove message on logger release
 	spinner := message.NewProgressSpinner("Fetching info for %d images. %s", imageCount, longer)
 	defer spinner.Stop()
-	switch c := len(cfg.ImageList); {
-	case c > 15:
-		l.Info("fetching info for images. This step may take a couple of minutes to complete", "count", c, "destination", cfg.DestinationDirectory)
-	case c > 5:
-		l.Info("fetching info for images. This step may take several seconds to complete", "count", c, "destination", cfg.DestinationDirectory)
-	default:
-		l.Info("fetching info for images", "count", c, "destination", cfg.DestinationDirectory)
-	}
+	l.Info("fetching info for images", "count", imageCount, "destination", cfg.DestinationDirectory)
 
 	logs.Warn.SetOutput(&message.DebugWriter{})
 	logs.Progress.SetOutput(&message.DebugWriter{})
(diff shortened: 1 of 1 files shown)


src/internal/packager/images/pull.go
```diff
@@ -354,24 +356,37 @@ func CleanupInProgressLayers(ctx context.Context, img v1.Image) error {
...
 func SaveSequential(ctx context.Context, cl clayout.Path, m map[transform.Image]v1.Image) (map[transform.Image]v1.Image, error) {
+	l := logger.From(ctx)
 	saved := map[transform.Image]v1.Image{}
 	for info, img := range m {
 		annotations := map[string]string{
 			ocispec.AnnotationBaseImageName: info.Reference,
 		}
+		// NOTE(mkcp): It's ok if the metadata is empty / err, they're just for debug logs.
+		wStart := time.Now()
+		size, _ := img.Size() //nolint:errcheck
```
AustinAbro321, line 367: IMO we shouldn't skip these errors. Zarf has been checking `img.Size()` on push for the progressbar until recently when we removed it, and I've never seen an error from it. Likely if this breaks, there is something wrong with the image / with the pull

src/internal/packager/images/pull.go
```diff
@@ -354,24 +356,37 @@ func CleanupInProgressLayers(ctx context.Context, img v1.Image) error {
...
+		l.Info("saving image", "ref", info.Reference, "size", size, "method", "sequential")
 		if err := cl.AppendImage(img, clayout.WithAnnotations(annotations)); err != nil {
 			if err := CleanupInProgressLayers(ctx, img); err != nil {
 				message.WarnErr(err, "failed to clean up in-progress layers, please run `zarf tools clear-cache`")
+				l.Error("failed to clean up in-progress layers. please run `zarf tools clear-cache`")
```
AustinAbro321, line 372: In Zarf now, `message.WarnErr` prints in the yellow warn format. I think it makes sense to keep in that format and just use `l.Error` for returned errors that stop the CLI entirely.

mkcp: Typically warns are used in systems to inform users but it's been handled automatically, or it don't require immediate action, e.g. we had to fallback to a slower method or a feature is deprecated. Failing to clean up here notes an actual Error and we should log it as that to users imo.

src/internal/packager/images/pull.go
mkcp: My intent was avoiding modifying any of the retry and cleanup behavior but i'm fine returning an error here. Added in next commit

src/internal/packager/images/pull.go
AustinAbro321: that makes a lot of sense, thanks for the explanation!

AustinAbro321 (approved):

PR #3159, opened 2024-10-29
chore: add logger package deploy

AustinAbro321:
## Description

add slogger for package deploy. Example below. Note that we will not print connect strings or the credentials table after deploy for `--log-format="console"` after this is merged. Once we've refactored to packager2 it becomes easier to do that in the future. For now, since we have the commands `zarf connect list` and `zarf tools get-creds` which achieve the same affect, I feel comfortable with this change.   
image

## Related Issue

Relates to #2576 

## Checklist before merging


src/pkg/packager/deploy.go
```diff
@@ -89,12 +92,18 @@ func (p *Packager) Deploy(ctx context.Context) error {
 		return err
 	}
 	warnings = append(warnings, validateWarnings...)
+	for _, warning := range validateWarnings {
```
mkcp, line 95: note: any time I see a simple "take collection, apply function to each" I make this face 😃

src/pkg/packager/deploy.go
```diff
@@ -124,10 +133,12 @@ func (p *Packager) Deploy(ctx context.Context) error {
 	}
 	if len(deployedComponents) == 0 {
 		message.Warn("No components were selected for deployment.  Inspect the package to view the available components and select components interactively or by name with \"--components\"")
+		l.Warn("No components were selected for deployment.  Inspect the package to view the available components and select components interactively or by name with \"--components\"")
```
mkcp, line 136: nit: capitalization. admittedly this one is weird cause of the second sentence. I was doing lower first sentence, upper second.

src/pkg/packager/deploy.go
```diff
@@ -124,10 +133,12 @@ func (p *Packager) Deploy(ctx context.Context) error {
...
 	}
 
 	// Notify all the things about the successful deployment
 	message.Successf("Zarf deployment complete")
+	l.Debug("Zarf deployment complete", "duration", time.Since(start))
```
mkcp, line 141: Nit: capital. 👍 on the duration metric

src/pkg/packager/deploy.go
```diff
@@ -357,6 +378,7 @@ func (p *Packager) deployComponent(ctx context.Context, component v1alpha1.ZarfC
 		healthCheckContext, cancel := context.WithTimeout(ctx, p.cfg.DeployOpts.Timeout)
 		defer cancel()
 		spinner := message.NewProgressSpinner("Running health checks")
+		l.Info("running health checks")
```
mkcp, line 381: Non-blocking: Is there a piece of data we can use to link this to the component e.g. component name? If we grow this part of the system to be concurrent and healthchecks are interleaved together, we want the ordering to not matter for the logs to make sense.

src/pkg/packager/deploy.go
```diff
@@ -368,16 +390,22 @@ func (p *Packager) deployComponent(ctx context.Context, component v1alpha1.ZarfC
...
 	}
+	l.Debug("done deploying component", "name", component.Name, "duration", time.Since(start))
 	return charts, nil
 }
 
 // Move files onto the host of the machine performing the deployment.
-func (p *Packager) processComponentFiles(component v1alpha1.ZarfComponent, pkgLocation string) error {
+func (p *Packager) processComponentFiles(ctx context.Context, component v1alpha1.ZarfComponent, pkgLocation string) error {
+	l := logger.From(ctx)
+	l.Handler()
```
mkcp, line 400: Bug: Is this intentional? It doesn't look like this value is used.

src/pkg/packager/deploy.go
```diff
@@ -475,6 +509,7 @@ func (p *Packager) setupState(ctx context.Context) error {
 
 		// Try to create the zarf namespace
 		spinner.Updatef("Creating the Zarf namespace")
+		l.Info("creating the Zarf namespace")
```
mkcp, line 512: Non-blocking: Can we easily add the namespace name here?

src/pkg/packager/sources/tarball.go
```diff
@@ -100,12 +105,14 @@ func (s *TarballSource) LoadPackage(ctx context.Context, dst *layout.PackagePath
 	if !dst.IsLegacyLayout() {
 		spinner := message.NewProgressSpinner("Validating full package checksums")
 		defer spinner.Stop()
+		l.Info("validating package checksums")
```
mkcp, line 108: Non-blocking: Is there an easy piece of metadata we can attach this to? Package name or source maybe?

src/pkg/zoci/common.go
```diff
@@ -29,12 +31,15 @@ type Remote struct {
 
 // NewRemote returns an oras remote repository client and context for the given url
 // with zarf opination embedded
-func NewRemote(url string, platform ocispec.Platform, mods ...oci.Modifier) (*Remote, error) {
-	logger := slog.New(message.ZarfHandler{})
+func NewRemote(ctx context.Context, url string, platform ocispec.Platform, mods ...oci.Modifier) (*Remote, error) {
+	l := slog.New(message.ZarfHandler{})
+	if logger.Enabled(ctx) {
```
mkcp, line 36: After seeing this fn used a few times, this looks like an effective way to propagate feature flags through the system. Imo we should look to use this pattern in the future.

mkcp (changes requested):
A few nits around log captialization, some non-blocking comments about adding metadata to some log lines, and a potential bug/dead code about a `Handler()` value that doesn't seem to be used. Overall, good work!

src/pkg/packager/deploy.go
AustinAbro321: We're creating thee Zarf namespace here - the namespace called Zarf

src/pkg/packager/deploy.go
AustinAbro321: This one was on purpose. I've been capitalizing proper nouns like Zarf and Helm. I'm fine with either choice, lmk what you'd like to do going forward, as I currently have the word Helm capitalized a few times in this PR.

src/pkg/packager/deploy.go
AustinAbro321: Totally missed that good catch

src/pkg/packager/deploy.go
AustinAbro321: Yup added the chart name

mkcp (approved):

PR #3161, opened 2024-10-29
chore: feature flag to enable switching between logger for libaries

AustinAbro321:
## Description

When a library takes in a logger we currently have a way for it to respect the logging feature flag as we move to slog. This PR adds functionality to enable that

## Related Issue

Relates to #2576 

## Checklist before merging

### DIFF
--- a/src/internal/packager/images/pull.go
+++ b/src/internal/packager/images/pull.go
@@ -144,6 +144,7 @@ func Pull(ctx context.Context, cfg PullConfig) (map[transform.Image]v1.Image, er
 
 					// TODO(mkcp): Remove message on logger release
 					message.Warnf("Falling back to local 'docker', failed to find the manifest on a remote: %s", err.Error())
+					l.Warn("Falling back to local 'docker', failed to find the manifest on a remote", "error", err.Error())
 
 					// Attempt to connect to the local docker daemon.
 					cli, err := client.NewClientWithOpts(client.FromEnv)
--- a/src/internal/packager/template/template.go
+++ b/src/internal/packager/template/template.go
@@ -36,13 +36,10 @@ func GetZarfVariableConfig(ctx context.Context) *variables.VariableConfig {
 		return interactive.PromptVariable(variable)
 	}
 
-	var l *slog.Logger
 	if logger.Enabled(ctx) {
-		l = logger.From(ctx)
-	} else {
-		l = slog.New(message.ZarfHandler{})
+		return variables.New("zarf", prompt, logger.From(ctx))
 	}
-	return variables.New("zarf", prompt, l)
+	return variables.New("zarf", prompt, slog.New(message.ZarfHandler{}))
 }
 
 // GetZarfTemplates returns the template keys and values to be used for templating.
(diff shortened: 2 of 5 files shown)


src/internal/packager/images/pull.go
```diff
@@ -144,6 +144,7 @@ func Pull(ctx context.Context, cfg PullConfig) (map[transform.Image]v1.Image, er
 
 					// TODO(mkcp): Remove message on logger release
 					message.Warnf("Falling back to local 'docker', failed to find the manifest on a remote: %s", err.Error())
+					l.Warn("Falling back to local 'docker', failed to find the manifest on a remote", "error", err.Error())
```
AustinAbro321, line 147: Unrelated, but noticed this the other day. IMO needed otherwise it's hard to tell why you're pull failed (this happened to me the other day when my PAT expired)

src/internal/packager/template/template.go
```diff
@@ -35,7 +36,13 @@ func GetZarfVariableConfig(ctx context.Context) *variables.VariableConfig {
...
 	}
 
-	return variables.New("zarf", prompt, logger.From(ctx))
+	var l *slog.Logger
+	if logger.Enabled(ctx) {
+		l = logger.From(ctx)
+	} else {
+		l = slog.New(message.ZarfHandler{})
+	}
+	return variables.New("zarf", prompt, l)
```
mkcp, line 45: Nit: Usually elses are avoided in Go, we can use returns to express this section as: 

```suggestion
    // Use logger if enabled
	if logger.Enabled(ctx) {
		return variables.New("zarf", prompt, logger.From(ctx))
	}
	// Use message by default
	return variables.New("zarf", prompt, slog.New(message.ZarfHandler{}))
```

src/pkg/logger/logger.go
```diff
@@ -170,6 +170,30 @@ func WithContext(ctx context.Context, logger *slog.Logger) context.Context {
 	return context.WithValue(ctx, defaultCtxKey, logger)
 }
 
+// TODO (@austinabro321) once we switch over to the new logger completely the enabled key & logic should be deleted
```
mkcp, line 173: 👍

src/pkg/logger/logger.go
```diff
@@ -170,6 +170,30 @@ func WithContext(ctx context.Context, logger *slog.Logger) context.Context {
...
+
+var defaultCtxKeyEnabled = ctxKeyEnabled{}
+
+// WithLoggingEnabled allows stores a value to determine whether or not slog logging is enabled
+func WithLoggingEnabled(ctx context.Context, enabled bool) context.Context {
+	return context.WithValue(ctx, defaultCtxKeyEnabled, enabled)
+}
+
+// Enabled returns true if slog logging is enabled
+func Enabled(ctx context.Context) bool {
```
mkcp, line 184: There's nothing wrong with the other route of passing the flag in as config struct values, but this is consistent with the `From(ctx)` route we took with `*slog.Logger`. Better to stick to the one way to do this imo.

mkcp (dismissed):
Good cleanup. `Enabled()` is simple for callers so imo that turned out well. One non-blocking nit about elses - lmk if you revise and I'll re-approve asap. No need to use the suggestion commit either, it always gets weird around DCO.

PR #3163, opened 2024-10-29
chore: begin adding logger to `zarf dev`

mkcp:
## Description
This PR adds logger to `sha256sum`, `patch-git`, `lint` and `find-images`. There's a few more dev commands to update, which i'll grab in another PR.

## Related Issue
Relates to #2576 

## Checklist before merging

### DIFF
--- a/src/cmd/dev.go
+++ b/src/cmd/dev.go
@@ -41,7 +41,6 @@ var devCmd = &cobra.Command{
 	Short:   lang.CmdDevShort,
 }
 
-// FIXME(mkcp): Lots of message
 var devDeployCmd = &cobra.Command{
 	Use:   "deploy",
 	Args:  cobra.MaximumNArgs(1),
@@ -75,7 +74,6 @@ var devDeployCmd = &cobra.Command{
 	},
 }
 
-// FIXME(mkcp): Lots of message
 var devGenerateCmd = &cobra.Command{
 	Use:     "generate NAME",
 	Aliases: []string{"g"},
@@ -173,6 +171,7 @@ var devSha256SumCmd = &cobra.Command{
 
 		if helpers.IsURL(fileName) {
 			message.Warn(lang.CmdDevSha256sumRemoteWarning)
+			logger.From(cmd.Context()).Warn("this is a remote source. If a published checksum is available you should use that rather than calculating it directly from the remote link")
 
 			fileBase, err := helpers.ExtractBasePathFromURL(fileName)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


src/pkg/packager/prepare.go
```diff
@@ -260,6 +269,8 @@ func (p *Packager) findImages(ctx context.Context) (map[string][]string, error)
 			}
 		}
 
+		imgCompStart := time.Now()
+		l.Info("looking for image in component", "name", component.Name, "resourcesCount", len(resources))
```
AustinAbro321, line 273: image->images

src/pkg/packager/prepare.go
```diff
@@ -305,16 +318,23 @@ func (p *Packager) findImages(ctx context.Context) (map[string][]string, error)
 		}
 
 		spinner.Success()
+		l.Debug("done looking for image in component",
```
AustinAbro321, line 321: image->images

src/cmd/dev.go
```diff
@@ -104,20 +105,30 @@ var devTransformGitLinksCmd = &cobra.Command{
...
-		processedText := transform.MutateGitURLsInText(message.Warnf, pkgConfig.InitOpts.GitServer.Address, text, pkgConfig.InitOpts.GitServer.PushUsername)
+
+		// Set log func for transform
+		var logFn func(string, ...any)
+		logFn = message.Warnf
+		// Use logger if we can
+		if logger.Enabled(ctx) {
+			logFn = logger.From(ctx).Warn
+		}
+		processedText := transform.MutateGitURLsInText(logFn, gitServer.Address, text, gitServer.PushUsername)
```
AustinAbro321, line 131: This is actually expecting printf style logging so the message ends up looking like this. 
```bash
 Unable to transform the git url, using the original url we have: %s !BADKEY=https://*/*.git
```
You could instead use this to keep with the printf format
```go
		l := logger.From(ctx)
		logFn = slog.NewLogLogger(l.Handler(), slog.LevelWarn).Printf
```
Might be good to add a todo to this function to move to just use the slogger from context after logging is released also. Not sure why this was given it's own logging function. Doesn't looked to be called from anywhere outside of Zarf

src/pkg/packager/prepare.go
mkcp: Good catch, ty

src/cmd/dev.go
mkcp: Yeah that's bugged for sure -- I'm going to tear it out and make a todo for later. I can see the effort to make a logger you can pass in, but it just creates weird compatibility issues. the Printf on log.Logger is neat though, hadn't seen that before.

AustinAbro321 (approved):

PR #3166, opened 2024-10-31
fix: maintain agent mutate even when already mutated

### LINKED ISSUE #3146: Zarf-agent does not maintain mutate after it already has `zarf-agent: patched` label applied
### Environment
Device and OS: RHEL 9.4
App version: `v0.42.0
Kubernetes distro being used: RKE2, 1.31.1+rke2r2

### Steps to reproduce
1. Install fluxcd
1. Create gitrepo with the blow manifests
1. Wait for flux to re-apply the kustomization

### Expected result

The  `.spec.url` to point to the zarf docker registry

### Actual Result

The `.spec.url` reverts back to the url to `oci://registry1.dso.mil/bigbang`

### Comments

> We do not have a builtin template for the registry IP, but I'd like to solve this without one.
> Looks like the issue is that the Zarf agent checks for the `zarf-agent: patched` which is added during the initial mutation, and exits if it already exists. see [this block](https://github.com/zarf-dev/zarf/blob/df9c2120fbcd2d3c148c9bd095ae1faf55196501/src/internal/agent/hooks/flux-helmrepo.go#L50).
> I believe the mutation of every resource in the mutating webhook is idempotent, so we can just delete that check.

- Austin Abro, on Slack

### Manifests
`kustomization.yml`
```yaml
---
apiVersion: kustomize.toolkit.fluxcd.io/v1
kind: Kustomization
metadata:
  name: bigbang
  namespace: flux-system
spec:
  interval: 2m0s
  path: ./.
  prune: true
  sourceRef:
    kind: GitRepository
    name: flux-system
```

`bigbang.yml`
```yaml
---
apiVersion: source.toolkit.fluxcd.io/v1
kind: HelmRepository
metadata:
  name: registry1
  namespace: flux-system
spec:
  interval: 2m
  type: oci
  url: oci://registry1.dso.mil/bigbang
```

a1994sc:
## Description

Removes the checks on the agent webhook for effecting already patched gitops related resources:

- `fluxcd-oci-repo`
- `fluxcd-helm-repo`
- `fluxcd-git-repo`
- `argocd-repository`
- `argocd-application`

The reasoning for removing the this check is that during the regular lifecycle of gitops tools is that they will re-apply the manifests to make sure the cluster reflects the correct state; this means that fluxcd will override the mutated repos to point to something outside of `zarf`'s control, e.i. clear web endpoints that might not be accessible.

## Note

So I did not add tests for removing the check on pods... My logic is that pods are static through their lifecycle once they start, so `zarf` would only need to mutate them once.

## Related Issue

Fixes #3146

## Checklist before merging

### DIFF
--- a/src/internal/agent/hooks/flux-helmrepo.go
+++ b/src/internal/agent/hooks/flux-helmrepo.go
@@ -47,13 +47,6 @@ func mutateHelmRepo(ctx context.Context, r *v1.AdmissionRequest, cluster *cluste
 		return &operations.Result{Allowed: true}, nil
 	}
 
-	if src.Labels != nil && src.Labels["zarf-agent"] == "patched" {
-		return &operations.Result{
-			Allowed:  true,
-			PatchOps: nil,
-		}, nil
-	}
-
 	zarfState, err := cluster.LoadZarfState(ctx)
 	if err != nil {
 		return nil, err
--- a/src/internal/agent/hooks/argocd-application_test.go
+++ b/src/internal/agent/hooks/argocd-application_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/zarf-dev/zarf/src/internal/agent/operations"
 	"github.com/zarf-dev/zarf/src/types"
 	v1 "k8s.io/api/admission/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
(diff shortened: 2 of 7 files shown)


AustinAbro321 (changes requested):
Nice work so far! Since this PR is strictly removing code rather than adding code can you delete the test cases involving an already patched Zarf agent, both existing and new

a1994sc:
> Nice work so far! Since this PR is strictly removing code rather than adding code can you delete the test cases involving an already patched Zarf agent, both existing and new

So the thought process in why I added the tests for patching even if they already have the labels is so that it leaves some tests for future developers to not remove that functionality... I can remove the tests if you would still like

a1994sc:
Thank you Austin, I will update the PR to apply the changes!

a1994sc:
@AustinAbro321 Hey, updated the PR to remove the extra tests and added the logic for checking the url against the zarf state registry url. Let me know if that is good

AustinAbro321 (dismissed):
LGTM, but I'd like to get a review from another one of the @zarf-dev/maintainers before merge

a1994sc:
> LGTM, but I'd like to get a review from another one of the @zarf-dev/maintainers before merge

No problem! I figured it was a medium sized change so it is all good to get more eyes on things

src/internal/agent/hooks/flux-helmrepo.go
```diff
@@ -37,6 +37,14 @@ func NewHelmRepositoryMutationHook(ctx context.Context, cluster *cluster.Cluster
 // mutateHelmRepo mutates the repository url to point to the repository URL defined in the ZarfState.
 func mutateHelmRepo(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster.Cluster) (*operations.Result, error) {
 	l := logger.From(ctx)
+	var (
+		patches   []operations.PatchOperation
+		isPatched bool
+
+		isCreate = r.Operation == v1.Create
+		isUpdate = r.Operation == v1.Update
```
phillebaba, line 45: Define these variables where they are first used rather than at the top of the function.

phillebaba:
Small change otherwise things look good.

src/internal/agent/hooks/flux-helmrepo.go
a1994sc: I followed the pattern found in `src/internal/agent/hooks/flux-gitrepo.go`

```
	var (
		patches   []operations.PatchOperation
		isPatched bool

		isCreate = r.Operation == v1.Create
		isUpdate = r.Operation == v1.Update
	)
```
[`ref`](https://github.com/zarf-dev/zarf/blob/a21ed416fc2c2c1c80df97184a49433ee779b933/src/internal/agent/hooks/flux-gitrepo.go#L42-L48) but I can update and get that up tonight

a1994sc:
@AustinAbro321 and @phillebaba 

I updated the PR per @phillebaba comments

phillebaba (approved):

AustinAbro321 (approved):

PR #3168, opened 2024-10-31
chore: slogger for `zarf init`

AustinAbro321:
## Description

Refactors logging for Zarf init

## Related Issue

Relates to #2576 

## Checklist before merging


src/pkg/logger/logger.go
```diff
@@ -107,6 +135,15 @@ type Config struct {
 	Destination
 }
 
+// LogValue of config
+func (c Config) LogValue() slog.Value {
```
AustinAbro321, line 139: When we print cfg current it looks like this
```bash
logger successfully initialized cfg={-4 console 0xc0000de068}
```
now looks like
```bash
logger successfully initialized cfg.level=debug cfg.format=console cfg.Destination=os.Stderr
```

src/pkg/zoci/pull.go
```diff
@@ -52,11 +51,14 @@ func (r *Remote) PullPackage(ctx context.Context, destinationDir string, concurr
 	}
 	layersToPull = append(layersToPull, manifest.Config)
 
+	layerSize := oci.SumDescsSize(layersToPull)
+	// TODO (@austinabro321) change this and other r.Log() calls to the proper slog format
+	r.Log().Info(fmt.Sprintf("Pulling %s, size: %s", r.Repo().Reference, utils.ByteFormat(float64(layerSize), 2)))
```
AustinAbro321, line 56: Currently, the progress bar provides the user with the name and size of the package they're downloading. IMO since we're eliminating the progress bar, we should upgrade the message with the package name from debug->info

src/cmd/initialize.go
```diff
@@ -70,6 +71,9 @@ var initCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
+		// Since the new logger ignores pterm output the credential table is no longer printed on init.
+		// This note is the intended replacement, rather than printing creds by default.
```
mkcp, line 75: 👍

src/cmd/root.go
```diff
@@ -222,8 +222,6 @@ func setupMessage(cfg messageCfg) error {
 		message.InitializePTerm(io.Discard)
 		// Disable all progress bars and spinners
 		message.NoProgress = true
-		// Ensures no user input is needed while we maintain backwards compatibility with message
-		config.CommonOptions.Confirm = true
```
mkcp, line 226: 👍

src/pkg/cluster/injector.go
```diff
@@ -123,11 +127,15 @@ func (c *Cluster) StartInjection(ctx context.Context, tmpDir, imagesDir string,
 	}
 
 	spinner.Success()
+	l.Debug("done with injection", "duration", time.Since(start))
```
mkcp, line 130: 🤔 Not in scope for this PR even remotely, but I wonder how difficult it would be to start aggregating duration metrics from CI logs. We don't get a ton of runtime feedback cause _airgap_ so feel like we need to get the signal that we can.

src/pkg/cluster/state.go
```diff
@@ -154,17 +159,21 @@ func (c *Cluster) InitZarfState(ctx context.Context, initOptions types.ZarfInitO
 		initOptions.ArtifactServer.FillInEmptyValues()
 		state.ArtifactServer = initOptions.ArtifactServer
 	} else {
+		// TODO (@austinabro321) validate immediately in `zarf init` if these are set and not equal and error out if so
```
mkcp, line 162: 👍 Worth a follow up issue imo. I don't think this would fall under the packager changes so maybe a refactor would be in scope for that issue. I see a lot of deeply nested code which makes it difficult to reason about.

src/pkg/logger/logger.go
mkcp: Good change ✨

src/pkg/zoci/pull.go
mkcp: Def - thanks for digging into this. Thoughts on a follow up issue?

src/pkg/logger/logger.go
```diff
@@ -37,6 +37,22 @@ var (
 	Error = Level(slog.LevelError) // 8
 )
 
+// String returns the string representation of the Level.
+func (l Level) String() string {
```
mkcp, line 41: This is a fine step for now but if we end up needing anything more complex in the future, we should consider making Level a complex type. e.g.

```
type Level struct {
    Value int
    s string
} 

func (l Level) String() string {
    return s
}

// todo: Equal(level Level) bool {...} ?

var (
    Debug = Level{slog.LevelDebug, "debug"}
    ...
)
```

There's other valid ways to approach this too, just food for thought!

src/pkg/logger/logger.go
```diff
@@ -99,6 +115,18 @@ var (
 	DestinationNone Destination = io.Discard
 )
 
+// can't define method on Destination type
+func destinationString(d Destination) string {
```
mkcp, line 119: Similar problem to `Level` above, there's nothing wrong with this approach but it may run into issues if we need to grow it. We have some application logic not strictly covered by `io.Writer` so we may have to add to our `Desination` type at some point. Nonblocking, just something to consider.

mkcp (approved):
LGTM - added some feedback on growing the system over time and potential follow up issues.

src/pkg/zoci/pull.go
AustinAbro321: Should be already solved with this. Took the debug message and moved it to info. Diff isn't super clear since I moved the message so it would have size info.

src/pkg/cluster/state.go
AustinAbro321: Definitely - https://github.com/zarf-dev/zarf/issues/3188

src/pkg/cluster/injector.go
AustinAbro321: Yeah that'd be interesting, I believe we save log files somewhere or if not it's not hard to do with actions

src/pkg/logger/logger.go
AustinAbro321: Yeah fair

PR #3169, opened 2024-10-31
fear: allow versioning of init packages outside of binary version

### LINKED ISSUE #3148: Allow `ZarfInitConfig` to be versioned outside of being tied to binary version

> So we are running RKE2 in an air-gapped environment, so I'm having to write a custom `zarf-init` that is responsible for installing `rook-ceph` into the cluster... so because the init package version is tied to the binary used by the pipeline, when we upgrade ceph from `1.15.3` -> `1.15.4` we would need to override the version we already have published...

- **Given** when `.metadata.version` is defined in a `ZarfInitConfig` Package
- **Then** use that as the package version for the init package
- **Else** default to using the binary version if not present

### Additional context
`zarf init` would also need to accept an option path argument to allow passing in a different `zarf-init` package, but still have the default behavior to search for `zarf-init-<arch>-<version>.tar.zst`

a1994sc:
## Description

Allows for the versioning of `ZarfInitConfig` by setting `.metadata.version`; it defaults to using the existing logic for setting the version to whatever the binary version is.

## Related Issue

Fixes #3148

## Checklist before merging


src/config/lang/english.go
```diff
@@ -121,6 +121,9 @@ const (
 # Initializing without any optional components:
 $ zarf init
 
+# Initializing w/ a package of a non-standard version:
```
a1994sc, line 124: Not really sure how best to word this... Any input would be very grateful

src/cmd/initialize.go
```diff
@@ -227,3 +236,42 @@ func init() {
...
+			files = append(files, splitFiles...)
+			return files
+		},
+	}
+
+	if err := survey.AskOne(prompt, &path, survey.WithValidator(survey.Required)); err != nil {
+		return "", fmt.Errorf("package path selection canceled: %w", err)
+	}
+
+	return path, nil
```
a1994sc, line 276: Added to follow the pattern laid out by `zarf package deploy`; however seeing I want to leave the room for backward-compatibility I'm not sure if there is an `optional` setting that we change it too?

AustinAbro321 (approved):

PR #3171, opened 2024-11-01
refactor: debug template print

AustinAbro321:
## Description

We were manually putting together a json, refactored this to edit the object then marshall/log the json direct

## Related Issue

Relates to #2576 

## Checklist before merging

### DIFF
--- a/src/internal/packager/template/template.go
+++ b/src/internal/packager/template/template.go
@@ -147,7 +147,9 @@ func debugPrintTemplateMap(ctx context.Context, templateMap map[string]*variable
 func getSanitizedTemplateMap(templateMap map[string]*variables.TextTemplate) map[string]string {
 	sanitizedMap := make(map[string]string, len(templateMap))
 	for key, template := range templateMap {
-		if template.Sensitive {
+		if template == nil {
+			sanitizedMap[key] = ""
+		} else if template.Sensitive {
 			sanitizedMap[key] = "**sanitized**"
 		} else {
 			sanitizedMap[key] = template.Value
--- a/src/internal/packager/template/template.go
+++ b/src/internal/packager/template/template.go
@@ -7,6 +7,7 @@ package template
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"log/slog"
 	"strings"
@@ -102,7 +103,10 @@ func GetZarfTemplates(ctx context.Context, componentName string, state *types.Za
 		}
 	}
 
-	debugPrintTemplateMap(ctx, templateMap)
+	err = debugPrintTemplateMap(ctx, templateMap)
+	if err != nil {
+		return nil, err
+	}
 
 	return templateMap, nil
 }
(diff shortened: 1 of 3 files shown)


src/internal/packager/template/template.go
```diff
@@ -127,21 +131,27 @@ func generateHtpasswd(regInfo *types.RegistryInfo) (string, error) {
...
 }
 
-func debugPrintTemplateMap(ctx context.Context, templateMap map[string]*variables.TextTemplate) {
-	// TODO (@austinabro321) sanitize the template by making a copy and changing the actual keys
-	// then use json.MarshalIndent to create the json
-	debugText := "templateMap = { "
+func debugPrintTemplateMap(ctx context.Context, templateMap map[string]*variables.TextTemplate) error {
+	sanitizedMap := getSanitizedTemplateMap(templateMap)
+
+	b, err := json.MarshalIndent(sanitizedMap, "", "  ")
```
mkcp, line 137: ++ using stdlib when available

src/internal/packager/template/template_test.go
```diff
@@ -0,0 +1,34 @@
...
+	expected := map[string]string{
+		"###ZARF_GIT_AUTH_PULL###": "**sanitized**",
+		"###ZARF_GIT_AUTH_PUSH###": "**sanitized**",
+		"###ZARF_GIT_PULL###": "zarf-git-read-user",
+		"###ZARF_GIT_PUSH###": "zarf-git-user",
+		"###ZARF_REGISTRY###": "127.0.0.1:31999",
+	}
+
+	output := getSanitizedTemplateMap(input)
+	require.Equal(t, expected, output)
```
mkcp, line 33: There's clean data here but I'd recommend using a test table for this approach and running checks for the simplest test cases as well as the complex one. Empty values are great to test for as well. What if we pass in a nil map, empty map, or a map with a real key nil entry `variables.TextTemplate`?

```suggestion
tt := struct{
	name string
	in map[string]*variables.TextTemplate
	expect map[string]string
}{
	{
		// TODO cases: can nils input, empty in
	},
	{
		// TODO case: can handle map with non-sensitive value
	},
	{
		name: "can redact a sensitive value"
		in: map[string]*variables.TextTemplate{
			"###ZARF_GIT_AUTH_PULL###": {Sensitive: true, Value: "secret1"},
		},
		expect: map[string]string{
			"###ZARF_GIT_AUTH_PULL###": "**sanitized**",
		},
	},
	{
		name: "can handle both sensitive and non-sensitive values at once"
		in: map[string]*variables.TextTemplate{
			"###ZARF_GIT_AUTH_PULL###": {Sensitive: true, Value: "secret1"},
			"###ZARF_GIT_AUTH_PUSH###": {Sensitive: true, Value: "secret2"},
			"###ZARF_REGISTRY###":      {Sensitive: false, Value: "127.0.0.1:31999"},
			"###ZARF_GIT_PUSH###":      {Sensitive: false, Value: "zarf-git-user"},
			"###ZARF_GIT_PULL###":      {Sensitive: false, Value: "zarf-git-read-user"},
		},
		expect: map[string]string{
			"###ZARF_GIT_AUTH_PULL###": "**sanitized**",
			"###ZARF_GIT_AUTH_PUSH###": "**sanitized**",
			"###ZARF_GIT_PULL###": "zarf-git-read-user",
			"###ZARF_GIT_PUSH###": "zarf-git-user",
			"###ZARF_REGISTRY###": "127.0.0.1:31999",
		},
	}, 
}

for _, tc := range tt {
    t.Run(tc.name, func(t *testing.T) {
        // TODO Requires here 
    })
}
```
(psuedocode)

mkcp (changes requested):
Good change overall and glad you took on adding tests. I had some feedback abt. the approach in testing-with-data approach. There's some "unhappy path" cases to cover with nils and empties that we can cover easier with the test table pattern.

src/internal/packager/template/template_test.go
AustinAbro321: Good call! It would panic before if variables.TextTemplate was nil, fixed that and added test cases.

mkcp (approved):
LGTM

PR #3193, opened 2024-11-06
feat: exit on config file error

### LINKED ISSUE #3192: Errors in `zarf-config` file should not be ignored
Currently [cmd/common/viper.go](https://github.com/zarf-dev/zarf/blob/main/src/cmd/common/viper.go#L151) completely ignores errors in the zarf config file during `zarf package create`. It does this because one possible error is that no config file was found and the config file is optional. However this hides all kinds of errors that could be cause by an invalid config file and hiding this leads to difficult to debug unexpected behavior.

- **Given** an invalid `zarf-confif.yaml` (or toml or json)
- **When** `zarf package create` is run (or `zarf package deploy` or anything that uses configuration)
- **Then** zarf should report the error and exit with a non-zero error code

* If not an error & exit then at least a warning would be better.
* Technically a separate `zarf dev validate-config` would also be a viable solution

AustinAbro321:
## Description

When we have a config file error that isn't not found we shouldn't ignore it

## Related Issue

Fixes #3192 

## Checklist before merging

### DIFF
--- a/src/cmd/common/viper.go
+++ b/src/cmd/common/viper.go
@@ -7,6 +7,7 @@ package common
 import (
 	"context"
 	"errors"
+	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


src/cmd/common/viper.go
```diff
@@ -168,25 +169,24 @@ func isVersionCmd() bool {
...
 	}
 	if vConfigError != nil {
-		l.Error("unable to load config file", "error", vConfigError)
-		return
-	}
-	if cfgFile := v.ConfigFileUsed(); cfgFile != "" {
-		l.Info("using config file", "location", cfgFile)
+		return fmt.Errorf("unable to load config file: %w", vConfigError)
 	}
+	cfgFile := v.ConfigFileUsed()
```
AustinAbro321, line 187: We already return if a config file is found, so no need to check for an empty string

mkcp (approved):
LGTM

mkcp (approved):
Rereviewed as discussed

PR #3212, opened 2024-11-07
feat: introduce slog for zarf tools

AustinAbro321:
## Description

Introduces slog for zarf tools. Because of the complexity around cmd, I decided to make the crane commands use the default logger, as right now they don't receive flags properly. This should be fixed when cmd is refactored. 

Additionally the largest change here is around `zarf tools update-creds` which has a very unique output at the moment. Here are the changes side by side
image
image

## Related Issues

Relates to #2576

## Checklist before merging


src/pkg/message/credentials.go
```diff
@@ -60,43 +62,56 @@ func PrintCredentialTable(state *types.ZarfState, componentsToDeploy []types.Dep
...
 		fmt.Println(state.RegistryInfo.PullPassword)
 	default:
 		Warn("Unknown component: " + componentName)
+		l.Warn("unknown component", "component", componentName)
 	}
 }
 
 // PrintCredentialUpdates displays credentials that will be updated
-func PrintCredentialUpdates(oldState *types.ZarfState, newState *types.ZarfState, services []string) {
+func PrintCredentialUpdates(ctx context.Context, oldState *types.ZarfState, newState *types.ZarfState, services []string) {
```
mkcp, line 95: Coupling message together with logger should be avoided. What if instead we copied `PrintCredentialUpdates()` over to a helper in `cmd/tools/zarf.go` and then swapped out all of the message calls for logger calls? Then the new presentation logic stays in the presentation layer (cmd) and we don't intertwine both libraries.

src/pkg/message/credentials.go
```diff
@@ -60,43 +62,56 @@ func PrintCredentialTable(state *types.ZarfState, componentsToDeploy []types.Dep
 }
 
 // PrintComponentCredential displays credentials for a single component
-func PrintComponentCredential(state *types.ZarfState, componentName string) {
+func PrintComponentCredential(ctx context.Context, state *types.ZarfState, componentName string) {
```
mkcp, line 65: See comments below in on `PrintCredentialUpdates()`. Recommend duplicating the control flow, adding logger in its place, and moving the function nearer to the cmd it's used in.

mkcp (changes requested):
Most of the changes look good but I have reservations about coupling logger into message. Suggested an alternative.

mkcp (approved):
LGTM

Issue #3223, opened 2024-11-08
Send PTerm tables to os.Stdout not logger.DefaultDestination


mkcp:
### Is your feature request related to a problem? Please describe.
This is rly more of a bug that I started doing while I was adding in structured logs. Tables needed to be _reinitialized_ when pterm is feature flagged off -- when I put this hack together I passed them `logger.DefaultDestination` which happens to be stderr.

### Describe the behavior you'd like
Sending tables that are definitely "output for users" to stderr isn't very unixy, so let's instead send them to stdout. Let's also declare the writer _outside_ of logger. These are similar but not the same components, so there's no need to use a logger type for output.

Most if not all of of these table prints are well contained to the cmd layer. Here's some examples, this should not be considered an exhaustive list:
- https://github.com/zarf-dev/zarf/blob/36b779f13b4b2f46bc028f093ff1f2ca47b14c77/src/cmd/dev.go#L266
- https://github.com/zarf-dev/zarf/blob/36b779f13b4b2f46bc028f093ff1f2ca47b14c77/src/cmd/dev.go#L317
- https://github.com/zarf-dev/zarf/blob/36b779f13b4b2f46bc028f093ff1f2ca47b14c77/src/cmd/package.go#L240
- https://github.com/zarf-dev/zarf/blob/36b779f13b4b2f46bc028f093ff1f2ca47b14c77/src/cmd/package.go#L286

### Additional context
Ideally we don't have to use this re-initialize hack at all, but that may have to be in scope for a future issue.

AustinAbro321:
Fixed by https://github.com/zarf-dev/zarf/pull/3226

PR #3226, opened 2024-11-08
feat: render tables on stdout

mkcp:
## Description
This PR adjusts a hack from before, which made sure we still rendered tables even when message / pterm was disabled via feature flags.

Now, we render them to Stdout instead of rendering them on Stderr like diagnostic logs. There is a slightly insane hack here to retrieve the current writer from a new atomic package var, then reinitialize Pterm with os.Stdout, and finally initialize it back to the original value.

This hack should be addressed as soon as possible with output writers that can be defined and utilized locally and are threadsafe, at least a far as providing atomic stdout writes, if possible.

## Related Issue
Finishes #3223
Relates to #2576 

## Checklist before merging


mkcp:
Hold off on review or merge, I need to look over the failing e2e tests.

mkcp:
reworking this approach, we shouldn't have to update globals

mkcp:
Much better now - just chasing down failing tests.

AustinAbro321 (approved):

PR #3234, opened 2024-11-12
fix: add check during image discovery to make sure images are valid

a1994sc:
## Description

When using `zarf` to discover images with the Ironbank chart `kyverno-policies`, the `restrict-image-registries` policy, (simple version included in tests), has an entry of:

- `registry.dso.mil/*`

This is a false positive and cases cosign looks up to fail as that is not a valid image.

## Checklist before merging


a1994sc:
Related to #3253

src/pkg/packager/prepare.go
```diff
@@ -37,7 +38,8 @@ import (
 	"github.com/zarf-dev/zarf/src/types"
 )
 
-var imageCheck = regexp.MustCompile(`(?mi)"image":"([^"]+)"`)
+var imageCheck = regexp.MustCompile(`(?mi)"image":"((([a-z0-9._-]+)/)?([a-z0-9._-]+)(:([a-z0-9._-]+))?)"`)
+var imagePodCheck = regexp.MustCompile(`(?mi)^(([a-z0-9._-]+)/)?([a-z0-9._-]+)(:([a-z0-9._-]+))?$`)
```
brandtkeller, line 42: I believe this regex doesn't account for registry domain names or potentially a local docker registry (ip:port). While I am not a regex expert - I believe something to the effect of the suggestion below to accomplish the intent. 

```suggestion
var imagePodCheck = regexp.MustCompile(`(?mi)^(?:(?:[a-z0-9._-]+(?:\.[a-z]{2,})?|(?:\d{1,3}\.){3}\d{1,3})(?::\d+)?)/((?:[a-z0-9._-]+/)*)?([a-z0-9._-]+)(:([a-z0-9._-]+))?$`)
```

a1994sc: that regex did not workout.... _so_ I borrowed a regex from the `github.com/containers/image` team.... it is just the image reference section but it seems to be a fairly easy drop in replacement for my first regex...

I ran the failing test locally and it is still not happy...

```shell
TMPDIR=/tmp make test-e2e-without-cluster ARCH=amd64
```

So any help on that would be super helpful

brandtkeller:
> Any reason those two test are stuck?

GitHub reports "unexpected error" - so I'm going to give those a retry real quick.

a1994sc:
> > Any reason those two test are stuck?
> 
> GitHub reports "unexpected error" - so I'm going to give those a retry real quick.

Ah, thank you very much!!

brandtkeller (approved):
This definitely improves the baseline. I'll let another maintainer weigh-in if there is any patterns or constraints we want to follow with regards to the regex logic - otherwise LGTM.

a1994sc:
Thanks!! I was a little worried that because you all are migrating to `package2` it might have needed to be put on hold. Though very happy to help an awesome project!

mkcp (approved):
LGTM. Good test case and defense for the logic. Matching the license and incl. attribution makes accepting this easy. Nice job

a1994sc:
I would be happy to help with the effort to port over to the `package2` implementation, whenever that works starts

Issue #2902, opened 2024-08-16
Support for image volume source 


brandtkeller:
### Is your feature request related to a problem? Please describe.

Kubernetes 1.31 introduced [image volume](https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/) source in alpha.

As this is an associated OCI artifact - Zarf may require the need to be checking for images used as volumes during mutation in the agent. 

### Describe the solution you'd like

- **Given** a pod uses an image volume source 
- **When** the mutation of a pod ocurs
- **Then** the image reference is updated to the zarf registry

### Additional context

[KEP-4639](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/4639-oci-volume-source)

Understand this is an alpha feature that requires the `ImageVolume` feature gate - opening this issue for tracking intent to support if not intending to support until graduation.

AustinAbro321:
IMO given that this would be a simple, nonbreaking, change to the agent, this is something that doesn't need to wait until graduation to implement. IMO it'll be reasonable to add once tools like k3s and eks support 1.31

phillebaba:
This needs to be discussed, regarding if we want to support Alpha features. I am a fan of OCI volumes but this feature will change in the future. 

@AustinAbro321 this will not hit EKS when 1.31 gets there. Alpha features and for a while now Beta feature are disabled by default. We wont see any major adoption until it gets into a graduated state.

If we end up supporting this for a specific use case it should be marked with a big experimental flag that allows us to break it at any time.

willswire:
It [appears](https://github.com/kubernetes/enhancements/pull/5450) that Kubernetes v1.35 may include GA for `ImageVolume`. We may also want to keep a close eye on https://github.com/k3s-io/k3s/issues/12390 which will include the necessary update for containerd to support image volumes.

brandtkeller:
Thanks for the added context @willswire . Given that a target is approaching for GA - I believe it's important to start scoping this effort and work down the path of requirements (logic, testing, etc). 

@a1994sc you have an open PR for this - are you interested in pursuing this further? (if so I will assign you and provide more feedback on the PR directly)

a1994sc:
@brandtkeller I would be more then happy to be assigned this ticket!

I was thinking about circling back on this around when 1.35 released, but would be happy to test on that `1.34.x` of k3s that has the support

brandtkeller:
Article currently states "enabled by default" in 1.35

[...]

PR #3235, opened 2024-11-12
feat(agent): mutate read-only OCI volumes

a1994sc:
## Description

This adds the ability for `zarf` to mutate pods that use the upcoming `image-volume` spec

```yaml
spec:
  ...
  volumes:
  - name: volume
    image:
      reference: quay.io/crio/artifact:v1
```

to

```yaml
spec:
  ...
  volumes:
  - name: volume
    image:
      reference: 127.0.0.1:31999/crio/artifact:v1-zarf-2568457951
```

As it uses the container run for pulling the image this should work like all the other mutations; however I need to test on a cluster that has the feature-gate enabled. So I will be leaving this as a draft-pr.

## Related Links

- `kubernetes/enhancements` #[4639](https://github.com/kubernetes/enhancements/issues/4639)
- [`kuberenets docs`](https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/)

## Related Issues

- #2902

## Checklist before merging


a1994sc:
Off site convo: Austin and I came to the understanding that this likely will not be merged in until after the kubernetes enhancement is stable

brandtkeller:
As discussed in #2902 - I believe this will be a key enabler when the supporting feature is GA in kubernetes. We have the intended logic and unit testing available already and the chance to support exploratory efforts by end users.

I would like to recommend pursuing this early - in an attempt to have it ready for consumers who want to begin testing the potential use cases. 

What I believe we need:
- Review existing testing infrastructure - determine paths we have available to orchestrate an E2E test
  - This may be an out-of-band test such that we orchestrate the feature flag appropriately (may not be viable across all of our distribution testing).
  - Should include some verification that mutation occurred AND that the test results in operational access to the data
- Determine boundaries - Zarf shouldn't need to be aware of whether the cluster supports the feature or not.
- Documentation - such as a reference tutorial - can follow in future PR's .

Thanks for the continued support @a1994sc

a1994sc:
So this is some initial notes, as such they are going to go into the comments rather then docs until I can sort thru and pick out the useful bits, (also will be a bit scatter brained/unordered):

## Determine boundaries:
  - I think that this one will be the "hardest" to design around because I see the following paths for 
    - we check the nodes in the cluster have a compatible level of Container Runtime and use that to determine if we are going to mutate or not....
    - we have an opt-in flag that enables the image mutation: `--feature-gates=ImageVolumeMutation=true`
  - From a code prospective, this PR already has the required code changes/tests to mutate the pods during creation such that it will the Kubernetes API will be happy.

## Review existing testing infrastructure:
  - So this maybe need to me to talk with the `Maintainers` team to determine what resources we have to work with...
    - in an ideal work we should expand our testing to include other popular CNCF distros:
      - [`k0s`](https://github.com/k0sproject/k0s)
        - this distro still dependent on `v1` of `containerd` because of `helm`, however with the release of `helm` v4 on the 12th of November 2025 that project might be able to update soon.
        - [Issue](https://github.com/k0sproject/k0s/issues/4581)
      - [`microk8s`](https://github.com/canonical/microk8s)
        - Does seem that the team over on this distro is will be updating support for `v2` of `containerd` with the release of `v1.35` of k8s.
        - [Issue](https://github.com/canonical/microk8s/issues/5230)

## Documentation:
I think that at a high level we should make it know what common distros we test against, and document any known pitfalls when using less common distros or non-AWS cloud providers.

brandtkeller:
> I think that this one will be the "hardest" to design around because I see the following paths for
we check the nodes in the cluster have a compatible level of Container Runtime and use that to determine if we are going to mutate or not....
we have an opt-in flag that enables the image mutation: --feature-gates=ImageVolumeMutation=true

How does the kube-api respond when attempting to create a pod with an image volume source? I am curious if Zarf needs to care about whether the underlying cluster can support it. 

> So this maybe need to me to talk with the Maintainers team to determine what resources we have to work with...
in an ideal work we should expand our testing to include other popular CNCF distros...

I agree with these points fundamentally. I would maybe take a stance that they should be pursued as follow-on work. I believe it is reasonable that we have an existing distro test in place that we could modify (or duplicate).

reframing this PR as:
1. mutation code + unit tests (done)
2. E2E test

Dependent on timing - we could instrument a separate kind/k3d workflow test now that opts-into the feature flag and runs the test and later we move that E2E test under standard E2E testing when we are testing consistently on >=1.35. 

Thoughts?

a1994sc:
> How does the kube-api respond when attempting to create a pod with an image volume source? I am curious if Zarf needs to care about whether the underlying cluster can support it.

I believe when I originally tested this locally the kubernetes api accepted the pod with the image volume, (both with and without the mutation), but then the container runtime complained it didn't know what to do. Let me circle back to you and get a copy of the error to throw into this discussion.

> I agree with these points fundamentally. I would maybe take a stance that they should be pursued as follow-on work. I believe it is reasonable that we have an existing distro test in place that we could modify (or duplicate).

I do agree that moving adding new distros as its own task would be ideal, seeing we would likely benefit from testing on more distros beyond this specific issue.

a1994sc:
So reviewing the patch notes from [`Kubernetes v1.35: Timbernetes (The World Tree Release)`](https://kubernetes.io/blog/2025/12/17/kubernetes-v1-35-release/#final-call-for-containerd-v1-x), there is mention Kubernetes will be dropping support for `v1.x` for containerd; this is an important milestone for broad support of "Image Volume Source", as that will leave only `v2.0` as the only version that does not support the feature.

Looking at the [`Current state of containerd releases`](https://containerd.io/releases/#current-state-of-containerd-releases), it does seem like `v2.0` is already EoL, so any distros that currently use the LTS version of containerd (`k0s`), are likely to skip over that version and go straight to a version that supports the Image Volume Source.... So we may not need super rigorous tests cases....

a1994sc:
I will be playing to tackle getting this PR ready for review this week, ideally.

a1994sc:
Some current notes:

## Commands

```bash
$ k3d cluster create --image docker.io/rancher/k3s:v1.35.0-k3s1
# checkout pod-volume-mutate
$ make build
$ make init-package
$ ./build/zarf package deploy ./build/zarf-init-amd64-v0.68.1-10-g9b3f8c25.tar.zst
$ mkdir -p /tmp/pod-volume-image && cd /tmp/pod-volume-image
# create the below yaml
$ zarf package create
$ zarf package deploy --confirm ./zarf-package-pod-volume-image-amd64-1.0.0.tar.zst 
```

## Errors

```text
Error creating: Internal error occurred: replace operation does not apply: doc is missing path: /spec/volume/0/image/reference: missing value
```

This seems to indicate that the kubernetes api, does not know how to handle creating a pod with the volume mount. However it should be enabled by default in `v1.35`

## Debugging

```bash
$ kubectl explain pod.spec.volumes.image.reference
KIND:       Pod
VERSION:    v1

FIELD: reference <string>


DESCRIPTION:
    Required: Image or artifact reference to be used. Behaves in the same way as
...
```

Seems like the k8s api does know about the pod field...

## Files

`zarf.yaml`
```yaml
---
kind: ZarfPackageConfig
metadata:
  name: pod-volume-image
  version: 1.0.0
components:
  - name: baseline
    required: true
    manifests:
      - name: pod-volume-image
        namespace: pod-volume-image
        files:
          - deployment.yaml
    images:
      - docker.io/library/debian:latest
      # Possible images - baseline
      - quay.io/crio/artifact:v1
```

`deployment.yaml`
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: agent
  namespace: pod-volume-image
spec:
  replicas: 1
  selector:
    matchLabels:
      app: agent
  template:
    metadata:
      labels:
        app: agent
    spec:
      containers:
        - name: shell
          command:
            - "sleep"
            - "infinity"
          image: docker.io/library/debian:latest
          resources:
            limits:
              cpu: 500m
              memory: 256Mi
            requests:
              cpu: 100m
              memory: 64Mi
          volumeMounts:
            - name: volume
              mountPath: /volume
              subPath: dir
      volumes:
        - name: volume
          image:
            reference: quay.io/crio/artifact:v1
            pullPolicy: IfNotPresent
```

a1994sc:
So I do believe that this PR is ready for review, it should cover the basics of what is needed for mounting the image volumes.

It does not have anything for if the feature is not supported.... not sure the best way to tackle that tbh... I would be happy to hear some input

src/test/e2e/45_volume_image_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	pkgDefinitionPath := filepath.Join("src", "pkg", "packager", "testdata", "find-images", "pod-volume-image")
+	tmpdir := t.TempDir()
+	_, _, err := e2e.Zarf(t, "package", "create", pkgDefinitionPath, "-o", tmpdir)
+	require.NoError(t, err)
+
+	packageName := fmt.Sprintf("zarf-package-pod-volume-image-%s.tar.zst", e2e.Arch)
+	path := filepath.Join(tmpdir, packageName)
+
+	_, _, err = e2e.Zarf(t, "package", "deploy", path, "--confirm", "--skip-version-check")
```
AustinAbro321: Skip version check is not required for this feature. Same during remove

src/test/e2e/45_volume_image_test.go
```diff
@@ -0,0 +1,34 @@
...
+	require.NoError(t, err)
+
+	stdOut, _, err := e2e.Kubectl(t, "get", "pod", "-l", "app=agent", "-n", "pod-volume-image", "-o", "jsonpath={.items[].status.phase}")
+	require.NoError(t, err)
+	require.Equal(t, "Running", stdOut)
```
AustinAbro321: Rather than checking if a pod is running, a deployed pod should always be running unless there is a bug in the wait logic, we should check that the volume was properly mutated. An additional check we could do would be to ensure that the volume exists in the container with a kubectl shell command.

src/test/e2e/45_volume_image_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"fmt"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestVolumeImageMount(t *testing.T) {
```
AustinAbro321, line 16: Right now there are no test clusters running 1.35 and therefore no clusters actually doing the mutate correctly. My recommendation would be to add `--image rancher/k3s:v1.35.0-k3s1` to the k3d cluster create step in actions, then in the this test we could add the following code. As the other distros are updated to newer versions they will work with this and we can delete this check in the future

```go
	// Skip test if Kubernetes version is less than 1.35.0
	stdOut, _, err := e2e.Kubectl(t, "version", "--output=json")
	require.NoError(t, err)

	var versionInfo struct {
		ServerVersion struct {
			GitVersion string `json:"gitVersion"`
		} `json:"serverVersion"`
	}
	err = json.Unmarshal([]byte(stdOut), &versionInfo)
	require.NoError(t, err)

	k8sVersion, err := semver.NewVersion(versionInfo.ServerVersion.GitVersion)
	require.NoError(t, err)

	if k8sVersion.LessThan(semver.MustParse("v1.35.0")) {
		t.Skipf("Skipping test: Kubernetes version %s is less than 1.35.0", versionInfo.ServerVersion.GitVersion)
	}
```

AustinAbro321 (changes requested):
Nice work so far, some comments on the test

src/test/e2e/45_volume_image_test.go
a1994sc: That is a very clean way to handle that, I will update the tests. Thank you very much

src/test/e2e/45_volume_image_test.go
a1994sc: Gotcha, I had copied the test file from another, so wasn't super sure what was required.

src/test/e2e/45_volume_image_test.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	err = json.Unmarshal([]byte(stdOut), &versionInfo)
+	require.NoError(t, err)
+
+	k8sVersion, err := semver.NewVersion(versionInfo.ServerVersion.GitVersion)
+	require.NoError(t, err)
+
+	if k8sVersion.LessThan(semver.MustParse("v1.35.0")) {
+		t.Skipf("Skipping test: Kubernetes version %s is less than 1.35.0", versionInfo.ServerVersion.GitVersion)
+	} else {
```
AustinAbro321: nit: don't need the else block here, t.skip is essentially a return

AustinAbro321 (commented):
one minor nit otherwise lgtm. Also lets rename the title to follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). Maybe "feat(agent): mutate read-only OCI volumes"

a1994sc:
> one minor nit otherwise lgtm. Also lets rename the title to follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). Maybe "feat(agent): mutate read-only OCI volumes"

Do I need to change the PR name or should I condense my commits and use that commit message?

src/test/e2e/45_volume_image_test.go
a1994sc: Ah ok, I will update. I didn't know that it worked that way

AustinAbro321:
Condensing commits won't have an impact since we always use squash and merge. The pr title ends up being the commit in git history

a1994sc:
> Condensing commits won't have an impact since we always use squash and merge. The pr title ends up being the commit in git history

Ah, ok, I already updated the PR title; I will update the test this evening

src/internal/agent/hooks/pods.go
```diff
@@ -44,12 +44,20 @@ func parsePod(object []byte) (*corev1.Pod, error) {
...
+}
+
+func getAnnotationKey(ctx context.Context, image string) string {
+	annotationName := fmt.Sprintf("original-%s", image)
 	// The name segment is required and must be 63 characters or less, beginning and ending with
 	// an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.
 	// https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
 	if len(annotationName) > 63 {
-		logger.From(ctx).Debug("truncating container name to fit Kubernetes 63 character annotation name limit", "container", containerName)
+		logger.From(ctx).Debug("truncating container name to fit Kubernetes 63 character annotation name limit", "container", image)
```
brandtkeller, line 60: non-blocking - I wonder what the realistic likely hood of truncation is - and if we should evaluate a smaller trunctation + hash to prevent collision?

src/internal/agent/hooks/pods.go
```diff
@@ -120,6 +128,19 @@ func mutatePod(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster.Clu
 		patches = append(patches, operations.ReplacePatchOperation(path, replacement))
 	}
 
+	// update the image host for each volume that contains an "image" reference
+	for idx, volume := range pod.Spec.Volumes {
+		if volume.Image != nil {
```
brandtkeller, line 133: Should an errors occur during this process - the user usually faces a stalled deploy and has to go digging in the agent logs. 

I'd like to consider a modification:
Check for empty reference - this is an early return scenario that we can add traceability around:

```go
if volume.Image != nil {
    if volume.Image.Reference == "" {
		return nil, fmt.Errorf("volume %q (index %d) has an ImageVolumeSource with empty reference - this is invalid and must be specified",
			volume.Name, idx)
		}
}
```

I can't see other known error handling that we might want to include as of yet.

src/internal/agent/hooks/pods.go
```diff
@@ -120,6 +128,19 @@ func mutatePod(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster.Clu
...
+			path := fmt.Sprintf("/spec/volumes/%d/image/reference", idx)
+			replacement, err := transform.ImageTransformHost(registryURL, volume.Image.Reference)
+			if err != nil {
+				return nil, err
```
brandtkeller: similar to the above - great time to add more clarity and traceability:


```suggestion
				return nil, fmt.Errorf("failed to transform volume %q (index %d) image reference %q: %w", volume.Name, idx, volume.Image.Reference, err)
```

brandtkeller (changes requested):
Minor changes otherwise testing this locally works as expected - I was able to take the test package and deploy -> shell -> `/volume/zarf version` -> "v0.68.1" 

Very excited to see this operational - thanks for the persistence!

AustinAbro321 (approved):
lgtm

brandtkeller (approved):
LGTM

Issue #2865, opened 2024-08-10
Allow `zarf dev find-images` update `zarf.yaml` with discovered images


a1994sc:
### Is your feature request related to a problem? Please describe.

When running `zarf dev find-images` it only output to the console, it does not update the `zarf.yaml` file.

To not break the current functionality it would be nice to have a flag like `-u`/`--update-zarf-manifest` to have the update the manifest for us.

### Describe the solution you'd like

- **Given** an update to the zarf manifests/zarf charts
- **When** the `-u`/`--update-zarf-manifest` is provided
- **Then** the `zarf.yaml` file `.components[].images` array is updated.

### Additional context

This is mainly aimed at creating CI/CD around zarf and renovate, so when a helm package is updated via Renovate then zarf will update the manifest and other parts of the pipeline will push those changes.

PR #3236, opened 2024-11-13
chore: refactor `findImages` to remove print statements

a1994sc:
## Description

Refactor of `findImages` remove print states from inside the function. It now passes the image map to the `devFindImagesCmd` which prints a similar table as before.

A regression is that we lost the easy ability add the comment blocks for the `maybe` images and `cosign` artifacts.

## Related Issue

Relates to #2865

## Checklist before merging


src/cmd/common/table.go
```diff
@@ -14,6 +14,27 @@ import (
...
+
+// PrintComponentTable prints the components with found images.
+func PrintComponentTable(imagesMap map[string][]string) {
+	fmt.Println(COMPONENTS)
+	for component, images := range imagesMap {
+		fmt.Printf(NAME, component)
+		if len(images) > 0 {
+			fmt.Println(IMAGES)
+			for i := 0; i < len(images); i++ {
+				fmt.Printf(IMAGE, images[i])
```
a1994sc, line 32: I was not sure if printing them out with the `fmt.Print` functions was the correct way to handle this, I did not see a good log level to put it in the logger.

Please let me know if you want me to update this

phillebaba:
We are currently in the process of refactoring the pacakger function so find images will be completely re-written. During this refactoring printing will be deferred to the CLI.

a1994sc:
Ah ok, thank you @phillebaba for the clarification. I'm wanting to put some effort into addressing #2865 for automatically updating the `zarf.yaml` file. Should I assume that `findImages` will still return an array with the images that zarf discovered still?

a1994sc:
I can also close this PR if that is preferred for you

phillebaba:
I am not looking into changing the existing behavior for find images. So the refactor will still in the end print the found images. I would suggest bringing up the issue to change the behavior of find images at the next community meeting.

PR #3238, opened 2024-11-13
feature: Add ability for zarf to find oci artifacts for fluxcd ocirepo resource

a1994sc:
## Description

When trying to discover images with the `OCIRepositry` kind from FluxCD, `zarf` did not have logic for finding them with `zarf dev find-images`

## Checklist before merging


a1994sc:
Fixes #3255

phillebaba:
As with #3236 we are currently refactoring findImages so it will be hard to take this PR at this moment.

a1994sc:
OK, that works for me @phillebaba, thank you very much.

AustinAbro321:
@a1994sc Now that find images is refactored, would you be willing to recreate this PR in [the new implementation](https://github.com/zarf-dev/zarf/blob/main/src/internal/packager2/find_images.go)

a1994sc:
Updated the PR

src/internal/packager2/find_images.go
```diff
@@ -482,6 +490,30 @@ func appendToImageMap(imgMap map[string]bool, pod corev1.PodSpec) map[string]boo
...
+	s.Write([]byte(t[1]))
+
+	if repo.Reference.Tag != "" {
+		s.WriteString(":")
+		s.WriteString(repo.Reference.Tag)
+	} else if repo.Reference.Digest != "" {
+		s.WriteString("@")
+		s.WriteString(repo.Reference.Digest)
+	} else if repo.Reference.SemVer != "" || repo.Reference.SemverFilter != "" {
+		fmt.Errorf("Can not use semver or semverFilter with OCIRepository")
```
AustinAbro321, line 508:
```suggestion
		logger.From(ctx).Error("can not use semver or semverFilter with OCIRepository")
```
Pass context here and use the builtin logger

src/internal/packager2/find_images.go
```diff
@@ -482,6 +490,30 @@ func appendToImageMap(imgMap map[string]bool, pod corev1.PodSpec) map[string]boo
 	return imgMap
 }
 
+func appendToImageMapOCIRepo(imgMap map[string]bool, repo sourcev1beta2.OCIRepositorySpec) map[string]bool {
+	var regex = regexp.MustCompile(`oci://(.+)`)
+	var t = regex.FindStringSubmatch(repo.URL)
+
+	s := strings.Builder{}
+	// Pulls out just the url from the regex
+	s.Write([]byte(t[1]))
```
AustinAbro321, line 499: According to the flux docs the `oci://` is required, so I think it'd be simpler to do a `strings.TrimPrefix` here https://fluxcd.io/flux/components/source/ocirepositories/#url

AustinAbro321 (changes requested):
small comments, otherwise looks good. Nice work!

src/internal/packager2/find_images.go
a1994sc: Updated per comment

src/internal/packager2/find_images.go
a1994sc: Updated the logic to use the `strings.TrimPrefix` func

src/internal/packager2/find_images_test.go
```diff
@@ -139,6 +139,35 @@ func TestFindImages(t *testing.T) {
...
+		{
+			name:        "ocirepo-semver",
+			packagePath: "./testdata/find-images/flux-oci-repo-semver",
+			opts: FindImagesOptions{
+				SkipCosign: true,
+			},
+			expectedImages: []ComponentImageScan{
+				{
+					ComponentName: "baseline",
+					Matches:       []string{},
```
a1994sc, line 167: Not sure that I handled this error correctly... I tried setting a `expectedErr:` but it didn't seem to pass... so any input would be very grateful

AustinAbro321: Yeah `expectErr` won't work, because we're not returning an error when we run into the semver check just logging, which makes sense, given that we want the rest of find-images to run in this case. IMO just delete this test case.

src/internal/packager2/find_images.go
```diff
@@ -464,6 +472,23 @@ func appendToImageMap(imgMap map[string]bool, pod corev1.PodSpec) map[string]boo
...
 
+func appendToImageMapOCIRepo(ctx context.Context, imgMap map[string]bool, repo sourcev1beta2.OCIRepositorySpec) map[string]bool {
+	var url = strings.TrimPrefix(repo.URL, "oci://")
+
+	if repo.Reference.Tag != "" {
+		url = url + ":" + repo.Reference.Tag
+	} else if repo.Reference.Digest != "" {
+		url = url + "@" + repo.Reference.Digest
+	} else if repo.Reference.SemVer != "" || repo.Reference.SemverFilter != "" {
+		logger.From(ctx).Error("can not use semver or semverFilter with OCIRepository")
```
AustinAbro321, line 483: We should use the `Warn` function rather than the `Error` function. We should also give users the image url, and oci repository name so they can easily find it. (note: you'll have to pass in `sourcev1beta2.OCIRepository` for this) 
```suggestion
		logger.From(ctx).Warn("cannot create image reference with semver or semverFilter", "image", url, "OCIRepository", repo.Name)
```

a1994sc: image

a1994sc: Updated PR with your comments Austin

AustinAbro321 (changes requested):
Just remove the `flux-oci-repo-semver` directory and I will approve

AustinAbro321:
@a1994sc Running tests right now and if everything passes I'll approve. FYI we use a GitHub feature called merge queue which runs all tests against this branch with main merged into it before the actual merge happens, so no worries if your branch is not up to date.

AustinAbro321 (approved):

mkcp (approved):

PR #3241, opened 2024-11-15
Actually use emptydir mount when persistence is disabled

### LINKED ISSUE #1998: zarf init fails on multi-node minikube cluster
### Environment
Device and OS: MacBook Pro M2 Pro, MacOS
App version: v0.29.1
Kubernetes distro being used: MiniKube (Docker driver) v1.31.1, k8s v1.24.7
Other:

### Steps to reproduce
1. `minikube start --kubernetes-version 1.24.7 --nodes 3`
2. `zarf init --confirm`

### Expected result
It works

### Actual Result
The registry fails to deploy

Pod Logs (zarf-docker-registry):
```shell
Configuration error: error parsing /etc/docker/registry/config.yml: no storage configuration provided
Usage: 
  registry serve <config> [flags]
Flags:
  -h, --help=false: help for serve
Additional help topics:
```

It just never deploys, and the Helm deploy just keeps trying till it times out, spitting out essentially the same error as above.

### Severity/Priority
High: needed for our developer environment for testing HA application deployments. Not affecting prod (yet)

### Additional Information
We tried this as well, which resulted in 1 registry pod deploying, but the 2nd pod had the same error as listed above
`zarf init --storage-class=standard --set REGISTRY_PVC_ENABLED=false --confirm`

xyzzyz:
## Description
In #2124, an emptydir was mounted when `persistence.enabled: false` is set. However, this is not enough to make registry work: it crashloops as in issue #1998:
```
Configuration error: error parsing /etc/docker/registry/config.yml: no storage configuration provided
```

 This is fixed by setting the env variable specifying storage config even if persistence is disabled:

## Related Issue

Fixes #1998

## Checklist before merging


AustinAbro321:
Hey, really appreciate this PR, and this is a bug. The `persitence.enabled` block on there was added specifically with a s3 backend in mind. I'm not sure if this change would break that use case, but the team and I will need to add a test to the Zarf repo first before I'd feel comfortable with this change. Marking this as blocked for now, and I will circle back when that test is merged.

xyzzyz:
To provide some context, I use this to bootstrap Zarf on a cluster where there's no PV provisioner available. The way I do it is to install Rancher's  `local-path-provisioner` as part of Zarf init package:

```
components:
  - name: zarf-injector
    required: true
    import:
      path: zarf/packages/zarf-registry

  # Creates the temporary seed-registry with ephemeral PV to install local path provisioner
  - name: zarf-seed-registry
    required: true
    import:
      path: zarf/packages/zarf-registry
    actions:
      onDeploy:
        before:
          - cmd: echo "false"
            setVariables:
              - name: REGISTRY_PVC_ENABLED


  # Creates the temporary seed local path provisioner with ephemeral PV to install local path provisioner
  - name: seed-local-c-provisioner
    required: true
    import:
      path: zarf-local-path-provisioner

  - name: zarf-registry
    required: true
    import:
      path: zarf/packages/zarf-registry
    actions:
      onDeploy:
        before:
          - cmd: echo "true"
            setVariables:
              - name: REGISTRY_PVC_ENABLED

  # Reinstall local path provisioner to get its image inserted into permanent registry
  - name: local-path-provisioner
    required: true
    import:
      path: zarf-local-path-provisioner

  - name: zarf-agent
    required: true
    import:
      path: zarf/packages/zarf-agent
```

Without this, I'd need to make `local-path-provisioner` installation part of pre-Zarf process.

brandtkeller:
I'd like to revisit this PR and review two items:
1. If this modification should be able to dynamically handle the permutations
  1.a. It appears as though there are three possible scenarios: emptyDir, S3, and a k8s PVC.  There could be more room here to improve the flexibility to support the required environment variable when needed.
2. The intended behavior can be achieved via a workaround such as:
```yaml
 # zarf.yaml
  - name: zarf-seed-registry
    required: true
    import:
      url: oci://ghcr.io/zarf-dev/packages/init:v0.63.0
    charts:
      - name: docker-registry
        valuesFiles:
          - overrides/registry-values-longhorn.yaml
```

and 

```yaml
# overrides/registry-values-longhorn.yaml
persistence:
  enabled: false

extraEnvVars:
  - name: REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY
    value: "/var/lib/registry"

```

@xyzzyz are you still interested in pursuing this update or would you like to close it out?

PR #3259, opened 2024-11-21
feat: add logger to packager.Publish

mkcp:
## Description
This one ended up being really simple because most of the downstream code had already been ported to slog. I started updating the tests to use the new logging, but decided that it's better if we wait until we swap over to the default format. Added a couple review comments for discussion.

## Related Issue
Relates to #2576 

## Checklist before merging

### DIFF
--- a/src/pkg/packager/publish.go
+++ b/src/pkg/packager/publish.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/defenseunicorns/pkg/helpers/v2"
 	"github.com/defenseunicorns/pkg/oci"
@@ -17,6 +18,7 @@ import (
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
 	"github.com/zarf-dev/zarf/src/config"
 	"github.com/zarf-dev/zarf/src/pkg/layout"
+	"github.com/zarf-dev/zarf/src/pkg/logger"
 	"github.com/zarf-dev/zarf/src/pkg/message"
 	"github.com/zarf-dev/zarf/src/pkg/packager/creator"
 	"github.com/zarf-dev/zarf/src/pkg/packager/filters"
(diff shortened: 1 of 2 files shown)


src/pkg/packager/publish.go
```diff
@@ -123,5 +130,10 @@ func (p *Packager) Publish(ctx context.Context) (err error) {
 			return err
 		}
 	}
+	l.Info("packaged successfully published",
```
mkcp, line 133: Usually we print the success logs at debug, but this one is a network request that could take some time, so I opted to make it an info so users get some feedback. Not much happens after this so I could go either way.

src/test/e2e/50_oci_publish_deploy_test.go
```diff
@@ -48,7 +48,6 @@ func (suite *PublishDeploySuiteTestSuite) Test_0_Publish() {
 	ref := suite.Reference.String()
 	stdOut, stdErr, err := e2e.Zarf(suite.T(), "package", "publish", example, "oci://"+ref, "--plain-http")
 	suite.NoError(err, stdOut, stdErr)
-	suite.Contains(stdErr, "Published "+ref)
```
mkcp, line 51: Checking stderr for a specific string feels like a change detector -- any actual error in the process should be caught by the suite.NoError above. Initially I set these to match in both message and logger, but upon more consideration we should just delete these checks and make sure our error trees work. Thoughts?

src/pkg/packager/publish.go
AustinAbro321: I think info here makes sense, it gives the URL on the last line of logs which is nice since that's what the user will likely look for.  I'd prefer this line in `remote.PublishPackage` since `remote.PublishPackage` is called elsewhere. (If you run `zarf package create --output oci://registry`), but we have to use the printf format until we switch off old logs.

AustinAbro321: Talked synchronously, good with this here

AustinAbro321 (approved):

PR #3265, opened 2024-11-21
feat: add the new logger through Zarf wherever it is missing

### LINKED ISSUE #2576: Refactoring of logging

AustinAbro321:
## Description

Went back through the Zarf codebase and added the new logging commands to places where we may have missed on the first pass through. There are some errors that we will never reach or are very unlikely to reach where we run a `message.Debug`. In these situations, I just replaced message with the new logger now. 

Fixes #2576 

## Checklist before merging


src/internal/packager/helm/repo.go
AustinAbro321, line 1: whim of the autoformatter

src/pkg/interactive/components.go
```diff
@@ -25,6 +25,7 @@ func SelectOptionalComponent(component v1alpha1.ZarfComponent) (bool, error) {
 		return false, err
 	}
 	if component.Description != "" {
+		// TODO (@austinabro321) once we move interactiveness to CLI level we should replace this with logger.Info
```
mkcp, line 28: ++

mkcp (approved):
LGTM

mkcp (approved):

PR #3286, opened 2024-12-03
feat: allow init packages to be explicitly versioned

AustinAbro321:
## Description

In #3169 we allowed Zarf init packages to be explicitly versioned rather than always being the binary version. This behavior was not brought over to packager2

## Checklist before merging


mkcp (approved):
LGTM

PR #3293, opened 2024-12-05
chore: remove deprecated get git password command

AustinAbro321:
## Description

This removes the deprecated get git password command which has been deprecated for over a year. It was replaced by `zarf tools get-creds git`. This command was already broken because when the get-creds command was updated to use the `RunE` function, `get-git-password` was not updated so it panic'd

## Checklist before merging

### DIFF
--- a/src/cmd/dev.go
+++ b/src/cmd/dev.go
@@ -237,7 +237,7 @@ var devSha256SumCmd = &cobra.Command{
 }
 
 var devFindImagesCmd = &cobra.Command{
-	Use:     "find-images [ PACKAGE ]",
+	Use:     "find-images [ DIRECTORY ]",
 	Aliases: []string{"f"},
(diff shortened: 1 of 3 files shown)


src/cmd/dev.go
```diff
@@ -237,7 +237,7 @@ var devSha256SumCmd = &cobra.Command{
 }
 
 var devFindImagesCmd = &cobra.Command{
-	Use:     "find-images [ PACKAGE ]",
+	Use:     "find-images [ DIRECTORY ]",
```
AustinAbro321, line 240: sneaking this in a bit, but noticed this while writing the ZEPs. This runs on a directory not a package. Switching this to use the same verbiage used in `zarf package create`.

schristoff (approved):

PR #3303, opened 2024-12-09
fix: passing context into packager

AustinAbro321:
## Description

There is currently a small bug in the when using the `--log-format` flag that causes not all logs to be seen as context is not passed properly into the packager struct. 

Passing context into structs is not ideal, this was done to avoid breaking changes. The packager struct will be removed once #2969 is completed

## Checklist before merging


mkcp (approved):
🙏 LGTM

mkcp (approved):

PR #3305, opened 2024-12-09
test: remove bad remote image test

AustinAbro321:
## Description

Testing that Zarf will fail when a remote image is pulled is unnecessary

## Checklist before merging

### DIFF
--- a/src/test/e2e/00_use_cli_test.go
+++ b/src/test/e2e/00_use_cli_test.go
@@ -116,16 +116,6 @@ func TestUseCLI(t *testing.T) {
 		require.Contains(t, stdErr, expectedOutString, "The log level should be changed to 'debug'")
 	})
 
-	t.Run("zarf package to test bad remote images", func(t *testing.T) {
-		_, stdErr, err := e2e.Zarf(t, "package", "create", "src/test/packages/00-remote-pull-fail", "--confirm")
-		// expecting zarf to have an error and output to stderr
-		require.Error(t, err)
-		// Make sure we print the get request error (only look for GET since the actual error changes based on login status)
-		require.Contains(t, stdErr, "failed to find the manifest on a remote: GET")
-		// And the docker error
-		require.Contains(t, stdErr, "response from daemon: No such image")
-	})
-
 	t.Run("zarf package to test archive path", func(t *testing.T) {
 		t.Parallel()
 		stdOut, stdErr, err := e2e.Zarf(t, "package", "create", "packages/distros/eks", "--confirm")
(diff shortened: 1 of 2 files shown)


src/internal/packager/images/pull_test.go
```diff
@@ -84,6 +85,54 @@ func TestCheckForIndex(t *testing.T) {
 }
 
 func TestPull(t *testing.T) {
+	t.Parallel()
+	testCases := []struct {
+		name        string
+		ref         string
+		expectedErr string
```
mkcp, line 92: Often I'll do two different tables for happy and unhappy path. The logic often branches and that can get weird with the test type. This works fine for now though.

src/internal/packager/images/pull_test.go
```diff
@@ -84,6 +85,54 @@ func TestCheckForIndex(t *testing.T) {
...
+			pullConfig := PullConfig{
+				DestinationDirectory: destDir,
+				ImageList: []transform.Image{
+					ref,
+				},
+			}
+
+			pulled, err := Pull(context.Background(), pullConfig)
+			if tc.expectedErr != "" {
+				require.ErrorContains(t, err, tc.expectedErr)
```
mkcp, line 119: If you can do an errors.Is that compares a typed error to another that's ideal. Checking text can be fragile in the same way as checking console output. No change requested though, just something to keep in mind.

mkcp (approved):

src/internal/packager/images/pull_test.go
AustinAbro321: Yeah agree, when we things to happen with the error or have more to check than it just being there. Keeping this as is for now since it still feels simple

src/internal/packager/images/pull_test.go
AustinAbro321: I actually did decide to change it. The library (crane) doesn't expose these errors so I don't want tests to break if they change. In the future we can always update this if we have more test cases

src/internal/packager/images/pull_test.go
```diff
@@ -84,6 +85,53 @@ func TestCheckForIndex(t *testing.T) {
...
+		expectErr bool
+	}{
+		{
+			name: "pull an image",
+			ref:  "ghcr.io/zarf-dev/zarf/agent:v0.32.6@sha256:b3fabdc7d4ecd0f396016ef78da19002c39e3ace352ea0ae4baa2ce9d5958376",
+		},
+		{
+			name:      "error when pulling an image that doesn't exist",
+			ref:       "ghcr.io/zarf-dev/zarf/imagethatdoesntexist:v1.1.1",
+			expectErr: true,
```
mkcp, line 101: 👍

mkcp (approved):
LGTM

PR #3312, opened 2024-12-10
feat: apply `--no-color` to dev and console logging option, remove `console-no-color`

### LINKED ISSUE #3306: Remove `console-no-color` option on `--log-format` and make `console` and `dev` respect the `no-color` flag
Having both an option `console-no-color` on the `--log-format` flag and a `--no-color` flag is confusing. Once the new logging is enabled by default, a user who uses the `--no-color` flag might be confused when they still see color from the terminal when either the `console` or `dev` formats are used. Additionally, a user who uses `--log-format=console-no-color` may be surprised when there is still color to terminal. 

Instead when `--no-color` is true Zarf should turn off color from the `console` and `dev` formats. This means the `console-no-color` option can be removed

### Alternatives
- Get rid of the `--no-color` flag. When `--log-format=console-no-color` is selected use `no-color` for any pterm output. This is not as explicit, and may lead to unexpected behavior. For example, `json` also has no color, so users may be suprised to see color while using the `json` logging format. 
- When `console-no-color` is used set `--no-color` to true by default. When `--no-color` is used set the default output to `console-no-color`. Again this may not be as clear of a UX. 

## Additional context
While most of the pterm calls are going away with the slog refactor, pterm is still called for yaml and table prints which is why color could still be output. For example, the yaml printed during `zarf package inspect` or the tables printed by `zarf dev lint` and `zarf package list` contain color.

AustinAbro321:
## Related Issue

Fixes #3306 

## Checklist before merging

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,4 +6,4 @@ USER 65532:65532
 
 COPY --chown=65532:65532 --chmod=0700 "build/zarf-linux-$TARGETARCH" /zarf
 
-CMD ["/zarf", "internal", "agent", "--log-level=debug", "--log-format=console-no-color", "--no-log-file"]
+CMD ["/zarf", "internal", "agent", "--log-level=debug", "--log-format=console", "--no-color"]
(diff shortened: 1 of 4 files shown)


mkcp (approved):
Send it

Issue #2773, opened 2024-07-26
Rethink Cobra code structure to remove use of global variables


phillebaba:
### Describe what should be investigated or refactored

The getting started guide in the Cobra documentations demos the library by using the `init()` function to register the command and configure flags. Which has resulted in a lot of projects following this pattern for their own code. The subsequent examples in the documentation does not use this pattern.

Error propagation becomes a large challenge when using functions `init()` as errors cannot be returned and they are magically called when the module is loaded. Additionally testing can become a hassle. This has led to problems like #2756 because there is no easy path to deal with these problems. As the functions are called automatically there is no easy way to pass function parameters, which instead led to using global variables to pass critical information. This includes things like command flags, embedded file systems, and build flags. Using such a pattern creates loose coupling between where the data is set and consumed.

My suggestion is that we refactor the CLI code base a bit to remove any use of the `init()` function. This would mean following a pattern similar to  what has been suggested in this issue.

https://github.com/spf13/cobra/issues/1041#issuecomment-1050938423

To make things clean we could have a single function for each subcommand/package which would then be responsible for calling and configuring commands downstream. This way we could pass any dependencies down from main to the subcommand which needs it.

### Links to any relevant code

https://github.com/zarf-dev/zarf/blob/81f1c7085ea954fd0d8e1610bc516d2b342965e8/main.go#L43-L44

https://github.com/zarf-dev/zarf/blob/81f1c7085ea954fd0d8e1610bc516d2b342965e8/src/cmd/initialize.go#L179-L228

### Additional context

N/A

mkcp:
I wrote up a similar issue #3183 that ended up being a duplicate. Copying the context over to here, cause there's code links and an example in `zarf say`

```
### Describe what should be investigated or refactored
Right now generate most of Zarf's CLI commands and flags inside of cmd pkg inits. While these inits() do have a reliable order, the order is not at all obvious when reading the code. We can make the code more maintainable by being explicit about when and how we generate commands and flags. Let's move this functionality into root and call the generate functions linearly, rather than in a bunch of init()s.

One pain point in this migration is that the logical commands in Zarf, e.g. `zarf package ...`, are each grouped by file names but are all in the `cmd` pkg. The many init functions help group and split these up, so we take that pain back on by moving them together. Taking additional time to move commands into their own subpackages, e.g. `cmd/dev` will help avoid overpopulating `cmd` with implementation details of specific commands.

### Links to any relevant code
[...]

PR #3322, opened 2024-12-12
refactor: Create structs for each command to  better isolate commands

soltysh:
## Description
This refactor is a first step to isolating all the command, which should allow easier consumption of zarf in other places. 

## Related Issue
Relates to #2773

## Checklist before merging

### DIFF
--- a/src/cmd/connect.go
+++ b/src/cmd/connect.go
@@ -16,100 +16,121 @@ import (
 	"github.com/zarf-dev/zarf/src/pkg/utils/exec"
 )
 
-var (
+// ConnectOptions holds the command-line options for 'connect' sub-command.
+type ConnectOptions struct {
 	cliOnly bool
 	zt      cluster.TunnelInfo
-)
-var connectCmd = &cobra.Command{
-	Use:     "connect { REGISTRY | GIT | connect-name }",
-	Aliases: []string{"c"},
-	Short:   lang.CmdConnectShort,
-	Long:    lang.CmdConnectLong,
-	RunE: func(cmd *cobra.Command, args []string) error {
-		ctx := cmd.Context()
-		l := logger.From(ctx)
-		target := ""
-		if len(args) > 0 {
-			target = args[0]
-		}
-
-		spinner := message.NewProgressSpinner(lang.CmdConnectPreparingTunnel, target)
-		defer spinner.Stop()
-
-		c, err := cluster.NewCluster()
-		if err != nil {
-			return err
-		}
-
-		var tunnel *cluster.Tunnel
-		if target == "" {
-			tunnel, err = c.ConnectTunnelInfo(ctx, zt)
-		} else {
-			var ti cluster.TunnelInfo
-			ti, err = c.NewTargetTunnelInfo(ctx, target)
-			if err != nil {
-				return fmt.Errorf("unable to create tunnel: %w", err)
-			}
-			if zt.LocalPort != 0 {
-				ti.LocalPort = zt.LocalPort
-			}
-			tunnel, err = c.ConnectTunnelInfo(ctx, ti)
-		}
-
-		if err != nil {
-			return fmt.Errorf("unable to connect to the service: %w", err)
-		}
-
-		defer tunnel.Close()
-
-		if cliOnly {
-			spinner.Updatef(lang.CmdConnectEstablishedCLI, tunnel.FullURL())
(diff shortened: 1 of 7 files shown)


src/cmd/dev.go
```diff
@@ -31,378 +31,465 @@ import (
...
+		Short:   lang.CmdDevPatchGitShort,
+		Args:    cobra.ExactArgs(2),
+		RunE:    o.Run,
+	}
 
-		if helpers.IsURL(fileName) {
-			message.Warn(lang.CmdDevSha256sumRemoteWarning)
-			logger.From(cmd.Context()).Warn("this is a remote source. If a published checksum is available you should use that rather than calculating it directly from the remote link")
+	// TODO(soltysh): get rid of pkgConfig global
+	cmd.Flags().StringVar(&pkgConfig.InitOpts.GitServer.PushUsername, "git-account", types.ZarfGitPushUser, lang.CmdDevFlagGitAccount)
```
AustinAbro321, line 191: Yup the globals are moving away in #2969

src/cmd/connect.go
```diff
@@ -16,100 +16,121 @@ import (
...
 }
 
-var connectListCmd = &cobra.Command{
-	Use:     "list",
-	Aliases: []string{"l"},
-	Short:   lang.CmdConnectListShort,
-	RunE: func(cmd *cobra.Command, _ []string) error {
-		c, err := cluster.NewCluster()
+// Run performs the execution of 'connect' sub command.
+func (o *ConnectOptions) Run(cmd *cobra.Command, args []string) error {
```
mkcp, line 51: Good choice to isolate these out into function declarations on the pkg

src/cmd/connect.go
```diff
@@ -16,100 +16,121 @@ import (
...
+	l := logger.From(ctx)
+	target := ""
+	if len(args) > 0 {
+		target = args[0]
+	}
+
+	spinner := message.NewProgressSpinner(lang.CmdConnectPreparingTunnel, target)
+	defer spinner.Stop()
+
+	c, err := cluster.NewCluster()
```
mkcp, line 62: Definitely not in scope for this PR, but sometime in the future it'd be great to ensure every CLI command can call a clearly defined API fn that encapsulates all of the configuration. Connect in particular has a lot of logic that could be abstracted out of the cli layer, which made me think of this.

src/cmd/connect.go
```diff
@@ -16,100 +16,121 @@ import (
...
+	case err = <-tunnel.ErrChan():
+		return fmt.Errorf("lost connection to the service: %w", err)
+	}
 }
 
-func init() {
-	rootCmd.AddCommand(connectCmd)
-	connectCmd.AddCommand(connectListCmd)
+// ConnectOptions holds the command-line options for 'connect list' sub-command.
+type ConnectListOptions struct{}
```
mkcp, line 110: +1 good pattern

src/cmd/root.go
```diff
@@ -126,6 +116,31 @@ func run(cmd *cobra.Command, _ []string) {
...
+		Run:               run,
+	}
+
+	// TODO(soltysh): consider adding command groups
+	rootCmd.AddCommand(NewConnectCommand())
+	rootCmd.AddCommand(NewDestroyCommand())
+	rootCmd.AddCommand(NewDevCommand())
+	rootCmd.AddCommand(NewInitCommand())
+	rootCmd.AddCommand(NewInternalCommand(rootCmd))
+	rootCmd.AddCommand(NewPackageCommand())
```
mkcp, line 139: This is the biggest win of the refactor imo -- we don't need to understand what magic implicit invocation order you get from a bunch of init()s being executed. Clear and declarative 👍 

One note though, `zarf say`'s command should probably be added to this sequence. It's currently in Execute below, but it's better off grouped with everything else.

mkcp (commented):
This is a good direction for cmd imo. Lots of room for future improvements as well.

src/cmd/root.go
soltysh: Yes!

The other part I actually started considering is grouping commands, have a look at `kubectl` output, it groups commands into several groups (basic, deploy, cluster management, etc...). We don't have that many, yet, but if we want to consider this construct will make it very easy to do just that.

mkcp: That makes a lot of sense, i'm in favor!

AustinAbro321 (approved):
LGTM

PR #3325, opened 2024-12-12
feat: print warnings before confirm with slogger

AustinAbro321:
## Description
I realized many warnings before confirm actions weren't being printed. We were printing some of the pre-confirm warnings separately which leads to cleaner code, however practically speaking they are almost impossible to notice for users because everything before the `p.confirmAction` happens quickly and `p.confirmAction` prints out a yaml file. 

## Checklist before merging


mkcp (approved):
LGTM

PR #3334, opened 2024-12-16
test: move lint e2e test to unit test

AustinAbro321:
## Description

lint e2e test should be moved to a unit test for simplicity and to avoid checking output from the cli

## Checklist before merging

### DIFF
--- a/src/pkg/lint/lint.go
+++ b/src/pkg/lint/lint.go
@@ -133,9 +133,6 @@ func templateZarfObj(zarfObj any, setVariables map[string]string) ([]PackageFind
 				})
 			}
 		}
-		// if unSetTemplates {
-
-		// }
 		for key, value := range setVariables {
 			templateMap[fmt.Sprintf("%s%s###", templatePrefix, key)] = value
 		}
--- a/src/pkg/lint/lint.go
+++ b/src/pkg/lint/lint.go
@@ -119,7 +119,6 @@ func templateZarfObj(zarfObj any, setVariables map[string]string) ([]PackageFind
 			return err
 		}
 
-		var unSetTemplates bool
 		for key := range yamlTemplates {
 			if deprecated {
 				findings = append(findings, PackageFinding{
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -130,14 +130,6 @@ init-package: ## Create the zarf init package (must `brew install coreutils` on
 release-init-package:
 	$(ZARF_BIN) package create -o build -a $(ARCH) --set AGENT_IMAGE_TAG=$(AGENT_IMAGE_TAG) --confirm .
 
-# INTERNAL: used to build an iron bank version of the init package with an ib version of the registry image
```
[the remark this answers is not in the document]
mkcp: 👍

src/pkg/lint/lint.go
```diff
@@ -128,15 +127,12 @@ func templateZarfObj(zarfObj any, setVariables map[string]string) ([]PackageFind
 				})
 			}
 			if _, present := setVariables[key]; !present {
-				unSetTemplates = true
+				findings = append(findings, PackageFinding{
+					Description: fmt.Sprintf("package template %s is not set and won't be evaluated during lint", key),
```
mkcp, line 131: 👍

src/pkg/lint/lint_test.go
```diff
@@ -93,3 +95,67 @@ func TestFillObjTemplate(t *testing.T) {
...
+			Description:         "Image not pinned with digest",
+			Item:                "image-in-good-flavor-component:unpinned",
+			PackageNameOverride: "lint",
+			PackagePathOverride: ".",
+			Severity:            SevWarn,
+		},
+	}
+	cwd, err := os.Getwd()
+	require.NoError(t, err)
+	// TODO @austinabro321: remove this and parallelize the test once changing the working directory is no longer required
```
mkcp, line 153: Makes sense keeping the scope of changes manageable.

src/pkg/lint/lint_test.go
```diff
@@ -93,3 +95,67 @@ func TestFillObjTemplate(t *testing.T) {
 	require.ElementsMatch(t, expectedFindings, findings)
 	require.Equal(t, expectedComponent, component)
 }
+
+func TestLintPackageWithImports(t *testing.T) {
```
mkcp, line 99: In an ideal world we would testing more cases which are smaller. But this makes sense as an iterative step in the right direction.

mkcp (approved):
Looks good, added a few comments but no changes requested.

This is definitely outside the scope of the PR, but deleting e2e test 12 highlights to me that these should be named in their order. Add or remove one and all of the other tests are affected. If you agree, would you be interested in scoping out an issue to migrate e2e tests to no longer be numbered?

PR #3356, opened 2024-12-20
Suggestion: Update SilenceErrors note

RothAndrew:
If `SilenceErrors:` is false, Cobra will emit unstructured log lines. Since Zarf wants structured logs whenever possible it makes sense for this to be set to true. The value being true signifies that we will be emitting our own logs using `slog` when we are returning an error.


AustinAbro321:
Thanks @RothAndrew, please follow the [DCO instructions](https://github.com/zarf-dev/zarf/pull/3356/checks?check_run_id=34686899901) to sign your commit then we can merge

RothAndrew:
@AustinAbro321 done

AustinAbro321 (approved):

Issue #3194, opened 2024-11-06
image pull fail from corrupted Crane cache


AustinAbro321:
### Describe what should be investigated or refactored
Seeing a flake in the [test-external](https://github.com/zarf-dev/zarf/blob/main/.github/workflows/test-external.yml) workflow. Images are failing to be saved.

[Workflow run](https://github.com/zarf-dev/zarf/actions/runs/11709398708/job/32613351462)
Relevant logs: 
```bash
  •  Fetching info for 9 images. This step may take several seconds to complete.
  •  Fetched info for 9 images
  •  Pulling 9 images (0.00 Byte of 243.74 MBs)

 WARNING  Failed to save images in parallel, falling back to sequential save: All attempts fail:
          #1: error writing layer: expected blob size 3419706, but only wrote 3207362
          #2: error writing layer: expected blob size 3419706, but only wrote 3207362
     ERROR:  failed to create package: All attempts fail:
             #1: error writing layer: expected blob size 3419706, but only wrote 3207362
             #2: error writing layer: expected blob size 3419706, but only wrote 3207362
    common.go:33: 
        	Error Trace:	/home/runner/work/zarf/zarf/src/test/external/common.go:33
        	            				/home/runner/work/zarf/zarf/src/test/external/ext_in_cluster_test.go:165
```

AustinAbro321:
I've validated that this is not caused by by disk space as the error in this case will look different
```bash
failed to create package: All attempts fail:
             #1: error writing layer: write
             /tmp/zarf-2081439845/images/blobs/sha256/000f791482e95f5e804ace91e5d39e0d48723c758a6adc740738cc1f9cd296153189335322:
             no space left on device
             #2: error writing layer: write
             /tmp/zarf-2081439845/images/blobs/sha256/000f791482e95f5e804ace91e5d39e0d48723c758a6adc740738cc1f9cd296152478328656:
             no space left on devic
```

AustinAbro321:
I run a script to build the podinfo-flux package (what the test flakes on) 100 times in two different terminals in parallel. I was not able to repeat this error. 

@RothAndrew reported that a similar error happening to him during his day to day with a separate private package. It does not happen to him in the images are not in the Zarf cache. This follows because in our usual e2e tests we delete the zarf cache right away for storage purpose, likely that is what's causing the flake to only appear in the test-external workflow

RothAndrew:
[...]

PR #3358, opened 2024-12-20
fix: properly delete invalid layers during image pull

AustinAbro321:
## Description

Zarf was not properly deleting invalid layers during image pull

## Related Issue

Relates to #3194 - pretty sure this will fix it, but I will give it some time to see if the flake disappears

## Checklist before merging

### DIFF
--- a/src/internal/packager/images/pull.go
+++ b/src/internal/packager/images/pull.go
@@ -13,6 +13,7 @@ import (
 	"maps"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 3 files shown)


src/internal/packager/images/pull.go
```diff
@@ -331,12 +330,7 @@ func CleanupInProgressLayers(ctx context.Context, img v1.Image) error {
 			if err != nil {
 				return err
 			}
-			absPath, err := config.GetAbsCachePath()
-			if err != nil {
-				return err
-			}
-			cacheDir := filepath.Join(absPath, layout.ImagesDir)
```
AustinAbro321, line 338: The main issue was that we were expecting the contents to be under an "images" folder which does not exist in the crane cache

Issue #3180, opened 2024-11-04
pkgr refactor: Refactor Packager Create 


schristoff:
#2969

PR #3381, opened 2025-01-07
Add support for import chains in packager2

phillebaba:
## Description

This change adds import chains to the packager2 which were never implemented during the refactor.

## Related Issue

Relates to #3180

## Checklist before merging


AustinAbro321 (approved):
Very elegant solution, nice job

PR #3397, opened 2025-01-13
feat: default to new logging format

AustinAbro321:
## Description

This defaults to the new logging format. 

Additionally, this changes `zarf init` to no longer print out credentials, instead a call to run `zarf tools get-creds` is given

## Checklist before merging


src/cmd/root.go
```diff
@@ -188,7 +188,7 @@ func init() {
 
 	// Logs
 	rootCmd.PersistentFlags().StringVarP(&LogLevelCLI, "log-level", "l", v.GetString(common.VLogLevel), lang.RootCmdFlagLogLevel)
-	rootCmd.PersistentFlags().StringVar(&LogFormat, "log-format", v.GetString(common.VLogFormat), "[beta] Select a logging format. Defaults to 'console'. Valid options are: 'console', 'json', 'dev'")
+	rootCmd.PersistentFlags().StringVar(&LogFormat, "log-format", v.GetString(common.VLogFormat), "[beta] Select a logging format. Defaults to 'console'. Valid options are: 'console', 'json', 'dev', 'legacy'. The legacy option will be removed in a coming release")
```
mkcp, line 191: Imo, let's keep backwards compatibility for 2 releases so no one is blocked on upgrading for just the logs.

src/pkg/logger/logger.go
```diff
@@ -99,6 +99,9 @@ var (
 	FormatConsole Format = "console"
 	// FormatDev uses a verbose and pretty printing devslog handler
 	FormatDev Format = "dev"
+	// FormatLegacy indicates that the user is continuing to use the legacy logger. The same as FormatNone.
+	// This option will be removed in a coming release
+	FormatLegacy Format = "legacy"
```
mkcp, line 104: 👍

mkcp (approved):
Exciting! Let's ship it

PR #3398, opened 2025-01-13
fix: avoid errors when `--max-package-size` differs between create runs

### LINKED ISSUE #3266: Rerunning package create with --max-package-size leaves old files around
### Environment
Device and OS: Framework 16 - Ubuntu 24.04
App version: v0.43.1
Kubernetes distro being used: K3d
Other:

### Steps to reproduce
1. For a Zarf package with a size >10MB run the following:

```
zarf package create . --max-package-size=10
zarf package create . --max-package-size=1
```

### Expected result

A split Zarf package is made with only 1MB files

### Actual Result

A split Zarf package is made where the first few are 10MB and the next few are 1MB

image

### Severity/Priority

Medium (the workaround is to delete all of the Zarf parts before recreate)

### Additional Context

None

AustinAbro321:
## Description

As seen explained in #3266 the existing split file logic doesn't replace files it writes over leading to errors with create is run with different `--max-package-size` values. Additionally, if the package is split and there are more files than Zarf is expecting (likely leftovers from a previous run) Zarf will give the below error 
```bash
     ERROR:  failed to deploy package: unable to load the package: package is missing parts, expected 15, found
             29
```

## Related Issue

Fixes #3266

## Checklist before merging


src/internal/packager2/layout/create_test.go
```diff
@@ -14,8 +17,115 @@ import (
...
+			expectedLastFileSize: 8,
+			expectedFileCount:    205,
+			expectedSha256Sum:    "fe8460f4d53d3578aa37191acf55b3db7bbcb706056f4b6b02a0c70f24b0d95a",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			dir := t.TempDir()
```
mkcp, line 58: Nonblocking: this is more of a style thing, but usually I add some small intent comment (example below) on a series of logically similar operations. There's tradeoffs here with just making self-documenting code, so I'm curious where your opinions are at there.

src/internal/packager2/layout/create_test.go
```diff
@@ -14,8 +17,115 @@ import (
...
+			require.NoError(t, err)
+			_, err = f.Write(b)
+			require.NoError(t, err)
+			err = f.Close()
+			require.NoError(t, err)
+
+			err = splitFile(context.Background(), p, tt.chunkSize)
+			require.NoError(t, err)
+
+			_, err = os.Stat(p)
```
mkcp, line 76: e.g.

`// Ensure content exists`

src/internal/packager2/layout/create_test.go
```diff
@@ -14,8 +17,115 @@ import (
...
+}
+
+func TestSplitDeleteExistingFiles(t *testing.T) {
+	t.Parallel()
+	tempDir := t.TempDir()
+	inputFilename := filepath.Join(tempDir, "testfile.txt")
+	data := make([]byte, 50)
+	err := os.WriteFile(inputFilename, data, 0644)
+	require.NoError(t, err)
+	// Create many fake split files
```
mkcp, line 112: 👍

mkcp (approved):
LGTM

src/internal/packager2/layout/create_test.go
AustinAbro321: I think it's a reasonable thing to do. I try to break up my tests with indentation and the arrange-act-assert methodology. I don't think comments are always necessary, but when code is non intuitive I do think it helps to provide additional context.

In this case, this test case was actually from packager1 so I just copied it over, going forward I'll try to air on the side of more clarity.

PR #3403, opened 2025-01-14
feat: error when building a package with zero components

### LINKED ISSUE #3307: Error when building package with zero components
When a Zarf package has flavors set on all of the components, it is possible to build a package that has no components. This occurs either when no flavor is defined or when the set flavor does not match any component. The current behavior is to create the package and report everything was successful.

N/A

### Additional context

I  do not see a use case where someone would want to create a package without a component. It would be better to tell the user with an error that no components match the filter.

AustinAbro321:
## Related Issue

Fixes #3307

## Checklist before merging


phillebaba:
The current tests do not cover a package with components that are filtered out due to flavor filtering, which a believe is what is described in the issue.

AustinAbro321:
@phillebaba Would you suggest that I move the logic to `resolveImports`, or add a test to `loadPackage` to test it out with flavors? I'm leaning towards the latter since `lint.ValidatePackage` is an exported function so having the logic there may come in handy for library or after create use cases.

phillebaba:
I dont have a problem with having the existing test, but I would probably not close the issue with this PR. Adding a test to loadPackage would probably be the best way to solve this.

AustinAbro321:
Updated with a test on loadPackage

phillebaba (approved):

mkcp (approved):
LGTM - nice and simple 👍

PR #3413, opened 2025-01-17
feat: dev inspect definition

AustinAbro321:
## Description

Adds a `zarf dev inspect definition` command which shows the package definition after templating, flavors, and component imports are applied. 

First part of implementing ZEP [0008-rework-inspect](https://github.com/zarf-dev/proposals/blob/main/0008-rework-inspect/README.md)

## Checklist before merging


src/test/e2e/00_use_cli_test.go
```diff
@@ -20,6 +20,18 @@ import (
 func TestUseCLI(t *testing.T) {
 	t.Log("E2E: Use CLI")
 
+	// TODO once cmd is refactored to accept an io.Writer, move this test to DevInspectDefinitionOptions.Run()
+	t.Run("zarf package inspect definition", func(t *testing.T) {
```
soltysh, line 24: Nit: `zarf dev inspect definition`

soltysh (approved):
One minor nit, feel free to merge when you get that fixed

src/cmd/dev.go
```diff
@@ -50,10 +52,61 @@ func NewDevCommand() *cobra.Command {
...
+}
+
+// Run performs the execution of 'dev inspect definition' sub-command.
+func (o *DevInspectDefinitionOptions) Run(cmd *cobra.Command, args []string) error {
+	ctx := cmd.Context()
+	pkg, err := layout2.LoadPackage(ctx, setBaseDirectory(args), o.Flavor, o.SetVariables)
+	if err != nil {
+		return err
+	}
+	pkg.Build = v1alpha1.ZarfBuildData{}
```
phillebaba, line 102: Is this required to be set to an empty struct?

src/internal/packager2/layout/create.go
```diff
@@ -260,7 +260,8 @@ func CreateSkeleton(ctx context.Context, packagePath string, opt CreateOptions)
 	return buildPath, nil
 }
 
-func loadPackage(ctx context.Context, packagePath, flavor string, setVariables map[string]string) (v1alpha1.ZarfPackage, error) {
+// LoadPackage returns a validated package definition after flavors, imports, and variables are applied.
+func LoadPackage(ctx context.Context, packagePath, flavor string, setVariables map[string]string) (v1alpha1.ZarfPackage, error) {
```
phillebaba, line 264: You should use the following function instead.

https://github.com/zarf-dev/zarf/blob/15216b5480661e1284ab1338d9897fe2403a0c6f/src/internal/packager2/load.go#L38

The API may need to be cleaned up in the future.

src/cmd/dev.go
AustinAbro321: Technically no, but most of the data is missing at this point. IMO since the package isn't built yet, it makes sense to not include the build data

src/cmd/dev.go
```diff
@@ -50,10 +52,61 @@ func NewDevCommand() *cobra.Command {
 	cmd.AddCommand(NewDevFindImagesCommand(v))
 	cmd.AddCommand(NewDevGenerateConfigCommand())
 	cmd.AddCommand(NewDevLintCommand(v))
+	cmd.AddCommand(NewDevInspectCommand(v))
 
 	return cmd
 }
 
+// NewDevInspectCommand creates the `dev inspect` sub-command.
+func NewDevInspectCommand(v *viper.Viper) *cobra.Command {
```
soltysh, line 61: Let's wait with exporting these for the community call later today, and based on the decision from there, make it private.

soltysh (approved):

src/cmd/dev.go
phillebaba: Well would the build data not be empty already here?

AustinAbro321: No migrations and I believe one other field get values in LoadPackage

phillebaba (dismissed):

phillebaba (approved):

Issue #3224, opened 2024-11-08
Add output formats to more commands than just zarf version


mkcp:
### Is your feature request related to a problem? Please describe.
With the addition of structured logging #2576 it's become clear that structured data has user value in outputs not just diagnostic logs. Currently most output, like in `zarf package list` is printed using pterm tables.

### Describe the behavior you'd like
In addition to tables, being able to output json, yaml, or potentially other formats would be hugely valuable for users to script and parse zarf's output. Some utilites like `gh` and `vault` provide filtering as well. This allows users to specify only the fields they want in their output, cutting out steps in external tools later. Filtering is more of a nice to have and maybe more appropriate for a second iteration.

@Racer159 Raised the detail that `zarf version` has an `--output / -o flag` already and supports rendering to json and yaml. That's a good starting point for us extending this behavior into other commands.

### Additional context
This shouldn't be too difficult to implement because, at least as far as I could tell writing this, all of the table printing is well contained to `cmd`. Swapping out the printer from message, or extracting it into a more configurable layer shouldn't require any significant architecture changes - just pass the data to a new renderer.

mkcp:
@rjferguson21 Appreciate the request, let us know what you think and if there's anything you'd like us to add.

mkcp:
Some prior work for ideas:

Vault -format with tables, json, yaml, and jsonx (scroll down to `-format`)]: https://developer.hashicorp.com/vault/docs/commands
k8s jsonpath https://kubernetes.io/docs/reference/kubectl/jsonpath/
GH -json formatting options: https://cli.github.com/manual/gh_help_formatting

mkcp:
Related #3226

AustinAbro321:
Since we already have a `--output` flag for specifying the location of a created artifact, for example `zarf package create --output`, and a `--log-format` flag, I think it would make sense to call this flag `--output-format`

AustinAbro321:
Closing this as we have the `--output-format` flag for `zarf package list` and `zarf tools get-creds` from #3415

PR #3415, opened 2025-01-21
feat: introduce output format flag for `zarf tools get-creds` and `zarf package list`

AustinAbro321:
## Description

Introduces the `--output-flag` for `zarf tools get-creds` and `zarf package list`. Additionally deprecates the `--output` flag on zarf version to be `--output-format`. There are other commands that print tables in the future such as `zarf connect list` and `zarf dev lint`. `get-creds` and `list` were chosen because the output of these commands is more likely to be used in automation.

## Related Issue

Relates to #3224

## Checklist before merging


src/pkg/message/credentials.go
```diff
@@ -22,61 +22,19 @@ const (
...
-func PrintCredentialTable(state *types.ZarfState, componentsToDeploy []types.DeployedComponent) {
-	if len(componentsToDeploy) == 0 {
-		componentsToDeploy = []types.DeployedComponent{{Name: "git-server"}}
-	}
-
-	// Pause the logfile's output to avoid credentials being printed to the log file
-	if logFile != nil {
-		logFile.Pause()
-		defer logFile.Resume()
-	}
```
AustinAbro321, line 35: this was not ported over, but logFile is only used with the legacy logger, and since this now goes to stdout it doesn't go to the logfile anyway

src/cmd/package.go
```diff
@@ -406,56 +409,103 @@ func (o *packageInspectOptions) run(cmd *cobra.Command, args []string) error {
...
+			return err
+		}
+		fmt.Fprintln(o.outputWriter, string(output))
+	case "yaml":
+		output, err := goyaml.Marshal(packageList)
+		if err != nil {
+			return err
+		}
+		fmt.Fprint(o.outputWriter, string(output))
+	case "table":
```
mkcp, line 498: Not a blocker, but the table rendering can be done in a reusable way with reflection. [One example](https://stackoverflow.com/questions/67608323/extract-field-names-of-a-struct-and-put-them-in-a-slice-of-strings)

src/cmd/version.go
```diff
@@ -34,14 +45,16 @@ func newVersionCommand() *cobra.Command {
 		RunE:    o.run,
 	}
 
-	cmd.Flags().StringVarP(&o.outputFormat, "output", "o", "", "Output format (yaml|json)")
+	cmd.Flags().VarP(&o.outputFormat, "output-format", "o", "Output format (yaml|json)")
+	cmd.Flags().VarP(&o.outputFormat, "output", "", "Output format (yaml|json)")
+	cmd.Flags().MarkDeprecated("output", "output is deprecated. Please use --output-format instead")
```
mkcp, line 50: 👍

mkcp (approved):
Small nit and notes for future improvements. will reapprove

mkcp (approved):
LGTM

soltysh (approved):
lgtm

PR #3416, opened 2025-01-21
feat: deprecate and replace `zarf package inspect` with child commands `zarf package inspect definition|sbom|images`

AustinAbro321:
## Description

feat: change `zarf package inspect` to a parent command, `zarf package inspect definition|sbom|images`

This also removes the ability to view SBOMs directly in browser with the `--sbom` flag

## Related Issue

[ZEP#9](https://github.com/zarf-dev/proposals/pull/9)

## Checklist before merging


src/cmd/package.go
```diff
@@ -376,47 +379,172 @@ func (o *PackageInspectOptions) PreRun(_ *cobra.Command, _ []string) {
...
+		imagesOpts := PackageInspectImagesOptions{
+			SkipSignatureValidation: pkgConfig.PkgOpts.SkipSignatureValidation,
+		}
+		return imagesOpts.Run(cmd, args)
+	}
+
+	definitionOpts := PackageInspectDefinitionOptions{
+		SkipSignatureValidation: pkgConfig.PkgOpts.SkipSignatureValidation,
+	}
+	return definitionOpts.Run(cmd, args)
```
soltysh, line 406: Good stuff :+1:

src/cmd/package.go
```diff
@@ -376,47 +379,172 @@ func (o *PackageInspectOptions) PreRun(_ *cobra.Command, _ []string) {
...
+func NewPackageInspectSBOMCommand(v *viper.Viper) *cobra.Command {
+	o := &PackageInspectSBOMOptions{}
+	cmd := &cobra.Command{
+		Use:   "sbom [ PACKAGE ]",
+		Short: "Output the package SBOM (Software Bill Of Materials) to the specified directory",
+		Args:  cobra.MaximumNArgs(1),
+		RunE:  o.Run,
+	}
+
+	cmd.Flags().BoolVar(&o.SkipSignatureValidation, "skip-signature-validation", false, lang.CmdPackageFlagSkipSignatureValidation)
```
soltysh, line 425: I haven't implemented that bit yet, but ideally you want to have `NewPackageInspectSBOMOptions` function which folks can invoke and get default options for a command. Then when you wire your flags here instead of providing defaults directly, you're just using the field from options struct. 

A nice example of that is in `kubectl api-resources` [command](https://github.com/kubernetes/kubernetes/blob/42c4ddbe2d8efe95481d10ccfc9ca098a71e83e9/staging/src/k8s.io/kubectl/pkg/cmd/apiresources/apiresources.go#L97-L122). Notice how `NewAPIResourceOptions` is passed with IO streams, and then each flag is automatically wired to the default value of a particular options. 

The main benefit of such an approach is that you have only a single source of truth wrt defaults.

src/cmd/package.go
```diff
@@ -376,47 +379,172 @@ func (o *PackageInspectOptions) PreRun(_ *cobra.Command, _ []string) {
...
+}
+
+// PackageInspectSBOMOptions holds the command-line options for 'package inspect sbom' sub-command.
+type PackageInspectSBOMOptions struct {
+	SkipSignatureValidation bool
+	SBOMOutputDir           string
+}
+
+// NewPackageInspectSBOMCommand creates the `package inspect sbom` sub-command.
+func NewPackageInspectSBOMCommand(v *viper.Viper) *cobra.Command {
```
soltysh, line 416: [Same comment](https://github.com/zarf-dev/zarf/pull/3413#discussion_r1925551341) as in #3413 about exporting these functions.

src/cmd/package.go
```diff
@@ -376,47 +379,172 @@ func (o *PackageInspectOptions) PreRun(_ *cobra.Command, _ []string) {
...
+// Run performs the execution of 'package inspect images' sub-command.
+func (o *PackageInspectImagesOptions) Run(cmd *cobra.Command, args []string) error {
+	ctx := cmd.Context()
+
+	src, err := choosePackage(ctx, args)
+	if err != nil {
+		return err
+	}
+
+	cluster, _ := cluster.NewCluster() //nolint:errcheck
```
soltysh, line 489: Whenever explicitly ignoring error it's good to explain why you're doing so. I'm guessing that since this is only showing the contents of the  file we don't care about the error.

src/cmd/package.go
AustinAbro321: I went ahead and implemented this

src/cmd/package.go
AustinAbro321: We actually don't know at this point yet if a cluster is needed, which is why we avoid the error. Added context in comments

src/cmd/package.go
soltysh: Great!

soltysh (dismissed):
lgtm

soltysh (approved):
lgtm

PR #3421, opened 2025-01-22
chore: unexport cmd functions besides NewZarfCommand

### LINKED ISSUE #2765: Allow for proper embedding of Zarf CLI

Currently Zarf exports only the `ExecuteC` receiver from its `rootCmd`. In order for Zarf to be properly embedded like the CLIs it embeds (`helm`, `crane`, `yq`, etc...), it needs to expose its `rootCmd`.

- **Given** another project wants to embed the Zarf CLI
- **When** that project imports `cmd.NewRootCmd`
- **Then** the project is able to embed Zarf

https://github.com/defenseunicorns/uds-cli/blob/dff77e30dd81363a3ff44cc10e8d1d5186ad4297/src/cmd/vendored.go#L70-L79 

is how it is currently done in other projects, but this can have strange behavior and it not the most cobra native way to embed CLIs, escpecially since both CLIs _could_ be natively embedded w/ `cmd.AddCommand`.

AustinAbro321:
## Description

This will unexport all commands besides `NewZarfCommand`. This keeps the surface area of breaking changes small, while giving users a while to directly embed Zarf with Cobra. 

This flattens the src/cmd package as well which enables unexporting functions, and makes the structure simpler all around. We no longer need a separate `common` package to avoid import cycles. 

## Related Issue

Fixes #2765 (Technically the issue could have already been closed, it's more intentional this time though)
 

## Checklist before merging


soltysh (approved):
lgtm

PR #3432, opened 2025-01-24
test: avoid flake in test external

AustinAbro321:
## Description

There is a flake we often get in test external #3194 relating to a broken Zarf cache. Likely, this will not get resolved until we either stop pulling packages in parallel (unlikely as it would make pulling signifigantly slower), or move away from crane (planned but will take time). Adding this in as a band aid to avoid flakes

## Checklist before merging


mkcp (approved):
LGTM

PR #3440, opened 2025-01-27
fix: display correct size with units during image pull

### LINKED ISSUE #3439: Image size message is misleading
### Environment
Device and OS: 0.47.0

The new logging format in 0.47.0 includes a message like below for each image. 
`saving image ref=nvcr.io/nvidia/cloud-native/gpu-operator-validator:v23.9.2 size=3249 method=concurrent`
This size is misleading as it is the size of the image manifests which is a json representation of the different parts of the image, whereas the user will be more interested in the total size of the image they are pulling. Additionally, we should add a unit such as kb / mb to the image size

AustinAbro321:
## Description

Fixes #3439 

## Checklist before merging

### DIFF
--- a/src/internal/packager/images/pull.go
+++ b/src/internal/packager/images/pull.go
@@ -228,11 +228,11 @@ func Pull(ctx context.Context, cfg PullConfig) (map[transform.Image]v1.Image, er
 				img = cache.Image(img, cache.NewFilesystemCache(cfg.CacheDirectory))
 			}
 
-			manifest, err := img.Manifest()
+			size, err := getSizeOfImage(img)
 			if err != nil {
-				return fmt.Errorf("unable to get manifest for %s: %w", refInfo.Reference, err)
+				return fmt.Errorf("failed to get size of image: %w", err)
 			}
-			totalBytes.Add(manifest.Config.Size)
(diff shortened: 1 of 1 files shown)


src/internal/packager/images/pull.go
```diff
@@ -397,6 +391,32 @@ func CleanupInProgressLayers(ctx context.Context, img v1.Image, cacheDirectory s
 	return eg.Wait()
 }
 
+func getSizeOfImage(img v1.Image) (int64, error) {
```
mkcp, line 394: tidy 👍

src/internal/packager/images/pull.go
```diff
@@ -421,7 +442,7 @@ func SaveSequential(ctx context.Context, cl clayout.Path, m map[transform.Image]
 		saved[info] = img
 		l.Debug("done saving image",
 			"ref", info.Reference,
-			"size", size,
+			"size-in-bytes", size,
```
mkcp, line 445: thoughts on using "bytes" or "imageBytes" for these tags?

mkcp (approved):
Nonblocking nit, LGTM

mkcp (approved):

PR #3449, opened 2025-01-29
fix: avoid false positives in import cycle detection

### LINKED ISSUE #3448: Package can't import a component from a different package that shares a dependency
### Environment
Zarf version: v0.47.0

### Steps to reproduce
1. setup the following files/folder structsure:
```.
├── child-component
│   └── zarf.yaml
├── jq
│   └── zarf.yaml
└── zarf.yaml
```

zarf.yaml:
```yaml
---
kind: ZarfPackageConfig
metadata:
  name: example-package

components:
  # JQ
  - name: jq-amd64
    required: true
    import:
      path: jq

  - name: jq-darwin
    required: true
    import:
      path: jq

  - name: child-jq-command
    required: true
    import:
      path: child-component

  - name: jq-command    
    actions:
      onDeploy:
        defaults:
          mute: true
        before:
          - cmd: |
              echo "I'm doing something cool with JQ too!"
              echo "{}" | ./bin/jq
```

./jq/zarf.yaml:
```yaml
---
kind: ZarfPackageConfig
metadata:
  name: jq
  description: Deploys the jq bin to the ./bin directory for use with other packages.

components:
  - name: jq-amd64
    only:
      localOS: linux
    required: true
    description: Deploy ./bin/jq bin for use in other components
    files:
      - source: https://github.com/stedolan/jq/releases/latest/download/jq-linux64
        target: ./bin/jq
        executable: true
  - name: jq-darwin
    only:
      localOS: darwin
    required: true
    description: Deploy ./bin/jq bin for use in other components
    files:
      - source: https://github.com/stedolan/jq/releases/latest/download/jq-osx-amd64
        target: ./bin/jq
        executable: true
```

./child-component/zarf.yaml
```yaml
---
kind: ZarfPackageConfig
metadata:
  name: child-component

components:
  # JQ
  - name: jq-amd64
    required: true
    import:
      path: ../jq

  - name: jq-darwin
    required: true
    import:
      path: ../jq

  - name: child-jq-command
    actions:
      onDeploy:
        defaults:
          mute: true
        before:
          - cmd: |
              echo "I'm doing something cool with JQ"
              echo "{}" | ./bin/jq
```

2.  run `zarf package create`

### Expected result
Package successfully creates as the package does not import the any component twice.

### Actual Result
Package fails to create with the following error:
```
ERR failed to create package: package jq imported in cycle by child-component in component jq-amd64
```

### Severity/Priority
High

I can't build packages anymore

### Additional Context

AustinAbro321:
## Description

Replaces the map cycle detection with a stack-based approach. The current approach caused false positives if multiple components import the same sub-package in a specific directory hierarchy because `filepath.Join` will clean the path causing it to look like the same import is occurring multiple times.  Another potential solution is to use `os.PathSeparator` instead of `filepath.Join` to avoid cleaned filepaths

This PR instead uses a stack and only checks the current chain of imports to avoid false positives.

## Related Issue

Fixes #3448 

## Checklist before merging


mkcp (approved):
LGTM

PR #3463, opened 2025-02-04
test: delete e2e test for examples/package-flavors

AustinAbro321:
## Description

This deletes the e2e test for package flavors. The other test in [10_component_flavor_test.go](https://github.com/zarf-dev/zarf/blob/33d8a2a2dc84b89913cc74d45157a04f6230c5ec/src/test/e2e/10_component_flavor_test.go#L37) already covers flavors and `layout2.LoadFromTar` more in depth. We have unit tests for verifying package builds with flavors as well. 

I think the package-flavors example does a good job of showing off the `--flavor` feature, but doesn't make for a good e2e test. Especially since it pulls images from dockerhub which can lead to flakes on TOOMANYREQUESTS. 

## Checklist before merging

### DIFF
--- a/src/test/e2e/10_component_flavor_test.go
+++ b/src/test/e2e/10_component_flavor_test.go
@@ -15,25 +15,6 @@ import (
 	layout2 "github.com/zarf-dev/zarf/src/internal/packager2/layout"
 )
 
-func TestFlavorExample(t *testing.T) {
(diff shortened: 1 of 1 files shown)


mkcp (approved):
Always good to have a lean test suite 👍

PR #3470, opened 2025-02-05
fix: avoid error when building package importing skeleton with remote components

### LINKED ISSUE #3469: Remote skeleton components fail when imported
### Environment
Device and OS: v0.47.0-v0.48.0
App version:
Kubernetes distro being used:
Other:

### Steps to reproduce
1. `zarf package create` on https://github.com/defenseunicorns/delivery-zarf-init/ (standard custom init package)

### Expected result

Build succeeds

### Actual Result
```bash
2025-02-05 09:11:28 ERR failed to create package: component zarf-injector not found
```

### Severity/Priority

High, custom init packages and any other skeleton relying on remote skeleton components fail

### Additional Context

By remote skeleton components I am referring to a skeleton component which doesn't contain any local files, local manifests, or local Helm charts. These components instead have resources that are pulled when the package is imported such as images, Helm charts in repos, or remote files

AustinAbro321:
## Description

During the create refactor the logic was changed to error when pulling a component from a skeleton package without an associated layer in the oci manifests. This leads to errors when a component is imported that only has remote resources. A tarball isn't created for that component in the skeleton package, and therefore a layer is not created for that component. This changes the logic to follow how it was originally in [importchain.fetchOCISkeleton](https://github.com/zarf-dev/zarf/blob/33d8a2a2dc84b89913cc74d45157a04f6230c5ec/src/pkg/packager/composer/oci.go#L48)

## Related Issue

Fixes #3469 

## Checklist before merging

### DIFF
--- a/src/internal/packager2/layout/import.go
+++ b/src/internal/packager2/layout/import.go
@@ -190,8 +190,8 @@ func fetchOCISkeleton(ctx context.Context, component v1alpha1.ZarfComponent, pac
 	}
 	componentDesc := manifest.Locate(filepath.Join(layout.ComponentsDir, fmt.Sprintf("%s.tar", name)))
 	var tarball, dir string
-	// If the descriptor was not found all resources in the component are remote
-	// We add an empty addition to the cache in this case
+	// If the descriptor for the component tarball was not found then all resources in the component are remote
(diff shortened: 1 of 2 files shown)


mkcp (approved):
LGTM

PR #3471, opened 2025-02-05
chore(deps): group golang dependencies in dependabot

AustinAbro321:
## Description

Golang often releases their changes together. Generally they work with one another and often update the same dependencies so this avoids the double review and merge conflict errors

## Checklist before merging


mkcp (approved):
LGTM

PR #3472, opened 2025-02-05
fix: pulling not respecting "uncompressed" setting in metadata

### LINKED ISSUE #3464: `.metadata.uncompressed` is not being respected when pulled from OCI repo
### Environment
Device and OS: Linux
App version: v0.48.0
Other: n/a

### Steps to reproduce
1. `zarf package pull oci://ghcr.io/a1994sc/zarf/k3s:1.31.5-k3s1`

### Expected result

- `zarf-package-k3s-amd64-1.31.5-k3s1.tar`

### Actual Result

- `zarf-package-k3s-amd64-1.31.5-k3s1.tar.zst`

### Visual Proof
- https://github.com/a1994sc/zarf-package-k3s

### Additional Context
See [`zarf.yaml`](https://github.com/a1994sc/zarf-package-k3s/blob/567813da82cc42b659f679038be1b244dc9958e7/zarf.yaml)

a1994sc:
## Description

Address `zarf package pull` not respecting `.metadata.uncompressed` package setting

## Related Issue

Fixes #3464

## Checklist before merging


src/internal/packager2/pull.go
```diff
@@ -84,6 +84,22 @@ func Pull(ctx context.Context, src, dir, shasum string, filter filters.Component
 	if err != nil && !errors.Is(err, os.ErrNotExist) {
 		return err
 	}
+	if strings.HasSuffix(name, ".tar") {
```
AustinAbro321, line 87: We want to avoid another unarchive and archive, as this can be quite costly depending on how big the package is, I've heard of 100+GB packages for example. Instead I believe this would be solved by changing the constant `.data.tar.zst` and the function `nameFromMetadata` depending on the value of .metadata.compressed

AustinAbro321 (changes requested):

src/internal/packager2/pull.go
a1994sc: I left some mental notes in the issue that I will bring over:

https://github.com/zarf-dev/zarf/issues/3464#issuecomment-2635484062

> [R]eviewing the code a bit, I think that it has something to do with:
> https://github.com/zarf-dev/zarf/blob/33d8a2a2dc84b89913cc74d45157a04f6230c5ec/src/internal/packager2/pull.go#L49
> though it does lead to a bit of a chicken and egg problem as we would need to determine the file extension by checking the `.metadata.uncompressed` but the code needs a local file to check...

So I guess is there a cleaner way to check the `zarf.yaml` of a remote artifact other then checking locally?

AustinAbro321: Ah gotcha, yeah packages published to https will already be compressed/uncompressed correctly, whereas oci packages are not compressed as the files comprising the packages are in oci layers. I would move the `nameFromMetadata` functions into `pullOCI` and `pullHTTP` and have these functions take in a directory rather than complete path then name / rename accordingly.

a1994sc: OK, I will give it a try over the weekend and see if that helps

a1994sc: hm, doing some thinking and should the updated function look something like this?

```golang
func pullOCI(ctx context.Context, src, tmpPath, shasum string, filter filters.ComponentFilterStrategy) (bool, string, error)
```

basically it still return the `isPartial` `bool` the `error` but adds a new `tarPath` `string`

src/internal/packager2/pull.go
```diff
@@ -113,48 +113,52 @@ func pullOCI(ctx context.Context, src, tarPath, shasum string, filter filters.Co
...
-		return false, err
+		return false, "", err
 	}
 	err = archiver.Archive(allTheLayers, tarPath)
 	if err != nil {
-		return false, err
+		return false, "", err
 	}
-	return isPartial, nil
+	return isPartial, tarPath, nil
```
a1994sc, line 161: @AustinAbro321 does this look like a the direction that you wanted to go?

src/internal/packager2/pull.go
a1994sc: I'm not sure it is the prettiest way to go about it, but I think that I added the logic needed to determine the file type with both `http(s)` and `oci` pulls

src/internal/packager2/pull.go
AustinAbro321: Ah actually, I think you just need to move this block outside of `supportsFiltering`. 
```go
	pkg, err := remote.FetchZarfYAML(ctx)
	if err != nil {
		return false, "", err
	}
	if !pkg.Metadata.Uncompressed {
		tarPath = fmt.Sprintf("%s.zst", tarPath)
	}
```

a1994sc: OK, I should be able to have that update in the next little bit.

a1994sc: Updated the PR let me know of that works for you, thank you.

AustinAbro321 (dismissed):
LGTM, nice work. Going to ask another member of my team to give it a look before merge

a1994sc:
I will see if I can track down that windows test failing....

src/internal/packager2/pull.go
```diff
@@ -113,87 +114,121 @@ func pullOCI(ctx context.Context, src, tarPath, shasum string, filter filters.Co
...
-		return err
+		return "", err
 	}
 	if received != shasum {
-		return fmt.Errorf("shasum mismatch for file %s, expected %s but got %s", tarPath, shasum, received)
+		return "", fmt.Errorf("shasum mismatch for file %s, expected %s but got %s", tarPath, shasum, received)
 	}
-	return nil
+
+	mtype, err := mimetype.DetectFile(tarPath)
```
AustinAbro321, line 210: Windows doesn't allow renaming of an already opened file. Rather than `defer f.close` you will have to run `f.close` anywhere after the `io.Copy`

a1994sc: gotcha... I will have an update to the function before this afternoon for the team to review

a1994sc: I moved the `f.Close()` to after 

```go
_, err = io.Copy(f, resp.Body)
```
and added `f.Close()` on error returns. Let me know if that helps

a1994sc: Gotcha, I will see if I can encapsulate that logic

a1994sc: Moved that logic to its own function, I believe that should address the windows issues.... _fingures crossed_

AustinAbro321 (dismissed):
LGTM, nice work

a1994sc:
Thank you!

src/internal/packager2/pull.go
```diff
@@ -113,54 +114,104 @@ func pullOCI(ctx context.Context, src, tarPath, shasum string, filter filters.Co
...
+	return isPartial, tarPath, nil
 }
 
-func pullHTTP(ctx context.Context, src, tarPath, shasum string) error {
+func pullHTTP(ctx context.Context, src, tarDir, shasum string) (string, error) {
 	if shasum == "" {
-		return errors.New("shasum cannot be empty")
+		return "", errors.New("shasum cannot be empty")
+	}
+	tmpDir, err := utils.MakeTempDir(config.CommonOptions.TempDirectory)
```
brandtkeller, line 169: This `tmpDir` logic looks unused and redundant given that `tarDir` is a derivative of the same logic - can we remove this?

a1994sc: both `pullOCI` and `pullHTTP` destroy either temporary directories after they are done with their respective

```golang
defer os.Remove(tmpDir)
```

so we need to "copy" them to the `tmpDir` location so we don't try to reference a destroyed file.

AustinAbro321: I see the need for it in `pullOCI` since we need a place to store the individual layers before we archive them, but in `pullHTTP` we pull the binary down as is. It looks like only place `tmpDir` is used in `pullHTTP` is in `os.Remove`

a1994sc: Oh, I see... I'm very sorry, I thought that @brandtkeller was talking about it in the context of the greater `Pull` function.... Yes, I can see the unused and redundant `tmpDir`. Testing that it wasn't magically being used and will update the PR soon

a1994sc: Updated and the code works without that, thank you for the catch.

brandtkeller (approved):

AustinAbro321 (approved):

PR #3479, opened 2025-02-07
fix: injector build process

AustinAbro321:
## Description

Replacing the injector build process which currently doesn't work at all with cross. This enables building binaries for arm64 and amd64 systems using docker, keeping the process simple. However , at least AFAIK, with cross is not compatible with arm64 mac

## Checklist before merging


mkcp (dismissed):
LGTM

mkcp (approved):

PR #3482, opened 2025-02-07
feat: improve skeleton package messaging

AustinAbro321:
## Description

We've heard complaints about skeleton packages being confusing, I expect them to be deprecated and their use case along custom init packages to be replaced some day, but for now I wanted to make it more clear to users that they are publishing a skeleton package and what that means. 

## Checklist before merging


mkcp (approved):
LGTM

PR #3487, opened 2025-02-10
3480 zarf connect default

### LINKED ISSUE #3480: zarf connect external dependency - xdg-open not found in $PATH
`zarf connect` on a headless system attempts to execute `xdp-open` on the system as an external dependency to the Zarf CLI and subsequently fails. The `--cli-only` flag circumvents this problem by attempting to open the default system browser. I believe the default behaviors should be switched.

See [exec.LaunchURL()](https://github.com/zarf-dev/zarf/blob/main/src/pkg/utils/exec/exec.go#L153)

- **Given** zarf defaults to encapsulated execution 
- **When** zarf connect is executed
- **Then** it defaults to the `--cli-only` logic and browser automation is exposed with a flag separately 

Remove support for launching the browser automatically and only support the headless output with associated tunnel link.

### Additional context
I believe Zarf should always default to execution without requirements on external/system dependencies. This could mean flipping from defaulting to opening a browser to defaulting to the `--cli-only` behavior and expecting

brandtkeller:
## Description

Propose a default `zarf connect` workflow that is not reliant upon an existing system dependency - instead opting for a more headless-friendly approach. 

Considering documenting this as an element of design for functionality within zarf - IE default workflows (or all workflows) being fully encapsulated logic that do not have external-to-the-binary dependencies. 

## Related Issue

Fixes #3480

## Checklist before merging


AustinAbro321 (approved):
LGTM welcome to the club!

PR #3493, opened 2025-02-12
chore(release): install cross before injector build

AustinAbro321:
## Description

We reworked the injector process to use cross, but didn't install cross before building 

## Checklist before merging


AustinAbro321:
Build was successful - https://github.com/zarf-dev/zarf/actions/runs/13287839378/job/37100764939 (job failed because only main is allowed to publish to aws)

PR #3494, opened 2025-02-12
test: remove dependency on `make build-examples` in e2e tests

AustinAbro321:
## Description

Currently e2e tests rely on certain examples being built before the test suite is run. This change makes e2e test build an example before they use it, decreasing storage requirements and making individual tests more encapsulated and easier to run locally.

Looking at the action performance metrics, it seems like e2e with cluster tests are generally taking about the same amount of time +- 1 minute. e2e without cluster tests seem to be faster by about 2 minutes

## Checklist before merging


src/test/external/ext_out_cluster_test.go
```diff
@@ -143,17 +147,14 @@ func (suite *ExtOutClusterTestSuite) Test_1_Deploy() {
 func (suite *ExtOutClusterTestSuite) Test_2_DeployGitOps() {
 	// Deploy the flux example package
 	temp := suite.T().TempDir()
-	defer func() {
-		suite.NoError(os.RemoveAll(temp), "unable to remove temporary directory")
```
AustinAbro321, line 147: suit.T().TempDir() always gets cleaned up by go so this isn't needed

mkcp (approved):
Nice job! Better test isolation, on-demand resource builds, and it's faster.

PR #3517, opened 2025-02-19
feat: deprecate config file types other than yaml and toml

### LINKED ISSUE #3516: Deprecate configuration file types other than yaml and toml
We currently allow the default viper extensions - `{"json", "toml", "yaml", "yml", "properties", "props", "prop", "hcl", "tfvars", "dotenv", "env", "ini"}`. Many of these are inconvenient to use in the context of Zarf and supporting all of them adds complexity to a user by introducing unnecessary choices. 

Additionally, Zarf may add custom parsing logic to config files to enable case sensitive map keys, which Viper does not support by default, see - https://github.com/zarf-dev/proposals/pull/16. 

The argument could be made to eliminate all but one choice. However, toml is useful for removing the layers of tabs that exist in yaml. Especially since configuration options tend to have three+ layers such as `package.deploy.set`. Yaml is useful for specifying maps and keeping parity with Helm charts. Many times the same a user will want to specify a value from a Helm values file, and keeping the option to copy and paste is convenient.

AustinAbro321:
## Related Issue

Fixes #3516

## Checklist before merging


mkcp (approved):
LGTM

PR #3540, opened 2025-02-24
fix: do not create SBOM for packages without SBOMable material

### LINKED ISSUE #3524: Zarf always populates sbom.tar artifact even if no SBOMs exist
### Environment
Device and OS: `darwin arm64`
App version: `0.48.1`
Other:

After a recent change in Zarf, it seems like an `sbom.tar` artifact is being created and populated for every package, even if no SBOM exists. This tar file contains a single file `compare.html`

I believe this was introduced [in this commit](https://github.com/zarf-dev/zarf/blob/3eab9287f90cd63cfa32a902b1033e1d66207b6b/src/internal/packager2/layout/sbom.go#L97-L100). The in-line comment above the call to `createSBOMCompareAsset()` says to only add the compare tool if there are SBOMs but that doesn't seem to be something we check before calling the function [or within the function](https://github.com/zarf-dev/zarf/blob/3eab9287f90cd63cfa32a902b1033e1d66207b6b/src/internal/packager2/layout/sbom.go#L265-L267)

When we noticed this in our project, we wrote [this small hotfix](https://github.com/defenseunicorns/uds-cli/pull/1066/commits/388ae9ac973e19d3a797c6b75f89aaa3a1ec1678) to check if there was SBOM content we needed to care about.

### Steps to reproduce
1. Build the following Zarf Package
```
kind: ZarfPackageConfig
metadata:
  name: sbom-gen

components:
  - name: silly-goofy-mood
    actions:
      onCreate:
        before:
          - cmd: echo "Hello World!"
```

NOTE: Even though I am using different `inspect` commands. This is not a side effect of updating the inspect command.

### Expected result
No SBOM generated as of `Zarf 0.46.0`
NOTE: The `sbom.tar` file exists within the package, but it is empty.

Image

### Actual Result

SBOM output exists as of `Zarf v0.48.1` (But I believe it was introduced in earlier)

Image

AustinAbro321:
## Description

do not create SBOM for packages without SBOMable material, looks like a bug from the create refactor

## Related Issue

Fixes #3524

## Checklist before merging


src/internal/packager2/remove.go
```diff
@@ -167,5 +167,6 @@ func Remove(ctx context.Context, opt RemoveOptions) error {
 		}
 	}
 
+	l.Info("package successfully removed", "name", pkg.Metadata.Name)
```
AustinAbro321, line 170: Adding this in here as an aside. Someone gave me feedback that `zarf package remove` feels strange since it doesn't give any confirmation that the remove happened successfully, I would agree

src/internal/packager2/layout/create_test.go
```diff
@@ -227,6 +227,43 @@ func TestLoadPackageErrorWithoutCompatibleFlavor(t *testing.T) {
...
+func writePackageToDisk(t *testing.T, pkg v1alpha1.ZarfPackage, dir string) {
+	t.Helper()
+	b, err := goyaml.Marshal(pkg)
+	require.NoError(t, err)
+	path := filepath.Join(dir, ZarfYAML)
+	err = os.WriteFile(path, b, 0700)
+	require.NoError(t, err)
+}
+
+func TestGetSBOM(t *testing.T) {
```
mkcp, line 239: 👍

src/internal/packager2/create.go
```diff
@@ -64,6 +67,11 @@ func Create(ctx context.Context, packagePath string, opt CreateOptions) error {
 
 	if opt.SBOMOut != "" {
```
brandtkeller, line 68: I believe I traced the logic between the earlier `CreatePackage()` and this point - and am curious if there is any value in augmenting this conditional with `&& !opt.SkipSBOM` ?

Sounds like a niche edge-case. But potentially conflicting arguments (from either CLI or SDK use) could potentially result in the SBOM being skipped but `SBOMOut` not being empty and therefor this error printing? 

No hard opinions on this from me personally - but stood-out as a likely scenario for this error to be hit (especially if the SDK use had a static path mentioned etc).

src/internal/packager2/layout/package.go
```diff
@@ -121,8 +121,14 @@ func (p *PackageLayout) Cleanup() error {
...
 }
 
+// ErrNoSBOMAvailable is returned when a user tries to access a package SBOM, but it is not available
+var ErrNoSBOMAvailable = errors.New("zarf package does not have an SBOM available")
+
 // GetSBOM outputs the SBOM data from the package to the give destination path.
 func (p *PackageLayout) GetSBOM(destPath string) (string, error) {
+	if !p.Pkg.IsSBOMAble() {
+		return "", ErrNoSBOMAvailable
+	}
```
mkcp, line 131: Let's declare `ErrNoSBOMAvailable` as a custom error type rather than a named error string.  The message could pass back the package name or sha as metadata e.g. :

```
...
(e *NoSBOMAvailableError) Error() string {
    return fmt.Sprintf("zarf package does not have an SBOM available, package=%s", e.package)
}
```
https://gobyexample.com/custom-errors

mkcp (changes requested):
ErrNoSBOMAvailable has a chance to return some debugging metadata from PackageLayout.GetSBOM. Recommend declaring it as a custom error type. Everything else looks good - test case especially.

src/internal/packager2/create.go
AustinAbro321: This is a great point, it brings into question where the validation should take place, but I think if a user had both the `--skip-sbom` flag and the `sbom-out` flags set they should error. I'm not sure if I should add that validation to the CLI level, the function level, or both. Leaning towards add it to the `Create` function

src/internal/packager2/layout/package.go
AustinAbro321: I like this, will implement now

mkcp (approved):
LGTM

PR #3546, opened 2025-02-26
Remove support for legacy logging

### LINKED ISSUE #3500: Complete deprecation of Message logging
Currently, we support the old logging behavior with message under `--log-format='legacy'`. We intend to fully remove this behavior once it is no longer necessary to maintain.

Related #2717

### Additional context
TODO

mkcp:
## Description
Within, we've removed `log-format="legacy"` from the  `cmd` layer while keeping `message` set up properly to print non-log messages like Yaml and tables. Next, we've chase down every `message` log statement, spinner, and progress bar in the main zarf codebase. 

## Related Issue

Fixes #3500 
Relates to #2576

## Checklist before merging

### DIFF
--- a/src/cmd/dev.go
+++ b/src/cmd/dev.go
@@ -234,7 +234,8 @@ func newDevPatchGitCommand() *cobra.Command {
 	return cmd
 }
 
-func (o *devPatchGitOptions) run(_ *cobra.Command, args []string) error {
+func (o *devPatchGitOptions) run(cmd *cobra.Command, args []string) error {
+	l := logger.From(cmd.Context())
 	host, fileName := args[0], args[1]
 
 	// Read the contents of the given file
@@ -249,7 +250,6 @@ func (o *devPatchGitOptions) run(_ *cobra.Command, args []string) error {
 	// Perform git url transformation via regex
 	text := string(content)
 
-	l := logger.Default()
 	processedText := transform.MutateGitURLsInText(l.Warn, gitServer.Address, text, gitServer.PushUsername)
 
 	// Print the differences
--- a/src/cmd/dev.go
+++ b/src/cmd/dev.go
@@ -25,7 +25,6 @@ import (
 	layout2 "github.com/zarf-dev/zarf/src/internal/packager2/layout"
 	"github.com/zarf-dev/zarf/src/pkg/lint"
 	"github.com/zarf-dev/zarf/src/pkg/logger"
-	"github.com/zarf-dev/zarf/src/pkg/message"
 	"github.com/zarf-dev/zarf/src/pkg/packager"
 	"github.com/zarf-dev/zarf/src/pkg/transform"
 	"github.com/zarf-dev/zarf/src/pkg/utils"
@@ -250,8 +249,9 @@ func (o *devPatchGitOptions) run(_ *cobra.Command, args []string) error {
 	// Perform git url transformation via regex
 	text := string(content)
 
-	// TODO(mkcp): Currently uses message for its log fn. Migrate to ctx and slog
-	processedText := transform.MutateGitURLsInText(message.Warnf, gitServer.Address, text, gitServer.PushUsername)
+	// FIXME(mkcp): Currently uses message for its log fn. Migrate to ctx and slog
+	l := logger.Default()
+	processedText := transform.MutateGitURLsInText(l.Warn, gitServer.Address, text, gitServer.PushUsername)
 
 	// Print the differences
 	// TODO(mkcp): Uses pterm to print text diffs. Decouple from pterm after we release logger.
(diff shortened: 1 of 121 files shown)


src/cmd/version.go
```diff
@@ -18,24 +18,33 @@ import (
...
 )
 
 type versionOptions struct {
 	outputFormat outputFormat
 	outputWriter io.Writer
 }
 
-func newVersionOptions() *versionOptions {
+func newVersionOptions(writer ...io.Writer) (*versionOptions, error) {
+	if len(writer) > 1 {
```
AustinAbro321, line 29: nit: IMO it would read cleaner if we accepted a single writer and then checked if it was nil

src/pkg/packager/interactive.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/zarf-dev/zarf/src/pkg/utils"
 )
 
+// TODO(mkcp): Implementation is heavily dependent on Message
```
AustinAbro321, line 23: I don't see anything in here we can't remove, everything only shows up when the `--legacy` flag is used

src/pkg/zoci/common.go
```diff
@@ -51,6 +46,12 @@ func NewRemote(ctx context.Context, url string, platform ocispec.Platform, mods
 	return &Remote{remote}, nil
 }
 
+// String provides a string representation of the Remote's reference.
+// REVIEW(mkcp): Does this function make sense? I use it for info logging in zoci/copier.go
+func (r *Remote) String() string {
+	return r.Repo().Reference.String()
```
AustinAbro321, line 52: I don't feel super strongly, but I'd prefer removing this method and keeping the calls here explicit. `OrasRemote` also contains the manifest and platform so just calling the string method it wouldn't be clear what it will output without looking at the code.

src/cmd/root.go
AustinAbro321, line 1: There's one more todo / call to remove to message here in the `Execute` function

AustinAbro321 (changes requested):
A few comments, but this is pretty close

src/cmd/root.go
```diff
@@ -209,28 +174,27 @@ func init() {
...
 	}
 
-	v := getViper()
+	vpr := getViper()
 
 	// Logs
-	rootCmd.PersistentFlags().StringVarP(&LogLevelCLI, "log-level", "l", v.GetString(VLogLevel), lang.RootCmdFlagLogLevel)
-	rootCmd.PersistentFlags().StringVar(&LogFormat, "log-format", v.GetString(VLogFormat), "[beta] Select a logging format. Defaults to 'console'. Valid options are: 'console', 'json', 'dev', 'legacy'. The legacy option will be removed in a coming release")
-	rootCmd.PersistentFlags().BoolVar(&SkipLogFile, "no-log-file", v.GetBool(VNoLogFile), lang.RootCmdFlagSkipLogFile)
-	rootCmd.PersistentFlags().BoolVar(&message.NoProgress, "no-progress", v.GetBool(VNoProgress), lang.RootCmdFlagNoProgress)
```
AustinAbro321, line 218: Would it make sense to keep this flag around for a bit and mark it as deprecated? Users probably have this flag in their pipeline, so their commands will fail when it's gone. I'm okay with it either way, but wanted to call this out.

src/pkg/zoci/common.go
mkcp: Yeah that's fair, this is a bit of a leaky abstraction to try to save a few nested object calls. Deleted in latest.

src/cmd/version.go
mkcp: Fair, I hear what you're saying. I think I agree and I'll add that my choice to use a variadic function here is overly clever. My intent was adding an optional writer arg without breaking callers' expectations of a default writer when calling `newVersionOptions()` with empty args, but that's not actually an issue for how `newVersionOptions` is used.

Looking into it more, there is bug here until we check for nil like you're saying. Simplified the fn with a nil->default in latest revision.

src/internal/packager/helm/repo.go
```diff
@@ -239,7 +222,8 @@ func (h *Helm) DownloadPublishedChart(ctx context.Context, cosignKeyPath string)
 
 	// Set up the chart chartDownloader
 	chartDownloader := downloader.ChartDownloader{
-		Out:            spinner,
+		// TODO(mkcp): Shouldn't rely on a global mutable var. Pass in a writer here somehow, or at least make atomic?
+		Out:            message.OutputWriter,
```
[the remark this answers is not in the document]
mkcp: That's a sensible way to handle it -- appreciate the context. Resolved in latest

src/cmd/dev.go
```diff
@@ -250,11 +249,10 @@ func (o *devPatchGitOptions) run(_ *cobra.Command, args []string) error {
 	// Perform git url transformation via regex
 	text := string(content)
 
-	// TODO(mkcp): Currently uses message for its log fn. Migrate to ctx and slog
-	processedText := transform.MutateGitURLsInText(message.Warnf, gitServer.Address, text, gitServer.PushUsername)
+	l := logger.Default()
```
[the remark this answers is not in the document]
mkcp: Good catch, resolved in latest 👍

src/cmd/root.go
mkcp: Good catch, paragraph log on program error removed in latest.

src/pkg/packager/interactive.go
mkcp: 🫡 sendin it

AustinAbro321 (approved):
LGTM

Issue #1684, opened 2023-05-08
Create a top-level security best practices page`


Racer159:
### Describe what should be investigated or refactored
We have a few security recommendations within our documentation but @Noxsios mentioned that it would be nice to have a top-level page that aggregates/links this information so that people can more easily review it.

### Additional context
Doing this would enable more people to be more secure when using Zarf.

mkcp:
We have a best practices section now on the website (#2335), still need to add a section to prune the cache.

PR #3551, opened 2025-03-03
2335 docs best practices

### LINKED ISSUE #2335: Document workflow for deploying a newer version of the init package onto an existing zarf clusters
Currently we don't have much on the docs site explaining how zarf clusters in should be updated in the day two scenario. I believe, it should just be redeploying the init package, then the new zarf package, but we should make sure that info is easy to find. We should also talk about rolling back the deployment of a package to an earlier package. Not sure if there are any pitfalls there or other common day two "traps" one might fall into. 

The best practices page proposed in #1684 could be a decent spot, otherwise we could have a page dedicated to day two advice

brandtkeller:
## Description

Need for documentation of the `zarf init` day 2 operations. Following helm with the use of a `Best Practices` section of the docs that we can expand for other opportunities. 

## Related Issue

Fixes #2335

Relates to #1684

## Checklist before merging


site/src/content/docs/best-practices/upgrading-zarf.mdx
```diff
@@ -0,0 +1,61 @@
+---
+title: "Upgrading Zarf"
+---
+
+Optimally the version of Zarf performing the `package create` and `package deploy` operations aligns to the version of [Init Package]() that is installed within the cluster.
```
AustinAbro321, line 5: Missing the [Init Package] link here

site/src/content/docs/best-practices/upgrading-zarf.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+## 2. Re-initialize the cluster
+
+- **Init:** The init process can be executed to upgrade a cluster from one version to another.
+
+  ```bash
+  zarf init --confirm
+  ```
+
+- **Deploy:** The Init Package can also be deployed directly with `package deploy` - allowing for explicit deployment specification.
```
AustinAbro321, line 37: I think it's worth mentioning here that custom init packages can be deployed this way

site/src/content/docs/best-practices/upgrading-zarf.mdx
```diff
@@ -0,0 +1,61 @@
...
+  zarf package deploy <zarf-init-package tarball> --confirm
+  ```
+
+  **Note:** The use of package deploy for zarf init operations will not expose any configuration for external git/registry options.
+
+## 3. Validate Zarf Infrastructure
+
+- **Validate:** Ensure the following zarf resources are healthy:
+  - Registry if deployed
+  - Agent pods
```
AustinAbro321, line 49:
```suggestion
- **Validate:** Ensure the following zarf resources are healthy:
  - Registry
  - Agent pods
  - Git server if deployed
```
In the default init package the Git Server is an optional component, but the rest of the services are required.

AustinAbro321 (changes requested):
Few small comments, adding this will be a plus.

site/src/content/docs/best-practices/upgrading-zarf.mdx
brandtkeller: agreed!

site/src/content/docs/best-practices/upgrading-zarf.mdx
brandtkeller: The registry wouldn't be deployed is an external registry is used correct?

AustinAbro321: Ah yeah totally forgot about that

site/src/content/docs/best-practices/upgrading-zarf.mdx
```diff
@@ -0,0 +1,62 @@
...
+  - Git server (if deployed)
+
+  ```bash
+  zarf tools kubectl get pods -n zarf
+  ```
+
+# Special Considerations
+
+After the original `zarf init` has taken place - any following initialize actions will use the same `secret` for obtaining information provided by any `zarf init` flags.
```
AustinAbro321, line 58:
```suggestion
After the original `zarf init` has taken place - any following initialize actions will use the values in the `zarf-state` secret in the `zarf` namespace to obtain information for Zarf infrastructure such as the registry. For example, the `--registry-pull-password` flag is ignored on subsequent runs.
```
Thoughts on adding some more detail here? I think specifying that this is for flags related infrastructure and not flags like `--components` would make this more clear.

AustinAbro321 (approved):
LGTM

PR #3556, opened 2025-03-04
feat: recording deployed component status in package deploy

### LINKED ISSUE #3553: Report component status for each deployed component
This [PR](https://github.com/zarf-dev/zarf/pull/3066/files#diff-0e9c236b38b2c2443c4b5b092c9fc667aee2cc1fbb88ee84a980f65c3b0457f0) removed the Status and ObservedGeneration from the DeployedComponent struct. I was using this to provide incremental deployment status for a package based on what components had been deployed. Could that be added back into the DeployedComponent struct?

#### Deploying
- **Given** a package is deploying 
- **When** a package component starts
- **Then** the DeployedComponent is updated with a ComponentStatusDeploying

#### Succeeded
- **Given** a package is deploying 
- **When** a package component succeeds
- **Then** the DeployedComponent is updated with a ComponentStatusSucceeded

#### Failed
- **Given** a package is deploying 
- **When** a package component fails
- **Then** the DeployedComponent is updated with a ComponentStatusFailed

sgettys:
## Description
Adds back in deployed component status when deploying a package. 
...

## Related Issue
Fixes #3553 

## Checklist before merging

### DIFF
--- a/src/pkg/packager/deploy.go
+++ b/src/pkg/packager/deploy.go
@@ -151,6 +151,7 @@ func (p *Packager) deployComponents(ctx context.Context) ([]types.DeployedCompon
 
 	// Process all the components we are deploying
 	for _, component := range p.cfg.Pkg.Components {
+		packageGeneration := 1
 		// Connect to cluster if a component requires it.
 		if component.RequiresCluster() {
 			timeout := cluster.DefaultTimeout
@@ -162,10 +163,16 @@ func (p *Packager) deployComponents(ctx context.Context) ([]types.DeployedCompon
 			if err := p.connectToCluster(connectCtx); err != nil {
 				return nil, fmt.Errorf("unable to connect to the Kubernetes cluster: %w", err)
 			}
+			// If this package has been deployed before, increment the package generation within the secret
+			if existingDeployedPackage, _ := p.cluster.GetDeployedPackage(ctx, p.cfg.Pkg.Metadata.Name); existingDeployedPackage != nil {
+				packageGeneration = existingDeployedPackage.Generation + 1
+			}
 		}
 
 		deployedComponent := types.DeployedComponent{
-			Name: component.Name,
+			Name:               component.Name,
+			Status:             types.ComponentStatusDeploying,
+			ObservedGeneration: packageGeneration,
 		}
 
 		// Ensure we don't overwrite any installedCharts data when updating the package secret
@@ -201,7 +208,7 @@ func (p *Packager) deployComponents(ctx context.Context) ([]types.DeployedCompon
 
 		if deployErr != nil {
 			onFailure()
-
+			deployedComponents[idx].Status = types.ComponentStatusFailed
 			if p.isConnectedToCluster() {
 				if _, err := p.cluster.RecordPackageDeployment(ctx, p.cfg.Pkg, deployedComponents); err != nil {
 					message.Debugf("Unable to record package deployment for component %q: this will affect features like `zarf package remove`: %s", component.Name, err.Error())
@@ -213,6 +220,7 @@ func (p *Packager) deployComponents(ctx context.Context) ([]types.DeployedCompon
 
 		// Update the package secret to indicate that we successfully deployed this component
 		deployedComponents[idx].InstalledCharts = charts
+		deployedComponents[idx].Status = types.ComponentStatusSucceeded
 		if p.isConnectedToCluster() {
 			if _, err := p.cluster.RecordPackageDeployment(ctx, p.cfg.Pkg, deployedComponents); err != nil {
 				message.Debugf("Unable to record package deployment for component %q: this will affect features like `zarf package remove`: %s", component.Name, err.Error())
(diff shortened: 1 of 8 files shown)


src/pkg/packager/deploy.go
```diff
@@ -162,10 +163,16 @@ func (p *Packager) deployComponents(ctx context.Context) ([]types.DeployedCompon
...
+			// If this package has been deployed before, increment the package generation within the secret
+			if existingDeployedPackage, _ := p.cluster.GetDeployedPackage(ctx, p.cfg.Pkg.Metadata.Name); existingDeployedPackage != nil {
+				packageGeneration = existingDeployedPackage.Generation + 1
+			}
 		}
 
 		deployedComponent := types.DeployedComponent{
-			Name: component.Name,
+			Name:               component.Name,
+			Status:             types.ComponentStatusDeploying,
```
AustinAbro321, line 174: component status deploying is not used. Note that in the original implementation the package secret was updated before the deployment started. 

This also means you will have to update the test statements in the actions at `src/test/packages/22-git-data/zarf.yaml`

AustinAbro321 (commented):
Comment to make sure we sure the deploying status is utilized. As discussed synchronously an e2e test would be good here, given the difficultly of adding a unit test in deploy components atm I think it's reasonable to not add any for this change.

src/pkg/packager/deploy.go
sgettys: I'm not sure what you mean by not used. This updates the component to "deploying" before the deploy actually starts then will transition to failed or succeeded based on the actual deployComponent call on line 197. 

This shouldn't change the length of the deployedComponents list, just adds back the Status field so I think that the test statements in `src/test/packages/22-git-data/zarf.yaml` are still correct

sgettys: I can add an update to that test to cover the deployedComponent status though and make sure it's correct!

sgettys: Being able to "catch" a "deploying" status is going to be tough for the e2e tests

AustinAbro321: Yeah but updating the `deployedComponents` object doesn't update the secret in the cluster. We'll need a call to `p.cluster.RecordPackageDeployment(ctx, p.cfg.Pkg, deployedComponents)` before the deployment starts to do that. 

Once we start making the call to `p.cluster.RecordPackageDeployment(ctx, p.cfg.Pkg, deployedComponents)` before the component is deployed, the numbers in the test statements of `src/test/packages/22-git-data/zarf.yaml` will increase by 1 because we'll have started tracking the deployment before the actions start rather than after

sgettys: Ah ok makes sense!

src/test/e2e/37_component_status_test.go
```diff
@@ -0,0 +1,74 @@
...
+					}
+				}
+				require.NoError(t, err)
+				time.Sleep(2 * time.Second)
+			}
+		}
+	}()
+	_, _, err = e2e.Zarf(t, "package", "deploy", path, "--confirm")
+	require.NoError(t, err)
+	defer func() {
```
brandtkeller, line 68: Could be my ignorance and I don't see this pattern used elsewhere - would we want to potentially use `t.CleanUp()` instead of defer in the event an error occurred and the require statement exits the test?

src/test/e2e/37_component_status_test.go
```diff
@@ -0,0 +1,74 @@
...
+	// Stop channel getting the zarf state
+	stop := make(chan bool)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		deployingSeen := false
+		for {
+			select {
+			case <-stop:
```
brandtkeller, line 36: For sake of test resilience - would it make sense to have a timeout case as well?

brandtkeller (commented):
Adding some thoughts for the test at hand.

src/test/e2e/37_component_status_test.go
sgettys: Ah you're right, I thought I saw that pattern but I must have been hallucinating!

src/pkg/packager/deploy.go
```diff
@@ -180,7 +187,12 @@ func (p *Packager) deployComponents(ctx context.Context) ([]types.DeployedCompon
 
 		deployedComponents = append(deployedComponents, deployedComponent)
 		idx := len(deployedComponents) - 1
-
+		if p.isConnectedToCluster() {
+			if _, err := p.cluster.RecordPackageDeployment(ctx, p.cfg.Pkg, deployedComponents); err != nil {
```
AustinAbro321, line 191: You'll need to send `packageGeneration` into this function to set `DeployedPackage.Generation`. Otherwise on subsequent deploys the package will always start at 1.

AustinAbro321 (changes requested):
Nice test, this is close. Run `make lint-go` to fix the lint errors locally.

AustinAbro321 (approved):
LGTM

brandtkeller (approved):

PR #3557, opened 2025-03-05
feat: add more logging to packager2.Pull

mkcp:
## Description
Previously, Pull would only report errors to users by passing nested errors up the chain from ORAS. This focused more on internal library details than actionable user behavior. This PR is a first pass at adding debugging context with metadata and some info, error, and debug logs to packager2.Pull. 

## Checklist before merging

### DIFF
--- a/src/internal/packager2/pull.go
+++ b/src/internal/packager2/pull.go
@@ -7,12 +7,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/zarf-dev/zarf/src/pkg/logger"
 	"io"
 	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/defenseunicorns/pkg/helpers/v2"
 	"github.com/defenseunicorns/pkg/oci"
@@ -31,6 +33,11 @@ import (
 
 // Pull fetches the Zarf package from the given sources.
 func Pull(ctx context.Context, src, dir, shasum string, filter filters.ComponentFilterStrategy, publicKeyPath string, skipSignatureValidation bool) error {
+	l := logger.From(ctx)
+	start := time.Now()
+	defer func() {
+		l.Debug("done Pull", "src", src, "dir", dir, "duration", time.Since(start))
+	}()
 	u, err := url.Parse(src)
 	if err != nil {
 		return err
@@ -52,11 +59,13 @@ func Pull(ctx context.Context, src, dir, shasum string, filter filters.Component
 	isPartial := false
 	switch u.Scheme {
 	case "oci":
+		l.Info("starting pull from oci source", "src", src, "digest", shasum)
 		isPartial, tmpPath, err = pullOCI(ctx, src, tmpDir, shasum, filter)
 		if err != nil {
 			return err
 		}
 	case "http", "https":
+		l.Info("starting pull from http(s) source", "src", src, "digest", shasum)
 		tmpPath, err = pullHTTP(ctx, src, tmpDir, shasum)
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


src/internal/packager2/pull.go
```diff
@@ -31,6 +33,11 @@ import (
 
 // Pull fetches the Zarf package from the given sources.
 func Pull(ctx context.Context, src, dir, shasum string, filter filters.ComponentFilterStrategy, publicKeyPath string, skipSignatureValidation bool) error {
+	l := logger.From(ctx)
+	start := time.Now()
+	defer func() {
+		l.Debug("done Pull", "src", src, "dir", dir, "duration", time.Since(start))
```
brandtkeller, line 39: I like the intent here as I think (correct me) it meets some of the requests of late as exposing more information for consumers on what occurred and the results. 

is `done Pull` as eloquent of a message as we require? Concise and straight to the point.

src/internal/packager2/pull.go
```diff
@@ -112,12 +122,14 @@ func pullOCI(ctx context.Context, src, tarDir, shasum string, filter filters.Com
...
 	arch := config.GetArch()
-	remote, err := zoci.NewRemote(ctx, src, oci.PlatformForArch(arch))
+	platform := oci.PlatformForArch(arch)
+	remote, err := zoci.NewRemote(ctx, src, platform)
 	if err != nil {
 		return false, "", err
 	}
 	desc, err := remote.ResolveRoot(ctx)
 	if err != nil {
+		l.Error("unable to resolve oci descriptor", "os", platform.OS, "arch", platform.Architecture, "error", err)
```
AustinAbro321, line 132: The average user probably won't know what an OCI descriptor is. maybe something like `failed to pull package for platform {OS}/{ARCH}. 

I also think the error below is misleading. ResolveRoot is trying to get the manifest not the index.json. IMO combining this line and the below line would be the way to go for now

src/internal/packager2/pull.go
```diff
@@ -31,6 +33,11 @@ import (
 
 // Pull fetches the Zarf package from the given sources.
 func Pull(ctx context.Context, src, dir, shasum string, filter filters.ComponentFilterStrategy, publicKeyPath string, skipSignatureValidation bool) error {
+	l := logger.From(ctx)
+	start := time.Now()
+	defer func() {
```
AustinAbro321, line 38: Is the goal here to print this message even when we error?

mkcp: yeah, it saves from having to branch out the same debug end several times to track the duration. Just experimenting with it, not attached to this way of doing so.

src/internal/packager2/pull.go
mkcp: I also feel like our error here is misleading, could you clarify what you mean by combining the two lines though?

src/internal/packager2/pull.go
mkcp: Ideally this would point to the function itself, e.g. `done packager2.Pull ...` but packager2 is a moving target so I just named it Pull for now. --log-format="dev" will also give us line numbers, IIRC.

src/internal/packager2/pull.go
AustinAbro321: Combining these two lines
```go
		l.Error("unable to resolve oci descriptor", "os", platform.OS, "arch", platform.Architecture, "error", err)
		return false, "", fmt.Errorf("could not fetch images index: %w", err)
```
maybe something like
```go
return false, "", fmt.Errorf("could not find package at %s with architecture %s: %w", err, src, platform.Architecture)
```
Leaving out the `platform.OS` might make sense as well since from a user perspective since it's always `MultiOS` for Zarf packages which feels like more of an implementation detail.

src/internal/packager2/pull.go
AustinAbro321: I'd prefer it at the end of the function. I don't think we'd need to branch out for this, and I think it makes sense to avoid the done pull message when an error is returned.

mkcp: Ya makes sense, resolved in latest rev.

src/internal/packager2/pull.go
mkcp: That's fair, definitely agreed on OS not being actionable info for users. I'll go ahead and take out the error log and we'll rely on the err log near main. Fixed in latest

src/internal/packager2/pull.go
mkcp: Added the pkg prefix in latest

AustinAbro321 (approved):

Issue #3113, opened 2024-10-17
Unable to use OCI artifacts that are not all image layers


bcvanmeurs:
### Environment
Device and OS: MacOS
App version: v0.41.0
Kubernetes distro being used: Kind
Other:

### Steps to reproduce
1. Specify a non container image as an image for a zarf component. I used [ORAS](https://oras.land/) to create an OCI artefact and tried to add it to a package.
2. Run the create package command 

### Expected result

I expected the OCI artefact to be packaged like normal container images.
As the [documentation](https://docs.zarf.dev/ref/components/#container-images) says: "The image list is not limited to containers, any OCI image following the [Image Manifest specification](https://github.com/opencontainers/image-spec/blob/main/manifest.md) can be pulled"

### Actual Result

Package fails to create

### Visual Proof (screenshots, videos, text, etc)

```
Pulling 6 images ERROR: failed to create package: All attempts fail:
    #1: mismatched fs layers (1) and diff ids (0)
```

### Severity/Priority
Our use case relies on pulling artefacts from the container registry with initialisation containers. This is not possible at the moment 

### Additional Context

[Link to Zarf slack thread](https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1728918320871789)

bcvanmeurs:
Context from Austin Abro:
I ran into a similar issue previously with Helm OCI objects. It's an upstream crane issue when crane tries to cache non container OCI images. We implemented a workaround that checks if an [image consists entirely of image layers before caching](https://github.com/zarf-dev/zarf/blob/bd657a87d4065d1466aa01ad3660542c46688b4c/src/internal/packager/images/pull.go#L173-L179). However, perhaps we should have instead looked at the media type of the config. Looking at the [OCI spec](https://github.com/opencontainers/image-spec/blob/main/manifest.md#image-manifest-property-descriptions), that seems to be a more true way of identifying whether or not an OCI image is a container image.

bcvanmeurs:
I am happy to give it a try and create a PR if someone can help me with some pointer how to implement it.

AustinAbro321:
@bcvanmeurs What does the manifest of this image look like for you ?

For example, the image manifest of Helm chart objects has both a different config media type and a different layer media type. 
```json
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.cncf.helm.config.v1+json",
[...]

Issue #20, opened 2021-09-07
Ephemeral/parallelizable E2E tests


RothAndrew:
As a developer of Zarf, I want the E2E tests to happen on an ephemeral server, so that I can ensure that they are always running on a clean environment and so more than one test suite can run concurrently

---

Migrated from https://repo1.dso.mil/platform-one/big-bang/apps/product-tools/zarf/-/issues/34

PR #3559, opened 2025-03-07
feat: replace Crane usage for Image pull and Push with ORAS

### LINKED ISSUE #3194: image pull fail from corrupted Crane cache

### LINKED ISSUE #3434: Migrate off of crane for pulling and pushing OCI images
Crane has been an instrumental library to Zarf. It is responsible one of the most core features of our product, pulling and pushing images. However, we've had several issues while using crane. In particular, not accepting context, concurrent pulls and caching of non container OCI images tend to cause trouble. See:

- https://github.com/google/go-containerregistry/issues/1955
- https://github.com/google/go-containerregistry/issues/1941
-  #3113 
- #2594

### Alternatives
We should consider alternatives to fix these issues, and open ourselves up to further improvements. 
- [ORAS](https://github.com/oras-project/oras) (OCI registry as storage) is a project that has become popular in the past few years or so. It is a go library for interacting with OCI registries. We use ORAS already to publish and pull Zarf packages. 
- https://github.com/containers/image - this is the library that [skopeo](https://github.com/containers/skopeo) uses for image pulls. I have not looked deeply into the code, but it is worth looking at how another successful tool with a similar mission does pulls and pushes

### Additional context
Moving off Crane will present challenges. Crane also supports the oci-dir format, which syft uses to scan images local, this is how SBOMs are created during `zarf package create`. The Crane CLI is embedded directly into Zarf, removing it entirely will no doubt cause a breaking change in the workflow of some users. The Crane CLI has functionality to pull images from the local Docker daemon which would need replacement as well

AustinAbro321:
## Description

This replaces the Crane library uses for image pulls and pushes with ORAS. This does not change `zarf tools crane` operations.

Proposal - [ZEP-0019: Refactor Image operations #20](https://github.com/zarf-dev/proposals/pull/20)

This PR also enables the Zarf cache during E2E tests which helps verify that #3194 is truly getting solved.

Crane has special behavior for localhost. If localhost is used for image pulls or pushes it will try to connect over https, and fallback to http if that fails. This behavior was brought into Zarf since ORAS does not have it. 

Release notes:
Image pulls and pushes now respect the `--oci-concurrency` flag to determine the number of layers to pull/push in parallel. In previous versions of Zarf every image layer was pulled and pushed concurrently and the `--oci-concurrency` flag only applied to `zarf package publish` and `zarf package pull`. Additionally, the default value of `--oci-concurrency` changed from three to six. 

Zarf will no longer pull images in parallel, you may see a slowdown if you have a fast internet connection. We changed this because concurrent image pulls were causing reliability issues for some users. Generally image pulls are bottlenecked by the network so too many concurrent pulls can degrade performance. 

The `--plain-http` flag is now respected when connected to registries for image pulls and pushes.

Docker v1 images are no longer supported. Dockerhub completely stopped serving v1 images in June 2019, it's unlikely you will be effected by this, but if you are using a particularly old registry you may want to double check.

The format of the cache is changing. You are encouraged to run `zarf tools clear-cache` as existing image layers will not do anything. 

## Related Issue

Fixes #3434
Fixes #3194
Relates to #3113
Relates to #2104

## Checklist before merging


RothAndrew:
let's gooooooo! so excited for this lol

src/pkg/cluster/tunnel.go
```diff
@@ -153,6 +154,15 @@ func (c *Cluster) ConnectToZarfRegistryEndpoint(ctx context.Context, registryInf
 		if tunnel, err = c.NewTunnel(ZarfNamespaceName, SvcResource, ZarfRegistryName, "", 0, ZarfRegistryPort); err != nil {
 			return "", tunnel, err
 		}
+	} else if dns.IsServiceURL(registryInfo.Address) {
```
AustinAbro321, line 157: In the packager2 mirror refactor the push logic was changed to take Kubernetes services as input, this brings this function in parity with that.

src/test/e2e/39_crane_to_oras_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/zarf-dev/zarf/src/pkg/utils/exec"
+)
+
+func TestCraneToORAS(t *testing.T) {
```
AustinAbro321, line 18: A relatively novel test, this pulls down the latest Zarf version using Crane and makes sure that packages built with Crane still work with ORAS and vice versa.

src/internal/packager2/layout/sbom.go
```diff
@@ -82,6 +82,18 @@ func generateSBOM(ctx context.Context, pkg v1alpha1.ZarfPackage, buildPath strin
 		}
 	}
 
+	// Stereoscope creates extra layers outside of the OCI formatted directory in the cache
```
AustinAbro321, line 85: It seems extra layers were also getting created with Crane, but since this follows the Crane format for cached layers, we never noticed. https://github.com/anchore/stereoscope/issues/387

src/cmd/viper.go
```diff
@@ -213,7 +213,7 @@ func setDefaults() {
 	v.SetDefault(VLogFormat, string(logger.FormatConsole))
 
 	// Package defaults that are non-zero values
-	v.SetDefault(VPkgOCIConcurrency, 3)
+	v.SetDefault(VPkgOCIConcurrency, 6)
```
soltysh, line 216: Why this change? Is ORAS more performant?

src/internal/packager/images/cache/cache.go
```diff
@@ -0,0 +1,153 @@
+/*
+Copyright The ORAS Authors.
```
soltysh, line 2: License update?

src/internal/packager/images/common.go
```diff
@@ -28,10 +32,88 @@ type PullConfig struct {
...
+const (
+	DockerLayer                    = "application/vnd.docker.image.rootfs.diff.tar.gzip"
+	DockerUncompressedLayer        = "application/vnd.docker.image.rootfs.diff.tar"
+	OCILayer                       = "application/vnd.oci.image.layer.v1.tar+gzip"
+	OCILayerZStd                   = "application/vnd.oci.image.layer.v1.tar+zstd"
+	OCIUncompressedLayer           = "application/vnd.oci.image.layer.v1.tar"
+	DockerForeignLayer             = "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip"
+	OCIRestrictedLayer             = "application/vnd.oci.image.layer.nondistributable.v1.tar+gzip"
+	OCIUncompressedRestrictedLayer = "application/vnd.oci.image.layer.nondistributable.v1.tar"
+)
```
soltysh, line 53: Why defining those rather than just use these https://github.com/opencontainers/image-spec/blob/main/specs-go/v1/mediatype.go ?

src/internal/packager/images/pull.go
```diff
@@ -7,506 +7,393 @@ package images
...
+		return nil, fmt.Errorf("failed to get credentials: %w", err)
+	}
+	client := &auth.Client{
+		Client:     retry.DefaultClient,
+		Cache:      auth.NewCache(),
+		Credential: credentials.Credential(credStore),
+	}
+	platform := &ocispec.Platform{
+		Architecture: cfg.Arch,
+		OS:           "linux",
```
soltysh, line 77: Add a TODO here, we'll likely want to expose this as a configuration option to allow non-linux images to be pulled sometime in the future.

src/internal/packager/images/pull.go
```diff
@@ -7,506 +7,393 @@ package images
...
-			defer shaLock.Unlock()
-			for _, layer := range layers {
-				digest, err := layer.Digest()
-				if err != nil {
-					return fmt.Errorf("unable to get digest for image layer: %w", err)
+			fetchOpts := oras.DefaultFetchBytesOptions
+			fetchOpts.FetchOptions.TargetPlatform = platform
+			desc, b, err := oras.FetchBytes(ectx, localRepo, overriddenRef, fetchOpts)
+			if err != nil {
+				if strings.Contains(err.Error(), "toomanyrequests") {
```
soltysh, line 121: Add a TODO here, we should consider a backoff logic for those cases. k8s has several primitives [here](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/wait/backoff.go) that we cold just re-use.

src/internal/packager/images/pull.go
```diff
@@ -7,506 +7,393 @@ package images
...
-			fetched[refInfo] = img
-
+			// Both oci and docker manifest types can be marshalled into a manifest
+			// https://github.com/oras-project/oras-go/blob/853e0125ccad32ff691e4ed70e156c7619021bfd/internal/manifestutil/parser.go#L37
+			var manifest ocispec.Manifest
+			if err := json.Unmarshal(b, &manifest); err != nil {
+				return err
+			}
+			size := getSizeOfImage(desc, manifest)
+			m.Lock()
```
soltysh, line 148: Nit: can you rename that lock to be more meaningful.

src/internal/packager2/layout/create.go
```diff
@@ -169,7 +166,7 @@ func CreatePackage(ctx context.Context, packagePath string, opt CreateOptions) (
 		l.Info("generating SBOM")
 		err = generateSBOM(ctx, pkg, buildPath, sbomImageList)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to generate SBOM: %w", err)
```
soltysh, line 169: :+1: for adding context for the error

soltysh (commented):
I haven't looked too deeply into code correctness, I trust others will. But I really like the simplicity behind the oras. I also looked at https://github.com/zarf-dev/zarf/pull/3584 where you're combining the two approaches and I have a general comment that I'd probably create separate set of pull and push files with just the new behavior. Introduce a new interface and then depending on that config you're setting in the preRun you'd be able to instantiate/use one or the other implementation. But keep them in separate files for ease of removing in the near future.

src/cmd/viper.go
AustinAbro321: Crane pushes and pulls all layers concurrently always, and had no way to respect this flag so for container operations the concurrency is actually decreasing. 

OCI concurrency flag is used for Zarf package OCI operations as well. The ORAS CLI defaults to three, so that's what we went with originally for Zarf packages, but I think three is conservative. Skopeo for example, goes with six and that feels like a more sane inbetween.

src/internal/packager/images/cache/cache.go
AustinAbro321: Yeah I wanted to ask about this, should I just add both the Zarf license header and ORAS header? I got the code from here, https://github.com/oras-project/oras/blob/main/internal/cache/target.go.

AustinAbro321: Related, I created issue https://github.com/oras-project/oras-go/issues/881, to ask for a cache implementation to be exposed by the library, I'll add a comment there for future tracking

src/cmd/viper.go
Noxsios: won't block, but prob a good idea to call out in release notes the default has increased

there's a higher likelihood with 6 coroutines of pulling multiple large layers at a time, saturating a users network connection + causing higher memory usage

src/internal/packager2/layout/create.go
```diff
@@ -142,17 +142,14 @@ func CreatePackage(ctx context.Context, packagePath string, opt CreateOptions) (
 			RegistryOverrides:    opt.RegistryOverrides,
 			CacheDirectory:       filepath.Join(cachePath, ImagesDir),
 		}
-		pulled, err := images.Pull(ctx, pullCfg)
+		manifests, err := images.Pull(ctx, pullCfg)
```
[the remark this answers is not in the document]
AustinAbro321: Great catch, it was not intentional. Updated to pass concurrency throughout the system

Noxsios (changes requested):

src/internal/packager/images/common.go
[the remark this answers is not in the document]
Noxsios: This is a sane and logical progression, but just in case users are still using v1 registries, it is worth making sure everyone knows this.

src/internal/dns/dns.go
```diff
@@ -46,3 +48,49 @@ func ParseServiceURL(serviceURL string) (string, string, int, error) {
...
+// Inspired from https://github.com/google/go-containerregistry/blob/098045d5e61ff426a61a0eecc19ad0c433cd35a9/pkg/name/registry.go
+
+// Detect more complex forms of local references.
+var reLocal = regexp.MustCompile(`.*\.local(?:host)?(?::\d{1,5})?$`)
+
+// Detect the loopback IP (127.0.0.1)
+var reLoopback = regexp.MustCompile(regexp.QuoteMeta("127.0.0.1"))
+
+// Detect the loopback IPV6 (::1)
+var reipv6Loopback = regexp.MustCompile(regexp.QuoteMeta("::1"))
```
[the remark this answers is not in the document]
AustinAbro321: I just copied this code over from Crane, so I think I'll keep it as is for simplicity sake

src/internal/packager/images/pull.go
```diff
@@ -7,506 +7,399 @@ package images
...
 
-			cacheImg, err := utils.OnlyHasImageLayers(img)
+			repo.Reference, err = registry.ParseReference(overriddenRef)
 			if err != nil {
 				return err
 			}
-			if cacheImg && cfg.CacheDirectory != "" {
-				img = cache.Image(img, cache.NewFilesystemCache(cfg.CacheDirectory))
-			}
+			repo.PlainHTTP = cfg.PlainHTTP || dns.IsLocalhost(repo.Reference.Registry)
```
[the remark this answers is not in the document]
AustinAbro321: I'm not sure, I considered doing a boolptr with plain http to and letting localhost be fallback when it's not set. I wanted to keep parity with Crane for now, and then evaluate further if it's something we want to keep. I will see if I can get an https connection on localhost, and cause this to error, and see if that's what happens in current Zarf / Crane too.

src/pkg/utils/image.go
```diff
@@ -25,20 +25,24 @@ func LoadOCIImage(imgPath string, refInfo transform.Image) (v1.Image, error) {
...
+			(manifest.Annotations[ocispec.AnnotationRefName] == refInfo.Path+refInfo.TagOrDigest && refInfo.Host == "docker.io") {
 			// This is the image we are looking for, load it and then return
-			return layoutPath.Image(manifest.Digest)
+			img, err := layoutPath.Image(manifest.Digest)
+			if err != nil {
+				return nil, fmt.Errorf("failed to lookup image %s: %w", refInfo.Reference, err)
+			}
+			return img, nil
 		}
 	}
```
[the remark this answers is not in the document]
AustinAbro321: Yeah it would, initially I thought it was only used at create time, but I missed that it's used by the injector code so it'd break older init packages. I believe we could get by with only checking for the `AnnotationBaseImageName`, but I don't think there is any harm in checking for AnnotationRefName as well which may enable moving away from the extra `AnnotationBaseImageName` annotation in the future

src/internal/packager/images/push.go
```diff
@@ -6,97 +6,149 @@ package images
...
-		pushImage := func(img v1.Image, name string) error {
-			if tunnel != nil {
-				return tunnel.Wrap(func() error { return crane.Push(img, name, pushOptions...) })
-			}
-			return crane.Push(img, name, pushOptions...)
-		}
+	src, err := oci.NewWithContext(ctx, cfg.SourceDirectory)
+	if err != nil {
+		return fmt.Errorf("failed to instantiate oci directory: %w", err)
+	}
```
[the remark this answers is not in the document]
AustinAbro321: Yeah I like making a separate function for it. 

We have to do it both on pull and push. So we can both push old packages with new Zarf and push new packages with old Zarf.

src/internal/packager/images/pull.go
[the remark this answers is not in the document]
AustinAbro321: I updated this to do a fallback similar to Crane. First we will ping the registry using https, if that doesn't work and the user either is connecting to localhost or used the `--plain-http` flag we will connect try using http. Curious on your thoughts here @Noxsios

src/internal/packager/images/pull.go
```diff
@@ -7,506 +7,399 @@ package images
...
+		var pullSrc oras.ReadOnlyTarget
+		var err error
+		repo := &orasRemote.Repository{}
+		repo.Reference, err = registry.ParseReference(imageInfo.registryOverrideRef)
+		if err != nil {
+			return fmt.Errorf("failed to parse image reference %s: %w", imageInfo.registryOverrideRef, err)
 		}
-		wStart := time.Now()
-		size, err := getSizeOfImage(img)
+		repo.PlainHTTP = cfg.PlainHTTP || dns.IsLocalhost(repo.Reference.Registry)
```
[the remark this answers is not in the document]
AustinAbro321: I'm unsure if I want to keep this logic as is or move it over to the fallback logic implemented on push. I think I'd rather keep it as is and keep it simple on create. Most registries redirect http pulls to https seemingly. If anyone runs into an issue on create, we should be able to update this to fallback as well.

src/internal/packager/images/push.go
```diff
@@ -6,97 +6,149 @@ package images
...
+	if c != nil {
+		var err error
+		registryURL, tunnel, err = c.ConnectToZarfRegistryEndpoint(ctx, cfg.RegInfo)
 		if err != nil {
 			return err
 		}
-		toPush[refInfo] = img
+		if tunnel != nil {
+			defer tunnel.Close()
+		}
```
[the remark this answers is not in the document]
Noxsios: ah yeah, I forgot this is a `var` pointer populated by a later function. The entire tunnel + cluster + registryURL logic feels a bit jank to me. But that's a fix for a later time, this code is fine for now. Feel free to close/resolve.

src/internal/packager/images/pull.go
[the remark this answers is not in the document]
Noxsios: I'm in the camp of perform no magic logic.

It's not clear in using the tool that it would automagically downgrade to HTTP when talking _only_ to localhost like things.

Crane does so to make local testing easier / not need `--plain-http` flag each time but imo since there is no HTTP spec wide rule / documentation for how this downgrade logic works, we should avoid it. There _is_ an HTTP upgrade spec, using 301/302 but not a downgrade redirect.

The user specified `localhost:555/nginx:latest` w/o any `--plain-http`, we try `https://localhost:555/v2/nginx/manifests/latest`, and fail if its supposed to be `http`.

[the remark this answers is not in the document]
Noxsios: But I don't believe I am the best authority on this matter. I'll start a discussion thread and update this comment w/ results.

mkcp (approved):
LGTM, nice work! License matches and the included attribution should be sufficient.

mkcp (approved):
LGTM, discussed final review synced up

PR #3560, opened 2025-03-07
fix: support for ephemeral container mutation

### LINKED ISSUE #2153: EphemeralContainers via kubectl debug bypass Zarf webhook
### Environment
Device and OS: macOS
App version: 0.31.1

### Steps to reproduce
1. Zarf init
2. Deploy some things with zarf
3. `kubectl debug -it -n test test-pod --image busybox` (with some pod in cluster) 

### Expected result

Zarf would mutate the image to be one from zarf's registry.

### Actual Result

The update to the pod bypasses zarf's webhook.

### Severity/Priority

Not critical, user can always specify the zarf image registry directly.

### Additional Context

There are two issues at play:
- Zarf's webhook needs `pods/ephemeralcontainers` added as a resource (see https://www.github.com/kyverno/kyverno/issues/2821 as an example of this)
- Even with ^ added `kubectl debug` call would not get patched because the pod itself already went through the patching lifecycle - https://github.com/defenseunicorns/zarf/blob/06c8e53e6a88ced00f1d8bc769e3d3f8efe7751a/src/internal/agent/hooks/pods.go#L50

brandtkeller:
## Description

See the related issue - Ephemeral Containers are not explicitly supported by the zarf agent for mutation. This adds support by ensuring the `pods/ephemeralcontainers` subresource is being included in the `MutatingWebhookConfiguration`. 

largely this should always be an operation that is being done on a pod that has already been patched - therefore we need to enable an exception for a pre-patched pod to still allow mutation of the ephemeral containers even if previously patched. 

## Changes
A pod cannot be created or updated with ephemeral containers defined declaratively. The original code for ephemeral containers should not be reachable - therefor moved it out of the primary pod mutation logic and into the ephemeral containers mutation logic. 

## Related Issue

Fixes #2153 

## Checklist before merging


brandtkeller:
Current side-effect - image annotations get updated as a not-entirely-idempotent workflow. Working to address this.

AustinAbro321:
@brandtkeller To give some more context, the reason the agent is not fully idemopotent is because of the crchash we append to images, see https://docs.zarf.dev/ref/init-package/#image-mutation-to-unique-hashed-tags. If the agent was to mutate these resources twice, the crchash would no longer be correct. Open to other ways of checking besides a patched label, but I know that confused me when I first saw it

brandtkeller:
Thanks for the context @AustinAbro321. Agree on multiple mutation side-effects. 

I do believe that the transform has an [early return](https://github.com/zarf-dev/zarf/blob/main/src/pkg/transform/image.go#L34) in the case of the `crchash` scenario that would prevent re-compute? Does this sound correct or is there something I am not considering?

AustinAbro321:
Great point, I thought I remembered running into a gotcha here, maybe it was just the annotations

AustinAbro321 (commented):
I think we should add to the E2E tests to ensure the kube API server is sending us the information we expect, I.E. an empty sub resource on non ephemeral container requests and responding to the changes in the `webhook.yaml` properly. 

Other than that looks good

src/test/e2e/38_ephemeral_container_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// deploy the above package
+	stdOut, stdErr, err = e2e.Zarf(t, "package", "deploy", path, "--confirm")
+	require.NoError(t, err, stdOut, stdErr)
+
+	// using a pod the package deploys - run a kubectl debug command
+	stdOut, stdErr, err = e2e.Kubectl(t, "debug", "test-pod", "-n", "test", "--image=busybox:1.36", "--profile", "general")
+	require.NoError(t, err, stdOut, stdErr)
+
+	// there is no native 'wait' logic for ephemeral containers
```
brandtkeller, line 36: This is a bit "cludgy" - if there is a better pattern here - I am open to it. Otherwise there is some race condition between the kubectl `debug` command and `get` command such that it returns an empty string (possibly getting information too quickly before the mutation and admission has occurred).

src/test/e2e/38_ephemeral_container_test.go
```diff
@@ -0,0 +1,73 @@
...
+		require.NoError(t, err)
+		if podStdOut != "" {
+			t.Log("Ephemeral container detected!", podStdOut)
+			ephemeralContainer = podStdOut
+			break
+		}
+
+		// Check timeout
+		if time.Since(startTime) > timeout {
+			t.Log("Timeout reached! Ephemeral container not found.")
```
AustinAbro321, line 54: IMO we should simply error `t.Fail` here

src/test/e2e/38_ephemeral_container_test.go
```diff
@@ -0,0 +1,73 @@
...
+		time.Sleep(interval)
+	}
+
+	t.Log("Ephemeral Container: ", ephemeralContainer)
+
+	// ensure the image used contains the internal zarf registry (IE mutated)
+	require.Contains(t, ephemeralContainer, "127.0.0.1:31337/library/busybox:1.36-zarf-")
+
+	// cleanup - should perform cleanup in the event of pass or fail - separate from defer
+	t.Cleanup(func() {
```
AustinAbro321, line 69: this needs to be moved up to directly after deploy. If we fail on require.Contains here for example, the cleanup will never register

AustinAbro321 (changes requested):
A few small tests things

src/test/e2e/38_ephemeral_container_test.go
```diff
@@ -0,0 +1,73 @@
...
+	stdOut, stdErr, err = e2e.Zarf(t, "package", "deploy", path, "--confirm")
+	require.NoError(t, err, stdOut, stdErr)
+
+	// using a pod the package deploys - run a kubectl debug command
+	stdOut, stdErr, err = e2e.Kubectl(t, "debug", "test-pod", "-n", "test", "--image=busybox:1.36", "--profile", "general")
+	require.NoError(t, err, stdOut, stdErr)
+
+	// there is no native 'wait' logic for ephemeral containers
+	timeout := 10 * time.Second // Timeout after 10 seconds
+	interval := 2 * time.Second // Check every 2 seconds
```
AustinAbro321, line 38: subjective nit: I'm generally okay with comments in tests, especially when giving detail on a block is needed within the context of a test or what the code is doing, but IMO comments like these are redundant.

src/test/e2e/38_ephemeral_container_test.go
AustinAbro321: Looks like should work 
```bash
kubectl wait --for=condition=Ready pod/my-debug-pod --timeout=10s
```

mkcp: Definitely favor an await-style API over polling and checking 👍

src/test/e2e/38_ephemeral_container_test.go
```diff
@@ -0,0 +1,72 @@
...
+	require.NoError(t, err, stdOut, stdErr)
+	packageName := fmt.Sprintf("zarf-package-basic-pod-%s-0.0.1.tar.zst", e2e.Arch)
+	path := filepath.Join(tmpdir, packageName)
+
+	// deploy the above package
+	stdOut, stdErr, err = e2e.Zarf(t, "package", "deploy", path, "--confirm")
+	require.NoError(t, err, stdOut, stdErr)
+
+	// cleanup - should perform cleanup in the event of pass or fail
+	t.Cleanup(func() {
```
mkcp, line 33: This may be a matter of taste, but I'd rather we decouple cleanup from the order of the test and don't ensure that it didn't error. I'm assuming that this call will error out if it doesn't find `"basic-pod"` to delete, meaning we always have to call it after `package deploy` Making cleanup idempotent (doesn't matter how, where, or how many times we we call it, it just woks) makes the test more resilient to change.

I'm suggesting we should move this near the top of the function and declare all the expected resources we want to clean up after - errors can be ignored.

src/test/e2e/38_ephemeral_container_test.go
```diff
@@ -0,0 +1,72 @@
...
+	require.NoError(t, err, stdOut, stdErr)
+
+	// there is no native 'wait' logic for ephemeral containers
+	timeout := 10 * time.Second
+	interval := 2 * time.Second
+	startTime := time.Now()
+
+	var ephemeralContainer string
+
+	for {
```
mkcp, line 49: Not requesting changes, just food for thought. With a wait API, this problem can be expressed as a blocking future/promise. See "Use Channels as Futures/Promises" here. https://go101.org/article/channel-use-cases.html

src/test/e2e/38_ephemeral_container_test.go
brandtkeller: The issue here being that the pod status never changes and the use of `debug` is adding a `ephemeral container`  to an existing pod.

src/test/e2e/38_ephemeral_container_test.go
brandtkeller: love these kind of recommendations - I am quite vanilla in my goroutines so I took a stab.

src/test/e2e/38_ephemeral_container_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestEphemeralContainers(t *testing.T) {
+	t.Log("E2E: Ephemeral Containers mutation")
+
+	// cleanup - should perform cleanup in the event of pass or fail
+	t.Cleanup(func() {
+		e2e.Zarf(t, "package", "remove", "basic-pod", "--confirm") //nolint:errcheck
```
mkcp, line 21: 👍

src/test/e2e/38_ephemeral_container_test.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+
+			t.Log("Waiting for ephemeral...")
+		}
+	}()
+
+	var ephemeralContainer string
+
+	select {
+	case ephemeralContainer = <-resultCh:
```
mkcp, line 69: 👍 Clean

mkcp (dismissed):
LGTM - nice job

src/test/e2e/38_ephemeral_container_test.go
```diff
@@ -0,0 +1,78 @@
...
+		t.Log("Ephemeral Container: ", ephemeralContainer)
+	case <-timeoutCh:
+		t.Log("Timeout reached! Ephemeral container not found.")
+		t.Fail()
+	}
+
+	// Ensure the image used contains the internal zarf registry (IE mutated)
+	require.Contains(t, ephemeralContainer, "127.0.0.1:31337/library/busybox:1.36-zarf-")
+}
```
AustinAbro321, line 78: I tried it out and it seems we can replace this block with the following Kubectl wait command. The CRC checksum is consistent, it's based on the image name, alternatively we could use a shasum
```go
	stdOut, stdErr, err = e2e.Kubectl(t, "wait", "--namespace=test", "--for=jsonpath={.spec.ephemeralContainers[*].image}=127.0.0.1:31337/library/busybox:1.36-zarf-2140033595",
		"pod/test-pod", "--timeout=10s")
````

File: src/test/packages/38-ephemeral-container/zarf.yaml [hunk 10]
```diff
@@ -0,0 +1,17 @@
...
+components:
+  - name: busybox
+    required: true
+    manifests:
+      - name: busybox
+        namespace: test
+        files:
+          - pod.yaml
+    images:
+      - busybox:1.36
```
AustinAbro321, line 17: Unfortunately, github actions runners tend to get rate limited from dockerhub pretty quickly so we tend to keep image pulls to ghcr. Probably something we should fix, but for now I'd recommend using `ghcr.io/stefanprodan/podinfo:6.4.0` or something else from ghcr.

src/test/e2e/38_ephemeral_container_test.go
brandtkeller: Great catch - I had not thought about `--for` in that wait command...

AustinAbro321 (approved):
LGTM

PR #3562, opened 2025-03-10
chore(docs): zarf principles draft

### LINKED ISSUE #3552: Document the core Zarf principles
## Purpose
The guidelines and principles for which Zarf is both developed as well as future changes should stem from a set of design principles such that new requests or development can be measured for adherence with the design principles. 

This also mitigates new feature suggestions that may have unforeseen implications from being seen as subjectively reviewed (for what does or does not get implemented) and instead may provide more objectives measures for reference and review. 

Following inspirations from sources such as:
- [aeron](https://github.com/aeron-io/aeron/wiki/Design-Principles)

## Potential Targets
- Airgap native: Ensuring all functionality in zarf supports an airgapped ecosystem by default. Configurations can be exposed which rely on external/host dependencies - but no new functionality should ever prohibit zarf from being able to perform the lifecycle of actions without said dependency. 
- Tooling Inclusion: Selection of tools which get vendored into Zarf and the justification

brandtkeller:
## Description

WIP: drafting some principles for discussion. 

Zarf needs objective criteria for reviewing new or modified behavior when submitted from any entity. 

This will better allow capability review without the potential for interpreting the decision as subjective. 

## Related Issue

Fixes #3552

Relates to #

## Checklist before merging

### DIFF
--- a/site/src/content/docs/contribute/principles.mdx
+++ b/site/src/content/docs/contribute/principles.mdx
@@ -1,0 +1,49 @@
+---
+title: Zarf Principles
+---
+
+Zarf is a tool for continuous software delivery to disconnected environments. To ensure that Zarf remains true to its mission and continues to provide value for its users, the following design principles serve as a framework for evaluating new features and capabilities.
+
+## 1. Airgap-First Functionality
+Zarf is designed for airgapped environments, meaning that all functionality should operate without requiring external dependencies. This includes:
+- Ensuring that all packaged software and dependencies are self-contained.
+- Avoiding reliance on external repositories or registries during runtime.
+- Providing a seamless user experience for both online and offline operations.
+
+## 2. Minimal and Secure Binary
+The Zarf binary must remain lightweight and secure while offering core functionality. Considerations include:
+- Baking in only essential tools necessary for package management, deployment, and validation.
+- Avoiding unnecessary bloat by carefully selecting embedded utilities.
+- Ensuring security best practices, including dependency verification and signed releases.
+
+## 3. Clear Scope Within Kubernetes Environments
+Zarf operates in Kubernetes clusters but does not attempt to replace existing Kubernetes tooling. Instead, Zarf should:
+- Focus on enabling package management and deployment in airgapped clusters.
+- Complement, rather than compete with, Kubernetes-native solutions (e.g., Helm, Kustomize, and Operator patterns).
+- Facilitate cluster initialization and ongoing software management without enforcing opinionated infrastructure choices.
+
+## 4. Declarative and Reproducible Workflows
+To maintain consistency in deployments, Zarf should emphasize:
+- Declarative package definitions that allow for repeatable deployments.
+- Versioned and auditable package formats that ensure integrity across different environments.
+- Support for GitOps workflows where applicable.
+
+## 5. Operational Simplicity
+Zarf should remain simple to use and operationalize, with considerations such as:
+- Clear and concise configuration files that reduce complexity.
+- A user-friendly CLI that abstracts away unnecessary details while allowing expert-level control when needed.
+- Automation and scripting capabilities that streamline usage in constrained environments.
+
+## 6. Compatibility and Extensibility
+Zarf should be flexible enough to work in various environments while maintaining its core mission. This means:
+- Supporting multiple architectures and operating systems where feasible.
+- Allowing extensibility through custom package configurations and integrations.
+- Ensuring compatibility with industry standards to facilitate interoperability.
+
+## 7. Community and Open Source Focus
+As an open-source project, Zarf should prioritize community involvement and transparency by:
+- Encouraging feedback and contributions from users and maintainers.
+- Providing clear documentation and guidance for adoption.
+- Adhering to open governance principles to guide long-term sustainability.
+
+These principles help guide the evolution of Zarf while ensuring that new features and capabilities align with its core mission. By adhering to these guidelines, Zarf can continue to serve as a robust, reliable tool for software delivery.


site/src/content/docs/contribute/principles.mdx
```diff
@@ -0,0 +1,49 @@
...
+The Zarf binary must remain lightweight and secure while offering core functionality. Considerations include:
+- Baking in only essential tools necessary for package management, deployment, and validation.
+- Avoiding unnecessary bloat by carefully selecting embedded utilities.
+- Ensuring security best practices, including dependency verification and signed releases.
+
+## 3. Clear Scope Within Kubernetes Environments
+Zarf operates in Kubernetes clusters but does not attempt to replace existing Kubernetes tooling. Instead, Zarf should:
+- Focus on enabling package management and deployment in airgapped clusters.
+- Complement, rather than compete with, Kubernetes-native solutions (e.g., Helm, Kustomize, and Operator patterns).
+- Facilitate cluster initialization and ongoing software management without enforcing opinionated infrastructure choices.
```
Racer159, line 23: It may be worth calling out that in complementing the K8s environment it should also "fit in" from a user and dev experience perspective - i.e. Zarf shouldn't invent its own workflows unless that workflow is just something that doesn't exist (Zarf Values <> Zarf Variables would be a good example there).

site/src/content/docs/contribute/principles.mdx
```diff
@@ -0,0 +1,49 @@
...
+To maintain consistency in deployments, Zarf should emphasize:
+- Declarative package definitions that allow for repeatable deployments.
+- Versioned and auditable package formats that ensure integrity across different environments.
+- Support for GitOps workflows where applicable.
+
+## 5. Operational Simplicity
+Zarf should remain simple to use and operationalize, with considerations such as:
+- Clear and concise configuration files that reduce complexity.
+- A user-friendly CLI that abstracts away unnecessary details while allowing expert-level control when needed.
+- Automation and scripting capabilities that streamline usage in constrained environments.
```
Racer159, line 35: Developer simplicity I think is also key here - there are special things like the `zarf dev find-images` commands that really help developers build up their packages and gather resources.

site/src/content/docs/contribute/principles.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+## 6. Compatibility and Extensibility
+Zarf should be flexible enough to work in various environments while maintaining its core mission. This means:
+- Supporting multiple architectures and operating systems where feasible.
+- Allowing extensibility through custom package configurations and integrations.
```
Racer159, line 40: Worth calling out its use as a library explicitly here? (since "integrations" could be many things and some of them we may not prefer (i.e. a python script that generates Zarf packages and drives Zarf's CLI would be an "integration" but IMO that would be treated differently from a Go CLI importing Zarf as a library)

site/src/content/docs/contribute/principles.mdx
```diff
@@ -0,0 +1,49 @@
...
+- Versioned and auditable package formats that ensure integrity across different environments.
+- Support for GitOps workflows where applicable.
+
+## 5. Operational Simplicity
+Zarf should remain simple to use and operationalize, with considerations such as:
+- Clear and concise configuration files that reduce complexity.
+- A user-friendly CLI that abstracts away unnecessary details while allowing expert-level control when needed.
+- Automation and scripting capabilities that streamline usage in constrained environments.
+
+## 6. Compatibility and Extensibility
```
Racer159, line 37: It may be worth calling out specific expectations around stability of interfaces - i.e. what is "stable" in the CLI (probably a lot but not `zarf internal`?) or what is exposed in the library

Racer159 (commented):
just a few comments for thought

site/src/content/docs/contribute/principles.mdx
brandtkeller: Great point - added an item to reflect this.

site/src/content/docs/contribute/principles.mdx
brandtkeller: Added!

site/src/content/docs/contribute/principles.mdx
brandtkeller: I give mention to the SDK now (Added) - there is likely room for further specification but we still need to establish that scope in other ways as well.

site/src/content/docs/contribute/principles.mdx
brandtkeller: I agree that we should certainly do that more broadly - still looking for other opportunities within this document that can provide that.

AustinAbro321 (commented):
This document looks great, and I'm generally in agreement with the principals. "measured by" includes many metrics, both qualitative and quantitative. Is the plan to work towards measuring and reporting these metrics over time? Or is the intent to say that these are the concrete impacts that these goals should have on Zarf?

brandtkeller:
>  Is the plan to work towards measuring and reporting these metrics over time? Or is the intent to say that these are the concrete impacts that these goals should have on Zarf?

I'd say it's a mix of both that we can refine and report on iteratively where valuable. Some are guardrails - IE "Adoption of signed releases and integrity verification methods" <- we already do this but we want to ensure we don't violate this explicitly. Others are continual signals that need feedback loops - "Ensuring compatibility with industry standards". Then we get to things we can measure -> binary/package size / time to first successful deployment/ etc. 

Broad enough to help us find applicable paths for new suggestions or existing logic.

mkcp (approved):
LGTM!

site/src/content/docs/contribute/principles.mdx
Racer159: yeah - I think the level you added is likely good for a principles doc and then other docs / information can help shape the rest.

site/src/content/docs/contribute/principles.mdx
Racer159: Yeah I agree it is kind of a meta-point - it may be best for that to be in the community section? i.e. making sure that the stability expectations are understood by the community (i.e. they look like Helm / K8s expectations) and that there is a path for contribution that includes some levels of experimentation to grow feedback.

AustinAbro321 (approved):

site/src/content/docs/contribute/principles.mdx
brandtkeller: I'm going to resolve this currently in favor of hopefully codifying the intent around this ambiguity alongside the feature flag work. If that doesn't coalesce then we certainly could use a callout.

PR #3568, opened 2025-03-11
feat: accept version flag for `zarf tools download-init`

### LINKED ISSUE #3366: Allow `zarf tools download-init` to download a specific zarf version

Currently, when using `zarf tools download-init`, it downloads the latest version. This can be problematic if we're building packages that aren't compatible with the latest version of Zarf. It would be beneficial to specify the version in the command:

```bash
zarf tools download-init -v v0.45.0
```

- **Given** a state

  ```bash
  zarf tools download-init -v v0.45.0
  ```

- **When** an action is taken

  When the `-v` or `--version` flags are passed

- **Then** something happens

  The desired version of Zarf is downloaded

Currently, the alternative is to use the `gh` CLI or `curl`. While that works, adding this functionality to `zarf tools download-init` would be more convenient.

jbrewer3:
## Description
Updates to allow downloading of versioned zarf init packages
...

## Related Issue

Fixes #3366

## Checklist before merging


brandtkeller:
`make docs-and-schema` should fix the error - otherwise this looks functional from my local testing.

brandtkeller (changes requested):
Adds a neat capability to the `download-init` command that many may find useful. 

Given the parameters remained the same - I could not find an attack vector that this change might enable. 

Arbitrary strings are validated by the OCI logic - if the tag doesn't exist the logs indicate as such. That does not mean there isn't potential for misuse.

Anytime we introduce accepting input from the user - I would prefer we perform input validation. In `src/pkg/zoci/utils.go` this might look something like:

```
// GetInitPackageURL returns the URL for the init package for the given version with strict validation.
func GetInitPackageURL(version string) (string, error) {
    // Strict SemVer validation (vMAJOR.MINOR.PATCH with optional pre-release and metadata)
    var validVersion = regexp.MustCompile(`^v\d+\.\d+\.\d+(-[a-zA-Z0-9._-]+)?(\+[a-zA-Z0-9._-]+)?$`)

    if !validVersion.MatchString(version) {
        return "", fmt.Errorf("invalid version: must follow SemVer and start with 'v' (e.g., v0.50.0)")
    }

    return fmt.Sprintf("ghcr.io/zarf-dev/packages/init:%s", version), nil
}

```

This would prevent any unnecessary network calls and provide more explicit errors. 

Open to feedback from other maintainers as well.

AustinAbro321:
@brandtkeller I like that suggestion, and I think it could be made even simpler by using the semver package we already use in Zarf 
```go
ver, err := semver.NewVersion(config.CLIVersion)
if err != nil {
  return fmt.Errorf("invalid version %s: %w", version, err)
}
```

brandtkeller (approved):

PR #3571, opened 2025-03-13
fix: change Helm chart dependency verification to "verify never" from "verify if possible"

### LINKED ISSUE #3565: Failed to load keyring
### Environment
Device and OS: Microsoft Windows 10 Enterprise
App version: v0.49.0
Kubernetes distro being used: -
Other: -

### Steps to reproduce
1. Create `ZarfPackageConfig` with a component that use a local helm chart. This local helm chart have a dependency (in my case `cloudnative-pg`).
2. Run the command to list images: `zarf dev find-images`

### Expected result
The images are correctly listed.

### Actual Result
Zarf complains about the GPG keyring.

Console logs:

```
λ zarf dev find-images ./ai-package
2025-03-11 11:16:30 INF using config file location=C:\Users\jrial95\Sources\ai-deployment\zarf-config.toml
2025-03-11 11:16:30 INF using build directory path=./ai-package
2025-03-11 11:16:30 INF processing local helm chart name=ai-platform version=0.1.0 path=components\ai-platform\helm-chart
2025-03-11 11:16:40 ERR unable to find images: unable to package the chart ai-platform: unable to package the "ai-platform" chart: unable to build dependencies for the chart: could not download https://github.com/cloudnative-pg/charts/releases/download/cloudnative-pg-v0.23.1/cloudnative-pg-0.23.1.tgz: failed to load keyring: open C:\Users\202837\.gnupg\pubring.gpg: The system cannot find the file specified.
```

### Severity/Priority
Medium

### Additional Context
None

AustinAbro321:
## Description

When downloading Helm charts directly specified in a `zarf.yaml`  the verify option is set to `verifyNever`, however, when downloading dependent charts Zarf sets the verifier to `verifyIfPossible`. Zarf does not leave a way to for users to set their keyring path besides the environment variable `GNUPGHOME`.

The verify functionality is opt in the Helm CLI and the downloader will fail when `verifyIfPossible` is set and the user doesn't have gpg installed or setup properly. Given this, we should set `verifyIfPossible` to `verifyNever`

If in the future users want the option to verify their charts this could be done through additional flags and we could make the functionality consistent with regular charts and dependent charts.  

## Related Issue

Fixes #3565 

## Checklist before merging


brandtkeller (approved):
Agree with the proposed changes and the summary regarding the potential that this could be an opt-in feature for zarf. It should be complete across the whole lifecycle - not simply dependency charts.

PR #3584, opened 2025-03-18
experimental: allow both Crane and ORAS images as a backend

AustinAbro321:
## Description

This is created as a branch off  #3559 to see what it would look like if we temporarily allowed both Crane and ORAS to work within Zarf by environment variable

## Checklist before merging


src/internal/packager2/images/common.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/zarf-dev/zarf/src/pkg/transform"
+	"github.com/zarf-dev/zarf/src/types"
+)
+
+type pullKeyEnabled struct{}
+
+var defaultPullKeyEnabled = pullKeyEnabled{}
+
+// WithNewClientEnabled allows stores a value to determine whether the new puller is enabled
+func WithNewClientEnabled(ctx context.Context, enabled bool) context.Context {
```
AustinAbro321, line 28: This is how to specify whether or not to use the new client

AustinAbro321:
Closing this as we're opting not to go with this strategy

PR #3380, opened 2025-01-07
chore: switch to new aws resources

AustinAbro321:
## Description

Don't merge! Using this to try out the new AWS account

## Checklist before merging

- [ ] Test, docs, adr added or updated as needed
- [ ] [Contributor Guide Steps](https://github.com/zarf-dev/zarf/blob/main/CONTRIBUTING.md#developer-workflow) followed

PR #3600, opened 2025-03-24
chore: update init s3 publish path

AustinAbro321:
## Description

Update init s3 publish path

This was something I forgot to do in #3380

## Checklist before merging


brandtkeller (approved):
Verified bucket exists - LGTM

PR #3602, opened 2025-03-24
chore: update cosign key to new aws environment

AustinAbro321:
## Description

This updates the cosign key to the key in the new environment, related to #3380. 

This also updates the dos-games package to a new version. Once this PR is in I will publish the 1.2 version of the dos-games package. There will be a follow up PR after this new version is published to update the docs to point to the new version of the dos-games package

## Checklist before merging


main.go
```diff
@@ -12,13 +12,9 @@ import (
 	"syscall"
 
 	"github.com/zarf-dev/zarf/src/cmd"
-	"github.com/zarf-dev/zarf/src/config"
 	"github.com/zarf-dev/zarf/src/pkg/lint"
 )
 
-//go:embed cosign.pub
-var cosignPublicKey string
```
AustinAbro321, line 20: IIRC this was used before the transition from defenseunicorns to zarf-dev to check packages from the defense unicorn GitHub org. We removed this check, but didn't remove the variable, which is currently not doing anything

src/test/e2e/00_use_cli_test.go
```diff
@@ -32,10 +32,10 @@ func TestUseCLI(t *testing.T) {
 		require.Contains(t, stdOut, string(b))
 	})
 
-	t.Run("zarf prepare sha256sum <local>", func(t *testing.T) {
```
mkcp, line 35: Not in scope for this, but should we start deprecating prepare? I was surprised to see it, needed to double check that the alias still existed.

mkcp (approved):
LGTM

src/test/e2e/00_use_cli_test.go
AustinAbro321: Yeah it would make sense to deprecate it. I've been using issues like this for deprecations - https://github.com/zarf-dev/zarf/issues/3420

Issue #3530, opened 2025-02-20
Packager2 FindImages


brandtkeller:
### Describe what should be investigated or refactored
Packager2 `FindImages()`

### Links to any relevant code
[func (p *Packager) FindImages(ctx context.Context) (map[string][]string, error)](https://pkg.go.dev/github.com/defenseunicorns/zarf/src/pkg/packager#Packager.FindImages)

a1994sc:
The link provided links to the Defense Unicorns version of zarf, not the current version:

[func (p *Packager) FindImages(ctx context.Context) (map[string][]string, error)](https://pkg.go.dev/github.com/zarf-dev/zarf/src/pkg/packager#Packager.FindImages)

PR #3609, opened 2025-03-26
refactor: remove statefulness from Helm package

### LINKED ISSUE #3610: Refactor Helm package
1. Most of the functions are pointer receiver methods `Helm` object, which has about 15 fields. The methods are not stateful, or at least don't need to be. They generally don't mutate the Helm object in ways that matter, but since we're using a pointer receiver it's impossible to know if the first call effects the second when two calls are made sequentially.  
2. It's hard to know which fields are required to make a call because there are about 15 fields on the Helm object and several of them are only needed for one method in the package. Additionally there is the field `cfg` of type `*types.PackagerConfig` has about 100 sub fields, only two of these sub fields are actually needed
4. Confusing code flows. Taking a look at the below code would you know when chartOverride would be set? It happens when a Helm chart is created from a Zarf manifest. Because this code happens about a few calls deep from the public function `InstallOrUpgradeChart`, it's not intuitive that this is why that escape hatch is there. Instead, we could have `InstallOrUpgradeChart` accept a `*chart.Chart`. It's an extra function call for the caller, but overall less parameters to `InstallOrUpgradeChart` and much more clear. 
```go
func (h *Helm) loadChartData() (*chart.Chart, chartutil.Values, error) {
	var (
		loadedChart *chart.Chart
		chartValues chartutil.Values
		err         error
	)

	if h.chartOverride == nil {
		// If there is no override, get the chart and values info.
		loadedChart, err = h.loadChartFromTarball()
		if err != nil {
			return nil, nil, fmt.Errorf("unable to load chart tarball: %w", err)
		}

		chartValues, err = h.parseChartValues()
		if err != nil {
			return loadedChart, nil, fmt.Errorf("unable to parse chart values: %w", err)
		}
	} else {
		// Otherwise, use the overrides instead.
		loadedChart = h.chartOverride
		chartValues = h.valuesOverrides
	}

	return loadedChart, chartValues, nil
}
```

https://github.com/zarf-dev/zarf/tree/b7197b23efc25517216c08fb7aa6ab1104869184/src/internal/packager/helm

### Additional context
This refactor should happen before #3530

AustinAbro321:
## Description

The Helm package is hard to work with. I go in detail in the problems here #3610

## Related Issue

fixes #3610
Relates to #3530 

## Checklist before merging


src/internal/packager/helm/chart.go
```diff
@@ -28,74 +31,87 @@ import (
...
 	"github.com/zarf-dev/zarf/src/internal/healthchecks"
+	"github.com/zarf-dev/zarf/src/internal/packager/template"
 	"github.com/zarf-dev/zarf/src/pkg/message"
 	"github.com/zarf-dev/zarf/src/types"
 )
 
 // Use same default as Helm CLI does.
 const maxHelmHistory = 10
 
+type InstallUpgradeOpts struct {
```
mkcp, line 43: opts ++

I'm curious if we can simplify this further by snapshotting both `cluster`and `state`and passing in immutable copies rather than the entire world.

src/internal/packager/helm/chart.go
```diff
@@ -294,128 +248,106 @@ func (h *Helm) UpdateReleaseValues(ctx context.Context, updatedValues map[string
 		return nil
 	}
 
-	return fmt.Errorf("unable to find the %s helm release", h.chart.ReleaseName)
+	return fmt.Errorf("unable to find the %s helm release", chart.ReleaseName)
 }
 
-func (h *Helm) installChart(ctx context.Context, postRender *renderer) (*release.Release, error) {
+func installChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, chart *chart.Chart, chartValues chartutil.Values,
```
mkcp, line 254: Nit: Not always necessary for a private fn, but it's complex enough that it should have a doccomment.

src/internal/packager/helm/chart.go
```diff
@@ -294,128 +248,106 @@ func (h *Helm) UpdateReleaseValues(ctx context.Context, updatedValues map[string
...
-		return nil, fmt.Errorf("unable to load chart data: %w", err)
-	}
-
 	// Perform the loadedChart installation.
-	return client.RunWithContext(ctx, loadedChart, chartValues)
+	return client.RunWithContext(ctx, chart, chartValues)
 }
 
-func (h *Helm) upgradeChart(ctx context.Context, lastRelease *release.Release, postRender *renderer) (*release.Release, error) {
+func upgradeChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, chart *chart.Chart, chartValues chartutil.Values,
```
mkcp, line 283: Nit: doccomment

src/internal/packager/helm/chart.go
```diff
@@ -294,128 +248,106 @@ func (h *Helm) UpdateReleaseValues(ctx context.Context, updatedValues map[string
...
-	}
-
 	// Perform the loadedChart upgrade.
-	return client.RunWithContext(ctx, h.chart.ReleaseName, loadedChart, chartValues)
+	return client.RunWithContext(ctx, zarfChart.ReleaseName, chart, chartValues)
 }
 
-func (h *Helm) rollbackChart(name string, version int) error {
-	client := action.NewRollback(h.actionConfig)
+func rollbackChart(name string, version int, actionConfig *action.Configuration, timeout time.Duration) error {
```
mkcp, line 316: Nit: doccomment

src/internal/packager/helm/chart.go
```diff
@@ -294,128 +248,106 @@ func (h *Helm) UpdateReleaseValues(ctx context.Context, updatedValues map[string
...
-
-		chartValues, err = h.parseChartValues()
-		if err != nil {
-			return loadedChart, nil, fmt.Errorf("unable to parse chart values: %w", err)
-		}
-	} else {
-		// Otherwise, use the overrides instead.
-		loadedChart = h.chartOverride
-		chartValues = h.valuesOverrides
+func LoadChartData(zarfChart v1alpha1.ZarfChart, chartPath string, valuesPath string, valuesOverrides map[string]any) (*chart.Chart, chartutil.Values, error) {
```
mkcp, line 335: Nit: doccomment

src/internal/packager/helm/chart.go
```diff
@@ -294,128 +248,106 @@ func (h *Helm) UpdateReleaseValues(ctx context.Context, updatedValues map[string
...
 
+	chartValues, err := parseChartValues(zarfChart, valuesPath, valuesOverrides)
+	if err != nil {
+		return loadedChart, nil, fmt.Errorf("unable to parse chart values: %w", err)
+	}
 	return loadedChart, chartValues, nil
 }
 
-func (h *Helm) migrateDeprecatedAPIs(ctx context.Context, latestRelease *release.Release) error {
+func migrateDeprecatedAPIs(ctx context.Context, c *cluster.Cluster, actionConfig *action.Configuration, latestRelease *release.Release) error {
```
mkcp, line 348: nit: doccomment

src/internal/packager/helm/chart.go
mkcp: Also, I'm typically in favor of commenting public fields on options structs. Usually I try to explain the capture the intended usage for each field. It can get repetitive though - kind of depends on the style guide for the team. Thoughts?

src/internal/packager/helm/common.go
```diff
@@ -5,84 +5,31 @@
...
-	actionConfig   *action.Configuration
-	variableConfig *variables.VariableConfig
-	state          *types.ZarfState
-}
-
-// Modifier is a function that modifies the Helm config.
-type Modifier func(*Helm)
-
-// New returns a new Helm config struct.
-func New(chart v1alpha1.ZarfChart, chartPath string, valuesPath string, mods ...Modifier) *Helm {
```
mkcp, line 54: Note: Package names that match the type and constructor functions for types that have zero values that can't be used are a good practice. We don't have to bring it back for this PR but it's something to consider in the future as we continue to improve the codebase.

src/internal/packager/helm/post-render.go
```diff
@@ -32,86 +31,71 @@ import (
 )
 
 type renderer struct {
-	*Helm
+	chart v1alpha1.ZarfChart
+
+	adoptExistingResources bool
+	cluster                *cluster.Cluster
```
mkcp, line 37: Does renderer need to mutate `cluster.Cluster`, `types.ZarfState`, `action.Configuration`, and `variables.VariableConfig` or can we pass these in as copies?

src/internal/packager/helm/post-render.go
```diff
@@ -32,86 +31,71 @@ import (
...
 	connectStrings types.ConnectStrings
 	namespaces     map[string]*corev1.Namespace
 }
 
-func (h *Helm) newRenderer(ctx context.Context) (*renderer, error) {
-	rend := &renderer{
-		Helm:           h,
-		connectStrings: types.ConnectStrings{},
-		namespaces:     map[string]*corev1.Namespace{},
+func newRenderer(ctx context.Context, chart v1alpha1.ZarfChart, adoptExistingResources bool, c *cluster.Cluster, airgapMode bool, state *types.ZarfState,
```
mkcp, line 47: nit: doccomment, private function but there's meaningful complexity here

src/internal/packager/helm/post-render.go
```diff
@@ -32,86 +31,71 @@ import (
...
+	} else if rend.adoptExistingResources {
 		namespace.Labels = cluster.AdoptZarfManagedLabels(namespace.Labels)
-		rend.namespaces[h.chart.Namespace] = namespace
+		rend.namespaces[rend.chart.Namespace] = namespace
 	}
 
 	return rend, nil
 }
 
 func (r *renderer) Run(renderedManifests *bytes.Buffer) (*bytes.Buffer, error) {
```
mkcp, line 85: Doccomment

src/internal/packager/helm/repo.go
```diff
@@ -251,7 +252,7 @@ func (h *Helm) DownloadPublishedChart(ctx context.Context, cosignKeyPath string)
 	}
 
 	// Download the file into a temp directory since we don't control what name helm creates here
-	temp := filepath.Join(h.chartPath, "temp")
```
mkcp, line 254: Noticed that we don't have to reach into the `helm` struct to access this chartPath. It's just in lexical scope. I just think that's neat. 😄

src/internal/packager/helm/repo.go
```diff
@@ -379,7 +380,7 @@ func (h *Helm) buildChartDependencies() error {
 	return nil
 }
 
-func (h *Helm) loadAndValidateChart(location string) (loader.ChartLoader, *chart.Chart, error) {
+func loadAndValidateChart(location string) (loader.ChartLoader, *chart.Chart, error) {
```
mkcp, line 383: Note: Does `*chart.Chart` have to be a reference or can we pass back a copy?

Also, I think we can get away without a doccomment here, it's fairly self-documenting

src/internal/packager/helm/template.go
```diff
@@ -0,0 +1,155 @@
...
+
+func newTemplateRenderer(actionConfig *action.Configuration, vc *variables.VariableConfig) (*templateRenderer, error) {
+	rend := &templateRenderer{
+		actionConfig:   actionConfig,
+		variableConfig: vc,
+	}
+	return rend, nil
+}
+
+func (tr *templateRenderer) Run(renderedManifests *bytes.Buffer) (*bytes.Buffer, error) {
```
mkcp, line 108: Nit: doccomment

src/internal/packager/helm/template.go
```diff
@@ -0,0 +1,155 @@
...
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/releaseutil"
+
+	"github.com/zarf-dev/zarf/src/config"
+	"github.com/zarf-dev/zarf/src/pkg/message"
+)
+
+// TemplateChart generates a helm template from a given chart.
+func TemplateChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, chart *chart.Chart, values chartutil.Values,
```
mkcp, line 31: Clean 👍

src/internal/packager/helm/utils.go
```diff
@@ -1,87 +0,0 @@
-// SPDX-License-Identifier: Apache-2.0
```
mkcp, line 1: migrating `utils` functions closer to the data ++

src/internal/packager/helm/zarf.go
```diff
@@ -61,20 +62,20 @@ func (h *Helm) UpdateZarfRegistryValues(ctx context.Context) error {
...
 
 // UpdateZarfAgentValues updates the Zarf agent deployment with the new state values
-func (h *Helm) UpdateZarfAgentValues(ctx context.Context) error {
+func UpdateZarfAgentValues(ctx context.Context, opts InstallUpgradeOpts) error {
 	l := logger.From(ctx)
 	spinner := message.NewProgressSpinner("Gathering information to update Zarf Agent TLS")
 	defer spinner.Stop()
 
-	deployment, err := h.cluster.Clientset.AppsV1().Deployments(cluster.ZarfNamespaceName).Get(ctx, "agent-hook", metav1.GetOptions{})
+	deployment, err := opts.Cluster.Clientset.AppsV1().Deployments(cluster.ZarfNamespaceName).Get(ctx, "agent-hook", metav1.GetOptions{})
```
mkcp, line 78: FUTURE: This is a lot of .'s and mutation but not something we need to fix right away.

src/internal/packager2/layout/package.go
```diff
@@ -326,11 +328,7 @@ func validatePackageIntegrity(pkgLayout *PackageLayout, isPartial bool) error {
 	}
 
 	if len(packageFiles) > 0 {
-		// TODO (phillebaba): Replace with maps.Keys after upgrading to Go 1.23.
-		filePaths := []string{}
-		for k := range packageFiles {
-			filePaths = append(filePaths, k)
-		}
+		filePaths := slices.Collect(maps.Keys(packageFiles))
```
mkcp, line 331: TIL `maps.Keys(...)`

src/pkg/packager/prepare.go
```diff
@@ -158,18 +158,11 @@ func (p *Packager) findImages(ctx context.Context) (map[string][]string, error)
 		resources := []*unstructured.Unstructured{}
 		matchedImages := map[string]bool{}
 		maybeImages := map[string]bool{}
-		for _, chart := range component.Charts {
+		for _, zarfChart := range component.Charts {
```
mkcp, line 161: Q: Why zarfChart here? We're within zarf, so I would expect it to be first class and helmChart to gets prefixed.

mkcp (commented):
Comments within. Some nits and some design considerations

src/internal/packager/helm/repo.go
AustinAbro321: Technically it does not, but since Helm expects to receive a *chart.Chart it feels weird to dereference then reference again

src/internal/packager/helm/chart.go
AustinAbro321: I'm 50/50. I'm also okay with not commenting every field, but leaning towards comments. Cluster for example, feels pretty obvious in the context, it's a struct that can really only be used to connect to a k8s cluster. I like the way you framed capturing the intended use of the fields. I added comments with that in mind.

src/internal/packager/helm/chart.go
AustinAbro321: Added doccomments to many of the functions, but I'm struggling to think of a comment for install / upgrade / rollback that would be a value add. These functions are really just setting the fields for the Helm install / upgrade / rollback function then executing it.

mkcp: Fair, that in itself can be a code smell indicating a leaky abstraction but it's not something that has to be addressed in this PR given the scope outlined in the issue.

src/internal/packager/helm/chart.go
mkcp: Re: immutable data: This was intended as an architectural consideration for the future, no need to balloon scope on this PR to address is now. I'll make that more clear in my comments next time 😄

Re: doccomments: Nice, yeah that strikes a good balance and aligns well with the guidance in the `Types` section here: https://tip.golang.org/doc/comment#Types

PR #3611, opened 2025-03-26
chore: update dos-games references to use new version

AustinAbro321:
## Description

Published a new package signed with the new key here - https://github.com/zarf-dev/zarf/actions/runs/14089306390

This updates the code to use that new package

## Checklist before merging


brandtkeller (approved):
LGTM - tested locally without issue. I don't see any missed tags besides the `1.0.0-$arch` reference in `11_oci_pull_inspect_test.go` which I imagine is supposed to stay unchanged.

AustinAbro321:
@brandtkeller yea it's a regression test for the old OCI publish format

PR #3614, opened 2025-03-27
fix(inspect): only pull required files for inspect on OCI packages

### LINKED ISSUE #3572: Zarf package inspect unnecessarily pulls down package
### Environment
Device and OS:
N/A
App version:
0.48.1
Kubernetes distro being used:
N/A
Other:
Expected behavior observed in zarf version 0.32.2

### Steps to reproduce
1. Run zarf apckage inspect against a package stored in OCI For example:
`zarf package inspect oci://ghcr.io/defenseunicorns/packages/uds/gitlab:17.9.1-uds.1-upstream` 

### Expected result
zarf displays the package config without downloading the entire package

### Actual Result
zarf pulls the entire package locally and then displays the pacakge config

Actual behavior:

Image

Note: The above screenshot shows 0.46.0, but i also tested against `0.48.1` and the behavior is the same. 

Expected behavior: 

Image

### Severity/Priority
Medium

### Additional Context
This is a regression from the expected behavior, which makes these commands far less usable as packages can be very large so getting the output of the command can take several minutes rather than a couple seconds. Unsure the exact zarf version where the behavior changed. The issue impacts all of the inspect commands including retrieving the sboms.

brandtkeller:
## Description

Fixing the ability to inspect packages stored in OCI without needing to pull the full package. 

Originally I had sought to only retrieve the required artifacts from the package in the OCI registry - but for maintaining the package signature validation requirements - opted to stick to the common patterns with the packager2 `layout` and writing/loading from disk. 

This currently adds an `Inspect` option to `Load` and defaults to retrieving all "required" artifacts from the package. This includes the zarf.yaml, checksums, signature and sbom. This could be further improved to only retrieve the SBOM when required- but I chose to minimize the inspect logic cascading too far. 

Ideally much of this could be fetched/stored in memory - I think this would be an improvement to consider if we look at decoupling the validation logic from being a disk activity. 

Overall opting to stick to a strategy that aligns with current load operations such that all packages on disk are not impacted and new logic simply accounts for the OCI nuance. 

## Related Issue

Fixes #3572

Relates to #

## Checklist before merging


src/internal/packager2/layout/package.go
```diff
@@ -101,9 +102,12 @@ func LoadFromDir(ctx context.Context, dirPath string, opt PackageLayoutOptions)
...
 	}
-	err = validatePackageIntegrity(pkgLayout, opt.IsPartial)
-	if err != nil {
-		return nil, err
+	// do not validate integrity if inspecting, as all files may not be available.
+	if !opt.Inspect {
+		err = validatePackageIntegrity(pkgLayout, opt.IsPartial)
+		if err != nil {
+			return nil, err
+		}
```
AustinAbro321, line 110: Would `validatePackageIntegrity` work for inspect if opt.IsPartial is set to true?

src/internal/packager2/load.go
```diff
@@ -53,6 +54,25 @@ func LoadPackage(ctx context.Context, opt LoadOptions) (*layout.PackageLayout, e
...
+				SkipSignatureValidation: opt.SkipSignatureValidation,
+				IsPartial:               isPartial,
+				Inspect:                 true,
+			}
+			pkgLayout, err := layout.LoadFromDir(ctx, path, layoutOpt)
+			if err != nil {
+				return nil, err
+			}
+			return pkgLayout, nil
+		}
```
AustinAbro321, line 75: Opinion based, but I think it would be better to keep the inspect logic in a separate function. The package you get back from inspect is going to be quite different than any other package you get back (no components) so it feels a bit out of the scope of the function to pick it up. 

Keeping it in it's own function would probably also make it simpler to pull or not pull the SBOM

AustinAbro321 (commented):
A few comments on potential simplifications, otherwise the core logic seems to be good.

src/internal/packager2/layout/package.go
brandtkeller: I will give that a shot. If memory serves me right - I believe the SHA comparison between the aggregate checksum and the checksums file won't match as this isn't a complete package.

brandtkeller:
Closing as out-dated - continuing work in another branch

PR #3617, opened 2025-03-28
fix: support importing components with multiple charts

willswire:
## Description

When composing components, if the imported component has more than one chart, Zarf is unable to load the chart because it looks for `.zarf-cache/oci/dirs/abcdf...wxyz/chart` instead of `.zarf-cache/oci/dirs/abcdf...wxyz/charts`.

## Related Issue

Fixes #

Relates to #

## Checklist before merging


AustinAbro321:
@willswire Thanks for the PR! Could you make an issue with a simple way to reproduce this bug

willswire:
Closed via https://github.com/zarf-dev/zarf/issues/3623

PR #3618, opened 2025-03-30
fix(mirror): improve ux of mirror with state fallback

### LINKED ISSUE #2389: zarf package mirror-resources should either default to the zarf registry / zarf git server or require either --registry-url or --git-url
Running `zarf package mirror-resources` without any CLI options may be confusing to users as the error message looks like this currently. 
registry: 
```bash 
 WARNING  Retrying (1/3) in 5s: GET
          https://index.docker.io/v2/stefanprodan/podinfo/manifests/6.4.0-zarf-2985051089:
          MANIFEST_UNKNOWN: manifest unknown; unknown tag=6.4.0-zarf-298505108
```
git repo:
```bash
 WARNING  Retrying (1/3) in 5s: failed to push the git repo "zarf-public-test-1143224168": unable to
          push repo to the gitops service: repository not found
```
Users might expect that we default to the zarf registry / server or they may think something is wrong with their package or images. Likely we should either default to the zarf registry and zarf git server or fail if those are not set and they are using them 

- **Given** I have a package that has images
- **When** I run `zarf package mirror-resources zarf-package-my-package` without specifying the --registry-url
- **Then** I get an error message "zarf package mirror-resources requires registry to be set to mirror images`
OR
- **Given** I have a package that has images
- **When** I run `zarf package mirror-resources zarf-package-my-package` without specifying the --registry-url
- **Then** the images are uploaded to the zarf registry

Similar for the git repo use case but repos instead of images and --git-url rather than --registry-url 

### Additional context
I am leaning towards implementing both. First checking if state exists and using it if it does. If state does not exist, fall back to requiring flags be used.

brandtkeller:
## Description

WIP - took some creative liberties with looking at the SDK and general utility in order to potential allow for more resilient workflows. 

## Related Issue

Fixes #2389

Relates to #

## Checklist before merging

### DIFF
--- a/site/src/content/docs/commands/zarf_package_mirror-resources.md
+++ b/site/src/content/docs/commands/zarf_package_mirror-resources.md
@@ -52,10 +52,12 @@ $ zarf package mirror-resources <your-package.tar.zst> \
       --git-push-username string        Username to access to the git server Zarf is configured to use. User must be able to create repositories via 'git push' (default "zarf-git-user")
       --git-url string                  External git server url to use for this Zarf cluster
   -h, --help                            help for mirror-resources
+      --images                          mirror only the images
       --no-img-checksum                 Turns off the addition of a checksum to image tags (as would be used by the Zarf Agent) while mirroring images.
       --registry-push-password string   Password for the push-user to connect to the registry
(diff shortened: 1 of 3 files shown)


brandtkeller:
Closing as stale - working to simplify

PR #3624, opened 2025-04-01
fix: set flavor on composed component

AustinAbro321:
## Description

A bug was introduced in #3597 which caused validation to improperly fail when a flavor is used by a child component but not used by a parent component. It stems from the fact that flavors are not set on the composed component when a child component has one. This PR updates the behavior to set the flavor on the newly composed

## Checklist before merging

### DIFF
--- a/src/internal/packager2/layout/import.go
+++ b/src/internal/packager2/layout/import.go
@@ -139,7 +139,6 @@ func resolveImports(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath, 
 	pkg.Constants = slices.CompactFunc(constants, func(l, r v1alpha1.Constant) bool {
 		return l.Name == r.Name
 	})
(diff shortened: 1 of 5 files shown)


src/internal/packager2/layout/import.go
```diff
@@ -139,7 +139,6 @@ func resolveImports(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath,
 	pkg.Constants = slices.CompactFunc(constants, func(l, r v1alpha1.Constant) bool {
 		return l.Name == r.Name
 	})
-	importStack = importStack[0 : len(importStack)-1]
```
AustinAbro321, line 142: Linter pointed out that this doesn't actually do anything

brandtkeller (approved):
LGTM - visual diff can be easily seen with `zarf dev inspect definition`

PR #3636, opened 2025-04-04
feat!: Add debug granularity to package create validation and update error message

mkcp:
## Description
This PR adds some debug logging and the package name to resolving and validating imports. We also append the package name to a flavor validation error - the error message is otherwise unchanged.

Before:
Screenshot 2025-04-04 at 14 08 18

After:
Screenshot 2025-04-04 at 14 02 16

## Related Issue
#3597 

## Checklist before merging


brandtkeller (approved):
I like this pattern. Given the `debug` logging i am not concerned about the UX of logs that may contain empty values in the key/value pairs (IE flavor).

mkcp:
Yeah, and in this case as well as many others in go there's signal in using empty value. There's different philosophies on zero / empty values in programming but golang tends towards making the zero value useful.

https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful

PR #3643, opened 2025-04-08
fix: oci-concurrency flag on init

AustinAbro321:
## Description

As a part of #3559 the oci-concurrency flag was made to work with container images, however the language was not changed to reflect this and it was never added to `zarf init`.

## Checklist before merging


mkcp (approved):
LGTM

brandtkeller (approved):

Issue #3342, opened 2024-12-17
Option to keep tmp directory during zarf package deploy


MiraDSaan:
### Is your feature request related to a problem? Please describe.
I was building a zarf package and my values file had an error.  There was an extra quote in one of the files.
The zarf error message refered to a file in the zarf temporary directory, however the file was deleted when zarf completed, so I was given a link to a file that did not exist.  The error message is not helpful in finding the error.

I tried using the following command in hopes the temp directory would be saved, but the tmp directory was still deleted.

```
zarf package deploy -l=debug --tmpdir=/tmp/test1/zarfpackage.tar.zst --confirm
...
#3: unable to load chart data: unable to parse chart values: failed to parse
             /tmp/test1/zarf-4146961286/components/chart/values/xxx-0.0.3-0: error converting YAML
             to JSON: yaml: line 15: found unexpected end of stream: 
```

The file `/tmp/test1/zarf-4146961286/components/chart/values/xxx-0.0.3-0` no longer exists, thereby making the error message useless.

### Describe the behavior you'd like

- **Given** I have a zarf package 
- **When** I run `zarf package deploy` and an error occurs in which zarf prints the name of the file in the error message
- **Then** zarf does NOT delete the temp directory which contains the file listed in the error message.

### Describe alternatives you've considered
(optional) A clear and concise description of any alternative solutions or features you've considered.

Two alternatives I thought of:

- **Given** I have a zarf package 
- **When** I run `zarf package deploy --save-tmpdir` 
- **Then** zarf does NOT delete the temp directory.

- **Given** I have a zarf package 
- **When** I use the `tmpdir` flag during `zarf package deploy --tmpdir=/tmp/mydir/` 
- **Then** zarf does NOT delete the temp directory which contains the file listed in the error message.

phillebaba:
I think that the underlying problem you are trying to solve to debug the output will be solved by https://github.com/zarf-dev/proposals/pull/9. Adding the ability to show manifest and value output would catch these issues.

Issue #2631, opened 2024-06-17
Feature request: Zarf dev package templating


zack-is-cool:
### Is your feature request related to a problem? Please describe.
It would be beneficial if zarf had the ability to template out what would be in a package at create time and also at deploy time with input variables. It is sometimes difficult to determine what will happen with a complex zarf package with component imports, overrides etc.

### Describe the solution you'd like

- **Given** I'm creating a zarf package
- **When** I type `zarf dev package template create` (or something like this) with `--set` variables or referencing a zarf.config
- **Then** Zarf will 
	- template and print the Zarf package with templated `###ZARF_PKG_TMPL`s rendered
	- template and print the values.yaml files with `###ZARF_VAR`s- defaults used if none provided by the user at create time
	- template and print all manifests that the helm charts would create with `###ZARF_VAR`s - defaults used if none provided by the user at create time
- **When** I type `zarf dev package template deploy`  (or something like this) with `--set` variables or referencing a zarf.config
- **Then** Zarf will 
  - template and print the Zarf package with templated `###ZARF_PKG_TMPL`s rendered
  - template and print the values.yaml files with `###ZARF_VAR`s- possibly defaults used
  -  template and print all manifests that the helm charts would create with `###ZARF_VAR`s - possibly defaults used if none provided by the user at deploy time

### Additional context
I feel like this workflow eases the friction of needing to actually have a kubernetes cluster up and running and allows you to debug zarf package templating logic before applying anything to a cluster - similar to the way `helm template` works, but with the additional zarf templating on top of it.. This makes the feedback loop smaller when debugging a zarf package that is failing to deploy.

JoeHCQ1:
I'll second this feature request. Not having an easy way to see the rendered yaml also made diagnosing this bug a bit tricky: https://github.com/defenseunicorns/zarf/issues/2692

Note that if the problem is in the values.yaml and it makes the yaml invalid, then a `ctrl-c` mid attempt to make it work will preserve the rendered yaml in your /tmp directory and you can view it there.

This command is also said to result in the full manifests because 'kind' is in all manifests:

```bash
zarf dev find-images . --flavor upstream --why "kind" > templated.txt
```

Final note: this zarf functionality should also be expanded into UDS since UDS is frequently the entrypoint to the package creation and deployment commands.

AustinAbro321:
[...]

Issue #3625, opened 2025-04-01
Add zarf label to all resources created by a package deploy


sgettys:
### Clear labeling of Zarf created resources
I want to be able to query k8s api and get a list of resources back that were created by deploying a zarf package. Something like `zarf-created-by=<package-name>` could be added to the resource labels.

### Describe the behavior you'd like
1. 
- **Given** a Zarf package
- **When** It is deployed
- **Then** All Kubernetes resources that are created by the deployment are labeled with the package name

2. 
- **Given** a Zarf package is deployed
- **When** `kubectl get <resource> -n <namespace> -l zarf-created-by=<pacakge-name>` is executed
- **Then** all resources created in the requested namespace by the specified package are returned

AustinAbro321:
I agree this would be a good feature. This label wouldn't fit neatly into any of the [Kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/), so a custom route is the way to go. 

We try to keep the `zarf.dev/` prefix on any of the custom labels that we add. IMO a good label would be `zarf.dev/package: my-package`.

Feel free to create a PR for this. You'll have to add the label to the post-renderer, IMO you'll also want to add the label to the namespace. I don't think it's necessary to add the label to the secrets the package creates.

PR #3653, opened 2025-04-10
feat: zarf package inspect manifests

AustinAbro321:
## Description

This implement `zarf package inspect manifests` from the rework inspect ZEP - https://github.com/zarf-dev/proposals/tree/main/0008-rework-inspect

## Related Issue

Relates to #2631

Fixes: #3342

## Checklist before merging


src/cmd/package.go
```diff
@@ -62,7 +62,20 @@ func newPackageCommand() *cobra.Command {
 	return cmd
 }
 
-type packageCreateOptions struct{}
+type packageCreateOptions struct {
```
AustinAbro321, line 65: We want to move to this strategy overall, I am doing it in create for this PR so that I can write unit tests at the CLI level rather than e2e tests

src/cmd/package.go
```diff
@@ -392,14 +413,86 @@ func (o *packageInspectOptions) run(cmd *cobra.Command, args []string) error {
 	return definitionOpts.run(cmd, args)
 }
 
-// PackageInspectSBOMOptions holds the command-line options for 'package inspect sbom' sub-command.
-type PackageInspectSBOMOptions struct {
+type packageInspectManifestsOpts struct {
+	skipSignatureValidation bool
+	components              string
```
brandtkeller, line 418: Interesting to see that the `components` flag is simply a string versus a slice of strings. Looks to be an existing pattern - but surprised me nonetheless given the `StringSliceVar` flag options. (No change required)

src/internal/packager2/inspect.go
```diff
@@ -0,0 +1,144 @@
...
+				helmChart, values, err := helm.LoadChartData(chart, chartDir, valuesDir, chartOverrides)
+				if err != nil {
+					return PackageInspectManifestResults{}, fmt.Errorf("failed to load chart data: %w", err)
+				}
+				chartTemplate, err := helm.TemplateChart(ctx, chart, helmChart, values, opts.KubeVersion, variableConfig)
+				if err != nil {
+					return PackageInspectManifestResults{}, fmt.Errorf("could not render the Helm template for chart %s: %w", chart.Name, err)
+				}
+				resources = append(resources, Resource{
+					Content:      chartTemplate,
```
brandtkeller, line 106: In my testing - I am seeing the `---` separator on helm charts missing a new line and therefor being merged with the content of the last resource of the last line. 
```suggestion
					Content:      fmt.Sprintf("%s\n", chartTemplate),,
```

example: 
```
        env:
        - name: LONGHORN_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
      restartPolicy: Never
      serviceAccountName: longhorn-service-account---
```

brandtkeller (changes requested):

src/cmd/package.go
```diff
@@ -62,7 +62,20 @@ func newPackageCommand() *cobra.Command {
 	return cmd
 }
 
-type packageCreateOptions struct{}
+type packageCreateOptions struct {
+	confirm                 bool
+	output                  string
+	differentialPackagePath string
+	setVariables            map[string]string
```
mkcp, line 69: Not in scope for this PR, but good opportunity in the future to declare a type for `setVariables`'s `map[string]string`. e.g. `setVariables map[variable.Key]variable.Value` `setVariables variable.Variables` or another alternative

src/cmd/package.go
```diff
@@ -62,7 +62,20 @@ func newPackageCommand() *cobra.Command {
...
+	sbom                    bool
+	sbomOutput              string
+	skipSBOM                bool
+	maxPackageSizeMB        int
+	registryOverrides       map[string]string
```
mkcp, line 74: See comment above, `map[string]string` is usually an indication that there's a compound type we can name and codify.

src/internal/packager2/inspect.go
AustinAbro321: Great catch!

src/cmd/package.go
AustinAbro321: Yeah and it seems like we can actually use StringSliceVar without it being a breaking change. If I were to guess the name reason it's a slice is because you can do globbing - `Globbing component names with '*' and deselecting 'default' components with a leading '-' are also supported`.

AustinAbro321: I guess either way even if you are using globbing it's still a list, trying to decide whether or not I should make an issue for this

brandtkeller: It's quite superfluous - I don't think there is any criticality so I'd argue no.

brandtkeller (commented):
I don't see an issue with the current logic. 

My only outstanding question is if you see any intent on supporting a deployed package?

Currently `images` and `definition` work on deployed packages. `sbom` could work if we delivered the SBOM to the cluster. Is our intent to continue that pattern and make inspect functional on deployed packages or are we beginning a pivot/split?

AustinAbro321:
@brandtkeller I believe it is not currently possible to know which resources are deployed by a Zarf package, but #3625 would give that ability. Once that's implemented we can evaluate if that is functionality we'd want to add. Helm already has the command [helm get manifest](https://helm.sh/docs/helm/helm_get_manifest/) and since we deploy everything through Helm it's possible to do this for manifests as well. 

Still, maybe the feature would have value for a user who knows they have charts deployed, but doesn't know which package deployed them. Personally, I'd like to see the request come in first, but I think it'd a reasonable feature to add.

brandtkeller:
@AustinAbro321 I think that is a fair constraint. My interest is solely with the story telling here. IS package meant to be applicable for packages on disk _AND_ deployed packages as well - or is it a case-by-case? 

Not something we need answered but if the answer is the former, then we should track it.

brandtkeller (approved):

PR #3658, opened 2025-04-11
feat: zarf dev inspect manifest

AustinAbro321:
## Description

This implement `zarf dev inspect manifests` from the rework inspect ZEP - https://github.com/zarf-dev/proposals/tree/main/0008-rework-inspect

## Related Issue

Relates to https://github.com/zarf-dev/zarf/issues/2631

## Checklist before merging


site/src/content/docs/commands/zarf_dev_inspect.md
```diff
@@ -33,4 +33,5 @@ Commands to get information about a Zarf package using a `zarf.yaml`
 
 * [zarf dev](/commands/zarf_dev/)	 - Commands useful for developing packages
 * [zarf dev inspect definition](/commands/zarf_dev_inspect_definition/)	 - Displays the fully rendered package definition
+* [zarf dev inspect manifests](/commands/zarf_dev_inspect_manifests/)	 - Template and output all manifests and charts referenced by the package definition
```
mkcp, line 36: 👍

src/internal/packager2/inspect.go
```diff
@@ -142,3 +145,178 @@ func PackageInspectManifests(ctx context.Context, pkgLayout *layout2.PackageLayo
...
+	Flavor             string
+	KubeVersion        string
+}
+
+type DevInspectManifestResults struct {
+	Resources []Resource
+}
+
+// DevInspectManifests returns manifests and Helm chart manifests after templating
+func DevInspectManifests(ctx context.Context, packagePath string, opts DevInspectManifestsOptions) (DevInspectManifestResults, error) {
```
mkcp, line 161: I'm not 100% on the API here. `Dev...` feels like a `cmd` concerns not `packager2` concerns, and `Package...` [L42](https://github.com/zarf-dev/zarf/pull/3658/files#diff-f12af45cbf249f1786d50fc2924a79099e8e712de195520443fe8bfa1b60a53bR42) repeats the go pkg name in a way. Thoughts on introducing an `inspect/` subdirectory and pkg in `internal/packager2`?

src/internal/packager2/inspect.go
```diff
@@ -142,3 +145,178 @@ func PackageInspectManifests(ctx context.Context, pkgLayout *layout2.PackageLayo
...
+	}
+	variableConfig := template.GetZarfVariableConfig(ctx)
+	variableConfig.SetConstants(pkg.Constants)
+	variableConfig.PopulateVariables(pkg.Variables, opts.DeploySetVariables)
+
+	tmpPackagePath, err := utils.MakeTempDir(config.CommonOptions.TempDirectory)
+	if err != nil {
+		return DevInspectManifestResults{}, err
+	}
+	defer os.RemoveAll(tmpPackagePath)
```
mkcp, line 178: 💭 Future concern but is it bad if we silently fail cleanup here? My expectation is that this should be fine by default in CI, and capturing this error might add flakiness. Maybe we add a `tmp.IsCleanupFrom(ctx)` configuration flag or something

src/cmd/package_test.go
```diff
@@ -189,6 +196,10 @@ func TestPackageInspectManifests(t *testing.T) {
 			}
 			packagePath := filepath.Join(tmpdir, fmt.Sprintf("zarf-package-%s-%s.tar.zst", tc.packageName, config.GetArch()))
 			err = opts.run(context.Background(), []string{packagePath})
+			if tc.expectedErr != "" {
+				require.EqualError(t, err, tc.expectedErr)
```
mkcp, line 200: nit: `require.ErrorContains` or alternative, we want to avoid tests that act as change detectors. https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html

src/internal/packager2/inspect.go
```diff
@@ -142,3 +145,178 @@ func PackageInspectManifests(ctx context.Context, pkgLayout *layout2.PackageLayo
...
+			Content:      string(content),
+			Name:         manifest,
+			ResourceType: ManifestResource,
+		})
+	}
+	return resources, nil
+}
+
+// getTemplatedChart returns a templated chart.yaml as a string after templating
+func getTemplatedChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, packagePath string, baseComponentDir string, variableConfig *variables.VariableConfig, kubeVersion string) (Resource, chartutil.Values, error) {
```
mkcp, line 270: good refactor 👍

mkcp (changes requested):
API concern and a nit. Nonblocking idea about tmpdir cleanup and CI

src/internal/packager2/inspect.go
AustinAbro321: In this case it was just something I forgot to do, but I believe we error on cleanup in other parts of the code. I think I'll go back today and turn the linter back on in packager2/ and fix situations like this.

mkcp: sgtm

src/internal/packager2/inspect.go
AustinAbro321: Good callout. ATM I don't see that strong of a reason to separate inspect from the rest of packager2, it share code / functionality to find images with something like `dev inspect manifests`, but `package inspect manifest` does share code with pull so I'm a bit weary to introduce dependency complexity.

Another alternative for the API concerns could be `InspectDefinitionManifests` and `InspectPackageManifests` rather than the current `PackageInspectManifests` and `DevInspectManifests`.

mkcp (approved):
LGTM

PR #3662, opened 2025-04-11
fix(publish): cleanup trailing manifest file

### LINKED ISSUE #3661: Publish creates a file with the package name
### Steps to reproduce
1. create a package `zarf package create examples/helm-charts`
2. publish said package

### Expected result

Package is published - no temporary or new files are created

### Actual Result

a file is being created with the package name

brandtkeller:
## Description

The `file.store` push operation that is handled during the ORAS logic creates a manifest file on the local filesystem. We can clean this up so that it is not hanging around after each publish.

## Related Issue

Fixes #3661 

## Checklist before merging

### DIFF
--- a/src/internal/packager2/layout/oci.go
+++ b/src/internal/packager2/layout/oci.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"os"
 	"sort"
 	"strings"
 	"time"
@@ -103,11 +104,13 @@ func (r *Remote) Push(ctx context.Context, pkgLayout *PackageLayout, concurrency
 	if err != nil {
 		return err
 	}
-	// here is where the created annotation is added
+	// here is where the manifest is created and written to the filesystem given the file.store Push() functionality
 	root, err := r.orasRemote.PackAndTagManifest(ctx, src, descs, manifestConfigDesc, annotations)
 	if err != nil {
 		return err
 	}
+	// remove this file after the push is complete
+	defer os.Remove(pkgLayout.Pkg.Metadata.Name)
 
 	copyOpts := r.orasRemote.GetDefaultCopyOpts()
 	copyOpts.Concurrency = concurrency


src/internal/packager2/layout/oci.go
```diff
@@ -103,12 +104,19 @@ func (r *Remote) Push(ctx context.Context, pkgLayout *PackageLayout, concurrency
...
 	root, err := r.orasRemote.PackAndTagManifest(ctx, src, descs, manifestConfigDesc, annotations)
 	if err != nil {
 		return err
 	}
 
+	defer func() {
+		// remove the dangling manifest file created by the PackAndTagManifest
+		// should this behavior change, we should expect this to begin producing an error
+		err2 := os.Remove(pkgLayout.Pkg.Metadata.Name)
+		err = errors.Join(err, err2)
```
mkcp, line 117: When using this pattern, we need to return `err` on [line 132](https://github.com/zarf-dev/zarf/pull/3662/files#diff-cb2deb33e1627365d3674b6c90a45161a6e4380d5759eb248672e7ac8b25d470R132) even on the happy path. There's a chance for us to exit normally without an error in the function body, then for an error to occur in the defer step. Upon giving this another look, there's actually a subtle preexisting bug on [line 70](https://github.com/zarf-dev/zarf/pull/3662/files#diff-cb2deb33e1627365d3674b6c90a45161a6e4380d5759eb248672e7ac8b25d470R69-R70]) because we return `nil` at the end of the function. Returning `err` at the end fixes both!

mkcp (changes requested):
Small bugfix in err var return but otherwise good to go!

src/internal/packager2/layout/oci.go
AustinAbro321: Just tried it out and it seems that the error is returned either way
image

mkcp: Oh that's really interesting, I guess err=nil in the return from the func gets overwritten by defer err! Nice catch

mkcp (approved):

src/internal/packager2/layout/oci.go
AustinAbro321: Though that's with named parameters (which aren't in this function) it won't work either way

AustinAbro321 (approved):
LGTM

PR #3678, opened 2025-04-15
feat: create helm chart for zarf-agent

### LINKED ISSUE #3620: zarf-agent custom tolerations
It seems the [zarf-registry](https://github.com/zarf-dev/zarf/blob/main/packages/zarf-registry) package supports setting custom tolerations, but there's no such ability with the [zarf-agent](https://github.com/zarf-dev/zarf/blob/main/packages/zarf-agent), especially because it's a set of manifests instead of a Helm chart. I'm currently trying to deploy Zarf to a boostrapping node that has a taint on it to only allow the pods necessary to get the cluster started. I intended to deploy Karpenter after zarf-init, but if non-tainted nodes need to exist for zarf-agent to work, I'm a bit stuck.

Is there a reason zarf-agent is a bunch of manifests instead of a Helm chart? If it was a chart, I could set an override.

zarf-agent configurable tolerations (helm charts)
- Given deploying zarf-init to a bootstrapping node.
- When I try to deploy karpenter after a zarf-init, there are non-tainted nodes needed for zarf-agent to work. I am trying to deploy zarf to a bootstrapping node that has a taint to only allow pods neccesary to get the cluster started.
- Then zarf-agent won't deploy.

custom init package

JeffResc:
## Description
Converts the zarf-agent manifests to a helm chart, allowing users to customize various values such as affinity and tolerations on the deployment.

## Related Issue

Fixes #3620

## Checklist before merging


packages/zarf-agent/chart/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+description: Zarf agent
+name: zarf-agent
+version: 1.0.0
```
brandtkeller, line 4: This _might_ align more to the zarf version
```suggestion
version: 0.51.0
```

JeffResc: I was matching [what was done with zarf-registry](https://github.com/zarf-dev/zarf/blob/19329edfb1b3e35515896920d39b9f6780dafc6d/packages/zarf-registry/chart/Chart.yaml#L4), but I agree that zarf-agent is more highly coupled with the actual zarf version

brandtkeller (changes requested):
I believe we're missing a baseline `values.yaml` for the chart (living beside the `Chart.yaml`) that would serve as a default values definition that `agent-values.yaml` could then overlay at the Zarf component level.

JeffResc:
> I believe we're missing a baseline `values.yaml` for the chart (living beside the `Chart.yaml`) that would serve as a default values definition that `agent-values.yaml` could then overlay at the Zarf component level.

I'm not sure there's a need to have any other default values outside of the ones provided by `agent-values.yaml`. Would an acceptable solution be to move the `agent-values.yaml` into the `chart/values.yaml` and remove the `valuesFiles` reference from `zarf.yaml`?

brandtkeller:
> I'm not sure there's a need to have any other default values outside of the ones provided by `agent-values.yaml`. Would an acceptable solution be to move the `agent-values.yaml` into the `chart/values.yaml` and remove the `valuesFiles` reference from `zarf.yaml`?

Speaking strictly from a helm chart correctness perspective - the current state will fail `helm lint` . Much like the `zarf-registry` package providing a valid chart which we then have values modified at the zarf component layer - I'd opt to follow the same pattern.

JeffResc:
Yeah, I agree. There should definitely be a baseline `values.yaml` in the `chart/` directory that I missed. Latest commit moves the values into that file and removes the overrides done at the Zarf package level.

packages/zarf-agent/zarf.yaml
```diff
@@ -19,19 +19,12 @@ components:
     required: true
     images:
       - "###ZARF_PKG_TMPL_AGENT_IMAGE_DOMAIN######ZARF_PKG_TMPL_AGENT_IMAGE###:###ZARF_PKG_TMPL_AGENT_IMAGE_TAG###"
-    manifests:
+    charts:
       - name: zarf-agent
+        releaseName: zarf-agent
```
brandtkeller, line 24: Bit of a hack - but if we modify this - given that the raw chart name has not changed since creation and that the hasher logic is quite old as well [link](https://github.com/zarf-dev/zarf/blob/main/src/internal/packager/helm/common.go#L38-L42) then we should have quite substantial backwards compatibility AND it should function now. 
```suggestion
        releaseName: zarf-d2db14ef40305397791454e883b26fc94ad9615d
```

I say we start here to begin - test other E2E functionality while continuing to get feedback on this as the solution.

brandtkeller:
Looks like some differences between the old/new resources with regards to formatting and tests written to validate an expected state. Let me know if you need assistance running them down.

brandtkeller:
Spent a few cycles digging and found that the problem isn't that there is a difference in the expected values - rather a lack of difference when performing the `update-creds` operation due to the agent release not being found and no error present in that case.

brandtkeller:
All checks passing - Believe this is now in a state where we can review the feasibility of the "workaround". I intend to think more about other options that might enable backwards compatibility without needing to keep specific conditional logic.

packages/zarf-agent/chart/Chart.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+description: Zarf agent
+name: zarf-agent
+version: 0.51.0
```
brandtkeller: we may consider if we care what this is initially set to?

brandtkeller: Changed to `0.1.0` for initial pass

JeffResc:
> Side note, this says it fixes https://github.com/zarf-dev/zarf/issues/3620, but doesn't add a variable for tolerations

While it's not exposed as a Zarf variable, there is now a helm value for tolerations, which should meet the use case @fingermustache and I were looking for. I'd also be happy to do a follow-up PR that maps the helm values to Zarf variables to make configuring them easier

PR #3682, opened 2025-04-16
refactor(cluster,state,types)!:  migrate types.ZarfState to state.State 

mkcp:
## Description
The goal of this PR is to apply golang best-practices to Zarf's resource tracking in a k8s cluster and eventually enable Zarf to handle more complex cluster operations.

Major changes are:
- Migrate `types.ZarfState` to `pkg/state/{state,state_test}.go`
- Add `state.MergeOptions` and simplify `state.Merge` params
- Add `cluster.InitStateOptions` and update `cluster.InitState` to take the new type.
- Fix boundaries between `cluster.Cluster` and ZarfState. Cluster method names have been updated and simplified. Some of these changes are breaking, if necessary we can provide backwards compatibility with a deprecation warnings that alias to the new method names.
- Migrate cluster constants and credential keys from`message` the `pkg/state`. The final home for some of these may be `pkg/cluster`, but we'll see. (Update: a follow-up issue for the remaining methods lives [here](https://github.com/zarf-dev/zarf/issues/3719))

Minor fixes:
- Remove redundant naming like "Zarf" showing up in method and type names. This is not fully complete, many of the cluster constants still contain the Zarf prefix (which makes sense because it reflects the prefix in the k8s secret store)

## Related Issue
Relates to #2709

## Checklist before merging


mkcp:
Github workflows are being weird and stuck on "expected", seeing if closing and reopening forces the workflows to rerun.

src/cmd/package.go
```diff
@@ -663,11 +663,11 @@ func (o *packageInspectImagesOptions) run(cmd *cobra.Command, args []string) err
 		return err
 	}
 
-	// The user may be pulling the package from the cluster or using a built package
+	// The user may be pulling the package from the c or using a built package
```
AustinAbro321, line 666: accidental find and replace

src/cmd/package.go
```diff
@@ -717,11 +717,11 @@ func (o *packageInspectDefinitionOptions) run(cmd *cobra.Command, args []string)
...
```
AustinAbro321, line 720: accidental find and replace

src/cmd/zarf_tools.go
```diff
@@ -102,23 +102,23 @@ func (o *getCredsOptions) complete(ctx context.Context) error {
...
-	state, err := o.cluster.LoadZarfState(ctx)
+	s, err := o.cluster.Load(ctx)
 	if err != nil {
 		return err
 	}
 	// TODO: Determine if this is actually needed.
-	if state.Distro == "" {
-		return errors.New("zarf state secret did not load properly")
+	if s.Distro == "" {
+		return errors.New("zarf s secret did not load properly")
```
AustinAbro321, line 111: accidental find and replace

src/internal/agent/hooks/flux-helmrepo_test.go
```diff
@@ -175,7 +176,7 @@ func TestFluxHelmMutationWebhook(t *testing.T) {
 			code: http.StatusOK,
 		},
 		{
-			name: "should not mutate URL if it has the same hostname as Zarf state internal repo",
+			name: "should not mutate URL if it has the same hostname as Zarf s internal repo",
```
AustinAbro321, line 179: accidental find and replace

src/internal/agent/hooks/flux-helmrepo_test.go
```diff
@@ -146,7 +147,7 @@ func TestFluxHelmMutationWebhook(t *testing.T) {
 			code: http.StatusOK,
 		},
 		{
-			name: "should not mutate URL if it has the same hostname as Zarf state",
+			name: "should not mutate URL if it has the same hostname as Zarf s",
```
AustinAbro321, line 150: accidental find and replace

src/pkg/cluster/cluster.go
```diff
@@ -35,9 +42,22 @@ const (
...
+}
+
+// NewWithWait wraps NewClusterWithWait to provide a shorthand alias.
+func NewWithWait(ctx context.Context) (*Cluster, error) {
+	return NewClusterWithWait(ctx)
+}
+
+// New wraps NewCluster to provide a shorthand alias.
+func New(ctx context.Context) (*Cluster, error) {
+	return NewCluster(ctx)
```
AustinAbro321, line 60: While I understand the strategy to avoid breaking changes, I personally don't think we're at the point where we should do this for the cluster package. It's unlikely that more than a few projects consume these functions. Also the parameters of `NewCluster` are changing anyway so the user would still have to update for a breaking change.

src/pkg/cluster/cluster.go
```diff
@@ -134,3 +154,204 @@ func WatcherForConfig(cfg *rest.Config) (watcher.StatusWatcher, error) {
...
+	// Information about the container registry Zarf is going to be using
+	RegistryInfo types.RegistryInfo
+	// Information about the artifact registry Zarf is going to be using
+	ArtifactServer types.ArtifactServerInfo
+	// StorageClass of the k8s cluster Zarf is initializing
+	StorageClass string
+}
+
+// Init takes initOptions and hydrates a cluster's state from InitOptions.
+func (c *Cluster) Init(ctx context.Context, opts InitOptions) error {
```
AustinAbro321, line 173: IMO `Init`, `Load`, and `Save` should change to `InitState`, `LoadState`, and `SaveState`. Given that the cluster package deals with several different data types, it won't be super intuitive for users to know that these functions are dealing with Zarf state. 

For example, a user might be confused if the `cluster.New()` and `cluster.Save()` deal with separate data types.

src/pkg/message/credentials.go
AustinAbro321, line 1: All of the functions in this file can be deleted, they are replaced by functionality in the new logger.

AustinAbro321 (changes requested):
Some comments about naming, overall a good refactor.

src/pkg/cluster/cluster.go
mkcp: Yeah that is fair, and I agree that re-adding `State` into the cluster methods that return a `State` is a better API than what exists in the PR now. In future iterations I wonder whether we can't simplify `cluster` even more - `New`, `InitState`, and `LoadState` all feel like similar operations... not sure yet, much to consider.

src/pkg/cluster/cluster.go
mkcp: I agree with that, updated in latest to go with `cluster.New()`

src/pkg/message/credentials.go
mkcp: I agree but feel like this is out of scope for this PR. IMO, doing this well would mean implementing LogValuer for slog fields that need to be redacted, which is sets a new pattern in `pkg/logger`. I've scoped out a follow-up issue here https://github.com/zarf-dev/zarf/issues/3719.

src/internal/packager/helm/common.go
```diff
@@ -88,7 +88,7 @@ func loadChartFromTarball(chart v1alpha1.ZarfChart, chartPath string) (*chart.Ch
 	// Get the path the temporary helm chart tarball
 	sourceFile := StandardName(chartPath, chart) + ".tgz"
 
-	// Load the loadedChart tarball
+	// LoadState the loadedChart tarball
```
AustinAbro321, line 91: There's several comments where Load changes to LoadState from a find and replace

mkcp: Oh man, I keep doing this 🤣 I try to scan through the diff visually before commit, but it's still easy to miss stuff. Fixed in latest

src/pkg/message/credentials.go
AustinAbro321: perfect

PR #3713, opened 2025-04-21
fix: ensure variable uniqueness during compose

### LINKED ISSUE #3707: zarf duplicates variables/constants in output
### Environment

Device and OS: MBP, macOS - deploying on remote EKS
App version: latest, 0.52.1
Other: Also experienced in GitHub CI on k3d

### Steps to reproduce

1. Deploy a package, specifically one with imported components.
2. Observe the output, looking at variables/constants.

### Expected result

Accurate list of variables/constants without duplicates.

### Actual Result

Many variables/constants are duplicated (or more). It appears this may be based on usage of `import`s in the zarf packages - output seems to include the variables/constants per component, so components imported from the same zarf package have their variables duplicated. Of particular annoyance is that `currently set to` only shows on the first instance of the var, making the UX bad to identify specific vars and what they are set to.

Example with zarf init constants:

```console
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
constants:  static values set by the package author

- name: REGISTRY_IMAGE
  value: library/registry

- name: REGISTRY_IMAGE_TAG
  value: 2.8.3

- name: REGISTRY_IMAGE
  value: library/registry

- name: REGISTRY_IMAGE_TAG
  value: 2.8.3

- name: REGISTRY_IMAGE
  value: library/registry

- name: REGISTRY_IMAGE_TAG
  value: 2.8.3

- name: AGENT_IMAGE
  value: zarf-dev/zarf/agent

- name: AGENT_IMAGE_TAG
  value: v0.51.0

- name: GITEA_IMAGE
  value: gitea/gitea:1.21.5-rootless
```

Also can be viewed in zarf's own CI: https://github.com/zarf-dev/zarf/actions/runs/14536071354/job/40784555858#step:7:256 (note the variables are the worst, with some vars being 3-4x).

### Severity/Priority

Low/Medium - poor UX but functionality is all unaffected.

AustinAbro321:
## Description

There is a bug in compose that causes duplicates variables to appear. `slices.CompatFunc` was used to de-duplicate variables, however since it only compares consecutive elements there were a lot of opportunities for misses. 

## Related Issue

Fixes #3707

## Checklist before merging

### DIFF
--- a/src/internal/packager2/layout/import.go
+++ b/src/internal/packager2/layout/import.go
@@ -148,7 +148,6 @@ func resolveImports(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath, 
 
 	pkg.Components = components
 	// Keep only first occurrence of each variable
-	pkg.Variables = []v1alpha1.InteractiveVariable{}
 	for _, v := range variables {
 		if slices.IndexFunc(pkg.Variables, func(existing v1alpha1.InteractiveVariable) bool {
 			return existing.Name == v.Name
@@ -158,7 +157,6 @@ func resolveImports(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath, 
 	}
 
 	// Keep only first occurrence of each constant
-	pkg.Constants = []v1alpha1.Constant{}
 	for _, c := range constants {
 		if slices.IndexFunc(pkg.Constants, func(existing v1alpha1.Constant) bool {
 			return existing.Name == c.Name
--- a/src/internal/packager2/layout/import.go
+++ b/src/internal/packager2/layout/import.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"slices"
 	"time"
 
 	"github.com/zarf-dev/zarf/src/pkg/logger"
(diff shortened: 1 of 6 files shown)


src/internal/packager2/layout/import.go
```diff
@@ -147,12 +147,23 @@ func resolveImports(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath,
...
 	pkg.Components = components
-	pkg.Variables = slices.CompactFunc(variables, func(l, r v1alpha1.InteractiveVariable) bool {
-		return l.Name == r.Name
-	})
-	pkg.Constants = slices.CompactFunc(constants, func(l, r v1alpha1.Constant) bool {
-		return l.Name == r.Name
-	})
+	// Keep only first occurrence of each variable
+	for _, v := range variables {
+		if slices.IndexFunc(pkg.Variables, func(existing v1alpha1.InteractiveVariable) bool {
```
brandtkeller, line 152: Agree with your assessment - the variables and constants are not used at scale to be a determinant consideration. 

For fun - at fist glance I want to reach for a `map[string]bool` here for uniqueness which I believe would O(n)?

brandtkeller (approved):

src/internal/packager2/layout/import.go
AustinAbro321: Yeah I originally was going to do that, but since maps are unordered, I didn't want the zarf.yaml to not be fully reproducible

brandtkeller: The map would only be for uniqueness checks though yes? Storage of the variable would still be appended to the `pkg.variables` I believe?

mkcp (approved):
LGTM - I like that you got a fix out quickly and are exploring functional data-driven patterns here.

In a future PR, let's see how we can codify the intended behavior in unit tests so we can fail fast and cheaply. That's probably in scope for an explicit tech debt cleanup PR though!

src/internal/packager2/layout/import.go
AustinAbro321: You are correct, I just updated it to use maps instead

PR #3683, opened 2025-04-16
feat: allow unused flavors with a warning

AustinAbro321:
## Description

In #3597 I added validation for flavors, however I did not anticipate that users intentionally created packages with unused flavors to set different tags on the package. As we don't have an alternative at the moment, this PR allows unused flavors again, but with a warning

## Checklist before merging

- [ ] Test, docs, adr added or updated as needed
- [ ] [Contributor Guide Steps](https://github.com/zarf-dev/zarf/blob/main/CONTRIBUTING.md#developer-workflow) followed

PR #3714, opened 2025-04-21
fix: stop warning users when flavors are not used

AustinAbro321:
## Description

In #3683 I implemented flavors with a warning if a flavor is not used. However there was a mistake in the logic so it warns even when flavor aren't used

## Checklist before merging


brandtkeller (approved):
LGTM - tested a few variations and the warning feels appropriate.

PR #3717, opened 2025-04-22
feat: use websockets for port forwards

### LINKED ISSUE #3715: Zarf should move to websockets for port-forward behavior
As of kubernetes 1.31, any streaming traffic is moving from SPDY to websockets. To keep in line with kubernetes and to possibly solve the issue of image pushes failing when using aws-ssm, zarf's tunnel for pushing images should attempt to use websockets as well.

- **Given** a cluster with kubernetes 1.31 or later
- **When** an image is pushed to the zarf registry
- **Then** websockets is used as the base protocol for the tunneling, rather than SPDY

### Additional context
[Blog post on the transition](https://kubernetes.io/blog/2024/08/20/websockets-transition/#why-change-the-streaming-protocol)

AustinAbro321:
## Description

This updates the tunnel package to use the Kubernetes portforward websocket feature. This should resolve long standing issues in the community that makes Zarf incompatible with tools like aws ssm as the current port forward method does not work well with proxies or gateways. 

## Related Issue

Fixes #3715

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/tunnel.go
+++ b/src/pkg/cluster/tunnel.go
@@ -21,7 +21,6 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/portforward"
 	"k8s.io/client-go/transport/spdy"
-	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 
 	"github.com/avast/retry-go/v4"
 	"github.com/defenseunicorns/pkg/helpers/v2"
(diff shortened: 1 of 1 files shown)


src/pkg/cluster/tunnel.go
```diff
@@ -530,3 +530,23 @@ func (tunnel *Tunnel) getAttachablePodForService(ctx context.Context) (string, e
...
 }
+
+// taken from https://github.com/kubernetes/kubernetes/blob/680ea07dbb2c6050d13b93660fa4d27d2d28d6eb/staging/src/k8s.io/kubectl/pkg/cmd/portforward/portforward.go#L139-L156
+func createDialer(method string, url *url.URL, config *rest.Config) (httpstream.Dialer, error) {
+	transport, upgrader, err := spdy.RoundTripperFor(config)
+	if err != nil {
+		return nil, err
+	}
+	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: transport}, method, url)
+	if !cmdutil.PortForwardWebsockets.IsDisabled() {
```
soltysh, line 541: I'd drop this if, since you always want to have that full fallback mechanism enabled. At least for quite a few future releases until k8s decides entirely to drop that.

AustinAbro321: Yeah I wasn't sure if there was some niche use case where a user would still want to force the old port forwarder, assuming not I'll go ahead and remove this

brandtkeller (approved):
LGTM - good find

soltysh (approved):
lgtm

PR #3718, opened 2025-04-22
feat: add retries to image push

AustinAbro321:
## Description

In #3559 we moved from Crane to ORAS. As part of that move, the retry logic was removed on image push.  ORAS does have builtin retries but these do not account for failures on the port-forward. For example, an HPA may spin down the Zarf registry and break, these situations can only be saved by retrying and creating another tunnel. By adding back in retries at the Zarf level, we can account for these situations. 

To manually test a pod failure, I restarted the Zarf registry mid push. I verified that I got a failure doing this on current Zarf, and verified that this will recover when trying with the change in this PR
image

## Checklist before merging

### DIFF
--- a/src/internal/packager/images/push.go
+++ b/src/internal/packager/images/push.go
@@ -96,8 +96,8 @@ func Push(ctx context.Context, cfg PushConfig) error {
 		return copyImage(ctx, src, remoteRepo, srcName, dstName, cfg.OCIConcurrency, defaultPlatform)
 	}
 
-	err = retry.Do(func() error {
-		for _, img := range cfg.ImageList {
+	for _, img := range cfg.ImageList {
+		err = retry.Do(func() error {
 			l.Info("pushing image", "name", img.Reference)
 			// If this is not a no checksum image push it for use with the Zarf agent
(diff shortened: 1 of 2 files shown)


src/internal/packager/images/push.go
```diff
@@ -68,53 +49,98 @@ func Push(ctx context.Context, cfg PushConfig) error {
...
+			var err error
+			plainHTTP, err = shouldUsePlainHTTP(ctx, registryURL, client)
 			if err != nil {
 				return err
 			}
+		}
 
-			if err = pushImage(img.Reference, offlineNameCRC); err != nil {
-				return err
+		pushImage := func(srcName, dstName string) error {
```
mkcp, line 86: Declaring this as a private package func is also a valid approach, and a good way to test how portable logic in your function is (e.g. how easy it is to change should requirements change in the future). However, the params get much more complicated without using the parent func's scope, so this feels like a reasonable tradeoff here. IMO this is about the upper limit of how complex a locally scoped function variable should be.

mkcp (approved):
LGTM, nice work 💯

Issue #3435, opened 2025-01-27
FluxCD helm release fails when using ocirepo resource as chart reference


a1994sc:
### Environment
App version: `v0.47.0`
Kubernetes distro being used: `kind`, `rke2`

### Steps to reproduce
1. Create files in the `Additional Context` section
2. Package those files and deploy to a zarf cluster with fluxcd installed

### Expected result

The `helm-controller` installs the `0.13.0` version of the `external-secrets` helm-release

### Actual Result

Helm throws a version mismatch error because of the `-zarf-` and crc32 checksum

### Visual Proof (screenshots, videos, text, etc)

```text
artifact revision 0.13.0-zarf-3677108702 does not match chart version 0.13.0
```

### Resolution

I think that `zarf` should allow the ability to remove the crc32 checksum from resources when a certain annotation is provided:
`zarf.dev/remove-checksum`

### Additional Context
`kustomization.yml`
```yaml
---
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: external-secrets
resources:
  - namespace.yml
  - helm-release.yml
  - oci-repo.yml
configMapGenerator:
  - name: hr-values-external-secrets
    files:
      - values.yaml
    options:
      disableNameSuffixHash: true
```
`helm-release.yml`
```yaml
---
apiVersion: helm.toolkit.fluxcd.io/v2
kind: HelmRelease
metadata:
  name: external-secrets
spec:
  interval: 1m
  chartRef:
    kind: OCIRepository
    name: ghcr-io-external-secrets
  valuesFrom:
    - kind: ConfigMap
      name: hr-values-external-secrets
      optional: false
```
`namespace.yml`
```yaml
---
apiVersion: v1
kind: Namespace
metadata:
  name: external-secrets
```
`oci-repo.yml`
```yaml
---
apiVersion: source.toolkit.fluxcd.io/v1beta2
kind: OCIRepository
metadata:
  name: ghcr-io-external-secrets
spec:
  interval: 12h
  layerSelector:
    mediaType: "application/vnd.cncf.helm.chart.content.v1.tar+gzip"
    operation: copy
  url: oci://ghcr.io/external-secrets/charts/external-secrets
  ref:
    tag: 0.13.0
```
`values.yaml`
```yaml
---
# yaml-language-server: $schema=https://raw.githubusercontent.com/external-secrets/external-secrets/refs/heads/main/deploy/charts/external-secrets/values.schema.json
[...]

PR #3720, opened 2025-04-23
feat: disable crc32 hashing on helm charts for resource `ocirepo`

a1994sc:
## Description

I have brought up using an annotation in #3436 however there was some concern about keeping the agent simple, which makes sense to me. So I'm not sure that is will be any better but I did want see what the maintainers and community thought about checking the `MediaType` of the `LayerSelector` (if present) and not preforming the hashing if it is a known helm type.

### Additional comments

When I brought this up during a community meeting earlier in the year, we thought it might be a possible idea to query the internal registry and look at the media-type in there; I think that it might be a bit more reliable to check against the available meta-data versus making a network call and holding up the admission process

## Related Issue

Relates to #3435

## Checklist before merging

### DIFF
--- a/src/internal/agent/hooks/flux-ocirepo.go
+++ b/src/internal/agent/hooks/flux-ocirepo.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/defenseunicorns/pkg/helpers/v2"
 	"github.com/fluxcd/pkg/apis/meta"
@@ -97,7 +98,18 @@ func mutateOCIRepo(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster
 			patchedURL = fmt.Sprintf("%s:%s", patchedURL, src.Spec.Reference.Tag)
 		}
 
-		patchedSrc, err := transform.ImageTransformHost(registryAddress, patchedURL)
+		var (
+			patchedSrc string
+			err        error
+		)
+
+		// Do not include a crc32 hash if the MediaType starts with known helm selector
+		if (src.Spec.LayerSelector != nil) && strings.HasPrefix(src.Spec.LayerSelector.MediaType, "application/vnd.cncf.helm.chart.content.v1.tar+gzip") {
+			patchedSrc, err = transform.ImageTransformHostWithoutChecksum(registryAddress, patchedURL)
+		} else {
+			patchedSrc, err = transform.ImageTransformHost(registryAddress, patchedURL)
+		}
+
 		if err != nil {
 			return nil, fmt.Errorf("unable to transform the OCIRepo URL: %w", err)
 		}
--- a/src/internal/agent/hooks/flux-ocirepo_test.go
+++ b/src/internal/agent/hooks/flux-ocirepo_test.go
@@ -119,6 +119,44 @@ func TestFluxOCIMutationWebhook(t *testing.T) {
 			code: http.StatusOK,
 		},
 		{
+			name: "mute helm chart, but don't include crc32 hash",
+			admissionReq: createFluxOCIRepoAdmissionRequest(t, v1.Update, &flux.OCIRepository{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "mutate-this",
+				},
+				Spec: flux.OCIRepositorySpec{
+					URL: "oci://ghcr.io/stefanprodan/charts/podinfo",
+					LayerSelector: &flux.OCILayerSelector{
+						MediaType: "application/vnd.cncf.helm.config.v1+json",
+					},
+					Reference: &flux.OCIRepositoryRef{
+						Tag: "6.4.0",
+					},
+				},
+			}),
+			patch: []operations.PatchOperation{
+				operations.ReplacePatchOperation(
+					"/spec/url",
+					"oci://127.0.0.1:31999/stefanprodan/charts/podinfo",
+				),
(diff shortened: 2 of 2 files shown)


src/internal/agent/hooks/flux-ocirepo.go
```diff
@@ -97,7 +98,18 @@ func mutateOCIRepo(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster
...
+			patchedSrc string
+			err        error
+		)
+
+		// Do not include a crc32 hash if the MediaType starts with known helm selector
+		if (src.Spec.LayerSelector != nil) && strings.HasPrefix(src.Spec.LayerSelector.MediaType, "application/vnd.cncf.helm.config") {
+			patchedSrc, err = transform.ImageTransformHostWithoutChecksum(registryAddress, patchedURL)
+		} else {
+			patchedSrc, err = transform.ImageTransformHost(registryAddress, patchedURL)
+		}
```
brandtkeller, line 111: Using the metadata on hand feels appropriate. I see why reaching out to the registry to determine this `MediaType` might feel more magical but this does prevent an additional network call. If we document this requirement explicitly then it provides a path forward. 

A more magical solution would not be a breaking change. Open to thoughts from others,

AustinAbro321: For now, I would prefer this solution. The issue it is that the layer selector field is optional, but I do prefer this to a request to to the registry

src/internal/agent/hooks/flux-ocirepo.go
```diff
@@ -97,7 +98,18 @@ func mutateOCIRepo(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster
...
 		}
 
-		patchedSrc, err := transform.ImageTransformHost(registryAddress, patchedURL)
+		var (
+			patchedSrc string
+			err        error
+		)
+
+		// Do not include a crc32 hash if the MediaType starts with known helm selector
+		if (src.Spec.LayerSelector != nil) && strings.HasPrefix(src.Spec.LayerSelector.MediaType, "application/vnd.cncf.helm.config") {
```
AustinAbro321, line 107: From searching github it looks like most of the time the string is `application/vnd.cncf.helm.chart.content.v1.tar+gzip`. Could you change the prefix you're searching for to `application/vnd.cncf.helm`

src/internal/agent/hooks/flux-ocirepo.go
a1994sc: Hm, for some reason I thought that you needed to specify the `.spec.layerSelector` when dealing with `ocirepo` and helm charts... Might have been when they first came out, but flux matured enough to be smart about which layer it uses?

However, now that I know it is optional when dealing with helm charts, I do agree that it is a little clunky...

I kinda hate to retread the same ground, but I do that this would be a cleared up with using a label or annotation instead:

- It would be opt-in, so it won't effect existing work-loads
- Could be applied to any other resource type, `pod`, `gitrepo`, `application`, etc. as a overall pattern.
- Is also a kubernetes native way to handle mutations, e.i. istio

src/internal/agent/hooks/flux-ocirepo.go
a1994sc: I will update the PR with this comment

src/internal/agent/hooks/flux-ocirepo.go
a1994sc: @AustinAbro321 / @brandtkeller  do you have any thoughts about using the labels as it is less "magic" as the package author will be declare what needs to include the crc32 hash or not?

brandtkeller: I would love to see an implementation and compare them side-by-side if possible.

a1994sc: In #3436 I used an annotation, but the logic would be similar

a1994sc: @brandtkeller I created another using the label for the mutation logic:

https://github.com/zarf-dev/zarf/pull/3789

AustinAbro321: I am going to be put some more thought into this later this week. My initial instinct is that we should avoid adding an annotation as it's something that we'll have to support forever, and if we are going to add it we'll likely want to add it to all resources effected by the agent. It'll be confusing if an annotation works for one resource and not another. 

I'd like to evaluate if it's possible for an upstream solution. If not, I think checking the registry with a fallback to defaulting to using the crc checksum would be reasonable.

a1994sc: Well I do understand the overhead part of this, I personally think that is a choose by the `fluxcd` team to check the `.annotations."org.opencontainers.image.version"` when using the `.spec.ref.tag` in the ocirepo resource.

I think that another possible solution would when we push the helm charts, we could edit the annotation to be the same as the crc32 hashed tag. That might be "cleaner" as it would not effect any of the mutation logic

a1994sc: ok, so I think that flux checks the `Chart.yaml` version to determine what is the "correct" version of the helm chart.

In pr #3800 I tested to see if changing the annotation to include the crc32 hashing how flux is still giving the same error:

> artifact revision 20.0.2-zarf-2681845985 does not match chart version 20.0.2

I think we need a label to disable the hashing or to remove the crc32 hashing....

PR #3738, opened 2025-04-25
refactor: deploy

### LINKED ISSUE #3528: Packager2 Deploy
Packager2 `Deploy()` 

[func (p *Packager) Deploy(ctx context.Context) error](https://pkg.go.dev/github.com/defenseunicorns/zarf/src/pkg/packager#Packager.Deploy)

### Additional context
See #3408

AustinAbro321:
## Description

Behavior changes: 
- Previously SBOMs would always be placed in a temporary directory even if the `--confirm` flag was used. I feel it's unlikely  that a person who desires to view the SBOMs would deploy with `--confirm` then try to look at the SBOMs quickly before the package is finished deployed and the SBOMs are erased. 
- Previously Zarf would warn if the version of the init package that the cluster was initialized with was less than v0.26.0. Given that this was released on April 19th, 2023 and we have not added any additional warnings, this is not called in `packager2.deploy`

## Related Issue

Fixes #3528 

## Checklist before merging


src/test/e2e/31_checksum_and_signature_test.go
```diff
@@ -36,7 +36,7 @@ func TestChecksumAndSignature(t *testing.T) {
 	// Test that we get an error when trying to deploy a package without providing the public key
 	stdOut, stdErr, err = e2e.Zarf(t, "package", "deploy", pkgName, "--confirm")
 	require.Error(t, err, stdOut, stdErr)
-	require.Contains(t, stdErr, "failed to deploy package: unable to load the package: package is signed but no key was provided - add a key with the --key flag or use the --skip-signature-validation flag and run the command again")
+	require.Contains(t, stdErr, "unable to load package: package is signed but no key was provided")
```
AustinAbro321, line 39: We'll have to decide how descriptive we want error messages to be on deploy. I would say it's reasonably intuitive to look at flags given the new message. A reasonable counter would be that Zarf is constructed in such a way that it can be deployed by an inexperienced person, though the typical deployer is likely more often a pipeline nowadays.

Do we consider the original message overkill or should it be included for more clarity. I lean to keep the error message smaller, but I can see it either way.

mkcp: I'm leaning towards the smaller error message approach that you mentioned and that's in the current branch. 

For this particular test assert, checking that an error occurred is plenty. The test case and inputs are clear and we get no additional information or test coverage by requiring a error string.

AustinAbro321: IMO for e2e tests it's worth considering checking stderr for a specific string. There is a fair amount that could go wrong during deploy, so this guards against the test failing for some other reason

src/internal/packager2/deploy.go
```diff
@@ -0,0 +1,704 @@
...
+		if component.RequiresCluster() {
+			timeout := cluster.DefaultTimeout
+			if pkgLayout.Pkg.IsInitConfig() {
+				timeout = 5 * time.Minute
+			}
+			connectCtx, cancel := context.WithTimeout(ctx, timeout)
+			defer cancel()
+			if !d.isConnectedToCluster() {
+				var err error
+				d.c, err = cluster.NewWithWait(ctx)
```
brandtkeller, line 129: Should this be utilizing `connectCtx`? 
```suggestion
				d.c, err = cluster.NewWithWait(connectCtx)
```

src/internal/packager2/deploy.go
```diff
@@ -0,0 +1,704 @@
...
+
+	// Always init the state before the first component that requires the cluster (on most deployments, the zarf-seed-registry)
+	if component.RequiresCluster() && d.s == nil {
+		applianceMode := false
+		for _, component := range pkgLayout.Pkg.Components {
+			if component.Name == "k3s" {
+				applianceMode = true
+			}
+		}
+		err := d.c.InitState(ctx, cluster.InitStateOptions{
```
brandtkeller, line 226: Noticing that `StorageClass` is not utilized here and there is no option in `DeployOpts` in order to propagate use of the `--storage-class` init flag. Is this intentional?

src/internal/packager2/deploy.go
```diff
@@ -0,0 +1,704 @@
...
+			}
+			connectCtx, cancel := context.WithTimeout(ctx, timeout)
+			defer cancel()
+			if !d.isConnectedToCluster() {
+				var err error
+				d.c, err = cluster.NewWithWait(ctx)
+				if err != nil {
+					return nil, err
+				}
+				if err := d.verifyPackageIsDeployable(connectCtx, pkgLayout.Pkg); err != nil {
```
brandtkeller, line 133: Appreciate that this is further surfaced

brandtkeller (changes requested):
Minor items of feedback

src/internal/packager2/deploy.go
AustinAbro321: Yes, really great catch

src/cmd/package.go
```diff
@@ -225,30 +226,155 @@ func (o *packageDeployOptions) preRun(_ *cobra.Command, _ []string) {
...
+func confirmDeploy(ctx context.Context, pkgLayout *layout2.PackageLayout, setVariables map[string]string) (err error) {
+	l := logger.From(ctx)
+
+	err = utils.ColorPrintYAML(pkgLayout.Pkg, getPackageYAMLHints(pkgLayout.Pkg, setVariables), true)
+	if err != nil {
+		return fmt.Errorf("unable to print package definition: %w", err)
 	}
+
+	if pkgLayout.Pkg.IsSBOMAble() && !pkgLayout.ContainsSBOM() {
+		l.Warn("this package does NOT contain an SBOM.  If you require an SBOM, please contact the creator of this package to request a version that includes an SBOM.",
```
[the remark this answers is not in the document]
AustinAbro321: An SBOM also won't exist if images and files aren't used in a package, but that's tracked by the `IsSBOMAble`. So yeah this log is somewhat outdated from a time where Zarf assumed that the deployer and creator were always different. Will update.

brandtkeller (dismissed):
Thinking about implicit vs explicit options use and what we get for free in some of our other packages global options. 

Nonetheless I did not find an area of feature parity not being met. This work side-by-side with the older logic is a great point of comparison. 

Great work!

src/cmd/initialize.go
```diff
@@ -117,18 +112,39 @@ func (o *initOptions) run(cmd *cobra.Command, _ []string) error {
 	pkgConfig.PkgOpts.SetVariables = helpers.TransformAndMergeMap(
 		v.GetStringMapString(VPkgDeploySet), pkgConfig.PkgOpts.SetVariables, strings.ToUpper)
 
-	pkgClient, err := packager.New(&pkgConfig, packager.WithSource(src), packager.WithContext(ctx))
-	if err != nil {
-		return err
+	loadOpt := packager2.LoadOptions{
```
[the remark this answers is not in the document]
AustinAbro321: Yeah I definitely agree. I am going to add config here. I think it would even be worth making architecture required for OCI pulls instead of setting it in LoadPackage.

brandtkeller (approved):
LGTM

brandtkeller (dismissed):

mkcp (approved):
updates LGTM

PR #3739, opened 2025-04-25
test: improve archive test

Noxsios:
## Description

Cuts down `TestUseCLI/zarf_package_to_test_archive_path` from 60s to 3s by not using the eks distro package (100Mb) as the test package. Instead a test package is created with similarly structured data. The eks distro package is still built and deployed in nightly.

## Checklist before merging

### DIFF
--- a/src/test/packages/00-archive-path/zarf.yaml
+++ b/src/test/packages/00-archive-path/zarf.yaml
@@ -4,7 +4,6 @@ metadata:
   name: archive-path
 
 x-untar: &base-untar
-  name: untar
   required: true
   files:
     - source: archive.tar
@@ -27,14 +26,14 @@ x-untar: &base-untar
           mute: true
 
 components:
-  - only:
+  - name: untar-linux
+    only:
       localOS: linux
     <<: *base-untar
-    name: untar-linux
-  - only:
+  - name: untar-darwin
+    only:
       localOS: darwin
     <<: *base-untar
-    name: untar-darwin
   - name: untar-windows
     required: true
     only:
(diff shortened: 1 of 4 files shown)


src/test/packages/00-archive-path/zarf.yaml
Noxsios, line 1: I can put this into the test itself, but it was a good exercise seeing what needed to change between *nix and windows

src/test/packages/00-archive-path/zarf.yaml
```diff
@@ -0,0 +1,59 @@
...
+      onFailure:
+        - cmd: rm -f archive/content/* archive.tar
+      after:
+        - cmd: rm -f archive/content/* archive.tar && rmdir archive/content && rmdir archive
+          mute: true
+
+components:
+  - name: untar
+    only:
+      flavor: linux
```
AustinAbro321, line 31: We should use `only.localOS` here

src/test/packages/00-archive-path/zarf.yaml
AustinAbro321, line 1: nit: could we rename the `t.Run` and this file to extract path over archive path to make it more clear what we're testing

AustinAbro321: I'd prefer creation + cleanup in go as I think it's easier to read / maintain, but I won't block on it

AustinAbro321 (changes requested):
It is great that we are making a specific package for the feature under test here, and it's a pattern we should use always going forward. A few small notes, otherwise looks good

src/test/packages/00-archive-path/zarf.yaml
Noxsios: I had it there before but for some reason it wasn't applying on my windows machine, ill try again

PR #3742, opened 2025-04-28
test: avoid dockerhub flake in agent ignore test

AustinAbro321:
## Description

I've frequently noticed a flake happening around our agent ignore test. https://github.com/zarf-dev/zarf/actions/runs/14672162347/job/41181212190. The reason for this flake is that Kuberentes is trying to pull from dockerhub, but is getting rate limited. By switching to ghcr we can avoid the rate limit

## Checklist before merging

### DIFF
--- a/src/test/packages/26-agent-ignore/manifests/deployment.yaml
+++ b/src/test/packages/26-agent-ignore/manifests/deployment.yaml
@@ -1,20 +1,20 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: httpd-deployment
+  name: podinfo-deployment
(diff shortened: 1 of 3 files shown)


brandtkeller (dismissed):
LGTM - not worth a change but if we ever need to trim some weight for image pulls (on create or in this instance deploy) we can also use the alpine image in our org packages.

AustinAbro321:
@brandtkeller I ended up using this suggestion. I like the pattern of using our own images and I think this is more explicit. I added another tag that way the comment on the image would still be true.

brandtkeller (approved):

PR #3743, opened 2025-04-28
feat: remove user submitted digest from OCI pull log

### LINKED ISSUE #3622: Remove digest field from logs if --shasum is not set

When you run `zarf package create` with the `--shasum` flag you should see something like `digest=some-digest` in your logs, but when you run `zarf package create` without `--shasum` you just see `digest=` with no value. I believe it would be helpful to remove the empty digest field if you dont pass shasum. Seeing the empty digest field in the logs indicates to me that something is wrong when that isn't actually the case.

AustinAbro321:
## Description

Agree with below issue that the empty digest log can be confusing. I could have added to the log with conditionals, but given that the shasum is strictly user input and is more traditionally given on oci objects with the `@sha256:..` format I believe it's simpler to remove. 

For example a user could run
```bash
starting pull from oci source src=oci://ghcr.io/austinabro321/helm-charts:0.0.1@sha256:bce489695b38ac0e6c3660c8fbc8498aee9d2180db9887d62d79fd3c7508e00c digest=
```

They are using a digest in this situation, but since it's on the OCI object it looks like it's empty.

I kept the log on `http(s)` pulls as the digest is required 

## Related Issue

Fixes #3622

## Checklist before merging

### DIFF
--- a/src/internal/packager2/pull.go
+++ b/src/internal/packager2/pull.go
@@ -63,7 +63,7 @@ func Pull(ctx context.Context, src, dir, shasum, architecture string, filter fil
 	isPartial := false
 	switch u.Scheme {
 	case "oci":
-		l.Info("starting pull from oci source", "src", src, "digest", shasum)
+		l.Info("starting pull from oci source", "src", src)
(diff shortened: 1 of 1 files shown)


brandtkeller (approved):
LGTM

PR #3744, opened 2025-04-28
chore: remove invopop/jsonschema dependency from root go.mod

AustinAbro321:
## Description

In #2886 the json schema was separated in another directory under hack. However the main zarf go.mod still contained the jsonschema dependency because it was implementing one of the methods to add extra items to the schema outside of what was possible with json/jsonschema tags. 

This was originally added to the schema so users would know that package templates for imports don't work, however since compose runs before the schema checks these schema checks were not reached. By adding explicit validation we can remove the jsonschema dependency and make the error message show up in the relevant workflows.

## Checklist before merging

### DIFF
--- a/hack/schema/go.mod
+++ b/hack/schema/go.mod
@@ -6,7 +6,7 @@ replace github.com/zarf-dev/zarf => ../..
 
 require (
 	github.com/invopop/jsonschema v0.13.0
-	github.com/zarf-dev/zarf v0.38.2
+	github.com/zarf-dev/zarf v0.0.0-local // grabbed from local
 )
 
 require (
--- a/src/internal/packager2/layout/import.go
+++ b/src/internal/packager2/layout/import.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/zarf-dev/zarf/src/pkg/logger"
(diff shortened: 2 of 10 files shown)


hack/schema/go.mod
```diff
@@ -6,7 +6,7 @@ replace github.com/zarf-dev/zarf => ../..
 
 require (
 	github.com/invopop/jsonschema v0.13.0
-	github.com/zarf-dev/zarf v0.38.2
+	github.com/zarf-dev/zarf v0.0.0-local // grabbed from local
```
AustinAbro321, line 9: more intuitive this way imo

src/internal/packager2/layout/import.go
```diff
@@ -175,6 +176,9 @@ func resolveImports(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath,
 
 func validateComponentCompose(c v1alpha1.ZarfComponent) error {
 	errs := []error{}
+	if strings.Contains(c.Import.Path, v1alpha1.ZarfPackageTemplatePrefix) || strings.Contains(c.Import.URL, v1alpha1.ZarfPackageTemplatePrefix) {
```
brandtkeller, line 179: As a part of the packager2 SDK I appreciate this. Can I confirm though that you won't actually hit this code from the CLI as linting will always occur before? For `package create` I believe this is true but want to make sure I'm not overlooking a potential workflow.

AustinAbro321: My original description wasn't very clear, I reworded it. Users can hit this error now as linting happens after compose. This also means that users were not hitting this error before, so the error message wouldn't directly suggest that package templates were not rendered in the workflow

brandtkeller: Thanks for the clarity. Part of my issue was testing locally with `path: ###ZARF_PKG_TMPL_PATH###` instead of having it properly wrapped `path: "###ZARF_PKG_TMPL_PATH###"`

brandtkeller (approved):
validated locally and traced the workflow to see that this error will be produced prior to linting for added clarity. LGTM.

PR #3746, opened 2025-04-28
fix: retry on images and components

Noxsios:
## Description

Restore some retry logic changed in #3559 

## Related Issue

Relates to #2104 , #1568 , #1218 

## Checklist before merging

### DIFF
--- a/src/internal/packager/images/push.go
+++ b/src/internal/packager/images/push.go
@@ -219,7 +219,7 @@ func copyImage(ctx context.Context, src *oci.Store, remote oras.Target, srcName 
 	copyOpts.Concurrency = concurrency
 	copyOpts.WithTargetPlatform(desc.Platform)
 	copyOpts.PreCopy = func(_ context.Context, desc ocispec.Descriptor) error {
-		logger.From(ctx).Debug("pushing", "mediaType", desc.MediaType, "digest", helpers.First30Last30(desc.Digest.String()))
+		logger.From(ctx).Debug("pushing", "blob", desc)
 		return nil
 	}
 	_, err = oras.Copy(ctx, src, srcName, remote, dstName, copyOpts)
--- a/src/internal/packager/images/push.go
+++ b/src/internal/packager/images/push.go
@@ -124,7 +124,8 @@ func Push(ctx context.Context, cfg PushConfig) error {
 
 				err = retry.Do(
 					func() error { return pushImage(img, offlineNameCRC) },
-					retry.OnRetry(func(_ uint, err error) {
+					retry.OnRetry(func(attempt uint, err error) {
+						l.Debug("retrying image push", "attempt", attempt, "error", err)
 						if err != nil && tunnel != nil {
 							_ = tunnel.Reconnect(ctx)
 						}
@@ -136,9 +137,6 @@ func Push(ctx context.Context, cfg PushConfig) error {
 				if err != nil {
 					return err
 				}
-				if err := pushImage(img, offlineNameCRC); err != nil {
-					return err
-				}
 			}
 
 			// To allow for other non-zarf workloads to easily see the images upload a non-checksum version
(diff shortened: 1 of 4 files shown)


src/pkg/cluster/tunnel.go
```diff
@@ -458,14 +477,15 @@ func (tunnel *Tunnel) establish(ctx context.Context) (string, error) {
...
-	dialer, err := createDialer("POST", portForwardCreateURL, tunnel.restConfig)
+	dialer, err := createDialer(http.MethodPost, portForwardCreateURL, tunnel.restConfig)
 	if err != nil {
 		return "", fmt.Errorf("unable to create the dialer %w", err)
 	}
 
 	// Construct a new PortForwarder struct that manages the instructed port forward tunnel.
 	ports := []string{fmt.Sprintf("%d:%d", localPort, tunnel.remotePort)}
-	portforwarder, err := portforward.New(dialer, ports, tunnel.stopChan, tunnel.readyChan, tunnel.out, tunnel.out)
+	errOut := &slogWriter{l, fmt.Sprintf("tunnel error output for %s: ", ports)}
```
Noxsios, line 487: i can delete this debug writer if yall want, the tunnel doesn't really print useful information most of the time that isnt captured in the errChan

src/pkg/cluster/tunnel.go
```diff
@@ -402,7 +401,27 @@ func (tunnel *Tunnel) FullURL() string {
...
-	close(tunnel.stopChan)
+	if tunnel.stopChan == nil {
+		return
+	}
+	select {
+	case <-tunnel.stopChan:
+	default:
+		close(tunnel.stopChan)
+	}
+}
```
Noxsios, line 412: prevents panics if `Close` is called multiple times

src/pkg/cluster/tunnel.go
daveworth: I would prefer we do that - we may want similar functionality but this goes against some structured logging standards - prefixing error messages with strings points me to a code smell of not using a well-named field to carry context.

src/internal/packager/images/push.go
```diff
@@ -134,13 +147,29 @@ func Push(ctx context.Context, cfg PushConfig) error {
...
 			pushed = append(pushed, img)
 		}
 		return nil
-	}, retry.Context(ctx), retry.Attempts(uint(cfg.Retries)), retry.Delay(500*time.Millisecond))
+	}, retry.Context(ctx), retry.Attempts(uint(cfg.Retries)), retry.Delay(500*time.Millisecond), retry.OnRetry(func(attempt uint, _ error) {
+		if attempt == uint(cfg.Retries) {
+			cfg.ResponseHeaderTimeout = 60 * time.Second // this should really never happen
+		}
+		l.Debug("retrying component image(s) push", "response_timeout", cfg.ResponseHeaderTimeout)
+	}))
```
AustinAbro321, line 172: This if statement will never be reached, it starts at attempt 0. Verified with below code
```go
func main() {
	retry.Do(func() error {
		return fmt.Errorf("error")
	}, retry.Attempts(3), retry.OnRetry(func(attempt uint, _ error) {
		fmt.Println("attempt", attempt)
		if attempt == uint(3) {
			fmt.Println("here")
		}
	}))
}
```

src/pkg/cluster/tunnel.go
AustinAbro321: Interesting, I would have thought errors like below were coming from the `tunnel.out` in latest Zarf, I didn't realize it's being sent to `io.Discard`. I guess Kubernetes must have a println somewhere
image

I don't know what information the tunnel would give, I would be fine with slogging it to debug, but I don't feel strongly either way.

AustinAbro321 (changes requested):
One small comment otherwise looks good

src/internal/packager/images/push.go
```diff
@@ -122,7 +125,17 @@ func Push(ctx context.Context, cfg PushConfig) error {
 					return err
 				}
 
-				if err = pushImage(img, offlineNameCRC); err != nil {
+				err = retry.Do(
+					func() error { return pushImage(img, offlineNameCRC) },
+					retry.OnRetry(func(_ uint, err error) {
+						cfg.OCIConcurrency = 1
```
AustinAbro321, line 131: Thoughts on changing `ociconcurrency` to a local variable defined at the start of the component level retry? 

This way if it's a problem with the port forward then this will fail both attempts and concurrency will be reset to the user submitted value. If it's a problem with the iops / network it'll be set to one for the subsequent retries.

src/internal/packager/images/push.go
Noxsios: verified fix with

```go
func main() {
	attempts := uint(3)

	retry.Do(func() error {
		fmt.Println("do")
		return fmt.Errorf("error")
	}, retry.Attempts(attempts), retry.OnRetry(func(attempt uint, _ error) {
		fmt.Println("on retry", attempt)
		if attempts > 2 && attempt == attempts-2 {
			fmt.Println("last try")
		}
	}))
}
```

src/internal/packager/images/push.go
Noxsios: I think thats a good balance, ill implement

AustinAbro321 (approved):
LGTM

PR #3750, opened 2025-04-30
feat: initial support for IPv6 only

touchardv:
## Description

Initial support for IPv6 only mode with zarf.
Use of HTTP and localhost is kept in IPv6 mode (to avoid the need for dealing with registry configuration in the k8s container runtime) by using a socket proxy (currently a basic `socat` Alpine container image).

It is a work-in-progress and some further changes are required to:
- [X] ~~Package the proxy container image with the injector registry~~ (already supported by zarf)
- [X] Update the Rust injector to support serving the proxy image.
- [X] Fix the broken tests.
- [X] Add additional tests to cover the changes.
- [ ] Find an alternative solution for running the proxy (???)

## Related Issue

Fixes #

Relates to https://github.com/zarf-dev/zarf/issues/3455

## Checklist before merging

### DIFF
--- a/src/config/config.go
+++ b/src/config/config.go
@@ -56,7 +56,7 @@ var (
 	CLIArch string
 
 	// ZarfSeedPort is the NodePort Zarf uses for the 'seed registry'
-	ZarfSeedPort string
+	ZarfSeedPort int
 
 	CosignPublicKey string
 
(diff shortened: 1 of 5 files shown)


touchardv:
Closing, replaced by https://github.com/zarf-dev/zarf/pull/3938

PR #3751, opened 2025-04-30
feat: respect context during HTTP file downloads

### LINKED ISSUE #2647: Several functions in Zarf need to be more context aware and return an error after context is cancelled

AustinAbro321:
## Description

respect context during HTTP file downloads

## Related Issue

fixes #2647

## Checklist before merging

### DIFF
--- a/src/pkg/utils/network.go
+++ b/src/pkg/utils/network.go
@@ -101,10 +101,13 @@ func httpGetFile(ctx context.Context, url string, destinationFile *os.File) (err
 	l.Info("download start", "url", url)
 	start := time.Now()
 
-	// Get the data
-	resp, err := http.Get(url)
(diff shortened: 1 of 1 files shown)


brandtkeller (approved):
LGTM - improves the baseline and allows extension with a client if we need to further configure execution.

PR #3752, opened 2025-04-30
making vars for gitea chart repo and zarf injector

### LINKED ISSUE #3722: Make injector repo and gitea chart path configurable
We are currently composing and creating a version of our own zarf init, but references for artifacts, images, and charts have to be proxied. There is a way to do this registries using `--registry-override`, and for the gitea image using `--set AGENT_IMAGE_DOMAIN`, but there are two conspicuous ones where it is not configurable:
- In `packages/zarf-registry/zarf.yaml`, the location of the injector binary is `zarf-init-resources.s3.us-east-1.amazonaws.com`; if this was a variable with a default, it would make things a little easier
- In `packages/gitea/zarf.yaml`, `dl.gitea.io` is not overrideable, at least not easily.

Both of these values should be set as variables in their respective zarf.yaml files, and set to defaults with the current values. They should thus be overrideable using a `--set` flag on `zarf package create`

I am currently using a sed command to do this instead:
```bash
        sed -i-bk 's|zarf-init-resources\.s3\.us-east-1\.amazonaws.com|our-proxied-path.com|' packages/zarf-registry/zarf.yaml
        sed -i-bk 's|dl\.gitea\.io|our-proxied-path.com|' packages/gitea/zarf.yaml
        zarf package create . --confirm ...
```

chiqomar:
## Description

Makes the gitea helm chart path (dl.gitea.io/charts) and the path to the zarf injector binary (zarf-init-resources.s3.us-east-1.amazonaws.com) variables in zarf-config.toml, so that we replace them as necessary, like other paths already can.

## Related Issue

Fixes #3722 

## Checklist before merging

- [ x] Test, docs, adr added or updated as needed
- [ x] [Contributor Guide Steps](https://github.com/zarf-dev/zarf/blob/main/CONTRIBUTING.md#developer-workflow) followed


zarf-config.toml
```diff
@@ -6,6 +6,7 @@ agent_image_tag = 'local'
 
 # Tag for the zarf injector binary to use
 injector_version = '2025-03-24'
+injector_domain = 'zarf-init-resources.s3.us-east-1.amazonaws.com'
```
brandtkeller, line 9: When I execute `zarf dev inspect definition .` at the root of the repository - I get the following for the injector 
```yaml
- name: zarf-injector
  description: |
    Bootstraps a Kubernetes cluster by cloning a running pod in the cluster and hosting the registry image.
    Removed and destroyed after the Zarf Registry is self-hosting the registry image.
  required: true
  only:
    cluster:
      architecture: arm64
  files:
  - source: packages/zarf-registry/https:/zarf-init-resources.s3.us-east-1.amazonaws.com/injector/2025-03-24/zarf-injector-arm64
    shasum: f68cf097ace34bdd04f6b1571a8d8ab4e04a27614dd7e752199ff09e21f4089b
    target: "###ZARF_TEMP###/zarf-injector"
    executable: true
```

We need to determine why this is malformed.

brandtkeller (changes requested):
Appreciate the contribution. Couple items to resolve, otherwise this will be good to go.

zarf-config.toml
AustinAbro321: Looks like it's adding prepending the source path with the path to the imported package, without first checking if the file is a URL

AustinAbro321: Ah because templating happens after imports, Zarf doesn't see the file as a URL during imports and therefore adds the path to the imported component

brandtkeller:
Appreciate the patience @chiqomar - I believe #4070 removed the injector templating for source in favor of explicit source. We may be able to move forward with this updated in the PR.

brandtkeller:
@chiqomar Do you still intend to work on this feature?

brandtkeller:
Closing this PR as stale - feel free to open another PR if interested in the changes.

PR #3754, opened 2025-05-01
fix: set default user for git and registry state during `zarf package deploy`

AustinAbro321:
## Description

There is a bug right now when a user runs `zarf package deploy <init-package>` it won't set the default push user or git repository. I believe this was introduced when cmd was refactored to not use `init()`s at startup. The `init()` seemed to be changing state for `zarf init`, but this behavior was also required by `zarf package deploy`. Regardless, it makes more sense for these default functions to not know what Zarf sets to defaults at the cmd level. 

## Checklist before merging

### DIFF
--- a/src/types/k8s.go
+++ b/src/types/k8s.go
@@ -146,6 +146,10 @@ func (gs *GitServerInfo) FillInEmptyValues() error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


brandtkeller (approved):
Establishing defaults - aligns with the current functionality - LGTM

PR #3757, opened 2025-05-01
feat(mirror-resources): support for State fallback without flags

### LINKED ISSUE #2389: zarf package mirror-resources should either default to the zarf registry / zarf git server or require either --registry-url or --git-url

brandtkeller:
## Description

Support for falling-back to zarf state when using `zarf package mirror-resources` without supplying URL flags otherwise returning an error that is clear.

## Related Issue

Fixes #2389 

## Checklist before merging

### DIFF
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -348,17 +348,17 @@ func (o *packageMirrorResourcesOptions) run(cmd *cobra.Command, args []string) (
 		images += len(component.Images)
 		repos += len(component.Repos)
 	}
-	logger.From(ctx).Debug("Package contains images and repos", "images", images, "repos", repos)
+	logger.From(ctx).Debug("package contains images and repos", "images", images, "repos", repos)
 
 	// We don't yet know if the targets are internal or external
 	c, _ := cluster.New(ctx) //nolint:errcheck
 
 	if images == 0 && o.mirrorImages {
-		logger.From(ctx).Warn("No images found in package to mirror")
+		logger.From(ctx).Warn("no images found in package to mirror")
 	}
 
 	if o.mirrorImages && images > 0 {
-		logger.From(ctx).Info("Mirroring images", "images", images)
+		logger.From(ctx).Info("mirroring images", "images", images)
 		if pkgConfig.InitOpts.RegistryInfo.Address == "" {
 			// if empty flag & zarf state available - execute
 			// otherwise return error
@@ -366,7 +366,7 @@ func (o *packageMirrorResourcesOptions) run(cmd *cobra.Command, args []string) (
 			if err != nil {
 				return fmt.Errorf("no registry URL provided and no zarf state found")
 			}
-			logger.From(ctx).Debug("No registry URL provided, using zarf state", "address", state.RegistryInfo.Address)
+			logger.From(ctx).Debug("no registry URL provided, using zarf state", "address", state.RegistryInfo.Address)
 			pkgConfig.InitOpts.RegistryInfo = state.RegistryInfo
 		}
 		mirrorOpt := packager2.MirrorOptions{
@@ -386,17 +386,17 @@ func (o *packageMirrorResourcesOptions) run(cmd *cobra.Command, args []string) (
 	}
 
 	if repos == 0 && o.mirrorRepos {
-		logger.From(ctx).Warn("No git repositories found in package to mirror")
+		logger.From(ctx).Warn("no git repositories found in package to mirror")
 	}
 
 	if o.mirrorRepos && repos > 0 {
-		logger.From(ctx).Info("Mirroring repos", "repos", repos)
+		logger.From(ctx).Info("mirroring repos", "repos", repos)
 		if pkgConfig.InitOpts.GitServer.Address == "" {
 			state, err := c.LoadState(ctx)
 			if err != nil {
 				return fmt.Errorf("no git URL provided and no zarf state found")
 			}
-			logger.From(ctx).Debug("No git URL provided, using zarf state", "address", state.GitServer.Address)
+			logger.From(ctx).Debug("no git URL provided, using zarf state", "address", state.GitServer.Address)
 			pkgConfig.InitOpts.GitServer = state.GitServer
 		}
 
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -27,7 +27,6 @@ import (
 
 	"github.com/zarf-dev/zarf/src/config"
 	"github.com/zarf-dev/zarf/src/config/lang"
-	"github.com/zarf-dev/zarf/src/internal/dns"
 	"github.com/zarf-dev/zarf/src/internal/packager2"
 	"github.com/zarf-dev/zarf/src/pkg/cluster"
 	"github.com/zarf-dev/zarf/src/pkg/lint"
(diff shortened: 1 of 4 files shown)


mkcp (dismissed):
Clean! LGTM

src/cmd/package.go
```diff
@@ -336,6 +328,31 @@ func (o *packageMirrorResourcesOptions) run(cmd *cobra.Command, args []string) (
 		err = errors.Join(err, pkgLayout.Cleanup())
 	}()
 
+	// We don't yet know if the targets are internal or external
+	c, _ := cluster.New(ctx) //nolint:errcheck
+
+	if pkgConfig.InitOpts.RegistryInfo.Address == "" {
```
AustinAbro321, line 334: We should check if the package has images before this validation, and check if the package has repos before the git server validation. 

For example, a user without an initialized cluster mirroring a package that doesn't include git repositories will get an error here if they only specify the `--registry-*` args.

src/cmd/package.go
```diff
@@ -336,6 +328,31 @@ func (o *packageMirrorResourcesOptions) run(cmd *cobra.Command, args []string) (
...
+		// if empty flag & zarf state available - execute
+		// otherwise return error
+		state, err := c.LoadState(ctx)
+		if err != nil {
+			return fmt.Errorf("no registry URL provided and no zarf state found")
+		}
+		// default registry address will be 127.0.0.1:31999 - we need to update this in order to tunnel
+		state.RegistryInfo.Address = "http://zarf-docker-registry.zarf.svc.cluster.local:5000"
```
AustinAbro321, line 342: This will actually still work as with localhost nodeport addresses, so you can keep using what is in state. This way this will work with external registries as well

AustinAbro321 (changes requested):
A few notes, very excited for this change

src/cmd/package.go
AustinAbro321: Agree that those would be useful flags

brandtkeller: I think addressing that gap is important enough to include as additional scope - WIP

src/cmd/package.go
```diff
@@ -336,20 +342,76 @@ func (o *packageMirrorResourcesOptions) run(cmd *cobra.Command, args []string) (
...
+		err = packager2.MirrorImages(ctx, mirrorOpt)
+		if err != nil {
+			return err
+		}
+	}
+
+	if o.reposOnly {
+		if repos == 0 {
+			return fmt.Errorf("no repos found in package to mirror")
+		}
```
AustinAbro321, line 390: This is a breaking change as if a user runs `zarf package mirror-resources` without either the `--images` or `--repos` flag and their package only has either images or repos then the command would fail. As a user I'd expect the command to silently succeed.

src/cmd/package.go
```diff
@@ -336,20 +342,76 @@ func (o *packageMirrorResourcesOptions) run(cmd *cobra.Command, args []string) (
...
+	for _, component := range pkgLayout.Pkg.Components {
+		images += len(component.Images)
+		repos += len(component.Repos)
+	}
+	logger.From(ctx).Debug("Package contains images and repos", "images", images, "repos", repos)
+
+	// We don't yet know if the targets are internal or external
+	c, _ := cluster.New(ctx) //nolint:errcheck
+
+	if o.imagesOnly {
```
AustinAbro321, line 356: nit: a common case will be that both `ImageOnly` and `reposOnly` are true, which I feel could make these names misleading. Thoughts on re-naming to something like `o.mirrorImages` and `o.mirrorRepos`?

src/cmd/package.go
brandtkeller: Added a `warn` log for each scenario but otherwise still a successful exit code

AustinAbro321: Maybe `Info` would be more appropriate? Nothing went wrong per say. Or I feel `Warn` would be appropriate if we only logged when a user used either the `--repos` or `--images` flag and they didn't have the specified resource

brandtkeller: style guide would help here. There can be a lot in logging and `Info` might obfuscate an unintentional result? versus `Warn`

```bash
dev@dev:~/work/zarf$ ./build/zarf package mirror-resources zarf-package-git-data-amd64-0.0.1.tar.zst
2025-05-02 18:15:28 INF using config file location=/home/dev/work/zarf/zarf-config.toml
2025-05-02 18:15:28 WRN No images found in package to mirror
2025-05-02 18:15:28 INF Mirroring repos repos=9
2025-05-02 18:15:28 INF pushing repository to server repo=https://github.com/zarf-dev/zarf-public-test.git server=http://127.0.0.1:39711
2025-05-02 18:15:28 INF pushing repository to server repo=https://dev.azure.com/defenseunicorns/zarf-public-test/_git/zarf-public-test server=http://127.0.0.1:46253
2025-05-02 18:15:28 INF pushing repository to server repo=https://github.com/zarf-dev/zarf-public-test.git@v0.0.1 server=http://127.0.0.1:39859
2025-05-02 18:15:28 INF pushing repository to server repo=https://github.com/zarf-dev/zarf-public-test.git@refs/tags/v0.0.1 server=http://127.0.0.1:37995
2025-05-02 18:15:28 INF pushing repository to server repo=https://dev.azure.com/defenseunicorns/zarf-public-test/_git/zarf-public-test@v0.0.1 server=http://127.0.0.1:39677
2025-05-02 18:15:29 INF pushing repository to server repo=https://github.com/zarf-dev/zarf-public-test.git@refs/heads/dragons server=http://127.0.0.1:34441
2025-05-02 18:15:29 INF pushing repository to server repo=https://dev.azure.com/defenseunicorns/zarf-public-test/_git/zarf-public-test@refs/heads/dragons server=http://127.0.0.1:33961
2025-05-02 18:15:29 INF pushing repository to server repo=https://github.com/zarf-dev/zarf-public-test.git@01a23218923f24194133b5eb11268cf8d73ff1bb server=http://127.0.0.1:45897
2025-05-02 18:15:30 INF pushing repository to server repo=https://dev.azure.com/defenseunicorns/zarf-public-test/_git/zarf-public-test@01a23218923f24194133b5eb11268cf8d73ff1bb server=http://127.0.0.1:42151
```

More than willing to change it over to `Info`

AustinAbro321: I can see the argument for it either way, and won't block on it but I feel it's very unlikely that a user would want to mirror resources in a package and not realize the package doesn't have the resource they want to mirror. If they're running this command it's likely with a specific purpose in mind. 

Side note, the warning is currently always issued if the package has no repos, even if only `--images` was used.

brandtkeller: I'll advocate for warn here but agree that the log presence is a rough edge that we can solve with a quick additional conditional.

src/cmd/package.go
AustinAbro321, line 1: nit: standard in Zarf is to keep logs lowercase unless they start with a proper noun like `Zarf` or `Helm`, otherwise lgtm

AustinAbro321 (approved):
LGTM

PR #3760, opened 2025-05-02
fix: template variables in values file in `zarf package inspect value-files`

AustinAbro321:
## Description

Fixes a bug `zarf package inspect values-files` does not template the value templates

## Checklist before merging


brandtkeller (approved):
LGTM - validated locally that this is now templating accordingly.

PR #3767, opened 2025-05-06
feat: cache OCI packages

### LINKED ISSUE #2033: Cache OCI packages for redeployment
When deploying OCI references over an over again (common in dev workflows) you currently need to redownload the package over and over again even if the shasums have not changed.

- **Given** I have an OCI package
- **When** I run `zarf package deploy oci://reference:tag`
- **Then** Zarf will download the layers to cache and unpack them from there instead of tmp.
- **And When** I rerun  `zarf package deploy oci://reference:tag`
- **Then** Zarf will deploy from the cache instead of redownloading (assuming the remote shas for the tag have not changed)

We could tell people to pull then deploy (and this is a valid work around for now) but it is not quite the experience many people would expect.

### Additional context
We need to consider disk management and may need to provide better ways to look into and prune `zarf` cache as a part of this issue.

AustinAbro321:
## Description

This caches all layers of Zarf package inside the already used Zarf cache (by default `~/.zarf-cache/images`). If a layer already exists in the local cache, it will be taken from there. If not, the layer will be grabbed from the upstream per usual. 

## Related Issue

Fixes #2033 

## Checklist before merging


AustinAbro321:
Trying out with a Zarf package, [longhorn](https://github.com/zarf-dev/zarf/tree/main/examples/longhorn) which is 1.15 GB took about ~30-60 seconds for me to pull with current Zarf. After using the cache this drops down ~6 seconds.

src/internal/packager2/layout/create.go
```diff
@@ -145,7 +145,7 @@ func CreatePackage(ctx context.Context, packagePath string, opt CreateOptions) (
 			ImageList:             componentImages,
 			Arch:                  pkg.Metadata.Architecture,
 			RegistryOverrides:     opt.RegistryOverrides,
-			CacheDirectory:        filepath.Join(cachePath, ImagesDir),
+			CacheDirectory:        filepath.Join(cachePath, zoci.ImageCacheDirectory),
```
mkcp, line 148: Thoughts on changing `images.PullConfig.CacheDirectory` to

```
type CacheConfig struct {
  Directory string
}
```

Users may want to access to varying cache patterns in the future (for example, skip cache) and this allows us to add keys without introducing breaking existing behaviors.

mkcp (approved):
LGTM, open question about cache configuration but otherwise impl looks good

PR #3771, opened 2025-05-07
chore(dependencies): update crane

AustinAbro321:
## Description

There was a breaking change in the docker dependency that causes Crane to not work with 28+ https://github.com/google/go-containerregistry/issues/2072. Crane has fixed this issue, but has not released, I assume they don't want to break people who can't update to Docker 28. Because the only commit since their last release has been fixing the Docker breaking change, I feel comfortable pinning to a commit

## Checklist before merging


brandtkeller (approved):
Sounds fair - I am going to assume new releases will still get flagged by dependabot and thus this won't fly under the radar.

Issue #3525, opened 2025-02-20
Packager Refactor - Parent Issue


brandtkeller:
### Describe what should be investigated or refactored
As part of schema work it has been identified that src/packager intertwines "deploy time" and "package create time" logic. 

We will begin by refactoring the following pieces of src/packager into src/packager2 - this will allow us to re-implement this functionality without making breaking changes to Zarf in between releases. We will then swap over the calls to packager.() functions to packager2.() and deprecate src/packager functionality for src/packager2 (while renaming packager2 to packager)

Decisions Made:

- To refactor "Opts" out of the larger Packager struct, Opts will be moved into their respective "primary file". All exported functions will take its "related" Opts struct as a function argument.
  - E.g. - [InspectOpts](https://github.com/zarf-dev/zarf/blob/cf4e9891d531b792849394f80830363411242fd9/src/types/packager.go#L27) is moved into the [inspect.go](https://github.com/zarf-dev/zarf/pull/2990/files#diff-ce1d4a01e228f255d9bda4038e0ba0c7dc1d8f798d3629503d4e137076e6f90aR14) file within packager2

### Additional context
Doing this work should consider the following targets:
- Better unit testing
- Less global state
- Less side effects from calls such as `os.Chdir`
- Cleaner API for library users
- Support for standard cross-architecture workflows

brandtkeller:
See #2969 for historical reference

PR #3773, opened 2025-05-07
feat(packager): make packager2.Pull public

mkcp:
## Description
This pr moves `packager2.Pull` to `packager` and replaces the original implementation. It also adds a struct for optional params. Getting some testhttp flakiness that might be due to running in parallel, checking in CI

## Related Issue
Relates to #3525 

## Checklist before merging


mkcp:
PR is more or less ready to go. There's a few REVIEW comments in pull_test.go that mark commented-out t.Parallel calls. Once those are addressed this should be good for merged.

AustinAbro321 (approved):

mkcp:
Discussed more, and we decided to take a slightly different approach here than the original PR description. Now, `packager.Pull` provides the same API as after the full upgrade from `packager2` (see #3525) so this may be utilized in the next Zarf release and should remain stable. Also, after more discussion the `destination` directory is now a required parameter.

Issue #3774, opened 2025-05-08
Give users the ability to turn off the Zarf cache


AustinAbro321:
### Is your feature request related to a problem? Please describe.
Currently it is not possible to turn off the zarf-cache. If you give `zarf package create` a bad path or an empty path such as `zarf packge create --zarf-cache=""` then the `zarf-cache` will be set back to default.

Instead we should introduce a flag `--no-cache`. This would also apply to `zarf package pull` and `zarf init` (zarf init caches the init package in the zarf cache when it is downloaded). 

This could be helpful for users who have limited disk space. Though there may be some commands which use a temporary directory as they hold necessary files in the cache. For instance importing a skeleton package requires the cache. 

### Describe the behavior you'd like

- **Given** a zarf.yaml with images
- **When** a user runs `zarf package create --no-cache`
- **Then** image layers do not get cached

### Describe alternatives you've considered
 If `--zarf-cache` is set to empty then we could avoid using the cache, this may be confusing though as usually empty for a directory input is the cwd.

AustinAbro321:
Skeleton packages are pulled into the cache, but in this case they require the cache. Instead they should pull to a tmpdir, and use the new cache in `zoci.NewRemote()`. That way `--no-cache` can be used as this will still work
https://github.com/zarf-dev/zarf/blob/d45eb73ac557daed0d2618be1947498b17926c01/src/internal/packager2/layout/import.go#L219

AustinAbro321:
I'm not certain that we want an explicit `--no-cache`, it may be better to warn instead of error when a user provides a `--zarf-cache` directory we can't create, then in those cases don't cache the resources.  This would open up users to run `zarf package create --zarf-cache=/dev/null` for instance

PR #3775, opened 2025-05-08
feat: avoid using a cache in `zarf package pull` is cachepath is not set

AustinAbro321:
## Description

`zarf package pull` recently started caching layers as of #3767. The `config.CommonOptions.CachePath` is always set during CLI pulls, however since we are not setting the cachepath in several unit tests, running unit tests on main creates several folders in the users local directory. 

This does allow users to run `zarf package pull --zarf-cache=""` to avoid caching the package entirely, and they can run `zarf package pull --zarf-cache="."` to cache in the current directory. 

This relates to #3774 as Zarf should have more intuitive options to allow users to cache or not cache. Prioritizing this as tests are messy on main currently. 

## Checklist before merging

### DIFF
--- a/src/pkg/zoci/common.go
+++ b/src/pkg/zoci/common.go
@@ -37,20 +37,23 @@ type Remote struct {
 // with zarf opination embedded
 func NewRemote(ctx context.Context, url string, platform ocispec.Platform, mods ...oci.Modifier) (*Remote, error) {
 	l := logger.From(ctx)
-	absCachePath, err := config.GetAbsCachePath()
-	if err != nil {
-		return nil, err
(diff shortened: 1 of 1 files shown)


brandtkeller (approved):
LGTM - critical path to resolving issues. I imagine we still have some future iteration to perform around the `config` package and interweaving it with `packager2`

AustinAbro321:
Excellent point @brandtkeller. We could pass through the common opts through the library instead of relying on a global. I think at some point we'll want to do this pretty much across the board. It does make function signatures longer and a calls a bit more cumbersome on caller, but it is overall more explicit and clear. For instance, I may have caught the unit test problem early if the signature cache was a required parameter

PR #3782, opened 2025-05-09
feat: use builtin ORAS cache for skeleton packages

AustinAbro321:
## Description

In #3767 we created a generic cache for ORAS layers. This is used by default, but we also have a special implementation of caching specifically for skeleton packages. This PR removes that implementation so only the generic cache is used. By doing this we get closer to enabling the ability for users to turn off the cache. 
## Checklist before merging


AustinAbro321:
Closing this, may get back to it in the future, but not a priority atm

PR #3797, opened 2025-05-14
fix: avoid panic during wait-for HTTP server

AustinAbro321:
## Description

Currently there is a panic when an HTTP server that someone is waiting for returns an error code that is not the expected error code. This is because it will try to call the `.Error()` function on a nil err.

## Checklist before merging

### DIFF
--- a/src/pkg/utils/wait.go
+++ b/src/pkg/utils/wait.go
@@ -32,10 +32,11 @@ func isJSONPathWaitType(condition string) bool {
 // ExecuteWait executes the wait-for command.
 func ExecuteWait(ctx context.Context, waitTimeout, waitNamespace, condition, kind, identifier string, timeout time.Duration) error {
 	l := logger.From(ctx)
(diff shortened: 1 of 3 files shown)


src/pkg/utils/wait.go
```diff
@@ -166,10 +167,14 @@ func waitForNetworkEndpoint(ctx context.Context, resource, name, condition strin
 				if condition == "success" {
 					// Try to get the URL and check the status code.
 					resp, err := http.Get(url)
+					if err != nil {
+						l.Debug(err.Error())
+						continue
+					}
 
 					// If the status code is not in the 2xx range, try again.
-					if err != nil || resp.StatusCode < 200 || resp.StatusCode > 299 {
```
brandtkeller, line 171: This would certainly lead to the issue and be problematic - nice debug

brandtkeller (approved):
LGTM - direct change to critical path - improved logging along the way.

PR #3806, opened 2025-05-15
fix: respect `.metadata.architecture` on package create output to registry

AustinAbro321:
## Description

There is a bug wherein if the package has a predefined architecture in `.metadata.architecture` it will not be respected when the package is published. 

## Checklist before merging


src/internal/packager2/create_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			reg := createRegistry(t, ctx)
+			config.CLIArch = tt.arch
+			err := Create(ctx, tt.path, CreateOptions{
+				Output: fmt.Sprintf("oci://%s", reg.String()),
+			})
+			require.NoError(t, err)
+			layout := pullFromRemote(t, ctx, fmt.Sprintf("%s/create-arch:0.0.1", reg.String()), tt.arch)
```
mkcp, line 41: parameterize the pull target here (generate from tt.path or add field to test case?), similar concern to above re: adding new test cases and running into unexpected dependencies in test logic.

mkcp (changes requested):

mkcp (approved):
LGTM, pending CI

PR #3811, opened 2025-05-15
feat: add checking in remote repo if manifest is a chart

### LINKED ISSUE #3435: FluxCD helm release fails when using ocirepo resource as chart reference

a1994sc:
## Description

Makes call to zarf registry when trying to mutate `ocirepo` resource, and checks if the manifest is a helm chart

## Related Issue

Fixes #3435

## Checklist before merging

### DIFF
--- a/src/internal/agent/hooks/flux-ocirepo.go
+++ b/src/internal/agent/hooks/flux-ocirepo.go
@@ -12,6 +12,7 @@ import (
 	"github.com/defenseunicorns/pkg/helpers/v2"
 	"github.com/fluxcd/pkg/apis/meta"
 	flux "github.com/fluxcd/source-controller/api/v1beta2"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/zarf-dev/zarf/src/config"
 	"github.com/zarf-dev/zarf/src/config/lang"
 	"github.com/zarf-dev/zarf/src/internal/agent/operations"
@@ -19,6 +20,14 @@ import (
 	"github.com/zarf-dev/zarf/src/pkg/logger"
 	"github.com/zarf-dev/zarf/src/pkg/transform"
 	v1 "k8s.io/api/admission/v1"
+	"oras.land/oras-go/v2"
+	orasRemote "oras.land/oras-go/v2/registry/remote"
+	"oras.land/oras-go/v2/registry/remote/auth"
+	orasRetry "oras.land/oras-go/v2/registry/remote/retry"
+)
+
+const (
+	HelmMediaTypeManifest = "application/vnd.cncf.helm.config.v1+json"
 )
 
 // NewOCIRepositoryMutationHook creates a new instance of the oci repo mutation hook.
@@ -97,7 +106,39 @@ func mutateOCIRepo(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster
 			patchedURL = fmt.Sprintf("%s:%s", patchedURL, src.Spec.Reference.Tag)
 		}
 
-		patchedSrc, err := transform.ImageTransformHost(registryAddress, patchedURL)
+		var (
+			patchedSrc string
+			err        error
+		)
+
+		client := &auth.Client{
+			Client: orasRetry.DefaultClient,
+			Cache:  auth.NewCache(),
+			Credential: auth.StaticCredential(registryAddress, auth.Credential{
+				Username: zarfState.RegistryInfo.PullUsername,
+				Password: zarfState.RegistryInfo.PullPassword,
+			}),
+		}
+
+		registry := &orasRemote.Repository{
+			PlainHTTP: true,
+			Client:    client,
+		}
+
+		// maybe should wrap such that if it can not make the connection to the registry it will still mutate?
+		_, b, err := oras.FetchBytes(ctx, registry, registryAddress, oras.DefaultFetchBytesOptions)
+
+		var manifest ocispec.Manifest
+		if err := json.Unmarshal(b, &manifest); err != nil {
+			return nil, err
+		}
+
+		if isChart(manifest.Config.MediaType) {
+			patchedSrc, err = transform.ImageTransformHostWithoutChecksum(registryAddress, patchedURL)
+		} else {
+			patchedSrc, err = transform.ImageTransformHost(registryAddress, patchedURL)
+		}
+
 		if err != nil {
 			return nil, fmt.Errorf("unable to transform the OCIRepo URL: %w", err)
 		}
@@ -146,4 +187,12 @@ func populateOCIRepoPatchOperations(repoURL string, isInternal bool, ref *flux.O
 	}
 
 	return patches
+}
+
+func isChart(mediaType string) bool {
+	switch mediaType {
+	case HelmMediaTypeManifest:
+		return true
+	}
+	return false
 }


src/internal/agent/hooks/flux-ocirepo.go
```diff
@@ -97,7 +106,39 @@ func mutateOCIRepo(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster
...
+			}),
+		}
+
+		registry := &orasRemote.Repository{
+			PlainHTTP: true,
+			Client:    client,
+		}
+
+		// maybe should wrap such that if it can not make the connection to the registry it will still mutate?
+		_, b, err := oras.FetchBytes(ctx, registry, registryAddress, oras.DefaultFetchBytesOptions)
```
a1994sc, line 129: because I added this network call, the unit tests are failing... any advise on how to update the unit tests would be greatly appreciated

brandtkeller: This is a great question - i am unfamiliar if we have examples to pull from for mocking. 

in `src/internal/packager2/publish_test.go` we have an example of an in-memory registry. Given there was an appropriate artifact for each test to put in there you could actually run mutation where you check an actual registry.

Let me know if that is still a blocker and I'll try to think about more options.

a1994sc: Hm, OK thanks Brandt, I will see if I can rework the tests over the weekend.

a1994sc:
Hey team, I believe that this PR is in a better state to be reviewed. I have fixed the unit tests to maintained backwards-compatibility with the current logic, e.i. if A) it can not reach the zarf registry, or B) if the media type is not of type `helm`. I also included a new unit test for checking with an in-memory registry, shoot-out to @brandtkeller for the awesome help both here in the PR and in DM's .

Please let me know if there are any changes you would like to make

AustinAbro321 (changes requested):
This is an excellent start, thank you for doing research and driving this forward!

src/internal/agent/hooks/flux-ocirepo.go
```diff
@@ -97,11 +101,29 @@ func mutateOCIRepo(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster
...
 		}
 
+		// Always patch with crc32 hashing
 		patchedSrc, err := transform.ImageTransformHost(registryAddress, patchedURL)
 		if err != nil {
 			return nil, fmt.Errorf("unable to transform the OCIRepo URL: %w", err)
 		}
 
+		// Get the media type of the oci image
+		mediaType, err := getManifestConfigMediaType(ctx, zarfState, patchedSrc)
```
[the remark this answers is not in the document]
a1994sc: oh, so set `.webhooks[name: agent-flux-ocirepo.zarf.dev ].timeout` to 20 seconds?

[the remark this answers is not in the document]
a1994sc: I believe that I updated the webhook as described and lowered the timeout for the agent

src/internal/agent/hooks/flux-ocirepo_test.go
```diff
@@ -35,6 +35,111 @@ func createFluxOCIRepoAdmissionRequest(t *testing.T, op v1.Operation, fluxOCIRep
 	}
 }
 
+func TestFluxOCIHelmMutationWebhook(t *testing.T) {
```
AustinAbro321: Instead of a new function here we should move these into the existing function. We actually want the registry to be setup for the other test as currently the test takes an extra 10 seconds so it can timeout

src/internal/agent/hooks/flux-ocirepo_test.go
```diff
@@ -35,6 +35,111 @@ func createFluxOCIRepoAdmissionRequest(t *testing.T, op v1.Operation, fluxOCIRep
...
+	t.Parallel()
+
+	ctx := context.Background()
+	url, err := setupRegistry(t, ctx)
+	if err != nil {
+		panic(err)
```
AustinAbro321: User `require.NoError` over panic

src/internal/agent/hooks/flux-ocirepo.go
```diff
@@ -21,6 +22,11 @@ import (
 	v1 "k8s.io/api/admission/v1"
 )
 
+const (
+	HelmMediaTypeManifest = "application/vnd.cncf.helm.config.v1+json"
+	WebhookTimeout        = 10 * time.Second
```
AustinAbro321: nit: This isn't the WebHookTimeout. Maybe registryFetchTimeout would fit better?

src/internal/agent/hooks/flux-ocirepo.go
```diff
@@ -21,6 +22,11 @@ import (
 	v1 "k8s.io/api/admission/v1"
 )
 
+const (
```
AustinAbro321, line 25: These constants don't need to be exported

src/internal/agent/hooks/common_test.go
```diff
@@ -0,0 +1,147 @@
...
+		},
+		{
+			Host: "ghcr.io",
+			Path: "stefanprodan/manifests/podinfo",
+			Tag:  "6.9.0",
+		},
+		{
+			Host: "ghcr.io",
+			Path: "stefanprodan/podinfo",
+			Tag:  "6.9.0",
```
AustinAbro321: The chart and manifests should be quick, but pulling down this container image can take a while, try using this container image instead https://github.com/orgs/zarf-dev/packages/container/package/images%2Fhello-world which is super small.

src/internal/agent/hooks/flux-ocirepo.go
```diff
@@ -97,11 +103,33 @@ func mutateOCIRepo(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster
 			patchedURL = fmt.Sprintf("%s:%s", patchedURL, src.Spec.Reference.Tag)
 		}
 
+		// Always patch with crc32 hashing
```
AustinAbro321: nit: somewhat misleading comment since we unpatch it later

src/internal/agent/hooks/common_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/zarf-dev/zarf/src/pkg/state"
+	"github.com/zarf-dev/zarf/src/pkg/transform"
+	"github.com/zarf-dev/zarf/src/test/testutil"
+	"github.com/zarf-dev/zarf/src/types"
+	"oras.land/oras-go/v2"
+	"oras.land/oras-go/v2/registry/remote"
+)
+
+func populateLocalRegistry(t *testing.T, ctx context.Context, localUrl string, artifact transform.Image) error {
```
AustinAbro321: nit: since these are test functions we can use `require.NoError` directly instead of returning an error

AustinAbro321 (changes requested):
A few nits, and some test cleanup to be done. Overall changes are looking good.

src/internal/agent/hooks/flux-ocirepo_test.go
a1994sc: Tbh, the separate function was mostly so that I can focus my tests on the function I was creating, so I can merge that back into one function without too much issue

src/internal/agent/hooks/common_test.go
a1994sc: Fair, I mostly just went with `ghcr.io/stefanprodan/podinfo` because it was a known entity, and was on `ghcr.io` not `docker.io`. 

I will get that updated today

src/internal/agent/hooks/flux-ocirepo_test.go
a1994sc: So I update the tests into one, had to change some of the existing tests to match the new url.... but I'm having issues with the test that use a custom `svc` such as here:

https://github.com/zarf-dev/zarf/blob/420c4c98388579b461c1714e157548ef4dc9b009/src/internal/agent/hooks/flux-ocirepo_test.go#L317-L336

Any input on how to update the tests would be great!

src/internal/agent/hooks/flux-ocirepo_test.go
```diff
@@ -257,12 +333,43 @@ func TestFluxOCIMutationWebhook(t *testing.T) {
...
+			// a little hacky, but overrides the svc endpoint when testing with a test svc
+			if tt.svc == nil {
+				s = &state.State{RegistryInfo: types.RegistryInfo{
+					Address: url,
+				}}
+			} else {
+				s = &state.State{RegistryInfo: types.RegistryInfo{
+					Address: fmt.Sprintf("%s:%d", tt.svc.Spec.ClusterIP, tt.svc.Spec.Ports[0].Port),
+				}}
+			}
```
a1994sc: I have some mixed feelings on how I implemented this part, where I override the test endpoint of the zarf docker registry, the tests do now pass... but Please let me know if you want me to rework it

site/src/content/docs/ref/init-package.mdx
```diff
@@ -153,7 +153,7 @@ The `zarf-agent` is a [Kubernetes Mutating Webhook](https://kubernetes.io/docs/r
...
 
-The `zarf-agent` modifies the following [flux](https://fluxcd.io/flux/) resources: [GitRepository](https://fluxcd.io/docs/components/source/gitrepositories/), [OCIRepository](https://fluxcd.io/flux/components/source/ocirepositories/), & [HelmRepository](https://fluxcd.io/flux/components/source/helmrepositories/) to point to the local Git Server or Zarf Registry. HelmRepositories are only modified if the `type` key is set to `oci`.
+The `zarf-agent` modifies the following [flux](https://fluxcd.io/flux/) resources: [GitRepository](https://fluxcd.io/docs/components/source/gitrepositories/), [OCIRepository](https://fluxcd.io/flux/components/source/ocirepositories/), & [HelmRepository](https://fluxcd.io/flux/components/source/helmrepositories/) to point to the local Git Server or Zarf Registry. HelmRepositories are only modified if the `type` key is set to `oci`. During the mutation of OCIRepositories, a call is made to the zarf registry to determine what media type the oci artifact, if the artifact is a helm chart the mutation will __NOT__ include the crc32 hash as including the hash interferes with the Flux deployment of the chart.
```
AustinAbro321:
```suggestion
The `zarf-agent` modifies the following [flux](https://fluxcd.io/flux/) resources: [GitRepository](https://fluxcd.io/docs/components/source/gitrepositories/), [OCIRepository](https://fluxcd.io/flux/components/source/ocirepositories/), & [HelmRepository](https://fluxcd.io/flux/components/source/helmrepositories/) to point to the local Git Server or Zarf Registry. HelmRepositories are only modified if the `type` key is set to `oci`. During the mutation of OCIRepositories, a call is made to the zarf registry to determine the media type of the OCI artifact. If the artifact is a helm chart the mutation will __NOT__ include the crc32 hash as including the hash interferes with the Flux deployment of the chart.
```

src/internal/agent/hooks/common_test.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func setupRegistry(t *testing.T, ctx context.Context, port int, artifacts []transform.Image, copyOpts oras.CopyOptions) (string, error) {
+	localUrl := testutil.SetupInMemoryRegistry(ctx, t, port)
+
+	localReg, err := remote.NewRegistry(localUrl)
+	localReg.PlainHTTP = true
+	if err != nil {
+		return "", err
+	}
```
AustinAbro321: localReg is never used

AustinAbro321 (changes requested):
A few small tweaks, otherwise looks good

src/internal/agent/hooks/flux-ocirepo_test.go
a1994sc: _thanks github for not saying that this wasn't visible to anyone else..._

AustinAbro321: A cleaner solution could be to move up `setupRegistry` above the table tests, then add a field `registryAddress` to `admissionTest`. Then we can avoid the if statement, and each entry could own it's address

src/internal/agent/hooks/flux-ocirepo.go
```diff
@@ -97,11 +103,34 @@ func mutateOCIRepo(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster
...
 		patchedSrc, err := transform.ImageTransformHost(registryAddress, patchedURL)
 		if err != nil {
 			return nil, fmt.Errorf("unable to transform the OCIRepo URL: %w", err)
 		}
 
+		timeoutCtx, cancel := context.WithTimeout(ctx, registryFetchTimeout)
+		defer cancel()
+
+		// Get the media type of the oci image
+		mediaType, err := getManifestConfigMediaType(timeoutCtx, zarfState, patchedSrc)
```
brandtkeller, line 116: @AustinAbro321 what are your thoughts on completeness of this solution and support of oci 1.1? 

I'd be willing to let this slide for now and writing an issue - but I believe the more complete solution would be primarily supporting `artifactType` and then falling back to `config.mediaType`. 

Thoughts?

brandtkeller: Could be as simple as:

```
mt := manifest.ArtifactType       // OCI 1.1+
if mt == "" {
    mt = manifest.Config.MediaType // pre-1.1 fallback
}
```

src/internal/agent/hooks/flux-ocirepo_test.go
brandtkeller: I intend to dig into this more intentionally asap - I am curious if this pattern tests the logic accordingly or is working-around it - but I need to step through the code. Will report back.

brandtkeller (commented):
I like the direction - appreciate your patience. Planning to step through the testing shortly but wanted to pass on some thoughts.

src/internal/agent/hooks/flux-ocirepo.go
AustinAbro321: This article does suggest that it's the way of the future https://oras.land/docs/concepts/artifact/, however Helm does not add an artifactType as of now. When they do add one it's fairly likely it won't be the same string as the media type so the check could start failing if they do add it. I think it's safer for now that we assume there's no artifact type then add it in the future

src/internal/agent/hooks/flux-ocirepo_test.go
brandtkeller: Appreciate the patience. Core functionality looks great but I'd like to revisit this test so as to have the purpose of the `svc` object in the `admissionTest` appropriately testing our logic for internal service lookup. 

Spent some time mapping these changes to the test on `main` and do believe this is working-around the issue and not testing service discovery logic. 

Stepping through the test -> [NewOCIRepositoryMutationHook](https://github.com/zarf-dev/zarf/blob/main/src/internal/agent/hooks/flux-ocirepo.go#L25) -> [mutateOCIRepo](https://github.com/zarf-dev/zarf/blob/main/src/internal/agent/hooks/flux-ocirepo.go#L28) -> [GetServiceInfoFromRegistryAddress](https://github.com/zarf-dev/zarf/blob/main/src/internal/agent/hooks/flux-ocirepo.go#L68) -> [serviceInfoFromNodePortURL](https://github.com/zarf-dev/zarf/blob/main/src/pkg/cluster/secrets.go#L167) -> [content](https://github.com/zarf-dev/zarf/blob/main/src/pkg/cluster/tunnel.go#L268-L304)

The workaround currently pre-sets the state to the internal IP:port - which isn't a valid state. So it fails in the [early detection](https://github.com/zarf-dev/zarf/blob/main/src/pkg/cluster/tunnel.go#L281) stages.

I'd like to see registry still set as the nodeport url `127.0.0.1:31999` and then identifying the internal service:port as intended. The ReplacePatchOperation should then be expecting the original `127.0.0.1:31999`. 

All of that said - it probably puts you back in the original spot. I can help run through the debugging more but do believe some of the root of the issue is that `localhost` was used in some of the tests for not mutating and`serviceInfoFromNodePortURL()` cannot actually handle that as expected (maybe a bug - maybe expected - needs refinement). 

Appreciate your bearing with us as we get better about responding to these PR's.

AustinAbro321:
I tried the example here, but it seems that it's not connecting to the registry. I'm not sure why. Were you able to get this to work? I added a log to make sure that state is being pulled correctly and it seems that it is 
image

a1994sc:
~I am seeing that as well... I will debug it and push an update~

Correction.... I'm seeing the following error:

image

src/internal/agent/hooks/common.go
```diff
@@ -13,3 +27,43 @@ func getLabelPatch(currLabels map[string]string) operations.PatchOperation {
...
+
+	ref, err := registry.ParseReference(imageAddress)
+	if err != nil {
+		return "", err
+	}
+
+	http, err := images.ShouldUsePlainHTTP(ctx, ref.Registry, client)
+	if err != nil {
+		return "", err
+	}
```
AustinAbro321: Easy to miss but the problem is that we're using the imageAddress rather than the registry address for credentials. ORAS requires that staticCredentials only take the host. 
```suggestion
	ref, err := registry.ParseReference(imageAddress)
	if err != nil {
		return "", err
	}
	client := &auth.Client{
		Client: orasRetry.DefaultClient,
		Cache:  auth.NewCache(),
		Credential: auth.StaticCredential(ref.Registry, auth.Credential{
			Username: zarfState.RegistryInfo.PullUsername,
			Password: zarfState.RegistryInfo.PullPassword,
		}),
	}

	http, err := images.ShouldUsePlainHTTP(ctx, ref.Registry, client)
	if err != nil {
		return "", err
	}
```

a1994sc:
Ok, that works from me, Thanks @AustinAbro321

brandtkeller (approved):
lgtm - we did just merge the migration of the agent orchestration from manifests to helm chart - so this will require a quick conflict resolve.

AustinAbro321 (approved):

a1994sc:
🎉🎉🎉🎉

Issue #3433, opened 2025-01-24
Implement and use v1beta1 schema


AustinAbro321:
### Is your feature request related to a problem? Please describe.
The [ADR-0026](https://github.com/zarf-dev/zarf/blob/main/adr/0026-schema.yaml) describes the v1beta1 schema. Once packager is refactored to make this change easier, the new schema should be introduced and used through Zarf. Specific changes are covered in the ADR.

This issue should be closed when users are able to create packages with `apiVersion: v1beta1` in their zarf.yaml

nick-adkins:
Sub-issue: https://github.com/zarf-dev/zarf/issues/2245

nick-adkins:
Another sub-issue: https://github.com/zarf-dev/zarf/issues/2059

Issue #2245, opened 2024-01-23
Refactor `charts` definition in `zarf.yaml`


Racer159:
### Describe what should be investigated or refactored
It would be worth taking some time to refactor the `charts` definition in the `zarf.yaml` to enhance user understanding of the keys such as `repoName` and `gitPath` which are usually non-intuitive for users.  `repoName` really means the name of a chart _in_ a Helm Repo and `gitPath` means the path of a chart _in_ a `git` repository which are not obvious to folks not familiar with Zarf.

### Links to any relevant code
https://github.com/defenseunicorns/zarf/blob/f039affe91b664f601b274a5e66d29bf030ca75f/src/types/component.go#L92

### Additional context
These fields have come up with questions a few times in user questions in our `k8s` slack.

AustinAbro321:
propose a schema that looks like below. This is based on the examples/helm-charts within the Zarf repository. Non changed fields were taken out for brevity. Exactly one of the following field will exist for each `components.charts`.

```yaml
- name: podinfo-repo-new
  helmRepo:
    url: https://stefanprodan.github.io/podinfo
    name: podinfo # replaces repoName since it's only applicable for helm chart repositories
    version: 6.4.0

- name: podinfo-git-new
  git:
    url: https://stefanprodan.github.io/podinfo@6.4.0
    path: charts/podinfo
    # no version field, use chart.yaml version field at that git tag

- name: podinfo-oci-new
  oci:
    url: oci://ghcr.io/stefanprodan/charts/podinfo
    version: 6.4.0

- name: podinfo-local-same
  local:
   path: chart
    # no version field, use local chart.yaml version
```

AustinAbro321:
It should be noted that in the current Zarf schema the `version` flag is used for both OCI charts and git charts. We should leave these flags off of the git sub object and instead we should grab the information from the chart.yaml in the git repo / local directory. Git repos should specify their version with the @ syntax instead of using the version field both for the chart version and the upstream tag. 

https://github.com/zarf-dev/zarf/blob/10bf7465c4b0fc0b0477bb55f5189cd2b85d0d01/src/internal/packager/helm/repo.go#L44-L53

https://github.com/zarf-dev/zarf/blob/10bf7465c4b0fc0b0477bb55f5189cd2b85d0d01/src/internal/packager/helm/repo.go#L156-L163

AustinAbro321:
[...]

PR #3815, opened 2025-05-16
fix(helm): Allow empty version for localpath chart

### LINKED ISSUE #3813: When specifying a helm chart via a localPath, the version field should not be required
Right now components that use helm charts from a localPath are required to specify the version as well.  This is redundant since the chart obviously has a chart version in the Chart.yaml file.  What is worse is that the version specified below can be different than the chart which is at best misleading.

```yaml
- name: splash
   localPath: ../charts/splash
   version: "0.0.1" # THIS SHOULD NOT BE REQUIRED WHEN localPath IS SPECIFIED
   namespace: default
   valuesFiles:
      - splash-values.yaml
```

brandtkeller:
## Description

chart version can be any value for `localPath` helm charts and this is used in the naming of the helm archive within the zarf package. This can result in confusion and inaccuracy of artifact versions were one to inspect/decompress a package. 

Given that a chart name in the `ZarfPackageConfig` must be unique (and is enforced during create) - then allowing for version being not present should not create any collisions.

This PR makes the following assumptions:
- `localPath` charts should not require a version be specified
-  return an error when expected and actual versions mismatch. 

This PR helps decide on paths forward. 

## Related Issue

Fixes #3813

## Checklist before merging

### DIFF
--- a/src/internal/packager/helm/common.go
+++ b/src/internal/packager/helm/common.go
@@ -75,6 +75,9 @@ func ChartFromZarfManifest(manifest v1alpha1.ZarfManifest, manifestPath, package
 
 // StandardName generates a predictable full path for a helm chart for Zarf.
 func StandardName(destination string, chart v1alpha1.ZarfChart) string {
+	if chart.Version == "" {
+		return filepath.Join(destination, chart.Name)
+	}
 	return filepath.Join(destination, chart.Name+"-"+chart.Version)
 }
 
--- a/src/internal/packager/helm/repo.go
+++ b/src/internal/packager/helm/repo.go
@@ -87,15 +87,9 @@ func PackageChartFromLocalFiles(ctx context.Context, chart v1alpha1.ZarfChart, c
 	}
 
 	if chart.Version != "" && chart.Version != loadedChart.Metadata.Version {
-		// warn the user - alternatively we could return an error here if the expectation is improved provenance
-		l.Warn("chart version mismatch",
-			"expected", chart.Version,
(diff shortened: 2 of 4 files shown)


AustinAbro321:
I agree with this change, however we must consider how this will effect packages created and deployed with different versions of Zarf. A package using localpath without a version, will not deploy on earlier versions of Zarf.

The error will be: 
```go
2025-05-22 10:49:41 ERR failed to deploy package: unable to deploy component "demo-helm-charts": open /tmp/zarf-3011270750/components/demo-helm-charts/values/podinfo-local--0: no such file or directory
```

I don't think we have this written down anywhere (we should write it down), but a goal with Zarf has been to not assume people have the same Zarf version for create and deploy. This change is also proposed by a comment in #2245, which will happen with #3433

Adding this for context, open to other solutions besides waiting for the next schema version

brandtkeller:
@AustinAbro321 superb callout and catch - I will get that note added to the draft principles document. Otherwise I fully agree that this should not be merged as-is. I will think about compatibility otherwise we may need to close this out.

brandtkeller:
More digging - the delta between having no version specified and the backwards compatibility across create/deploy creates a constraint where we cannot make this change without it being a breaking one. version is used in the underlying package filesystem - removing it means that older versions of zarf could not deploy a newer package. 

Even if we load the chart - obtain the version and add it to the name in the package fllesystem as expected during create - deploy will still not have a defined chart version in the `zarf.yaml` manifest to use during load. 

Closing this as incomplete.

PR #3836, opened 2025-05-21
fix(archive): skip format autodetection for tar format

### LINKED ISSUE #3835: component archive compression misidentified as brotli
### Steps to reproduce
1. build the [identity-authorization](https://github.com/defenseunicorns/uds-core/blob/main/packages/identity-authorization/zarf.yaml) package from uds-core
2. Attempt to decompress or deploy the package

### Expected result

Can decompress or deploy as expected

### Actual Result

fails due to:
```
ERROR:  failed to deploy bundle: unable to load the package: unable to perform decompression: unable to extract "/var/folders/91/ysgtzhdx4lx24fydyk48qchc0000gn/T/zarf-2254187512/components/authservice.tar": archive/tar: invalid tar header
```

brandtkeller:
## Description

We have a reproducible example of a component archive `component.tar` being misidentified as having the `brotli` compression and thus failing package deploy and tools archiver decompress due to having an invalid tar header.

This relates to: https://github.com/mholt/archives/issues/29 

Because the raw Brotli format lacks a fixed magic number, decoding 16 bytes of almost any non-Brotli data (including a plain .tar file) often succeeds without error. Meanwhile, the library’s TAR header detector can fail on non-USTAR archives (e.g. V7-style tarballs that don’t include the “ustar” signature), so the only positive match is from the Brotli stream test. The net result: a .tar file is misidentified as Brotli, passed to the Brotli decoder (which outputs junk), and then fed to the Tar extractor—hence the “invalid tar header” error.

Intending to file an issue against `mholt/archives` for this - in the meantime we can circumvent auto-detection on `.tar` files by using the tar extactor directly. 

## Related Issue

Fixes #3835

Relates to #

## Checklist before merging

### DIFF
--- a/src/pkg/archive/archive.go
+++ b/src/pkg/archive/archive.go
@@ -42,6 +42,94 @@ const (
 	filePerm = 0o644
 )
 
+var archivers = map[string]archives.Archiver{
+	// Plain archives
+	extensionZip: archives.Zip{},
+	extensionTar: archives.Tar{},
+
+	// Compressed TAR-based archives (all need Extraction set)
+	extensionGz: archives.CompressedArchive{
+		Compression: archives.Gz{},
+		Archival:    archives.Tar{},
+		Extraction:  archives.Tar{},
+	},
+	extensionTgz: archives.CompressedArchive{
+		Compression: archives.Gz{},
+		Archival:    archives.Tar{},
+		Extraction:  archives.Tar{},
+	},
+	extensionBz2: archives.CompressedArchive{
+		Compression: archives.Bz2{},
+		Archival:    archives.Tar{},
+		Extraction:  archives.Tar{},
+	},
+	extensionTbz2: archives.CompressedArchive{
+		Compression: archives.Bz2{},
+		Archival:    archives.Tar{},
+		Extraction:  archives.Tar{},
+	},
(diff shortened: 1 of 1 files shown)


src/pkg/archive/archive.go
```diff
@@ -367,17 +367,28 @@ func unarchive(ctx context.Context, src, dst string) (err error) {
...
+		format, input, err = archives.Identify(ctx, src, file)
+		if err != nil {
+			return fmt.Errorf("unable to identify archive %q: %w", src, err)
+		}
+		var ok bool
+		// Assert that it supports extraction
+		extractor, ok = format.(archives.Extractor)
+		if !ok {
+			return fmt.Errorf("unsupported format for extraction: %T", format)
+		}
```
mkcp, line 391: Blocking: the happy path (identify format and assert extraction) should not be in an else block here off of the guard case (if .tar suffix, skip). This can be fixed by wrapping this entire section in a function that returns (archives.Extractor, io.Reader, error) and returning early with extractor `archives.Tar{}` if we have a tar suffix.

mkcp (changes requested):

AustinAbro321 (approved):

Issue #3844, opened 2025-05-22
Zarf tools wait-for unable to use SELECTOR - always returns true


benmountjoy111:
### Environment
Device and OS: MacOS 15.4.1
App version: v0.55.4
Kubernetes distro being used: v1.30.3+rke2r1
Other:

### Steps to reproduce
1.  Run `zarf tools wait-for pod -n kube-system kube-proxy-<node/hash>` (using pod NAME)
2. Observe zarf correctly waits for `pod/kube-proxy-<node/hash>`
3. Run `zarf tools wait-for pod -n kube-system component=kube-proxy` (using pod SELECTOR)
4. Observe zarf waits for `pod/ with label `component=kube-proxy`. This appears to be working but it is not actually waiting on the correct resource. 
5. To prove this assumption, run `zarf tools wait-for pod -n kube-system component=this-label-does-not-exist`
6. Observe zarf waits for `pod/ with label `component=this-label-does-not-exist` and returns a successful return code (in that zarf believes this pod with the label exists, but it does not). 

According to zarf, the `wait-for` command's usage should allow a SELECTOR in place of a NAME

```
$ zarf tools wait-for --help
By default Zarf will wait for all Kubernetes resources to be ready before completion of a component during a deployment.
This command can be used to wait for a Kubernetes resources to exist and be ready that may be created by a Gitops tool or a Kubernetes operator.
You can also wait for arbitrary network endpoints using REST or TCP checks.

Usage:
  zarf tools wait-for { KIND | PROTOCOL } { NAME | SELECTOR | URI } { CONDITION | HTTP_CODE } [flags]
```
NOTE: this bug is present on all k8s resources (deployments, daemonsets, statefulsets, etc.) anytime a SELECTOR is used.

### Expected result
Zarf correctly waits-for resources. If the resource doesn't exist, zarf correctly returns a non-zero return code.

### Actual Result
Zarf **always** return success (return code of 0) when using a SELECTOR.

### Visual Proof (screenshots, videos, text, etc)
See attached screenshot of all 3 commands, and the --help / usage of zarf tools wait-for

### Severity/Priority
Medium/Low

### Additional Context
Add any other context or screenshots about the technical debt here.

PR #3846, opened 2025-05-22
feat(utils): support user interrupts in utils.ExecuteWait and add debug logging

mkcp:
## Description
Previously, `zarf tools wait-for` would require two ctrl-c's from a user to interrupt the process. This PR adds a `ctx.Done()` to `utils.ExecuteWait`'s select and some extra debugging to help trace #3844.

## Related Issue
Relates to #3844

## Checklist before merging

### DIFF
--- a/src/pkg/utils/wait.go
+++ b/src/pkg/utils/wait.go
@@ -90,11 +90,16 @@ func ExecuteWait(ctx context.Context, waitTimeout, waitNamespace, condition, kin
 		select {
 		case <-expired:
 			return errors.New("wait timed out")
+		case <-ctx.Done():
+			return errors.New("received interrupt")
 
 		default:
 			// Check if the resource exists.
+			l.Debug("checking resource existence", "namespace", namespaceFlag, "kind", kind, "identifier", identifier)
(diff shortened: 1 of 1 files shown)


AustinAbro321 (approved):
LGTM, we should also do a follow up to have `waitForNetworkEndpoint` to accept ctx as well.

mkcp:
@AustinAbro321 sgtm, should be an easy addition.

src/pkg/utils/wait.go
```diff
@@ -90,11 +90,16 @@ func ExecuteWait(ctx context.Context, waitTimeout, waitNamespace, condition, kin
 		select {
 		case <-expired:
 			return errors.New("wait timed out")
+		case <-ctx.Done():
+			return errors.New("received interrupt")
```
brandtkeller, line 94: Not blocking - curious if you believe propagation with `ctx.Err()` included in the error returned (or even checking for specific errors such as `context.Canceled would net any long-term value?

We don't use `context.WithTimeout` in this function so we wouldn't get the benefit of coupling and checking for them together.

brandtkeller (commented):
Continually improving the baseline!

PR #3854, opened 2025-05-23
chore: lint src/internal/packager

AustinAbro321:
## Description

src/internal/packager was excluded from lint on the basis that the packages would be replaced in the packager refactor (#3525), however only src/internal/packager/sbom will be replaced. This PR lints the long lived packages that were excluded.  

## Checklist before merging


mkcp (approved):
LGTM

PR #3859, opened 2025-05-26
chore: update gitea chart to `12.3.0`

a1994sc:
## Description

Updates `gitea` chart to 12.3.0:

- The upstream chart replaces `redis` with `valkey`, so I disabled the `valkey` cluster.
- Changed url to use OCI chart
- Changed `gitea_image` to use `ghcr.io` endpoint to avoid rate-limiting
  - Might want to consider changing `registry_image_*` to use ghcr.io endpoints

## Related Issue

Fixes #N/A

## Checklist before merging

### DIFF
--- a/packages/gitea/zarf.yaml
+++ b/packages/gitea/zarf.yaml
@@ -59,7 +59,7 @@ components:
       - name: gitea
         releaseName: zarf-gitea
         url: oci://registry-1.docker.io/giteacharts/gitea
-        version: 12.0.0
+        version: 12.1.2
         namespace: zarf
         valuesFiles:
           - gitea-values.yaml
--- a/packages/gitea/zarf.yaml
+++ b/packages/gitea/zarf.yaml
@@ -59,7 +59,7 @@ components:
       - name: gitea
         releaseName: zarf-gitea
         url: oci://registry-1.docker.io/giteacharts/gitea
-        version: 12.1.2
+        version: 12.3.0
         namespace: zarf
         valuesFiles:
           - gitea-values.yaml
--- a/packages/gitea/zarf.yaml
+++ b/packages/gitea/zarf.yaml
@@ -58,8 +58,8 @@ components:
     charts:
       - name: gitea
         releaseName: zarf-gitea
-        url: https://dl.gitea.io/charts
-        version: 10.1.1
+        url: oci://registry-1.docker.io/giteacharts/gitea
+        version: 12.0.0
         namespace: zarf
         valuesFiles:
           - gitea-values.yaml
--- a/zarf-config.toml
+++ b/zarf-config.toml
@@ -11,4 +11,4 @@ registry_image = 'library/registry'
 registry_image_tag = '3.0.0'
 
 # The image reference to use for the optional git-server Zarf deploys
-gitea_image = 'ghcr.io/go-gitea/gitea:1.24.3-rootless'
+gitea_image = 'ghcr.io/go-gitea/gitea:1.24.6-rootless'
--- a/zarf-config.toml
+++ b/zarf-config.toml
@@ -11,4 +11,4 @@ registry_image = 'library/registry'
 registry_image_tag = '3.0.0'
 
 # The image reference to use for the optional git-server Zarf deploys
-gitea_image = 'gitea/gitea:1.21.5-rootless'
+gitea_image = 'ghcr.io/go-gitea/gitea:1.23.8-rootless'
--- a/zarf-config.toml
+++ b/zarf-config.toml
@@ -11,4 +11,4 @@ registry_image = 'library/registry'
 registry_image_tag = '3.0.0'
 
 # The image reference to use for the optional git-server Zarf deploys
-gitea_image = 'ghcr.io/go-gitea/gitea:1.23.8-rootless'
+gitea_image = 'ghcr.io/go-gitea/gitea:1.24.3-rootless'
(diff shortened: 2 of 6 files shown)


brandtkeller:
Hey @a1994sc , would you mind rebasing this when you have a moment? I'd like to see where this lands for review currently.

a1994sc:
Oh rebase, not merge, yea I can update that later today

brandtkeller (changes requested):
E2E without cluster workflows are failing with a [hardcoded gitea image](https://github.com/zarf-dev/zarf/blob/main/src/test/e2e/06_create_sbom_test.go#L73)

I'd be fine with updating this line again unless you see a more dynamic option available.

a1994sc:
ok, I think I saw that recent too, so I can update that.

brandtkeller:
reviewing the delta of updates between the indicated application version as well as the helm chart versions. No action required by the PR.

Appreciative of catching the update on this.

a1994sc:
Not that is a very exhaustive change, but I have been rolling the latest version of the helm chart/docker image internally without too much issue. However I understand my needs aren't going to fit for the whole zarf user base

brandtkeller:
Coming back to this (finally) - appreciate the patience. Gitea as an init package component should likely get more attention than we give it to-date. 

@a1994sc would you mind updating this PR with a merge commit or rebase (however you prefer). I'd love to get this over the line soon. 

Recognizing gitea has released more updates since. I am good with keeping this update as-is or updating the artifacts to the latest releases.

a1994sc:
@brandtkeller I think it would be "better" to update the artifacts to the latest, so I will update that in the next little bit.

a1994sc:
@brandtkeller Updated

brandtkeller (approved):
This is a big jump - appreciate @a1994sc for calling this out.

Performed testing locally:
- zarf init (current) -> deploy flux -> deploy argo -> zarf init (with these changes) -> no resulting issues. 
- Tested re-deploying flux/argo without issue
- Tested fresh install of these updates w/ flux/argo packages -> no issues

Compared installation of gitea from current `init` package by comparing the outputs of `zarf package inspect manifests`. None of the kubernetes resources look modified in a way that should be a concern.

Additionally read through the release notes for releases since 1.21.5 and - while there were many changes - none stood out as alarming. 

 Zarf should look to automate the process (mend perhaps?) in order to keep update cycles closer.

a1994sc:
Renovate (/Mend) would be my suggest for a way to keep this package updated.

You would either need to "self-host" (e.i. renovate action on github) to run a [`postupgradetasks`](https://docs.renovatebot.com/configuration-options/#postupgradetasks) that runs the hack script I created, or you could use the hosted solution then have an action runs the hack script and commits the changes to the branch.

Both have either strength and weaknesses.

packages/gitea/zarf.yaml
```diff
@@ -58,8 +58,8 @@ components:
     charts:
       - name: gitea
         releaseName: zarf-gitea
-        url: https://dl.gitea.io/charts
-        version: 10.1.1
+        url: oci://registry-1.docker.io/giteacharts/gitea
+        version: 12.3.0
```
AustinAbro321, line 62: I'd like to avoid using registry1, when we used to rely on registry1 for e2e tests we frequently saw flakes from their infrastructure going down. Do we see any significant advantages to them over upstream?

zarf-config.toml
```diff
@@ -11,4 +11,4 @@ registry_image = 'library/registry'
 registry_image_tag = '3.0.0'
 
 # The image reference to use for the optional git-server Zarf deploys
-gitea_image = 'gitea/gitea:1.21.5-rootless'
+gitea_image = 'ghcr.io/go-gitea/gitea:1.24.6-rootless'
```
AustinAbro321, line 14: Nice find, makes the ever present TOOMANYREQUESTS less likely

packages/gitea/zarf.yaml
a1994sc: So this somewhat stems from my own preference to use the OCI artifacts over the http chart repos... but at the moment the gitea team are not cross publishing their helm charts over to `ghcr.io`

zarf-config.toml
a1994sc: Yes sir... really just removed one call to docker hub for another, e.i. gitea image for the gitea chart.

packages/gitea/zarf.yaml
AustinAbro321: Ahh, yeah generally I do prefer using OCI artifacts as well, but my experience with registry1 infra hasn't been great. I would ask that we revert to using upstream for now and if an official OCI source is published we switch it then

a1994sc: Sure, I can swap that out this evening... should I make the version a variable? e.i. `GITEA_CHART_VERSION` ?

brandtkeller: are we possibly mixing up `registry-1.docker.io` with `registry1.dso.mil` or has there really been flakes with dockers infrastructure?

a1994sc: When you go to the page for `giteacharts/gitea` the install box specifies `registry-1.docker.io`. I think it will work with just `docker.io` though

https://hub.docker.com/r/giteacharts/gitea

AustinAbro321: I was mixing up registry-1.docker.io with registry1.dso.mil :facepalm:, sorry. Please disregard

a1994sc: Well I think that we have seen in the pipeline where too many calls to `docker.io` can cause `TOOMANYREQUESTS ` errors

AustinAbro321: True, but that is more rare than the error I was thinking of. With registry1.dso.mil the site was often just down.

AustinAbro321 (approved):

PR #3869, opened 2025-05-28
bug(create): ensure that when given an output directory Archive creates the directory for the user

### LINKED ISSUE #3860: Package creation to non-existent output folder fails in 0.55.x
### Environment
Device and OS: MBP, macOS
App version: 0.55.x (0.55.2 and 0.55.5)
Other:

### Steps to reproduce
1. Create a zarf config file with a defined output directory:
```yaml
package:
  create:
    output: build
```
2. Attempt to create a basic zarf package such as the below (`zarf p create --confirm`):
```yaml
kind: ZarfPackageConfig
metadata:
  name: test

components:
  - name: test
    required: true
    images:
      - busybox
```

### Expected result

Package creates, automatically creating the `build` directory.

### Actual Result

`build` directory is not created, package creation fails with this error:
```
2025-05-27 08:41:06 ERR failed to create package: unable to create package: failed to create build/zarf-package-test-arm64.tar.zst: open build/zarf-package-test-arm64.tar.zst: no such file or directory
```

### Severity/Priority

Low/Medium

### Additional Context

Marking this as a possible bug because this is a regression - in our packages we have been "relying" on this behavior for CI builds for a year+, if I revert to 0.54.0 the folder is autocreated for me.

mkcp:
## Description
This PR adds a check in `packager2/layout/*PackageLayout.Archive()` to ensure dirPath is initialized. If it isn't, it will create the directory for the user. It should also preserve the directory if it already exists.

## Related Issue

Fixes #3860

## Checklist before merging

### DIFF
--- a/src/internal/packager2/layout/package.go
+++ b/src/internal/packager2/layout/package.go
@@ -179,12 +179,30 @@ func (p *PackageLayout) GetImageDir() string {
 }
 
 func (p *PackageLayout) Archive(ctx context.Context, dirPath string, maxPackageSize int) error {
-	packageName := fmt.Sprintf("%s%s", sources.NameFromMetadata(&p.Pkg, false), sources.PkgSuffix(p.Pkg.Metadata.Uncompressed))
+	// Ensure output path exists
+	// NOTE(mkcp): We may want to move this behavior to archive.Compress
+	isPresent, err := exists(dirPath)
+	if err != nil {
+		return err
+	}
+	if !isPresent {
+		err = os.MkdirAll(dirPath, os.ModePerm)
+		if err != nil {
+			return err
+		}
+	}
+	// Generate path to package tarball
+	name := sources.NameFromMetadata(&p.Pkg, false)
+	suffix := sources.PkgSuffix(p.Pkg.Metadata.Uncompressed)
+	packageName := fmt.Sprintf("%s%s", name, suffix)
 	tarballPath := filepath.Join(dirPath, packageName)
-	err := os.Remove(tarballPath)
+
+	// Overwrite package tarball if it already exists
+	err = os.Remove(tarballPath)
 	if err != nil && !errors.Is(err, os.ErrNotExist) {
 		return err
 	}
+
 	logger.From(ctx).Info("writing package to disk", "path", tarballPath)
 	files, err := os.ReadDir(p.dirPath)
 	if err != nil {
@@ -339,4 +357,16 @@ func validatePackageSignature(ctx context.Context, pkgLayout *PackageLayout, pub
 		return fmt.Errorf("package signature did not match the provided key: %w", err)
 	}
 	return nil
+}
+
+// exists checks if a file is present at the given path.
+func exists(path string) (bool, error) {
+	_, err := os.Stat(path)
+	if err == nil {
+		return true, nil
+	}
+	if errors.Is(err, fs.ErrNotExist) {
+		return false, nil
+	}
+	return false, err
 }


src/internal/packager2/layout/package.go
```diff
@@ -179,12 +179,30 @@ func (p *PackageLayout) GetImageDir() string {
 }
 
 func (p *PackageLayout) Archive(ctx context.Context, dirPath string, maxPackageSize int) error {
-	packageName := fmt.Sprintf("%s%s", sources.NameFromMetadata(&p.Pkg, false), sources.PkgSuffix(p.Pkg.Metadata.Uncompressed))
+	// Ensure output path exists
+	// NOTE(mkcp): We may want to move this behavior to archive.Compress
```
AustinAbro321, line 183: Yeah I tested it out and on previous version of Zarf users could run commands like `zarf tools archiver compress dir-to-compress dir-not-exists/my-file.tar.zst` so I think the behavior should live there

src/internal/packager2/layout/package.go
```diff
@@ -179,12 +179,30 @@ func (p *PackageLayout) GetImageDir() string {
 }
 
 func (p *PackageLayout) Archive(ctx context.Context, dirPath string, maxPackageSize int) error {
```
AustinAbro321, line 181: Tests are failing because `dirPath` will be empty when the `--output` flag is not used so we're running `os.MkdirAll` on an empty string

AustinAbro321 (changes requested):
Left some comments. Another potential way to solve this would be to add this block to the top of `archives.Compress`. This should be easier to unit test as well
```go
	if err := os.MkdirAll(filepath.Dir(dest), helpers.ReadExecuteAllWriteUser); err != nil {
		return fmt.Errorf("failed to create parent directory for %s: %w", dest, err)
	}
```

src/internal/packager2/layout/package.go
mkcp: Makes sense!

src/internal/packager2/layout/package.go
mkcp: Ah! Definitely a gap in input validation. Moving the parent directory creation down a level of abstraction helps with that problem

mkcp:
> Left some comments. Another potential way to solve this would be to add this block to the top of `archives.Compress`. This should be easier to unit test as well
> 
> ```go
> 	if err := os.MkdirAll(filepath.Dir(dest), helpers.ReadExecuteAllWriteUser); err != nil {
> 		return fmt.Errorf("failed to create parent directory for %s: %w", dest, err)
> 	}
> ```

Yeah that's a more elegant solution for sure, revising the PR now.

src/internal/packager2/layout/package.go
```diff
@@ -340,3 +346,15 @@ func validatePackageSignature(ctx context.Context, pkgLayout *PackageLayout, pub
 	}
 	return nil
 }
+
+// exists checks if a file is present at the given path.
+func exists(path string) (bool, error) {
```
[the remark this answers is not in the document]
mkcp: yup! missed it, deleted in latest

AustinAbro321 (approved):
LGTM

src/pkg/archive/archive_test.go
```diff
@@ -37,6 +37,83 @@ func readFile(t *testing.T, path string) string {
...
+			name:      "errors when both sources and dest are empty",
+			expectErr: true,
+		},
+		{
+			name:      "errors when given a dest but empty sources",
+			dest:      "err-archive.tar.gz",
+			expectErr: true,
+		},
+		{
+			name:      "errors when given sources but empty dest",
```
brandtkeller, line 78: This doesn't hit the intended error path for code coverage for empty dest given the `filepath.Join()` used in the test logic below. 

A permutation of this is still a valid testcase nonetheless as it represents a destination being a directory rather than a target but we should find a better way to augment the error as "dest is a directory" is less clear than "the destination is not a valid archive" or similar?

brandtkeller (changes requested):
One code path worth consideration otherwise functioning happy paths locally.

src/pkg/archive/archive_test.go
mkcp: It's true that the test environment setup does mean we're not actually passing an empty string into the Compress function here. Just so we're not blocking a critical bug fix, how about I go ahead and comment that test case out and we can figure out a path forward for the empty destination case in a follow up PR?

mkcp: Reframed the problem slightly and updated+added some test cases. lmk what you think

src/pkg/archive/archive.go
```diff
@@ -135,6 +136,19 @@ type CompressOpts struct{}
...
 func Compress(ctx context.Context, sources []string, dest string, _ CompressOpts) (err error) {
+	if len(sources) == 0 {
+		return fmt.Errorf("sources cannot be empty")
+	}
+	if dest == "" {
+		return fmt.Errorf("dest cannot be empty")
+	}
+
+	// Ensure dest parent directories exist
+	err = os.MkdirAll(filepath.Dir(dest), helpers.ReadExecuteAllWriteUser)
```
brandtkeller, line 147: nit: we have a package const `dirPerm` that we have been using to help drive package isolation given this is not only an internal library. I believe in doing so we adhere closely to [dependencies in go](https://go.dev/talks/2012/splash.article#TOC_7.) for imports and help prevent potential consumer issues for cyclic imports.

brandtkeller (commented):
One passing thought for the use of this as an SDK

PR #3872, opened 2025-05-29
refactor: separate out create / layout to avoid cyclic dependency with zoci

### LINKED ISSUE #3887: Refactor src/internal/packager2/layout to avoid cyclic import or duplication with zoci
As the packager refactor nears towards closing we need to eliminate the current duplication between src/internal/packager2/layout and src/pkg/zoci. Currently, both create an ORAS remote and preform a package push. 

If `layout` called the ZOCI package to push packages to get rid of the duplication then there would be a cyclic import. The layout package creates packages, and therefore uses ZOCI to pull skeleton packages during component imports. The ZOCI package, in order to know which files to pull and push uses the PackageLayout object and constant fields in the layout package. See the diagram below for a visual of this:

Image

### Option 1

One way to solve this would be to break the layout package into three different packages:
- load - this package validates, resolves imports, and composes components to return a "loaded" Zarf package that's ready for create. 
- parse - this package parses Zarf packages and applies schema migrations.
- layout - this package contains the layout constants and PackageLayout Object, and assembles a package layout

This imports would then look like this: 

Image

### Pros: 
- By limiting the scope of the `layout` package to strictly during deploy, we limit the imports that it makes. Constants like `layout.ZARFYaml` or `layout.ComponentsDir` are used pretty heavily across the codebase. Limiting the scope of the `layout` package makes it less likely for future cyclic imports. 
- Better separation of concerns, the layout package shouldn't need to know how to do schema migrations for a zarf.yaml

### Option 2:
Move all of ZOCI into layout.

#### Pros:
- Simple & easy to implement.
#### Cons: 
- Layout is doing more, and importing more. There isn't a good separation of concerns. Also increases the likelihood of future cyclic import problems.  

### Option 3:

Place the layout constants and PackageLayout type in separate packages. Create an interface that `layout` adheres to so that zoci can use it without a cyclic import. 

Image

#### Pros: 
- The layout constants which are heavily used across Zarf, have no chance of a cyclic import. 
#### Cons: 
- The layout data, and Package Layout object would exist in separate packages, even though they're tightly coupled. 
- Create + Package Layout are still together, so any future package that create uses, can't use Package Layout without using an interface to avoid cyclic imports.

### Related Issue

Relates to https://github.com/zarf-dev/zarf/issues/3525

AustinAbro321:
## Description

Preforms Option1 in #3887

## Related Issue

Fixes #3887
Relates to #3525 

## Checklist before merging


src/internal/packager2/parse/load.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"context"
+
+	goyaml "github.com/goccy/go-yaml"
+	"github.com/zarf-dev/zarf/src/api/v1alpha1"
+	"github.com/zarf-dev/zarf/src/pkg/logger"
+)
+
+// ZarfPackage parses the yaml passed as a byte slice and applies potential schema migrations.
+func ZarfPackage(ctx context.Context, b []byte) (v1alpha1.ZarfPackage, error) {
```
mkcp: Does this need its own package? ParseZarfPackage seems to be the function name, maybe it can be a private func in load/parse.go?

src/internal/packager2/parse/deprecated.go
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2021-Present The Zarf Authors
 
-package layout
+package parse
```
mkcp: Similar to parse.ZarfPackage, I think these can be private functions in `load` - maybe `load/parse.go`?

src/internal/packager2/parse/load.go
AustinAbro321: I originally had the same thought, but it can't be in load since both `load` and `packageLayout` use `parse.ZarfPackage`. It would create a new cyclic dependency. 
Image

src/internal/packager/pkgcfg/load.go
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+// Package pkgcfg loads and applies schema migrations to zarf.yaml files.
+package pkgcfg
```
mkcp: nit: does this need to be nested in `packager`? They both work on v1alpha1.ZarfPackage but the behavior isn't necessarily _only_ useful for packager operations.

src/internal/packager/pkgcfg/load.go
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
```
mkcp: nit: This is a simple pkg, we can merge everything together. Rename load.go to `pkgcfg.go`

src/internal/packager/pkgcfg/deprecated_test.go
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
```
mkcp: nit: rename file to `pkgcfg_test.go`

src/internal/packager/pkgcfg/deprecated.go
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
```
mkcp: nit: merge `deprecated.go` file contents with `load.go` and rename to `pkgcfg.go`.

mkcp (changes requested):
a few nits to resolve then good to merge!

PR #3873, opened 2025-05-29
fix: unchecked err when populating variables for deploy

AustinAbro321:
## Description

packager2 has not been running golint. There will be a follow up PR to address this and fix every lint error, however this error struck me as something that could cause a bug on main, so I'm creating a quick fix now. 

## Checklist before merging


mkcp (approved):
LGTM

brandtkeller (approved):

PR #3882, opened 2025-06-02
refactor: move zoci into layout

AustinAbro321:
## Description
This executes on an alternative mentioned in #3872.

Goal is to break the dependency cycle with between src/internal/packager2/layout and src/pkg/zoci

## Checklist before merging


AustinAbro321:
Closing this in favor of #3872

PR #3888, opened 2025-06-03
refactor: packager2 api cleanup

AustinAbro321:
## Description

This changes some of the packager2 API before we make it public:

- Removes reliance on global state (besides `config.CommonOptions.TempDirectory`)
- Changes API so all required fields are parameters instead of option structs
- Remove result structs in favor of simple returns
- opt->opts where it wasn't already

This also catches a regression where `zarf package inspect manifests` was not passing components. It introduces a test to avoid this regression in the future

## Related Issue

Relates to #3525 

## Checklist before merging


src/internal/packager2/generate.go
```diff
@@ -21,30 +22,33 @@ import (
...
 	// Path to the Helm chart in the git repository
 	GitPath string
 	// Kube version to provide to the Helm chart
 	KubeVersion string
 }
 
 // Generate a Zarf package definition using information about a Helm chart.
-func Generate(ctx context.Context, opts *GenerateOptions) (pkg v1alpha1.ZarfPackage, err error) {
+func Generate(ctx context.Context, packageName, url, version string, opts GenerateOptions) (pkg v1alpha1.ZarfPackage, err error) {
+	if packageName == "" {
```
mkcp, line 33: good input validation

src/cmd/dev.go
```diff
@@ -314,6 +314,11 @@ func (o *devDeployOptions) run(cmd *cobra.Command, args []string) error {
 		OptionalComponents: pkgConfig.PkgOpts.OptionalComponents,
 		Timeout:            pkgConfig.DeployOpts.Timeout,
 		Retries:            pkgConfig.PkgOpts.Retries,
+		OCIConcurrency:     config.CommonOptions.OCIConcurrency,
+		RemoteOptions: packager2.RemoteOptions{
```
mkcp: Change requested:

The struct data
```
packager2.RemoteOptions{
		PlainHTTP:             config.CommonOptions.PlainHTTP,
		InsecureSkipTLSVerify: config.CommonOptions.InsecureSkipTLSVerify,
}
```
shows up many times in this PR, which indicates to me that there's a missing abstraction. This can be addressed with a simple function wrapper, e.g.: `func DefaultRemoteOptions() RemoteOptions {...}`
or maybe more contextual like `DefaultCLIRemoteOptions` or `DefaultTestRemoteOptions`.

Typically we'd declare the function next to the `RemoteOptions` type declaration (packager2.go). For the test data though, that would get declared in a `x_test.go` file like any other test helper. The `config.CommonOptions` values appear to be the default in many different contexts so i'd recommend the first suggestion: `DefaultRemoteOptions`.

There's an argument for removing the dependency on `config.CommonOptions` and keeping the default config data local to packager2, but the `config` pkg is a broader discussion that I don't think is in scope for this PR.

AustinAbro321: Yeah definitely agree on removing the dependency on config.CommonOptions or at least 99% of it. First I'm removing it from the SDK, and then I think we'll be primed to remove it from the CLI.

AustinAbro321: Gave the CLI and test each their own default function for remote options, lmk what you think

Issue #3629, opened 2025-04-02
Zarf package conflicts


cade-thomas:
Deploy two packages with the same name to two different namespaces and then look at zarf package list
the last one you deploy with override the existing one.

### Environment
App version: v0.49.0
Kubernetes distro being used: rke2
Other:

### Steps to reproduce
1.Create two namespaces and two zarf packages with the same name but targeting different namespaces 
2.  deploy a package A to namespace A with name A
3. deploy a package B to namespace B with name A
### Expected result
zarf tracks these as two separate packages so opps like zarf package remove does not have unexpected results
### Actual Result
zarf seems to track these as one package even with them being in two separate namespaces 
### Additional Context
You could do what k8s does with CRD names and make them fully qualified is some way like packageName.namespace.

AustinAbro321:
Packages can deploy to several namespaces so we can't qualify them by namespace. As far as I know there is not a way to qualify that a package being deployed is definitely a different package then a package already in cluster and not the next upgrade for the current package. 

Perhaps we could warn a user if their upgraded package does not contain the same components as the current package in the cluster before they deploy?

cade-thomas:
Just an idea but maybe somehow track the relationship between namespaces and packages. If a package is deployed and it covers multiple namespaces track that to some capacity the same way you would a package deploying to a single namespace. This way if two packages with the same name (but lets say different helm values) were deployed to the same cluster zarf could tell they originated from different packages and treat them separately to avoid conflicts with commands like zarf package remove

mkcp:
Related #2407

mkcp:
Related #3639

ntwkninja:
is this resolved with the implementation of [ZEP-0017](https://github.com/zarf-dev/proposals/pull/18)?

AustinAbro321:
@ntwkninja Not exactly. ZEP-0017 covers the case where you have the same package and you want to deploy it twice. However, a cluster could see two different Zarf packages that coincidentally have the same name. This issue is also one of the reasons that we made the requirement for namespace overrides to only be used for a package with one namespace. Otherwise, it gets tricky to track the package state.

https://github.com/zarf-dev/proposals/issues/26 will track improving this in Zarf.

PR #3889, opened 2025-06-03
feat(deploy): support for package namespace override flag

### LINKED ISSUE #2407: Ability to control deploy-to namespace using a Zarf Variable

I have to deploy the same zarf package more than once to the same cluster, using a different namespace each time. The number of deployments needed and namespace names are both not always known in advance.

- **Given** I have a Zarf variable called `NAMESPACE` and a zarf.yaml component like:

```yaml
components:
  - name: my-app
    required: true
    charts:
    - name: my-app
      localPath: ./
      namespace: "###ZARF_VAR_NAMESPACE###"
      version: 1.2.3
      valuesFiles:
      - values.yaml
```

- **When** `zarf package deploy --set NAMESPACE=foo` is run and followed by `zarf package deploy --set NAMESPACE=bar`
- **Then** "my-app" is deployed to the namespace "foo", followed by a 2nd instance of "my-app" being deployed to the namespace "bar".

- Using create-time variables, but that would require managing multiple built .tar.zst files which is undesirable
- Using multiple optional components, but that's (a) an ugly hack, and (b) requires both the namespace names as well as the maximum possible number of deployments to be known in advance.

brandtkeller:
## Description

This pull request adds support for a `--namespace` flag in the deploy operation - allowing users with a single-namespace package to deploy the package multiple times to multiple isolated namespaces. This is the implementation for proposal [ZEP-0017](https://github.com/zarf-dev/proposals/tree/main/0017-chart-namespace-overrides).

## Constraints
- This does not support package namespace overrides for `ZarfInitConfig` packages. 

## Design Choices
- Use of `ZarfPackage` methods for validating a package has only declared a single namespace & updating all package components to the target namespace
- Add `NamespaceOverride` as a new field under `DeployedPackage` 
  - Use of functional options for setting namespace - while it is exposed - helps provide a mechanism where the signature of existing cluster methods do not need to be modified. 
  - Exposing the `NamespaceOverride` field with json tags allows us to store it in the package state secret. Any call to `cluster.GetDeployedZarfPackages` will have the information innately.
  - Ensures secret delineation should not conflict by _not_ using `package name` solely for secret identification. 
  - formats overriden state secret with `zarf-package-my-package-override-namespacename` such that the collisions should only occur in a very specific edgecase.
- Logic/Testing for config file override is included
- Updates `DeployedPackage` and remaining "k8s" types to `state` package.
- Backwards compatible with older versions of Zarf - allowing for package list/inspect operations without panic. 

## Related Issue

Fixes #2407

## Checklist before merging


src/test/e2e/40_namespace_override_test.go
```diff
@@ -0,0 +1,67 @@
...
+	stdOut, stdErr, err = e2e.Zarf(t, "package", "create", "src/test/packages/40-namespace-override/multi-ns", "-o", tmpDir, "--confirm")
+	require.NoError(t, err, stdOut, stdErr)
+
+	// Attempt to deploy the bad package - should fail
+	stdOut, stdErr, err = e2e.Zarf(t, "package", "deploy", badPackage, "--namespace", "test4", "--confirm")
+	require.Error(t, err, stdOut, stdErr)
+
+	// Remove each zarf package
+	for _, pkg := range []string{"test-package", "test-package-test2", "test-package-test3"} {
+		stdOut, stdErr, err = e2e.Zarf(t, "package", "remove", pkg, "--confirm")
```
brandtkeller: Note: `zarf package remove` targeting the deployed package in the cluster state works as intended - this does introduce a gap where `zarf package remove` should also get a `--namespace` flag when using the tarball as the remove target. Thoughts?

src/api/v1alpha1/package.go
```diff
@@ -88,6 +90,48 @@ func (pkg ZarfPackage) IsSBOMAble() bool {
 	return false
 }
 
+// ValidateNamespaces checks to ensure the package has no more than the specified number of namespaces
+func (pkg *ZarfPackage) ValidateNamespaces(maxNamespaces int) error {
```
AustinAbro321: It feels like there is some business logic bleeding into this function. Out of context of the `--namespace` flag there is nothing invalid in having many namespaces. Maybe something like `getUniqueNamespaceCount` would be a more pure focus on the package type? 

Also we should make this a non-pointer method to keep line with the other methods.

src/config/lang/english.go
```diff
@@ -281,6 +281,7 @@ $ zarf package mirror-resources <your-package.tar.zst> --repos \
 	CmdPackageDeployValidateArchitectureErr            = "this package architecture is %s, but the target cluster only has the %s architecture(s). These architectures must be compatible when \"images\" are present"
 	CmdPackageDeployValidateLastNonBreakingVersionWarn = "The version of this Zarf binary '%s' is less than the LastNonBreakingVersion of '%s'. You may need to upgrade your Zarf version to at least '%s' to deploy this package"
 	CmdPackageDeployInvalidCLIVersionWarn              = "CLIVersion is set to '%s' which can cause issues with package creation and deployment. To avoid such issues, please set the value to the valid semantic version for this version of Zarf."
+	CmdPackageDeployFlagNamespace                      = "Target namespace for package deployment. "
```
AustinAbro321: We should add a note that the package must deploy to a single namespace  in order to use this flag

src/test/e2e/40_namespace_override_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNamespaceOverride(t *testing.T) {
+	t.Log("E2E: Namespace override")
+	tmpDir := t.TempDir()
+
+	goodPackage := fmt.Sprintf("%s/zarf-package-test-package-%s-0.0.1.tar.zst", tmpDir, e2e.Arch)
+	badPackage := fmt.Sprintf("%s/zarf-package-test-package-fail-%s-0.0.1.tar.zst", tmpDir, e2e.Arch)
```
AustinAbro321: nit: lets make this two separate tests. I know the good/bad package pattern is already present in our e2e tests, but IMO it makes things harder to read and switches contexts.

src/test/e2e/40_namespace_override_test.go
AustinAbro321: Good catch, and yeah I would agree the `--namespace` flag would make sense there

src/config/lang/english.go
AustinAbro321: I'm thinking we should also mark this feature with an [Alpha]. It would give room for a breaking change someday if we implement a system where several namespaces are fully dynamic.

brandtkeller: Agreed

src/api/v1alpha1/package.go
brandtkeller: Good catch - I agree with that sentiment.

src/test/e2e/40_namespace_override_test.go
brandtkeller: I'm going to look at that now - see if there are any implications.

AustinAbro321: One implication I'm thinking about is that re-using name in this way could lead to accidental state overrides. Let's say I have a package named `keycloak-dev`, and I have a different package `keycloak` that I deploy with `zarf package deploy <my-package> --namespace=dev` the packages would overwrite each other states. This is already a problem in Zarf (#3629), but this could exasperate the problem. 

Maybe we should have a more explicit mechanism for this field? Then add something like `namespace override` as a column in `zarf package list`. It would mean that everything that accepts a cluster source must also accept the `--namespace` flag, as of right now this is `zarf package inspect definition/images` and `zarf package remove`. It would also make it easier for users to tell which packages were deployed this way. What do you see as the pros/cons?

brandtkeller: That is a really good point and one I hadn't seen historical context on - thank you for bringing it up. 

I'm going to work around that information as it would very likely change my approach from the name alone to something more prescriptive.

brandtkeller: New process implemented to make all of override naming/storage/state more unique/explicit.

src/internal/packager2/deploy.go
```diff
@@ -50,6 +49,12 @@ type DeployOpts struct {
 	Retries int
 	// Number of layers to push concurrently per image
 	OCIConcurrency int
+	// Whether to use plainHTTP when connecting to the registry
+	PlainHTTP bool
+	// Whether or not to skipTLSVerify when connecting to the registry
+	InsecureTLSSkipVerify bool
```
AustinAbro321: These are already here with RemoteOptions

src/config/lang/english.go
```diff
@@ -281,17 +281,20 @@ $ zarf package mirror-resources <your-package.tar.zst> --repos \
...
+	CmdPackageDeployFlagNamespace                      = "[Alpha] Override the namespace for package deployment. Requires the package to have only one distinct namespace defined."
 
 	CmdPackageMirrorFlagComponents = "Comma-separated list of components to mirror.  This list will be respected regardless of a component's 'required' or 'default' status.  Globbing component names with '*' and deselecting components with a leading '-' are also supported."
 	CmdPackageMirrorFlagNoChecksum = "Turns off the addition of a checksum to image tags (as would be used by the Zarf Agent) while mirroring images."
 
 	CmdPackageInspectFlagSbomOut    = "Specify an output directory for the SBOMs from the inspected Zarf package"
 	CmdPackageInspectFlagListImages = "List images in the package (prints to stdout)"
+	CmdPackageInspectFlagNamespace  = "[Alpha] Override the namespace for package inspection. Requires the package to have been deployed using the namespace override."
```
AustinAbro321: Requires feels like the wrong verb here. Maybe something like `Applicable only to in-cluster packages deployed to a dynamic namespace`. Ditto for remove

src/cmd/package.go
```diff
@@ -285,7 +288,15 @@ func (o *packageDeployOptions) run(cmd *cobra.Command, args []string) (err error
...
-func deploy(ctx context.Context, pkgLayout *layout2.PackageLayout, opts packager2.DeployOpts) ([]types.DeployedComponent, error) {
+func deploy(ctx context.Context, pkgLayout *layout2.PackageLayout, opts packager2.DeployOpts) ([]state.DeployedComponent, error) {
+	// Update component namespaces here prior to confirmation when overriding
+	if opts.NamespaceOverride != "" {
+		nsCount := pkgLayout.Pkg.GetUniqueNamespaceCount()
+		if nsCount > 1 {
+			return nil, fmt.Errorf("package contains %d namespaces, cannot override namespace to %s", nsCount, opts.NamespaceOverride)
+		}
+		pkgLayout.Pkg.SetPackageNamespace(opts.NamespaceOverride)
+	}
```
AustinAbro321: I get why we validate and set the package namespace early here, but we should also validate and set the package namespace at the CLI level. Currently, Deploy accepts the `NamespaceOverride` opt, but if someone is calling the function purely at the SDK level, then it will only have an impact on the secret, not on the namespace the package is deployed to. To avoid duplication, one option would be to make a common function in packager2 both deploy and the cli can call.

src/cmd/package_test.go
AustinAbro321, line 1: Lets set the `namespaceOverride` field in one of these tests

src/internal/packager2/remove.go
```diff
@@ -22,17 +23,24 @@ import (
...
 	"github.com/zarf-dev/zarf/src/pkg/cluster"
-	"github.com/zarf-dev/zarf/src/types"
 )
 
 // RemoveOptions are the options for Remove.
 type RemoveOptions struct {
-	Cluster                 *cluster.Cluster
-	Filter                  filters.ComponentFilterStrategy
+	// Source is the source of the package
+	Source string
```
AustinAbro321: Remove already takes a source

src/internal/packager2/load.go
```diff
@@ -36,6 +37,8 @@ type LoadOptions struct {
 	SkipSignatureValidation bool
 	Filter                  filters.ComponentFilterStrategy
 	Output                  string
+	// [ALPHA] Namepsace override for package deployment
+	NamespaceOverride string
```
AustinAbro321: This should be a parameter on `cluster.GetDeployedPackage` or maybe that function should get it's own opts struct. Since namespaceOverrides is only relevant to cluster pulls we shouldn't make it an option here.

src/pkg/cluster/zarf.go
```diff
@@ -62,12 +61,19 @@ func (c *Cluster) GetDeployedZarfPackages(ctx context.Context) ([]types.Deployed
 
 // GetDeployedPackage gets the metadata information about the package name provided (if it exists in the cluster).
 // We determine what packages have been deployed to the cluster by looking for specific secrets in the Zarf namespace.
-func (c *Cluster) GetDeployedPackage(ctx context.Context, packageName string) (*types.DeployedPackage, error) {
-	secret, err := c.Clientset.CoreV1().Secrets(state.ZarfNamespaceName).Get(ctx, config.ZarfPackagePrefix+packageName, metav1.GetOptions{})
+func (c *Cluster) GetDeployedPackage(ctx context.Context, packageName string, opts ...state.DeployedPackageOptions) (*state.DeployedPackage, error) {
```
AustinAbro321, line 64: We should test some of the new logic here in `TestGetDeployedPackage`

AustinAbro321 (changes requested):
I like the change in direction, a few suggestions + calls for unit tests

src/cmd/package_test.go
brandtkeller: Added

src/internal/packager2/load.go
brandtkeller: Agreed - thought about options more but I didn't want to obfuscate LoadOptions or embed them if possible.

src/pkg/cluster/zarf.go
brandtkeller: Added

src/internal/packager2/remove.go
brandtkeller: merge conflict addition - corrected

src/config/lang/english.go
brandtkeller: That is fair! modified to potentially meet this intent.

src/cmd/package.go
brandtkeller: Great points - I tested a few different strategies for locality of the namespace override identifiers and ultimately landed on your thoughts as well. Isolating the logic in a common function and then duplicating across CLI/SDK to meet parity while primarily isolating the core logic to `DeployOptions` and `Deploy()`.

site/src/content/docs/commands/zarf_package_remove.md
```diff
@@ -24,6 +24,7 @@ zarf package remove { PACKAGE_SOURCE | PACKAGE_NAME } --confirm [flags]
       --components string           Comma-separated list of components to remove.  This list will be respected regardless of a component's 'required' or 'default' status.  Globbing component names with '*' and deselecting components with a leading '-' are also supported.
       --confirm                     REQUIRED. Confirm the removal action to prevent accidental deletions
   -h, --help                        help for remove
+  -n, --namespace string            [Alpha] Override the namespace for package removal. Applicable only to packages deployed using the namespace override flag.
```
AustinAbro321: If we're adding the `-n` shorthand we should add it to the other commands. Also we should just say the "namespace flag" instead of "namespace override flag" imo.

AustinAbro321 (changes requested):
One small nit, otherwise lgtm

Racer159:
:smiling_imp: 
image
image
image

brandtkeller:
@Racer159 you raised a good point that a constraint behind this is that it does not support init-packages to-date. We'd need to rewire all of the hardcoded assumptions around the `zarf` namespace, state, etc.

AustinAbro321:
Yeah excellent call out @Racer159. A similar concern is that other packages might have operators or actions expecting that the package deploys to a particular namespace. 

IMO we should provide a metadata field that allows users to enable/disable the ability to override the package namespace . Maybe `allowNamespaceOverride`. I'm undecided if `allowNamespaceOverride` should default to false or true, but I'm leaning towards false. On one hand this feature is not dangerous in the majority of situations since it's only applicable to single namespaces. On the other hand, this is an alpha feature and if we open up in the future to allowing X namespace overrides it could get more dangerous.

src/pkg/packager/packager2.go
```diff
@@ -53,3 +53,16 @@ func generateValuesOverrides(chart v1alpha1.ZarfChart, componentName string, var
 	// This corrects the logic to ensure that chartOverrides and valuesOverrides are merged correctly.
 	return helpers.MergeMapRecursive(chartOverrides, valuesOverrides), nil
 }
+
+// OverridePackageNamespace overrides the package namespace if the package contains only one unique namespace
+func OverridePackageNamespace(pkg v1alpha1.ZarfPackage, namespace string) error {
+	// disallow override on init packages while account for future kinds
+	if pkg.Kind != v1alpha1.ZarfPackageConfig {
```
brandtkeller, line 63: Added a check to match the current constraint for overrides not supporting InitPackages (yet).

AustinAbro321: What do you think of a metadata field on the Zarf package such as `allowNamespaceOverride`? It'd be unlikely but feasible that a user has a custom init package that could deploy to a single namespace. However, the more likely scenario that we should address is packages built with implicit knowledge of their namespace either through actions, an operator, or another mechanism. In those situations, we should give package creators a lever to turn off namespace overrides. 

If we go this route we should also consider if we want the field to default to false and force package creators to opt into it.  On one hand namespace overrides are not dangerous since it's only applicable to single namespace packages. On the other hand, this is an alpha feature and if we open up in the future to allowing namespace overrides in a package with multiple namespaces it could get messy.

brandtkeller: Valid suggestion - I appreciate consideration of having a mechanism to disable namespace overrides. I'd like to keep scope minimal as to warrant more feedback from the community.

Given this is not a security consideration (rather more UX) I am inclined to propose a default that allows namespace overrides to occur unless otherwise declared - preferably using the bool default value. 

`disableNamespaceOverride`?

AustinAbro321: FWIW we did get that feedback on the original proposal, though that was in the context of multiple configurable namespaces - https://github.com/zarf-dev/proposals/pull/18#discussion_r1952999991 

Yeah I'm fine with that. Generally we've been trying to avoid introducing fields in the schema starting with a negative (I.E. NoWait will become Wait in the next schema version). We can use boolptrs to make the default true or false. Though since we need a verb either way (I don't think just `namespaceOverrides` is clear enough) I don't feel too strongly on `allow` vs `disable`

brandtkeller: Good context and I do agree that it is a reasonable extension of scope - That thread might have concluded without getting the information ultimately included in the ZEP itself (Or I missed it when mapping requirements). 

words are hard and a non-negative wasn't easily identified - but I do like to stick to the style. Given that we can unmarshall to true by default and check for nil I think boolptr here is reasonable (albeit not an option I will typically reach for).

AustinAbro321: IMO we should get rid of the check on init packages in general and set the `allowNamespaceOverride` field as false in the init package zarf.yaml

src/pkg/packager/packager2.go
```diff
@@ -53,3 +53,21 @@ func generateValuesOverrides(chart v1alpha1.ZarfChart, componentName string, var
 	// This corrects the logic to ensure that chartOverrides and valuesOverrides are merged correctly.
 	return helpers.MergeMapRecursive(chartOverrides, valuesOverrides), nil
 }
+
+// OverridePackageNamespace overrides the package namespace if the package contains only one unique namespace
+func OverridePackageNamespace(pkg v1alpha1.ZarfPackage, namespace string) error {
+	if pkg.Metadata.AllowNamespaceOverride != nil {
```
AustinAbro321: We should make a boolean method like `v1alpha1.AllowNamespaceOverride()`, that way the logic of whether the field defaults to true lives with the package type and subsequent calls are simpler. 

We do something similar with `IsRequired` https://github.com/zarf-dev/zarf/blob/bc926ee32c09d7f4947701b3b142b92e89a239d7/src/api/v1alpha1/component.go#L90

src/pkg/message/connect.go
```diff
@@ -0,0 +1,27 @@
...
+package message
+
+import (
+	"fmt"
+
+	"github.com/zarf-dev/zarf/src/pkg/state"
+)
+
+// PrintConnectStringTable prints a table of connect strings.
+func PrintConnectStringTable(connectStrings state.ConnectStrings) {
```
AustinAbro321: I think this is leftover from a merge. We moved the `PrintConnectStringTable` function to src/cmd

src/pkg/packager/packager2.go
brandtkeller: I was thinking a similar thing and I did not land on a strong opinion. Yes those who deploy a custom init package _could_ attempt to override the namespace - it would fail whether they set `allowNamespaceOverride` to false _or_  they did not and it naturally fails because there is no support for the current hardcoded assumptions. 

I kept the explicit check for now as we know there is work required to support this AND it provides explicit feedback to the user regarding support rather than more ambiguous language of the process failing. Open to other opinions though.

src/api/v1alpha1/package.go
```diff
@@ -95,6 +95,33 @@ func (pkg ZarfPackage) IsSBOMAble() bool {
 	return false
 }
 
+// GetUniqueNamespaceCount returns the number of unique namespaces in the package.
+func (pkg ZarfPackage) GetUniqueNamespaceCount() int {
```
mkcp: nit: Get/set function names can usually be made more descriptive.

name -> `UniqueNamespaceCount` or return slice of unique namespaces and users len(`UniqueNamespaces`).

src/api/v1alpha1/package.go
```diff
@@ -95,6 +95,33 @@ func (pkg ZarfPackage) IsSBOMAble() bool {
...
+		}
+		for _, manifest := range component.Manifests {
+			uniqueNamespaces[manifest.Namespace] = struct{}{}
+		}
+	}
+	return len(uniqueNamespaces)
+}
+
+// SetPackageNamespace updates all existing namespaces to the provided one
+func (pkg ZarfPackage) SetPackageNamespace(namespace string) {
```
mkcp: nit: method exists on type `ZarfPackage` so no need to say it sets package namespaces. also getter/setter names

name -> `UpdateAllNamespaces` / `UpdateAllComponentNamespaces`

mkcp (commented):
added some nits, great overall direction!

src/pkg/packager/packager2.go
AustinAbro321: Yeah that's fair. Technically it's possible for a custom init package be built in such a way that namespaceOverrides won't break it, but it is super unlikely anyone's building an init package in that way. It'd be good feedback to hear if they are. 

I wouldn't mind both keeping the init check and setting the `allowNamespaceOverride` on the main zarf.yaml to be more explicit and provide an example of the flag being used, but I wouldn't block on it

brandtkeller: Agree %100 adding it for reference. The coupling of the init package components to potentially hardcoded resources (state namespace, webhook ClientConfig targets, etc) have me interested to see exactly what it would take. Certainly not impossible but warrants it's own issue/scope.

src/pkg/packager/packager2.go
brandtkeller: Appreciate this reference

src/pkg/message/connect.go
brandtkeller: good catch

src/api/v1alpha1/package.go
brandtkeller: thanks for the suggestion

src/api/v1alpha1/package.go
brandtkeller: thanks for the suggestion

AustinAbro321 (approved):
LGTM, great job

PR #3894, opened 2025-06-05
refactor: standardize packager2 API and decouple CLI from SDK

AustinAbro321:
## Description

Currently some functions such as deploy and mirror take a `layout.PackageLayout` object, whereas other functions like `InspectPackageResources` and `Remove` call `LoadPackage`. This PR standardizes the API so that functions on an already built package take a package layout. This reduces the number of fields in the Options struct for each relevant function, in an effort to make it more clear what each function does.

This PR also decouples the 1:1 relationship between SDK and CLI. For example, instead of `InspectPackageDefinition` being a wrapper around `GetPackageFromSourceOrCluster` the CLI calls `GetPackageFromSourceOrCluster` directly. Similarly, `packager2.MirrorImages` which currently only calls `pushImagesToRegistry` is deleted in favor of exposing `packager.PushImagesToRegistry` directly.

Also standardized option structs to end with Options over opts

Relates to #3525 

## Checklist before merging

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -50,6 +50,8 @@ linters:
         - name: error-strings
         - name: error-naming
         - name: exported
+          arguments:
+            - "disableStutteringCheck"
         - name: if-return
         - name: increment-decrement
         - name: var-naming
--- a/src/internal/packager2/layout/assemble.go
+++ b/src/internal/packager2/layout/assemble.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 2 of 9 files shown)


src/internal/packager2/deploy.go
```diff
@@ -333,33 +331,25 @@ func (d *deployer) deployComponent(ctx context.Context, pkgLayout *layout.Packag
...
-			RegistryInfo:          d.s.RegistryInfo,
-			ImageList:             refs,
-			PlainHTTP:             opts.PlainHTTP,
-			NoChecksum:            noImgChecksum,
-			Arch:                  pkgLayout.Pkg.Build.Architecture,
-			Retries:               opts.Retries,
-			InsecureSkipTLSVerify: opts.InsecureSkipTLSVerify,
-		}
-		err := images.Push(ctx, pushConfig)
+		imagePushOpts := ImagePushOptions{
```
brandtkeller, line 334: Appreciate how this cleans up the options and logic.

brandtkeller (approved):
I love how well this makes a lot of the work for Packager2 and surrounding SDK/CLI coalesce into an intuitive bundle. I've been reaching at times to think about what felt off about the current patterns and this PR really improved what was bothering me.

PR #3898, opened 2025-06-06
refactor: move split file logic to it's own package

AustinAbro321:
## Description

This moves the split file logic into it's own package, and stops the reliance on the types.ZarfPackageSplitData

## Checklist before merging

### DIFF
--- a/src/internal/split/split.go
+++ b/src/internal/split/split.go
@@ -142,8 +142,8 @@ func File(ctx context.Context, srcPath string, chunkSize int) (err error) {
 	return nil
 }
 
-// ReAssembleFile reconstructs many split files into a single file then deletes the split files
-func ReAssembleFile(src, dest string) (err error) {
+// ReassembleFile takes a directory containing split files, reassembles those files into the destination, then the split files.
+func ReassembleFile(src, dest string) (err error) {
 	pattern := strings.Replace(src, ".part000", ".part*", 1)
 	splitFiles, err := filepath.Glob(pattern)
 	if err != nil {
--- a/src/internal/split/split.go
+++ b/src/internal/split/split.go
@@ -1,0 +1,210 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+// Package split splits and re-assembles files
+package split
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/zarf-dev/zarf/src/pkg/logger"
+)
+
+// FileData contains info about a split file.
+type FileData struct {
+	// The sha256sum of the file
+	Sha256Sum string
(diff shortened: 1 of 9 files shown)


src/internal/split/split.go
```diff
@@ -0,0 +1,210 @@
...
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/zarf-dev/zarf/src/pkg/logger"
+)
+
+// FileData contains info about a split file.
+type FileData struct {
```
mkcp: nit: having "data" in a struct is a bit redundant, for a similar reason that we don't add suffixes to fields like "usernameStr string". File is fine here, but it does collide with the function name below.

src/internal/split/split_test.go
```diff
@@ -0,0 +1,120 @@
...
+			err = json.Unmarshal(b, &data)
+			require.NoError(t, err)
+			require.Equal(t, tt.expectedFileCount, data.Count)
+			require.Equal(t, int64(tt.fileSize), data.Bytes)
+			require.Equal(t, tt.expectedSha256Sum, data.Sha256Sum)
+		})
+	}
+}
+
+func TestSplitDeleteExistingFiles(t *testing.T) {
```
mkcp, line 98: No changes needed, but this test is a bit odd to me. It's more testing an implementation detail than the API itself. Not rly in scope for this refactor, but something for us to consider for the future.

src/internal/split/split_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSplitFile(t *testing.T) {
```
mkcp, line 16: 👀

src/internal/split/split.go
```diff
@@ -0,0 +1,210 @@
...
+	Sha256Sum string
+	// The size of the file in bytes
+	Bytes int64
+	// The number of parts the file is split into
+	Count int
+}
+
+// File splits a file into several parts. part000 always holds the split.FileData.
+// The remaining parts hold a chunkSize number of bytes of the original file.
+func File(ctx context.Context, srcPath string, chunkSize int) (err error) {
```
mkcp: nit: `File()` is contextualized by the package name here: e.g. `split.File(ctx, path, chunkSize)` but creates the type name collision above. I'm in favor of `split.Split()` or `split.SplitFile()` here. 

As an aside, functions named after data are fine, e.g. `File()`, `NewFile()`, `Files()`, or when the func _returned_ data of that kind fwiw. this func is just executed for side effects though)

src/internal/split/split.go
AustinAbro321: I ended up naming it `split.SplitFileMetadata`. I also added the rule to the linter `disableStutteringCheck`. In this package, I think the repetitive name is fine.

mkcp: Metadata is just data with a prefix on it, the same problem is there! :P

mkcp (approved):

PR #3901, opened 2025-06-09
feat: print action command output to std while cmd is running

AustinAbro321:
## Description

This changes actions to output the cmd std to the users terminal while the command is running. The command stdout/stdin is still output in a structured log at the debug level, but is now longer output in a structured log at the info level 

## Checklist before merging


mkcp (approved):
LGTM

PR #3902, opened 2025-06-09
chore!: delete packager and related dead code

AustinAbro321:
## Description

There will be a follow up after this PR to make src/internal/packager2 public so there is an SDK replacement for every packager function deleted here. 

This also deletes much of the message package and some other dead code

Relates to #3525 

## Checklist before merging


brandtkeller (approved):
No current outliers for removed code found in a manual review of this branch. Zarf binary matrix builds successfully.

PR #3904, opened 2025-06-10
feat: track progress on image pull/push

### LINKED ISSUE #3633: Give occasional progress updates during image pull and pushes
Zarf does not give progress for image pulls and pushes. Some images can be quite large especially as the AI industry often packs large models into images. There is a Kubernetes beta feature for image pod volumes. This will increase the likelihood of large images. As a design principal, we would like a user to have to go no more than 5-10 seconds without feedback from the system. Image pulls & pushes are most likely to be leave a user with no feedback.

- **Given** A user has a `zarf.yaml` with an image that takes over five seconds to download
- **When** they run `zarf package create` and it takes more than five seconds to download the image
- **Then** they see a log message that looks like below
` image download in progress... completion=x%, remaining=xGB

An alternative would be to re-introduce progress bars. There is an avenue to do this since the ORAS cli has progress bars while calling the same functions we call with oras-go, however log messages will be much simpler to implement and likely will get users most of what they are looking for.

AustinAbro321:
## Description

This introduces occasional log statements to give progress during OCI pulls and pushes (images and Zarf packages).

Implementation was loosely inspired by the progressbar implementation in the [ORAS CLI](https://github.com/oras-project/oras/tree/b28a30be912e9e3ae741b7a801f4110f96a1011f/internal/progress). Goal was to create simple a simple solution that avoids leaving users without feedback from the system for more than five seconds. UX was designed to be similar to how opentofu gives occasional progress feedback when something takes a while. 

Views:
image pull on create:
image
image push on deploy (lowered interval for screenshot because local pushes are fast):
image
package pull:
image
package publish:
image

## Related Issue

Fixes #3633

## Checklist before merging


src/internal/packager/images/pull.go
AustinAbro321, line 1: It's easier to review this with whitespace hidden

src/internal/packager/images/progress.go
```diff
@@ -0,0 +1,133 @@
...
+		percentComplete := float64(bytesRead) / float64(totalBytes) * 100
+		remaining := float64(totalBytes) - float64(bytesRead)
+		l.Info(msg, "name", imageName, "complete", fmt.Sprintf("%.1f%%", percentComplete), "remaining", utils.ByteFormat(remaining, 2))
+	}
+}
+
+const defaultProgressInterval = 5 * time.Second
+
+// StartReporting starts the reporting goroutine
+func (tt *TrackedTarget) StartReporting() {
```
brandtkeller: Given that the push is governed by the caller-supplied context - I wonder if this should accept context and if we could remove more of the bespoke `stopReports` in favor of a `ctx.Done()`

src/internal/packager/images/progress.go
```diff
@@ -0,0 +1,133 @@
...
+	"oras.land/oras-go/v2"
+)
+
+// Report defines a function to log progress
+type Report func(bytesRead, totalBytes int64)
+
+// DefaultReport returns a default report function
+func DefaultReport(l *slog.Logger, msg string, imageName string) Report {
+	return func(bytesRead, totalBytes int64) {
+		percentComplete := float64(bytesRead) / float64(totalBytes) * 100
```
brandtkeller, line 31: Minor - it may be worthwhile to check that `totalBytes` is not zero and what to do in that situation

src/internal/packager/images/progress.go
```diff
@@ -0,0 +1,133 @@
...
+		remaining := float64(totalBytes) - float64(bytesRead)
+		l.Info(msg, "name", imageName, "complete", fmt.Sprintf("%.1f%%", percentComplete), "remaining", utils.ByteFormat(remaining, 2))
+	}
+}
+
+const defaultProgressInterval = 5 * time.Second
+
+// StartReporting starts the reporting goroutine
```
brandtkeller: What happens if StartReporting is called twice? is there any potential state that needs to be tracked?

src/internal/packager/images/progress.go
```diff
@@ -0,0 +1,133 @@
...
+	reportInterval time.Duration
+	bytesRead      *atomic.Int64
+	totalBytes     int64
+
+	stopReports chan struct{}
+	wg          sync.WaitGroup
+}
+
+// NewTrackedTarget creates a new TrackedTarget
+func NewTrackedTarget(target oras.Target, totalBytes int64, reporter Report) *TrackedTarget {
```
brandtkeller, line 116: Is there any value in making interval configurable?

src/internal/packager/images/progress.go
```diff
@@ -0,0 +1,133 @@
...
+		reportInterval: defaultProgressInterval,
+		bytesRead:      &atomic.Int64{},
+		totalBytes:     totalBytes,
+		stopReports:    make(chan struct{}),
+	}
+}
+
+// Push wraps the target push method with an appropriate tracked reader.
+func (tt *TrackedTarget) Push(ctx context.Context, desc ocispec.Descriptor, content io.Reader) error {
+	tReader := &trackedReader{
```
brandtkeller, line 132: Maybe a nit - I think we could potentially do without the temporary `trackedReader` 

```
if wt, ok := content.(io.WriterTo); ok {
    tracked := &trackedWriterToReader{writerTo: wt, bytesRead: &tt.bytesRead}
    return tt.Target.Push(ctx, desc, tracked)
}
return tt.Target.Push(ctx, desc, &trackedReader{reader: content, bytesRead: &tt.bytesRead})
```

brandtkeller (commented):
I love improvements that hits the UX. Minor items that come to mind - also believe some of this could be potentially unit tested in some capacity?

src/internal/packager/images/progress.go
AustinAbro321: Fair enough, added a warn + return

src/internal/packager/images/progress.go
AustinAbro321: Yeah it should definitely accept a context. Though to get rid of the `StopReporting` call we'd have to make a separate context for `StartReporting` then cancel it. I like the feel of the `defer StopReporting` slightly more, but I don't feel too strongly either way

src/internal/packager/images/progress.go
AustinAbro321: I like embedding the trackedReader into the `trackedWritertoReader` struct this way we don't have to maintain another `Read` method on the `trackedWritertoReader`, it'll just use the `trackedReader` object.

src/internal/packager/images/progress.go
AustinAbro321: It would duplicate the log calls. Not sure if I should add a boolean like `reportingStarted` on the Tracker object and return if so. 

On one hand we wouldn't want duplicate logging calls to happen for the same object. On the other hand, there is no situation where `startReporting` should be called twice without `StopReporting` called inbetween, so having the check for duplicate calls makes it easier for mistakes like that to go unnoticed. LMK what you think.

src/internal/packager/images/progress.go
AustinAbro321: Maybe in the future, given that this is internal and we want the same interval everywhere. I didn't expose the interval

AustinAbro321:
@brandtkeller Separated out the Tracker object and created a unit test for it. I wanted to avoid creating a registry or mocking an `oras.target` for the test. I feel it's a reasonable test, but a bit jank. There might be way to better structure the code for easier unit tests, open to alternative ideas.

src/internal/packager/images/progress_test.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+// newTestTracker creates a Tracker instance with a customizable report interval for testing.
+func newTestTracker(bytesRead, totalBytes int64, reporter Report, interval time.Duration) *Tracker {
```
mkcp, line 14: ++

src/internal/packager/images/progress.go
brandtkeller: I think that is fair - my brain picks up on tickers for potential memory leaks but I remembered that this was resolved in [go 1.23](https://go.dev/doc/go1.23#timer-changes)

PR #3906, opened 2025-06-11
feat: expose packager(2) as public package

### LINKED ISSUE #3525: Packager Refactor - Parent Issue

AustinAbro321:
## Description

Renames packager2 and moves it out of internal

## Related Issue

Fixes #3525 

## Checklist before merging


mkcp (approved):
LGTM

PR #3907, opened 2025-06-11
refactor: pass cachepath through packager functions

AustinAbro321:
## Description

This passes the Zarf cache to all relevant functions on the SDK level rather than relying on global state. Long term we should also get rid of global state at the cmd level, but that's out of scope for this PR. 

## Checklist before merging


AustinAbro321:
Yeah I think there is room in both unit and e2e test to improve cache testing

PR #3910, opened 2025-06-12
feat(sdk): intuitive default path when using `./zarf` actions

AustinAbro321:
## Description

Currently if a library user calls `packager.Create` on a package with an action calling `./zarf` like this
```go
    actions:
      onCreate:
        before:
          - cmd: ./zarf version
```

The command will re run the go program calling the Zarf SDK. This is because actions will call `GetFinalExecutableCommand` to replace `./zarf`, and instead of replacing it with the zarfCLI it will replace it with whatever the go program running the Zarf SDK is. 

There is already an override for this with the `config.ActionsUseSystemZarf` global, but it is set to false by defualt. This PR updates it to be set to true by default, but set to false main() that way by default it will be true for library users and false for CLI users. 

## Checklist before merging

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,6 +20,7 @@ import (
 var zarfSchema embed.FS
 
 func main() {
+	// This ensures `./zarf` actions call the current Zarf binary over the system Zarf binary
 	config.ActionsUseSystemZarf = false
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
--- a/main.go
+++ b/main.go
@@ -12,6 +12,7 @@ import (
 	"syscall"
 
 	"github.com/zarf-dev/zarf/src/cmd"
+	"github.com/zarf-dev/zarf/src/config"
 	"github.com/zarf-dev/zarf/src/pkg/lint"
 )
 
@@ -19,6 +20,7 @@ import (
 var zarfSchema embed.FS
 
 func main() {
+	config.ActionsUseSystemZarf = false
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	signalCh := make(chan os.Signal, 1)
--- a/src/pkg/utils/io_test.go
+++ b/src/pkg/utils/io_test.go
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+// Package utils provides generic helper functions.
+package utils
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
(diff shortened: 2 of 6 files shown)


main.go
```diff
@@ -12,13 +12,16 @@ import (
...
 	"github.com/zarf-dev/zarf/src/cmd"
+	"github.com/zarf-dev/zarf/src/config"
 	"github.com/zarf-dev/zarf/src/pkg/lint"
 )
 
 //go:embed zarf.schema.json
 var zarfSchema embed.FS
 
 func main() {
+	// This ensures `./zarf` actions call the current Zarf binary over the system Zarf binary
```
mkcp, line 23: 👍

src/pkg/utils/io_test.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/zarf-dev/zarf/src/config"
+)
+
+// GetFinalExecutableCommand returns the final path to the Zarf executable including and library prefixes and overrides.
+func TestGetFinalExecutableCommand(t *testing.T) {
```
mkcp, line 16: good test

ivan-pinto:
>There is already an override for this with the config.ActionsUseSystemZarf global, but it is set to false by defualt. This PR updates it to be set to true by default, but set to false main() that way by default it will be true for library users and false for CLI users.

I believe this should be the other way around. A CLI user will require Zarf to be installed in its system and therefore the default here can be `true`, the library user has no such requirement (Zarf installed in its system) and should therefore be `false`.

AustinAbro321:
When `config.ActionsUseSystemZarf` global is false it calls the current application binary. We know this will work in the Zarf CLI as it will call the Zarf binary, however in a library application `./zarf` wouldn't call the Zarf binary, it would call the calling application. 
Take this code for example which sets `config.ActionsUseSystemZarf` to false. If you run this with a zarf.yaml if your cwd, and if it includes a `./zarf` action, it will re-run the application infinitely. 
```
//go:embed zarf.schema.json
var zarfSchema embed.FS

var cache = "/home/austin/.zarf-cache"

func main() {
	lint.ZarfSchema = zarfSchema
	config.ActionsUseSystemZarf = false
	definitionPath := "."
	ctx := logger.WithContext(context.Background(), logger.Default())
	_, err := packager.Create(ctx, definitionPath, "build", packager.CreateOptions{
		CachePath: cache,
		SkipSBOM:  true,
	})
	if err != nil {
		panic(err)
	}
}
```
image

You can still set the `config.ActionsUseSystemZarf` how you want, or set `config.ActionsCommandZarfPrefix` but I'm guessing `config.ActionsUseSystemZarf=false` default is not what most users will be looking for.

PR #3911, opened 2025-06-12
feat(sdk): return final package ref after publish

AustinAbro321:
## Description

Returns the final path after publish. This makes for a more intuitive experience on the SDK side, where you can publish your package, then use the new reference directly.

It bodes well for this PR that the tests consistently get simplified. Most of the time, in test, after we call this function we put together the ref to do more with it anyway. 

## Checklist before merging


mkcp (dismissed):
LGTM

mkcp (approved):

PR #3914, opened 2025-06-13
feat!(connect): support for listen address on zarf connect

### LINKED ISSUE #3913: Add all kubectl port-forward capabilities to zarf connect
`zarf connect` does not expose all of the capabilities of `kubectl port-forward`. Specifically I would like to be able to specify the listen address for a `zarf connect` so I can bind to more than just `127.0.0.1`

- **Given** A zarf package is deployed with a connect label
- **When** `zarf connect <package> --address 0.0.0.0` is ran 
- **Then** a kubectl pf is created that binds to all interfaces

brandtkeller:
## Description

Adds the ability to specify listening addresses for binding the tunnel connection. This allows users to specify other addresses than the hardcoded `localhost/127.0.0.1` in such a way as to bind to all interfaces for connectivity on the zarf CLI machine. 

IE `zarf connect <target> --address 0.0.0.0` to bind to all interfaces. 

Primarily this replaces the use of `portforward.New()` which has `localhost` hardcoded and calls `portforward.NewOnAddresses()` with directly calling `portfoward.NewOnAddresses()` and ensuring we pass in `localhost` everywhere except for when overriding.

Minor Note - added `ls` alias to `zarf connect list` command as this follows the same patterns seen elsewhere (`zarf package ls`)

The goal was minimal impact to tunnel function signature - that said, updates were required to support the tunnel endpoint methods now potentially returning many.

## Related Issue

Fixes #3913

## Checklist before merging


brandtkeller:
Moving to draft - going to align to `kubectl` for passing a comma-separated list of addresses as opposed to a single address.

src/cmd/internal.go
```diff
@@ -251,8 +251,9 @@ func (o *internalCreateReadOnlyGiteaUserOptions) run(cmd *cobra.Command, _ []str
 		return err
 	}
 	defer tunnel.Close()
-	tunnelURL := tunnel.HTTPEndpoint()
-	giteaClient, err := gitea.NewClient(tunnelURL, s.GitServer.PushUsername, s.GitServer.PushPassword)
+	// tunnel is created with the default listenAddress - there will only be one endpoint until otherwise supported
+	tunnelURLs := tunnel.HTTPEndpoints()
+	giteaClient, err := gitea.NewClient(tunnelURLs[0], s.GitServer.PushUsername, s.GitServer.PushPassword)
```
AustinAbro321, line 260: I think it would read better to have a function like `FirstEndpoint` for this, which errors if there isn't an endpoint. We will always have at least one based on how you setup the defaults, but this would just future proof it to 100% never panicing.

src/pkg/cluster/tunnel.go
```diff
@@ -49,12 +49,13 @@ const (
...
 type TunnelInfo struct {
-	LocalPort    int
-	RemotePort   int
-	Namespace    string
-	ResourceType string
-	ResourceName string
-	urlSuffix    string
+	LocalPort     int
+	RemotePort    int
+	ListenAddress []string
```
AustinAbro321: nit: this should probably be plural, `ListenAddresses`

src/cmd/internal.go
brandtkeller: I am good with making that addition - I felt retaining the `Endpoint()` method and simply returning the "first" arbitrarily as potentially deceiving (rather the user should decide which they want/need) - but the addition of a `FirstEndpoint()` is still reasonable.

brandtkeller: Thinking on this more - and the distinction between `tunnel.Endpoints()` and `tunnel.HTTPEndpoints()` and `tunnel.FullURLs()` -> I am a bit torn here. 

I don't think it makes sense to create tunnel.First*()` methods for each. For sake of error traceability I do think it may make sense to return errors on zero present - otherwise let the consumer consume the endpoints as they see fit.

AustinAbro321: Yeah I think that's reasonable. Though instead of an error I think we should check if the list is nil (go automatically treats lists of 0 length as nil)

brandtkeller: > instead of an error I think we should check if the list is nil (go automatically treats lists of 0 length as nil)

Can you elaborate? 

I generally understand the length check to be more idiomatic as it is safe for nil and non-nil states (such as empty non-nil states). 

But maybe you are suggesting something else such as not returning the error?

AustinAbro321: Sorry yeah I had some incorrect assumptions on how slices work and my suggestion was unclear. I'm suggesting that we only return a list then callers can error if the len is 0, before using the first item in the list.

brandtkeller: ah! I am good with that - change in progress.

PR #3934, opened 2025-06-21
chore: wrap mutex around appending to failed repository updates

joonas:
## Description

Appending to `repoFailList` from multiple goroutines is not concurrency safe, so let's wrap a `sync.Mutex` around the append.

## Related Issue

Fixes #

Relates to #

## Checklist before merging

### DIFF
--- a/src/cmd/helm/repo_update.go
+++ b/src/cmd/helm/repo_update.go
@@ -123,13 +123,16 @@ func updateCharts(repos []*repo.ChartRepository, out io.Writer, failOnRepoUpdate
 	fmt.Fprintln(out, "Hang tight while we grab the latest from your chart repositories...")
 	var wg sync.WaitGroup
 	var repoFailList []string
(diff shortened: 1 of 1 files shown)


brandtkeller:
I don't think this is an unreasonable change - upstream implemented [similar](https://github.com/helm/helm/pull/30958) improvements - going to tap @AustinAbro321 for zarf/helm context on this one as for what our sync plan is on these files from helm.

AustinAbro321:
Yeah, we got https://github.com/helm/helm/pull/13617 merged into Helm which will allow us to replace the quasi fork we have in Zarf with a simple call to Helm's `NewRootCommand`. That PR isn't getting backported into Helm 3 so we'll have to wait for Helm 4 (early 2025). In the meantime this seems like a reasonable addition. Thanks @joonas

AustinAbro321 (approved):

Issue #3817, opened 2025-05-17
Support `zarf init` into existing multi-architecture clusters


CJLove:
### Is your feature request related to a problem? Please describe.
Running `zarf init` with an existing multi-architecture cluster there can be failures when containers are deployed to nodes with an incompatible architecture.

Nodes in the cluster are labeled according to their architecture (e.g. `cputype`=amd64 or arm64)

### Describe the behavior you'd like

- **Given** a multi-architecture kubernetes clusters with nodes labeled
- **When** `zarf init` is performed 
- **Then** zarf successfully deploys its pods  into the cluster on compatible nodes 

### Describe alternatives you've considered
Given that nodes are labeled with their architecture it would be sufficient to somehow specify node selector information to the `zarf init` process. 

### Additional context
Beyond the initial zarf deployment it would be up to the user to create zarf packages with multi-arch support or with suitable node selection support in their helm charts.

a1994sc:
Just throwing my thoughts into this, but I wonder if an MVP would be to have the zarf pods only be scheduled on the appropriate architecture?

Example:
- when you have a AMD64 init package, the injection, docker registry, and agent include a 
  - ```yaml
    nodeSelector: 
      kubernetes.io/arch: "amd64"
    ```

Honestly I would love see multi-arch zarf packages

CJLove:
> Just throwing my thoughts into this, but I wonder if an MVP would be to have the zarf pods only be scheduled on the appropriate architecture?
> 
> Example:
> 
> * when you have a AMD64 init package, the injection, docker registry, and agent include a
>   
>   * nodeSelector: 
>       kubernetes.io/arch: "amd64"
> 
> Honestly I would love see multi-arch zarf packages

I agree, having a nodeSelector as you suggest would allow the `zarf init` to succeed on a multi-arch cluster.

brandtkeller:
Agree with the above sentiment - The init package should expose configuration that makes deployment more dynamic when required - while also considering source code logic changes that should use the architecture of the init package when performing selection of the injector image/node.

PR #3935, opened 2025-06-21
feat: add arch variable to variables

a1994sc:
## Description

This PR adds in the zarf variable `ZARF_ARCHITECTURE`, this helps with deploying the zarf package into a multi-arch cluster.

## Related Issue

Relates to #3817

## Checklist before merging

### DIFF
--- a/src/internal/packager/helm/zarf.go
+++ b/src/internal/packager/helm/zarf.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/zarf-dev/zarf/src/pkg/state"
-	"github.com/zarf-dev/zarf/src/pkg/variables"
 
 	"helm.sh/helm/v3/pkg/action"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -98,8 +97,6 @@ func UpdateZarfAgentValues(ctx context.Context, opts InstallUpgradeOptions) erro
 		return err
 	}
 
-	arch := deployment.Spec.Template.Spec.NodeSelector["kubernetes.io/arch"]
-
 	// List the releases to find the current agent release name.
 	listClient := action.NewList(actionConfig)
 	releases, err := listClient.Run()
@@ -131,11 +128,6 @@ func UpdateZarfAgentValues(ctx context.Context, opts InstallUpgradeOptions) erro
 			applicationTemplates, err := template.GetZarfTemplates(ctx, "zarf-agent", opts.State)
 			if err != nil {
 				return fmt.Errorf("error setting up the templates: %w", err)
-			}
-			if arch != "" {
-				applicationTemplates["###ZARF_PKG_ARCHITECTURE###"] = &variables.TextTemplate{
-					Value: arch,
-				}
 			}
 			opts.VariableConfig.SetApplicationTemplates(applicationTemplates)
 
--- a/src/internal/packager/helm/zarf.go
+++ b/src/internal/packager/helm/zarf.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/zarf-dev/zarf/src/pkg/state"
+	"github.com/zarf-dev/zarf/src/pkg/variables"
 
 	"helm.sh/helm/v3/pkg/action"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -97,6 +98,8 @@ func UpdateZarfAgentValues(ctx context.Context, opts InstallUpgradeOptions) erro
 		return err
 	}
 
+	arch := deployment.Spec.Template.Spec.NodeSelector["kubernetes.io/arch"]
+
 	// List the releases to find the current agent release name.
 	listClient := action.NewList(actionConfig)
 	releases, err := listClient.Run()
@@ -128,6 +131,11 @@ func UpdateZarfAgentValues(ctx context.Context, opts InstallUpgradeOptions) erro
 			applicationTemplates, err := template.GetZarfTemplates(ctx, "zarf-agent", opts.State)
 			if err != nil {
 				return fmt.Errorf("error setting up the templates: %w", err)
+			}
+			if arch != "" {
+				applicationTemplates["###ZARF_PKG_ARCHITECTURE###"] = &variables.TextTemplate{
+					Value: arch,
+				}
 			}
 			opts.VariableConfig.SetApplicationTemplates(applicationTemplates)
 
--- a/src/pkg/cluster/injector.go
+++ b/src/pkg/cluster/injector.go
@@ -263,7 +263,7 @@ func (c *Cluster) getInjectorImageAndNode(ctx context.Context, architecture stri
 		if err != nil {
 			return "", "", err
 		}
-		if nodeDetails.Status.NodeInfo.Architecture == "" || nodeDetails.Status.NodeInfo.Architecture != architecture {
+		if nodeDetails.Status.NodeInfo.Architecture != "" && nodeDetails.Status.NodeInfo.Architecture != architecture {
 			continue
 		}
 		if nodeDetails.Status.Allocatable.Cpu().Cmp(*resReq.Requests.Cpu()) < 0 ||
--- a/src/pkg/cluster/injector.go
+++ b/src/pkg/cluster/injector.go
@@ -36,7 +36,7 @@ import (
 )
 
 // StartInjection initializes a Zarf injection into the cluster.
-func (c *Cluster) StartInjection(ctx context.Context, tmpDir, imagesDir string, injectorSeedSrcs []string, registryNodePort int) error {
+func (c *Cluster) StartInjection(ctx context.Context, tmpDir, imagesDir string, architecture string, injectorSeedSrcs []string, registryNodePort int) error {
 	l := logger.From(ctx)
 	start := time.Now()
 	// Stop any previous running injection before starting.
@@ -56,7 +56,7 @@ func (c *Cluster) StartInjection(ctx context.Context, tmpDir, imagesDir string, 
 			corev1.ResourceCPU:    resource.MustParse("1"),
 			corev1.ResourceMemory: resource.MustParse("256Mi"),
 		})
-	injectorImage, injectorNodeName, err := c.getInjectorImageAndNode(ctx, resReq)
+	injectorImage, injectorNodeName, err := c.getInjectorImageAndNode(ctx, architecture, resReq)
 	if err != nil {
 		return err
 	}
@@ -245,7 +245,7 @@ func (c *Cluster) createPayloadConfigMaps(ctx context.Context, tmpDir, imagesDir
 }
 
 // getImagesAndNodesForInjection checks for images on schedulable nodes within a cluster.
-func (c *Cluster) getInjectorImageAndNode(ctx context.Context, resReq *v1ac.ResourceRequirementsApplyConfiguration) (string, string, error) {
+func (c *Cluster) getInjectorImageAndNode(ctx context.Context, architecture string, resReq *v1ac.ResourceRequirementsApplyConfiguration) (string, string, error) {
 	// Regex for Zarf seed image
 	zarfImageRegex, err := regexp.Compile(`(?m)^127\.0\.0\.1:`)
 	if err != nil {
@@ -262,6 +262,9 @@ func (c *Cluster) getInjectorImageAndNode(ctx context.Context, resReq *v1ac.Reso
 		nodeDetails, err := c.Clientset.CoreV1().Nodes().Get(ctx, pod.Spec.NodeName, metav1.GetOptions{})
 		if err != nil {
 			return "", "", err
+		}
+		if nodeDetails.Status.NodeInfo.Architecture != architecture {
+			continue
 		}
 		if nodeDetails.Status.Allocatable.Cpu().Cmp(*resReq.Requests.Cpu()) < 0 ||
 			nodeDetails.Status.Allocatable.Memory().Cmp(*resReq.Requests.Memory()) < 0 {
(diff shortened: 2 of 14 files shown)


a1994sc:
~I need to add logic to the injector to use the arch of the package.~

Added injector logic

a1994sc:
Hm, so I after reviewing the logs from the last e2e test I see that for the arch value was not being passed along correctly. I updated the logic such that the injector, docker registry, and gitea all get deployed.... however the agent gets deployed fine once, however during part of the `make test-e2e-with-cluster ARCH=amd64` the agent pods get into a pending state because the `kubernetes.io/arch` is set to `""`... any input would be super helpful

a1994sc:
image

a1994sc:
I updated this PR so now `ZARF_ARCHITECTURE` is only set for the life-cycle of the package deployment.

Edit:

Though, I am getting the following error when doing an end-2-end test:

```text
4s          Warning   FailedCreate        replicaset/source-controller-869c75dd44             Error creating: Internal error occurred: failed calling webhook "agent-pod.zarf.dev": failed to call webhook: Post "https://agent-hook.zarf.svc:443/mutate/pod?timeout=10s": tls: failed to verify certificate: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "ca.private.zarf.dev")
```

AustinAbro321:
This seems like it'll be a huge value add for Zarf, thanks for the PR. Once CI is passing, I'll try to test locally and send you an updated eks.yaml which would let us spin up a multi-arch cluster so we can have a proper e2e test.

a1994sc:
Yea no problem! This does not fully address having a truly multi-arch support, but is a step in that direction.

(A bit selfishly I would love to see zarf running a multi-arch cluster at home, RPI and intel NUC)

a1994sc:
Todo: update the variable to `ZARF_PKG_ARCHITECTURE` per convo on slack channel

ref: https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1752503299318649?thread_ts=1752502999.838289&cid=C03B6BJAUJ3

src/pkg/packager/deploy.go
```diff
@@ -626,6 +633,7 @@ func setupState(ctx context.Context, c *cluster.Cluster, pkg v1alpha1.ZarfPackag
 	if s == nil {
 		return nil, errors.New("cluster state should not be nil")
 	}
+	s.Architecture = pkg.Metadata.Architecture
```
AustinAbro321: I had not realized that this was a field on state. Looks like it was added at least a few years ago, but was not used. With how we're setting it here, it doesn't do anything. Another route we could take would be setting it in `InitState` and if state already exists then pulling it from the init package secret. But I think that's not necessary, at least as of now and we can delete this line.

src/pkg/packager/deploy.go
```diff
@@ -351,6 +351,13 @@ func (d *deployer) deployComponent(ctx context.Context, pkgLayout *layout.Packag
 	}
 
 	applicationTemplates, err := template.GetZarfTemplates(ctx, component.Name, d.s)
+
+	// This variable needs to be set from the package metadata architecture
+	// This might be able to be expanded to include all the .metadata settings
+	applicationTemplates["###ZARF_PKG_ARCHITECTURE###"] = &variables.TextTemplate{
+		Value: pkgLayout.Pkg.Metadata.Architecture,
```
AustinAbro321: Can you replace all the uses of `pkgLayout.Pkg.Metadata.Architecture` with `pkgLayout.Pkg.Build.Architecture`

src/test/e2e/25_helm_test.go
```diff
@@ -64,6 +64,10 @@ func testHelmChartsExample(t *testing.T) {
 	defer e2e.CleanFiles(t, fmt.Sprintf("zarf-package-helm-charts-local-tgz-%s-0.0.1.tar.zst", e2e.Arch))
 
 	// Create a package that needs dependencies
+	stdOut, stdErr, err = e2e.Zarf(t, "tools", "helm", "repo", "add", "gitlab", "https://charts.gitlab.io/")
```
AustinAbro321, line 67: I imagine the tests previously assumed someone already had this installed, in that case this is a good change

AustinAbro321 (changes requested):
A few small notes, I will make it an action item on my side to get you a new eks.yaml that can properly test a multi-node cluster with Zarf

src/test/e2e/25_helm_test.go
a1994sc: Yea, as far as I can tell it might be included in the github runners, but I was failing at this locally...

Helm does not return an error code when it already has a registry with the name so this should be a fairly safe thing to keep in

src/pkg/packager/deploy.go
a1994sc: Is that because that there is no guarantee that a package will have a `.metadata.arch` ?

src/pkg/packager/deploy.go
a1994sc: That sounds good to me.

src/pkg/cluster/injector.go
AustinAbro321, line 1: Because of the way we manually find a node to select the injector with, this label actually won't do anything. Instead you should use `nodeDetails.Status.NodeInfo.Architecture`

AustinAbro321 (changes requested):
Realized something after https://github.com/zarf-dev/zarf/actions/runs/16351367206/job/46203688496 failed

src/pkg/packager/deploy.go
AustinAbro321: It looks like we always make them equal, at least we do now, but I think it also just reads a bit better. We know this is set by the system and not the user since it's on deploy.

src/pkg/cluster/injector.go
a1994sc: @AustinAbro321 I updated this to use `nodeDetails.Status.NodeInfo.Architecture`. I had to rebase the branch to accept the changes in #4004 so let me know if you need any help with updating #4001

src/pkg/cluster/injector.go
```diff
@@ -263,6 +263,9 @@ func (c *Cluster) getInjectorImageAndNode(ctx context.Context, resReq *v1ac.Reso
 		if err != nil {
 			return "", "", err
 		}
+		if nodeDetails.Status.NodeInfo.Architecture != architecture {
```
AustinAbro321: Out of an abundance of caution we should not continue if `nodeDetails.Status.NodeInfo.Architecture`=="". Hopefully that never happens, but I don't want to break anybody if they're on a cluster where this isn't set for some reason.

a1994sc: So I believe that this is a result of using a bare-bones "cluster" for tests. I believe that a real node will always have an Architecture.... I will do some research to verify that though

a1994sc: Sorry I meant to leave this sooner:

image

https://kubernetes.slack.com/archives/C0BP8PW9G/p1753214412824019

The above convo was from the #sig-node channel on the k8s slack.... My understanding is that unless someone crafts a malicious node, the architecture field will be populated.

AustinAbro321: There does seem to be a pretty strong level of confidence here, still I think I'd lean towards checking for the empty string just in case. Fair chance it never comes up, but it doesn't cost us anything

a1994sc: Sure, I think that should be good to add as a sanity check. I can add that in this evening

a1994sc:
@AustinAbro321 is there anything else I can do for this PR?

AustinAbro321:
@a1994sc Apology for the delay, I realized there was a gap in how our AWS infra was setup that required some changes to allow PRs to make EKS clusters. I have a workflow in .github/workflows/nightly-eks-multi-arch.yml that should run through and do a multi arch test. Could you copy that file, minus line 6 that makes it run on pull requests. Workflow is running now, it's passed before I just made some small changes to the structure of the file.

src/pkg/cluster/injector.go
```diff
@@ -263,6 +263,9 @@ func (c *Cluster) getInjectorImageAndNode(ctx context.Context, resReq *v1ac.Reso
 		if err != nil {
 			return "", "", err
 		}
+		if nodeDetails.Status.NodeInfo.Architecture == "" || nodeDetails.Status.NodeInfo.Architecture != architecture {
+			continue
+		}
```
AustinAbro321:
```suggestion
		if nodeDetails.Status.NodeInfo.Architecture != "" && nodeDetails.Status.NodeInfo.Architecture != architecture {
			continue
		}
```
Suggestion to change the behavior to not continue if architecture is empty. That way on the off chance there's some distro or version that doesn't fill in this value, nothing breaks for existing users

a1994sc: I will have to review, but I think that there was an issue with the injector logic not finding the right node with the suggested config.... but I will review and see if I had some bad logic before

src/internal/packager/helm/zarf.go
```diff
@@ -129,6 +132,11 @@ func UpdateZarfAgentValues(ctx context.Context, opts InstallUpgradeOptions) erro
 			if err != nil {
 				return fmt.Errorf("error setting up the templates: %w", err)
 			}
+			if arch != "" {
```
AustinAbro321: Looking at this code closer, I think we should move this template into  `GetZarfTempaltes` so it's consistent with the rest of the Zarf templates. Architecture can be set on state then move through the system like that. Sorry, I think you may have had that originally and I asked you to change it.

a1994sc: Actually, this specific code might be able to be removed here, as it was making sure the agent pods get scheduled on the right arch. Though with pr #3678 changing it over to a helm chart, I think we can safely remove this.... as even with an upgrade from an older version, the arch is still injected with this section:

https://github.com/zarf-dev/zarf/blob/103cedc4276abd9d16dc6f0fcf8b7136785f0186/packages/zarf-agent/chart/templates/deployment.yaml#L19-L20

As for moving this back into `GetZarfTemplates` I think that should be fine to be honest, I have something in my head that was indicating it might have been an issue. I will see if I can A) remember and report back or B) update the PR to have back in `GetZarfTemplates`

a1994sc: @AustinAbro321 updated the PR per your comments, so please let me know if there is anything else that needs to be addressed

AustinAbro321:
Confirmed locally the test-upgrade failure is not a flake. If we go from an older `zarf init` to the `zarf init` on this branch then it errors.

a1994sc:
I got bogged down by org stuff, so I will see about identifying the upgrade issues this week

a1994sc:
@AustinAbro321 would expanding `state.Merge()` and writing that new state to the `zarf-state` secret be the correct way to set the `architecture`?

https://github.com/zarf-dev/zarf/blob/b337191b08434776a31a9371bc670373a5a541cf/src/pkg/state/state.go#L269-L277

AustinAbro321:
@a1994sc Since state.Merge essentially exists for update-creds we shouldn't make an update there. In that case we can assume the architecture in state will be the same as before since a new package is not being deployed. Only time it'd ever change is if someone in a multi architecture cluster deploys one version of the init package with one architecture then deploys another version of the init packager with a different architecture. 

The reason the upgrade test is failing currently is because architecture is only set when state is nil. We need to set architecture outside of the nil check so that is always set even on subsequent init calls. Otherwise we will see this in the podspec which will never match to a node. 
```yaml
  nodeSelector:
    kubernetes.io/arch: ""
```

a1994sc:
Ah ok, well is there an existing way to write the architecture of the package to the state? And have it persist? I have tried:

```golang
		if d.s.Architecture == "" {
			d.s.Architecture = pkgLayout.Pkg.Build.Architecture
		}
```

But that does not persist in the `zarf-state` secret 

edit: add "real" code example

a1994sc:
@AustinAbro321 I believe that I fixed the issue with this PR. Can you, or another maintainer, run the tests to verify?

AustinAbro321:
I'm thinking about the complexity of this PR, and also about a recent issue we got https://github.com/zarf-dev/zarf/issues/4168 which would somewhat conflict with this PR as written. 

I would recommend taking a simpler approach here and adding a Zarf variable called NODE_SELECTOR to each of our init package charts. Zarf would no longer automatically set the node selector, the user would need to use the `--set` flag. I think the manual step for a small percentage of users is worth it to lower the complexity in Zarf. The only go changes with this approach would be with the injector

a1994sc:
Hey @AustinAbro321 , I feel like the spirit of this PR would be better served long term by either PR #4136 or an extension of that feature what exposes the various metadata variables from the package information.

Can you please let me know your thoughts?

AustinAbro321:
@a1994sc Agreed. I think a strategy where we expose NODE_SELECTOR as a variable would be best. I don't think we always want it there automatically as we'll likely implement something similar to https://github.com/zarf-dev/zarf/issues/4168, but users having the option to add a node selector will be good.

a1994sc:
Alright, if I'm understanding you right @AustinAbro321 here is what I would like to do for next steps:
- Close this PR
- Open a new PR that exposes the `nodeSelector` field for both the agent and registry helm charts.

Does that wound good to you?

AustinAbro321:
@a1994sc Yup and could be useful for other use cases too. The only thing is that it still wouldn't fully cover this use case since the injector could still start on a node with the wrong architecture, but I think exposing node selector is a good start.

a1994sc:
Hm, what I can do as a little bit of a middle ground, is to keep the logic to check the arch of pkg versus the node, so that at we can only schedule on compatible nodes

PR #3956, opened 2025-07-01
feat: implement a registry proxy

AustinAbro321:
## Description

WIP! Using this as a testing ground to setup some testing and see how this works with different distros

## Checklist before merging


packages/zarf-registry/chart/templates/injector-daemonset.yaml
```diff
@@ -0,0 +1,86 @@
...
+            periodSeconds: 2
+            successThreshold: 1
+            timeoutSeconds: 1
+          volumeMounts:
+            - mountPath: /zarf-init/zarf-injector
+              name: init
+              subPath: zarf-injector
+            - mountPath: /zarf-seed
+              name: seed
+            {{- range $i := until (int .Values.proxy.injector.payLoadConfigMapAmount) }}
```
brandtkeller, line 54: broader opportunity here - [Projected Volumes](https://kubernetes.io/docs/concepts/storage/projected-volumes) offer a more manageable way to combine many configmaps into a single volume. instead of N Volume mounts and N volumes for the configmaps - it would be 1 volumeMount and 1 Projected volume that defines the N configmaps. 

I think we're still a ways from the podspec hitting the maximum length - but this could clean-up the approach.

brandtkeller (commented):
Incremental thought - working through code review and local testing

brandtkeller:
Kind of a broad question - so the result may be out of scope but still pertinent for discussion. 

Do we still need a seed registry? 

As I understand it - the proxy is pulling from the injector - which is now a persistent daemonset. This is actually more resilient in practice. 

The zarf-docker-registry still pulls from the seed registry. I'm curious if we could/should look at the implications here and coalesce on a more resilient strategy.

packages/zarf-registry/chart/templates/injector-daemonset.yaml
AustinAbro321: Neat idea. I'd be interested in exploring this as a future enhancement. I feel like I need to understand projected volumes better to understand the benefit we'd gain from the switch. I also want to keep the injector daemonset and the regular injector in sync and don't want to make this PR too big.

AustinAbro321:
@brandtkeller Great question. The seed registry is referring to the injector which in the registry proxy case is long lived. The component `zarf-seed-registry` is looked at in go code and with the nodeport solution it first spins up the injector, with the registry proxy solution it adds the injector configmaps to the cluster

This is where it gets tricky, the next Helm install is in the `zarf-registry` component. This Helm install is almost the same, but it changes the registry image so that the original registry can point to itself rather than the seed registry. You are correct in that when the registry proxy is used it's unnecessary and less resilient to make the registry point to itself. 

The simple option is to dynamically set the registry image to use the seed registry when `.proxy.enabled` is true. 

A more complex option could be to look at maintaining two init packages. Though as they are right now I don't feel very confident in the custom init package lifecycle. I think having a separate package / chart would also make it tough to go from nodeport to registry proxy on an already initialized cluster.

brandtkeller:
> @brandtkeller Great question. The seed registry is referring to the injector which in the registry proxy case is long lived. The component `zarf-seed-registry` is looked at in go code and with the nodeport solution it first spins up the injector, with the registry proxy solution it adds the injector configmaps to the cluster
> 
> This is where it gets tricky, the next Helm install is in the `zarf-registry` component. This Helm install is almost the same, but it changes the registry image so that the original registry can point to itself rather than the seed registry. You are correct in that when the registry proxy is used it's unnecessary and less resilient to make the registry point to itself.
> 
> The simple option is to dynamically set the registry image to use the seed registry when `.proxy.enabled` is true.
> 
> A more complex option could be to look at maintaining two init packages. Though as they are right now I don't feel very confident in the custom init package lifecycle. I think having a separate package / chart would also make it tough to go from nodeport to registry proxy on an already initialized cluster.

Great distinctions here - I am certainly interested in seeing if we can evaluate the true need of the seed-registry in the long term. This solution as I understand it has the proxy image pull from the injector (which is great - injector is now persistent in this model) but still relies on the seed registry to reset the image. a better behavior would be for the injector to always serve both the proxy and the registry and skip the seed registry - but this might lead to the afformentioned need to maintain multiple init packages... why would this be?

src/cmd/initialize.go
```diff
@@ -57,6 +57,13 @@ func newInitCommand() *cobra.Command {
 	cmd.Flags().StringVar(&pkgConfig.PkgOpts.OptionalComponents, "components", v.GetString(VInitComponents), lang.CmdInitFlagComponents)
 	cmd.Flags().StringVar(&pkgConfig.InitOpts.StorageClass, "storage-class", v.GetString(VInitStorageClass), lang.CmdInitFlagStorageClass)
 
+	cmd.Flags().BoolVar(&pkgConfig.InitOpts.RegistryInfo.ProxyMode, "registry-proxy", false, "connect to the Zarf registry over a DaemonSet proxy")
+	cmd.Flags().IntVar(&pkgConfig.InitOpts.SeedRegistryHostPort, "injector-hostport", v.GetInt(VInitSeedRegistryHostPort),
+		"the hostport that the long lived DaemonSet injector will use when the registry is running in proxy mode")
+	// While this feature is in early alpha we will hide the flags
+	cmd.Flags().MarkHidden("registry-proxy")
```
brandtkeller: Does our model for feature flag implementation support more dynamic options for flag declaration? 

IE a lookup for `--features=registry-proxy=true` being set before that flags are even registered?

src/cmd/initialize.go
```diff
@@ -146,6 +158,7 @@ func (o *initOptions) run(cmd *cobra.Command, _ []string) error {
 		OCIConcurrency:         config.CommonOptions.OCIConcurrency,
 		SetVariables:           pkgConfig.PkgOpts.SetVariables,
 		StorageClass:           pkgConfig.InitOpts.StorageClass,
+		SeedRegistryHostPort:   pkgConfig.InitOpts.SeedRegistryHostPort,
```
brandtkeller: we're setting this from the `--injector-hostport` flag - but there is a clear distinction in zarf terminology between the injector and the seed registry correct?

src/internal/feature/feature.go
```diff
@@ -225,6 +226,13 @@ func init() {
 			Since:   "v0.60.0",
 			Stage:   Alpha,
 		},
+		{
+			Name:        RegistryProxy,
+			Description: "Enables setting Zarf init with the registry proxy flag",
+			Enabled:     false,
+			Since:       "v0.60.0",
```
brandtkeller: Just flagging - Update prior to merge.

src/pkg/cluster/cluster.go
```diff
@@ -331,3 +343,71 @@ func (c *Cluster) SaveState(ctx context.Context, s *state.State) error {
 	}
 	return nil
 }
+
+// GetIPFamily returns the IP family of the cluster, can be ipv4, ipv6, or dual.
+func (c *Cluster) GetIPFamily(ctx context.Context) (_ state.IPFamily, err error) {
```
brandtkeller, line 357: ++ appreciative of how pragmatic this approach is given the potential variables involved with other methods of determination.

src/pkg/cluster/cluster.go
```diff
@@ -331,3 +343,71 @@ func (c *Cluster) SaveState(ctx context.Context, s *state.State) error {
...
+	for _, family := range ipFamilies {
+		switch family {
+		case corev1.IPv4Protocol:
+			hasIPv4 = true
+		case corev1.IPv6Protocol:
+			hasIPv6 = true
+		}
+	}
+
+	if hasIPv4 && hasIPv6 {
```
brandtkeller: I believe this is pretty strong but makes an assumption that it has both _and_ both have been assigned. It leaves the latter some room for error that we may want to consider.

.github/workflows/test-proxy.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Download build artifacts
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+        with:
+          name: build-artifacts
+          path: build/
+
+      - name: setup kind config
```
brandtkeller, line 75: do we have resources available to emulate a multi-node cluster? given daemonsets - I'd be interested in validating multi-node regularly.

brandtkeller (commented):
continuing with some questions - mapping the daemonset image and ports currently.

src/cmd/initialize.go
AustinAbro321: Yeah you're right it seems like the seed registry represents the initial install distribution:v3 registry before it's upgraded to point back to itself. Updated terminology to use injector

src/pkg/cluster/cluster.go
AustinAbro321: I believe it will be fine given that the service sets the field `.spec.IPFamilyPolicy` to `preferDualStack`, which will assign on both stacks if dual stack is available https://kubernetes.io/docs/concepts/services-networking/dual-stack/#services

.github/workflows/test-proxy.yml
AustinAbro321: Great suggestion, I believe we will have the resources. Updated the kind-config so we will see

src/cmd/initialize.go
AustinAbro321: No it doesn't, and I'm not sure cobra will allow for it. Though I prefer hiding flags as a manual action, I think sometimes we'll want flags hidden for alpha features and sometimes we won't

.github/workflows/test-proxy.yml
AustinAbro321: Looks like multi node works fine in actions.

AustinAbro321:
> a better behavior would be for the injector to always serve both the proxy and the registry and skip the seed registry - but this might lead to the afformentioned need to maintain multiple init packages... why would this be?

@brandtkeller  FWIW in this model I've changed the registry deployment to always point to the injector when the proxy is in use since the injector is still long lived. However, since we still have the seed registry component in the base zarf.yaml we do an unnecessary Helm upgrade. If we had a separate package we could spin up the registry in this model without the seed registry, but we'd have to have another root zarf.yaml that doesn't declare the seed registry component

packages/zarf-registry/chart/templates/service.yaml
```diff
@@ -9,13 +9,20 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 spec:
-  type: {{ .Values.service.type }}
+  {{- if .Values.proxy.enabled }}
+  type: ClusterIP
+  {{- else }}
+  type: NodePort
+  {{- end }}
+  ipFamilyPolicy: PreferDualStack
```
soltysh, line 17: What this buys us?

src/cmd/initialize.go
soltysh: > IE a lookup for `--features=registry-proxy=true` being set before that flags are even registered?

Theoretically that's achievable, but I think it's up to the author to decide whether one wants to expose a flag or not. Also, you'd most likely fall into manually parsing flags, since we're constructing commands prior to any parsing, so too much overhead for the outcome.

The one thing I'd definitely do is to add an explicit information in the description of a flag, that this is alpha functionality which requires setting `--features=registry-proxy=true`.

src/pkg/cluster/cluster.go
```diff
@@ -331,3 +343,71 @@ func (c *Cluster) SaveState(ctx context.Context, s *state.State) error {
...
+	hasIPv6 := false
+
+	for _, family := range ipFamilies {
+		switch family {
+		case corev1.IPv4Protocol:
+			hasIPv4 = true
+		case corev1.IPv6Protocol:
+			hasIPv6 = true
+		}
+	}
```
soltysh: How about:
```golang
hasIPV4 := slices.Contains(ipFamilies, IPv4Protocol)
hasIPV6 := slices.Contains(ipFamilies, IPv6Protocol)
```
instead of the entire block above?

src/pkg/cluster/cluster.go
```diff
@@ -331,3 +343,71 @@ func (c *Cluster) SaveState(ctx context.Context, s *state.State) error {
...
+
+	if hasIPv4 && hasIPv6 {
+		return state.IPFamilyDualStack, nil
+	} else if hasIPv6 {
+		return state.IPFamilyIPv6, nil
+	} else if hasIPv4 {
+		return state.IPFamilyIPv4, nil
+	}
```
soltysh, line 413: Nit, but in the above case something like:
```golang
switch {
case hasIPv4 && hasIPv6:
    return state.IPFamilyDualStack, nil
case hasIPv6:
    return state.IPFamilyIPv6, nil
case hasIPv4:
    return state.IPFamilyIPv4, nil
default 
    return "", fmt.Errorf("unable to determine IP family of cluster")
}
```

src/pkg/cluster/injector.go
```diff
@@ -292,6 +296,81 @@ func (c *Cluster) getInjectorImageAndNode(ctx context.Context, resReq *v1ac.Reso
...
 
+// GetInjectorDaemonsetImage gets the image that is most likely to be accessible from all nodes
+// It first grabs the smallest image with pause in the name. This should be the pause container which every node must have access to
+// If there are no pause images then it grabs the smallest image.
+func (c *Cluster) GetInjectorDaemonsetImage(ctx context.Context) (string, error) {
+	l := logger.From(ctx)
+
+	var injectorImage string
+	err := retry.Do(func() error {
+		nodes, err := c.Clientset.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
```
soltysh, line 363: This might be expensive query to perform, how about just limiting this to control-plane nodes? iow. with `node-role.kubernetes.io/control-plane` labels?

src/pkg/state/state.go
```diff
@@ -185,28 +215,36 @@ type RegistryInfo struct {
 	PullPassword string `json:"pullPassword"`
 	// URL address of the registry
 	Address string `json:"address"`
-	// Nodeport of the registry. Only needed if the registry is running inside the kubernetes cluster
+	// Nodeport of the registry. Only needed if the internal Zarf registry is used and connected with over a nodeport service.
 	NodePort int `json:"nodePort"`
 	// Secret value that the registry was seeded with
 	Secret string `json:"secret"`
+	// ProxyMode is true if the registry made available through a DaemonSet proxy.
+	ProxyMode bool `json:"proxyMode"`
```
soltysh: Long term, I'm not 100% sure we won't add another mode for injector. So to look forward, I'm inclined to say that rather than introducing one-off boolean, it'll be more preferable to introduce InjectorMode or something like that, and that would accept values such as `NodePort` (the current default) and the new `Proxy`. `""` (empty) being identical to `NodePort`. This way most of the code you're writing in this PR will need to be change to use `switch` statements rather than ifs, which would make it easier to add additional modes in the future, if we decide to do so.

soltysh (commented):
I've left several comments, the biggest one being to switch away from `ProxyMode` being boolean and introduce an enum representing various modes we support. 

I haven't played with the functionality, yet, I'm planning to go back there tomorrow and provide more feedback.

src/pkg/state/state.go
AustinAbro321: Great idea, it makes the code more readable and future proof. Updated

packages/zarf-registry/chart/templates/service.yaml
AustinAbro321: In a dual stack cluster, I think it's possible that this would choose IPv6, but as far as I can tell you can only tell socat to listen to one stack like below. When the chart is dual stack if this service where to pick IPv6, then socat wouldn't be able to pull from it
```bash
            - tcp4-listen:{{ .Values.service.nodePort }},reuseaddr,fork
            - tcp4:{{ template "docker-registry.fullname" . }}:5000
```

src/pkg/cluster/injector.go
AustinAbro321: How expensive are you thinking? Given that this only happens once during `zarf init` I think an extra 10-20 seconds would be fine for a more accurate image selection.

src/pkg/cluster/injector.go
```diff
@@ -35,18 +35,25 @@ import (
 	v1ac "k8s.io/client-go/applyconfigurations/core/v1"
 )
 
+var zarfImageRegex = regexp.MustCompile(`(?m)^(127\.0\.0\.1|\[::1\]):`)
+
 // StartInjection initializes a Zarf injection into the cluster.
```
brandtkeller: nit: doccomments defining context for the returned integer

src/pkg/cluster/injector.go
brandtkeller: Also curious about the constraints or limitations present with only looking at the control-plane (IE thinking managed kubernetes). 

More broadly I am curious how this responds across Kubernetes upgrades... as I imagine there is no guarantee of common image references. Do we know how we expect this to respond?

src/pkg/cluster/injector_test.go
```diff
@@ -292,3 +294,144 @@ func TestGetInjectorImageAndNode(t *testing.T) {
...
+								SizeBytes: 500000,
+							},
+						},
+					},
+				},
+			},
+			expectedImage: "k8s.gcr.io/pause:3.7",
+		},
+		{
+			name: "falls back to smallest image when no pause images",
```
brandtkeller: I am tracing the logic - but am curious about the response of this test. 

The `expectedImage` is `alpine:latest` but that image wasn't available on `node2` - how do we expect the daemonset to be operational if it doesn't exist on all nodes?

brandtkeller (commented):
Review nearing completion - adding more thoughts and questions to help bridge a gap in my understanding.

src/pkg/cluster/injector_test.go
AustinAbro321: Generally it shouldn't ever get to this point, but the idea is that there should pretty much always be pause in the name of the pause image for the cluster. In the case that there is not we choose the smallest image in the cluster and expect that it is the pause image. Since it is a pre-requisite that all nodes must have access to the pause image then we expect them to still be able to pull the image even if it has not already been pulled on the node.

src/pkg/cluster/injector.go
AustinAbro321: Yeah it's a great question and definitely is something we should have an answer for. This logic should pretty much always choose the pause image. However, it could be the case that during an upgrade of Kubernetes the pause image goes from 3.10 to 3.11. If someone installed the new nodes with access to a registry that only has pause image 3.11 then those new nodes would not be able to spin up the injector since it would still try to use 3.10 as the base image. 

If all of the nodes start to use 3.11 as their pause image during a Kubernetes upgrade then all the nodes should now be using that image. Still, this logic might not choose it as currently it only chooses the smallest pause image.  One way we could probably mitigate that would be to choose the pause image that is both small (current pause image is 736kB) and the latest version available on a node. In that case users might still would have to re run `zarf init` though. The pause image is updated generally ~ once or twice a year. 

Also documented this option here in the ZEP https://github.com/zarf-dev/proposals/tree/main/0033-registry-proxy#stricter-process-for-determining-injector-image around allowing this to be overwritten by the user.

src/pkg/cluster/injector_test.go
brandtkeller: I think availability of the pause image being available is reasonable. I am less comfortable with the assumption of being able to pull it - that is a big assumption for use cases and is certainly not going to the case for airgapped clusters (k3s/rke2/k0s in particular - need to look at others). 

Not that this should be a blocker - but I do think we should pay particular attention to this constraint - as versions across node upgrades will most likely not match.

AustinAbro321: Possible I'm missing something, but there should always be access to the current pause image defined by the kubelet on each node even in the airgap. A node can't run pods unless it has access to the pause image. 

But yeah as you state, there is some uncertainty around how well this will work when upgrading a cluster. This conversation did make me realize that the same cluster could have nodes with different kubelet versions and thus different pause images. Theoretically, this shouldn't matter too much if the newest pause image is used, but if the version of the pause image we choose is phased out and the cluster no longer has access to it then things could break on upgrades / new nodes.

AustinAbro321: I updated the logic so that it chooses the latest pause image, however the fact that nodes might have different pause images presents a pretty big wrinkle. The way K3s generally loads in images for an airgap install is by loading in an OCI layout where containerd can pull from (https://docs.k3s.io/installation/airgap?airgap-install=High+Availability+Configuration). It looks like K0s does something similar (https://docs.k0sproject.io/v1.21.1+k0s.0/airgap-install/). Talos instructs users to push images to the OCI registry (https://www.talos.dev/v1.11/advanced/air-gapped/). 

I'm starting to re-consider if a controller to spin up an injector when the proxy can't pull an image might be the way to go. Especially since a daemonset might not work if nodes don't have the same images on them. It's an alternative I discussed here - https://github.com/zarf-dev/proposals/tree/main/0033-registry-proxy#dynamically-manage-injector-lifecycle-with-controller

packages/zarf-registry/chart/templates/proxy-daemonset.yaml
```diff
@@ -0,0 +1,64 @@
...
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+              - ALL
+            readOnlyRootFilesystem: true
+            runAsNonRoot: true
+    # IPV6 does not support rewriting packets to localhost so it needs to use hostNetwork over hostPort
+    {{- if eq (include "proxy.hostNetwork" .) "true" }}
+      dnsPolicy: ClusterFirstWithHostNet
+      hostNetwork: true
```
brandtkeller, line 53: for posterity (as discussed previously) - we'll want to flag any "risky" configurations in documentation such that they are discoverable. Any item that a policy enforcement engine might block.

packages/zarf-registry/chart/templates/injector-daemonset.yaml
```diff
@@ -0,0 +1,86 @@
+{{ if .Values.proxy.enabled }}
+apiVersion: apps/v1
+kind: DaemonSet
```
brandtkeller, line 3: Thinking about daemonset implications and resources. We may have to emphasize in the future documentation the constraints for proper runtime _on each node_ when the proxy is enabled.  

Previously we only needed a single node for zarf operations. I imagine init requires the daemonset is healthy before success - but also thinking through situations such as a node `cordon/drain/uncordon` or other cluster events that may present interesting behaviors.

src/pkg/cluster/injector_test.go
brandtkeller: I believe that is a reasonable point of investigation given what we now know.

More broadly - this test is meant to validate that we can determine the appropriate daemonset image - which means that it needs to be an image common across the cluster in order for the daemonset to operate on each node. 

If we keep the daemonset approach - I think we need to adjust the testing - otherwise a controller would likely make it OBE.

brandtkeller (commented):
pending comments for posterity

packages/zarf-registry/chart/templates/injector-daemonset.yaml
```diff
@@ -0,0 +1,93 @@
...
+  selector:
+    matchLabels:
+      app: zarf-daemonset-injector
+  template:
+    metadata:
+      labels:
+        app: zarf-daemonset-injector
+        # Don't mutate this pod
+        zarf.dev/agent: ignore
+    spec:
```
brandtkeller, line 16: Much like the agent and registry - we could add [Priority Class](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/) to this daemonset for preemption.

Mainly looking to ensure we've established these workloads as most critical to operations given that other workloads will be dependent.

```yaml
priorityClassName: system-node-critical
```

packages/zarf-registry/chart/templates/proxy-daemonset.yaml
```diff
@@ -0,0 +1,64 @@
...
+  selector:
+    matchLabels:
+      app: zarf-registry-proxy
+  template:
+    metadata:
+      labels:
+        app: zarf-registry-proxy
+        # Don't mutate this pod
+        zarf.dev/agent: ignore
+    spec:
```
brandtkeller, line 16: Same note as above for `priorityClassName`

packages/zarf-registry/chart/templates/injector-daemonset.yaml
AustinAbro321: Great callout!

src/pkg/cluster/injector_test.go
AustinAbro321: Updated the tests mirror the situations we might come across in a real cluster better, lmk if you're looking for any other testing.

packages/zarf-registry/chart/templates/injector-daemonset.yaml
AustinAbro321: Great callout, we'll definitely want docs pages over the implications of a Daemonset

packages/zarf-registry/chart/templates/proxy-daemonset.yaml
AustinAbro321: Yeah I will make a note for a future Issue / PR to document this feature more. Having a note for why this is necessary will be nice for someone looking to justify it from a policy enforcement engine

.github/workflows/test-proxy.yml
```diff
@@ -0,0 +1,102 @@
...
+
+      - name: Setup Kind
+        run: |
+          kind delete cluster && kind create cluster --config=kind-nftables-config.yml
+
+      - name: Make Zarf executable
+        run: |
+          chmod +x build/zarf
+
+      - name: Run tests
```
brandtkeller, line 96: `init` should not complete if it is not successful... So this feels like a valid test. I am curious if we potentially see any outliers where resources might not be entirely healthy?

src/internal/feature/feature.go
```diff
@@ -225,6 +226,13 @@ func init() {
 			Since:   "v0.60.0",
 			Stage:   Alpha,
 		},
+		{
+			Name:        RegistryProxy,
+			Description: "Enables setting Zarf init with the registry proxy flag",
+			Enabled:     false,
+			Since:       "v0.63.0",
```
brandtkeller: Once more bump on `since` before we ship.

src/pkg/cluster/injector.go
```diff
@@ -347,6 +352,127 @@ func (c *Cluster) getInjectorImageAndNode(ctx context.Context, resReq *v1ac.Reso
 	return "", "", fmt.Errorf("no suitable injector image or node exists")
 }
 
+// GetInjectorDaemonsetImage gets the image that is most likely to be accessible from all nodes
+// It first grabs the latest version pause image with semver 3 or 4, under 1MiB, and with pause in the name.
+// If there are no valid pause images then it grabs the smallest image.
+func (c *Cluster) GetInjectorDaemonsetImage(ctx context.Context) (string, error) {
```
brandtkeller, line 358: Stemming from the testing thread - I would like to discuss/document how we handle the daemonset constraints. 

A daemonset requires a single consistent image reference across all nodes. The tests indicate that this method will return  a string and no error in the event it determines an image - even if that is not a consistent image across all nodes. 

Do we want to accept this constraint - or rather should this return an error? I don't see how the daemonset could become operational if that is the case. My gut says we should stick to the most explicit behavior - so as to iron out inconsistencies with the core logic under best-case scenarios and then make the logic more flexible - but I am open to opinions.

AustinAbro321: In my opinion, we should accept this constraint given that the `node.Status.Images` does not reflect all images available to the node, but rather all images that have been pulled by the node. `node.Status.Images` is always empty if a node has not yet run any pods, for instance.

Often users will either have a registry that is available to their nodes, or will have manually added the images to the proper location in the node where their container runtime will look. In both of these cases, the image will not show up in `node.Status.Images`, but will be pullable. This logic assumes that some nodes may not have yet pulled an image, and so it intentionally targets the pause image since all nodes must have access to the pause image to run any pods. The only potential pitfall is if nodes are running different versions of Kubernetes then it's possible that the pause image selected might not be universally accessible across nodes.

.github/workflows/test-proxy.yml
AustinAbro321: `init` will not complete if resources are not ready, so I feel decent about it. Open to ideas as to what to add to this. As the feature grows, such as with mtls, and the ability to switch between registry modes, I could see us converting this into a more detailed go test.

src/pkg/cluster/injector.go
brandtkeller: That sounds reasonable to me - given that it is a best guess instead of a known quantity. Additionally we intend to expose the ability to set the image explicitly correct?

AustinAbro321: Yea, it's alternative in the proposal pending user feedback. https://github.com/zarf-dev/proposals/tree/main/0033-registry-proxy#stricter-process-for-determining-injector-image. I'd be open to amending it as a harder requirement in a follow up given that some additional constraints have been exposed.

PR #3961, opened 2025-07-02
feat: optimize package OCI pulls

AustinAbro321:
## Description

Currently we archive a package after it's pulled, only to unarchive it after. This changes the logic to only archive a package pull when necessary saving time. 

Testing on the uds-core package, it saves about 8 seconds on my machine https://github.com/defenseunicorns/uds-core/pkgs/container/packages%2Fuds%2Fcore

## Checklist before merging

### DIFF
--- a/src/pkg/packager/pull.go
+++ b/src/pkg/packager/pull.go
@@ -103,7 +103,7 @@ type pullOCIOptions struct {
 	RemoteOptions
 }
 
-func pullOCI(ctx context.Context, opts pullOCIOptions) (_ *layout.PackageLayout, _ error) {
+func pullOCI(ctx context.Context, opts pullOCIOptions) (*layout.PackageLayout, error) {
 	if opts.Shasum != "" {
 		opts.Source = fmt.Sprintf("%s@sha256:%s", opts.Source, opts.Shasum)
 	}


src/internal/packager/images/progress_test.go
```diff
@@ -35,7 +35,7 @@ func TestTracker_ReportingCycle(t *testing.T) {
 		lastTotalBytes.Store(totalBytes)
 	}
 
-	testInterval := 10 * time.Millisecond
+	testInterval := 20 * time.Millisecond
```
AustinAbro321, line 38: This flaked on windows. If it happens again we may just rework this test, but I figure I'll try this to start

src/internal/packager/images/progress_test.go
```diff
@@ -35,7 +35,7 @@ func TestTracker_ReportingCycle(t *testing.T) {
 		lastTotalBytes.Store(totalBytes)
 	}
 
-	testInterval := 10 * time.Millisecond
+	testInterval := 30 * time.Millisecond
```
AustinAbro321, line 38: This flaked on windows. If it happens again we may just rework this test, but I figure I'll try this to start

mkcp: Yeah that's weird, it's possible this is an unrelated timing issue that was paved over by the wait from compress and uncompress. Something to track but not a blocker imo

src/pkg/packager/pull.go
```diff
@@ -91,85 +91,80 @@ func Pull(ctx context.Context, source, destination string, opts PullOptions) (st
...
 
-func pullOCI(ctx context.Context, opts pullOCIOptions) (_ bool, _ string, err error) {
-	tmpDir, err := utils.MakeTempDir(config.CommonOptions.TempDirectory)
-	if err != nil {
-		return false, "", err
-	}
-	defer func() {
-		err = errors.Join(err, os.RemoveAll(tmpDir))
-	}()
+func pullOCI(ctx context.Context, opts pullOCIOptions) (_ *layout.PackageLayout, _ error) {
```
mkcp: nit: don't name vars if they're all dogsledded
```suggestion
func pullOCI(ctx context.Context, opts pullOCIOptions) (*layout.PackageLayout, error) {
```

src/pkg/packager/publish_test.go
```diff
@@ -37,25 +37,18 @@ func pullFromRemote(ctx context.Context, t *testing.T, packageRef string, archit
...
 		CachePath:     cachePath,
 	}
-	_, tarPath, err := pullOCI(context.Background(), pullOCIOpts)
-	require.NoError(t, err)
-
-	layoutActual, err := layout.LoadFromTar(ctx, tarPath, layout.PackageLayoutOptions{
-		Filter:        filters.Empty(),
-		PublicKeyPath: publicKeyPath,
-	})
+	pkgLayout, err := pullOCI(ctx, pullOCIOpts)
```
mkcp, line 48: This just makes sense - it's always nice when the test case/usage gets simpler alongside a speedup.

mkcp (approved):
LGTM

PR #3963, opened 2025-07-02
fix!: allow creation of differential package from OCI source

### LINKED ISSUE #3960: Creating differential packages referencing Zarf packages in an OCI registry fails
### Environment
Device and OS: Mac Studio M1 Ultra - MacOS Sequoia 15.5
App version: v0.55.6

### Steps to reproduce
I will preface this with the fact that I am not sure if the behaviour I am looking for is expected/ supported, and that differential packages are still a beta feature. But I am going off of the docs here [differential-packages](https://docs.zarf.dev/ref/packages/#differential-packages), specifically the section:

> You can use this flag to point to an already built package you have locally or to a **package that has been previously [published](https://docs.zarf.dev/tutorials/6-publish-and-deploy#publish-package) to a registry**.

1. Create Zarf package as normal and publish to OCI registry.
2. Make changes to zarf.yaml (such as image updates etc, and bump the version in the zarf.yaml).
3. Attempt to create Zarf differential package by running `zarf package create . --differential oci://registry_url/path/to/package:v1.2.3`

### Expected result
A differential Zarf package between the version referenced in the remote OCI registry and the changes that have been made locally is created.

### Actual Result
Running the command `zarf package create . --differential oci://registry_url/path/to/package:v1.2.3` produces the error message `ERR failed to create package: unable to perform decompression: unable to open archive "oci://registry_url/path/to/package:v1.2.3": open oci://registry_url/path/to/package:v1.2.3: no such file or directory`

N/A

### Severity/Priority
I am able to workaround this issue by pulling the remote image locally and creating a differential, but it would reduce complexity in workflow if I was able to do this directly from remote.

### Additional Context
I have verified that auth to the remote OCI registry works, and that the referenced package `oci://registry_url/path/to/package:v1.2.3` is valid. I have done this by using the command `zarf package inspect manifests oci://registry_url/path/to/package:v1.2.3` and getting a valid and successful output.

If I pull the Zarf package `oci://registry_url/path/to/package:v1.2.3` locally, and then run `zarf package create . --differential package:v1.2.3` the differential package is created successfully.

AustinAbro321:
## Description

There is a regression which blocks users from making a differential package from an OCI source. 

This does include a breaking change in the SDK for `layout.assemblePackage`. It's possible to keep this backwards compatible, but IMO it makes sense to change the field here given that:
- Callers are much more likely to call `packager2.create` which still accepts a source, than the more granular `layout.assemblePackage`
- It's simple to get the package definition before `layout.assemblePackage` with `packager2.LoadPackage`

## Related Issue

Fixes #3960

## Checklist before merging

### DIFF
--- a/src/pkg/packager/create.go
+++ b/src/pkg/packager/create.go
@@ -12,6 +12,7 @@ import (
 	"github.com/defenseunicorns/pkg/helpers/v2"
 	"github.com/defenseunicorns/pkg/oci"
 
+	"github.com/zarf-dev/zarf/src/api/v1alpha1"
 	"github.com/zarf-dev/zarf/src/pkg/logger"
 	"github.com/zarf-dev/zarf/src/pkg/packager/layout"
 	"github.com/zarf-dev/zarf/src/pkg/packager/load"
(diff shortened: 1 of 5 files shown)


mkcp (approved):
LGTM - just make sure to highlight the breaking change in our next patch notes

PR #3979, opened 2025-07-09
feat(deploy): actions setVariables to deployed component

### LINKED ISSUE #3975: VariableConfig access for deployed packages missing
### Description

In the previous `packager` implementation - the `Packager` client - used to perform Deploy - exposed a [GetVariableConfig()]() function that could retrieve the state of VariableConfig such that you could access any variables that were "set" during the deploy - afterwords.

This is not available currently and prevents library consumers from accessing the data about a given package that has been deployed. 

### Severity/Priority

High Priority

### Additional Context

`VariableConfig` handles the collection of `setVariables` used primarily in `actions` - it may be suitable that we include any `setVariables` into [DeployedComponents]() such that these can be accessed separately from other information that may not be needed. 

`packager.Deploy()` already returns a slice of DeployedComponents - we may be able to add information for consumers to access post-deployment.

brandtkeller:
## Description

Separate implementation of including variables set during component actions. 

`packager.Deploy()` returns a slice of `DeployedComponents`. If we were to add additional information to the deployment process for each component - then we could include that natively in the return. 

This modifies the signature of the `actions` package functions to support returning variables set such that they can be used on a component-by-component basis or aggregated for a whole package after deployment. 
## Related Issue

Fixes #3975 
## Checklist before merging

### DIFF
--- a/src/pkg/packager/actions/actions.go
+++ b/src/pkg/packager/actions/actions.go
@@ -7,6 +7,7 @@ package actions
 import (
 	"context"
 	"fmt"
+	"maps"
 	"path/filepath"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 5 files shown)


brandtkeller:
Closing as not selected for current path currently.

PR #3986, opened 2025-07-14
feat: allow the creation of skeleton flavored packages

### LINKED ISSUE #3985: Create `flavor`s of `skeleton` packages

It is not possible to create a skeleton project if all the components have `.only.flavor` set:

Example `zarf.yaml` from DU:
 
https://github.com/defenseunicorns/uds-package-renovate/blob/main/zarf.yaml

- **Given** add `--flavor blah`
- **Then** a skeleton package is created with the skeleton components

- This might be a dead issue per #3970

a1994sc:
## Description

Allows the creation of flavored skeleton packages.

## Related Issue

Fixes #3985

## Checklist before merging

### DIFF
--- a/src/config/lang/english.go
+++ b/src/config/lang/english.go
@@ -307,6 +307,7 @@ $ zarf package publish ./path/to/dir oci://my-registry.com/my-namespace
 	CmdPackagePublishFlagSigningKey         = "Private key for signing or re-signing packages with a new key. Accepts either a local file path or a Cosign-supported key provider"
 	CmdPackagePublishFlagSigningKeyPassword = "Password to the private key used for publishing packages"
 	CmdPackagePublishFlagConfirm            = "Confirms package publish without prompting. Skips prompt for the signing key password"
+	CmdPackagePublishFlagFlavor             = "The flavor of components to include in the resulting package. This will only effect skeleton packages uploaded, will not change non-skeleton packages published."
 
 	CmdPackagePullShort   = "Pulls a Zarf package from a remote registry and save to the local file system"
 	CmdPackagePullExample = `
--- a/src/pkg/packager/layout/assemble.go
+++ b/src/pkg/packager/layout/assemble.go
@@ -196,6 +196,7 @@ func AssemblePackage(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath 
 type AssembleSkeletonOptions struct {
 	SigningKeyPath     string
 	SigningKeyPassword string
+	Flavor             string
 }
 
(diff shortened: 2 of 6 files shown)


a1994sc:
Leaving in draft mode because idk if this would end up being dead code

src/pkg/packager/layout/assemble.go
```diff
@@ -207,8 +208,13 @@ func AssembleSkeleton(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath
 		return nil, err
 	}
 
-	for _, component := range pkg.Components {
-		err := assembleSkeletonComponent(ctx, component, packagePath, buildPath)
+	// To remove the flavor value, as the flavor is configured by the tag uploaded to the registry
```
AustinAbro321, line 211: Are you removing the flavor from the component so that it can be imported by a package that is not using the same flavor? That's probably the way to go just want to make sure

a1994sc: Yes sir, if I create a flavor of `chocolate` people could only create `chocolate` flavored children projects

src/config/lang/english.go
```diff
@@ -307,6 +307,7 @@ $ zarf package publish ./path/to/dir oci://my-registry.com/my-namespace
 	CmdPackagePublishFlagSigningKey         = "Private key for signing or re-signing packages with a new key. Accepts either a local file path or a Cosign-supported key provider"
 	CmdPackagePublishFlagSigningKeyPassword = "Password to the private key used for publishing packages"
 	CmdPackagePublishFlagConfirm            = "Confirms package publish without prompting. Skips prompt for the signing key password"
+	CmdPackagePublishFlagFlavor             = "The flavor of components to include in the resulting package. This will only effect skeleton packages uploaded, will not change non-skeleton packages published."
```
AustinAbro321: I think we could reach your intent, in a broader way.
```suggestion
	CmdPackagePublishFlagFlavor             = "The flavor of components to include in the resulting package. The flavor will be appended to the package tag"
```

a1994sc: Fair, "wording" is not my strong suite so I'm happy to update that

AustinAbro321:
Approved after make docs and schema is run

a1994sc:
oh, right because I did the github auto-magic commit it did not update the docs... I will get that up in the next little bit

AustinAbro321 (approved):
This is an intuitive extension of skeleton packages, nice work.

mkcp (approved):

PR #4001, opened 2025-07-17
test: multi arch eks

AustinAbro321:
## Description

This PR is not intended to be merged, but instead intended to merge with #3935. Testing out the eks.yaml for that PR here.

## Checklist before merging


src/pkg/cluster/injector.go
```diff
@@ -263,6 +263,9 @@ func (c *Cluster) getInjectorImageAndNode(ctx context.Context, resReq *v1ac.Reso
 		if err != nil {
 			return "", "", err
 		}
+		if nodeDetails.Status.NodeInfo.Architecture == "" || nodeDetails.Status.NodeInfo.Architecture != architecture {
+			continue
+		}
```
AustinAbro321:
```suggestion
		if nodeDetails.Status.NodeInfo.Architecture != "" && nodeDetails.Status.NodeInfo.Architecture != architecture {
			continue
		}
```
Suggestion to change the behavior to not continue if architecture is empty. That way on the off chance there's some distro or version that doesn't fill in this value, nothing breaks for existing users

AustinAbro321:
Closing this as the plan will likely be to add arch as a variable to the base zarf.yaml

PR #4002, opened 2025-07-17
feat: allow for configurable image push logic in the library

Racer159:
## Description

This is an [alpha] level suggestion to allow a little more direct control over image pushing in the Zarf deployment API.

This is useful if you know the registry you are pushing to already has the images you need and you want to slim down parts of the Zarf package deploy.

## Related Issue

Fixes #N/A

## Checklist before merging


src/pkg/packager/deploy.go
```diff
@@ -55,6 +55,8 @@ type DeployOptions struct {
 	NamespaceOverride string
 	// Remote Options for image pushes
 	RemoteOptions
+	// [alpha] Whether to skip image push
+	NoImagePush bool
```
mkcp, line 59: I know this corresponds directly to the param name in `deployComponent` but let's call the SDK-level configuration something active, like`DisableImagePush`. 

cc @zarf-dev/maintainers

AustinAbro321:
@Racer159 I know we discussed alternatives in a separate channel. Is there still more you're looking to solve through this PR ?

brandtkeller:
Closing after discussion with no intent to pursue further. 

For posterity - alternatives were evaluated to mutate the in-memory package to remove images as needed via SDK:
```golang
	for idx := range pkgLayout.Pkg.Components {
		pkgLayout.Pkg.Components[idx].Images = []string{}
	}
```

PR #4004, opened 2025-07-18
fix: avoid injector and registry randomly landing on same nodeport

### LINKED ISSUE #2357: CI flake: "provided port is already allocated" occassionally happens during e2e cross cluster tests
We have seen this flake a few times, it is unclear why it is happening. 
In all of the cases the error is `provided port is already allocated` 
In the following two cases the error is on the below line. 
- https://github.com/defenseunicorns/zarf/actions/runs/8100430794/job/22138566751
- https://github.com/defenseunicorns/zarf/actions/runs/8160962762/job/22308786570
```
_, initStdErr, err := e2e.Zarf("init", "--components="+initComponents, "--nodeport", "31337", "-l", "trace", "--confirm")
require.NoError(t, err)
```
We also see a similar error here on here on test 99: 
- https://github.com/defenseunicorns/zarf/actions/runs/8012041029/job/21886586280
```
stdOut, stdErr, err = e2e.Zarf("init", "--components=k3s", "--confirm")
```

AustinAbro321:
## Description

Fixes #2357 using option 2 in my most recent comment. 

I tested this with k3d by making only two ports available:
```bash
k3d cluster create --k3s-arg "--service-node-port-range=31998-31999@server:0" --k3s-arg="--disable=traefik@server:0"
```

After a try or two you'll something like
```bash
2025-07-18 10:58:59 INF adding archived binary configmaps of registry image to the cluster
2025-07-18 10:59:06 INF injector service NodePort conflicts with registry NodePort, recreating service conflictingPort=31999
2025-07-18 10:59:07 INF injector service NodePort conflicts with registry NodePort, recreating service conflictingPort=31999
2025-07-18 10:59:09 INF deploying component name=zarf-seed-registry
2025-07-18 10:59:09 INF processing Helm chart name=docker-registry version=1.0.0 source=Zarf-generated
```
and it continue successfully. 

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/injector.go
+++ b/src/pkg/cluster/injector.go
@@ -396,7 +396,6 @@ func buildInjectionPod(nodeName, image string, payloadCmNames []string, shasum s
 // createInjectorNodeportService creates the injector service on an available port different than the registryNodePort service
 func (c *Cluster) createInjectorNodeportService(ctx context.Context, registryNodePort int) (*corev1.Service, error) {
 	l := logger.From(ctx)
-
 	for {
 		svcAc := v1ac.Service("zarf-injector", state.ZarfNamespaceName).
 			WithSpec(v1ac.ServiceSpec().
@@ -415,17 +414,14 @@ func (c *Cluster) createInjectorNodeportService(ctx context.Context, registryNod
 		assignedNodePort := int(svc.Spec.Ports[0].NodePort)
 		if assignedNodePort == registryNodePort {
 			l.Info("injector service NodePort conflicts with registry NodePort, recreating service", "conflictingPort", assignedNodePort)
-
 			err = c.Clientset.CoreV1().Services(state.ZarfNamespaceName).Delete(ctx, "zarf-injector", metav1.DeleteOptions{})
 			if err != nil {
 				return nil, err
 			}
 
 			time.Sleep(500 * time.Millisecond)
-
 			continue
 		}
-
 		return svc, nil
 	}
 }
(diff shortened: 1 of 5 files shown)


src/pkg/cluster/injector.go
```diff
@@ -400,3 +392,36 @@ func buildInjectionPod(nodeName, image string, payloadCmNames []string, shasum s
 
 	return pod
 }
+
+// createInjectorNodeportService creates the injector service on an available port different than the registryNodePort service
+func (c *Cluster) createInjectorNodeportService(ctx context.Context, registryNodePort int) (*corev1.Service, error) {
+	l := logger.From(ctx)
+	for {
```
brandtkeller: unbounded retrying - while unlikely here - is still possible. Could we incorporate a retries limit or a timeout context?

AustinAbro321:
@brandtkeller Yeah the tradeoff with being static is that any number we choose could be an existing number that the user already has defined in their cluster, so we'd have to make the injector NodePort configurable. I'd rather not introduce another CLI option to minimize complexity for users

src/pkg/cluster/injector.go
AustinAbro321: Yeah great call, made a retry limit of 10

brandtkeller:
> @brandtkeller Yeah the tradeoff with being static is that any number we choose could be an existing number that the user already has defined in their cluster, so we'd have to make the injector NodePort configurable. I'd rather not introduce another CLI option to minimize complexity for users

%100 - and this is non-blocking for sure - but the conversation interesting. Say we were to build-in an expectation that the injector service is the same as the registry - 1 (then we decide if we want to continue decrementing on conflict). 

What does this alternative provide?
1. random is inherently difficult to debug - whereas a more static behavior can be observable. 
2. many secure environments require static lists of ports and protocols used - even if just temporarily - zarf might be violating the list or creating a scenario where the port wasn't pre-exposed on nodes/firewall ahead of time and prevents zarf from working accordingly. 

Food for thought?

AustinAbro321:
Ah yeah, I hadn't thought about secure systems where the port must be known in advance. If someone came to us with that request, I'd be inclined to add it, though in the absence of that I'd rather keep it how it is now.

brandtkeller (approved):
lgtm

PR #4011, opened 2025-07-21
feat: add logic to add image to possible list when auth is needed

### LINKED ISSUE #4007: dev find-images fails when not logged into registry that requires auth
### Environment
Device and OS: Fedora
App version: v0.58.0

#### Files

`zarf.yaml`

```yaml
kind: ZarfPackageConfig
metadata:
  name: fuzzy-registry1
  version: 1.0.0
components:
  - name: baseline
    required: true
    manifests:
      - name: fuzzy-registry1
        namespace: default
        files:
          - configmap.yaml
```

`configmap.yaml`

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: rook-ceph-operator-config
  namespace: rook-ceph
data:
  ROOK_CSI_ATTACHER_IMAGE: registry1.dso.mil/ironbank/opensource/kubernetes-sigs/sig-storage/csi-attacher:v4.8.1
```

#### Steps

1. `mv ~/.docker/config.json ~/.docker/config.json.backup`
2. `zarf dev find-images --skip-cosign .`

### Expected result

```yaml
components:
  - name: baseline
    images:
      # Possible images - baseline
      - registry1.dso.mil/ironbank/opensource/kubernetes-sigs/sig-storage/csi-attacher:v4.8.1
```

### Actual Result

```yaml
components:
```

### Effected Code

https://github.com/zarf-dev/zarf/blob/87e13f5d7492bad80acb76433d6d8c75ceffd810/src/pkg/packager/find_images.go#L236-L250

### Additional Context

At present there is no check on if an image is "valid" when it is a "Match" (e.i. if it is discovered in a pod spec `image` field.). So it is an odd delta between the possible and confirmed matches. I understand it is likely a result of false positives like with `example.com` and what not... but as package creator it is confusing when I change images from an upstream to, in this case, registry1 image and Zarf does not discover the change unless I have a docker config with the auth.

An MVP would be to change the `l.Debug` to `l.Warn` so it gives an output like this:

Image

Another option would be to include a flag that would ignore the crane check and just add it to the list.

a1994sc:
## Description

Updates logic that when trying to use `zarf dev find-images` a package, it will silently fail if the registry, in this case registry1.dso.mil, needs auth to check it.

## Related Issue

Fixes #4007

## Checklist before merging

### DIFF
--- a/src/cmd/dev.go
+++ b/src/cmd/dev.go
@@ -683,8 +683,10 @@ func (o *devFindImagesOptions) run(cmd *cobra.Command, args []string) error {
 
 	componentDefinition := "\ncomponents:\n"
 	for _, finding := range imagesScans {
-		if len(finding.Matches) > 0 {
+		if len(finding.Matches)+len(finding.PotentialMatches)+len(finding.CosignArtifacts) > 0 {
 			componentDefinition += fmt.Sprintf("  - name: %s\n    images:\n", finding.ComponentName)
+		}
+		if len(finding.Matches) > 0 {
 			for _, image := range finding.Matches {
 				componentDefinition += fmt.Sprintf("      - %s\n", image)
 			}
--- a/src/pkg/packager/find_images.go
+++ b/src/pkg/packager/find_images.go
@@ -39,6 +39,8 @@ import (
 var (
 	imageCheck      = regexp.MustCompile(`(?mi)"image":"((([a-z0-9._-]+)/)?([a-z0-9._-]+)(:([a-z0-9._-]+))?)"`)
 	imageFuzzyCheck = regexp.MustCompile(`(?mi)["|=]([a-z0-9\-.\/:]+:[\w.\-]*[a-z\.\-][\w.\-]*)"`)
+	shaCheck        = regexp.MustCompile(`(?mi)sha256:[a-fA-F0-9]{64}`)
+	statusCheck     = regexp.MustCompile(`(?mi)status code 40[0-9]`)
 )
 
 // FindImagesOptions declares the parameters to find images.
@@ -240,6 +242,12 @@ func FindImages(ctx context.Context, packagePath string, opts FindImagesOptions)
 				if descriptor, err := crane.Head(image, images.WithGlobalInsecureFlag()...); err != nil {
 					// Test if this is a real image, if not just quiet log to debug, this is normal
 					l.Debug("suspected image does not appear to be valid", "error", err)
+					// statusCheck is find if the error has an 40x error code
+					// shaCheck is remove false positives of sha256:aaaaa....
+					if statusCheck.FindString(err.Error()) != "" && shaCheck.FindString(image) == "" {
+						l.Debug("adding image even though registry check failed")
+						validMaybeImages = append(validMaybeImages, image)
+					}
 				} else {
 					// Otherwise, add to the list of images
 					l.Debug("imaged digest found", "digest", descriptor.Digest)
--- a/src/pkg/packager/find_images.go
+++ b/src/pkg/packager/find_images.go
@@ -37,10 +37,11 @@ import (
 )
 
 var (
-	imageCheck      = regexp.MustCompile(`(?mi)"image":"((([a-z0-9._-]+)/)?([a-z0-9._-]+)(:([a-z0-9._-]+))?)"`)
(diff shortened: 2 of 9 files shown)


src/cmd/dev.go
```diff
@@ -683,8 +683,10 @@ func (o *devFindImagesOptions) run(cmd *cobra.Command, args []string) error {
 
 	componentDefinition := "\ncomponents:\n"
 	for _, finding := range imagesScans {
-		if len(finding.Matches) > 0 {
+		if len(finding.Matches)+len(finding.PotentialMatches)+len(finding.CosignArtifacts) > 0 {
```
a1994sc, line 686: This was a slight rework because at present will render like the following:

image

src/pkg/packager/find_images.go
```diff
@@ -240,6 +242,12 @@ func FindImages(ctx context.Context, packagePath string, opts FindImagesOptions)
 				if descriptor, err := crane.Head(image, images.WithGlobalInsecureFlag()...); err != nil {
 					// Test if this is a real image, if not just quiet log to debug, this is normal
 					l.Debug("suspected image does not appear to be valid", "error", err)
+					// statusCheck is find if the error has an 40x error code
+					// shaCheck is remove false positives of sha256:aaaaa....
+					if statusCheck.FindString(err.Error()) != "" && shaCheck.FindString(image) == "" {
```
brandtkeller: This looks good - one additional edge case is that a `dial tcp connect: connection refused` error will still be silent. 

For example:
```
./build/zarf dev find-images src/pkg/packager/testdata/find-images/fuzzy-registry-auth/ --skip-cosign -l debug
```

Will not return the same entries as the test case - as the logs indicate

```
2025-07-22 23:00:09 DBG found possible fuzzy match kind=ConfigMap value=localhost:65000/sig-storage/csi-node-driver-registrar:v2.13.0
2025-07-22 23:00:09 DBG found possible fuzzy match kind=ConfigMap value=localhost:65000/sig-storage/csi-resizer:v1.13.2
2025-07-22 23:00:09 DBG found possible fuzzy match kind=ConfigMap value=localhost:65000/sig-storage/csi-snapshotter:v8.2.1
2025-07-22 23:00:09 DBG suspected image does not appear to be valid error=Get "https://localhost:65000/v2/": dial tcp 127.0.0.1:65000: connect: connection refused; Get "http://localhost:65000/v2/": dial tcp 127.0.0.1:65000: connect: connection refused
2025-07-22 23:00:09 DBG suspected image does not appear to be valid error=Get "https://localhost:65000/v2/": dial tcp 127.0.0.1:65000: connect: connection refused; Get "http://localhost:65000/v2/": dial tcp 127.0.0.1:65000: connect: connection refused
2025-07-22 23:00:09 DBG suspected image does not appear to be valid error=Get "https://localhost:65000/v2/": dial tcp 127.0.0.1:65000: connect: connection refused; Get "http://localhost:65000/v2/": dial tcp 127.0.0.1:65000: connect: connection refused
```

Should this be captured as well?

src/cmd/dev.go
brandtkeller: Good catch - this does fix formatting

src/pkg/packager/find_images_test.go
```diff
@@ -155,6 +162,49 @@ func TestFindImages(t *testing.T) {
...
+						"registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.13.0",
+						"registry.k8s.io/sig-storage/csi-provisioner:v5.2.0",
+						"registry.k8s.io/sig-storage/csi-resizer:v1.13.2",
+						"registry.k8s.io/sig-storage/csi-snapshotter:v8.2.1",
+					},
+				},
+			},
+		},
+		{
+			name:        "fuzzy-registry-auth",
```
brandtkeller, line 188: ++ appreciative of this test case

brandtkeller (changes requested):
Functions as intended with local testing - left one potential edge case for consideration otherwise looks like a great improvement.

src/pkg/packager/find_images.go
a1994sc: Personally I think that makes sense to me, especially because Zarf detects this as match without checking:

```yaml
apiVersion: apps/v1
kind: Pod
metadata:
  labels:
    app: agent
spec:
  containers:
    - name: agent
      image: localhost:65000/sig-storage/csi-snapshotter:v8.2.0
```

brandtkeller: Agreed!

mkcp (approved):
LGTM

PR #4023, opened 2025-07-25
chore: update case for find images

### LINKED ISSUE #4022: zarf dev find-images command returning ClusterBinding roleRef.name value as image.
### Environment
Device and OS: Darwin
App version: v0.59.0
Kubernetes distro being used: Docker for Desktop k8s
Other:

### Steps to reproduce

Create a zarf package that pulls down the keda helm chart in a component:

```
  - name: keda
    required: true
    charts:
      - name: keda
        version: "2.17.0"
        namespace: keda
        url: https://kedacore.github.io/charts
        repoName: keda
    images:
      - ghcr.io/kedacore/keda-admission-webhooks:2.17.0
      - ghcr.io/kedacore/keda-metrics-apiserver:2.17.0
      - ghcr.io/kedacore/keda:2.17.0
      # Cosign artifacts for images - keda
      - ghcr.io/kedacore/keda-admission-webhooks:sha256-a87c42275757c70d21921e0928164802fde6dbc167cc0f8f5a34880e20b74937.sig
      - ghcr.io/kedacore/keda-metrics-apiserver:sha256-167fd532bd432334ad27f082689a4d7d2edca7eae707e895e767d303cb8fc9a5.sig
      - ghcr.io/kedacore/keda:sha256-112fc427d933f8dd11f51e081bb5b96303951419de9573ea952b1648fc342bd3.sig
    actions:
      onDeploy:
        after:
          - wait:
              cluster:
                kind: deployment
                name: keda-operator
                namespace: keda
                condition: available
          - wait:
              cluster:
                kind: deployment
                name: keda-operator-metrics-apiserver
                namespace: keda
                condition: available
          - wait:
              cluster:
                kind: deployment
                name: keda-admission-webhooks
                namespace: keda
                condition: available
```
Then run the `zarf dev find-images` command on it and you will see that some strange `Possible Images` are included in the image manifest, in this case it gave me:
```
# Possible images - keda
      - system:auth-delegator
```
When I tried to determine why:
```
± zarf dev find-images --why system:auth-delegator --deploy-set IP_STACK_MODE=SingleStackIPv4
2025-07-25 11:14:54 INF using config file location=/REDACTED/zarf-config.yaml
....
repo=https://kedacore.github.io/charts
2025-07-25 11:15:01 INF looking for images in component name=keda resourcesCount=29
2025-07-25 11:15:01 INF looking up cosign artifacts for discovered images count=4
2025-07-25 11:15:07 INF processing Helm chart name=rabbitmq-cluster-operator version=4.3.21
2025-07-25 11:15:17 INF looking for images in component name=rabbitmq-operator resourcesCount=37
2025-07-25 11:15:18 INF looking up cosign artifacts for discovered images count=4
2025-07-25 11:15:25 INF processing Helm chart name=cloudnative-pg version=0.23.2 repo=https://cloudnative-pg.github.io/charts
....
component: keda
chart: keda
resource:

apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  labels:
    app.kubernetes.io/component: operator
    app.kubernetes.io/instance: keda
    app.kubernetes.io/name: keda-operator-system-auth-delegator
    app.kubernetes.io/part-of: keda-operator
    app.kubernetes.io/version: 2.17.0
    helm.sh/chart: keda-2.17.0
  name: keda-operator-system-auth-delegator
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: system:auth-delegator
subjects:
- kind: ServiceAccount
  name: keda-metrics-server
  namespace: collibra-control
```

### Expected result

I expect only images to be populated in the zarf configs image manifest after running the tool.

### Actual Result

A roleRef name is suggested as an image.

### Severity/Priority

Critical, we run the tool as part of our pipeline to ensure that images are up to date and the manifest includes all images require to deploy the chart in an air gapped environment. 

### Additional Context

- This issue was introduced with `v0.59.0`, `v0.58.0` does not suggest the same images.

a1994sc:
## Description

...

## Related Issue

Fixes #4022

## Checklist before merging


brandtkeller:
@zarf-dev/maintainers any thoughts for what the behavior here might look like?

AustinAbro321:
Thinking this over, I'd prefer reverting the functionality in #4011. Something I hadn't thought about is that even when an image doesn't exist in the registry it'll give a 401 error. This makes the registry call as is not useful, as we'll pretty much always get 401 errors for non-images. 

It'll be hard to maintain logic otherwise for fuzzy images, and we want the list here to be very accurate, even for maybe images. I've seen users with flows where they auto update their package using `zarf dev find-images`. Additionally, users will need to login to create a package with the image anyway. I think the most we'd want to do would be to warn the user that we failed to connect to auth to the registry.

brandtkeller:
In this case - should we opt to improve documentation to support:
1. `info` logs for inability to authenticate during 401's (already an error/debug log)
2. Update command docs and documentation to clarify we will only provide output if images for which the registry call returns information/successful.

There is certainly a trade-off here. I don't hold strong opinions but am curious if there is a subset that is still valuable here? IE put the more loose filtering behind a flag perhaps?

AustinAbro321:
Yeah I think both of those steps would be good. The issue I have with a flag, is that the user should be logged into their registry anyway, assuming that they want to create the package soon after finding images. We shouldn't ask them for another action on top of that, when they already very likely need to login

a1994sc:
Valid, I do think that my own issues came from some level of annoyance when using a new devbox. I do think that there is a weird delta from how zarf assumes that `.spec.containers[].image` is always valid.... however given the how many different ways images are formatted and what not, it is not an easy problem to solve for

a1994sc:
So I reverted the logic changes from pr 4011, left in the logic that fixes the improper formatting of the `zarf dev find-images`:

https://github.com/zarf-dev/zarf/pull/4011/files#r2220321491

As well as the `SetupInMemoryRegistryWithAuth` Function

src/pkg/packager/find_images.go
```diff
@@ -380,6 +371,9 @@ func processUnstructuredImages(ctx context.Context, resource *unstructured.Unstr
 		}
 		matchedImages = appendToImageMapOCIRepo(ctx, matchedImages, ociRepo)
 
+	case "ClusterRoleBinding", "RoleBinding", "ClusterRole", "Role":
+		l.Debug("found a known false positive type", "kind", resource.GetKind())
+
```
AustinAbro321: This should be deleted, the presence of one of these resources doesn't indicate that find-images will find a false positive among them.

src/pkg/packager/find_images.go
```diff
@@ -392,8 +386,13 @@ func processUnstructuredImages(ctx context.Context, resource *unstructured.Unstr
 	// Capture "maybe images" for all kinds
 	matches := imageFuzzyCheck.FindAllStringSubmatch(string(b), -1)
 	for _, group := range matches {
-		l.Debug("found possible fuzzy match", "kind", resource.GetKind(), "value", group[1])
-		maybeImages[group[1]] = true
+		switch resource.GetKind() {
+		case "ClusterRoleBinding", "RoleBinding", "ClusterRole", "Role":
+			l.Debug("found a known false positive type", "kind", resource.GetKind())
```
AustinAbro321: Not sure how I feel about us maintaining a list of resources that can't have images in them, but I'm leaning towards it being good. This could help avoid a bunch of extraneous logs if we decide to make the crane return message an info level message. I think we can say with certainty that these types won't have images in them.

src/pkg/packager/find_images_test.go
```diff
@@ -185,22 +178,19 @@ func TestFindImages(t *testing.T) {
 			},
 		},
 		{
-			name:        "fuzzy-registry-auth",
-			packagePath: "./testdata/find-images/fuzzy-registry-auth",
+			name:        "roles-bindings",
```
AustinAbro321: I don't think we should include this test because the behavior is not actually changing since we still call out to Crane. We could get rid of the changes you made and the test would still pass

src/pkg/packager/find_images.go
a1994sc: Ah..... yea forgot that I had started off with creating that list... but yea, I can pull that code.

a1994sc:
> Appreciate you making this! I do think we should get rid of the `SetupInMemoryRegistryWithAuth` function since we're not using it. It's also in git history if we want to add it back.

Removed that function, let me know if there are any other changes

src/pkg/packager/testdata/find-images/roles-bindings/zarf.yaml
AustinAbro321, line 1: Need to delete this file then lgtm

mkcp (approved):
LGTM

Issue #4031, opened 2025-07-30
Unable to upgrade zarf-agent with custom-named init packages


brandtkeller:
### Environment
Device and OS: Linux / Ubuntu
App version: v0.59.0
Kubernetes distro being used: Kind
Other:

### Steps to reproduce
1. Create a custom-named init package with name other than `init` at version v0.58.0
2. Deploy the custom init package
3. Create a custom-named init package with name other than `init` at version v0.59.0
4. Deploy the upgraded custom init package

### Expected result

custom named init package upgrades as expected

### Actual Result

upgrade fails on the `zarf-agent` 

```bash
ERR failed to deploy package: unable to deploy component "zarf-agent": release: not found
unable to install chart after 3 attempts: All attempts fail:
#1: Unable to continue with install: ServiceAccount "zarf" in namespace "zarf" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "zarf-d2db14ef40305397791454e883b26fc94ad9615d": current value is "zarf-7628fdd392e8726264ce1f894b809ebf386b6acc"
#2: Unable to continue with install: ServiceAccount "zarf" in namespace "zarf" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "zarf-d2db14ef40305397791454e883b26fc94ad9615d": current value is "zarf-7628fdd392e8726264ce1f894b809ebf386b6acc"
#3: Unable to continue with install: ServiceAccount "zarf" in namespace "zarf" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "zarf-d2db14ef40305397791454e883b26fc94ad9615d": current value is "zarf-7628fdd392e8726264ce1f894b809ebf386b6acc": if you need to remove the failed chart, use `zarf package remove`
```

### Severity/Priority

Medium

### Additional Context
This is due to the recent changes to migrating the zarf-agent from manifests to a helm chart.

brandtkeller:
Still investigating broader implications. Due to the [Manifest -> Chart translation](https://github.com/zarf-dev/zarf/blob/main/src/internal/packager/helm/common.go#L32) that Zarf performs using a unique name from the package/component/manifest name - this alters the name for the zarf-agent helm release for custom-named init packages. 

This was handled for everyone who uses the standard "init" name for their init package. 

[...]

PR #4034, opened 2025-07-31
4031 init package naming

brandtkeller:
## Description

Discovered that `zarf tools update-creds agent` does not work for any init packages where `metadata.name` does not equal `init` (IE all custom-named init packages). 

This was inadvertently fixed in #3678 for all releases >=v0.59.0. This change merely allows the ability to run this logic on clusters that are already initialized with a custom-named init package but may not have upgraded. 

Given the historical data on clusters that run into certificate update issues - this may be something we reference for some time and can point users to the next release as a version that resolves the ability to update certificates on all <= v0.58.0 zarf agents.  

## Related Issue

Relates to #4031

## Checklist before merging

### DIFF
--- a/src/internal/packager/helm/zarf.go
+++ b/src/internal/packager/helm/zarf.go
@@ -108,8 +108,9 @@ func UpdateZarfAgentValues(ctx context.Context, opts InstallUpgradeOptions) erro
 	found := false
 	for _, release := range releases {
 		// Update the Zarf Agent release with the new values
-		// Maintaining the "raw-init" release name for backwards compatibility
-		if release.Chart.Name() == "raw-init-zarf-agent-zarf-agent" {
+		// note: this should never work for custom named init packages - this should be made more generic
+		// Assumes there will only be one release in the zarf namespace with "zarf-agent" in the name
+		if strings.Contains(release.Chart.Name(), "zarf-agent") {
 			found = true
 			chart := v1alpha1.ZarfChart{
 				Namespace:   "zarf",
--- a/src/internal/packager/helm/zarf.go
+++ b/src/internal/packager/helm/zarf.go
@@ -108,7 +108,6 @@ func UpdateZarfAgentValues(ctx context.Context, opts InstallUpgradeOptions) erro
 	found := false
 	for _, release := range releases {
 		// Update the Zarf Agent release with the new values
-		// note: this should never work for custom named init packages - this should be made more generic
 		// Assumes there will only be one release in the zarf namespace with "zarf-agent" in the name
 		if strings.Contains(release.Chart.Name(), "zarf-agent") {
 			found = true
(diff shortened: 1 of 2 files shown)


src/internal/packager/helm/zarf.go
```diff
@@ -108,8 +108,8 @@ func UpdateZarfAgentValues(ctx context.Context, opts InstallUpgradeOptions) erro
 	found := false
 	for _, release := range releases {
 		// Update the Zarf Agent release with the new values
-		// Maintaining the "raw-init" release name for backwards compatibility
-		if release.Chart.Name() == "raw-init-zarf-agent-zarf-agent" {
+		// Assumes there will only be one release in the zarf namespace with "zarf-agent" in the name
```
AustinAbro321, line 111: Unlikely scenario, but want to call out that if someone updates to a newer init package then `zarf tools update-creds agent` won't work on older version of Zarf

brandtkeller: That is a fair point. Given that there is no real pressure to change the chart name - we could retain `raw-init-zarf-agent-zarf-agent` and allow older zarf binaries to still run that command on a newer init package.

brandtkeller: reverted the chart name change - theoretically this maintains the name and older binaries can run `update-creds agent` on clusters initialized with init packages named "init".

AustinAbro321:
I'm thinking about this, and I'm not sure it's an addition we should add. It's pretty unlikely another user would have `zarf-agent` another chart with zarf-agent in the name, but given that this is not something we've seen a request for it, it feels like we're introducing a risk without enough upside. I think I'd prefer to tell users to update their cluster with `zarf init` at the latest version. They will have to handle https://github.com/zarf-dev/zarf/issues/4031, but at some point they will have to do that either way.

brandtkeller:
@AustinAbro321 I think that is fair - I do wonder if there is a case where the corner is hard to back-out-of without some careful `kubectl` crafting. 

IE today we don't update agent tls certs on init package upgrades. 

So If an environment were to have an old version of a custom-named init package and the certificates expired. Could they run Zarf init with new artifacts and be able to update the certs? likely not if they had anything deploy _after_ the agent as something like the git server would fail to deploy (certs are still expired). 

It is quite niche - so I am not against closing this out.

AustinAbro321:
@brandtkeller I had not considered that case. It's an interesting situation. If I was advising a user to get around it, I would ask them to not deploy gitea if optional, or re-create their init package without it, so it'd be possible to get around but certainty tough. 

My gut feeling is still to close this out, as I do think it will be quite niche and I wouldn't 100% rule out someone having a custom component like `zarf-agent-crossplane`. Hard to know what's the right move here though.

brandtkeller:
It is a recoverable state - albeit will require a number of manual steps - but should not be too fraught with any potential to remove any persistent state. I would advocate for building broad paths to solving known issues in previous releases - but this is optimally very niche and resolved going forward. Closing until such a time as this needs to be revisited.

PR #4039, opened 2025-08-01
feat(feature): Implement Feature API

mkcp:
## Description
This PR implements the `internal/feature` API as described in ZEP-0035 Feature Flags. It creates a new package `feature` which allows maintainers to set default `Feature`s, and provides facilities for users to enable and disable features for use in the SDK. Before release we should ship at least one default feature flag, even a test/documentation noop feature struct, to enforce that setDefault can only be called once.

In a follow-up PR we'll implement:
- These same user facilities (`feature.SetUser`) will be applied in the cmd layer to allow users to configure their own features via CLI flags, Env vars, and on-disk config.

## Related Issue
Implements part of [ZEP-0035](https://github.com/zarf-dev/proposals/tree/main/0035-feature-flags)

## Checklist before merging


src/internal/feature/feature.go
```diff
@@ -0,0 +1,224 @@
...
+
+// All returns all flags from both Default and User.
+func All() map[Mode]map[Name]Feature {
+	m := make(map[Mode]map[Name]Feature)
+	m[Default] = AllDefault()
+	m[User] = AllUser()
+	return m
+}
+
+// AllDefault returns all features with from the Default set for this version of Zarf.
```
AustinAbro321:
```suggestion
// AllDefault returns all features from the Default set for this version of Zarf.
```

src/internal/feature/feature.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		// We don't actually need to check the error here because an empty f will have the same truthiness
+		return false
+	}
+	return bool(f.Enabled)
+}
+
+// Set takes a slice of one or many flags, inserting the features onto user-configured features. If a feature name is
+// provided that is already a part of the set, then Set will return an error.
+// TODO: Should we allow users to call this multiple times even if we don't allow them to overwrite features?
```
AustinAbro321: I think dis-allowing is fine. If there's a world where we need to allow multiple sets in the future then we could change it, but this is probably the most intuitive option since there's always an error on the second attempt

src/internal/feature/feature_test.go
```diff
@@ -0,0 +1,752 @@
...
+		{
+			name: "user enabled, default empty",
+			user: []Feature{
+				{Name: "foo", Enabled: true},
+			},
+			fName:  "foo",
+			expect: true,
+		},
+		{
+			name: " enabled, user empty",
```
AustinAbro321: nit: A few test names are missing the word default

AustinAbro321 (commented):
A few nits, otherwise looks like a solid strategy

src/internal/feature/feature_test.go
mkcp: Ah dang i thought i fixed all of these 😆

src/internal/feature/feature.go
mkcp: Yeah i'm happy with the current behavior as well.

The main edge case I could see this creating is when a user is calling the SDK in one "context" with one set of feature flags, and then needs a different set of flags for a different context later but within the same running process. The workaround is just to enable everything up front, but issues could still arise. Simple enough to ease the restriction later though.

src/internal/feature/feature.go
```diff
@@ -0,0 +1,213 @@
...
+
+// All returns all flags from both Default and User.
+func All() map[Mode]map[Name]Feature {
+	m := make(map[Mode]map[Name]Feature)
+	m[Default] = AllDefault()
+	m[User] = AllUser()
+	return m
+}
+
+// AllDefault returns all features from the Default set for this version of Zarf.
```
brandtkeller, line 167: Should this intentionally return the same map instance stored in `defaultFeatures` or should it be making/return a copy such that someone doesn't unintentionally (or otherwise) break the WORM guarantee and mutate a flag?

src/internal/feature/feature.go
```diff
@@ -0,0 +1,213 @@
...
+// Package feature provides feature flags.
+package feature
+
+import (
+	"fmt"
+	"sync/atomic"
+)
+
+// Atoms wrapping the default and user-set collections of Features. These do not require mutexes because they will each
+// be modified very few times, with the public API for each ensuring there only an empty set can be written to.
```
brandtkeller: nit: potentially some grammar fixes (Or I learn how to read). 

> ensuring there only an empty set can be written to

src/internal/feature/feature.go
```diff
@@ -0,0 +1,213 @@
...
+
+	// Feature not found in either set
+	return Feature{}, fmt.Errorf("feature not found: %s", name)
+}
+
+// GetDefault takes a flag Name and returns the Feature struct from the default set.
+func GetDefault(name Name) (Feature, error) {
+	f, ok := AllDefault()[name]
+	if !ok {
+		return f, fmt.Errorf("feature not found: %s", name)
```
brandtkeller: Both of these Get functions return the same error - is it worth delineating which set was searched in the error to make it easier to debug?

brandtkeller (commented):
Minor items for context or my education. Appreciative of the tests - they were easy to step through an observe intent.

src/internal/feature/feature.go
mkcp: Nope - this is an isolation bug and we should be returning a copy of the map! thank you for catching it!

src/internal/feature/feature.go
mkcp: Fixed! There -> that

src/internal/feature/feature.go
mkcp: Makes sense, updated

PR #4051, opened 2025-08-06
feat(feature): add cmd integration and "axolotl mode"

mkcp:
## Description
This PR adds `features` to the cmd layer of Zarf. This includes additional viper config, a setup step and user debug notification in `rootCmd.preRun()`, as well as some additional parsing and printing functions where needed.

It also adds the "feature" `"axolotl-mode"` to help test the new `feature` implementation. Enabling the easter egg `"axolotl-mode=true"` is a throwback to before the structured logging overhaul, and allows users to opt-in to the original logo print before CLI commands, printing out the same output as `zarf say`.

Of note is that the string-input for each of (`CLI:--features="...", ENV:ZARF_FEATURES="...", file:features = '...'`) deviates slightly from the current ZEP-0035 revision and the proposal will need to be updated after this is merged. I opted for one flag/key-name, `features` a list of KV pairs, with comma separated pairs and `=` separated: `feature_name=bool`.

I believe this matches expected cli conventions of golang projects better, and using this input string in all places keeps our parsing code much simpler. I also believe one could make an argument in favor of the overall UX for this approach, even though it doesn't use nested keys and maps like a user might expect for files. With one format and one key name, features are named and handled the same in each config source, and users can simply copy-paste their settings from one to another.

Examples below:

### Config file
Screenshot 2025-08-06 at 13 09 42
Screenshot 2025-08-06 at 13 24 25

### ENV var
Screenshot 2025-08-06 at 13 25 02

### CLI flag
Screenshot 2025-08-06 at 13 25 40

### Axolotl Mode enabled
Screenshot 2025-08-06 at 13 26 02

## Open questions
- Should we be warning users if they provide a user flag that isn't in the defaults? It's hard to imagine a feature that we don't have a described default state for. This didn't come up in the UX section of the proposal but it might be a good follow-up for Beta readiness. Adding this will require an extra read during cmd setup, but shouldn't be prohibitively difficult.
- Should we use `featureName={enabled|disabled}` rather than `featureName={true|false}`? I don't think the expected behavior is even slightly ambiguous, but the former matches our internal model better. It also requires more typing so, 🤷 i could go either way.
- Just a note, but worth mentioning that Features is still in Alpha and the API as well as cmd layers are all subject to change based on feedback!

EDIT:

The current state of the PR is a bit different than the examples. Now we use Viper's `StringToString` map type, which has different parsing rules depending on whether it's a flag, env var, or config file. This behavior is the same as other places we use StringToString, so there's no special UX here - just the same Zarf users are already accustomed to.

## Related Issue
Relates to https://github.com/zarf-dev/proposals/tree/main/0035-feature-flags

## Checklist before merging

### DIFF
--- a/src/cmd/root.go
+++ b/src/cmd/root.go
@@ -10,9 +10,8 @@ import (
 	"log/slog"
 	"os"
 	"slices"
+	"strconv"
 	"strings"
-
-	"github.com/zarf-dev/zarf/src/pkg/logger"
 
 	"github.com/pterm/pterm"
 	"github.com/spf13/cobra"
@@ -20,12 +19,17 @@ import (
 
 	"github.com/zarf-dev/zarf/src/config"
 	"github.com/zarf-dev/zarf/src/config/lang"
+	"github.com/zarf-dev/zarf/src/internal/feature"
+	"github.com/zarf-dev/zarf/src/pkg/logger"
 	"github.com/zarf-dev/zarf/src/types"
 )
 
 var (
 	// Default global config for the packager
 	pkgConfig = types.PackagerConfig{}
+	// Features is a string map of feature names to enabled state.
+	// Example: "foo=true,bar=false,baz=true"
+	features string
 	// LogLevelCLI holds the log level as input from a command
 	LogLevelCLI string
 	// LogFormat holds the log format as input from a command
@@ -68,6 +72,18 @@ func (o *outputFormat) Type() string {
 var rootCmd = NewZarfCommand()
 
 func preRun(cmd *cobra.Command, _ []string) error {
+	// Configure user defined Features
+	err := setupFeatures(features)
+	if err != nil {
+		return err
+	}
+	// Implement "axolotl-mode"
+	if feature.IsEnabled("axolotl-mode") {
+		if _, err = fmt.Fprintln(os.Stderr, logo()); err != nil {
+			return err
+		}
+	}
+
 	// If --insecure was provided, set --insecure-skip-tls-verify and --plain-http to match
 	if config.CommonOptions.Insecure {
 		config.CommonOptions.InsecureSkipTLSVerify = true
@@ -87,6 +103,9 @@ func preRun(cmd *cobra.Command, _ []string) error {
 	ctx := logger.WithContext(cmd.Context(), l)
 	cmd.SetContext(ctx)
 
+	// Print enabled features once we have a logger available
+	l.Debug("User-configured features:", "features", flattenUserFeatures())
+
 	// if --no-color is set, disable PTerm color in message prints
 	if IsColorDisabled {
 		pterm.DisableColor()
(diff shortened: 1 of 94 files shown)


src/cmd/root.go
```diff
@@ -174,6 +258,9 @@ func init() {
 
 	vpr := getViper()
 
+	// Features
+	rootCmd.PersistentFlags().StringVar(&features, "features", vpr.GetString(VFeatures), "[ALPHA] Provide a comma-separated list of feature names to bools to enable or disable. Ex. --features \"foo=true,bar=false,baz=true\"")
```
AustinAbro321: nit: Should we use StringToStringVar to make parsing slightly simpler

mkcp: Yes! I looked but wasn't able to find this api, thanks for the shout

mkcp: While experimenting with this, it looks like the key=value syntax doesn't work for env vars. That said, there may still be a way to set them from env vars with the StringToStringVar datatype and parser. It also changes the config file syntax to actually support maps in a way that I think i better UX, but will need to be documented. 

For the purposes of this PR and the alpha version of features let's just ship the same string input for each, and revisit StringToString for beta.

src/cmd/root.go
```diff
@@ -68,6 +72,18 @@ func (o *outputFormat) Type() string {
 var rootCmd = NewZarfCommand()
 
 func preRun(cmd *cobra.Command, _ []string) error {
+	// Configure user defined Features
+	err := setupFeatures(features)
```
brandtkeller, line 76: I am looking back to the proposal to understand how this is intended to be used throughout the CLI. 

Should this operation be a `PersistentPreRun` instead - allowing for `setupFeatures` to be defined in a single location `root.go`?

brandtkeller (commented):
review in progress - dropping notes iteratively.

src/cmd/root.go
brandtkeller: To elaborate - Allow `setupFeatures` to be invoked in a single location versus every subcommand `PreRun()`.

src/cmd/root.go
AustinAbro321: Given that we have other stringToString variables that have seen significant use such as `[package.create.set]`, I'd like to see us stick with that same format for the viper config files. That said, I agree that the default way that viper decodes these environment variables (a json string) is a very poor UX. I think a custom parser could work. Though I also believe that the CLI flag and config file are sufficient for now, and if we get user feedback that they'd like to use environment variables we can add that later.

mkcp: Yeah that's fair - given that users will already have to use JSON for env vars in this situations then it's at least a consistent UX. Updated in latest.

src/cmd/root.go
mkcp: This is a good callout - in our case the function named "preRun" is being executed as a PersistentPreRun. The naming is a bit off but the functionality should be there - otherwise other setup like logger wouldn't be applying on subcommands.

brandtkeller: ah that is deceiving but I see it now! looks good.

PR #4062, opened 2025-08-08
feat: add shorthands for all --confirm flags

### LINKED ISSUE #4061: Support `-c` alias for `--confirm` in `zarf package deploy/remove`
As a frequent user of both Zarf and UDS, I (and many others I've observed) often get tripped up by the inconsistency in the flags used for deployment confirmation.

- UDS **supports** `uds deploy <bundle> -c`
- Zarf **requires** `zarf package deploy <package> --confirm`

I propose adding `-c` as a short-hand alias for the existing `--confirm` flag on the `zarf package deploy` command.

This would allow for the following command syntax, which is more ergonomic and consistent with other tools in the ecosystem:

```bash
# Proposed new usage
zarf package deploy <package-name>.tar.zst -c
```

The current alternative is to remember to always type out the full --confirm flag. While functional, it's less efficient and, as noted, a common point of confusion for users.

### Additional context
This is a small change that would have a significant positive impact on the daily user experience. It's a simple step toward making the CLI more intuitive and forgiving.

This request is related to: #3756

rodneyxr:
## Description

This PR adds support for `-c` shorthand for all `--confirm` flags in zarf.

## Related Issue

Fixes #4061

## Checklist before merging


brandtkeller:
DCO needs to be signed - the action can provide more info

docs can be updated with `make docs-and-schema`

rodneyxr:
I've updated it. Hopefully it should be good now.

Issue #2103, opened 2023-10-26
Zarf custom package file name when writing to local disk


alexandrulita91:
### Is your feature request related to a problem? Please describe.
I have a scenario where I need to determine the package file name that will result from invoking zarf package pull/create. My current approach involves parsing the command's output, but this method may become unreliable in the future due to changes.

### Describe the solution you'd like

Add a new argument for the package pull/create commands, where the user is allowed to customize the package file name.
`format = "%p-%n-%a-%v.tar.zst"`
- p -> package prefix
- n -> metadata.name
- a -> architecture
- v -> version

**Input / Output example:**

- "%p-%n-%a-%v.tar.zst" -> zarf-package-test-amd64-0.1.0.tar.zst
- "my-package.tar.zst" -> my-package.tar.zst
- "my-package.tar" -> my-package.tar

### Describe alternatives you've considered
Add a new argument where the user can set the package file name.

ToddSlater:
We are also experiencing this requirement/issue in zarf 28. This should be an essential feature of the zarf package create command.

Noxsios:
Labeling this as "large" (`L`). Writing the code for the custom naming isnt prob too bad (could do some custom templating w/ `%`s or use Go `text/template`. Hard part would be double checking and ensuring our package search's do not heavily rely upon the current naming convention.

brandtkeller:
This still feels like a valid option for operations - but there remains the constraint of package naming being largely deterministic. Zarf will need to determine intent of package naming being more dynamic if we were to support this.

AustinAbro321:
One option could be allowing the the `--output` flag to rename the package file without changing the `.metadata.name` of the package. If the value given to output ends with `.tar` or `.zst` it would be considered a file and moved to that path. 

We could also allow go templating in that value similar to the work we've done in #3946, so an output string might look like --output=build/zarf-package-{{ .Metdata.Name}}-{{ .Metadata.Version }}.tar.zst

PR #4063, opened 2025-08-08
feat!: add flavor in file name

### LINKED ISSUE #3881: Include flavor to name of package created

When creating a zarf package and specifying a `flavor` it would be helpful if the flavor being built was specified in the name of the created package.

ex: `zarf-package-ude2-amd64-apples-0.0.1.tar.zst` when creating the apples flavor and `zarf-package-ude2-amd64-apples-0.0.1.tar.zst` when no flavor is specified.

- **Given** a flavor is specified
- **When** a package is being created
- **Then** the name of the package will include the flavor

Currently we use additional scripts in the CI process to append the flavor to the package names

### Additional context

Link to slack discussion: https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1748453665111549

a1994sc:
## Description

This adds the `flavor` to the file name, when pulling a package.

```shell
zarf package pull oci://ghcr.io/defenseunicorns/packages/uds/renovate:41.57.1-uds.0-upstream
```

will now create a file like the following:

```text
zarf-package-renovate-amd64-upstream-41.57.1-uds.0.tar.zst
```

instead of

```text
zarf-package-renovate-amd64-41.57.1-uds.0.tar.zst
```

## Other changes

Added logic that includes the `.metadata.name` in the file name when pulling an init package, whose name is not `init`

```shell
zarf package pull --skip-signature-validation oci://ghcr.io/colonel-byte/zarf-init-slim:v0.60.0-upstream
```

will now create a file like the following:

```text
zarf-init-zarf-init-slim-amd64-upstream-v0.60.0.tar.zst
```

instead of

```text
zarf-init-amd64-upstream-v0.60.0.tar.zst
```

I can remove this logic as it is not directly related to the below ticket

## Related Issue

Fixes #3881

## Checklist before merging


AustinAbro321:
The biggest thing to consider here are workflows that automatically create + push or pull + deploy. This change would break those workflows. That said, I do like this change, and it should be fairly easy for users to figure out what the issue is in their automation. Still, I'd like to here thoughts from other members of @zarf-dev/maintainers.

a1994sc:
It could be a feature flag, that Kit added recently, `fancy-names`, and then work it in as the default?

AustinAbro321:
While that is an option, I'd rather avoid it. Feature flags have to be documented, maintained, and deprecated. Thinking about the user flows, I'm not sure the better name is convenient enough for me to make sure I have the flag set. Similarly, if I ran into the problem I wouldn't want to disable the feature flag, I'd rather use the new name for my package.

mkcp:
IMO this is a good change for us to make. Two of the same package with different flavors do not necessarily have the same behavior (if so why use a flavor?) so being able to tell which is which and not having them collide on disk feels intuitive. There's lots of breaking changes to consider, many that we may not expect, so as much community feedback as possible on this change is crucial.

Regarding feature flagging it, it would add some amount of safety to making the change where users could undo it. I don't see a lot of value in having it disabled by default to start though. This doesn't necessarily enable new workflows, so keeping it disabled wouldn't get us necessary feedback. This is a good nudge for me to get features into beta with the appropriate website docs and cli support for viewing all flags though!

a1994sc:
> Two of the same package with different flavors do not necessarily have the same behavior (if so why use a flavor?) so being able to tell which is which and not having them collide on disk feels intuitive.

This is true.... To use the original parent of Zarf, Defense Unicorn, they have two-three flavors, `upstream`, `registry1`, and `unicorn`, where the difference is just the source of the images.

> There's lots of breaking changes to consider, many that we may not expect, so as much community feedback as possible on this change is crucial.

This is very true, I can cross post this issue into the slack channel and see what people think as well.

> Regarding feature flagging it, it would add some amount of safety to making the change where users could undo it. I don't see a lot of value in having it disabled by default to start though.

I do like the idea of having an on by default feature flag, maybe a little bit of middle ground would be to announce it has a disabled feature with the note saying in N+2 (seems to be about about 1 month for minor version bumps) releases it will be enabled by default. Allowing for A) package maintainers to the required work ahead of time to update the CI/CD logic and B) gives the package maintainers the opportunity to disable it if they can not update either logic for the time being.

> This is a good nudge for me to get features into beta with the appropriate website docs and cli support for viewing all flags though!

Ha, yea, I will say it took me a bit to understand the idea behind feature gates, but I think it does open the door for some cool things.

If we as a community decided to go with the feature approach, I would be more then happy to be a bit of a guinea pig with implementing outside of the core maintainers (e.i Austin, Kit, and Brandt).

AustinAbro321:
@a1994sc The default feature flag is definitely a reasonable option, and it's good that the users will be able to control their workflows ahead of time. Ultimately, we'll have to decide if we think users will want to take the extra action ahead of time, and if we want to document it.    

I think it'd be good to ask this question in slack and see what people think. Generally Zarf has said that we're okay with breaking changes on create, but really try to avoid breaking changes for deployers. This ride that line a bit since it might break some pull+deploy workflows, but it's always in a non-destructive way so it should be alright.

Racer159:
Just a consideration (this is still useful as is thanks for the PR!) - it could be nice to do `zarf-package-<name>-<arch>-<version>-<flavor>.tar.zst` (flavor after version) instead to make it easier to map back and forth between OCI registries and tarballs (since that would then just be `zarf-package-<last-part-of-repo-name>-<platform-arch>-<tag>.tar.zst` without needing to parse/split on the tag to build the package name).

a1994sc:
I'm totally fine with that, I can update that this week. Are there any thoughts on the `zarf-init` change?

AustinAbro321:
Yeah I can see why a user would want to name their packages. I'm not sure I want to special case the name `init`. Could you create a separate issue for this?

a1994sc:
Absolutely, I will create a new issue/pr based off this current PR.

a1994sc:
@AustinAbro321 so I removed the custom names for the init packages, however left the flavor indicator in the file name, if that is ok

soltysh (approved):
I'm supportive of this change, even though this might be breaking change, which we have to be loud about, to ensure people are aware of the change. But switching to globbing when matching new files should handle that easily.

AustinAbro321 (approved):
Adding a ! in the title to denote breaking change. Looks good to me

a1994sc:
I mean a lot of both my public and private CI/CD for zarf package using something to the effect of `zarf package publish zarf-*.tar.zst` so I agree with using the globbing

brandtkeller (approved):
This change enhances the default and paves way for more conversations on package naming constraints ( IE #2103 ). We'll provide a mention in the release notes for those who might be impacted - LGTM.

PR #4064, opened 2025-08-08
feat!(publish): add retry logic to publish operations

### LINKED ISSUE #3787: Support retries on package publishing
Zarf currently supports a configurable number of retries for the package deploy command. This is useful for certain environments that are prone to zarf image pushes into the zarf registry timing out or failing.

Similarly, some environments can have issues that interrupt publishing packages to a registry partway through. It would be helpful if zarf supported configurable retries for package publishing like it does for deploys to provide a workaround for these kinds of issues.

While the correct fix would be to resolve whatever is causing the issue, many times the issue is outside the control of those using zarf in a given environment and having more tools to workaround things is nice.

- **Given** a package publish command
- **When** an error is encountered that would normally cause zarf to throw an error and stop
- **Then** zarf retries up to the configured retry limit

brandtkeller:
## Description

Adding support for retry logic to the package publish operations (package, skeleton, oci). This allows for more robust/resilient handling of publish operations - that may take time on larger packages - to allow retrying.

Proposing the retry logic live within the `zoci` pkg. 

For local package/skeleton - opting to include all remote interactions during the `PushPackage()` logic as it makes the operation more resilient to network flakiness - something that may very well be a consistent Zarf constraint.

For OCI to OCI copy - updating the logic to utilize `oras` directly and create logic that mirrors the `Push` operation. Willing to discuss this further and revert if we see a reason. 

Operations that push an artifact to a remote registry should return the descriptor to enable follow-on operations. 

## Related Issue

Fixes #3787

## Checklist before merging


src/pkg/zoci/copier.go
```diff
@@ -5,53 +5,83 @@
...
+			if copyErr != nil {
+				return copyErr
+			}
+
+			// 2) Update/tag the destination index to the source tag
+			return dst.OrasRemote.UpdateIndex(ctx, tag, publishedDesc)
+		},
+		retry.Attempts(uint(retries)),
+		retry.Delay(500*time.Millisecond),
+		retry.MaxDelay(8*time.Second),
```
mkcp: nit: extract delay and maxdelay durations to package consts

src/cmd/viper.go
```diff
@@ -98,6 +98,7 @@ const (
 
 	VPkgPublishSigningKey         = "package.publish.signing_key"
 	VPkgPublishSigningKeyPassword = "package.publish.signing_key_password"
+	VPkgPublishRetries            = "package.publish.retries"
```
AustinAbro321, line 105: We should set a default for this field like we do for the other retry fields. This way the value comes up on the help text side. Currently the default is three since it's set to that automatically on the SDK side when the value is 0.

src/pkg/zoci/push.go
```diff
@@ -28,89 +29,119 @@ import (
...
+			publishedDesc, copyErr := oras.Copy(ctx, src, root.Digest.String(), trackedRemote, "", copyOpts)
+			if copyErr != nil {
+				return copyErr
+			}
+
+			return r.OrasRemote.UpdateIndex(ctx, r.Repo().Reference.Reference, publishedDesc)
+		},
+		retry.Attempts(uint(retries)),
+		retry.Delay(500*time.Millisecond),
+		retry.MaxDelay(8*time.Second),
```
mkcp: nit: extract delay and maxdelay durations to package consts

src/pkg/zoci/copier.go
```diff
@@ -5,53 +5,83 @@
...
 
-	if err := dst.Repo().Manifests().PushReference(ctx, expected, bytes.NewReader(b), srcRoot.Digest.String()); err != nil {
-		return err
-	}
+	tag := src.Repo().Reference.Reference // keep the source tag on the destination
 
-	tag := src.Repo().Reference.Reference
-	if err := dst.UpdateIndex(ctx, tag, expected); err != nil {
-		return err
+	err = retry.Do(
```
mkcp, line 44: Good use of retry library and its tuning options in each of these 👍

src/pkg/zoci/push.go
```diff
@@ -28,89 +29,119 @@ import (
 const OCITimestampFormat = time.RFC3339
 
 // PushPackage publishes the zarf package to the remote repository.
-func (r *Remote) PushPackage(ctx context.Context, pkgLayout *layout.PackageLayout, concurrency int) (err error) {
+func (r *Remote) PushPackage(ctx context.Context, pkgLayout *layout.PackageLayout, concurrency int, retries int) (_ ocispec.Descriptor, err error) {
```
mkcp: This is a reasonable breaking change imo

src/cmd/viper.go
AustinAbro321, line 105: Separate note, what do you think of making the default 1? I imagine a strong majority of the time that someone has a failure when publishing it'll be that there was an authentication problem such as a 403

src/pkg/zoci/push.go
AustinAbro321: The descriptor returned here isn't used, was this addition intentional?

src/pkg/zoci/copier.go
```diff
@@ -5,53 +5,83 @@
...
-	"oras.land/oras-go/v2/content"
+	"oras.land/oras-go/v2"
+
+	retry "github.com/avast/retry-go/v4"
 )
 
-// CopyPackage copies a zarf package from one OCI registry to another
-func CopyPackage(ctx context.Context, src *Remote, dst *Remote, concurrency int) (err error) {
+// CopyPackage copies a zarf package from one OCI registry to another using ORAS with retry.
+func CopyPackage(ctx context.Context, src *Remote, dst *Remote, concurrency int, retries int) (err error) {
```
mkcp: While we're changing this public function, would it make sense to move concurrency and retries to a config struct? They have defaults so they're optional, and that might help avoid breaking changes in the future.

src/pkg/zoci/copier.go
```diff
@@ -5,53 +5,83 @@
...
 	l := logger.From(ctx)
 	if concurrency <= 0 {
 		concurrency = DefaultConcurrency
 	}
```
mkcp, line 30: retry.Do probably guards against this already, but we ought to check for negative `retries` here - similar to `concurrency`. Zero retries is probably ok and allowed.

src/pkg/zoci/copier.go
```diff
@@ -5,53 +5,83 @@
...
+		func() error {
+			l.Info("copying package",
+				"src", src.Repo().Reference.String(),
+				"dst", dst.Repo().Reference.String(),
+				"ref", srcRef,
+			)
+			source := src.Repo() // implements oras.ReadOnlyTarget
+			trackedDst := images.NewTrackedTarget(
+				dst.Repo(),
+				0, // unknown total for registry→registry copy
```
AustinAbro321: Maybe adding progress here should be a separate PR, currently this would not report progress. https://github.com/zarf-dev/zarf/blob/8067c8b31deea98610917cdcd918ede5ac0d20af/src/internal/packager/images/progress.go#L27-L30

It should be possible to get all the descs from the mainfest though then sum the total size.

AustinAbro321 (changes requested):
A few small items. Also before merging this I would like to ask the user if they still run into this when `--oci-concurrency=1`. I'll make a comment in the issue

brandtkeller:
> A few small items. Also before merging this I would like to ask the user if they still run into this when `--oci-concurrency=1`. I'll make a comment in the issue

User feedback is always good. I'd probably wager we're solving for a separate constraint here than what oci-concurrency offers.

AustinAbro321:
I think there's a fair likelihood it's related in a majority network flakes. I know that when we switched to oras from crane we enabled the `oci-concurrency`, setting for images pushes. Setting it to one completely solved the problem for some users. Later on we set the oci-concurrency to one automatically on retry and we've heard far fewer reports of image push issues since then. https://github.com/zarf-dev/zarf/blob/8067c8b31deea98610917cdcd918ede5ac0d20af/src/internal/packager/images/push.go#L151 It's something we should consider here if it solves the problem for users or makes the flake less likely. 

There are differences between the usual flows though so oci-concurrency could be unrelated. Usually image pushes will go through a Kubernetes port-forward whereas usually package publishes will not. The registry pod also occasionally is rolled so multiple attempts by default is especially important in that case.

src/cmd/viper.go
brandtkeller: That is a great suggestion.

src/pkg/zoci/push.go
brandtkeller: intentional! much how `oras.Copy()` returns the descriptor - we too should consider this pattern as we look at future utility of the referral api. We could revert that for this PR - but If I was making a breaking change I believe it relevant for the change.

brandtkeller:
> I think there's a fair likelihood it's related in a majority network flakes. I know that when we switched to oras from crane we enabled the `oci-concurrency`, setting for images pushes. Setting it to one completely solved the problem for some users. Later on we set the oci-concurrency to one automatically on retry and we've heard far fewer reports of image push issues since then.

I agree! I interpret this request as more of a roboustness feature as opposed to single-root-cause. Zarf should expect to run in network constrained environments. Having the ability to retry operations that rely on the network should likely be a pattern we replicate more broadly.

src/pkg/zoci/copier.go
brandtkeller: Good point - for all three publish functions we will check for negative given the type conversion to `uint` and it's resulting value.

src/pkg/zoci/copier.go
brandtkeller: Agreed - I mentioned including a few changes in favor of breaking changes - this one makes a lot of sense too.

AustinAbro321:
@brandtkeller That's fair, and I might be too stringent on the request that we must know what the issue is in the user case, but I do think it helps inform how we design the fallback mechanism. I'm a bit torn on if this should be a feature, but I do see the argument, especially if there is only one attempt by default

src/pkg/zoci/copier.go
brandtkeller: Added

src/pkg/zoci/copier.go
brandtkeller: added

src/pkg/zoci/copier.go
brandtkeller: I agree - did some looking at what we need and do feel it extends scope and would be well fit for a future PR given parity.

src/pkg/zoci/push.go
brandtkeller: To elaborate further - I believe the referral API is going to be a big enabler in the future with regards to artifacts that live _beside_ a package and should be discoverable. This is an attempt to include them in breaking changes now so that we can later integrate.

src/pkg/packager/create.go
```diff
@@ -97,7 +97,12 @@ func Create(ctx context.Context, packagePath string, output string, opts CreateO
 		if err != nil {
 			return "", err
 		}
-		err = remote.PushPackage(ctx, pkgLayout, opts.OCIConcurrency)
+		// Setting a default here for retries as the flag is less intuitive during create
+		publishOptions := zoci.PublishOptions{
+			Retries:        2,
```
AustinAbro321: IMO we should keep the default at 1 here, and if we get the feedback we can add a publish attempts flag or something along those lines

src/pkg/zoci/push.go
```diff
@@ -28,89 +29,119 @@ import (
...
+			l.Warn("retrying package push",
+				"attempt", n+1,
+				"max_attempts", opts.Retries,
+				"error", err,
+			)
+		}),
+	)
 	if err != nil {
-		return err
+		return ocispec.Descriptor{}, fmt.Errorf("publish failed after retries: %w", err)
```
AustinAbro321: Given that we're logging when we retry, I think it's reasonable to not include "after retries" here. Reads a bit awkward in the default case of one attempt

src/pkg/zoci/push.go
```diff
@@ -28,89 +29,119 @@ import (
...
+		retry.Delay(defaultDelayTime),
+		retry.MaxDelay(defaultMaxDelayTime),
+		retry.DelayType(retry.BackOffDelay), // exponential backoff
+		retry.LastErrorOnly(true),
+		retry.Context(ctx),
+		retry.OnRetry(func(n uint, err error) {
+			l.Warn("retrying package push",
+				"attempt", n+1,
+				"max_attempts", opts.Retries,
+				"error", err,
```
AustinAbro321: This is always done even on the last attempt (or first and only) 
```bash
../../build/zarf package publish . oci://ghcr.io/someone-else/
2025-08-14 16:58:16 INF loading skeleton package path=.
2025-08-14 16:58:16 INF pushing package to registry destination=ghcr.io/someone-else/helm-charts:0.0.1 architecture=skeleton size=294.59 KBs
2025-08-14 16:58:16 WRN retrying package push attempt=1 max_attempts=1 error=POST "https://ghcr.io/v2/someone-else/helm-charts/blobs/uploads/": response status code 403: denied: permission_denied: create_package
2025-08-14 16:58:16 ERR could not push package: publish failed after retries: POST "https://ghcr.io/v2/someone-else/helm-charts/blobs/uploads/": response status code 403: denied: permission_denied: create_package
```

AustinAbro321 (changes requested):
A few more comments based on the UX for this

src/pkg/packager/create.go
brandtkeller: good call - updated

src/pkg/zoci/push.go
brandtkeller: Updated!

src/pkg/zoci/push.go
brandtkeller: Good point - added a conditional for > 1. Considered changing language to "Attempts" but it created a misalignment.

AustinAbro321: I think the main issue though is that we say we are retrying after the last retry has already been preformed. For example: 

```sh
../../build/zarf package publish zarf-package-helm-charts-amd64-0.0.1.tar.zst oci://ghcr.io/someoneelse --retries=2
2025-08-15 14:51:07 INF pushing package to registry destination=ghcr.io/someoneelse/helm-charts:0.0.1 architecture=amd64 size=29.28 MBs
2025-08-15 14:51:07 WRN retrying package push attempt=1 max_attempts=2 error=POST "https://ghcr.io/v2/someoneelse/helm-charts/blobs/uploads/": response status code 403: denied: permission_denied: create_package
2025-08-15 14:51:08 INF pushing package to registry destination=ghcr.io/someoneelse/helm-charts:0.0.1 architecture=amd64 size=29.28 MBs
2025-08-15 14:51:08 WRN retrying package push attempt=2 max_attempts=2 error=POST "https://ghcr.io/v2/someoneelse/helm-charts/blobs/uploads/": response status code 403: denied: permission_denied: create_package
```

brandtkeller: fair point - I only fixed the first half. Updated to check for last attempt as well.

src/pkg/zoci/copier.go
brandtkeller: [Issue](https://github.com/zarf-dev/zarf/issues/4093) created to track adding this support.

AustinAbro321: gotcha, in that case let's remove the tracked destination from here so users don't see the warning `total bytes is a non-positive integer, can't report progress` during each reporting interval.

brandtkeller: Updated to remove the tracked destination.

src/pkg/packager/publish.go
```diff
@@ -38,6 +40,11 @@ func PublishFromOCI(ctx context.Context, src registry.Reference, dst registry.Re
 	l := logger.From(ctx)
 	start := time.Now()
 
+	// Allow zero but other sanitize retries if less-than zero
+	if opts.Retries < 0 {
+		opts.Retries = config.ZarfDefaultRetries
+	}
```
AustinAbro321: retry-go does allow 0 in retry.Attempts(), but it will retry infinitely until the function succeeds. Given that 0 will be the default at the SDK layer I believe we should change this. 

We might also want this to default to 1 instead of 3 as it's consistent with the CLI and the most common user flow.

brandtkeller: i agree with updating the default from 3 to 1. 

Intent was to support allowing the infinite option - but reflecting that feels like an sub-optimal configuration so I will get this updated.

src/cmd/viper.go
```diff
@@ -221,4 +222,7 @@ func setDefaults() {
 
 	// Deploy opts that are non-zero values
 	v.SetDefault(VPkgDeployTimeout, config.ZarfDefaultTimeout)
+
+	// Package publish opts that are non-zero values
+	v.SetDefault(VPkgPublishRetries, 1)
```
soltysh, line 227: Any particular reason why not sticking with the `config.ZarfDefaultRetries` which is currently 3 and applies everywhere else?

src/pkg/packager/publish.go
```diff
@@ -141,6 +164,11 @@ type PublishSkeletonOptions struct {
 func PublishSkeleton(ctx context.Context, path string, ref registry.Reference, opts PublishSkeletonOptions) (registry.Reference, error) {
 	l := logger.From(ctx)
 
+	// disallow and sanitize <= 0
+	if opts.Retries < 1 {
```
soltysh: This sounds more like a validation error rather than something we should silently fix for the user.

src/pkg/zoci/copier.go
```diff
@@ -5,53 +5,79 @@
...
-	if err != nil {
-		return err
-	}
-	l.Info("copying package",
-		"src", src.Repo().Reference.String(),
-		"dst", dst.Repo().Reference.String())
-	if err := oci.Copy(ctx, src.OrasRemote, dst.OrasRemote, nil, concurrency, nil); err != nil {
-		return err
+	if opts.OCIConcurrency <= 0 {
+		opts.OCIConcurrency = DefaultConcurrency
```
soltysh, line 21: Here, would be the place where I'd default the retries, if they are less than 1.

src/cmd/viper.go
soltysh: Just uncovered the answer in https://github.com/zarf-dev/zarf/pull/4064#discussion_r2288840791, so ignore it.

src/pkg/packager/publish.go
brandtkeller: From the SDK perspective this is establishing the default of one for the optional value. When not provided the option is the zero-value `0` - so we haven't been returning an error here. I am inclined to add a debug log here. Thoughts?

src/pkg/zoci/copier.go
brandtkeller: this is valid regardless of other changes

brandtkeller: Updated

src/pkg/packager/publish.go
soltysh: From a slack discussion with Brandt. My thinking is that `retries` is an input parameter, even though it's optional we should still validate, which is different from defaulting. Defaulting kicks in when the value is not set. But if a user explicitly sets an invalid value that should fail, not be silently defaulted.

brandtkeller: Added checking for negative values

soltysh (approved):
lgtm

Issue #3926, opened 2025-06-18
Data Injection Refactor


nick-adkins:
Data injection shells out to both kubectl and tar currently. This solution is both hard to test and relies on the presence of binaries on the host to function properly. The new refactored solution should remove any use of calling any external binaries.

nick-adkins:
Associated issue https://github.com/zarf-dev/zarf/issues/2422

nick-adkins:
Also associated issue #2265

AustinAbro321:
Prior to this being implemented we should implement #2181 especially since public registries often limit the size of images that can be pushed. For example ghcr.io has a limit of 10gb per layer. So users may want to skip this step and pack their image locally. It should be much much faster to pack a large image by tarball rather than from the daemon

willswire:
Recently ran into a use-case where it would have been nice to do data injections across Zarf packages (to allow for "data only" packages). This might be OBE with the consideration to move to image-based volumes, but wanted to chime in here with the user story.

AustinAbro321:
@willswire Thanks for the feedback! Would the strategy laid out in this document work for you? https://deploy-preview-4068--zarf-docs.netlify.app/best-practices/data-injections-migration/. I believe it's already an improvement over data injections. 

Looks like image based volumes should be stable in 1.35 https://github.com/kubernetes/enhancements/issues/4639

AustinAbro321:
Closing as data Injections is already deprecated. It will not be removed until after the v1alpha1 schema is removed (at least one year but likely a bit more). Data Injections will not be in the next schema version. See the migration guide here https://docs.zarf.dev/best-practices/data-injections-migration/

PR #4068, opened 2025-08-11
feat: deprecate data injections

AustinAbro321:
## Description

This deprecates data injections and documents a migration guide to move off of data injections. The kiwix example has been migrated to the strategy outlined in the guide. 

## Related Issue

Relates to #3926 

## Checklist before merging


site/src/content/docs/best-practices/data-injections-migration.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+**Key Changes:**
+- Replace data injection marker waiting logic with direct data copying.
+- Use your data container image.
+
+Your app should be ready to deploy. If there are any reasons that this method does not work for you, please comment in issue [#3926](https://github.com/zarf-dev/zarf/issues/3926)
+
+## Future Alternatives
+
+Once the new Kubernetes feature [OCI volume sources](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/4639-oci-volume-source) is considered stable, this document will recommend that method. OCI volumes sources, in beta as of Kubernetes 1.33, provide a simpler way to mount data from an image into a pod.
```
a1994sc: Was going to comment on using the `oci volume` source, glad to see it here

examples/kiwix/manifests/deployment.yaml
```diff
@@ -16,18 +16,16 @@ spec:
...
-          image: alpine:3.18
+          image: kiwix-data:local
           command:
-            [
-              "sh",
-              "-c",
-              # This command looks for the Zarf "data injection marker" which is a timestamped file that is injected after everything else and marks the injection as complete.
-              'while [ ! -f /data/###ZARF_DATA_INJECTION_MARKER### ]; do echo "waiting for zarf data sync" && sleep 1; done; echo "we are done waiting!"',
-            ]
+            - sh
```
brandtkeller, line 23: Something to call out in this example and maybe even more broadly in the migration docs is persistence requirements. 

Data injections require persistence - because your pod may be killed and then there is no process to re-inject the data. 

In this example - we no longer need a persistent volume - instead an `emptyDir` would be more than suitable given that we have a data image in the package/environment to re-use anytime the init-container runs again. 

There may be instances where you still want persistence - but likely that Is a broader kubernetes data management question.

brandtkeller (changes requested):
One passing thought for requirements - otherwise the current updates look good.

examples/kiwix/manifests/deployment.yaml
AustinAbro321: This is not something I thought of, and is an excellent call out thank you! I changed this example to use emptydir and added some notes to the guide

brandtkeller: My only other comment on this thread being if there is any reason to retain the `examples/kiwix/manifests/persistence.yaml` file?

AustinAbro321: I believe you're seeing `src/test/packages/23-data-injections/manifests/persistence.yaml`, which I kept around so we have an e2e test with data injections

brandtkeller: Yes! diff semantics got me.

Issue #4045, opened 2025-08-05
Improve injector testing and release process


AustinAbro321:
### Describe what should be investigated or refactored
The current testing for the injector is inadequate and releasing is cumbersome. 

The injector release process is to 
1. Merge changes to main.rs
2. run a workflow dispatch on main to release to an s3 bucket
3. Merge changes to the init package to use the new s3 bucket injector

No automatic testing happens at step 1.  To test the new injector you have to edit the zarf.yaml in the registry package to point to your local injector binary on your file system. It will be tested in step 3, but by then the broken code is already on main.  Additionally, I've heard complaints from multiple users that the injector in s3 is hard to track and update automatically using tools like renovate. 

For improved testing we should do the following: 
- Build an init package where on pull request it builds and deploys an init package using the local injector. 
- Build the injector and make sure it is less than 1024 mb. 

I believe the strongest option is to have another repo that only includes the injector. It will use a skeleton init package to preform the testing. Releases can be done totally separate from Zarf.  

Another option would be to do something similar to #2024, however the injector release process being tied to the Zarf release process presents a practical issue. We would release the injector and Zarf without updating Zarf to use the new injector. We could separate these, but the tag would still be seen by Go so it would confuse SDK users.

AustinAbro321:
The injector repo (https://github.com/zarf-dev/zarf-injector) has been created and this repo now uses that

PR #4070, opened 2025-08-11
chore: remove injector from this repo, use injector from zarf-dev/zarf-injector 

### LINKED ISSUE #2024: Add a tag for zarf injector releases to the codebase
We should do a tag-based release workflow for the zarf-injector as well so that it is easier to correlate the source code of a given `zarf-injector` binary back to it's source code.  This could work very similarly to the existing `release` workflow.

https://github.com/defenseunicorns/zarf/blob/main/.github/workflows/release.yml
https://github.com/defenseunicorns/zarf/blob/main/.github/workflows/build-rust-injector.yml

### Additional context
[Add any other context or screenshots about the technical debt here.](https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1695259616216399?thread_ts=1689908320.032309&cid=C03B6BJAUJ3)

AustinAbro321:
## Description

This changes Zarf to use the injector for the repo that now owns it https://github.com/zarf-dev/zarf-injector

## Related Issue

Fixes #2024, #4045 

## Checklist before merging

### DIFF
--- a/packages/zarf-registry/zarf.yaml
+++ b/packages/zarf-registry/zarf.yaml
@@ -111,9 +111,9 @@ components:
         architecture: amd64
     files:
       # Rust Injector Binary
-      - source: https://github.com/zarf-dev/zarf-injector/releases/download/###ZARF_PKG_TMPL_INJECTOR_VERSION###/zarf-injector-amd64
+      - source: https://github.com/zarf-dev/zarf-injector/releases/download/v0.1.0/zarf-injector-amd64
         target: "###ZARF_TEMP###/zarf-injector"
-        shasum: "###ZARF_PKG_TMPL_INJECTOR_AMD64_SHASUM###"
+        shasum: "97d2c3ec3c789a866c7bb88ed1d6819f8b279762f9398c8ec5ad70458f23bf85"
         executable: true
 
   - name: zarf-injector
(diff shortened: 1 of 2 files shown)


packages/zarf-registry/zarf.yaml
```diff
@@ -111,7 +111,7 @@ components:
         architecture: amd64
     files:
       # Rust Injector Binary
-      - source: https://zarf-init-resources.s3.us-east-1.amazonaws.com/injector/###ZARF_PKG_TMPL_INJECTOR_VERSION###/zarf-injector-amd64
+      - source: https://github.com/zarf-dev/zarf-injector/releases/download/###ZARF_PKG_TMPL_INJECTOR_VERSION###/zarf-injector-amd64
```
brandtkeller: Is there any utility to adding a variable for source base location? I don't see a scenario where someone needs the ability to use the older released versions for some time - but it is a good fallback and may enable development workflows more easily as well? Food for thought.

AustinAbro321: Someone actually tried to do this in https://github.com/zarf-dev/zarf/pull/3752/files, but it leaves us with this issue. https://github.com/zarf-dev/zarf/issues/3762/. 

Still I'd rather avoid it for now. If someone does want to use a custom injector then they could do what we are doing here https://github.com/zarf-dev/zarf-injector/blob/main/test/zarf.yaml#L6-L14. I believe a version variable will be simpler to update for renovate things too. 

I'm now thinking the flipside, should this be a variable at all? If you are a user setting this variable then you are already maintaining a custom init package, and so you have to set this. I've heard from at least three users that it's hard to track and update the injector, it would be easier now, but I'm not sure what value they get from deciding the version themselves given that they already import a version of the init package skeleton. Thoughts?

brandtkeller: greats points. 

I believe that is a valid question. The general premise of versions in zarf probably deserves some attention. 

I don't have a hard opinion on hardcoding - it may be beneficial. 

Should the injector zarf manifest live in this repository? or be managed and versioned in alignment with the injector artifacts in the other repository?

AustinAbro321: That's an interesting question. I think the injector components should stay in this repo. The injector repo wouldn't use the Zarf package it creates, and a package with only the injector binaries in them, wouldn't be useful on it's own.

soltysh (approved):
lgtm

packages/zarf-registry/zarf.yaml
brandtkeller: I think that's fair - I'd like to explore more about the software delivery angle in future and best practices after we look at imports and skeletons more.

PR #4078, opened 2025-08-13
feat: allow interactive remove

AustinAbro321:
## Description

This allows users to preform an interactive remove when `--confirm` is not set

## Related Issue

Fixes #

Relates to #

## Checklist before merging

### DIFF
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -1132,7 +1132,6 @@ func newPackageRemoveCommand(v *viper.Viper) *cobra.Command {
 	}
 
 	cmd.Flags().BoolVarP(&config.CommonOptions.Confirm, "confirm", "c", false, lang.CmdPackageRemoveFlagConfirm)
-	_ = cmd.MarkFlagRequired("confirm")
 	cmd.Flags().StringVar(&pkgConfig.PkgOpts.OptionalComponents, "components", v.GetString(VPkgDeployComponents), lang.CmdPackageRemoveFlagComponents)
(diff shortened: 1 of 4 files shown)


src/cmd/package.go
```diff
@@ -1180,6 +1179,21 @@ func (o *packageRemoveOptions) run(cmd *cobra.Command, args []string) error {
...
 		NamespaceOverride: o.namespaceOverride,
 	}
+	logger.From(ctx).Info("loaded package for removal", "name", pkg.Metadata.Name)
+	err = utils.ColorPrintYAML(pkg, nil, false)
+	if err != nil {
+		return fmt.Errorf("unable to print package definition: %w", err)
+	}
+	if !config.CommonOptions.Confirm {
+		prompt := &survey.Confirm{
+			Message: "remove this Zarf package?",
```
mkcp: nit: 
```suggestion
			Message: "Remove this Zarf package?",
```

mkcp (dismissed):
LGTM with style nit to match other prompt text.

brandtkeller (approved):
Changes look good. Noticed an odd behavior in testing - where an invalid component name used with `--components` on remove resulted in the full package being removed - we should open a new issue for this.

rodneyxr:
> Changes look good. Noticed an odd behavior in testing - where an invalid component name used with `--components` on remove resulted in the full package being removed - we should open a new issue for this.

Just chiming in here but we also experienced this issue as well using zarf v0.58.0 while trying to remove a component within a uds core package. We thought it was expected behavior so I never reported this but it did cause some issues and broke our core deployment putting it in a funky state...

PR #4111, opened 2025-08-20
feat: remove injector in parallel

### LINKED ISSUE #3339: Run Zarf injector removal in parallel
There is potential to save about `20-30` seconds during `zarf init`. This is because Zarf wait for the injector to be completely removed after the registry is setup. This removal could take place while the agent and git server are being deployed to save time. 

The code will need to be refactored to allow Zarf to wait for the injector removal thread to finish after the rest of init is completed.

### Additional context
This should be implemented after deploy and init have been refactored, see - #2969

AustinAbro321:
## Description

This saves about 15-20 seconds on standard init. An additional 10 seconds or so if the git server is deployed.

## Related Issue

Fixes #3339

## Checklist before merging

### DIFF
--- a/src/pkg/packager/deploy.go
+++ b/src/pkg/packager/deploy.go
@@ -68,11 +68,11 @@ type DeployOptions struct {
 // deployer tracks mutable fields across deployments. Because components can create a cluster and create state
 // any of these fields are subject to change from one component to the next
 type deployer struct {
-	s                  *state.State
-	c                  *cluster.Cluster
-	vc                 *variables.VariableConfig
-	hpaModified        bool
-	stopInjectionGroup *errgroup.Group
+	s            *state.State
+	c            *cluster.Cluster
+	vc           *variables.VariableConfig
+	hpaModified  bool
+	cleanupGroup *errgroup.Group
 }
 
 // DeployResult is the result of a successful deploy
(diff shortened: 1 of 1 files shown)


src/pkg/packager/deploy.go
```diff
@@ -305,9 +314,11 @@ func (d *deployer) deployInitComponent(ctx context.Context, pkgLayout *layout.Pa
 
 	// Do cleanup for when we inject the seed registry during initialization
 	if isSeedRegistry {
-		if err := d.c.StopInjection(ctx); err != nil {
-			return nil, fmt.Errorf("failed to delete injector resources: %w", err)
-		}
+		l.Info("removing injector pod in parallel")
+		d.stopInjectionGroup, _ = errgroup.WithContext(ctx)
```
brandtkeller: My only nit with this is:
1. if we should use the derived context 
2. make this more generic to open for other goroutine utilization

Otherwise I don't have hard opinions. Open to thoughts from others.

brandtkeller (commented):
initial thoughts while testing.

src/pkg/packager/deploy.go
AustinAbro321: Agree on both points. I misunderstood the derived context and thought it would be cancelled early. Thanks and updated

AustinAbro321:
Almost certaintly this only takes a while because we're correctly handling a sigterm in the rust application so Kubernetes is waiting for the default grace period (30s). A better solution is to handle sigterm in the injector or lower the default grace period

PR #4112, opened 2025-08-20
chore: update dependabot github actions frequency & grouping

AustinAbro321:
## Description

This lowers the frequency of GitHub actions from dependabot to once a weekly to lower noise. It also groups together actions from the [actions organization](https://github.com/actions), as these are generally trusted and break very rarely so group reviews are not complex.

## Checklist before merging

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -21,8 +21,12 @@ updates:
           - "github.com/docker/*"
   - package-ecosystem: github-actions
     directory: /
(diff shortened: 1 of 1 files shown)


brandtkeller (approved):
I don't see any mention of cadence in the OpenSSF Scorecard [checks](https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool) - lgtm

PR #4131, opened 2025-08-27
experiment: downgrade helm to 3.17.2 at zarf latest

mkcp:
## Description
This is a bit of a shot in the dark but the user has an issue with maxmimum chart size that was imposed in 3.17.3. I'm just pushing this to make CI run and see if there's any blockers to the downgrade that I haven't caught locally.

## Checklist before merging


mkcp:
Experiment successful! Nothing major appears to be broken, that's not a guarantee that there's no subtle incompatibilities that our unit testing and e2e suite have not caught.

PR #4136, opened 2025-08-28
feat: Add Values feature

### LINKED ISSUE #3946: Zarf Values and Go-templating Parent issue
Zarf Values and go-templates are major UX and architectural shifts in how to parameterize and generate assets through Zarf. The core of these changes is the addition of the packages `internal/value` and `internal/template`. Though they're currently internal, the long term goal is to make these facilities public as part of Zarf's SDK and at least design them to that quality.

#### package value
`value` declares various types and methods that deal with arbitrarily nested basic go types and collections. Functions, channels, and the like are not supported. The package gravitates around the `value.Values` type, and the primary intent for `value.Values` is to give users ways to declare and customize data for the purpose of injecting into Zarf-managed assets. `value.Values` offers users the `Values.Set(...)` and `Values.Extract(...)` methods which take a `value.Path`. A Path allows basic types and collections can be selectively applied or retrieved e.g. `".here.is.myData"` - or simply get everything with a path of `"."`. `Values.DeepMerge(...Values)` also allows for arbitrarily many values maps to be combined together.

Currently, `value.Values` is _not_ threadsafe, though that might change in the future. The methods `Clone() Values` and `Snapshot() map[string]any` would help with various threadsafe usecases, as well as a NewValues constructor method. While I was experimenting with this, a simple internal `RWMutex` seemed like the best path forward, as most uses of Values read many keys at once and writes tend to happen infrequently. `sync.Map` is not recommended because the sheer number of type assertions to handle nested collections would be a royal nightmare. More granular locking over individual keys would help optimize for extensive writes, but that's highly unlikely to match Zarf's usecase, even with concurrent implementations of package methods, or parallel components / actions.

#### package template
`template` can be thought of as "just" a Zarf-specific wrapper around text/template with the `Apply` functions, and offers the type `template.Objects` to assist in extracting Zarf context into predictable locations for users importing Values and other data. The `Objects.With...` methods offer facilities to map whatever templating context is necessary and available for the template apply.

While not in scope for Values and go-templating's GA, over time I do think we can merge the existing `###ZARF...###` style templates under packager and variables _into_ `internal/template`. The current vars, consts, and template behavior is so critical to how Zarf works and how users build their packages that I expect we will need to offer a long tail of support for it. That doesn't mean we have to keep their current _architecture_ in place, and as we gracefully deprecate parts of the behavior and wrap these deprecated components behind feature flags.

#### Alpha:
See description of: #4136 

#### Beta:
Scope of follow-up work described below. Each of these will be added to the Values alpha until it reaches a point of reasonable stability. Then we'll shift it over to beta as a kind of "GA Candidate" and aim to minimize further breaking changes.
#4222 
#4224 
#4225 
#4226
https://github.com/zarf-dev/zarf/issues/4282

#### GA:
- User feedback critical, maybe create a GitHub discussion and funnel slack discussion there?
- Replace uses of variables in the init package with values https://github.com/zarf-dev/zarf/issues/4328
- Update docs.zarf.dev: https://docs.zarf.dev/ref/values/
- Blog post

#### Post-GA:
- Continue phasing out variables in examples, docs, and other materials.

mkcp:
## Description

WIP, see #3946 for scope of implementation.

## Related Issue
Fixes #3946 

## Checklist before merging

### DIFF
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -23,6 +23,7 @@ import (
 	goyaml "github.com/goccy/go-yaml"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"github.com/zarf-dev/zarf/src/internal/value"
 	"github.com/zarf-dev/zarf/src/pkg/packager"
 	"oras.land/oras-go/v2/registry"
 
@@ -70,6 +71,7 @@ type packageCreateOptions struct {
 	output                  string
 	differentialPackagePath string
 	setVariables            map[string]string
+	valuesFiles             []string
 	sbom                    bool
 	sbomOutput              string
 	skipSBOM                bool
@@ -114,6 +116,7 @@ func newPackageCreateCommand(v *viper.Viper) *cobra.Command {
 	cmd.Flags().IntVarP(&o.maxPackageSizeMB, "max-package-size", "m", v.GetInt(VPkgCreateMaxPackageSize), lang.CmdPackageCreateFlagMaxPackageSize)
 	cmd.Flags().StringToStringVar(&o.registryOverrides, "registry-override", v.GetStringMapString(VPkgCreateRegistryOverride), lang.CmdPackageCreateFlagRegistryOverride)
 	cmd.Flags().StringVarP(&o.flavor, "flavor", "f", v.GetString(VPkgCreateFlavor), lang.CmdPackageCreateFlagFlavor)
+	cmd.Flags().StringSliceVarP(&o.valuesFiles, "values", "v", v.GetStringSlice(VPkgCreateValues), lang.CmdPackageCreateFlagValuesFiles)
 
 	cmd.Flags().StringVar(&o.signingKeyPath, "signing-key", v.GetString(VPkgCreateSigningKey), lang.CmdPackageCreateFlagSigningKey)
 	cmd.Flags().StringVar(&o.signingKeyPassword, "signing-key-pass", v.GetString(VPkgCreateSigningKeyPassword), lang.CmdPackageCreateFlagSigningKeyPassword)
@@ -156,7 +159,14 @@ func (o *packageCreateOptions) run(ctx context.Context, args []string) error {
 	}
 
 	v := getViper()
+
+	// Merge SetVariables and config variables.
 	o.setVariables = helpers.TransformAndMergeMap(v.GetStringMapString(VPkgCreateSet), o.setVariables, strings.ToUpper)
+
+	values, err := value.ParseFiles(o.valuesFiles)
+	if err != nil {
+		return err
+	}
 
 	cachePath, err := getCachePath(ctx)
 	if err != nil {
@@ -175,6 +185,7 @@ func (o *packageCreateOptions) run(ctx context.Context, args []string) error {
 		DifferentialPackagePath: o.differentialPackagePath,
 		RemoteOptions:           defaultRemoteOptions(),
 		CachePath:               cachePath,
+		Values:                  values,
 	}
 	_, err = packager.Create(ctx, baseDir, o.output, opt)
 	// NOTE(mkcp): LintErrors are rendered with a table
@@ -190,6 +201,7 @@ func (o *packageCreateOptions) run(ctx context.Context, args []string) error {
 
 type packageDeployOptions struct {
 	namespaceOverride string
+	valuesFiles       []string
 }
 
 func newPackageDeployCommand(v *viper.Viper) *cobra.Command {
@@ -214,6 +226,7 @@ func newPackageDeployCommand(v *viper.Viper) *cobra.Command {
 
 	cmd.Flags().IntVar(&pkgConfig.PkgOpts.Retries, "retries", v.GetInt(VPkgRetries), lang.CmdPackageFlagRetries)
 	cmd.Flags().StringToStringVar(&pkgConfig.PkgOpts.SetVariables, "set", v.GetStringMapString(VPkgDeploySet), lang.CmdPackageDeployFlagSet)
+	cmd.Flags().StringSliceVarP(&o.valuesFiles, "values", "v", v.GetStringSlice(VPkgDeployValues), lang.CmdPackageDeployFlagValuesFiles)
 	cmd.Flags().StringVar(&pkgConfig.PkgOpts.OptionalComponents, "components", v.GetString(VPkgDeployComponents), lang.CmdPackageDeployFlagComponents)
 	cmd.Flags().StringVar(&pkgConfig.PkgOpts.Shasum, "shasum", v.GetString(VPkgDeployShasum), lang.CmdPackageDeployFlagShasum)
 	cmd.Flags().StringVarP(&o.namespaceOverride, "namespace", "n", v.GetString(VPkgDeployNamespace), lang.CmdPackageDeployFlagNamespace)
@@ -237,8 +250,15 @@ func (o *packageDeployOptions) run(cmd *cobra.Command, args []string) (err error
 	}
 
 	v := getViper()
+
+	// Merge SetVariables and config variables.
 	pkgConfig.PkgOpts.SetVariables = helpers.TransformAndMergeMap(
 		v.GetStringMapString(VPkgDeploySet), pkgConfig.PkgOpts.SetVariables, strings.ToUpper)
+
+	values, err := value.ParseFiles(o.valuesFiles)
+	if err != nil {
+		return err
+	}
 
 	cachePath, err := getCachePath(ctx)
 	if err != nil {
@@ -269,6 +289,7 @@ func (o *packageDeployOptions) run(cmd *cobra.Command, args []string) (err error
 		Retries:                pkgConfig.PkgOpts.Retries,
 		OCIConcurrency:         config.CommonOptions.OCIConcurrency,
 		SetVariables:           pkgConfig.PkgOpts.SetVariables,
+		Values:                 values,
 		NamespaceOverride:      o.namespaceOverride,
 		RemoteOptions:          defaultRemoteOptions(),
 	}
@@ -634,6 +655,7 @@ type packageInspectValuesFilesOptions struct {
 	kubeVersion             string
 	setVariables            map[string]string
 	outputWriter            io.Writer
+	valuesFiles             []string
 }
 
 func newPackageInspectValuesFilesOptions() *packageInspectValuesFilesOptions {
@@ -659,6 +681,7 @@ func newPackageInspectValuesFilesCommand() *cobra.Command {
 	cmd.Flags().StringVar(&o.components, "components", "", "comma separated list of components to show values files for")
 	cmd.Flags().StringVar(&o.kubeVersion, "kube-version", "", lang.CmdDevFlagKubeVersion)
 	cmd.Flags().StringToStringVar(&o.setVariables, "set", v.GetStringMapString(VPkgDeploySet), lang.CmdPackageDeployFlagSet)
+	cmd.Flags().StringSliceVarP(&o.valuesFiles, "values", "v", v.GetStringSlice(VPkgDeployValues), lang.CmdPackageDeployFlagValuesFiles)
 
 	return cmd
 }
@@ -669,7 +692,14 @@ func (o *packageInspectValuesFilesOptions) run(ctx context.Context, args []strin
 		return err
 	}
 	v := getViper()
+
+	// Merge SetVariables and config variables.
 	o.setVariables = helpers.TransformAndMergeMap(v.GetStringMapString(VPkgDeploySet), o.setVariables, strings.ToUpper)
+
+	values, err := value.ParseFiles(o.valuesFiles)
+	if err != nil {
+		return err
+	}
 
 	cachePath, err := getCachePath(ctx)
 	if err != nil {
@@ -697,6 +727,7 @@ func (o *packageInspectValuesFilesOptions) run(ctx context.Context, args []strin
 	resourceOpts := packager.InspectPackageResourcesOptions{
 		SetVariables: o.setVariables,
 		KubeVersion:  o.kubeVersion,
+		Values:       values,
 	}
 	resources, err := packager.InspectPackageResources(ctx, pkgLayout, resourceOpts)
 	if err != nil {
@@ -720,6 +751,7 @@ type packageInspectManifestsOptions struct {
 	components              string
 	kubeVersion             string
 	setVariables            map[string]string
+	valuesFiles             []string
 	outputWriter            io.Writer
 }
 
@@ -745,6 +777,7 @@ func newPackageInspectShowManifestsCommand() *cobra.Command {
 	cmd.Flags().StringVar(&o.components, "components", "", "comma separated list of components to show manifests for")
 	cmd.Flags().StringVar(&o.kubeVersion, "kube-version", "", lang.CmdDevFlagKubeVersion)
 	cmd.Flags().StringToStringVar(&o.setVariables, "set", v.GetStringMapString(VPkgDeploySet), lang.CmdPackageDeployFlagSet)
+	cmd.Flags().StringSliceVarP(&o.valuesFiles, "values", "v", v.GetStringSlice(VPkgDeployValues), lang.CmdPackageDeployFlagValuesFiles)
 
 	return cmd
 }
@@ -755,7 +788,14 @@ func (o *packageInspectManifestsOptions) run(ctx context.Context, args []string)
 		return err
 	}
 	v := getViper()
+
+	// Merge SetVariables and config variables.
 	o.setVariables = helpers.TransformAndMergeMap(v.GetStringMapString(VPkgDeploySet), o.setVariables, strings.ToUpper)
+
+	values, err := value.ParseFiles(o.valuesFiles)
+	if err != nil {
+		return err
+	}
 
 	cachePath, err := getCachePath(ctx)
 	if err != nil {
@@ -783,6 +823,7 @@ func (o *packageInspectManifestsOptions) run(ctx context.Context, args []string)
 	resourceOpts := packager.InspectPackageResourcesOptions{
 		SetVariables: o.setVariables,
 		KubeVersion:  o.kubeVersion,
+		Values:       values,
 	}
 
 	resources, err := packager.InspectPackageResources(ctx, pkgLayout, resourceOpts)
--- a/src/internal/feature/feature.go
+++ b/src/internal/feature/feature.go
@@ -204,6 +204,7 @@ func featuresToMap(fs []Feature) map[Name]Feature {
 var (
 	// AxolotlMode declares the "axolotl-mode" feature
 	AxolotlMode Name = "axolotl-mode"
+	Values      Name = "values"
 )
 
 func init() {
@@ -223,6 +224,15 @@ func init() {
 				"This fun feature is intended to help with testing feature flags.",
 			Enabled: false,
 			Since:   "v0.60.0",
+			Stage:   Alpha,
+		},
+		{
+			Name: Values,
+			Description: "Enabling \"values\" allows for Helm-like values files to be imported. Values are intended to" +
+				"be used for advanced configuration of Zarf components and offer a simple and more flexible " +
+				"alternative to Variables .",
+			Enabled: false,
+			Since:   "v0.61.0",
 			Stage:   Alpha,
 		},
 	}
--- a/src/internal/value/value.go
+++ b/src/internal/value/value.go
@@ -1,0 +1,57 @@
+package value
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/goccy/go-yaml"
+	"github.com/zarf-dev/zarf/src/internal/feature"
+)
+
+type Values map[string]any
+
+// TODO: Implement, also we're eventually going to have to schema check here.
+func ParseFiles(vfs []string) (_ Values, err error) {
+	m := make(map[string]any)
+
+	// No files given
+	if len(vfs) <= 0 {
+		return m, nil
+	}
+
+	// Ensure feature.Values is enabled
+	if !feature.IsEnabled(feature.Values) {
+		return nil, fmt.Errorf("-f or --values provided but \"%s\" feature is not enabled. Run again with --features=\"%s=true\"", feature.Values, feature.Values)
+	}
+
+	// Ensure files exist
+	err = validateFiles(vfs)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, vf := range vfs {
+		f, err := os.Open(vf)
+		defer func(f *os.File) {
+			err := f.Close()
+			if err != nil {
+				err = fmt.Errorf("failed to close values file %s: %w", vf, err)
+			}
+		}(f)
+		err = yaml.NewDecoder(f).Decode(m)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return m, nil
+}
+
+// REVIEW: Do we care about empty files? Here? Small UX tradeoff whether or not to fail on empty files
+func validateFiles(vfs []string) error {
+	for _, vf := range vfs {
+		if _, err := os.Stat(vf); os.IsNotExist(err) {
+			return fmt.Errorf("values file %s does not exist", vf)
+		}
+	}
+	return nil
+}
--- a/src/pkg/packager/deploy.go
+++ b/src/pkg/packager/deploy.go
@@ -19,10 +19,12 @@ import (
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
 	"github.com/zarf-dev/zarf/src/config"
 	"github.com/zarf-dev/zarf/src/config/lang"
+	"github.com/zarf-dev/zarf/src/internal/feature"
 	"github.com/zarf-dev/zarf/src/internal/healthchecks"
 	"github.com/zarf-dev/zarf/src/internal/packager/helm"
 	"github.com/zarf-dev/zarf/src/internal/packager/images"
 	"github.com/zarf-dev/zarf/src/internal/packager/template"
+	"github.com/zarf-dev/zarf/src/internal/value"
 	"github.com/zarf-dev/zarf/src/pkg/cluster"
 	"github.com/zarf-dev/zarf/src/pkg/logger"
 	"github.com/zarf-dev/zarf/src/pkg/packager/actions"
@@ -43,6 +45,8 @@ import (
 type DeployOptions struct {
 	// Deploy time set variables
 	SetVariables map[string]string
+	// Values are values added at deploy time to Helm charts.
+	value.Values
 	// Whether to adopt any pre-existing K8s resources into the Helm charts managed by Zarf
 	AdoptExistingResources bool
 	// Timeout for Helm operations
@@ -85,6 +89,12 @@ func Deploy(ctx context.Context, pkgLayout *layout.PackageLayout, opts DeployOpt
 	l := logger.From(ctx)
 	l.Info("starting deploy", "package", pkgLayout.Pkg.Metadata.Name)
 	start := time.Now()
+
+	if len(opts.Values) > 0 && !feature.IsEnabled(feature.Values) {
+		return DeployResult{}, fmt.Errorf("values passed in but \"%s\" feature is not enabled. Run again with --features=\"%s=true\"", feature.Values, feature.Values)
+	}
+	// TODO(mkcp): Add values handling
+
 	if opts.NamespaceOverride != "" {
 		if err := OverridePackageNamespace(pkgLayout.Pkg, opts.NamespaceOverride); err != nil {
 			return DeployResult{}, err
--- a/src/cmd/viper.go
+++ b/src/cmd/viper.go
@@ -88,6 +88,7 @@ const (
 	VPkgCreateDifferential       = "package.create.differential"
 	VPkgCreateRegistryOverride   = "package.create.registry_override"
 	VPkgCreateFlavor             = "package.create.flavor"
+	VPkgCreateValues             = "package.create.values"
 
 	// Package deploy config keys
 
@@ -97,6 +98,7 @@ const (
 	VPkgDeployTimeout    = "package.deploy.timeout"
 	VPkgDeployNamespace  = "package.deploy.namespace"
 	VPkgRetries          = "package.deploy.retries"
+	VPkgDeployValues     = "package.deploy.values"
 
 	// Package publish config keys
 
--- a/src/config/lang/english.go
+++ b/src/config/lang/english.go
@@ -270,6 +270,7 @@ $ zarf package mirror-resources <your-package.tar.zst> --repos \
 	CmdPackageCreateFlagDifferential          = "[beta] Build a package that only contains the differential changes from local resources and differing remote resources from the specified previously built package"
 	CmdPackageCreateFlagRegistryOverride      = "Specify a map of domains to override on package create when pulling images (e.g. --registry-override docker.io=dockerio-reg.enterprise.intranet)"
 	CmdPackageCreateFlagFlavor                = "The flavor of components to include in the resulting package (i.e. have a matching or empty \"only.flavor\" key)"
+	CmdPackageCreateFlagValuesFiles           = "[alpha] Values files to use for templating and Helm overrides. Multiple files can be passed in as a comma separated list, and the flag can be provided multiple times."
 	CmdPackageCreateCleanPathErr              = "Invalid characters in Zarf cache path, defaulting to %s"
 
 	CmdPackageDeployFlagConfirm                        = "Confirms package deployment without prompting. ONLY use with packages you trust. Skips prompts to review SBOM, configure variables, select optional components and review potential breaking changes."
@@ -282,6 +283,7 @@ $ zarf package mirror-resources <your-package.tar.zst> --repos \
 	CmdPackageDeployValidateLastNonBreakingVersionWarn = "The version of this Zarf binary '%s' is less than the LastNonBreakingVersion of '%s'. You may need to upgrade your Zarf version to at least '%s' to deploy this package"
 	CmdPackageDeployInvalidCLIVersionWarn              = "CLIVersion is set to '%s' which can cause issues with package creation and deployment. To avoid such issues, please set the value to the valid semantic version for this version of Zarf."
 	CmdPackageDeployFlagNamespace                      = "[Alpha] Override the namespace for package deployment. Requires the package to have only one distinct namespace defined."
+	CmdPackageDeployFlagValuesFiles                    = CmdPackageCreateFlagValuesFiles
 
 	CmdPackageMirrorFlagComponents = "Comma-separated list of components to mirror.  This list will be respected regardless of a component's 'required' or 'default' status.  Globbing component names with '*' and deselecting components with a leading '-' are also supported."
 	CmdPackageMirrorFlagNoChecksum = "Turns off the addition of a checksum to image tags (as would be used by the Zarf Agent) while mirroring images."
--- a/src/internal/value/value_test.go
+++ b/src/internal/value/value_test.go
@@ -1,0 +1,1 @@
+package value
--- a/src/pkg/packager/create.go
+++ b/src/pkg/packager/create.go
@@ -11,6 +11,8 @@ import (
 
 	"github.com/defenseunicorns/pkg/helpers/v2"
 	"github.com/defenseunicorns/pkg/oci"
+	"github.com/zarf-dev/zarf/src/internal/feature"
+	"github.com/zarf-dev/zarf/src/internal/value"
 
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
 	"github.com/zarf-dev/zarf/src/pkg/logger"
@@ -21,10 +23,11 @@ import (
 
 // CreateOptions are the optional parameters to create
 type CreateOptions struct {
-	Flavor                  string
-	RegistryOverrides       map[string]string
-	SigningKeyPath          string
-	SigningKeyPassword      string
+	Flavor             string
+	RegistryOverrides  map[string]string
+	SigningKeyPath     string
+	SigningKeyPassword string
+	// TODO(mkcp): Give SetVariables a type
 	SetVariables            map[string]string
 	MaxPackageSizeMB        int
 	SBOMOut                 string
@@ -34,10 +37,16 @@ type CreateOptions struct {
 	CachePath               string
 	// applicable when output is an OCI registry
 	RemoteOptions
+	value.Values
 }
 
 // Create takes a path to a directory containing a ZarfPackageConfig and returns the path to the created package
 func Create(ctx context.Context, packagePath string, output string, opts CreateOptions) (_ string, err error) {
+	if len(opts.Values) > 0 && !feature.IsEnabled(feature.Values) {
+		return "", fmt.Errorf("values passed in but \"%s\" feature is not enabled. Run again with --features=\"%s=true\"", feature.Values, feature.Values)
+	}
+	// TODO(mkcp): Add values handling
+
 	if opts.SkipSBOM && opts.SBOMOut != "" {
 		return "", fmt.Errorf("cannot skip SBOM creation and specify an SBOM output directory")
 	}
--- a/src/pkg/packager/inspect.go
+++ b/src/pkg/packager/inspect.go
@@ -13,8 +13,10 @@ import (
 	"github.com/defenseunicorns/pkg/helpers/v2"
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
 	"github.com/zarf-dev/zarf/src/config"
+	"github.com/zarf-dev/zarf/src/internal/feature"
 	"github.com/zarf-dev/zarf/src/internal/packager/helm"
 	"github.com/zarf-dev/zarf/src/internal/packager/template"
+	"github.com/zarf-dev/zarf/src/internal/value"
 	"github.com/zarf-dev/zarf/src/pkg/packager/layout"
 	"github.com/zarf-dev/zarf/src/pkg/packager/load"
 	"github.com/zarf-dev/zarf/src/pkg/state"
@@ -44,10 +46,16 @@ type Resource struct {
 type InspectPackageResourcesOptions struct {
 	SetVariables map[string]string
 	KubeVersion  string
+	value.Values
 }
 
 // InspectPackageResources templates and returns the manifests, charts, and values files in the package as they would be on deploy
 func InspectPackageResources(ctx context.Context, pkgLayout *layout.PackageLayout, opts InspectPackageResourcesOptions) (_ []Resource, err error) {
+	if len(opts.Values) > 0 && !feature.IsEnabled(feature.Values) {
+		return []Resource{}, fmt.Errorf("values passed in but \"%s\" feature is not enabled. Run again with --features=\"%s=true\"", feature.Values, feature.Values)
+	}
+	// TODO(mkcp): Add values handling
+
 	s, err := state.Default()
 	if err != nil {
 		return nil, err


mkcp:
Hunting down the root cause for the regression in `TestVariables` - that behavior should not have changed with this PR. After that this should be good for review.

src/cmd/package.go
```diff
@@ -237,9 +242,18 @@ func (o *packageDeployOptions) run(cmd *cobra.Command, args []string) (err error
 	}
 
 	v := getViper()
+
+	// Merge SetVariables and config variables.
 	pkgConfig.PkgOpts.SetVariables = helpers.TransformAndMergeMap(
 		v.GetStringMapString(VPkgDeploySet), pkgConfig.PkgOpts.SetVariables, strings.ToUpper)
 
+	// Load files supplied by --values / -v or a user's zarf-config.{yaml,toml}
+	// REVIEW: Should we also load valuesFiles supplied via URL on the CLI?
```
mkcp, line 298: There's a few places we could load remote valuesfiles and I would like to have some clear idea on how, or at least a few options to follow, set up in a follow-up issue before this is merged.

src/cmd/package.go
```diff
@@ -663,8 +680,17 @@ func (o *packageInspectValuesFilesOptions) run(ctx context.Context, args []strin
 		return err
 	}
 	v := getViper()
+
+	// Merge SetVariables and config variables.
 	o.setVariables = helpers.TransformAndMergeMap(v.GetStringMapString(VPkgDeploySet), o.setVariables, strings.ToUpper)
 
+	// Load files supplied by --values / -v
+	// REVIEW: Should we also load valuesFiles supplied via URL on the CLI?
```
mkcp: Same "When and how to process remote valuesfiles" problem as above.

src/internal/feature/feature.go
```diff
@@ -225,6 +226,15 @@ func init() {
 			Since:   "v0.60.0",
 			Stage:   Alpha,
 		},
+		{
+			Name: Values,
+			Description: "Enabling \"values\" allows for Helm-like values files to be imported. Values are intended to" +
+				"be used for advanced configuration of Zarf components and offer a simple and more flexible " +
+				"alternative to Variables .",
+			Enabled: false,
+			Since:   "v0.61.0",
```
mkcp: TODO: Before merging, update to upcoming minor release

src/internal/value/value.go
```diff
@@ -0,0 +1,254 @@
...
+func (p Path) Validate() error {
+	if p == "" || !strings.HasPrefix(string(p), ".") {
+		return fmt.Errorf("invalid path format: %s", p)
+	}
+	return nil
+}
+
+// ParseFilesOptions provides optional configuration for ParseFiles
+type ParseFilesOptions struct {
+	// TODO(mkcp): Add schema check. Maybe here in parsing, or later in the process like templating?
```
mkcp, line 42: Schema checking is an important part of the ZEP and should be a prerequisite for beta. The scope of this PR is already pretty broad so I'd prefer to solve this in a follow-up. The key does already exist on the packagedefinition though.

src/pkg/packager/actions/actions.go
```diff
@@ -14,36 +15,43 @@ import (
 	"time"
 
 	"github.com/defenseunicorns/pkg/helpers/v2"
+	"github.com/goccy/go-yaml"
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
-	"github.com/zarf-dev/zarf/src/internal/packager/template"
+	ptmpl "github.com/zarf-dev/zarf/src/internal/packager/template"
```
mkcp, line 20: This is _not_ in scope for this PR but we do we have a few adjacent concepts of "templating" in Zarf's codebase. While we work on sunsetting variables I think it would be good to see if consolidating packager and variable templating into the `internal/template` package is viable.

mkcp:
Marking this ready to review because the code should be ready to go, pending any review changes. I added a lightweight e2e test yesterday that caught a package-relative path bug so running the tests on that fix.

src/cmd/package.go
AustinAbro321: I think we shouldn't give users the option to load values from URLs. It's not a request we've received from users for variables, and I think the use case is thinner since deploy is often done in the airgap.

In connected environments I think there's also some risk to deploy time injection. If a remote url was taken over and malicious values were added for example.

src/internal/value/value.go
```diff
@@ -0,0 +1,254 @@
...
+	start := time.Now()
+	defer func() {
+		logger.From(ctx).Debug("values parsing complete",
+			"duration", time.Since(start),
+			"files", len(paths))
+	}()
+
+	if ctx == nil {
+		return Values{}, errors.New("context cannot be nil")
+	}
```
AustinAbro321: nit: While there's no harm in checking, I think the go community has generally standardized around not checking if ctx == nil. For example, Helm, Flux, or Kubernetes generally don't do input validation on ctx

There's also this blurb in https://pkg.go.dev/context

> Do not pass a nil [Context](https://pkg.go.dev/context#Context), even if a function permits it. Pass [context.TODO](https://pkg.go.dev/context#TODO) if you are unsure about which Context to use.

src/cmd/package.go
mkcp: I definitely see the advantages of cutting remote values files from this iteration (likely through GA), especially since it's straightforward to revisit later with more exploratory work.

@brandtkeller do you have any thoughts on whether we should cut values from urls?

brandtkeller: Ultimately I think we've had a few great discussions on decomposing and de-scoping ZEP requirements as an area of active improvement. 

Given that and initial stress testing - my opinion would be to opt toward not allowing remote values file retrieval in the initial alpha release. This will provide us the opportunity to examine _what_ people really want from remote storage of these configuration artifacts _outside_ of the package. I don't know if we have much evidence currently to support a stance.

mkcp: Yeah, sounds good. It's something we can revisit in beta requirements, though I do think if we do end up including it post-GA is probably a more reasonable timeframe. The user behavior, implementation, and overall security posture/safeguards need to be carefully considered. Trimming out all of the remote values files references in this PR and making note of it in the parent issue.

src/cmd/package.go
mkcp: Remote values files not in scope for values alpha

examples/values-helm-overrides/README.md
AustinAbro321, line 1: Something we do for the docs site is have an x-mdx key that essentially works as the readme in the docs site, instead of having readmes alongside examples. Take a look at [dos-games](https://github.com/zarf-dev/zarf/blob/main/examples/dos-games/zarf.yaml#L21) example. Open to changing that in the future

examples/values-templating/zarf.yaml
AustinAbro321, line 1: Not blocking, but I'd personally prefer to see these combined into one example. I think having two components, one for charts and one for manifests, both utilizing values paints a nice story. If we go that route I think we can delete the TF component. I'd like to delete TF references from Zarf entirely, early on the project encouraged people to use terraform with Zarf actions, and that is a pattern I'd like us to stop promoting

src/internal/template/template.go
```diff
@@ -0,0 +1,149 @@
...
+
+// WithPackage takes a v1alpha1.ZarfPackage and makes Metadata, Constants, and Build available on the Objects map.
+func (o Objects) WithPackage(pkg v1alpha1.ZarfPackage) Objects {
+	// Check for fields that should be set on pkg to see if the sub-obj is available
+	if pkg.Metadata.Name != "" {
+		o.WithMetadata(pkg.Metadata)
+	}
+	if pkg.Build.User != "" {
+		o.WithBuild(pkg.Build)
+	}
```
AustinAbro321, line 95: Not necessarily against adding these, but it does feel a bit like a bit of scope creep. I'm not sure I can think of a use case for any of these fields other than .build.architecture and .build.flavor.  Curious if you had particular use cases in mind.

src/internal/template/template.go
```diff
@@ -0,0 +1,149 @@
...
+	// Are any Constants set? Load them
+	if len(pkg.Constants) > 0 {
+		o.WithConstants(pkg.Constants)
+	}
+	return o
+}
+
+// ApplyToCmd takes a string cmd and fills in any templates.
+func ApplyToCmd(ctx context.Context, cmd string, objs Objects) (string, error) {
```
AustinAbro321: nit: thoughts on naming this something like `ApplyToString`? I don't think there's anything in this function that makes it specific to commands

src/internal/value/value.go
```diff
@@ -0,0 +1,254 @@
...
+			ok, err := helpers.IsTextFile(path)
+			if err != nil {
+				return nil, err
+			}
+			if ok {
+				// Ensure file exists
+				// REVIEW: Do we actually care about empty files here? Small UX tradeoff whether or not to fail on empty files
+				if _, err := os.Stat(path); errors.Is(err, os.ErrNotExist) {
+					return nil, err
+				}
```
AustinAbro321: It looks like helpers.IsTextFile will already error if the file doesn't exist. I'm okay with succeeding on empty files, but I do agree that erroring on an empty file is the right move

src/internal/value/value.go
```diff
@@ -0,0 +1,254 @@
...
+// Values provides a map of keys to values for use in templating and Helm overrides.
+type Values map[string]any
+
+// Path starts with a . and represents a specific key in a nested hierarchy of keys. For example, .resources.limits.cpu
+// resolves the value for "cpu" within the keyspace of Values.
+type Path string
+
+// Validate inspects the string stored at Path and ensures it's valid.
+func (p Path) Validate() error {
+	if p == "" || !strings.HasPrefix(string(p), ".") {
```
AustinAbro321, line 34: Agree with having validation here, I think we should also add validation at create time, this way the user gets faster feedback / doesn't ship a package that won't deploy.  Here's an example of how we build in regex validation  elsewhere in the schema https://github.com/zarf-dev/zarf/blob/main/src/api/v1alpha1/component.go#L10. 

Also fine with this being a follow up issue

src/test/packages/42-values/zarf.yaml
AustinAbro321, line 1: IMO it would be good to also include `setValues` with actions in the e2e test here alongside a small chart that deploys another configmap.

src/internal/template/template.go
```diff
@@ -0,0 +1,149 @@
...
+	return o.WithValues(values)
+}
+
+// WithValues takes a value.Values and makes it available in templating Objects.
+func (o Objects) WithValues(values value.Values) Objects {
+	o[objectKeyValues] = values
+	return o
+}
+
+// WithMetadata takes the v1alpha1.ZarfMetadata section of a created package and makes it available in templating Objects.
```
mkcp, line 53: REVEIW: Thinking thru this more, I'm starting to wonder if in a followup we should nest the Metadata and Build keys under `.Package`, like `{{ .Package.Metadata }}`. Also, same for {{ .Component.x }} fields like name and namespace.

a1994sc: If things under `.Package` are exposed it would help things like my `architecture` related PR/issue

src/internal/value/value.go
mkcp: Good call, I'll add this to the beta release reqs

src/internal/template/template.go
mkcp: Yeah absolutely - I had the same thought when working through the component.file.target problem. I'll revise in this PR.

src/internal/template/template.go
mkcp: Really just nice to haves and preempting requests for more template objects from users in the future. Helm gives users access to .Release metadata, like .Release.name - it isn't the exact same thing as packagedefinition fields, but is a similar idea.

src/internal/template/template.go
mkcp: That's great to hear, I'm leaning towards keeping these in though the objects' key structure may change before the beta version!

src/internal/value/value.go
mkcp: Yeah that's fair, historically I've worked on codebases where "never panic" takes priority, but I'm open to either coding standard so long as it's consistently applied. It's well known that nil contexts aren't valid.

examples/values-templating/zarf.yaml
mkcp: Yeah I can see the advantages of this - especially removing the TF component. Early on I copied over the Variables example as an end-to-end proof of concept. I do think that this has outgrown the need for it though. I'll resolve the blocking feedback first then consider if this should be resolved now or in a follow-up revision.

src/internal/template/template.go
a1994sc: yea, it doesn't need to be in this PR but would be great down the line!

src/test/e2e/42_values_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestValues(t *testing.T) {
+	t.Log("E2E: Values")
+
+	// Don't run this test in appliance mode
+	if e2e.ApplianceMode {
+		return
+	}
```
AustinAbro321: We still want to run this test in appliance mode. Appliance mode is when we are using the built in k3s cluster

AustinAbro321:
One thing that will be import in regards to values is how they are handled in component imports. If we take a look at how variables are handled in this [component-imports test](https://github.com/zarf-dev/zarf/tree/main/src/pkg/packager/load/testdata/import/variables), we see that the value from the parents take priority over child values. I think it would make sense to keep the same strategy. 

It's fine to do in a separate PR, but I think it should be done before the feature flag is enabled by default. I can see users getting very confused if their values are silently not being forwarded to the next component.

src/internal/value/value.go
AustinAbro321: Yeah ctx is definitely exception to the rule around that. And yeah mostly just saying this cause that's how it's done everywhere else in the codebase so it'd be inconsistent

src/api/v1alpha1/component.go
```diff
@@ -172,6 +178,12 @@ type ZarfChartVariable struct {
 	Path string `json:"path"`
 }
 
+// ZarfChartValue maps a Zarf Value key to a Helm Value.
+type ZarfChartValue struct {
+	SourcePath string `json:"sourcePath,omitempty"`
+	TargetPath string `json:"targetPath,omitempty"`
```
AustinAbro321: omitempty makes fields optional, but both of these fields are required
```suggestion
	SourcePath string `json:"sourcePath"`
	TargetPath string `json:"targetPath"`
```

src/api/v1alpha1/component.go
```diff
@@ -124,6 +124,10 @@ type ZarfFile struct {
 	Symlinks []string `json:"symlinks,omitempty"`
 	// Local folder or file to be extracted from a 'source' archive.
 	ExtractPath string `json:"extractPath,omitempty"`
+	// [alpha]
+	// Template enables go-templates inside manifests. This is useful for parameterizing fields that the value will be
+	// known at deploy-time. See documentation for Zarf Values for how to set these values.
+	Template bool `json:"template,omitempty"`
```
AustinAbro321: One strategy I've liked when using a boolean is having a boolptr and a function like `isTemplate` or `shouldTemplate` here and for manifests. This way instead of the default always be false, we can control the default. For example, with manifests in the next schema version, they will be gotemplatable by default. 

We do something similar with the required field and [isRequired](https://github.com/zarf-dev/zarf/blob/07c7c7cc0dd242e367b3aa7df91d7c67863143d2/src/api/v1alpha1/component.go#L87). This allows the object to own what the default of the value should be when a bool is used

src/pkg/packager/actions/actions.go
```diff
@@ -114,6 +128,38 @@ retryCmd:
 				}
 			}
 
+			// If an output value is defined, parse the result and set it to values map.
+			for _, v := range action.SetValues {
```
AustinAbro321, line 131: It's pretty hard to unit test actions since they shell out. It might be good to e2e test this part

src/pkg/packager/layout/assemble.go
```diff
@@ -875,3 +881,53 @@ func createReproducibleTarballFromDir(dirPath, dirPrefix, tarballPath string, ov
...
+	src := file
+	if !filepath.IsAbs(src) {
+		src = filepath.Join(packagePath, file)
+	}
+	// Validate src
+	if _, err := os.Stat(src); err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("values file %s not found", src)
+		}
+		return fmt.Errorf("unable to access values file %s: %w", src, err)
```
AustinAbro321: nit, the error message we receive back from os.Stat should already be pretty clear, I don't think we should separate them

src/pkg/packager/packager2.go
```diff
@@ -29,28 +34,68 @@ func getPopulatedVariableConfig(ctx context.Context, pkg v1alpha1.ZarfPackage, s
...
+				"path", chartValue.SourcePath,
+				"error", err,
+				"component", componentName,
+				"chart", chart.Name,
+			)
+			continue
+		}
+
+		// Strip off the leading dot from the target path.
+		targetPath := chartValue.TargetPath[1:]
```
AustinAbro321: A single dot should be valid for both `sourcePath` and `targetPath`. Would be good to get a test for this behavior. Ref: https://github.com/zarf-dev/proposals/blob/main/0021-zarf-values/README.md#proposal

AustinAbro321 (changes requested):
I like the architecture and unit tests. A couple more notes on behavior and tests

src/test/packages/42-values/zarf.yaml
mkcp: Good call, fixed in latest

src/internal/template/template.go
mkcp: Fixed in latest - ended up going with Apply because the get string then return string case feels like a reasonable default. Also needed to update the comments here from a previous shape of the API (before going with template.Objects it was templateContext which was worse for several reasons).

src/pkg/packager/packager2.go
mkcp: That's a good catch, missed that behavior from the proposal while implementing the nested keys part. I have good sense for how to implement it, but i'll scope this out for a follow-up for the beta target cause it's pretty easily added and tested in isolation.

src/pkg/packager/layout/assemble.go
mkcp: revised in latest

src/pkg/packager/packager2.go
AustinAbro321: IMO it's worth adding in this PR. I see it as a bug in the behavior.

mkcp: That's fair - as I explored a solution to this more, this ended up being one of those problems that once you solve it a bunch of other things click into place.

I made sure both extract and set operations on Values support root paths. This ended up enabling some reuse and simplification in several places Values are used - both helm overrides and SetValues parsing with `values.Set` and the overall merge functionality. I also went over some places where unit tests were missing, so the behavior should be covered on a few system layers.

src/internal/value/value.go
mkcp: revised in latest

src/pkg/packager/actions/actions.go
mkcp: This is covered in e2e now

examples/values-helm-overrides/README.md
mkcp: That's a good callout - it wasn't immediately clear to me what those keys were for. Overall I'd rather we have readmes (and generate the site docs from there). It's a well-known convention and self-documenting for users (and maintainers, turns out!). I'll condense the readme into the x-mdx keys for now though.

src/api/v1alpha1/component.go
mkcp: This tracks - updated in latest and defaulting to false

brandtkeller:
Appreciative of the dissection of future work and maturation. If it is not already - we should capture it in ZEP/ZEP issue. 

Particularly the `package inspect` `--values` support and how powerful that will be for package development.

src/pkg/packager/layout/assemble.go
```diff
@@ -882,3 +881,31 @@ func createReproducibleTarballFromDir(dirPath, dirPrefix, tarballPath string, ov
...
+	if !filepath.IsAbs(src) {
+		src = filepath.Join(packagePath, file)
+	}
+	// Validate src
+	if _, err := os.Stat(src); err != nil {
+		return fmt.Errorf("unable to access values file %s: %w", src, err)
+	}
+
+	// Copy file to pre-archive package
+	dst := filepath.Join(buildPath, file)
```
brandtkeller: Observation and a question:

If a relative path is included - the file could be copied outside of the target temporary directory:

Example inputs:
```
kind: ZarfPackageConfig
metadata:
  name: values-templating
  description: Example nginx package to demonstrate Zarf Values templating
values:
  files:
    - values.yaml
    - ../helm-charts/values.yaml
```

```
2025-10-03 19:02:19 DBG copying values file src=examples/values-templating/values.yaml dst=/tmp/zarf-1589920477/values.yaml
2025-10-03 19:02:19 DBG copying values file src=examples/helm-charts/values.yaml dst=/tmp/helm-charts/values.yaml
```

It may be that this should be `src` instead of `file` - given how the logic responds to relative paths (See below)

```suggestion
	dst := filepath.Join(buildPath, src)
```

Although this would lead to the file being copied into the package based on the path - this feels undesirable. I wonder if we want to control _where_ these values are stored within the package. Using the recommendation above could lead to potential file overwrites unintentionally that i believe we want to control. 

Much like how we handle directories for specific purposes (IE sboms) we may do similarly for values?

mkcp:
@brandtkeller that's a great idea, and I expect that all of that exploration will end up fueling the documentation (workflows, migration guides etc.) that we publish on our website. I still need to dig into the inspect part myself but it's on the follow-ups for beta.

brandtkeller (changes requested):
Dropping one observation as I continue review

src/pkg/packager/layout/assemble.go
mkcp: Ya that relative path behavior is definitely a bug - we don't guard against that in this branch. This was originally src, and the file copying behavior there did end up causing issues with files not being found in CI. Using the filename fixed it as you can see in [this commit](https://github.com/zarf-dev/zarf/pull/4136/commits/8669d72c2f2ac058082c34e0c876cece0a4ea3f7).

How about we call filepath.Clean(file) on the file to ensure we're not jumping out of the intended destination, and then potentially let subdirectories like `/values` emerge by convention?

brandtkeller: that sounds reasonable!

examples/values-templating/zarf.yaml
mkcp: I opted to merge the examples together in this PR, same with removing the TF component.

src/pkg/packager/layout/assemble.go
mkcp: fixed in latest revision

brandtkeller: I am still seeing the same behavior:
```
2025-10-03 20:08:43 DBG copying values file src=examples/values-templating/values.yaml dst=/tmp/zarf-1418523624/values.yaml
2025-10-03 20:08:47 DBG copying values file src=examples/helm-charts/values.yaml dst=/tmp/helm-charts/values.yaml
```

mkcp: Ah! ok I see what's missing here, let me make sure the destination always includes the tmp-package root

mkcp: It was simpler than that - Clean just doesn't get rid of `..` at the start of a filepath. I added a guard clause against that so we're enforcing that values files are sourced from within a packagedefinition (arbitrary nesting of dirs is fine). Then, any `..`'s within the path will be trimmed by the clean call.

brandtkeller: This sound fair - optimally it's a big edgecase but otherwise it's a feedback mechanism to understand how users organize their projects with regards to `zarf.yaml` locality.

src/api/v1alpha1/package.go
```diff
@@ -247,3 +272,11 @@ type ZarfBuildData struct {
 	// The flavor of Zarf used to build this package.
 	Flavor string `json:"flavor,omitempty"`
 }
+
+// ZarfValues provides top-level imports for
```
brandtkeller: Incomplete/TODO doccomments

src/internal/template/template.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	return o
+}
+
+// Apply takes a string, fills in the templates with the given Objects, and returns a new string.
+func Apply(ctx context.Context, s string, objs Objects) (string, error) {
+	l := logger.From(ctx)
+	l.Debug("applying templates", "str", s)
+
+	tmpl, err := ttmpl.New("cmd").Funcs(sprig.TxtFuncMap()).Parse(s)
```
brandtkeller: I believe it would be beneficial to explicitly set the option for how we handle missing keys. 

```
tmpl.Option("missingkey=default")
```

This does raise the question for _how_ we should be handling missing keys. Testing currently defaults to the above implicitly (zero value).  Do we believe this is the best path or do we want more fail-fast scenarios?

mkcp: After thinking this through a bit, I feel that the better overall experience here would be erroring out to users if a key they're trying to template in doesn't exist. It provides quicker feedback against typos, and encourages users to set defaults with sprig, e.g. `{{ .Values.fooNotExistIsFine | default ""  }}`.

There was actually a moment in when testing e2e in CI where this would've saved me about 30-40mins, rather than having to run and then rerun to ensure the value I was looking for was set correctly.

mkcp: Revised in latest

src/api/v1alpha1/package.go
mkcp: fixed in latest

src/pkg/packager/layout/assemble.go
```diff
@@ -882,3 +881,36 @@ func createReproducibleTarballFromDir(dirPath, dirPrefix, tarballPath string, ov
 		return nil
 	})
 }
+
+func copyValuesFile(ctx context.Context, file, packagePath, buildPath string) error {
```
brandtkeller, line 891: I think there is a small but relevant risk with this logic. When we don't control the destination (currently arbitrarily placed based on filepath) we introduce the risk that some operations could corrupt the package. 

IE the `components` directory in a package should contain specific mappings from the manifest. Yet if I include:

```yaml
kind: ZarfPackageConfig
metadata:
  name: values-templating
  description: Example nginx package to demonstrate Zarf Values templating
values:
  files:
    - values.yaml
    - components/values-with-helm-chart/other-values.yaml
```

I could write additional information to that directory. It's a small risk - but the structure of a package is something that I would want to create guardrails around.

src/internal/value/value.go
```diff
@@ -0,0 +1,272 @@
...
+// - Do we take a json or byte array, a map[string]any, or a specific json.schema type?
+// - Do we want to return a list of errors, some specific schema fail datatype, or some other type?
+// - Surely there's libraries for this which have their own opinionated inputs for the schema and return types
+// func checkSchemaStub(_ Values, _ string) []error {
+// 	return nil
+// }
+
+// DeepMerge merges one or more Values maps recursively into the receiver via mutation.
+// Later maps in the variadic arguments take precedence over earlier ones.
+func (v Values) DeepMerge(sources ...Values) {
```
brandtkeller, line 129: ++ this replicates the Helm behaviors well to my knowledge. (IE slice replacement behaviors)

src/pkg/packager/load/load.go
```diff
@@ -53,6 +58,24 @@ func PackageDefinition(ctx context.Context, packagePath string, opts DefinitionO
...
+	// Resolve values file paths relative to the package path
+	valueFilePaths := make([]string, len(pkg.Values.Files))
+	for i, vf := range pkg.Values.Files {
+		valueFilePaths[i] = filepath.Join(packagePath, vf)
+	}
+	values, err := value.ParseFiles(ctx, valueFilePaths, value.ParseFilesOptions{})
+	if err != nil {
+		return v1alpha1.ZarfPackage{}, fmt.Errorf("failed to parse values files: %w", err)
+	}
+	opts.Values.DeepMerge(values)
```
brandtkeller: This execution of `DeepMerge` merges the package manifests values files onto the (assuming) provided command line values. 

`deploy.go` does the opposite and is what I assumed the correct stance. 

Is this intentional?

src/pkg/packager/deploy.go
```diff
@@ -109,8 +118,28 @@ func Deploy(ctx context.Context, pkgLayout *layout.PackageLayout, opts DeployOpt
...
+	for i, vf := range pkgLayout.Pkg.Values.Files {
+		valueFilePaths[i] = filepath.Join(pkgLayout.DirPath(), vf)
+	}
+	vals, err := value.ParseFiles(ctx, valueFilePaths, value.ParseFilesOptions{})
+	if err != nil {
+		return DeployResult{}, err
+	}
+	// Package defaults are overridden by deploy values.
+	vals.DeepMerge(opts.Values)
+	logger.From(ctx).Debug("package values", "values", vals)
```
brandtkeller: nit: we have `l` for use:

```suggestion
	l.Debug("package values", "values", vals)
```

src/pkg/packager/deploy.go
```diff
@@ -501,12 +532,15 @@ func (d *deployer) installCharts(ctx context.Context, pkgLayout *layout.PackageL
...
-		valuesOverrides, err := generateValuesOverrides(chart, component.Name, d.vc, opts.ValuesOverridesMap)
+		valuesOverrides, err := generateValuesOverrides(ctx, chart, component.Name, overrideOpts{
+			variableConfig:     d.vc,
+			values:             d.vals,
+			valuesOverridesMap: opts.ValuesOverridesMap,
+		})
 		if err != nil {
 			return installedCharts, err
 		}
+		logger.From(ctx).Debug("overrides generated", "values", valuesOverrides, "count", len(valuesOverrides))
```
brandtkeller:
```suggestion
		l.Debug("overrides generated", "values", valuesOverrides, "count", len(valuesOverrides))
```

src/pkg/packager/deploy.go
```diff
@@ -523,6 +557,12 @@ func (d *deployer) installCharts(ctx context.Context, pkgLayout *layout.PackageL
 		if err != nil {
 			return installedCharts, fmt.Errorf("failed to load chart data: %w", err)
 		}
+		logger.From(ctx).Debug("loaded chart",
```
brandtkeller:
```suggestion
		l.Debug("loaded chart",
```

src/internal/template/template.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	b := &bytes.Buffer{}
+	if err = tmpl.Execute(b, objs); err != nil {
+		return "", err
+	}
+	return b.String(), nil
+}
+
+// ApplyToFile load a file path at src, fills in the templates with the given Objects, then writes the file to dst.
+func ApplyToFile(ctx context.Context, src, dst string, objs Objects) error {
```
brandtkeller, line 123: Later use of this function reveals intentional overwrite of files. 

Is there any risk that we overwrite a file that is potentially used again with other values and can no longer be used as expected? Should we handle this differently?

src/pkg/packager/load/load.go
mkcp: Good catch on this, that merge precedence did get flipped but it turns out this is a vestige from earlier experiments w/ go-templating in the packagedefinition and why this didn't get caught in any of the tests. Removed this in latest

src/pkg/packager/deploy.go
mkcp: Revised each of these in latest

src/internal/template/template.go
mkcp: I do think that is more of a usage concern rather than something inherit in the API. For the overwriting semantics, ideally we'd be able to write the file once, but with setValues we need to wait to fill templates in until prior components are done executing. Whether we write the templated version in place or conditionally copy the pre-templated files to another location then write them again it's the same number of writes. 

As for using the file again - we currently don't have a need to apply templates in multiple steps for files, and doing so would likely require filling in some fields but not others. That would be a different API call entirely so we could split the file into chunks. The main example I could think of is progressive templating a package definition (e.g. metadata first, then other sections in a defined order). To solve that I'd be more inclined to apply each one at a time with the string Apply, or (probably more maintainable) have a new intermediate type for a partially templated package that we process with its own Apply function.

src/pkg/packager/layout/assemble.go
mkcp: I think that's a pretty fair suggestion - do you know if we have a list of protected filenames and directories to go off of here? This seems like a pattern that's likely to repeat in other parts of the system, and I'm inclined to have a generic list and validation fn for it in package.

AustinAbro321:
One thing that users will expect are the [Helm builtin function](https://github.com/helm/helm/blob/600eb487f11c002a131e88836cfb41b10dff99ac/pkg/engine/funcs.go#L54). They seem to be private methods, but are probably simple enough to copy over in most cases. Some of these functions especially the `toYaml` method are used super often in charts so that variables can be expanded upon. Otherwise maps aren't very usable (though I could be missing something). For example, the Zarf registry chart uses `toYaml` for `resources` which I believe is a common pattern -> https://github.com/zarf-dev/zarf/blob/7525abe99b4dbb6e829779278a9a7c6f77f99bab/examples/helm-charts/chart/templates/deployment.yaml#L185. Also used in the famous [podinfo example](https://github.com/stefanprodan/podinfo/blob/master/charts/podinfo/templates/deployment.yaml#L198)

I think it'd be reasonable to do this in a follow up PR, but it should be implemented before we make values enabled by default imo.

mkcp:
@AustinAbro321 Yeah, I think that's a pretty reasonable thing to add in a follow-up PR. I believe the sprig text funcs are set up to merge cleanly with additional funcmap entries.

src/pkg/packager/layout/assemble.go
brandtkeller: [layout](http://github.com/zarf-dev/zarf/blob/main/src/pkg/packager/layout/layout.go) is where my mind reaches for first - but primarily standardizes the naming as opposed to protecting them.

src/internal/template/template.go
brandtkeller: I believe that is fair. My thought was primarily around the operation and whether we should consider writing the templated file to a temporary location such that the artifact as it exists pre-template is available for re-use (niche problem to be fair). 

Described another way - When we decompress the tarball for deployment - what is our standard around modifying the items in the temporary directory? Given that cleanup will occur and the source of truth is still maintained in the tarball - do we modify them in place or apply handlers around processing the artifact without modification (tmpdir for example).

src/pkg/packager/layout/assemble.go
mkcp: Returning to this - having some generic guards in place seems useful not just for this but other places we copy files as well. I added checking functions in `layout.go` for reserved filenames as well as reserved directories like `components`. The directory check may block some amount of user expression in their package definitions, but it's something we can iterate on. The filename check should be straightforward though.

AustinAbro321: Rather than maintaining a list of files we can't collide with, what are your thoughts on instead having a top level directory called `values` where we place all create time values files. This would eliminate the chance the collisions.

src/pkg/packager/deploy.go
```diff
@@ -73,20 +79,23 @@ type deployer struct {
...
 	vc          *variables.VariableConfig
+	vals        value.Values
 	hpaModified bool
 }
 
 // DeployResult is the result of a successful deploy
 type DeployResult struct {
 	DeployedComponents []state.DeployedComponent
 	VariableConfig     *variables.VariableConfig
+	Values             value.Values
```
AustinAbro321, line 90: This is never set

src/pkg/packager/packager.go
```diff
@@ -29,29 +33,70 @@ func getPopulatedVariableConfig(ctx context.Context, pkg v1alpha1.ZarfPackage, s
...
+		}
+		if chartValue.TargetPath[0] != '.' {
+			return nil, fmt.Errorf("targetPath \"%s\" must start with a dot", chartValue.TargetPath)
+		}
+
+		// Extract value from source path in values
+		sourceValue, err := opts.values.Extract(value.Path(chartValue.SourcePath))
+		if err != nil {
+			// Log warning but don't fail - source path might not exist
+			logger.From(ctx).Warn("unable to extract value from path",
```
AustinAbro321: Maybe we should error here now as well? Feels somewhat inconsistent that we we error on templates not found in manifests, but don't error here

src/internal/template/template.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/Masterminds/sprig/v3"
+	"github.com/zarf-dev/zarf/src/api/v1alpha1"
+	"github.com/zarf-dev/zarf/src/internal/value"
+	"github.com/zarf-dev/zarf/src/pkg/logger"
+	"github.com/zarf-dev/zarf/src/pkg/variables"
+)
+
+const missingKeyDefault = "missingkey=error"
+
+var defaultFuncs = sprig.TxtFuncMap()
```
AustinAbro321: Taking a look at how Helm does things, they delete the sprig functions env and expandenv. I think that'd be a good thing to do from a security perspective. 
https://github.com/helm/helm/blob/f19bb9cd4c99943f7a4980d6670de44affe3e472/pkg/engine/funcs.go#L47-L48

src/pkg/packager/actions/actions.go
```diff
@@ -315,3 +336,27 @@ func MatchAllRegex(regex *regexp.Regexp, str string) []func(string) string {
 	}
 	return funcs
 }
+
+// parseAndSetValue parses the output string according to the setValue type and sets it in the values map.
+func parseAndSetValue(output string, setValue v1alpha1.SetValue, values value.Values) error {
```
AustinAbro321, line 341: If you look at where this is called if type is empty the error gets eaten and you get an error like this 
```bash
failed to deploy package: unable to deploy component "test-values-component": unable to run component before action: command "echo \"whatever\"" failed after 0 retries
```
It's a problem of the tryCmd function which could use some refactoring, but I think it would make sense to move the l.Warn from the actionRun call to the err check after the function is called for now

src/pkg/packager/actions/actions.go
```diff
@@ -315,3 +336,27 @@ func MatchAllRegex(regex *regexp.Regexp, str string) []func(string) string {
...
+	case v1alpha1.SetValueJSON:
+		var parsed any
+		if err := json.Unmarshal([]byte(output), &parsed); err != nil {
+			return fmt.Errorf("failed to parse JSON output for setValue %q: %w", setValue.Key, err)
+		}
+		val = parsed
+	case v1alpha1.SetValueString:
+		val = output
+	default:
+		return fmt.Errorf("unknown setValue type: %s", setValue.Type)
```
AustinAbro321: Thoughts on defaulting to string? It's not detailed in the proposal. As a user I think I would generally expect string to be the default unless otherwise specified, but I think both paths are reasonable. 

If we don't default to string then we should remove the `omitEmpty` from `.setValues.Type` since it's a required field in that case.

mkcp: Yeah I'm in favor of this, revised in latest.

src/pkg/packager/deploy.go
mkcp: Good bug catch, fixed.

src/internal/template/template.go
mkcp: Makes sense, updated in latest. Also added a todo for porting over the Helm builtins.

src/pkg/packager/packager.go
mkcp: It is a bit inconsistent - this uses the prior strategy where values didn't necessarily have to exist, so someone could optimistically map a value over. I'm alright shifting this to an error and have done so, though users may end up wanting defaults and/or the ability to not override if the field is empty. Something to track as the feature gains adoption

src/pkg/packager/actions/actions.go
mkcp: Take a look at the latest revision and let me know if that's what you meant. The whole retry section is a bit confusing but it ought to warn print the error results now.

src/pkg/packager/layout/assemble.go
mkcp: I am open to changing the build destination of package values, but I'd like to continue this convo in a follow up PR/scope of work. There are some considerations like whether we should enforce where users store their values files ahead of time, or copy them in from any import location and flatten / deduplicate / preserve reference dir structure.

AustinAbro321: I think it's worth evaluating in this PR because changing the package structure is tricky. Any packages created before the structure is changed won't work with newer versions of Zarf and vice versa. 

The questions you raise are fair, but imo it makes more sense to answer those questions in future issues / PRs package wide and stick with the current patterns when copying in manifests/valuesfiles/files. The current structure can be seen in [layout.go](https://github.com/zarf-dev/zarf/blob/307d965b91bd261162da857a5caa2e64bac6c6ad/src/pkg/packager/layout/layout.go#L4) where any user submitted files are placed in pre-defined directories. This simplifies the code by avoiding collisions and also lends structure when examining a decompressed package. 

Definitely possible I'm missing downsides of the current approach or there are upsides to this approach I'm not understanding.

src/pkg/packager/actions/actions.go
AustinAbro321: Yup that's what I meant, also the fact that we default on empty string makes this much more unlikely to be ran into now

src/pkg/packager/packager.go
AustinAbro321: Yeah I'm still about 50/50 on the decision and interested in feedback. I think it could be a nice boon and help tighten feedback loops. The only thing that's not ideal is that fields need to be set even if there is the `default` key used in go templates. 

Either way it'll be good to get user feedback on this. If a significant amount don't like it, it's much easier to turn off requiring fields then to turn on requiring fields.

src/pkg/packager/actions/actions.go
```diff
@@ -78,6 +86,12 @@ func runAction(ctx context.Context, basePath string, defaultCfg v1alpha1.ZarfCom
 		cmdEscaped = helpers.Truncate(cmd, 60, false)
 	}
 
+	// Apply go-templates in cmds
+	cmd, err = template.Apply(ctx, cmd, tmplObjs)
+	if err != nil {
+		l.Error("could not template cmd", "cmd", cmdEscaped, "err", err.Error())
```
AustinAbro321: Do we want to just log the error instead of returning it here? I know that for variable templating we only log the error, but I think that was more so a reflection of Zarf's stance on errors in the past where we skipped a lot of them.

src/cmd/package.go
```diff
@@ -261,6 +264,7 @@ func newPackageDeployCommand(v *viper.Viper) *cobra.Command {
 	cmd.Flags().BoolVar(&o.adoptExistingResources, "adopt-existing-resources", false, lang.CmdPackageDeployFlagAdoptExistingResources)
 	cmd.Flags().DurationVar(&o.timeout, "timeout", v.GetDuration(VPkgDeployTimeout), lang.CmdPackageDeployFlagTimeout)
 
+	cmd.Flags().StringSliceVarP(&o.valuesFiles, "values", "v", v.GetStringSlice(VPkgDeployValues), lang.CmdPackageDeployFlagValuesFiles)
```
brandtkeller: Non-blocking: we recently introduced a [modified getter](https://github.com/zarf-dev/zarf/blob/main/src/cmd/viper.go#L230-L239) for string slices that resolves some inconsistencies across env/flag/config-file.

src/test/e2e/42_values_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// Create the package
+	stdOut, stdErr, err := e2e.Zarf(t, "package", "create", src, "-o", tmpdir, "--skip-sbom", "--confirm", "--features=\"values=true\"")
+	require.NoError(t, err, stdOut, stdErr)
+
+	packageName := fmt.Sprintf("zarf-package-test-values-%s.tar.zst", e2e.Arch)
+	path := filepath.Join(tmpdir, packageName)
+
+	// Deploy the package
+	stdOut, stdErr, err = e2e.Zarf(t, "package", "deploy", path, "--confirm")
```
AustinAbro321: I think it would be good to test `--values` here. One of the harder things to unit test since it's at the cmd level and deploy.

examples/values-templating/zarf.yaml
```diff
@@ -0,0 +1,95 @@
...
+  - name: EXAMPLE_VAR
+    default: "bar"
+
+components:
+  - name: values-with-manifest
+    description: This component demonstrates templating manifest files by deploying an nginx page to the cluster
+    required: true
+    images:
+      # NOTE: Pkgtemplates aren't yet available with value and go-templates.
+      - "nginx:latest"
```
brandtkeller: Nit: for both here and in the deployment - we should use the long-name for images. Container runtimes such as CRI-O are beginning to enforce this due to ambiguity in registry location. This should likely be something Zarf detects in the future.

[reference](https://www.replicated.com/blog/breaking-change-kubernetes-1-34-tightens-image-name-resolution-rules----what-it-means-for-you)


```suggestion
      - "docker.io/library/nginx:latest"
```

Additionally if any scanners - [KICS](https://www.kics.io/index.html) for example - scan our source code I expect it to flag a finding for "latest" tags.

AustinAbro321: Yeah Zarf handles `:latest`, but generally I think it's preferred to avoid it in examples

src/cmd/package.go
mkcp: Nice! - updated in latest

examples/values-templating/zarf.yaml
mkcp: Makes sense -- updated in branch

src/pkg/packager/actions/actions.go
mkcp: Yeah - probably better to error here rather than matching the existing behavior. updated

src/test/e2e/42_values_test.go
mkcp: Yep, makes sense to get full coverage. Updated in latest

src/pkg/packager/layout/assemble.go
mkcp: That's fine, I went ahead and updated it by constraining package values files to the values subdirectory. That handles the enforcement part of the problem. Deduplication is another problem that will arise in package imports, but I can solve that by updating the existing PR.

In the future, it's clearly documented that alpha features are subject to change - and the whole point is giving ourselves flexibility to iterate and experiment. Users have to opt-in with a feature flag to try it out - we shouldn't fear making breaking changes or let that slow us down.

examples/values-templating/nginx-deployment.yaml
```diff
@@ -0,0 +1,55 @@
...
+  # Sprig example: Use configured replica count with a sensible default
+  replicas: {{ .Values.app.replicas | default 1 }}
+  template:
+    metadata:
+      labels:
+        app: nginx
+    spec:
+      containers:
+      - name: nginx
+        image: nginx:latest
```
brandtkeller: quickfix - this image needs to match the defined image in the zarf manifest

src/pkg/packager/layout/layout.go
```diff
@@ -45,3 +48,29 @@ const (
 	DataComponentDir      ComponentDir = "data"
 	ValuesComponentDir    ComponentDir = "values"
 )
+
+// ContainsReservedFilename checks if a path uses a reserved Zarf filename
+func ContainsReservedFilename(path string) error {
```
AustinAbro321: These funcs are no longer used

mkcp: Oh shoot i thought i got those, nice catch. fixed

examples/values-templating/nginx-deployment.yaml
mkcp: Thank you - definitely missed a spot! Fixed in latest

brandtkeller (approved):
Appreciate your persistence in advocating for alpha features to ship early for feedback while maintaining an open mind to required changes. 

LGTM!

PR #4158, opened 2025-09-08
fix: update dependabot npm ignores to compromised library versions

mkcp:
## Description
This PR adds a list of compromised package versions as as result of a supply chain attack, described here: https://www.aikido.dev/blog/npm-debug-and-chalk-packages-compromised. Zarf was not affected by this attack, and many of the compromised package versions have been removed as of the time of this PR. The CVE has also been flagged and registered, so it's likely that Dependabot already won't generate PRs for versions.

However, out of an abundance of caution, we're intentionally removing these packages from Dependabot updates to ensure we don't pull in a known-bad package. 

## Checklist before merging

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -32,8 +32,9 @@ updates:
     schedule:
       interval: daily
     ignore:
-      # NOTE(mkcp): These packages were compromised as part of a supply chain attack. Zarf was not affected and many of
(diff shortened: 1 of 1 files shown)


mkcp:
Each of the compromised versions have been resolved upstream so we can close this PR out safely without having to ignore these versions.

Issue #4160, opened 2025-09-09
Unclear guidance around switching from in-cluster zarf registry to external zarf registry


chaospuppy:
## Overview
The process of switching from a cluster-internal zarf registry to an external zarf registry is a bit unclear, and there are a few messages along the way that are slightly confusing.

## Steps to reproduce
1. Deploy UDS Core (or anything) to a cluster without specifying any zarf init overrides (i.e. using internal cluster registry)
2. Provide zarf init overrides to change to an external cluster
set `INIT_REGISTRY_URL`, `INIT_REGISTRY_PUSH_[USERNAME|PASSWORD]`, `INIT_REGISTRY_PULL_[USERNAME_PASSWORD]` via `uds-config`, via `--set`, or, presumably, while rerunning `zarf init`
3. See the following messages indicated an external registry has been detected:

<img width="1202" height="186" alt="Image" src="https://github.com/user-attachments/assets/0d29526d-1c80-45d6-87b5-ab5a7773e763" />
4. Notice warning about needing to run `uds zarf tools update-creds registry`
5. Run `uds zarf tools update-creds registry`, with environment/config init overrides still present, assuming that these would be picked up
6. Notice that the registry url does not change in `update-creds` command output:

<img width="790" height="222" alt="Image" src="https://github.com/user-attachments/assets/54f90a19-bf4a-4526-bb63-df1184176294" />

## Suggestions
Ultimately with support via slack it became apparent that the recommended approach for migrating from in-cluster to external registry is this process:
1. Use the zarf tools update-creds registry command to update the Zarf state with `--registry-push-username={PUSH_USERNAME} --registry-push-password={PUSH_PASSWORD} --registry-pull-username={PULL_USERNAME} --registry-pull-password={PULL_PASSWORD} --registry-url={REGISTRY_URL}` set
2. Update UDS config/environment vars/`--set` to match
3. Redeploy all packages or mirror images from the old registry to the new one
4. Run zarf package remove init --components zarf-registry,zarf-seed-registry,zarf-injector (if the registry was internal)

This guidance works and should be captured somewhere, while the bug that results in the `skipping init package component since external registry information was provided` message being emitted despite there being no intended change to the registry used should be fixed.

brandtkeller:
Thanks @chaospuppy - both enhancements to the process as well as documentation around best practices would be valuable here.

chaospuppy:
[...]

PR #4161, opened 2025-09-09
feat!: error early on changes to zarf services during a subsequent init

### LINKED ISSUE #3189: error out if users try to change Zarf deployed services credentials on re-init
When a user attempts to update the credentials of Zarf-deployed service using `zarf init` and flags such as `--artifact-*`, `--registry-*`, or `--git` with values different from the existing ones, they receive a warning that these services cannot be updated during a re-init. Instead of ignoring these flags, we should require users to execute the command with flags that are not ignored.

- **Given** I have a cluster already initialized with a registry pull password of "original-value"
- **When** I run zarf init --registry-pull-password="new-value"
- **Then** Zarf errors out immediately and tells me to use `zarf tools update-creds` 

One alternative to consider would be actually allowing the credentials to be changed on re-init. IMO better to keep things simple for now and keep these commands separate, even if we implemented this issue, we could always change to allowing in the future if users ask for it.

AustinAbro321:
## Description

Something that has confused many users is that if they try to change their registry credentials on a subsequent init, it will not work. They do get an error message, but it can be easily missed, and so it's better to return early and force the user to either use `zarf tools update-creds` or not call the command with flags that won't work 

## Related Issue

Fixes #3189
Relates to #4160

## Checklist before merging


src/pkg/cluster/cluster.go
```diff
@@ -265,17 +265,6 @@ func (c *Cluster) InitState(ctx context.Context, opts InitStateOptions) (*state.
 		s.RegistryInfo = opts.RegistryInfo
 		opts.ArtifactServer.FillInEmptyValues()
 		s.ArtifactServer = opts.ArtifactServer
-	} else {
-		// TODO (@austinabro321) validate immediately in `zarf init` if these are set and not equal and error out if so
```
mkcp, line 269: 👍

PR #4162, opened 2025-09-09
feat(registry-proxy)!: built-in mtls support

### LINKED ISSUE #2146: Improve security of zarf registry NodePort
Currently the zarf registry uses a NodePort which a) exposes an unneeded external listening port on every node and b) likely uses insecure communication between nodes.

Item a) can actually be mitigated today with [the `--nodeport-addresses=127.0.0.0/8` `kube-proxy` setting](https://kubernetes.io/docs/concepts/services-networking/service/#service-nodeport-custom-listen-address), but please note that this disables NodePorts entirely external to your nodes, this is not a solution for all users. However, if users do not need externally reachable NodePorts, this works today and disables the externally listening port.

One solution is to use a proper HTTPS registry, but this requires reconfiguring the CRI with the registry's likely-self-signed certificate. Today zarf is CRI agnostic and requires no manual configuration, perhaps this could be an alternate manual setting with the user manually reconfiguring their CRI, but I would like to see a secure configuration by default.

Another idea is to use a ClusterIP HTTPS registry and a  DaemonSet reverse proxy pod listening on a `127.0.0.1:31999` hostPort (HTTP, localhost only). Zarf uses a localhost HTTP registry for zero configuration with the CRI which this would still support (and HTTP is not insecure on localhost). Communication between nodes would be secured with HTTPS.

(A hostPort is preferable to a hostNetwork here. It's least privilege, we don't want the permissions of the host network namespace. IT supports specifying `hostIP: 127.0.0.1` so it will not be reachable externally, unlike NodePorts which depend on the `--nodeport-addresses` setting. It may be an issue that the hostPort will be inaccessible during registry restarts, a rolling restart like a NodePort supports is not possible. I'm also not sure what this startup will look like for new nodes without the DaemonSet pod yet on them and other things, needs some research.)

There are also a number of other in-cluster registries like https://github.com/XenitAB/spegel (which appears to be lighter weight than Dragonfly and Kraken). Spegel appears to do a very similar `hostPort` only binding as above, it's worth reviewing their configuration regardless. If we like one of these registries well enough to recommend we should test it, they should be compatible with zarf as an external registry and could be installed during zarf init, some minor work may be needed to bootstrap images.

AustinAbro321:
## Description

Fixes #2146

The breaking change in this PR is that by default during the next `zarf init --registry-mode=proxy` run mtls will be added and enforced. This shouldn't break anything within Zarf, but if a user expects to be able to talk to the registry with plainHTTP they may have to start integrating with the cert secrets. 

This changes the mutatingwebhook to mutate the registry to a service DNS name instead of a clusterIP. The reason this is done is because it would be very complex to specify a server host of a clusterIP, especially since the clusterIP could change if the service was regenerated.  

Certs will be automatically generated on init. If the lifetime of the cert is less than 50% of it's remaining lifetime then the cert will be re-generated. 

Two Kubernetes secrets will be generated. `zarf-registry-client-tls` and `zarf-registry-server-tls`, which will both be type `kubernetes.io/tls`. These secrets will create `ca.crt`, `tls.crt`, and `tls.key` as is the standard format. The `zarf-registry-client-tls` secret will be added to every namespace during deploy, similar to how the image pull secrets are added to namespaces. 

Since the registry now has a CA any clients calling it will now check the CA cert from the server. Older versions of Zarf won't recognize the CA since it's self signed.  In order to push to the registry they'll have to set `--insecure-skip-tls-verify`.  This is expected and okay especially since we're starting with this in alpha and if someone runs `zarf init --registry-proxy` then the new version of Zarf is already available in their airgap environment.  Newer versions of Zarf will automatically pull the CA from cluster when registry proxy is used.  Older versions of Zarf will not be able to push to the registry unless `--insecure-skip-tls-verify` is used.

## Checklist before merging

### DIFF
--- a/src/test/proxy/proxy_test.go
+++ b/src/test/proxy/proxy_test.go
@@ -43,10 +43,22 @@ func (suite *RegistryProxyTestSuite) Test_0_RegistryProxyInit() {
 	suite.NoError(err, "zarf-registry-client-tls secret should exist")
 }
 
-func (suite *RegistryProxyTestSuite) Test_1_UpdateCredsUpdatesMTLSSecrets() {
+func (suite *RegistryProxyTestSuite) Test_1_DeployRegularPackage() {
+	tmpdir := suite.T().TempDir()
+	stdOut, stdErr, err := e2e.Zarf(suite.T(), "package", "create", "examples/dos-games", "-o", tmpdir)
+	suite.NoError(err, stdOut, stdErr)
+
+	deployPath := filepath.Join(tmpdir, fmt.Sprintf("zarf-package-dos-games-%s-1.3.0.tar.zst", runtime.GOARCH))
+	stdOut, stdErr, err = e2e.Zarf(suite.T(), "package", "deploy", deployPath, "--confirm")
+	suite.NoError(err, stdOut, stdErr)
+
+	stdOut, stdErr, err = e2e.Zarf(suite.T(), "package", "remove", deployPath, "--confirm")
+	suite.NoError(err, stdOut, stdErr)
+}
+
+func (suite *RegistryProxyTestSuite) Test_2_UpdateCredsUpdatesMTLSSecrets() {
 	ctx := suite.T().Context()
 
-	// Get the original client TLS secret from the zarf namespace
 	originalPKI, err := suite.cluster.GetRegistryClientMTLSCert(ctx)
 	suite.NoError(err)
 	suite.NotEmpty(originalPKI.Cert)
@@ -54,23 +66,21 @@ func (suite *RegistryProxyTestSuite) Test_1_UpdateCredsUpdatesMTLSSecrets() {
 	stdOut, stdErr, err := e2e.Zarf(suite.T(), "tools", "update-creds", "registry", "--confirm")
 	suite.NoError(err, stdOut, stdErr)
 
-	// Get the updated client TLS secret from the zarf namespace
 	updatedPKI, err := suite.cluster.GetRegistryClientMTLSCert(ctx)
 	suite.NoError(err)
 	suite.NotEmpty(updatedPKI.Cert)
 
-	// Verify that the certificate was regenerated
+	// Verify that the original certificate matches what was regenerated
 	suite.NotEqual(originalPKI.Cert, updatedPKI.Cert)
 
-	// Get the updated client TLS secret from the podinfo-oci namespace
-	updatedNamespaceSecret, err := suite.cluster.Clientset.CoreV1().Secrets("podinfo-oci").Get(ctx, cluster.RegistryClientTLSSecret, metav1.GetOptions{})
+	// Get the updated client TLS secret from the dos-games namespace
+	updatedNamespaceSecret, err := suite.cluster.Clientset.CoreV1().Secrets("dos-games").Get(ctx, cluster.RegistryClientTLSSecret, metav1.GetOptions{})
 	suite.NoError(err)
 
-	// Verify the secret in podinfo-oci namespace matches the zarf namespace
 	suite.Equal(updatedPKI.Cert, updatedNamespaceSecret.Data[cluster.RegistrySecretCertPath])
 }
 
-func (suite *RegistryProxyTestSuite) Test_2_Flux() {
+func (suite *RegistryProxyTestSuite) Test_3_OCIOpsPackage() {
 	tmpdir := suite.T().TempDir()
 	stdOut, stdErr, err := e2e.Zarf(suite.T(), "package", "create", "examples/podinfo-flux", "-o", tmpdir)
 	suite.NoError(err, stdOut, stdErr)


src/cmd/initialize.go
```diff
@@ -113,7 +113,6 @@ func newInitCommand() *cobra.Command {
 	cmd.Flags().IntVar(&o.ociConcurrency, "oci-concurrency", v.GetInt(VPkgOCIConcurrency), lang.CmdPackageFlagConcurrency)
 
 	// If an external registry is used then don't allow users to configure the internal registry / injector
-	cmd.MarkFlagsMutuallyExclusive("registry-url", "registry-mode")
```
AustinAbro321, line 119: now that we have `registryModeExternal` this is unnecessary

src/internal/packager/images/push.go
```diff
@@ -93,14 +98,19 @@ func Push(ctx context.Context, cfg PushConfig) error {
 				Password: cfg.RegistryInfo.PushPassword,
 			}),
 		}
-
-		client.Client.Transport, err = orasTransport(cfg.InsecureSkipTLSVerify, cfg.ResponseHeaderTimeout)
-		if err != nil {
-			return err
+		if cfg.RegistryInfo.RegistryMode == state.RegistryModeProxy {
+			client.Client.Transport, err = orasTransportWithClientCertsFromSecrets(ctx, cfg.Cluster)
```
brandtkeller: ++ this is a noteworthy improvement

src/pkg/cluster/cluster.go
```diff
@@ -303,6 +320,122 @@ func (c *Cluster) InitState(ctx context.Context, opts InitStateOptions) (*state.
...
+	if percentageRemainingLife < remainingLifeRenewalThreshold {
+		return true, nil
+	}
+	return false, nil
+}
+
+// generateOrRenewRegistryCerts creates CA, server, and client certificates for registry mTLS
+// and applies them to the cluster as Kubernetes secrets with bundled CA certificates.
+// Only generates certificates if they don't exist or have less than 50% remaining life.
+func (c *Cluster) generateOrRenewRegistryCerts(ctx context.Context) error {
```
brandtkeller: This is distinctly different from how we manage certificates with the zarf agent. I am curious what tradeoffs are present and/or if we should be re-evaluating how we manage certificates for consistency?

src/pkg/cluster/cluster.go
```diff
@@ -44,6 +44,16 @@ const (
 	NamespaceOverrideLabel string = "zarf.dev/namespace-override"
 )
 
+// Registry TLS secret and certificate names
+const (
+	RegistryServerTLSSecret = "zarf-registry-server-tls"
+	RegistryProxyTLSSecret  = "zarf-registry-proxy-tls"
```
brandtkeller: This is bloating scope a little - but want to discuss. 

I believe zarf should be an enabler for applications to integrate with the zarf registry - using security protocols whenever possible. 

I would be curious about:
1. make the zarf secret name for the client information more generic
2. add future requirements that make this something we can automatically inject into zarf managed namespaces
3. Manage rolling out-of-band

Thoughts?

AustinAbro321: 1. Great call out we should definitely make these generic. Changed `RegistryProxyTLSSecret` to `RegistryClientTLSSecret`. Also changed the push logic to use mtls whenever the client cert is available to make this more future proof. 
2. Yeah I think that would be a good future addition
3. By this do you mean allowing a path for users to create the secrets themselves?

AustinAbro321:
@brandtkeller For IPv6 testing, worth noting that .github/workflows/test-proxy.yml runs tests in a matrix for ipv4 and ipv6

src/pkg/cluster/cluster.go
AustinAbro321: It's an interesting topic. I do think there is a desire to change the agent cert so that it updates automatically on init https://github.com/zarf-dev/zarf/issues/3367 . I did verify that this will renew even when the cert is expired. One question is if we want to update the certs automatically on `zarf tools update-creds`? I think the answer is yes and I've added that. I like trying out both here for now, and if we don't run into any issues it could serve as a model for the agent later.

src/pkg/cluster/cluster.go
brandtkeller: Item 3 is largely covered by the mention of `zarf tools update-creds` but I was thinking about an issue to capture enhancing third-party integration. IE copy the secret into all zarf managed namespaces by default (as we do with the image pull secret for the registry) as well as an opt-in capability for rolling pods based on some annotation. 

that is all out-of-scope - what I really want to capture here is consistency. We manage secrets and certificates (as a  subset) and just want to ensure we evaluate how to be consistent in the approach where possible now or in the future (IE updating old logic).

brandtkeller:
Do we have a mental model for how this implementation responds to deployments on top of a service mesh?

AustinAbro321:
I am definitely not an expert in service mesh, however I think this would make a sidecar not very useful for the registry / registry-proxy since traffic will be encrypted between those. This would also likely be hard to configure ahead of time, unless isitio pre-dated Zarf in the cluster. 

I tried out using the registry proxy setup in an isitio ambient mode enabled cluster in [uds-core](https://github.com/defenseunicorns/uds-core), and it worked as expected.

src/internal/packager/helm/post-render.go
```diff
@@ -161,6 +162,24 @@ func (r *renderer) adoptAndUpdateNamespaces(ctx context.Context) error {
 		if err != nil {
 			return fmt.Errorf("problem applying git server secret for the %s namespace: %w", name, err)
 		}
+
+		// Copy mTLS client secret if it exists in the Zarf namespace
+		mtlsSecret, err := c.Clientset.CoreV1().Secrets(state.ZarfNamespaceName).Get(ctx, cluster.RegistryClientTLSSecret, metav1.GetOptions{})
+		if err != nil {
```
brandtkeller: Nit: There is a silent continue implied here that I don't think is clear at first glance. I think a better model would be:

```golang
mtlsSecret, err := c.Clientset.CoreV1().Secrets(state.ZarfNamespaceName).Get(ctx, cluster.RegistryClientTLSSecret, metav1.GetOptions{})
if kerrors.IsNotFound(err) {
    // No mTLS configured, skip
    continue 
}
if err != nil {
    return fmt.Errorf("failed to get mTLS client secret: %w", err)
}
// Now use mtlsSecret...
```

src/internal/agent/hooks/flux-helmrepo.go
```diff
@@ -82,13 +83,28 @@ func mutateHelmRepo(ctx context.Context, r *v1.AdmissionRequest, cluster *cluste
...
 	// Mutate the helm repo URL if necessary
 	if isCreate || (isUpdate && !isPatched) {
-		patchedSrc, err := transform.ImageTransformHost(registryAddress, src.Spec.URL)
-		if err != nil {
-			return nil, fmt.Errorf("unable to transform the HelmRepo URL: %w", err)
+		var patchedSrc string
+		if isPatchedClusterIP {
+			patchedSrc, err = transform.ImageTransformHostWithoutChecksum(registryAddress, src.Spec.URL)
+			if err != nil {
+				return nil, fmt.Errorf("unable to transform the HelmRepo URL: %w", err)
```
brandtkeller: nit: both error returns contain the same log - underlying function return errors may give us more context but some uniqueness here can optimize traceability.

.github/workflows/test-proxy.yml
```diff
@@ -90,6 +88,11 @@ jobs:
         run: |
           kind delete cluster && kind create cluster --config=kind-nftables-config.yml
 
+      - name: Setup golang
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
```
brandtkeller: minor - main is on a newer version

packages/zarf-registry/chart/templates/secret.yaml
```diff
@@ -9,7 +9,39 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
 type: Opaque
+stringData:
+  configData: |
```
brandtkeller: No issue with this moving but will note the implication of the data in the secret means that we (and potentially users) won't have access to modifying these with values. Will need to look at which configurations are now static here in the event someone was creating a custom init package that added values via merge perhaps?

(more of a mental note to review further)

src/internal/packager/helm/post-render.go
```diff
@@ -161,6 +162,24 @@ func (r *renderer) adoptAndUpdateNamespaces(ctx context.Context) error {
 		if err != nil {
 			return fmt.Errorf("problem applying git server secret for the %s namespace: %w", name, err)
 		}
+
+		// Copy mTLS client secret if it exists in the Zarf namespace
```
brandtkeller: Why only if it exists do we update it? Should we not be looking for the lack of existence and adding it as well?

src/pkg/pki/pki.go
```diff
@@ -170,28 +187,210 @@ func generateCert(host string, ca *x509.Certificate, caKey *rsa.PrivateKey, notA
...
-	if !certHas20PercentRemainingTime {
+	if remainingLife < 20 {
 		logger.From(ctx).Warn("the Zarf agent certificate is expiring soon, run `zarf tools update-creds agent` to update", "expiration", cert.NotAfter)
 	}
 	return nil
 }
+
+// GenerateMTLSCerts generates a complete set of mTLS certificates including CA, server cert, and client cert.
+// Returns two GeneratedPKI structs: one for the server (containing server cert, key, and CA) and one for the client (containing client cert, key, and CA).
+func GenerateMTLSCerts(serverDNSNames []string, serverCommonName string, clientCommonName string) (server GeneratedPKI, client GeneratedPKI, err error) {
```
brandtkeller: Would this logic be able to be expanded or modified if we wanted to extend it to support other mtls cert creation (gitea)?

src/pkg/cluster/cluster.go
```diff
@@ -309,6 +327,107 @@ func (c *Cluster) InitState(ctx context.Context, opts InitStateOptions) (*state.
...
+
+	return needsCARenewal || needsServerRenewal || needsClientRenewal, nil
+}
+
+// ApplyRegistryCerts applies the provided server and client certificates to the cluster as Kubernetes secrets.
+// Both the server and client PKI bundles should contain the same CA certificate.
+func (c *Cluster) ApplyRegistryCerts(ctx context.Context, serverPKI, clientPKI pki.GeneratedPKI) error {
+	l := logger.From(ctx)
+
+	serverSecret := v1ac.Secret(RegistryServerTLSSecret, state.ZarfNamespaceName).
```
brandtkeller, line 427: Should `serverSecret` and `clientSecret` have the zarf managed label here? I see this is the entrypoint for creating the zarf namespace secrets that are then copied elsewhere.

`cluster.UpdateZarfManagedMTLSSecrets()` will only update if the label is present.

src/internal/packager/helm/post-render.go
```diff
@@ -161,6 +162,24 @@ func (r *renderer) adoptAndUpdateNamespaces(ctx context.Context) error {
...
+		mtlsSecret, err := c.Clientset.CoreV1().Secrets(state.ZarfNamespaceName).Get(ctx, cluster.RegistryClientTLSSecret, metav1.GetOptions{})
+		if err != nil {
+			if !kerrors.IsNotFound(err) {
+				return fmt.Errorf("failed to get mTLS client secret: %w", err)
+			}
+		} else {
+			mtlsSecretApply := v1ac.Secret(mtlsSecret.Name, name).
+				WithData(mtlsSecret.Data).
+				WithType(mtlsSecret.Type).
+				WithLabels(mtlsSecret.Labels)
```
brandtkeller: Should we check these labels prior to copy? I know we've hit issues in the past with labels being nil. Ideally they won't be here - but I might almost default to re-declaring the zarf managed label as well.

src/internal/packager/helm/post-render.go
AustinAbro321: I disagree, since it's reasonable that a future addition to this block may be unrelated to mtls. For instance, if we wanted to apply another secret in this function the developer would have to know to add the secret before the mtls block otherwise it sometimes wouldn't be applied.

src/internal/packager/helm/post-render.go
AustinAbro321: I was using the existence of the secret to determine if this is a cluster that supports mtls. I don't think we always want to add it to the cluster, because a user's zarf install may not support mtls, however I do think there is a more clear check we could do. 

Two ideas: 
1. Check if the cluster supports mtls, if it does then error if the secret is not found
2. Store the `pki.GeneratedPKI` of the client cert on the registryInfo struct similar to how state tracks the AgentTLS cert. Duplicates the data somewhat, but stops us on relying on this specific secret. 

I'm going with option 1 for now, I think this is cleaner

src/pkg/pki/pki.go
AustinAbro321: Yes this function generates generic mTLS certs, it would be able to create certs for gitea as is.

packages/zarf-registry/chart/templates/secret.yaml
AustinAbro321: A user wouldn't have been able to change any of the data on deploy, but could do it with a custom init package by adding an additional values file. I do think that'd be a rare case, but maybe something we'd want to allow. 

Will think on this some more and play around with the structure. Long term do we want to expose this entire block? Maybe yes since we have such a wide range of users

AustinAbro321:
>Can this implementation handle upgrading an existing deployment (proxy or otherwise) to include mTLS in all zarf managed namespaces?

@brandtkeller Once this PR is in proxy installations will be upgraded to use mTLS automatically. There is not a way to use mTLS without the registry proxy as of this PR, but that could be future work

edit: originally mis-read your question, the above is true for the Zarf namespace. For existing deployments you can't currently do that with this PR. Going to stew on this a bit more, but I'm leaning towards `zarf tools update-creds` updating all namespaces to include the secret

src/pkg/cluster/cluster.go
AustinAbro321: Yes nice catch, fixed. Functionally the proxy e2e test with update creds was still passing because it's an "and" condition with the skip namespace. That does seem a bit weird now that I look at it, but for now better to stay in line with the other functions

src/internal/packager/helm/post-render.go
AustinAbro321: Yeah I think this code block was kinda awkward. I ended up redesigning it a bit so that we re-create the secret from scratch using the client PKI instead of copying it over so we avoid duplication and issues like you're pointing out

brandtkeller:
Do we want to include `zarf tools registry` commands in scope for the mtls changes? If not we'll want to call out the `--insecure` option being required.

AustinAbro321:
@brandtkeller Yes, good callout. It was a simple addition

AustinAbro321:
>Can this implementation handle upgrading an existing deployment (proxy or otherwise) to include mTLS in all zarf managed namespaces?

Updated the code to add mTLS secrets to all Zarf managed namespaces. This also handles updating all the secrets when the cert is renewed on init

src/cmd/zarf_tools.go
```diff
@@ -336,6 +336,13 @@ func (o *updateCredsOptions) run(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
+
+		if newState.RegistryInfo.MTLSStrategy == state.MTLSStrategyZarfManaged {
```
brandtkeller, line 340: observation: understanding the distinction here - because this operation performs both the re-generation of TLS information AND also updates the cluster secrets - we do not get any feedback with `zarf tools update-creds registry` that the TLS secrets are being modified. 

Feels like a constraint of not storing the secrets in state but I am curious if there is optionality to separate generation from "apply to the cluster" (typically referred to as "Update") in such a way so as to signal that change. 

Otherwise I do believe that advanced orchestrators could still implement a workflow that stages a change in the secret content before applying it cluster wide.

src/pkg/pki/pki.go
brandtkeller: Is the statically defined `GenerateCA("Zarf Registry CA")` going to present any issues?

brandtkeller (commented):
WIP - testing locally but dropping a few more comments before completing review.

src/pkg/pki/pki.go
AustinAbro321: Whoops I missed that, updated

src/cmd/zarf_tools.go
AustinAbro321: Yeah we did not call out that the registry certs were changing during the credential update confirmation prompt. I've updated this. Is there another behavior, logging, or strategy change you're looking for when updating the mtls secrets?

brandtkeller: Outside of indicating the change occurring - my only curiosity is how best we can ensure observability of synchronization across the cluster. We allow a lot of potentially intermediate state - IE someone adds a zarf ignore label to a previously managed namespace. Do we want to add any identifying metadata to the secret that indicates the namespaces match the source secret?

AustinAbro321: I wouldn't want to add a identifying metadata as Zarf would have to manage the metadata, and the metadata itself might get out of sync, especially in cases of ignored secrets.

Perhaps there is an opportunity for a broader look at this management in a future PR / issues. My goal for this PR was to mimic the behavior of the registry and git pull secrets

PR #4164, opened 2025-09-09
fix(ci): remove nightly environment

brandtkeller:
## Description

Nightly release workflow is currently failing due to the use of the `release` environment on the `create-release` job. 

> branch "main" is not allowed to deploy to release due to environment protection rules.

This removes the environment - as the protections to block what branches/tags can push to the environment prevent `main` and access to the environment secrets are not required.  

## Related Issue

Relates to #3928

## Checklist before merging


mkcp (approved):
LGTM

PR #4172, opened 2025-09-11
fix(deploy): track success/failed chart install status

### LINKED ISSUE #4171: Zarf package remove fails to remove unsuccessful helm releases
### Environment
Device and OS: macOS
App version: 0.60.0
Kubernetes distro being used: k3s in k3d, k8s 1.31.4
Other:

### Steps to reproduce
1. `zarf package deploy <package>` on package containing helm release that create resources which will fail to become healthy in the cluster
2. `zarf package remove <package> --confirm`

### Expected result
Helm releases are removed 

### Actual Result
Helm releases are not removed

### Severity/Priority
Medium(?)

brandtkeller:
## Description

This PR adds state tracking for component Installed Charts. 

There is currently a gap in the deploy lifecycle where a user attempts to deploy a package -  where a component chart subsequently fails to become healthy - and the deploy fails. 

### Chart State

This leaves users in a state where If this is an initial deploy - no installed charts are recorded for the component - and thus they cannot be removed with `zarf package remove`. You then have to cleanup resources with helm (not the embedded version) or kubectl (which is messy). This is a very common process for package development. 

Installed Chart state is not strictly required to solve this - but knowing which of the charts of a component failed on subsequent updates - outside of the CLI logs - could be a future value add. 

### Context Cancellation

When the scenario highlighted above occurs - developers may not want to wait for the timeout (up to multiple minutes) to occur. Given a common scenario - `errimagepull` occurs - the package developer will very quickly realize that they need to add a missing or incorrect image definition to a component. They (at least myself) will reach for `ctrl+c` to cancel deploy. 

The above changes for tracking failure will not handle this case natively. The context cancellation will propagate to the client-go operations and will not perform the recording in the package secret.

This PR additionally includes context handling to ensure the recording can finish before exiting gracefully. 

## Related Issue

Fixes #4171

## Checklist before merging


src/pkg/cluster/zarf.go
```diff
@@ -250,6 +250,54 @@ func (c *Cluster) GetInstalledChartsForComponent(ctx context.Context, packageNam
 	return installedCharts, nil
 }
 
+// MergeInstalledChartsForComponent merges the provided existing charts with the provided installed charts.
+func (c *Cluster) MergeInstalledChartsForComponent(existingCharts, installedCharts []state.InstalledChart, partial bool) []state.InstalledChart {
```
AustinAbro321: I'd like to see unit tests for this function, I also think it should be moved out of cluster package since it doesn't use the cluster object. I think either in packager2 deploy.go or the state package

src/internal/packager/helm/chart.go
```diff
@@ -148,12 +148,12 @@ func InstallOrUpgradeChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, ch
 		if err != nil {
 			return nil, "", fmt.Errorf("%w: unable to rollback: %w", installErr, err)
 		}
-		return nil, "", installErr
+		return nil, zarfChart.ReleaseName, installErr
```
AustinAbro321, line 151: I think a few lines above this we intend to return `zarfChart.ReleaseName` but don't

src/pkg/state/state.go
```diff
@@ -412,9 +418,14 @@ type DeployedComponent struct {
...
+// ChartStatus is the status of a Helm Chart release
+type ChartStatus string
+
 // InstalledChart contains information about a Helm Chart that has been deployed to a cluster.
 type InstalledChart struct {
 	Namespace      string         `json:"namespace"`
 	ChartName      string         `json:"chartName"`
 	ConnectStrings ConnectStrings `json:"connectStrings,omitempty"`
+	Status         ChartStatus    `json:"status"`
+	Tracked        bool           `json:"tracked"`
```
AustinAbro321: May be worth splitting the additional of InstalledChart.Status and InstalledChart.tracked into separate PRs

brandtkeller: agreed - removed `tracked` for the moment.

src/pkg/cluster/zarf.go
brandtkeller: Agreed - updated

src/internal/packager/helm/chart.go
brandtkeller: good catch!

src/pkg/state/state_test.go
```diff
@@ -283,3 +283,85 @@ func TestMergeStateAgent(t *testing.T) {
...
+
+func TestMergeInstalledChartsForComponent(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name            string
+		existingCharts  []InstalledChart
+		installedCharts []InstalledChart
+		expectedCharts  []InstalledChart
+	}{
```
AustinAbro321, line 295: nit but I think partial should be a field on this table test as well

brandtkeller: fair - `partial` has no functional purpose yet - but wanted to set us up for prune without having to make a breaking change. 

lookup starts with the existing state and then adds each chart that was installed by the component. For prune - partial represents that we won't want to mark a release as untracked because deploy may have failed before it was deployed. 

Open to removing it - but low overhead.

brandtkeller:
> Overall I think these changes look good, but I'd like to see an e2e test in [41_remove_test.go](https://github.com/zarf-dev/zarf/blob/main/src/test/e2e/41_remove_test.go) for this behavior

Added a test - great call.

PR #4175, opened 2025-09-12
ci: stop building examples

AustinAbro321:
## Description

Currently the action .github/actions/packages has a step to build examples. Only the nightly eks cluster and test-windows workflow uses it, however building examples is no longer needed since we now have every e2e test build the packages it needs within the test.  

## Checklist before merging


AustinAbro321:
Proof that eks workflow still works https://github.com/zarf-dev/zarf/actions/runs/17675533071/job/50236625703

PR #4178, opened 2025-09-14
feat!: add custom names of init package in tar ball

### LINKED ISSUE #4095: Allow the custom name of init packages, that are not name `init`, in the locally pull tar-ball

It is very unclear when working with different init packages, say one for a single node cluster and one for a multi-node cluster, to have them both be called `zarf-init-vX.X.X-amd64.tar.zst`.

- **Given** a `.metadata.name` of not `init` (for backwards compatibility with the main init package).
- **Then** the `.metadata.name` is appended to the local tar-bundle, such as `zarf-init-single-vX.X.X-amd64.tar.zst`

After pulling one zarf init package to manually rename it on disk, then pull the second init package.

a1994sc:
## Description

When the Zarf init package is not empty or when named `init`, this PR will now include that name in the pulled down tar ball.

## Related Issue

Fixes #4095

## Checklist before merging

### DIFF
--- a/src/pkg/packager/layout/package.go
+++ b/src/pkg/packager/layout/package.go
@@ -264,7 +264,11 @@ func (p *PackageLayout) FileName() (string, error) {
 	var name string
 	switch p.Pkg.Kind {
 	case v1alpha1.ZarfInitConfig:
-		name = fmt.Sprintf("zarf-init-%s", arch)
+		if p.Pkg.Metadata.Name == "" || p.Pkg.Metadata.Name == "init" {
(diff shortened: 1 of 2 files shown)


AustinAbro321:
Hmm I don't love special casing the default init package, but I do agree that currently it's not good behavior that name changing the init package doesn't affect the package name. Another option could be removing the specialness from the init package in general and naming it `zarf-package-init` like every other package. This is a bigger breaking change since no doubt many users have shell scripts that pick up the init package. I think the initial naming convention was designed without custom init packages in mind. 

What do the other @zarf-dev/maintainers think?

PR #4193, opened 2025-09-17
feat: merge chart properties

vincent-onebrief:
## Description

The `charts` key supports merging, but doesn't support merging all properties. This isn't covered by [the documentation](https://docs.zarf.dev/ref/components/#merge-strategies) well.  Being able to override these properties allows for more complex CI testing where the zarf package needs to reference a derivative chart specific to the PR (as an example).

## Checklist before merging


src/pkg/packager/load/import_test.go
```diff
@@ -60,6 +60,10 @@ func TestResolveImports(t *testing.T) {
 			path:   "./testdata/import/flavor",
 			flavor: "pistachio",
 		},
+		{
+			name: "skeleton properties are overridden",
+			path: "./testdata/import/override",
```
AustinAbro321:
```suggestion
			name: "chart version and url properties are not overridden",
			path: "./testdata/import/chart",
```
nit: small changes to naming

AustinAbro321 (changes requested):
Generally I agree with this change. It doesn't really make sense to specify the url and version in your base zarf.yaml and for it to be impossible to have an effect. Have a minor the test / directory name

PR #4194, opened 2025-09-17
feat: labeling all resources created by a zarf package

### LINKED ISSUE #3625: Add zarf label to all resources created by a package deploy

sgettys:
## Description

Adds labels to all resources created by a zarf package:
```
zarf.dev/package: <package name>
zarf.dev/namespace-override: <namespace if provided>
```

This will allow for easy filtering of resources associated with a zarf package using label selectors.

## Related Issue

Fixes #3625 

## Checklist before merging

### DIFF
--- a/src/internal/packager/helm/chart.go
+++ b/src/internal/packager/helm/chart.go
@@ -18,7 +18,6 @@ import (
 	"github.com/zarf-dev/zarf/src/pkg/variables"
 
 	"github.com/Masterminds/semver/v3"
-	"github.com/avast/retry-go/v4"
 	plutoversionsfile "github.com/fairwindsops/pluto/v5"
 	plutoapi "github.com/fairwindsops/pluto/v5/pkg/api"
 	goyaml "github.com/goccy/go-yaml"
@@ -95,35 +94,25 @@ func InstallOrUpgradeChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, ch
 	helmCtx, helmCtxCancel := context.WithTimeout(ctx, opts.Timeout)
 	defer helmCtxCancel()
 
-	err = retry.Do(func() error {
-		var err error
+	releases, histErr := histClient.Run(zarfChart.ReleaseName)
 
-		releases, histErr := histClient.Run(zarfChart.ReleaseName)
+	l.Debug("checking for existing helm deployment")
 
-		l.Debug("checking for existing helm deployment")
+	if errors.Is(histErr, driver.ErrReleaseNotFound) {
+		// No prior release, try to install it.
+		l.Info("performing Helm install", "chart", zarfChart.Name)
 
-		if errors.Is(histErr, driver.ErrReleaseNotFound) {
-			// No prior release, try to install it.
-			l.Info("performing Helm install", "chart", zarfChart.Name)
+		release, err = installChart(helmCtx, zarfChart, chart, values, opts.Timeout, actionConfig, postRender)
+	} else if histErr == nil && len(releases) > 0 {
+		// Otherwise, there is a prior release so upgrade it.
+		l.Info("performing Helm upgrade", "chart", zarfChart.Name)
 
-			release, err = installChart(helmCtx, zarfChart, chart, values, opts.Timeout, actionConfig, postRender)
-		} else if histErr == nil && len(releases) > 0 {
-			// Otherwise, there is a prior release so upgrade it.
-			l.Info("performing Helm upgrade", "chart", zarfChart.Name)
+		lastRelease := releases[len(releases)-1]
 
-			lastRelease := releases[len(releases)-1]
-
-			release, err = upgradeChart(helmCtx, zarfChart, chart, values, opts.Timeout, actionConfig, postRender, opts.Cluster, lastRelease)
-		} else {
-			return fmt.Errorf("unable to verify the chart installation status: %w", histErr)
-		}
-
-		if err != nil {
-			return err
-		}
-
-		return nil
-	}, retry.Context(ctx), retry.Attempts(uint(opts.Retries)), retry.Delay(500*time.Millisecond))
+		release, err = upgradeChart(helmCtx, zarfChart, chart, values, opts.Timeout, actionConfig, postRender, opts.Cluster, lastRelease)
+	} else {
+		return nil, zarfChart.ReleaseName, fmt.Errorf("unable to verify the chart installation status: %w", histErr)
+	}
 	if err != nil {
 		removeMsg := "if you need to remove the failed chart, use `zarf package remove`"
 		installErr := fmt.Errorf("unable to install chart after %d attempts: %w: %s", opts.Retries, err, removeMsg)
--- a/site/src/content/docs/commands/zarf_dev_deploy.md
+++ b/site/src/content/docs/commands/zarf_dev_deploy.md
@@ -29,7 +29,7 @@ zarf dev deploy [flags]
   -h, --help                               help for deploy
       --no-yolo                            Disable the YOLO mode default override and create / deploy the package as-defined
       --registry-override stringToString   Specify a map of domains to override on package create when pulling images (e.g. --registry-override docker.io=dockerio-reg.enterprise.intranet) (default [])
-      --retries int                        Number of retries to perform for Zarf operations like git/image pushes or Helm installs (default 3)
+      --retries int                        Number of retries to perform for Zarf operations like git/image pushes (default 3)
       --timeout duration                   Timeout for health checks and Helm operations such as installs and rollbacks (default 15m0s)
 ```
 
--- a/site/src/content/docs/commands/zarf_init.md
+++ b/site/src/content/docs/commands/zarf_init.md
@@ -75,7 +75,7 @@ $ zarf init --artifact-push-password={PASSWORD} --artifact-push-username={USERNA
       --registry-push-username string   Username to access to the registry Zarf is configured to use (default "zarf-push")
       --registry-secret string          Registry secret value
       --registry-url string             External registry url address to use for this Zarf cluster
-      --retries int                     Number of retries to perform for Zarf operations like git/image pushes or Helm installs (default 3)
+      --retries int                     Number of retries to perform for Zarf operations like git/image pushes (default 3)
       --set stringToString              Specify deployment variables to set on the command line (KEY=value) (default [])
       --skip-signature-validation       Skip validating the signature of the Zarf package
       --storage-class string            Specify the storage class to use for the registry and git server.  E.g. --storage-class=standard
--- a/site/src/content/docs/commands/zarf_package_deploy.md
+++ b/site/src/content/docs/commands/zarf_package_deploy.md
@@ -27,7 +27,7 @@ zarf package deploy [ PACKAGE_SOURCE ] [flags]
   -c, --confirm                     Confirms package deployment without prompting. ONLY use with packages you trust. Skips prompts to review SBOM, configure variables, select optional components and review potential breaking changes.
   -h, --help                        help for deploy
   -n, --namespace string            [Alpha] Override the namespace for package deployment. Requires the package to have only one distinct namespace defined.
-      --retries int                 Number of retries to perform for Zarf operations like git/image pushes or Helm installs (default 3)
+      --retries int                 Number of retries to perform for Zarf operations like git/image pushes (default 3)
       --set stringToString          Specify deployment variables to set on the command line (KEY=value) (default [])
       --shasum string               Shasum of the package to deploy. Required if deploying a remote https package.
       --skip-signature-validation   Skip validating the signature of the Zarf package
--- a/site/src/content/docs/commands/zarf_package_mirror-resources.md
+++ b/site/src/content/docs/commands/zarf_package_mirror-resources.md
@@ -64,7 +64,7 @@ $ zarf package mirror-resources <your-package.tar.zst> --repos \
       --registry-push-username string   Username to access to the registry Zarf is configured to use (default "zarf-push")
       --registry-url string             External registry url address to use for this Zarf cluster
       --repos                           mirror only the git repositories
-      --retries int                     Number of retries to perform for Zarf operations like git/image pushes or Helm installs (default 3)
+      --retries int                     Number of retries to perform for Zarf operations like git/image pushes (default 3)
       --shasum string                   Shasum of the package to pull. Required if pulling a https package. A shasum can be retrieved using 'zarf dev sha256sum <url>'
       --skip-signature-validation       Skip validating the signature of the Zarf package
 ```
--- a/site/src/content/docs/commands/zarf_package_publish.md
+++ b/site/src/content/docs/commands/zarf_package_publish.md
@@ -32,7 +32,7 @@ $ zarf package publish ./path/to/dir oci://my-registry.com/my-namespace
   -c, --confirm                     Confirms package publish without prompting. Skips prompt for the signing key password
   -f, --flavor string               The flavor of components to include in the resulting package. The flavor will be appended to the package tag
   -h, --help                        help for publish
-      --retries int                 Number of retries to perform for Zarf operations like git/image pushes or Helm installs (default 1)
+      --retries int                 Number of retries to perform for Zarf operations like git/image pushes (default 1)
       --signing-key string          Private key for signing or re-signing packages with a new key. Accepts either a local file path or a Cosign-supported key provider
       --signing-key-pass string     Password to the private key used for publishing packages
       --skip-signature-validation   Skip validating the signature of the Zarf package
--- a/site/src/content/docs/ref/deploy.mdx
+++ b/site/src/content/docs/ref/deploy.mdx
@@ -154,12 +154,6 @@ The default timeout for Helm operations in Zarf is 15 minutes.
 
 Use the `--timeout` flag with `zarf init` and `zarf package deploy` to modify the timeout duration.
 
-### Retry Policy
-
-Zarf retries install and upgrade operations up to three times by default if an error occurs.
-
-Use the `--retries` flag with `zarf init` and `zarf package deploy` to change the number of retry attempts.
-
 ### Rollback Process
 
 If attempts to upgrade a chart fail, Zarf tries to roll the chart back to its last successful release. During this rollback process:
--- a/src/config/lang/english.go
+++ b/src/config/lang/english.go
@@ -210,7 +210,7 @@ $ zarf init --artifact-push-password={PASSWORD} --artifact-push-username={USERNA
 	CmdPackageFlagConcurrency             = "Number of concurrent layer operations when pulling or pushing images or packages to/from OCI registries."
 	CmdPackageFlagFlagPublicKey           = "Path to public key file for validating signed packages"
 	CmdPackageFlagSkipSignatureValidation = "Skip validating the signature of the Zarf package"
-	CmdPackageFlagRetries                 = "Number of retries to perform for Zarf operations like git/image pushes or Helm installs"
+	CmdPackageFlagRetries                 = "Number of retries to perform for Zarf operations like git/image pushes"
 
 	CmdPackageCreateShort = "Creates a Zarf package from a given directory or the current directory"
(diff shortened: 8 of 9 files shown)


src/api/v1alpha1/package.go
```diff
@@ -47,6 +47,10 @@ const (
 	ZarfPackageConfig ZarfPackageKind = "ZarfPackageConfig"
 	// APIVersion the api version of this package.
 	APIVersion string = "zarf.dev/v1alpha1"
+	// PackageLabel is the label used to identify this package.
```
Racer159:
```suggestion
	// PackageLabel is the label used to identify the owning package.
```
Just a suggestion to align with K8s wording https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

src/internal/packager/helm/chart.go
```diff
@@ -54,6 +54,10 @@ type InstallUpgradeOptions struct {
 	Timeout time.Duration
 	// Retries for the helm install/upgrade
 	Retries int
+	// Pkg is the zarf package being installed
+	Pkg *v1alpha1.ZarfPackage
+	// NamespaceOverride is the namespace to use for the chart
```
Racer159:
```suggestion
	// NamespaceOverride is the namespace override to use for the chart
```
zarfChart.namespace is the namespace for the chart (which would still have a value even if the namespace override was empty)

src/internal/packager/helm/chart.go
```diff
@@ -80,7 +84,7 @@ func InstallOrUpgradeChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, ch
 		return nil, "", fmt.Errorf("unable to initialize the K8s client: %w", err)
 	}
 
-	postRender, err := newRenderer(ctx, zarfChart, opts.AdoptExistingResources, opts.Cluster, opts.AirgapMode, opts.State, actionConfig, opts.VariableConfig)
+	postRender, err := newRenderer(ctx, zarfChart, opts.AdoptExistingResources, opts.Cluster, opts.AirgapMode, opts.State, actionConfig, opts.VariableConfig, opts.Pkg, opts.NamespaceOverride)
```
Racer159: so much of `opts` is passed into `newRenderer` does it make sense to pass in `opts` as a whole?

src/internal/packager/helm/post-render.go
```diff
@@ -117,6 +121,19 @@ func (r *renderer) adoptAndUpdateNamespaces(ctx context.Context) error {
 			}
 		}
 
+		// Add the package label to the namespace
+		if r.pkg != nil {
```
Racer159: Did you get past the issue here with init package labeling?

src/test/e2e/26_simple_packages_test.go
```diff
@@ -72,10 +75,60 @@ func TestManifests(t *testing.T) {
 
 	path := filepath.Join(tmpdir, fmt.Sprintf("zarf-package-manifests-%s-0.0.1.tar.zst", e2e.Arch))
 
-	// Deploy the package
-	stdOut, stdErr, err = e2e.Zarf(t, "package", "deploy", path, "--confirm")
+	// Deploy the package. Need to adopt existing resources to make sure that the label update for the package name
+	// applies to any previously created namespaces
+	stdOut, stdErr, err = e2e.Zarf(t, "package", "deploy", path, "--confirm", "--adopt-existing-resources")
+	require.NoError(t, err, stdOut, stdErr)
```
Racer159: we should change the behavior to add the label to namespaces without this flag IMO

src/internal/packager/helm/post-render.go
sgettys: Let me test that out again. I think we should assume that the renderer can only run when pkg is NOT nil?

Racer159: we should though IMO we could cover this with tests - nil checks aren't bad (they help prevent outright crashes) but we should ensure that the behavior works as expected when the correct inputs are given.

sgettys: The post renderer runs as part of `UpdateZarfRegistryValues` and `UpdateZarfAgentValues` which run outside the context of a package

src/internal/packager/helm/post-render.go
```diff
@@ -111,19 +114,26 @@ func (r *renderer) adoptAndUpdateNamespaces(ctx context.Context) error {
...
+		// If the namespace doesn't exist then create it. If it does exist and is already managed by Zarf then update the labels with
+		// the new package and namespace override labels.
 		if !existingNamespace {
 			// This is a new namespace, add it
 			_, err := c.Clientset.CoreV1().Namespaces().Create(ctx, namespace, metav1.CreateOptions{})
 			if err != nil {
 				return fmt.Errorf("unable to create the missing namespace %s", name)
 			}
-		} else if r.adoptExistingResources {
+		} else if zarfManaged { // If the namespace is already managed by Zarf then update the labels with the new package and namespace override labels.
```
Racer159:
```suggestion
		} else if zarfManaged || r.adoptExistingResources { // If the namespace is already managed by Zarf then update the labels with the new package and namespace override labels.
```
the adopt should still adopt everything

src/api/v1alpha1/package.go
```diff
@@ -47,6 +47,10 @@ const (
 	ZarfPackageConfig ZarfPackageKind = "ZarfPackageConfig"
 	// APIVersion the api version of this package.
 	APIVersion string = "zarf.dev/v1alpha1"
+	// PackageLabel is the label used to identify the owning of package.
+	PackageLabel string = "zarf.dev/package"
+	// NamespaceOverrideLabel is the label used to identify the namespace override.
+	NamespaceOverrideLabel string = "zarf.dev/namespace-override"
```
AustinAbro321: A better place to put these constants would be in this block https://github.com/zarf-dev/zarf/blob/main/src/pkg/cluster/cluster.go#L31-L38

src/internal/packager/helm/post-render.go
```diff
@@ -40,32 +40,38 @@ type renderer struct {
...
-	if c == nil {
+func newRenderer(ctx context.Context, chart v1alpha1.ZarfChart, actionConfig *action.Configuration, opts InstallUpgradeOptions) (*renderer, error) {
+	if opts.Cluster == nil {
 		return nil, fmt.Errorf("cluster required to run post renderer")
 	}
 	if actionConfig == nil {
 		return nil, fmt.Errorf("action configuration required to run post renderer")
 	}
-	if variableConfig == nil {
+	if opts.VariableConfig == nil {
```
AustinAbro321: I'd rather not have this function take opts. While it's semantically nicer, generally opts shouldn't be required. Also one of the goals with opts is to avoid breaking changes in function signatures, but since this is private we don't have to worry about that.

src/internal/packager/helm/post-render.go
```diff
@@ -111,19 +117,26 @@ func (r *renderer) adoptAndUpdateNamespaces(ctx context.Context) error {
...
 			if serverNamespace.Name == name {
 				existingNamespace = true
+				if serverNamespace.Labels[state.ZarfManagedByLabel] == "zarf" {
+					zarfManaged = true
+				}
+				break
 			}
 		}
-
+		namespace.Labels = r.setPackageLabels(namespace.Labels)
```
AustinAbro321: One implication that I'm thinking about here is that if two packages deploy resources to the same namespace then whichever deploy was more recent will take over and the label will change. Also, Zarf and Helm do not delete namespaces as well so these labels will stick around after the packages are removed from the cluster. Helm for instance doesn't add the `app.kubernetes.io/name` label to namespaces it creates. Still, these aren't necessarily a deal breaker to adding labels to namespaces as these side effects are relatively inconsequential, but I would like to evaluate it.

Could you provide more info about your use case @sgettys @Racer159. Is having this label on a namespace useful?

src/test/e2e/26_simple_packages_test.go
```diff
@@ -76,6 +79,55 @@ func TestManifests(t *testing.T) {
...
+
+	deploymentList := &appsv1.DeploymentList{}
+	err = json.Unmarshal([]byte(stdOut), deploymentList)
+	require.NoError(t, err)
+	require.Len(t, deploymentList.Items, 3, "expected 3 deployments")
+
+	// Wait for all the deployments to be ready
+	stdOut, stdErr, err = e2e.Kubectl(t, "wait", "deployment", "-l", "zarf.dev/package=manifests", "--for=condition=Available", "--all-namespaces", "--timeout=1m")
+	require.NoError(t, err, stdOut, stdErr)
```
AustinAbro321: So this is a bit interesting, initially I was going to say that this is unnecessary since Zarf waits for resources to be ready by default, but it actually is necessary because the podinfo deployment has an hpa which _usually_ doesn't start scaling until after the deployment is already considered ready by Zarf. Probably it would be fine, but I don't want the hpa to take too long and not get caught by this wait, or make a third pod and fail the test. I think it's simpler to just test what we want in the dos-games test above.

src/pkg/packager/deploy.go
```diff
@@ -290,7 +291,7 @@ func (d *deployer) deployInitComponent(ctx context.Context, pkgLayout *layout.Pa
 
 	// Before deploying the seed registry, start the injector
 	if isSeedRegistry {
-		err := d.c.StartInjection(ctx, pkgLayout.DirPath(), pkgLayout.GetImageDirPath(), component.Images, d.s.RegistryInfo.NodePort)
+		err := d.c.StartInjection(ctx, pkgLayout.DirPath(), pkgLayout.GetImageDirPath(), component.Images, d.s.RegistryInfo.NodePort, &pkgLayout.Pkg)
```
AustinAbro321: I'd lean towards skipping the labeling of the injector. I can see the argument either way on if it'd be accurate, on one hand there is an injector component in the init package, on the other hand these resources are created by Zarf itself. Given that the injector is running for ~15 seconds, I'd rather air on keeping our code simpler unless there is a use case for it.

AustinAbro321 (changes requested):
Thanks for making this change, I think it will be a positive for the project. Added a few nits and questions around scope

src/internal/packager/helm/post-render.go
Racer159: We don't actually need the namespaces - we were just trying to be complete here - I would be fine with just not labeling namespaces actually - the need we had was for tracking Pods / Pod Owner resources.

src/pkg/packager/deploy.go
Racer159: I asked for this to be included to more accurately see the pods in a package - the injector is normally only around for a little bit _but_ there are cases where Zarf init fails at the injector step and leaves the injector running - seeing the injector pod come back in a label selection query (i.e. if you have controlled K8s API access) can be a signal that this was the stage that Zarf init failed on.

src/internal/packager/helm/post-render.go
sgettys: I initially had it that way then went to opts because of the signature size and the fact that it was basically all of the "ops". I'm fine to revert back to explicit fields though

src/internal/packager/helm/post-render.go
sgettys: My thought around this was this label would be useful to at least denote the latest zarf package to operate against that namespace assuming multiple packages operate against it. We also discussed having a list of packages and appending instead of replacing but that could be confusing especially since you'd have to manage the label removal on package remove. In general I'm a fan of "labeling everything" including namespaces and this will at the bare minimum be a highly visible way to know if zarf has operated against a namespace or not

src/pkg/packager/deploy.go
sgettys: We discussed this and came to the decision to label even the injector just in case something failed during init. The philosophy of "zarf labels resources it creates" is what drove this in combination with debuggability on init failures

src/test/e2e/26_simple_packages_test.go
sgettys: I do at least like selecting multiple resources from the same package and since "package pods" is really what I'm getting at with this whole labeling thing IMO this is a good testcase. I can make it less restrictive on number of pods though to account for hpa!

src/internal/packager/helm/post-render.go
AustinAbro321: I'm still somewhat torn, I could see it being useful on occasion. While it'd be rare, the argument against adding the label to namespaces is to avoid a user doing something such as making a script where they grab the namespaces deployed by their package, but suddenly having that script break when another package is deployed to that namespace. I agree that adding and removing different packages from the label is not worth the complexity. 

I'm curious what the other @zarf-dev/maintainers think.

src/pkg/packager/deploy.go
AustinAbro321: Fair enough

src/internal/packager/helm/post-render.go
sgettys: Now that we have this functionality available it will be very easy to add this in for namespaces down the road if the use cases start popping up. I'll remove for now

Racer159: My 2 cents are that it is easy to add back if we need to in the future and can be left out as @sgettys mentioned - I do like completeness but if there are unexpected side effects that folks need to take into account it is worth holding off until we see more usecases - a list would be more complete but has its own side effects (what happens when a package is removed and the namespace is owned by two packages?).

src/test/e2e/26_simple_packages_test.go
AustinAbro321: Sounds good, but you should check for one or more replicas. A flake happened here because the check occurred before the hpa bumped up the replica count to two. https://github.com/zarf-dev/zarf/actions/runs/18012570063/job/51253428537?pr=4194

sgettys: Yeah I added that. Getting ready to push

src/internal/packager/helm/zarf.go
```diff
@@ -28,6 +28,11 @@ import (
 
 // UpdateZarfRegistryValues updates the Zarf registry deployment with the new state values
 func UpdateZarfRegistryValues(ctx context.Context, opts InstallUpgradeOptions) error {
+	pkg, err := opts.Cluster.GetDeployedPackage(ctx, "init")
```
AustinAbro321: To make this work with custom named init packages we could use cluster.GetDeployedZarfPackages and grab the first package with `kind: ZarfInitConfig`

src/internal/packager/helm/chart.go
```diff
@@ -54,6 +54,10 @@ type InstallUpgradeOptions struct {
 	Timeout time.Duration
 	// Retries for the helm install/upgrade
 	Retries int
+	// Pkg is the zarf package being installed
+	Pkg *v1alpha1.ZarfPackage
```
AustinAbro321: `.Metdata.Name` is the only field used from the package currently. Are you thinking that you might want other fields from the package in the future? Maybe a `string` or `v1alpha1.ZarfMetadata` would be more appropriate.

src/pkg/cluster/cluster.go
```diff
@@ -154,6 +159,8 @@ type InitStateOptions struct {
 	ArtifactServer state.ArtifactServerInfo
 	// StorageClass of the k8s cluster Zarf is initializing
 	StorageClass string
+	// Pkg is the Zarf package being initialized
+	Pkg *v1alpha1.ZarfPackage
```
AustinAbro321: This is no longer needed since you're not labeling the namespace anymore

src/internal/packager/helm/zarf.go
sgettys: Can we guarantee only a single ZarfInitConfig type?

src/internal/packager/helm/chart.go
sgettys: Possibly, more of a future proofing in case we want to do more package based labeling in the future. I don't have a specific use case for it now though so I don't mind narrowing the scope of that variable

sgettys: I guess we can use the package name and the namespace override label if it exists to look up the package later! So I can trim this down to be just the name

src/internal/packager/helm/zarf.go
AustinAbro321: I have seen before situations where there are multiple `ZarfInitConfig` packages deployed. The most complete thing to do would be check for a package with ZarfInitConfig and a component named `zarf-registry`. Same with the agent update, check for a package with `ZarfInitConfig` and a component named `zarf-agent`. 

Technically still possible for there to be two packages with those component names deployed, but I can't imagine a use case for it.

src/internal/packager/helm/post-render.go
```diff
@@ -40,21 +40,22 @@ type renderer struct {
...
-	}
+func newRenderer(ctx context.Context, chart v1alpha1.ZarfChart, adoptExistingResources bool, c *cluster.Cluster, airgapMode bool, s *state.State, actionConfig *action.Configuration, variableConfig *variables.VariableConfig, pkgName string, namespaceOverride string) (*renderer, error) {
 	if actionConfig == nil {
 		return nil, fmt.Errorf("action configuration required to run post renderer")
 	}
 	if variableConfig == nil {
 		return nil, fmt.Errorf("variable configuration required to run post renderer")
 	}
+	if pkgName == "" {
+		return nil, fmt.Errorf("package required to run post renderer")
```
AustinAbro321:
```suggestion
		return nil, fmt.Errorf("package name required to run post renderer")
```

src/internal/packager/helm/post-render.go
```diff
@@ -178,6 +183,25 @@ func (r *renderer) editHelmResources(ctx context.Context, resources []releaseuti
 		if err := yaml.Unmarshal([]byte(resource.Content), rawData); err != nil {
 			return fmt.Errorf("failed to unmarshal manifest: %w", err)
 		}
+		// If the object is empty, it's a blank resource, so we skip it. If the package is nil, we're not in a package and we don't want to add labels.
```
AustinAbro321:
```suggestion
		// If the object is empty, it's a blank resource, so we skip it. If the package name is empty we don't add labels.
```

src/pkg/cluster/injector.go
```diff
@@ -36,7 +36,7 @@ import (
 )
 
 // StartInjection initializes a Zarf injection into the cluster.
-func (c *Cluster) StartInjection(ctx context.Context, tmpDir, imagesDir string, injectorSeedSrcs []string, registryNodePort int) error {
+func (c *Cluster) StartInjection(ctx context.Context, tmpDir, imagesDir string, injectorSeedSrcs []string, registryNodePort int, pkg *v1alpha1.ZarfPackage) error {
```
AustinAbro321: Lets also use a string for package name here

src/internal/packager/helm/zarf.go
```diff
@@ -179,3 +197,16 @@ func UpdateZarfAgentValues(ctx context.Context, opts InstallUpgradeOptions) erro
 	}
 	return nil
 }
+
+func findInitPackageWithComponent(pkgs []state.DeployedPackage, componentName string) string {
+	for _, pkg := range pkgs {
+		if pkg.Data.Kind == "ZarfInitConfig" {
```
AustinAbro321:
```suggestion
		if pkg.Data.Kind == v1alpha1.ZarfInitConfig {
```

AustinAbro321 (approved):
LGTM, nice work

src/internal/packager/helm/post-render.go
```diff
@@ -114,9 +117,11 @@ func (r *renderer) adoptAndUpdateNamespaces(ctx context.Context) error {
 		for _, serverNamespace := range namespaceList.Items {
 			if serverNamespace.Name == name {
 				existingNamespace = true
+				break
```
brandtkeller, line 120: ++ good catch

PR #4206, opened 2025-09-25
fix: respect insecure flag in zarf tools registry prune|catalog|copy

AustinAbro321:
## Description

Currently the insecure flag is not respected in three of our crane commands. This could present a problem when the registry is spun up in proxy mode with a self signed CA as users may want to use the `--insecure` flag to get around the self signed CA. See #4162 

## Checklist before merging


brandtkeller (approved):
LGTM

PR #4208, opened 2025-09-25
feat: make the registry overrides deterministic and ordered

richluby:
## Description

Provides an ordered, structured mechanism to specify registry overrides. A helper function is provided.

For context, the current implementation does not allow overriding a repository within a registry in a reliable manner. One example is docker.io and docker.io/library. To map both of these to an internal repository, (eg. docker.example.com/third-party), the distinction between `/library` and `/unofficial-image` must be available.

By internally representing the overrides as an array, we maintain the order of the registry overrides. This allows a user to leverage a longest-prefix-first ordering to override registry repositories in a predictable manner. Only the first prefix is respected instead of choosing the last entry in the array.

## Shortfalls

- The `types.RegistryOverride` type is not propagated to the `v1alpha1.ZarfBuildData`. This means we still lose the ordering when crossing this API boundary (eg. unpacking a `zarf` package). I can make that update if desired. I did not do so in an effort to minimize changes.

## Related Issue

Fixes https://github.com/zarf-dev/zarf/issues/4207

## Checklist before merging

### DIFF
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -154,23 +154,23 @@ func parseRegistryOverrides(overrides []string) ([]types.RegistryOverride, error
 	result := make([]types.RegistryOverride, len(overrides))
 	for i, mapping := range overrides {
 		source, override, found := strings.Cut(mapping, "=")
-        if !found {
-            return nil, fmt.Errorf("registry override missing '=': %s", mapping)
-        }
-        
-        if source == "" {
-            return nil, fmt.Errorf("registry override missing source: %s", mapping)
-        }
-        
-        if override == "" {
-            return nil, fmt.Errorf("registry override missing value: %s", mapping)
-        }
-        
-        if index := slices.IndexFunc(result, func(existing types.RegistryOverride) bool {
-            return existing.Source == source
-        }); index >= 0 {
-            return nil, fmt.Errorf("registry override has duplicate source: existing index %d, new index %d, source %s", index, i, source)
-        }
+		if !found {
+			return nil, fmt.Errorf("registry override missing '=': %s", mapping)
+		}
+
+		if source == "" {
+			return nil, fmt.Errorf("registry override missing source: %s", mapping)
+		}
+
+		if override == "" {
+			return nil, fmt.Errorf("registry override missing value: %s", mapping)
+		}
+
+		if index := slices.IndexFunc(result, func(existing types.RegistryOverride) bool {
+			return existing.Source == source
+		}); index >= 0 {
+			return nil, fmt.Errorf("registry override has duplicate source: existing index %d, new index %d, source %s", index, i, source)
+		}
 
 		result[i].Source = source
 		result[i].Override = override


brandtkeller:
Hey @richluby! thanks for the PR. I agree with changing this behavior but need to dive into the implications a bit more to determine if it is a breaking change across the configuration lifecycle. 

Would you be able to add some tests to validate the expected behaviors of `parseRegistryOverrides`?  especially with regards to ordering. 

`make docs-and-schema` should update documentation accordingly.

richluby:
> Hey @richluby! thanks for the PR. I agree with changing this behavior but need to dive into the implications a bit more to determine if it is a breaking change across the configuration lifecycle.

Sounds good to me!
 
> Would you be able to add some tests to validate the expected behaviors of `parseRegistryOverrides`? especially with regards to ordering.
> 
> `make docs-and-schema` should update documentation accordingly.

I've added some tests around parsing, and run the docs generator.

site/src/content/docs/commands/zarf_tools_helm.md
```diff
@@ -48,6 +48,9 @@ Subset of the Helm CLI that includes the repo and dependency commands for managi
 
 * [zarf tools](/commands/zarf_tools/)	 - Collection of additional tools to make airgap easier
 * [zarf tools helm dependency](/commands/zarf_tools_helm_dependency/)	 - manage a chart's dependencies
+* [zarf tools helm diff](/commands/zarf_tools_helm_diff/)	 - Preview helm upgrade changes as a diff
```
brandtkeller: Likely an oddity of the make command - curious why these updates to `zarf tools helm` are included.

src/cmd/package.go
```diff
@@ -143,6 +144,23 @@ func newPackageCreateCommand(v *viper.Viper) *cobra.Command {
...
 
+// Converts registry overrides to a structured type.
+// Input is of the following form:
+// []string{"docker.io/library=docker.example.com", "docker.io=docker.example.com"}
+func parseRegistryOverrides(overrides []string) ([]types.RegistryOverride, error) {
+	result := make([]types.RegistryOverride, len(overrides))
+	for i, override := range overrides {
+		source, override, found := strings.Cut(override, "=")
+		if !found {
+			return nil, fmt.Errorf("invalid override: missing '=': %s", override)
```
brandtkeller: I believe we can be more explicit in this error message to really highlight the point of error
```suggestion
			return nil, fmt.Errorf("invalid registry override: missing '=': %s", override)
```

src/cmd/package.go
```diff
@@ -143,6 +144,23 @@ func newPackageCreateCommand(v *viper.Viper) *cobra.Command {
 	return cmd
 }
 
+// Converts registry overrides to a structured type.
+// Input is of the following form:
+// []string{"docker.io/library=docker.example.com", "docker.io=docker.example.com"}
+func parseRegistryOverrides(overrides []string) ([]types.RegistryOverride, error) {
```
brandtkeller, line 153: I think there are a few validation gaps we can add some checking for in this function. Particularly empty strings that would evaluate `Source` or `Override` to be empty strings.

src/cmd/package.go
```diff
@@ -143,6 +144,23 @@ func newPackageCreateCommand(v *viper.Viper) *cobra.Command {
...
+	result := make([]types.RegistryOverride, len(overrides))
+	for i, override := range overrides {
+		source, override, found := strings.Cut(override, "=")
```
brandtkeller: this `override` variable returned from `Cut()` shadows the original override loop variable. I would delineate these from eachother

src/internal/packager/images/pull.go
```diff
@@ -76,11 +76,16 @@ func Pull(ctx context.Context, cfg PullConfig) (map[transform.Image]ocispec.Mani
 	}
 
 	imagesWithOverride := []imageWithOverride{}
+	// Iterate over all images, marking each one as overridden.
 	for _, img := range cfg.ImageList {
 		overriddenImage := img
-		for k, v := range cfg.RegistryOverrides {
-			if strings.HasPrefix(img.Reference, k) {
-				overriddenImage.Reference = strings.Replace(img.Reference, k, v, 1)
+		for _, v := range cfg.RegistryOverrides {
```
brandtkeller, line 82: If I understand this correctly - this indicates that longest prefix will win _when_ the order has longer prefixes first yes?

My gut is leaning toward removing that constraint and sorting the overrides such that we can guarantee longest match.

site/src/content/docs/commands/zarf_tools_helm.md
richluby: I wondered the same. I'm happy to remove them if that doesn't affect the check job status.

src/internal/packager/images/pull.go
richluby: I've added the sort to the parse function. I figured there it would be most consistent throughout execution. Doing so also allows programmatic callers to set ordering as necessary if the auto-sort doesn't meet an edge case we don't see from here.

One additional test case ensures multiple prefix-based overrides don't interfere with each other.

src/cmd/package.go
richluby: Good catch. With the additional error cases, it clarified the overall errors a bit as well.

src/cmd/package.go
richluby: I've added both edge cases as a reported error.

src/cmd/package.go
richluby: Agreed. Added.

site/src/content/docs/commands/zarf_tools_helm.md
brandtkeller: let's give it a try - I attempted generation of docs from this source and don't see the additional helm docs - curious as to how CI will respond.

brandtkeller: ah - looking at the commands I imagine you might have some local helm plugins installed that are being picked up. You should be able to remove the docs related to helm from this PR and the project can think more about how to ignore those.

richluby: Done!

src/cmd/package.go
```diff
@@ -143,6 +144,35 @@ func newPackageCreateCommand(v *viper.Viper) *cobra.Command {
 	return cmd
 }
 
+// Converts registry overrides to a structured type.
+// The result will be sorted in descending order.
+// Descending order guarantees the longest prefix will be sorted toward the beginning.
+//
+// Input is of the following form:
+// []string{"docker.io/library=docker.example.com", "docker.io=docker.example.com"}
+func parseRegistryOverrides(overrides []string) ([]types.RegistryOverride, error) {
```
brandtkeller, line 153: The only other item that stands out to me is guarantees on precedence. Do we know how we intend to handle duplicate sources (should they exist)?

Additionally considering if we should update the language for the command (currently states it is a `map`)

richluby: > The only other item that stands out to me is guarantees on precedence. Do we know how we intend to handle duplicate sources (should they exist)?

For precedence, I don't think it makes sense to have something like the following:

```
- docker.io/library=docker.example.com/internal
- docker.io/library=docker.example.com/third-party
```

Unless the user's desire there is to mirror the image to both of the internal repositories? Or as a fallback like, "If `internal` fails, then try `third-party`." 

In the first case, I think it only affects unpacking. In the second case, I would have to switch to `slices.StableSortFunc(...)` so that I can keep the appropriate ordering as specified by the user, and update the pull logic quite a bit to handle retries with different images.

In lieu of either of these options, I've added an error check to reject duplicate sources.

> Additionally considering if we should update the language for the command (currently states it is a map)

I've updated the language to use `mapping` instead of `map` to distinguish it some from the go `map` type. I think it is still a source->override map from the perspective of someone reading the docs even if in-memory its an array of structs.

src/cmd/package.go
```diff
@@ -143,6 +144,40 @@ func newPackageCreateCommand(v *viper.Viper) *cobra.Command {
...
+			return nil, fmt.Errorf("invalid registry override: missing '=': %s", mapping)
+		} else if source == "" {
+			return nil, fmt.Errorf("registry override must have a source: %s", mapping)
+		} else if override == "" {
+			return nil, fmt.Errorf("registry override must have a value: %s", mapping)
+		} else if index := slices.IndexFunc(result, func(existing types.RegistryOverride) bool {
+			return existing.Source == source
+		}); index >= 0 {
+			return nil, fmt.Errorf("registry source is duplicated: existing index: %d erroring index: %d source: %s", index, i, source)
+		}
```
brandtkeller, line 173: nit: Just to clean this up and align to updates we're making to the project - flattening and normalizing the error messages would improve the readability. 
```suggestion
        if !found {
            return nil, fmt.Errorf("registry override missing '=': %s", mapping)
        }
        
        if source == "" {
            return nil, fmt.Errorf("registry override missing source: %s", mapping)
        }
        
        if override == "" {
            return nil, fmt.Errorf("registry override missing value: %s", mapping)
        }
        
        if index := slices.IndexFunc(result, func(existing types.RegistryOverride) bool {
            return existing.Source == source
        }); index >= 0 {
            return nil, fmt.Errorf("registry override has duplicate source: existing index %d, new index %d, source %s", index, i, source)
        }
```

src/cmd/package_test.go
```diff
@@ -317,3 +318,126 @@ func TestPackageInspectValuesFiles(t *testing.T) {
 		})
 	}
 }
+
+// TestParseRegistryOverrides ensures that ordering is maintained for registry overrides.
+func TestParseRegistryOverrides(t *testing.T) {
```
brandtkeller, line 323: Could you add test cases that provide coverage for the error validation code paths?

brandtkeller (changes requested):
One additional go-back - appreciate the patience as I should have called this out earlier.

src/cmd/package_test.go
richluby: Of course. Added!

brandtkeller (approved):
I believe this is in a great state and enhances the override experience. Tagging @zarf-dev/maintainers for a second review.

PR #4211, opened 2025-09-26
feat(prune): add support for removing charts not tracked in state

### LINKED ISSUE #2992: Request: Support  `--prune` feature when execute `zarf package deploy` to delete needless helm chart or manifest
No.

Day 1: I defined my application's Helm charts and images using zarf.yaml, then packaged them into a .zst archive. I deployed it to my cluster in an air-gapped environment, and it worked very well
like this:

```yaml
kind: ZarfPackageConfig
metadata:
  name: example-app
  description: Deploy example application
  version: 0.1.1
  architecture: amd64

components:
  - name: app-gateway
    required: true
    charts:
      - name: app-gateway
        version: 1.0.2
        namespace: app-gateway
        url: https://xxx.com/artifactory/helm-dev/
        releaseName: app-gateway
        valuesFiles:
          - helm-values/values-app-gateway.yaml
      - name: httpd
        version: 10.0.1
        namespace: httpd
        url: https://xxx.com/artifactory/helm-dev/
        releaseName: httpd
        valuesFiles:
          - helm-values/values-httpd.yaml
    images:
      - xxx.com/app-gateway:v1.1
      - httpd:alpine3.18
```

Day 2: My application was updated, but only some images and manifests were changed. I repackaged my Zarf package, from 0.1.1 to 0.1.2, and executed `zarf package deploy` to deploy it to the cluster. It worked very well.

Day 3: My application was updated again, with more changes this time. I needed to remove the `httpd` Helm chart, so I created a new Zarf package, from 0.1.2 to 0.2.0, and executed `zarf package deploy` again. However, the resources associated with the `httpd` Helm chart were not deleted.

If this was the manifest file, I think it would be the same

Therefore, I wonder if it's possible to add a `--prune` option to `zarf package deploy` to support the deletion of unnecessary resources from the new version of the package.

This may also be integrated with the `zarf tools registry prune` to automatically upgrade applications and clean up unwanted resources

Of course, I can define different Helm charts and manifests as separate components and then use `zarf package remove --component` to delete specific components.

However, this is somewhat cumbersome. It would be more convenient if `zarf.yaml` could provide a way to declare certain components to be removed before deployment. This way, `zarf package deploy` could automatically remove those components, making the process more convenient.

brandtkeller:
## Description

Adds support for a `zarf package prune` command. This allows for the explicit removal of a component chart by name OR it can look for all pending/orphaned resources and provide a list. Requires confirmation.

## Related Issue

Fixes #2992

## Checklist before merging


src/cmd/package.go
```diff
@@ -1317,6 +1318,146 @@ func (o *packageRemoveOptions) run(cmd *cobra.Command, args []string) error {
...
+
+	return cmd
+}
+
+func (o *packagePruneOptions) preRun(_ *cobra.Command, _ []string) {
+	// If --insecure was provided, set --skip-signature-validation to match
+	if config.CommonOptions.Insecure {
+		o.skipSignatureValidation = true
+	}
+}
```
AustinAbro321: We have as a backwards compatibility shim for existing commands, I don't think we should introduce it for new commands. It's generally set to be removed in January https://github.com/zarf-dev/zarf/issues/3420

src/cmd/package.go
```diff
@@ -1317,6 +1318,146 @@ func (o *packageRemoveOptions) run(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+	filter := filters.Combine(
+		filters.ByLocalOS(runtime.GOOS),
+	)
+	cachePath, err := getCachePath(ctx)
+	if err != nil {
+		return err
+	}
+	c, _ := cluster.New(ctx) //nolint:errcheck
```
AustinAbro321: Should we avoid the error check here? It's not always needed for the package load, but prune will expect that a cluster is available later on, so this could be a way to fail faster if there's no cluster.

src/pkg/packager/prune.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/zarf-dev/zarf/src/pkg/state"
+)
+
+// PruneOptions are the options for Prune
+type PruneOptions struct {
+	Cluster           *cluster.Cluster
+	Component         string
+	Chart             string
+	Timeout           time.Duration
+	NamespaceOverride string
```
AustinAbro321: namespace override is set, but never used

src/cmd/package.go
```diff
@@ -1317,6 +1318,146 @@ func (o *packageRemoveOptions) run(cmd *cobra.Command, args []string) error {
...
+	pruneOpt := packager.PruneOptions{
+		Cluster:           c,
+		Timeout:           config.ZarfDefaultTimeout,
+		NamespaceOverride: o.namespaceOverride,
+		Component:         o.component,
+		Chart:             o.chart,
+	}
+
+	// First get all applicable orphaned or pending resources based on the component filter
+	pruneState, err := packager.GetPruneableCharts(deployedPackage, pruneOpt)
```
AustinAbro321: nit: all the sources I've found spell it "prunable"

src/pkg/packager/prune.go
```diff
@@ -0,0 +1,68 @@
...
+
+// PruneStateResult is the result of GetPruneableCharts
+type PruneStateResult struct {
+	PruneableCharts map[string][]state.InstalledChart
+}
+
+// GetPruneableCharts returns a list of installed charts that can be pruned per component.
+// This function serves as a coordination layer that delegates the filtering logic to the
+// DeployedPackage and wraps the result in PruneStateResult for the packager layer.
+func GetPruneableCharts(deployedPackage *state.DeployedPackage, opts PruneOptions) (PruneStateResult, error) {
```
AustinAbro321: IMO, we should delete this function. I think it's reasonable to expect SDK users to call functions in the sdk outside of packager such as `deployedPackage.GetPrunableCharts()`

src/pkg/state/state.go
```diff
@@ -399,6 +405,76 @@ func (d *DeployedPackage) GetSecretName() string {
 	return fmt.Sprintf("%s-%s", "zarf-package", d.Name)
 }
 
+// GetPruneableCharts returns a map of component names to lists of charts that can be pruned,
+// filtered by the provided component and chart names. If componentFilter is empty, all components
+// are searched. If chartFilter is empty, all charts are searched.
+func (d *DeployedPackage) GetPruneableCharts(componentFilter, chartFilter string) (map[string][]InstalledChart, error) {
```
AustinAbro321: Charts are set to orphaned only work when a chart is removed from a component, but not when a component is removed entirely. I believe this could confuse some users. This could also encourage users to add all their charts to a single component so they can use prune in the future if need be.

src/pkg/packager/prune.go
brandtkeller: Good call - I moved this up to the CLI.

src/cmd/package.go
brandtkeller: agree!

src/cmd/package.go
brandtkeller: agree

src/cmd/package.go
brandtkeller: that is great feedback

src/pkg/packager/prune.go
brandtkeller: I think that is fair. I kept the `PruneCharts()` function as it serves to integrate cluster/state/helm but open to thoughts.

src/pkg/state/state.go
brandtkeller: Could you elaborate more? is this a concern or a required change? 

Charts should only be able to be orphaned during deploys. Removes should remove all charts - active or orphaned. I do see room for confusion among the two commands but less of a distinct path forward.

AustinAbro321: Ah there's a misunderstanding. I mean when a component is removed from one package to another, rather than it being removed with `zarf package remove`. 

For example, if we started with components like so
```yaml
components:
  - name: test-component
    required: true
    charts:
      - name: chart-to-keep
        version: 1.0.0
        namespace: prune-test
        localPath: chart-to-keep
  - name: test-component-2
    charts:
      - name: chart-to-remove
        version: 1.0.0
        namespace: prune-test
        localPath: chart-to-remove
```

then in the next version of the package removed test-component-2. Functionally, I think as a user I'd expect prune to behave the same in both situations, but currently it will only track the chart if it was originally in a component that still exists. In my experience, more often than not a component has a single chart.

brandtkeller: Ah interesting - correct me if I understood wrong - not only does Zarf orphan InstalledCharts from state - it also orphans components from state as well?

AustinAbro321: Yea, when Zarf records a package deployment it overwrites the existing secret and only the components that were in the latest deployment are added to state.

PR #4220, opened 2025-09-30
fix(injector): improve node resources availability for injector

### LINKED ISSUE #4218: Zarf Injector not calculating node resources correctly
### Environment

zarf version: 0.62.0

### Issue
Currently the Zarf injector does a check to verify there is enough space on the node before deploying the injector pod. The issue in the current logic is that it utilizes "Allocatable CPU" to verify if there is enough space. Allocatable CPU doesn't include the pods running on the node. Instead it is the actual resources minus kube-reserved, system-reserved, and eviction-threshold. This means that the Zarf injector can be scheduled on a node without sufficient resources leading to a failed deployment

[Node Allocatable docs](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#node-allocatable)

Current logic:
```
		if nodeDetails.Status.Allocatable.Cpu().Cmp(minNodeCPU) < 0 ||
		   nodeDetails.Status.Allocatable.Memory().Cmp(minNodeMemory) < 0 {
			continue findImages
		}
```


```
│ Labels:           app=zarf-injector                                                                                                                                                                                                                                              │
│                   zarf.dev/agent=ignore                                                                                                                                                                                                                                          │
│ Annotations:      <none>                                                                                                                                                                                                                                                         │
│ Status:           Failed                                                                                                                                                                                                                                                         │
│ Reason:           OutOfcpu                                                                                                                                                                                                                                                       │
│ Message:          Pod was rejected: Node didn't have enough resource: cpu, requested: 500, used: 3430, capacity: 3920                                                                                                                                                            │
│ SeccompProfile:   RuntimeDefault        
```
### Severity/Priority
high

brandtkeller:
## Description

This introduces an additional iteration through node pods in order to get an accurate calculation for available resources. This provides a more accurate representation of resources available on a node - as the injector skips the scheduler - which would normally provide this information. 

## Related Issue

Fixes #4218

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/injector.go
+++ b/src/pkg/cluster/injector.go
@@ -344,7 +344,6 @@ func (c *Cluster) getInjectorImageAndNode(ctx context.Context, resReq *v1ac.Reso
 		l.Debug("no suitable image found on node", "node", node.Name)
 	}
 
-	l.Debug("failed to find suitable injector image or node")
 	return "", "", fmt.Errorf("no suitable injector image or node exists")
 }
 
--- a/src/pkg/cluster/injector.go
+++ b/src/pkg/cluster/injector.go
@@ -262,37 +262,66 @@ func (c *Cluster) getInjectorImageAndNode(ctx context.Context, resReq *v1ac.Reso
 	if err != nil {
 		return "", "", err
 	}
+
 	for _, pod := range podList.Items {
 		nodeDetails, err := c.Clientset.CoreV1().Nodes().Get(ctx, pod.Spec.NodeName, metav1.GetOptions{})
 		if err != nil {
 			return "", "", err
 		}
-		if nodeDetails.Status.Allocatable.Cpu().Cmp(*resReq.Requests.Cpu()) < 0 ||
-			nodeDetails.Status.Allocatable.Memory().Cmp(*resReq.Requests.Memory()) < 0 {
+
+		var usedCPU, usedMem int64
+		for _, npod := range podList.Items {
+			if npod.Spec.NodeName != nodeDetails.Name {
+				continue
+			}
+			// skip completed/failed pods
+			if npod.Status.Phase == corev1.PodSucceeded || npod.Status.Phase == corev1.PodFailed {
+				continue
+			}
+			for _, ctn := range npod.Spec.Containers {
+				usedCPU += ctn.Resources.Requests.Cpu().MilliValue()
+				usedMem += ctn.Resources.Requests.Memory().Value()
+			}
+			for _, ctn := range npod.Spec.InitContainers {
+				usedCPU += ctn.Resources.Requests.Cpu().MilliValue()
+				usedMem += ctn.Resources.Requests.Memory().Value()
+			}
+		}
+
+		availCPU := nodeDetails.Status.Allocatable.Cpu().MilliValue() - usedCPU
+		availMem := nodeDetails.Status.Allocatable.Memory().Value() - usedMem
+
+		reqCPU := resReq.Requests.Cpu().MilliValue()
+		reqMem := resReq.Requests.Memory().Value()
+
+		if availCPU < reqCPU || availMem < reqMem {
+			// not enough free capacity
 			continue
 		}
+
+		// Respect taints
 		if hasBlockingTaints(nodeDetails.Spec.Taints) {
 			continue
 		}
+
+		// Look for a non-Zarf image from containers/init/ephemeral containers
 		for _, container := range pod.Spec.Containers {
-			if zarfImageRegex.MatchString(container.Image) {
-				continue
+			if !zarfImageRegex.MatchString(container.Image) {
+				return container.Image, pod.Spec.NodeName, nil
 			}
-			return container.Image, pod.Spec.NodeName, nil
 		}
 		for _, container := range pod.Spec.InitContainers {
-			if zarfImageRegex.MatchString(container.Image) {
-				continue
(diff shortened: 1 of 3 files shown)


src/pkg/cluster/injector.go
```diff
@@ -262,37 +263,69 @@ func (c *Cluster) getInjectorImageAndNode(ctx context.Context, resReq *v1ac.Reso
...
 		nodeDetails, err := c.Clientset.CoreV1().Nodes().Get(ctx, pod.Spec.NodeName, metav1.GetOptions{})
 		if err != nil {
 			return "", "", err
 		}
-		if nodeDetails.Status.Allocatable.Cpu().Cmp(*resReq.Requests.Cpu()) < 0 ||
-			nodeDetails.Status.Allocatable.Memory().Cmp(*resReq.Requests.Memory()) < 0 {
+
+		l.Debug("calculating available resources", "node", nodeDetails.Name)
+		var usedCPU, usedMem int64
+		for _, npod := range podList.Items {
```
AustinAbro321: We can't use this podList, we have to get a new podlist from pods on the node. Or rather start by going through nodes and getting the pods on them.

src/pkg/cluster/injector.go
```diff
@@ -262,37 +263,69 @@ func (c *Cluster) getInjectorImageAndNode(ctx context.Context, resReq *v1ac.Reso
...
+				usedMem += ctn.Resources.Requests.Memory().Value()
+			}
+			for _, ctn := range npod.Spec.InitContainers {
+				usedCPU += ctn.Resources.Requests.Cpu().MilliValue()
+				usedMem += ctn.Resources.Requests.Memory().Value()
+			}
+		}
+
+		availCPU := nodeDetails.Status.Allocatable.Cpu().MilliValue() - usedCPU
+		availMem := nodeDetails.Status.Allocatable.Memory().Value() - usedMem
```
AustinAbro321: I'd like to see us use the builtin functions for these types. For example, we could do this add have access to Add(), Sub(), and Cmp()

```go
		availableCPU := nodeDetails.Status.Allocatable.Cpu().DeepCopy()
```

src/pkg/cluster/injector.go
```diff
@@ -262,37 +263,69 @@ func (c *Cluster) getInjectorImageAndNode(ctx context.Context, resReq *v1ac.Reso
...
+				continue
+			}
+			for _, ctn := range npod.Spec.Containers {
+				usedCPU += ctn.Resources.Requests.Cpu().MilliValue()
+				usedMem += ctn.Resources.Requests.Memory().Value()
+			}
+			for _, ctn := range npod.Spec.InitContainers {
+				usedCPU += ctn.Resources.Requests.Cpu().MilliValue()
+				usedMem += ctn.Resources.Requests.Memory().Value()
+			}
```
AustinAbro321: Should we check emphemeral containers as well?

src/pkg/cluster/injector.go
```diff
@@ -250,6 +250,7 @@ func (c *Cluster) createPayloadConfigMaps(ctx context.Context, tmpDir, imagesDir
 
 // getImagesAndNodesForInjection checks for images on schedulable nodes within a cluster.
 func (c *Cluster) getInjectorImageAndNode(ctx context.Context, resReq *v1ac.ResourceRequirementsApplyConfiguration) (string, string, error) {
```
AustinAbro321, line 253: With the extra complexity this introduces, I think it'd be worth creating a unit test for this function

src/pkg/cluster/injector.go
brandtkeller: I believe we accomplish the task by isolating by a node in the following lines - but there is a tradeoff here for efficiency. 

I did update the logic to optimize:
- get all nodes and all pods
- map pods by node

Iterate through nodes and process pods by node for available resources - continue early when possible.

src/pkg/cluster/injector.go
brandtkeller: good catch

src/pkg/cluster/injector.go
brandtkeller: Ah! great catch - I passed over those on `resource.Quantity` type initially - should be using them properly now.

src/pkg/cluster/injector.go
brandtkeller: Add multiple tests for successful and failing paths

src/pkg/cluster/injector.go
```diff
@@ -250,49 +250,102 @@ func (c *Cluster) createPayloadConfigMaps(ctx context.Context, tmpDir, imagesDir
...
+
+		if candidateImage != "" {
+			l.Debug("selected image for injector", "node", node.Name, "image", candidateImage)
+			return candidateImage, node.Name, nil
+		}
+
+		l.Debug("no suitable image found on node", "node", node.Name)
 	}
+
+	l.Debug("failed to find suitable injector image or node")
```
AustinAbro321: nit this is mostly a repeat of the error message

AustinAbro321 (approved):
LGTM

src/pkg/cluster/injector.go
brandtkeller: Agreed! Same wavelength - this was removed in [this commit](https://github.com/zarf-dev/zarf/pull/4220/commits/8ada1fc90b918dfa57193a6627d02eca24876792)

PR #4227, opened 2025-10-02
feat!: push to injector instead of adding configmaps to the cluster

AustinAbro321:
## Description

I consider this blocked for now as it conflicts with the implementation of [0033-registry-proxy](https://github.com/zarf-dev/proposals/tree/main/0033-registry-proxy)

This will not be officially ready for review until https://github.com/zarf-dev/zarf-injector/pull/40 is merged and released. 

Now that the Zarf injector no longer puts together configmaps and ensures all of those configmaps together have a certain sha creators of custom init package should be encouraged to sha their registry image. Someone with access to port-forward services in the cluster can push to the injector registry and override an image. 

If users try to `zarf package deploy` an older version of the init package with an init package generated after this PR is created then it will fail since the injector no longer has the configmap logic. I believe this is okay as users should use `zarf init` to deploy the init package. 

This also moves the injector to internal as it's very specific to the init deploy lifecycle.

Having a mechanism to push makes it easier to rely on additional images from the injector such as a proxy image, without making init take longer.

## Checklist before merging


AustinAbro321:
Closing this as we are pausing work on https://github.com/zarf-dev/zarf-injector/issues/41, as it currently conflicts with https://github.com/zarf-dev/proposals/tree/main/0033-registry-proxy

PR #4234, opened 2025-10-06
fix: cleanup temp directory on pull and differential package loads

### LINKED ISSUE #3705: `TMPDIR` not properly cleaned out after multiple runs/pulls/misc

The `$TMPDIR` get filled with various staging parts of zarf that do not get cleaned up properly.

### Additional context

```shell
$ ls $TMPDIR
test.yml        zarf-431492100  zarf-897738423  zarf-1091906704  zarf-1203438468  zarf-1502900713  zarf-1809210350  zarf-1939208001  zarf-2521010492  zarf-2542684518  zarf-2823046111  zarf-3739878045  zarf-3863847754  zarf-4268330893
zarf-183051952  zarf-720355854  zarf-974994221  zarf-1106273493  zarf-1241826381  zarf-1734318326  zarf-1893956948  zarf-2450909861  zarf-2537174603  zarf-2767991230  zarf-3312580097  zarf-3784231634  zarf-4225363707
```

```shell
$ du -sh $TMPDIR
63G     /tmp/play
```

AustinAbro321:
## Description

When `zarf package pull` is run or differential package is used we currently are not cleaning up the package we place in the temp directory. This PR corrects that 

## Related Issue

Fixes #3705

## Checklist before merging

### DIFF
--- a/src/pkg/packager/create.go
+++ b/src/pkg/packager/create.go
@@ -66,9 +66,9 @@ func Create(ctx context.Context, packagePath string, output string, opts CreateO
 		if err != nil {
 			return "", fmt.Errorf("failed to load differential package: %w", err)
 		}
-		defer func() {
-			err = errors.Join(err, pkgLayout.Cleanup())
-		}()
+		if err := pkgLayout.Cleanup(); err != nil {
+			return "", err
+		}
 		differentialPkg = pkgLayout.Pkg
 	}
 
--- a/src/pkg/packager/pull.go
+++ b/src/pkg/packager/pull.go
@@ -81,9 +81,9 @@ func Pull(ctx context.Context, source, destination string, opts PullOptions) (_ 
 	if err != nil {
 		return "", err
 	}
-	defer func() {
-		err = errors.Join(err, pkgLayout.Cleanup())
-	}()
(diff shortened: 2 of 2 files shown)


src/pkg/packager/create.go
```diff
@@ -66,6 +66,9 @@ func Create(ctx context.Context, packagePath string, output string, opts CreateO
 		if err != nil {
 			return "", fmt.Errorf("failed to load differential package: %w", err)
 		}
+		defer func() {
```
brandtkeller: observation: does this `defer` offer us any benefit over a direct `pkgLayout.Cleanup()` execution? 

One thought being that - in performing all of the cleanup at the end - we potentially allow the core logic to execute before potentially interrupting with any errors?

brandtkeller (approved):
One passing thought - otherwise LGTM

src/pkg/packager/create.go
AustinAbro321: Good thinking. On create we don't need any of the files since we only need the pkg spec, and during pull we are outputting the package so we can cleanup right away there too

brandtkeller (approved):
LGTM

PR #4236, opened 2025-10-06
feat(values): Add --set-values and --set-variables flags

### LINKED ISSUE #4224: Add --set-value flag for granular values key overrides
Users are accustomed to being able to set individual Zarf variables or Helm values directly. We should have parity, probably with a --set-value key that can be passed in multiple times, or with comma separated values.

Another thing to consider is that we'll eventually want the `--set` flag to provide this functionality for Values. Right now, --set sets Variables, so as a part of implementing this, we should add `--set-variable` and alias the existing `--set` flag to this. Eventually we'll swap from one to the other, either at Values GA or sometime after.

mkcp:
## Description
This PR adds the `--set-values` flag with e2e coverage and examples documentation and adds `--set-variables` which is an alias for the current behavior of `--set`. The reason for --set-variables is to have logical parity with the new flag, and swap over `--set` to `--set-values` at some point in the future. This allows us to already have a deprecation path ready for users, and to ensure the flags are self-documenting in the mean time.

Do not merge this PR until #4136 is merged into main.

## Related Issue
Fixes #4224 

## Checklist before merging

### DIFF
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -252,7 +252,10 @@ func (o *packageDeployOptions) run(cmd *cobra.Command, args []string) (err error
 
 	// Merge setValues CLI flags
 	o.setValues = helpers.TransformAndMergeMap(
-		v.GetStringMapString(VPkgDeploySetValues), o.setValues, nil)
+		v.GetStringMapString(VPkgDeploySetValues),
+		o.setValues,
+		func(s string) string { return s },
+	)
 
 	// Load files supplied by --values / -v or a user's zarf-config.{yaml,toml}
 	// REVIEW: Should we also load valuesFiles supplied via URL on the CLI?


site/src/content/docs/commands/zarf_dev_deploy.md
```diff
@@ -23,8 +23,8 @@ zarf dev deploy [flags]
 ```
       --adopt-existing-resources           Adopts any pre-existing K8s resources into the Helm charts managed by Zarf. ONLY use when you have existing deployments you want Zarf to takeover.
       --components string                  Comma-separated list of components to deploy.  Adding this flag will skip the prompts for selected components.  Globbing component names with '*' and deselecting 'default' components with a leading '-' are also supported.
-      --create-set stringToString          Specify package variables to set on the command line (KEY=value) (default [])
-      --deploy-set stringToString          Specify deployment variables to set on the command line (KEY=value) (default [])
+      --create-set stringToString          [alias for --set-variables] Specify package variables to set on the command line (KEY=value) (default [])
+      --deploy-set stringToString          [alias for --set-variables] Specify deployment variables to set on the command line (KEY=value) (default [])
```
mkcp, line 27: This was an unexpected result from changing the other flags. Need to look more into how dev deploy is generated and see about `--create-set-variables` `--deploy-set-variables` and `--deploy-set-values`

AustinAbro321:
This was completed by #4466

PR #4245, opened 2025-10-08
fix(create): allow multiple overrides from environment variable

### LINKED ISSUE #4241: Zarf Package Create Registry Override Reduced Functionality
### Environment
Device and OS: Ubuntu Linux
App version: `v0.63.0`
Kubernetes distro being used: N/A package create issue
Other:

### Steps to reproduce
1. Use the environment variable approach with a JSON string for `--registry-override` to override registries during zarf package create. i.e. `export ZARF_PACKAGE_CREATE_REGISTRY_OVERRIDE={"registry1":"registry1-proxy","registry2":"registry2-proxy",...}`
2. Attempt to create the zarf package with `zarf package create`

### Expected result

This works with earlier versions prior to the registry override change in `v0.63.0`. In particular, it allows passing multiple overrides using a single environment variable.

### Actual Result

```
2025-10-08 06:51:17 ERR error parsing registry override: registry override missing '=': "docker.io":"local-proxy/registry-1.docker.io"
```

### Severity/Priority

This prevents us from upgrading Zarf versions without reworking how we call Zarf to use multiple `--registry-override ...` arguments

### Additional Context

brandtkeller:
## Description

In #4208 zarf improved handling of registry overrides at the cost of changing from a map[string]string to a []string with regards to processing inputs. 

The use of `flags.StringArrayVar()` allows for a more atomic setting of overrides where each `--registry-override` should provide 1 override (not comma delimited). 

This requires users to use the flag multiple times (advised) or a configuration file with a list of overrides as such:
```yaml
package:
  create:
    registry_override:
      source1: val1
      source2: val2
```

or 

```toml
[package.create]
registry_override = ["source1=val1", "source2=val2"]
```

Ultimately this leaves those using the environment variable constrained - as you cannot use a comma-delimited list and cannot create multiple unique overrides. 

This change introduces some compatibility for environment variables by allowing comma-delimiting - which also adds support for doing so in the CLI flags. 

## Related Issue

Fixes #4241

## Checklist before merging

### DIFF
--- a/site/src/content/docs/commands/zarf_package_create.md
+++ b/site/src/content/docs/commands/zarf_package_create.md
@@ -23,20 +23,20 @@ zarf package create [ DIRECTORY ] [flags]
 ### Options
 
 ```
-  -c, --confirm                         Confirm package creation without prompting
-      --differential string             [beta] Build a package that only contains the differential changes from local resources and differing remote resources from the specified previously built package
-  -f, --flavor string                   The flavor of components to include in the resulting package (i.e. have a matching or empty "only.flavor" key)
-  -h, --help                            help for create
-  -m, --max-package-size int            Specify the maximum size of the package in megabytes, packages larger than this will be split into multiple parts to be loaded onto smaller media (i.e. DVDs). Use 0 to disable splitting.
-      --oci-concurrency int             Number of concurrent layer operations when pulling or pushing images or packages to/from OCI registries. (default 6)
-  -o, --output string                   Specify the output (either a directory or an oci:// URL) for the created Zarf package
-      --registry-override stringArray   Specify a mapping of domains to override on package create when pulling images (e.g. --registry-override docker.io=dockerio-reg.enterprise.intranet)
-  -s, --sbom                            View SBOM contents after creating the package
-      --sbom-out string                 Specify an output directory for the SBOMs from the created Zarf package
-      --set stringToString              Specify package variables to set on the command line (KEY=value) (default [])
-      --signing-key string              Private key for signing packages. Accepts either a local file path or a Cosign-supported key provider
-      --signing-key-pass string         Password to the private key used for signing packages
-      --skip-sbom                       Skip generating SBOM for this package
+  -c, --confirm                     Confirm package creation without prompting
+      --differential string         [beta] Build a package that only contains the differential changes from local resources and differing remote resources from the specified previously built package
+  -f, --flavor string               The flavor of components to include in the resulting package (i.e. have a matching or empty "only.flavor" key)
+  -h, --help                        help for create
+  -m, --max-package-size int        Specify the maximum size of the package in megabytes, packages larger than this will be split into multiple parts to be loaded onto smaller media (i.e. DVDs). Use 0 to disable splitting.
+      --oci-concurrency int         Number of concurrent layer operations when pulling or pushing images or packages to/from OCI registries. (default 6)
+  -o, --output string               Specify the output (either a directory or an oci:// URL) for the created Zarf package
+      --registry-override strings   Specify a mapping of domains to override on package create when pulling images (e.g. --registry-override docker.io=dockerio-reg.enterprise.intranet)
+  -s, --sbom                        View SBOM contents after creating the package
+      --sbom-out string             Specify an output directory for the SBOMs from the created Zarf package
+      --set stringToString          Specify package variables to set on the command line (KEY=value) (default [])
+      --signing-key string          Private key for signing packages. Accepts either a local file path or a Cosign-supported key provider
+      --signing-key-pass string     Password to the private key used for signing packages
+      --skip-sbom                   Skip generating SBOM for this package
 ```
 
 ### Options inherited from parent commands


brandtkeller:
@richluby Would appreciate your thoughts.

richluby:
> @richluby Would appreciate your thoughts.

Looks like it should work. I've not worked much with `viper`. What's the interaction like if a user has a CLI flag, an environment variable, and the config file all set (like a 12-factor layout)?

brandtkeller:
> Looks like it should work. I've not worked much with `viper`. What's the interaction like if a user has a CLI flag, an environment variable, and the config file all set (like a 12-factor layout)?

Given the same underlying target for all three - precedence in Viper is Flag > Env > File to my knowledge.

richluby:
> Given the same underlying target for all three - precedence in Viper is Flag > Env > File to my knowledge.

Makes sense to me. I tried out a few locally. My concern was specifically that viper might attempt to provide the concatenation of all three. It appears to only take the highest precedent, which is much simpler to support.

richluby:
`mapstructure.StringToSliceHookFunc(",")`

I found this. Looks like viper supports custom decoders, and even provide some [out of the box](https://sagikazarmark.hu/blog/decoding-custom-formats-with-viper/#builtin-comma-separate-string-example).

Its potentially less immediately apparent what's happening, but does fit in with viper naturally.

brandtkeller:
Added documentation as well as E2E and unit testing. Contemplated moving the logic around but chose to keep it in `parseRegistryOverrides()` in order to unit test.  There are a few implications custom decoders - I am open to an enhancement but keeping it simple for the moment.

site/src/content/docs/tutorials/7-custom-init-packages.mdx
```diff
@@ -81,6 +81,33 @@ $ zarf package create . --set AGENT_IMAGE_TAG=vX.X.X \
   --registry-override quay.io=quay.enterprise.corp
 ```
 
+Registry overrides can also be orchestrated with environment variables and/or configuration files:
```
AustinAbro321, line 84: IMO the focus of this page should be on how custom init packages work, maybe a different page could have documentation on how to use environment variables with viper?

src/test/e2e/25_helm_test.go
```diff
@@ -107,6 +109,20 @@ func testHelmChartsExample(t *testing.T) {
...
+	t.Log("E2E: Helm chart with overrides")
+	tmpdir := t.TempDir()
+
+	// Create a test package (with multiple overrides set from an environment variable)
+	// This should fail but validates that the environment variable is correctly parsed
+	t.Setenv("ZARF_PACKAGE_CREATE_REGISTRY_OVERRIDE", "ghcr.io=localhost:555/noway,docker.io=local-proxy/registry-1.docker.io")
+	stdOut, stdErr, err := e2e.Zarf(t, "package", "create", "examples/helm-charts", "-o", tmpdir, "--tmpdir", tmpdir, "--confirm")
+	require.Error(t, err, stdOut, stdErr)
+	require.Contains(t, string(stdErr), "localhost:555/noway")
+	t.Setenv("ZARF_PACKAGE_CREATE_REGISTRY_OVERRIDE", "")
```
AustinAbro321: This should be deferred

src/cmd/package.go
AustinAbro321, line 1: I think instead of the changes that we're doing here, we should change the command to use `StringSliceVar` instead of `StringArrayVar`. `StringSliceVar` splits on commas.  

`v.GetStringSlice` does not split on `,` but instead splits on empty space ` `. I think this is pretty confusing behavior on vipers end, but changing this set us up to support commas instead of whitespace whenever `v.GetStringSlice` is used. I'm not sure if there is a viper configuration that allows us to change this globally. Otherwise, I'd rather keep it as is so we're consistent with other projects that use viper, and other commands that we might use `StringSliceVar` (currently there are none in the project)

brandtkeller: I was going back and forth - beyond the implications of splitting - I think what I wanted to convey is choosing the option which is most atomic. 

`StringArrayVar` conveys a UX that should be clear and repeatable - as seen in the existing documentation which encourages --registry-override a=b --registry-override c=d, avoiding surprises from a single CSV flag.

When we look at splitting - `StringSliceVar` still succumbs to the same bug - it misses everything after the first comma.

site/src/content/docs/tutorials/7-custom-init-packages.mdx
brandtkeller: I agree there isn't a great fit for this today. Config files have documentation, variables have some documentation, but this could be broken out in some way but would require some larger docs alignment. This example was specific and so I didn't reach for a complete rewrite.

src/cmd/package.go
AustinAbro321: When I try it out, I am seeing that StringSliceVar splits the slice properly by comma. This is in line with the documentation. https://github.com/spf13/pflag/blob/6fcfbc9910e1af538fde31db820be7d1bec231e4/string_slice.go#L97-L106.

brandtkeller: If I add a debug log before/after parsing and switch to `StringSliceVar` -  when I export the environment variable with:
```bash
export ZARF_PACKAGE_CREATE_REGISTRY_OVERRIDE='docker.io=dockerio.enterprise.corp,ghcr.io=ghcr.enterprise.corp,quay.io=quay.enterprise.corp'
```

I see: 

```bash
2025-10-09 16:54:40 DBG registry overrides overrides=[docker.io=dockerio.enterprise.corp,ghcr.io=ghcr.enterprise.corp,quay.io=quay.enterprise.corp]
2025-10-09 16:54:40 DBG parsed registry overrides overrides=[{docker.io dockerio.enterprise.corp,ghcr.io=ghcr.enterprise.corp,quay.io=quay.enterprise.corp}]
```

That is a single element pre-parsing. Are you making any other modifications off main?

AustinAbro321: Ah yeah the environment variables on the viper side. Having StringSliceVar allows comma separated lists to work with flags. So `--registry-overrides=docker.io=dockerio.enterprise.corp,ghcr.io=ghcr.enterprise.corp,quay.io=quay.enterprise.corp'` will work, but it won't change environment variables. 

`v.GetStringSlice` splits using whitespace instead of commas. I don't love this choice by viper, but I also don't love the option of maintaining separate ways to use viper keys. If we introduce another slice flag we'd have to remember to split by commas or we'll have an inconsistent viper experience.

brandtkeller: You raise a fair point about getters standardization. 

This feels like a big inconsistency for our expectation of support around environment variables, flags, and configuration files. 

Could we maintain our own getter in this instance? annotate the gap we believe is present?

AustinAbro321: I think in this instance, given from my research it looks like whitespace is non-configurable as the splitter for string slices, and this is the first instance of string slices we are using, it's reasonable to use our own getter. 

Looking at this thread, and from my local testing, unmarshalling works https://github.com/spf13/viper/issues/380#issuecomment-1916465489. We could have a function like getStringSlice(v *viper.Viper, key string), and use that instead along with the unmarshall technique

brandtkeller: Updated!

PR #4252, opened 2025-10-09
feat(values): add schema validation

### LINKED ISSUE #4225: Implement JSON Schema validation for values files
Users should be able to provide JSON schemas for their valuesfiles to ensure type safety and validation. That includes the values declared in the file itself and any subsequently imported files that take precedence, values set directly via CLI flag (see #4224), and Values set during actions with setValue.

mkcp:
## Description
This PR adds support for json schema validation of values both at create and deploy time.

More tangetially, it also adds in a schema check on the v1alpha1 struct for ZarfChartValue mappings, ensuring they start with a `.`, are provided as `.` segmented sections, and are all valid characters for use in a golang string map.

Once #4268 is ready, I think it would make sense to do a schema validate there as well. Remove doesn't have access to the package value schema contents yet so adding it to remove would be more a of future concern.

## Related Issue
Fixes #4225
Relates to #3946 

## Checklist before merging

### DIFF
--- a/src/api/v1alpha1/component.go
+++ b/src/api/v1alpha1/component.go
@@ -190,8 +190,8 @@ type ZarfChartVariable struct {
 
 // ZarfChartValue maps a Zarf Value key to a Helm Value.
 type ZarfChartValue struct {
-	SourcePath string `json:"sourcePath" jsonschema:"pattern=^\\.[a-zA-Z0-9_-]+(\\.[a-zA-Z0-9_-]+)*$"`
-	TargetPath string `json:"targetPath" jsonschema:"pattern=^\\.[a-zA-Z0-9_-]+(\\.[a-zA-Z0-9_-]+)*$"`
+	SourcePath string `json:"sourcePath" jsonschema:"pattern=^(\\.|\\.[a-zA-Z0-9_-]+(\\.[a-zA-Z0-9_-]+)*)$"`
+	TargetPath string `json:"targetPath" jsonschema:"pattern=^(\\.|\\.[a-zA-Z0-9_-]+(\\.[a-zA-Z0-9_-]+)*)$"`
 }
 
 // ZarfManifest defines raw manifests Zarf will deploy as a helm chart.
--- a/zarf.schema.json
+++ b/zarf.schema.json
@@ -453,11 +453,11 @@
       },
       "properties": {
         "sourcePath": {
-          "pattern": "^\\.[a-zA-Z0-9_-]+(\\.[a-zA-Z0-9_-]+)*$",
+          "pattern": "^(\\.|\\.[a-zA-Z0-9_-]+(\\.[a-zA-Z0-9_-]+)*)$",
           "type": "string"
         },
         "targetPath": {
-          "pattern": "^\\.[a-zA-Z0-9_-]+(\\.[a-zA-Z0-9_-]+)*$",
+          "pattern": "^(\\.|\\.[a-zA-Z0-9_-]+(\\.[a-zA-Z0-9_-]+)*)$",
           "type": "string"
         }
       },


src/test/packages/42_values/basic/action-configmap.yaml
AustinAbro321: Files in 42_values are not used

src/internal/value/value.go
```diff
@@ -237,6 +238,77 @@ func (v Values) Set(path Path, newVal any) error {
...
+	}
+	// Convert backslashes to forward slashes for file URI
+	absPath = filepath.ToSlash(absPath)
+	schemaLoader := gojsonschema.NewReferenceLoader("file:///" + absPath)
+
+	// Convert Values to a document for validation
+	documentLoader := gojsonschema.NewGoLoader(v)
+
+	// Validate
+	result, err := gojsonschema.Validate(schemaLoader, documentLoader)
```
AustinAbro321, line 277: One quirk of validating at both create and deploy time is the `required` field. I think the advantage of the required field is that you can force deployers to set a field when you don't want them to use a default. However, currently when `required` is set you are forced to set the field during create time, but only if your package contains a values.yaml file. 

I think we should skip required errors during create. It looks like the jsonSchema errors do have a type() function which returns required, unfortunately, it is not an enum

src/test/e2e/42_values_test.go
```diff
@@ -64,3 +64,62 @@ func TestValues(t *testing.T) {
...
+	t.Run("invalid values fail schema validation at create time", func(t *testing.T) {
+		src := filepath.Join("src", "test", "packages", "42-values", "schema-invalid")
+		tmpdir := t.TempDir()
+
+		// Create should fail with invalid values
+		stdOut, stdErr, err := e2e.Zarf(t, "package", "create", src, "-o", tmpdir, "--skip-sbom", "--confirm", "--features=\"values=true\"")
+		require.Error(t, err, "expected error for invalid values")
+		// Check that the error message mentions validation failure
+		output := stdOut + stdErr
+		require.Contains(t, output, "values validation failed", "error should mention schema validation failure")
```
AustinAbro321: nit: error output will always be in stderr

AustinAbro321 (changes requested):
Mainly a question of create vs deploy "required" enforcement

src/test/e2e/42_values_test.go
mkcp: Fixed in latest

src/internal/value/value.go
mkcp: Yeah this is a good suggestion, I've gone ahead and added the a SkipRequired option and worked it through the call stack in a few places. It's set by default during create. Let me know what you think

AustinAbro321 (approved):
LGTM

src/pkg/packager/layout/assemble.go
```diff
@@ -163,6 +163,15 @@ func AssemblePackage(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath
 		}
 	}
 
+	// Copy schema file if specified
+	if pkg.Values.Schema != "" {
+		l.Debug("copying values schema file to package", "schema", pkg.Values.Schema)
+		// NOTE(mkcp): values schemas should have the same copy-constraints as values files.
```
AustinAbro321: Something I just noticed, is that there is an additional constraint, we'll want to check that the schema is a valid json schema. Perhaps the best way to do this is to make sure `gojsonschema.NewSchema()` doesn't error? It's possible that a user has a json schema, but provides no values files on create, so they wouldn't discover they have a fault schema until deploy time

src/internal/value/value.go
```diff
@@ -237,6 +238,99 @@ func (v Values) Set(path Path, newVal any) error {
...
+		l.Debug("schema validation complete",
+			"duration", time.Since(start),
+			"schemaPath", schemaPath)
+	}()
+
+	// Load the schema from file
+	// Convert to absolute path and ensure forward slashes for URI
+	absPath, err := filepath.Abs(schemaPath)
+	if err != nil {
+		return &SchemaValidationError{
```
AustinAbro321: IMO this and the error on `gojsonschema.validate` shouldn't use this typed error. It seems like there is a delineation between the schema failing to load and values not adhering to the schema. This presents itself in the `SchemaValidationError` struct which has both Err and errs making it feel a bit awkward.

src/pkg/packager/load/load.go
```diff
@@ -27,6 +28,9 @@ import (
 type DefinitionOptions struct {
 	Flavor       string
 	SetVariables map[string]string
+	// SkipRequiredValues ignores values schema validation errors when a "required" field is empty. Used when a package
+	// value should be supplied at deploy-time and doesn't have a default set in the package values.
+	SkipRequiredValues bool
```
AustinAbro321, line 33: Skeleton package publishes should also set this to true

src/test/e2e/42_values_test.go
AustinAbro321: We are still combining stdout/stderr

AustinAbro321 (changes requested):
Few more items I noticed / nits

src/test/e2e/42_values_test.go
mkcp: ahh, i see there were multiple instances of this and I only fixed one of them. Resolved in latest

src/pkg/packager/load/load.go
mkcp: added to PublishSkeleton's load.DefinitionOptions

src/internal/value/value.go
mkcp: That's fair, most of the intent was to associate the schemapath with it, but that can be added to the error message. Updated

src/pkg/packager/layout/assemble.go
mkcp: Makes sense to account for this edge case, updated in latest. I noticed that the copy function this PR is going to be removed soon with https://github.com/zarf-dev/zarf/pull/4403, so I've updated this PR in anticipation of that change. Now that we're copying and loading the schema directly from package root, I'm hoping any merge conflicts are minimal.

src/internal/value/value.go
```diff
@@ -237,6 +238,118 @@ func (v Values) Set(path Path, newVal any) error {
...
+	}
+
+	return nil
+}
+
+// SchemaValidationError represents an error when JSON schema validation fails
+type SchemaValidationError struct {
+	SchemaPath string
+	Errors     []gojsonschema.ResultError
+	Err        error
```
AustinAbro321: This err field is never set, I believe we can delete it and replace it's uses with the errors field

mkcp: Yeah that makes sense - Unwrap won't work out of the box with gojsonschema.ResultError, so for now I'm going to skip implementing Unwrap for errors.Is and errors.As. We can revisit it in the future if we end up needing it.

AustinAbro321 (dismissed):

AustinAbro321 (approved):

PR #4260, opened 2025-10-11
fix: wait shell escaping

### LINKED ISSUE #3844: Zarf tools wait-for unable to use SELECTOR - always returns true

vincent-onebrief:
## Description

Currently when waiting for a jsonpath expression with bash special characters, the args passed to `kubectl wait` aren't properly escaped.

```console
$ zarf tools wait-for -n onebrief clusters cnpg-postgresql \
    '{.status.conditions[?(@.type=="Ready")].status}=True' \
    -l debug \
    --timeout 3s
DBG logger successfully initialized cfg.level=debug cfg.format=console cfg.destination=os.Stderr cfg.color=true
DBG User-configured features: features=
INF using config file location=zarf-config.yaml
DBG using temporary directory tmpDir=
INF waiting for clusters/cnpg-postgresql in namespace onebrief to exist.
DBG checking resource existence namespace=-n onebrief kind=clusters identifier=cnpg-postgresql
DBG cmd done cmd=[-e -c zarf tools kubectl get -n onebrief clusters cnpg-postgresql] stdout=NAME              AGE   INSTANCES   READY   STATUS                     PRIMARY
cnpg-postgresql   26h   1           1       Cluster in healthy state   cnpg-postgresql-1
 stderr= error=<nil>
INF waiting for clusters/cnpg-postgresql in namespace onebrief to be {.status.conditions[?(@.type=="Ready")].status}=True.
DBG wait error error=exit status 2
DBG checking resource existence namespace=-n onebrief kind=clusters identifier=cnpg-postgresql
DBG cmd done cmd=[-e -c zarf tools kubectl get -n onebrief clusters cnpg-postgresql] stdout=NAME              AGE   INSTANCES   READY   STATUS                     PRIMARY
cnpg-postgresql   26h   1           1       Cluster in healthy state   cnpg-postgresql-1
 stderr= error=<nil>
INF waiting for clusters/cnpg-postgresql in namespace onebrief to be {.status.conditions[?(@.type=="Ready")].status}=True.
DBG wait error error=exit status 2
ERR wait timed out
```

The same `kubectl wait` succeeds without issue

```console
$ kubectl wait -n onebrief \
  clusters cnpg-postgresql \
  --for jsonpath='{.status.conditions[?(@.type=="Ready")].status}=True' \
  --timeout=5s
cluster.postgresql.cnpg.io/cnpg-postgresql condition met
```

This is due to

```log
DBG wait done cmd=[-e -c build/zarf-mac-apple tools kubectl wait -n onebrief clusters cnpg-postgresql --for jsonpath={.status.conditions[?(@.type=="Ready")].status}=True --timeout=5s] stdout= stderr=sh: -c: line 0: syntax error near unexpected token `(' error=exit status 2
DBG wait error error=exit status 2
```

## Related Issue

Fixes #3844

## Checklist before merging

### DIFF
--- a/src/pkg/utils/wait.go
+++ b/src/pkg/utils/wait.go
@@ -30,14 +30,18 @@ func isJSONPathWaitType(condition string) bool {
 	return true
 }
 
-var shellRegex = regexp.MustCompile(`[^\w@%+=:,./-]`)
+// unsafeShellCharsRegex matches any character that is NOT a letter, digit, underscore (/w) or shell safe special characters
+var unsafeShellCharsRegex = regexp.MustCompile(`[^\w@%+=:,./-]`)
 
-func ShellQuote(s string) string {
+// Source: https://github.com/alessio/shellescape/blob/v1.6.0/shellescape.go#L30-L42
+// SPDX-License-Identifier: MIT
+// Minor edits: Simplified for use case
+func shellQuote(s string) string {
 	if len(s) == 0 {
 		return "''"
 	}
 
-	if shellRegex.MatchString(s) {
+	if unsafeShellCharsRegex.MatchString(s) {
 		return fmt.Sprintf("'%s'", strings.ReplaceAll(s, "'", "'\"'\"'"))
 	}
 
@@ -61,7 +65,7 @@ func ExecuteWait(ctx context.Context, waitTimeout, waitNamespace, condition, kin
 	if isJSONPathWaitType(condition) {
 		waitType = "jsonpath="
 		// Ensure any conditions aren't shell escaped
-		condition = ShellQuote(condition)
+		condition = shellQuote(condition)
 	} else {
 		waitType = "condition="
 	}


vincent-onebrief:
I didn't find any tests related to `wait-for` and the actual CLI calls that get made

src/pkg/utils/wait.go
```diff
@@ -45,6 +46,8 @@ func ExecuteWait(ctx context.Context, waitTimeout, waitNamespace, condition, kin
 	// Check if waitType is JSONPath or condition
 	if isJSONPathWaitType(condition) {
 		waitType = "jsonpath="
+		// Ensure any conditions aren't shell escaped
+		condition = shellescape.Quote(condition)
```
AustinAbro321: Rather than adding another dependency, we should copy this function into Zarf. Seems like it's small enough to simply copy into this package. https://github.com/alessio/shellescape/blob/b09271781b1a504d1b88e5b44a28e8159b2efd74/shellescape.go#L30-L42

src/pkg/utils/wait.go
```diff
@@ -29,6 +30,20 @@ func isJSONPathWaitType(condition string) bool {
 	return true
 }
 
+var shellRegex = regexp.MustCompile(`[^\w@%+=:,./-]`)
+
+func ShellQuote(s string) string {
```
AustinAbro321: We should make this function private

src/pkg/utils/wait.go
```diff
@@ -29,6 +30,20 @@ func isJSONPathWaitType(condition string) bool {
 	return true
 }
 
+var shellRegex = regexp.MustCompile(`[^\w@%+=:,./-]`)
```
AustinAbro321:
```suggestion
// unsafeShellCharsRegex matches any character that is NOT a letter, digit, underscore (/w) or shell safe special characters
var unsafeShellCharsRegex = regexp.MustCompile(`[^\w@%+=:,./-]`)
```
small comment so it's easier to see what the regex does

src/pkg/utils/wait.go
```diff
@@ -128,7 +145,10 @@ func ExecuteWait(ctx context.Context, waitTimeout, waitNamespace, condition, kin
 				zarfCommand, namespaceFlag, kind, identifier, waitType, condition, waitTimeout)
 
 			// If there is an error, log it and try again.
-			if _, _, err := exec.Cmd(shell, append(shellArgs, zarfKubectlWait)...); err != nil {
+			waitCmd := append(shellArgs, zarfKubectlWait)
+			waitStdout, waitStderr, err := exec.Cmd(shell, waitCmd...)
+			l.Debug("wait done", "cmd", waitCmd, "stdout", waitStdout, "stderr", waitStderr, "error", err)
```
AustinAbro321, line 154:
```suggestion
			l.Debug("wait done", "cmd", waitCmd, "stdout", waitStdout, "stderr", waitStderr)
```
nit: we already print the error on the next line if it happens

src/pkg/utils/wait.go
brandtkeller: Additionally something you could help us create a reference for here is code attribution. Not something we have an example for today but we have been discussing. 

I'd love to see a couple lines that attribute the copied code

```suggestion
// Source: https://github.com/alessio/shellescape/blob/v1.6.0/shellescape.go
// SPDX-License-Identifier: MIT
// Minor edits: add any edits here or remove line
func shellQuote(s string) string {
```

src/pkg/utils/wait.go
vincent-onebrief: This matches the existing pattern for `zarfKubectlGet`

brandtkeller (approved):
This looks reasonable. We don't have much in the way of testing this for regressions currently but I won't block on that. Open to additional testing for anyone interested.

PR #4267, opened 2025-10-14
feat(values): package remove actions templating

### LINKED ISSUE #4222: Add values and templating support for package remove actions and inspect
### Context
Currently values and templating are only fully supported during package deployment. These feaures are necessary in remove actions to ensure proper cleanup, and inspect allows users to better understand how go-templates will be applied in their files.

## Related:
  - Parent issue: #3946
  - Docs: Need to clarify when values are needed for remove operations

mkcp:
## Description
This PR ensures go-templates can be used in remove actions. e2e coverage and examples references included

## Related Issue
Fixes #4222 
Relates to #3946 

## Checklist before merging

### DIFF
--- a/examples/values-templating/zarf.yaml
+++ b/examples/values-templating/zarf.yaml
@@ -114,5 +114,6 @@ x-mdx: |
     kubectl get configmap -n helm-overrides
 
     # Remove the package with values templating in remove actions
-    zarf package remove values-templating --confirm --features="values=true"
+    # Feel free to change --set-values to whatever you want!
+    zarf package remove values-templating --confirm --features="values=true" --set-values="site.name=Example,app.environment=test,site.organization=ZarfDev"
     ```


src/test/e2e/42_values_test.go
```diff
@@ -53,7 +53,15 @@ func TestValues(t *testing.T) {
...
-	// Remove the package
-	stdOut, stdErr, err = e2e.Zarf(t, "package", "remove", "test-values", "--confirm")
+	// Remove the package with values
+	valuesFile := filepath.Join(src, "values", "values.yaml")
+	stdOut, stdErr, err = e2e.Zarf(t, "package", "remove", "test-values", "--confirm", "--features=\"values=true\"", "--values", valuesFile, "--set-values", "removeKey=custom-remove-value")
 	require.NoError(t, err, stdOut, stdErr)
+
+	// Verify the remove actions used the values correctly
+	// Check that the default value from values.yaml was templated
+	require.Contains(t, stdOut, "REMOVE_TEST_VALUE=default-value", "remove action should have templated default value from values.yaml")
```
AustinAbro321: Should we pass the `override-values.yaml` file into remove? This doesn't check that the default values are templated since we're passing them in.

mkcp: Yeah, resusing the package values here would be confusing - it even tripped me up while I was revising this. I'll test instead for just the overrides and --set-values here, and make a note to clearly document what values are in scope during remove.

AustinAbro321: Now this tests that we can add override files, but we don't have a test that checks that the package defaults from `pkgLayout.Pkg.Values.Files` are getting loaded. Seems like there is actually a bug atm where they are not.

mkcp: package remove only works with deployed packages that are retrieved from the cluster, yeah? We don't reify a layout from a package build even if it were available locally, so we have the filenames from the packagedefinition but not the actual contents. There's a note about this in wayne's ZEP:

> Zarf Values would also be available to onRemove actions if they existed in package.remove.values or were set with -f/--values. These would template actions and be able to use setValues but would not map to Helm charts.

This is imo pretty reasonable, and is the behavior that is tested for in the branch.

I'm not against loading a package in remove if it's available - but that would require deeper changes to remove and is better addressed as its own issue scope.

AustinAbro321: Great point, I totally forget that the remove values will not be available! Yeah definitely it's own issue, I'll create that

AustinAbro321: https://github.com/zarf-dev/zarf/issues/4317

examples/values-templating/zarf.yaml
```diff
@@ -103,4 +112,7 @@ x-mdx: |
 
     # View the helm chart results
     kubectl get configmap -n helm-overrides
+
+    # Remove the package with values templating in remove actions
+    zarf package remove values-templating --confirm --features="values=true"
```
AustinAbro321: nit: we should probably add `--set-values` or a `zarf-config.yaml` as just running this errors

mkcp: Good catch! Updated in latest

AustinAbro321 (approved):

PR #4268, opened 2025-10-14
feat(values): dev inspect

mkcp:
## Description
This PR adds values to dev inspect, so users can see the results of the package values that they have loaded.

## Related Issue
Fixes https://github.com/zarf-dev/zarf/issues/4222
Relates to https://github.com/zarf-dev/zarf/issues/3946

## Checklist before merging

### DIFF
--- a/src/cmd/dev_test.go
+++ b/src/cmd/dev_test.go
@@ -12,7 +12,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
-	"github.com/zarf-dev/zarf/src/internal/feature"
+	"github.com/zarf-dev/zarf/src/pkg/feature"
 	"github.com/zarf-dev/zarf/src/pkg/lint"
 	"github.com/zarf-dev/zarf/src/pkg/utils"
 	"github.com/zarf-dev/zarf/src/test/testutil"
--- a/src/pkg/packager/inspect_test.go
+++ b/src/pkg/packager/inspect_test.go
@@ -7,8 +7,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
-	"github.com/zarf-dev/zarf/src/internal/feature"
 	"github.com/zarf-dev/zarf/src/internal/value"
+	"github.com/zarf-dev/zarf/src/pkg/feature"
 	"github.com/zarf-dev/zarf/src/pkg/lint"
 	"github.com/zarf-dev/zarf/src/test/testutil"
 )


src/pkg/packager/inspect.go
```diff
@@ -176,6 +179,8 @@ func templateValuesFiles(chart v1alpha1.ZarfChart, valuesDir string, variableCon
 type InspectDefinitionResourcesOptions struct {
 	CreateSetVariables map[string]string
 	DeploySetVariables map[string]string
+	ValuesFiles        []string
+	SetValues          map[string]string
```
AustinAbro321: In other SDK functions such as deploy and remove, we pass in `values.Values`. Is there a reason we pass in values files and set values here instead?

mkcp: Def, this change didn't make sense - missed it while trying to automate some of these additions. The parse and merge code now lives in the CLI layer, while the API layer only takes a `value.Values` - same as the rest of the implementations.

src/cmd/dev.go
```diff
@@ -242,9 +271,29 @@ func (o *devInspectValuesFilesOptions) run(ctx context.Context, args []string) e
 	if err != nil {
 		return err
 	}
+
+	// Parse values from files
+	values, err := value.ParseFiles(ctx, o.valuesFiles, value.ParseFilesOptions{})
```
AustinAbro321, line 277: Some duplication in this block, that can be pulled out into a common function. I know there's a few active PRs with blocks like this, so I think a follow up PR afterwards would be reasonable

src/pkg/packager/inspect.go
```diff
@@ -264,58 +269,91 @@ func InspectDefinitionResources(ctx context.Context, packagePath string, opts In
...
 		}
+
 		resources = append(resources, Resource{
 			Content:      string(content),
-			Name:         manifest,
+			Name:         manifestFile,
 			ResourceType: ManifestResource,
 		})
-		return nil
+		return err2
```
AustinAbro321: nit: I think it's clearer to return nil here. In this case, returning err2 is the same as returning nil since err2 is only set in the defer block. Technically, it could make a difference if we set err2, then continued with the rest of the function, but I think that's a pattern we wouldn't want to encourage.

src/cmd/dev.go
AustinAbro321, line 1: it appears we are never sending in the values files from  the top level `values.files` in the package definition

AustinAbro321, line 1: The `dev inspect` commands have existing patterns for unit tests which test both the cmd and packager layer, I think we should add some for values. Ref: https://github.com/zarf-dev/zarf/blob/main/src/cmd/dev_test.go#L20

AustinAbro321 (changes requested):

src/pkg/packager/inspect.go
```diff
@@ -176,8 +178,11 @@ func templateValuesFiles(chart v1alpha1.ZarfChart, valuesDir string, variableCon
 type InspectDefinitionResourcesOptions struct {
 	CreateSetVariables map[string]string
 	DeploySetVariables map[string]string
-	Flavor             string
-	KubeVersion        string
+	// Values are values passed in at inspect time. They can come from the CLI, user configuration, or set directly by
+	// API callers.
+	value.Values
```
brandtkeller, line 183: With the CLI as the function caller - it parses values files from configuration and set values - but we have not yet parsed the values that were declared within the package 

```yaml
kind: ZarfPackageConfig
metadata:
  name: values-templating
  description: Example nginx package to demonstrate Zarf Values templating
values:
  files:
    - values.yaml
```

Should we be parsing these similar to deploy-time to include in templating?

src/cmd/dev.go
brandtkeller: This was my comment below as well. You beat me to it 😆 .

mkcp: Ah! Yeah we definitely should - I'm not used to us unit testing our cmd layer but it's good practice.

src/cmd/dev.go
mkcp: It does how up pretty often. This makes sense to me as two steps in the CLI layer because it's coming from two different sources (rather than wrapping both load valuesfiles and setvalues in the same helper func). The setValue step would definitely benefit from abstracting out the loop and path parsing though. I'll make a followup issue for it

src/cmd/dev.go
mkcp: Yeah this branch is buggy - spinning a few plates rn and missed giving dev inspect the proper attention. I'm reworking the approach and making sure the tests are in place for the intended behavior.

src/pkg/packager/inspect.go
mkcp: Yes, this was a miss in the previous version of the PR. It's been changed significantly now and should be addressed

src/pkg/packager/inspect.go
mkcp: yeah i think you're right, i was trying to make sure that the deferred error always gets returned from the closure, but that should be handled by the assignment in defer already whether we return it at the end or not.

src/cmd/dev.go
mkcp: Resolved in latest!

mkcp: Reworked in latest revision

src/pkg/packager/inspect.go
```diff
@@ -297,25 +351,25 @@ func getTemplatedManifests(ctx context.Context, manifest v1alpha1.ZarfManifest,
 		return nil, err
 	}
 
-	return resources, nil
+	return resources, err
```
AustinAbro321:
```suggestion
	return resources, nil
```

AustinAbro321 (commented):
minor nit otherwise looks good

Issue #4284, opened 2025-10-17
Add support for Kustomize Plugins


JeffResc:
### Is your feature request related to a problem? Please describe.
There is no way to build Zarf packages that have Kustomize templates that use the [Kustomize Plugins](https://kubectl.docs.kubernetes.io/guides/extending_kustomize/) feature. Once the package is built, deploying the package is not an issue.

AustinAbro321:
This was added by #4285

Issue #3095, opened 2024-10-14
Cannot template numbers through kustomize


nbrns:
### Environment
Device and OS: Windows 11 / WSL2
App version: v0.41.0
Kubernetes distro being used: EKS / kind

### Steps to reproduce
1. Define a zarf package that tries to template a integer value through kustomize, e.g. deployment spec replica count

zarf.yaml
```yaml
kind: ZarfPackageConfig
metadata:
  name: podinfo
  version: 1.0.0
variables:
  - name: DEPLOYMENT_REPLICAS
    default: "3"
components:
  - name: podinfo
    required: true
    manifests:
      - name: podinfo
        kustomizations:
          - manifests
```

manifests/kustomization.yaml
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - podinfo.yaml
```

manifests/podinfo.yaml
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: podinfo-deployment
spec:
  replicas: ###ZARF_VAR_DEPLOYMENT_REPLICAS###
  selector:
    matchLabels:
      app: podinfo
  template:
    metadata:
      labels:
        app: podinfo
    spec:
      containers:
      - name: podinfo
        image: ghcr.io/stefanprodan/podinfo:latest
        ports:
        - containerPort: 9898
```

### Expected result

podinfo is running with 3 replicas.

### Actual Result

podinfo is running with 1 replica.

The reason for this is because Zarf first applies the kustomize build. The kustomize build eliminates the Zarf variables, as they are treated as comments. Thus, the replica property is set to `null`. This results in the default value of 1 for deployments. While this results in deployments not being scaled, with some CRDs, where the replica property is required, this even breaks packages.

mkcp:
We do want to fix this behavior, it relates to the in-flight Values feature but needs to be accounted for in the implementation. 

rel #3946

Issue #2236, opened 2024-01-19
Use `kustomize localize` for `kustomize` resources


Racer159:
### Is your feature request related to a problem?

**As** Ezra **I want** to have kustomize resources templated on deploy **so that** I can control the templates with deploy time Zarf variables.

### Describe the solution you'd like

- **Given** I have Zarf variables that control the templating of my kustomizations
- **When** I specify --set on zarf package deploy
- **Then** those variables are templated into the customizations
- **And Then** the kustomizations are built

### Describe alternatives you've considered
Currently we template on create so that all resources are pulled from remote sources - this reduces some flexibility for deployments though.  We also could redesign variables to work more cleanly but this would require more work and would still need to take this into account.

### Additional context
https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1705676553417059?thread_ts=1705676553.417059&cid=C03B6BJAUJ3

PR #4285, opened 2025-10-17
feat: add support for kustomize plugins

JeffResc:
## Description
Adds support for enabling Kustomize plugins, by specifying `enableKustomizePlugins: true` in Zarf manifest

## Related Issue

Relates to #4284

## Checklist before merging


AustinAbro321:
One thing I realized as I was trying this out is that the exec plugins still don't work since we don't set `buildOptions.PluginConfig.FnpLoadingOptions.EnableExec = true`.

I think I would be fine with it on create, but if we did ever set it on deploy, for example, so we could template kustomizations (#3095), it's a bit more dicey and hidden. Could you tell me which plugins you are looking to use this for ?

JeffResc:
Hey @AustinAbro321, thanks for looking into this! The plugins are custom and something I cannot share here, but they do not require `enableExec`

AustinAbro321:
The only hesitation I have currently is that this will make implementing #2236 more complex. Plugins assume files on disk and we wouldn't be able to assume that a custom plugin is on disk in the deploy environment. 

This is probably fine though, we could expose an option on the kustomization struct such as `buildOnCreate` and have that default to false. Plugin users could set it to true

src/internal/packager/kustomize/build.go
```diff
@@ -26,6 +26,8 @@ func Build(path string, destination string, kustomizeAllowAnyDirectory bool) err
 		buildOptions.LoadRestrictions = krustytypes.LoadRestrictionsNone
 	}
 
+	buildOptions.PluginConfig = krustytypes.MakePluginConfig(krustytypes.PluginRestrictionsNone, krustytypes.BploUseStaticallyLinked)
```
AustinAbro321: Let's introduce a field in ZarfManifest called `enableKustomizePlugins` and set this based on that. This is in line with current kustomize behavior which puts plugins behind the `--enable-alpha-plugins` flag. 

It also gives package creators who may not have the plugin available the signal that it's required for a certain component.

AustinAbro321 (changes requested):

src/internal/packager/kustomize/build.go
JeffResc: Good call, I think that's a fair addition to this PR since it's an alpha feature

JeffResc: Hey @AustinAbro321, this has been implemented and is ready for review

PR #4295, opened 2025-10-20
feat: improve error message when image platform not found

### LINKED ISSUE #4294: Improve Error Message for No Matching Manifest
When creating a Zarf package that includes multiple container images, if one of the images fails due to a missing manifest for the target architecture, the error message does not indicate which specific image caused the failure.  

This makes debugging difficult for large packages because the output only shows a digest (`sha256`) rather than the actual image reference.

When building larger Zarf packages, this problem compounds because other image fetches may fail with context cancellation errors, obscuring the actual root cause.

See this example run on UDS Core:
[GitHub Actions log with misleading context canceled errors](https://github.com/defenseunicorns/uds-core/actions/runs/18655834419/job/53184592413?pr=2040#step:8:215)

- **Given** a Zarf package build process that includes multiple container images
- **When** one or more images fail to fetch due to a missing manifest for the target architecture
- **Then** should display both the image reference and tag in the error message alongside the digest.  Example suggested output:
```
failed to create package: failed to fetch image with architecture arm: image: amd64/unit:python3.12 sha256:eb08c13116018c33639b3a82c0e605f5e3310f0b0f5ee785e36d5b4865a8e144: not found: no matching manifest was found in the manifest list
```

- Manually running docker manifest inspect on each image in a package to find the problem image.
- Correlating digests to images, which is tedious

### Additional context
Minimal reproduction

`zarf.yaml`
```yaml
kind: ZarfPackageConfig
metadata:
  name: test-bad-package
  description: "Test Bad Package"

components:
  - name: bad-package
    required: true
    description: "This package's first image did not have an arm manifest"
    images:
      - amd64/unit:python3.12  # no arm manifest
      - docker.io/nginx:latest # ok image
```

```console
zarf package create -a arm
2025-10-20 11:48:44 INF assembling package path=.
2025-10-20 11:48:44 INF fetching info for images count=2 destination=/var/folders/92/351y94vn3yn9_gzl_jw60j1h0000gn/T/zarf-1707802018/images
2025-10-20 11:48:46 ERR failed to create package: failed to fetch image with architecture arm: sha256:eb08c13116018c33639b3a82c0e605f5e3310f0b0f5ee785e36d5b4865a8e144: not found: no matching manifest was found in the manifest list
```

In this specific case `amd64/unit:python3.12` is the failing image with no arm64 architecture.

AustinAbro321:
## Related Issue

Fixes #4294

## Checklist before merging

### DIFF
--- a/src/internal/packager/images/pull.go
+++ b/src/internal/packager/images/pull.go
@@ -200,7 +200,7 @@ func Pull(ctx context.Context, cfg PullConfig) (map[transform.Image]ocispec.Mani
 				fetchOpts.FetchOptions.TargetPlatform = platform
 				desc, b, err = oras.FetchBytes(ectx, repo, image.overridden.Reference, fetchOpts)
 				if err != nil {
-					return fmt.Errorf("failed to fetch image with architecture %s: %w", platform.Architecture, err)
+					return fmt.Errorf("failed to fetch image %s with architecture %s: %w", image.overridden.Reference, platform.Architecture, err)
 				}
 			}
 


brandtkeller (approved):
LGTM

PR #4301, opened 2025-10-21
feat!(sign): add package sign command

### LINKED ISSUE #3959: Command to re-sign Zarf packages
At present a package can be signed by providing the signing key during package creation and re-signed by providing a signing key during publish. To locally get access to a re-signed package the user needs to publish a package with the new key and then redownload it. I would like a command to skip the publishing and redownloading so that an existing signed package can be re-signed directly through the CLI.

- **Given** an existing signed Zarf package
- **When** a CLI command is run to re-sign it with the proper keys
- **Then** the package will be re-signed with the new keys

Publishing to a registry while passing in the signing keys and pulling back down from that registry

brandtkeller:
## Description
Adds a new zarf package sign command that enables signing existing Zarf packages with cryptographic signatures using Cosign. This allows packages to be signed independently from the build process.

### Key Changes
New Command: `zarf package sign`
Functionality: Signs existing Zarf packages (local tarballs or OCI packages) with a private key
Key Features:
- Sign local package files or pull from OCI registries
- Output to local directory or publish directly to OCI registry
- Support for multiple key providers (local files, AWS KMS, GCP KMS, Azure KMS,)
- `--overwrite` flag to re-sign packages with a new key
- Optional signature validation before re-signing
### Core Implementation Changes
Package Layout:
- Refactored SignPackage() to use context and structured options
- Atomic signing process: updates Build.Signed metadata, writes signed zarf.yaml, then creates signature
- Proper rollback handling on signing failures
Build Metadata:
- Added Signed *bool field to ZarfBuildData to track package signing status
Cosign Utilities:
- New SignBlobOptions struct with ShouldSign() helper
- CosignSignBlobWithOptions() wrapper for consistent signing interface
- DefaultSignBlobOptions() for standard signing configuration
Assembly Process:
- Refactored to use new PackageLayout.SignPackage() method
- Removed standalone signPackage() function
- Signing now integrated with package layout lifecycle
## Testing & Documentation
E2E Test: New test case for package signing workflow
Unit Tests: Comprehensive coverage for signing logic (success, failure, rollback scenarios)
Examples
```bash
# Sign an unsigned package
zarf package sign zarf-package-demo-amd64-1.0.0.tar.zst --signing-key ./private-key.pem

# Re-sign with a new key
zarf package sign package.tar.zst --signing-key ./new-key.pem --overwrite

# Sign and publish to OCI
zarf package sign package.tar.zst --signing-key ./key.pem --output oci://ghcr.io/org/packages
```
## Related Issue

Fixes #3959

## Checklist before merging

### DIFF
--- a/src/api/v1alpha1/package.go
+++ b/src/api/v1alpha1/package.go
@@ -267,12 +267,12 @@ type ZarfBuildData struct {
 	DifferentialPackageVersion string `json:"differentialPackageVersion,omitempty"`
 	// List of components that were not included in this package due to differential packaging.
 	DifferentialMissing []string `json:"differentialMissing,omitempty"`
-	// The minimum version of Zarf that does not have breaking package structure changes.
-	LastNonBreakingVersion string `json:"lastNonBreakingVersion,omitempty"`
 	// The flavor of Zarf used to build this package.
 	Flavor string `json:"flavor,omitempty"`
 	// Whether this package was signed
 	Signed *bool `json:"signed,omitempty"`
+	// Requirements for specific package operations.
+	VersionRequirements []VersionRequirement `json:"versionRequirements,omitempty"`
 }
 
 // ZarfValues imports package-level values files and validation.
@@ -281,4 +281,12 @@ type ZarfValues struct {
 	Files []string `json:"files,omitempty"`
 	// Schema is a placeholder field for importing a .json.schema file for imported Values files.
 	Schema string `json:"schema,omitempty"`
+}
+
+// VersionRequirement specifies minimum version requirements for the package
+type VersionRequirement struct {
+	// The minimum version of Zarf required to use this package
+	Version string `json:"version"`
+	// Explanation for why this version is required
+	Reason string `json:"reason,omitempty"`
 }
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -79,6 +79,7 @@ type packageCreateOptions struct {
 	signingKeyPassword      string
 	flavor                  string
 	ociConcurrency          int
+	skipVersionCheck        bool
 	withBuildMachineInfo    bool
 }
 
@@ -117,6 +118,8 @@ func newPackageCreateCommand(v *viper.Viper) *cobra.Command {
 	cmd.Flags().IntVarP(&o.maxPackageSizeMB, "max-package-size", "m", v.GetInt(VPkgCreateMaxPackageSize), lang.CmdPackageCreateFlagMaxPackageSize)
 	cmd.Flags().StringSliceVar(&o.registryOverrides, "registry-override", GetStringSlice(v, VPkgCreateRegistryOverride), lang.CmdPackageCreateFlagRegistryOverride)
 	cmd.Flags().StringVarP(&o.flavor, "flavor", "f", v.GetString(VPkgCreateFlavor), lang.CmdPackageCreateFlagFlavor)
+	cmd.Flags().BoolVar(&o.skipVersionCheck, "skip-version-check", false, "Ignore version requirements when deploying the package")
+	_ = cmd.Flags().MarkHidden("skip-version-check")
 
 	cmd.Flags().StringVar(&o.signingKeyPath, "signing-key", v.GetString(VPkgCreateSigningKey), lang.CmdPackageCreateFlagSigningKey)
 	cmd.Flags().StringVar(&o.signingKeyPassword, "signing-key-pass", v.GetString(VPkgCreateSigningKeyPassword), lang.CmdPackageCreateFlagSigningKeyPassword)
@@ -218,6 +221,7 @@ func (o *packageCreateOptions) run(ctx context.Context, args []string) error {
 		RemoteOptions:           defaultRemoteOptions(),
 		CachePath:               cachePath,
 		IsInteractive:           !o.confirm,
+		SkipVersionCheck:        o.skipVersionCheck,
 		WithBuildMachineInfo:    o.withBuildMachineInfo,
 	}
 	pkgPath, err := packager.Create(ctx, baseDir, o.output, opt)
@@ -244,6 +248,7 @@ type packageDeployOptions struct {
 	optionalComponents      string
 	shasum                  string
 	skipSignatureValidation bool
+	SkipVersionCheck        bool
 	ociConcurrency          int
 	publicKeyPath           string
 }
@@ -277,6 +282,8 @@ func newPackageDeployCommand(v *viper.Viper) *cobra.Command {
 	cmd.Flags().StringVar(&o.shasum, "shasum", v.GetString(VPkgDeployShasum), lang.CmdPackageDeployFlagShasum)
 	cmd.Flags().StringVarP(&o.namespaceOverride, "namespace", "n", v.GetString(VPkgDeployNamespace), lang.CmdPackageDeployFlagNamespace)
 	cmd.Flags().BoolVar(&o.skipSignatureValidation, "skip-signature-validation", false, lang.CmdPackageFlagSkipSignatureValidation)
+	cmd.Flags().BoolVar(&o.SkipVersionCheck, "skip-version-check", false, "Ignore version requirements when deploying the package")
+	_ = cmd.Flags().MarkHidden("skip-version-check")
 
 	return cmd
 }
@@ -339,6 +346,7 @@ func (o *packageDeployOptions) run(cmd *cobra.Command, args []string) (err error
 		NamespaceOverride:      o.namespaceOverride,
 		RemoteOptions:          defaultRemoteOptions(),
 		IsInteractive:          !o.confirm,
+		SkipVersionCheck:       o.SkipVersionCheck,
 	}
 
 	deployedComponents, err := deploy(ctx, pkgLayout, deployOpts, o.setVariables, o.optionalComponents)
@@ -670,7 +678,6 @@ func newPackageInspectCommand(v *viper.Viper) *cobra.Command {
 	cmd.Flags().StringVar(&o.sbomOutputDir, "sbom-out", "", lang.CmdPackageInspectFlagSbomOut)
 	cmd.Flags().BoolVar(&o.listImages, "list-images", false, lang.CmdPackageInspectFlagListImages)
 	cmd.Flags().BoolVar(&o.skipSignatureValidation, "skip-signature-validation", false, lang.CmdPackageFlagSkipSignatureValidation)
-
 	return cmd
 }
 
@@ -1232,6 +1239,7 @@ type packageRemoveOptions struct {
 	confirm                 bool
 	optionalComponents      string
 	skipSignatureValidation bool
+	skipVersionCheck        bool
 	ociConcurrency          int
 	publicKeyPath           string
 	valuesFiles             []string
@@ -1258,6 +1266,8 @@ func newPackageRemoveCommand(v *viper.Viper) *cobra.Command {
 	cmd.Flags().StringVar(&o.optionalComponents, "components", v.GetString(VPkgDeployComponents), lang.CmdPackageRemoveFlagComponents)
 	cmd.Flags().StringVarP(&o.namespaceOverride, "namespace", "n", v.GetString(VPkgDeployNamespace), lang.CmdPackageRemoveFlagNamespace)
 	cmd.Flags().BoolVar(&o.skipSignatureValidation, "skip-signature-validation", false, lang.CmdPackageFlagSkipSignatureValidation)
+	cmd.Flags().BoolVar(&o.skipVersionCheck, "skip-version-check", false, "Ignore version requirements when removing the package")
+	_ = cmd.Flags().MarkHidden("skip-version-check")
 	cmd.Flags().StringSliceVarP(&o.valuesFiles, "values", "v", []string{}, "Path to values file(s) for removal actions")
 	cmd.Flags().StringToStringVar(&o.setValues, "set-values", map[string]string{}, "Set specific values via command line (format: key.path=value)")
 
@@ -1322,6 +1332,7 @@ func (o *packageRemoveOptions) run(cmd *cobra.Command, args []string) error {
 		Cluster:           c,
 		Timeout:           config.ZarfDefaultTimeout,
 		NamespaceOverride: o.namespaceOverride,
+		SkipVersionCheck:  o.skipVersionCheck,
 		Values:            vals,
 	}
 	logger.From(ctx).Info("loaded package for removal", "name", pkg.Metadata.Name)
@@ -1355,6 +1366,7 @@ type packagePublishOptions struct {
 	confirm                 bool
 	ociConcurrency          int
 	publicKeyPath           string
+	skipVersionCheck        bool
 	withBuildMachineInfo    bool
 }
 
@@ -1378,6 +1390,8 @@ func newPackagePublishCommand(v *viper.Viper) *cobra.Command {
 	cmd.Flags().StringVarP(&o.flavor, "flavor", "f", v.GetString(VPkgCreateFlavor), lang.CmdPackagePublishFlagFlavor)
 	cmd.Flags().IntVar(&o.retries, "retries", v.GetInt(VPkgPublishRetries), lang.CmdPackageFlagRetries)
 	cmd.Flags().BoolVarP(&o.confirm, "confirm", "c", false, lang.CmdPackagePublishFlagConfirm)
+	cmd.Flags().BoolVar(&o.skipVersionCheck, "skip-version-check", false, "Ignore version requirements when publishing the package")
+	_ = cmd.Flags().MarkHidden("skip-version-check")
 	cmd.Flags().BoolVar(&o.withBuildMachineInfo, "with-build-machine-info", v.GetBool(VPkgPublishWithBuildMachineInfo), lang.CmdPackageCreateFlagWithBuildMachineInfo)
 
 	return cmd
@@ -1425,6 +1439,7 @@ func (o *packagePublishOptions) run(cmd *cobra.Command, args []string) error {
 			RemoteOptions:        defaultRemoteOptions(),
 			CachePath:            cachePath,
 			Flavor:               o.flavor,
+			SkipVersionCheck:     o.skipVersionCheck,
 			WithBuildMachineInfo: o.withBuildMachineInfo,
 		}
 		_, err = packager.PublishSkeleton(ctx, packageSource, dstRef, skeletonOpts)
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/agnivade/levenshtein v1.2.1
 	github.com/anchore/clio v0.0.0-20250408180537-ec8fa27f0d9f
 	github.com/anchore/stereoscope v0.1.12
-	github.com/anchore/syft v1.36.0
+	github.com/anchore/syft v1.37.0
 	github.com/avast/retry-go/v4 v4.7.0
 	github.com/defenseunicorns/pkg/helpers/v2 v2.0.4
 	github.com/defenseunicorns/pkg/oci v1.3.0
@@ -145,7 +145,7 @@ require (
 	github.com/in-toto/attestation v1.1.2 // indirect
 	github.com/itchyny/gojq v0.12.17 // indirect
 	github.com/itchyny/timefmt-go v0.1.6 // indirect
-	github.com/jedib0t/go-pretty/v6 v6.6.8 // indirect
+	github.com/jedib0t/go-pretty/v6 v6.6.9 // indirect
 	github.com/lestrrat-go/blackmagic v1.0.4 // indirect
 	github.com/lestrrat-go/dsig v1.0.0 // indirect
 	github.com/lestrrat-go/dsig-secp256k1 v1.0.0 // indirect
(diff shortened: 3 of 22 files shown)


src/api/v1alpha1/package.go
```diff
@@ -271,6 +271,8 @@ type ZarfBuildData struct {
 	LastNonBreakingVersion string `json:"lastNonBreakingVersion,omitempty"`
 	// The flavor of Zarf used to build this package.
 	Flavor string `json:"flavor,omitempty"`
+	// Whether this package was signed
+	Signed *bool `json:"signed,omitempty"`
```
AustinAbro321, line 273: I think we should have an `IsSigned` method similar to our other boolean pointers. Makes it clear what the default should be when the pointer is nil, and it's more convenient to call

src/cmd/package.go
```diff
@@ -1536,6 +1538,173 @@ func (o *packagePullOptions) run(cmd *cobra.Command, args []string) error {
...
+}
+
+func (o *packageSignOptions) run(cmd *cobra.Command, args []string) error {
+	ctx := cmd.Context()
+	l := logger.From(ctx)
+	packageSource := args[0]
+
+	if o.signingKeyPath == "" {
+		return errors.New("--signing-key is required")
+	}
```
AustinAbro321, line 1627: nit: use cmd.MarkFlagRequired

src/pkg/packager/layout/package_test.go
```diff
@@ -200,3 +203,556 @@ func TestPackageFileName(t *testing.T) {
...
+		require.True(t, *updatedPkg.Build.Signed, "zarf.yaml should have signed:true")
+
+		// Also verify in-memory state matches
+		require.NotNil(t, pkgLayout.Pkg.Build.Signed)
+		require.True(t, *pkgLayout.Pkg.Build.Signed)
+	})
+}
+
+// TestPackageLayoutSignPackageValidation uses table-driven tests for validation scenarios
+func TestPackageLayoutSignPackageValidation(t *testing.T) {
```
AustinAbro321, line 548: I think these setup functions obscure the flow and make it hard to follow. Is it instead possible to set values in the test struct?

src/pkg/packager/layout/package.go
```diff
@@ -124,9 +125,136 @@ func (p *PackageLayout) ContainsSBOM() bool {
...
-	return signPackage(p.dirPath, signingKeyPath, signingKeyPassword)
+// SignPackage signs the zarf package using cosign with the provided options.
+// If the options do not indicate signing should be performed (no key material configured),
+// this is a no-op and returns nil.
+//
+// This function:
+// 1. Updates Pkg.Build.Signed = true in memory
+// 2. Writes the updated zarf.yaml (with signed:true) to a temporary file
+// 3. Signs the temporary file
+// 4. If signing succeeds, replaces the actual zarf.yaml with the signed version
```
AustinAbro321: Generally for doc comments on public functions I think we should highlight the outcomes caused by the functions rather than the implementation details. To me, uses a temporary file is an implementation detail.

src/pkg/packager/layout/package.go
```diff
@@ -124,9 +125,136 @@ func (p *PackageLayout) ContainsSBOM() bool {
...
+	if err != nil {
+		// Rollback in-memory state
+		p.Pkg.Build.Signed = originalSigned
+		return fmt.Errorf("failed to sign package: %w", err)
+	}
+
+	// Signing succeeded - now atomically replace the actual files
+
+	// Copy signed zarf.yaml from temp to actual location
+	err = helpers.CreatePathAndCopy(tmpZarfYAMLPath, zarfYAMLPath)
```
AustinAbro321: This is a path where the parent directory definitely already exists as we validate that p.dirPath already exists. I think it would be more explicit to use an os.WriteFile function or whichever os function allows overwrites

src/pkg/packager/layout/package.go
```diff
@@ -124,9 +125,136 @@ func (p *PackageLayout) ContainsSBOM() bool {
...
+	if err != nil {
+		// This is a critical error - signing succeeded but we can't update the file
+		// Keep the signed:true state as it reflects what we intended
+		return fmt.Errorf("failed to update %s after signing: %w", ZarfYAML, err)
+	}
+
+	// Copy signature from temp to actual location
+	err = helpers.CreatePathAndCopy(tmpSignaturePath, actualSignaturePath)
```
AustinAbro321: Same helpers.CreatePathAndCopy vs os.WriteFile comment

src/pkg/packager/layout/package.go
```diff
@@ -124,9 +125,136 @@ func (p *PackageLayout) ContainsSBOM() bool {
...
+	if err != nil {
+		// Revert the zarf.yaml if signature copy fails
+		// We need to restore the original zarf.yaml
+		p.Pkg.Build.Signed = originalSigned
```
AustinAbro321: Significant complexity is added to this function for the reverting logic. I think it could be simplified by only setting `p.Pkg.Build.Signed` at the end of the function. 

Still I wonder if the revert logic is worth having. File operations going awry is very hard to test. It's technically possible for this to work, but a writeFile would have to fail, then another writeFile will have to succeed. It seems very unlikely

src/api/v1alpha1/package.go
brandtkeller: Good catch - focus has been on setting but getting is going to be highly relevant in the verification workflows.

src/cmd/package.go
brandtkeller: Open to changing this - primarily setting the conditional here instead currently as we expect the required inputs to change when we enable more signing optionality. Certainly we can set and remove required flags but this was a quality gate for future implementation.

src/pkg/packager/layout/package.go
brandtkeller: I agree - I was being overly verbose here so that the nuance of this operation does not go unnoticed - signing and modifying the definition contains an interesting edgecase whereby the signed key needs to be set _before_ it is actually signed.

src/pkg/packager/layout/package.go
brandtkeller: We both need to set the signed field to `true` _and_ write the file for the signing logic to use said file _before_ signing the file - hence comments on file writing complexities that made still need some refinement.

src/pkg/packager/layout/package_test.go
brandtkeller: Good feedback - I've been thinking something about our test management is actually more ambiguous and thus I tried this model. 

The root of the issue being that a struct is difficult to provide flexibility in this scenario where we need to get defaults and isolate individual changes. We could use our standard helpers patterns but that leads to more external-from-the-test logic. fair tradeoff nonetheless.

src/pkg/packager/layout/package.go
brandtkeller: Moved the doccoment internal to the function - can be removed in the future.

src/pkg/packager/layout/package.go
brandtkeller: Updated to use rename

src/pkg/packager/layout/package.go
brandtkeller: Updated to use rename

src/pkg/packager/layout/package.go
AustinAbro321: Gotcha, missed that. 

I still struggle with the need for a rollback. When someone loads a package layout, it's always loaded into a temporary directory. For CLI users we will always fail the command either way. SDK users should do the same as they have no way to delineate against an error that they can continue from and an error which left their package in a bad state. While we could add custom errors, I think continuing on error is not a behavior we'd want to encourage.

src/pkg/utils/cosign.go
```diff
@@ -23,58 +24,391 @@ import (
...
 
-// CosignVerifyBlob verifies the zarf.yaml.sig was signed with the key provided by the flag
-func CosignVerifyBlob(ctx context.Context, blobRef, sigRef, keyPath string) error {
-	keyOptions := options.KeyOpts{KeyRef: keyPath}
-	cmd := &verify.VerifyBlobCmd{
-		KeyOpts:    keyOptions,
-		SigRef:     sigRef,
+// SignBlobOptions contains all options for cosign sign-blob operations.
+// This structure supports all cosign v3 sign-blob capabilities.
+type SignBlobOptions struct {
```
AustinAbro321, line 35: Are these options from a cosign struct?

src/test/e2e/12_package_signing_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestPackageSigning(t *testing.T) {
+	t.Log("E2E: Package Signing")
+
+	t.Run("Signing a basic package", func(t *testing.T) {
```
AustinAbro321, line 18: We can add `t.Parallel` here

src/pkg/utils/cosign.go
brandtkeller: From multiple - particularly key options and root options 

See [here](https://github.com/zarf-dev/zarf/pull/4301/files#diff-cc6c1a073c984367a4ed48d25fd8397593b06b5a581b5103fc1705a7b6dda389R279-R324)

src/pkg/packager/layout/package.go
brandtkeller: I agree that we don't need the rollback here. I have concerns that we desynchronize the package by not rolling back - but it would only be during complex use cases where the SDK user was trying to handle errors instead of re-building from scratch.

src/test/e2e/12_package_signing_test.go
brandtkeller: added! good catch

src/pkg/packager/layout/package.go
brandtkeller: Updated

src/pkg/packager/layout/package.go
```diff
@@ -124,9 +125,125 @@ func (p *PackageLayout) ContainsSBOM() bool {
...
+
+	// Save the original signed state in case we need to rollback
+	var originalSigned *bool
+	if p.Pkg.Build.Signed != nil {
+		val := *p.Pkg.Build.Signed
+		originalSigned = &val
+	}
+
+	// Create temporary directory for signing
+	tmpDir, err := utils.MakeTempDir(config.CommonOptions.TempDirectory)
```
AustinAbro321, line 176: Now that we are not rolling back from a temporary directory is there still value in using a temp directory? Seems like it would be more straight forward to directly sign

brandtkeller: Unless we take a separate stance then we currently still require the temporary location:

When attempting to sign a package - it is not yet signed - so we need to write a temporary manifest with the `build.signed: true` data -> then we sign it and overwrite the original manifest.

We could overwrite the existing manifest during this operation but if it fails then we would place inaccurate information in the manifest.

src/pkg/packager/layout/package_test.go
```diff
@@ -200,3 +203,556 @@ func TestPackageFileName(t *testing.T) {
...
+		opts.PassFunc = passFunc
+
+		err = pkgLayout.SignPackage(ctx, opts)
+		require.Error(t, err)
+
+		// Verify Signed field was not set
+		require.Nil(t, pkgLayout.Pkg.Build.Signed)
+	})
+
+	t.Run("Signed field set to true on success", func(t *testing.T) {
```
AustinAbro321: I think we can delete this test. It's the same as "successful signing", except it doesn't check the signedPath.

src/pkg/packager/layout/package.go
```diff
@@ -124,9 +125,125 @@ func (p *PackageLayout) ContainsSBOM() bool {
...
+		return fmt.Errorf("invalid package layout: %s is not a directory", p.dirPath)
+	}
+
+	// Verify zarf.yaml exists before signing
+	zarfYAMLPath := filepath.Join(p.dirPath, ZarfYAML)
+	if _, err := os.Stat(zarfYAMLPath); err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("cannot sign package: %s not found in package layout", ZarfYAML)
+		}
+		return fmt.Errorf("cannot access %s for signing: %w", ZarfYAML, err)
```
AustinAbro321: I think it's simpler to have a single error message here

src/pkg/packager/layout/package_test.go
AustinAbro321: Fair enough no change requested

src/pkg/utils/cosign.go
AustinAbro321: Would it be better to have a struct that inherits the values like so
```go
type SignBlobOptions struct {
	options.KeyOpts
	options.RootOptions
}
```

brandtkeller: I am not opposed to it. There are pros/cons but my current thought was SDK developer experience. Do we want to require SDK users to need a direct cosign import to accomplish these operations? I don't have a hard opinion but I like the perspective.

AustinAbro321: I think so, with the understanding that these keys are directly from cosign, they have confidence in using the cosign documentation and can look at other examples of cosign usage in the ecosystem. 

Importing the cosign objects also keeps us up to date in the case that fields on these structs change

src/pkg/packager/layout/package.go
AustinAbro321: Fair. I was initially thinking it's okay to leave a package in a bad state when we error, but yeah there could be cases where an SDK user continues using the package

src/pkg/packager/layout/package.go
```diff
@@ -124,9 +125,125 @@ func (p *PackageLayout) ContainsSBOM() bool {
...
+	// This ensures the zarf.yaml metadata accurately reflects the signed state and the
+	// signature is valid for the zarf.yaml content that includes signed:true.
+
+	l := logger.From(ctx)
+
+	// Check if signing should be performed based on the options
+	// this is a no-op as there may be many different ways to sign
+	// input validation should be performed in the calling function
+	if !opts.ShouldSign() {
+		l.Info("skipping package signing (no signing key material configured)")
```
AustinAbro321, line 148: Given that this line now prints on `zarf package create` for unsigned packages, I think there is a potential impact on user psychology. This will likely serve as a call to action for users who are not currently signing their package to look into what it takes to sign a package. I know there are improvements to signing packages in the pipeline. Keeping this as a debug log until we have newer, documented, signing methods we want to encourage, could be a boon on future signing adoption. User's will likely not look into signing twice.  

Not formally requesting a change, but I do think this is an interesting topic.

src/pkg/utils/cosign.go
brandtkeller: If we update the Default Options functions to return the required embedded cosign options then I think most of my concerns to DevEx are handled.

brandtkeller: Updated

src/pkg/packager/layout/package.go
brandtkeller: Updated

src/pkg/packager/layout/package_test.go
brandtkeller: Agreed

src/pkg/packager/layout/package.go
brandtkeller: I agree with the sentiment - although I think things like signing, SBOMs, and any other attestations will continue to be reminded and required as regulation continues or begins to require as much. Not overpromising current capabilities is always a good reminder.

src/cmd/package.go
```diff
@@ -1560,6 +1562,173 @@ func (o *packagePullOptions) run(cmd *cobra.Command, args []string) error {
...
+			// For OCI sources, use current working directory
+			wd, err := os.Getwd()
+			if err != nil {
+				return err
+			}
+			outputDest = wd
+		} else {
+			// For file sources, use the same directory as the source
+			outputDest = filepath.Dir(packageSource)
+		}
```
AustinAbro321, line 1655: As I'm trying out the UX of this command something that confused me was that when I ran the command on a local package it signed my package in place by default, however when I ran the command on a package in a registry, it gave me a copy of the package. I assumed that the package in the registry was signed and got confused when it wasn't. Thoughts on signing the package in the registry by default?

src/cmd/package.go
```diff
@@ -1560,6 +1562,173 @@ func (o *packagePullOptions) run(cmd *cobra.Command, args []string) error {
...
+	signOpts.KeyRef = o.signingKeyPath
+	signOpts.PassFunc = passFunc
+
+	err = pkgLayout.SignPackage(ctx, signOpts)
+	if err != nil {
+		return fmt.Errorf("failed to sign package: %w", err)
+	}
+
+	// Handle output - OCI or local file
+	if helpers.IsOCIURL(outputDest) {
```
AustinAbro321: This could be a future enhancement, but one option to eliminate duplication is using `packagePublishOptions.Run()`

src/pkg/utils/cosign.go
AustinAbro321, line 1: This is definitely out of scope, but I think it would be a good issue would be to move this file from utils to it's own cosign package. Ideally the package would be internal as well

src/cmd/package.go
brandtkeller: That is good UX feedback. I'd say that is a reasonable and maybe optimal expectation.

src/pkg/utils/cosign.go
brandtkeller: I am not against doing that now - it still provides surface area to support an embedded `zarf tools sign-blob` command if requested.

brandtkeller: Updated

src/cmd/package.go
brandtkeller: added

src/cmd/package.go
brandtkeller: Updated

brandtkeller:
Updated to include a few UX/DevEx Enhancements
- Re-publish to OCI when source is OCI by default
- Move cosign to an internal cosign package
- Remove the cosign required import for SDK DeVex with a small helper

src/pkg/packager/layout/package.go
```diff
@@ -124,9 +126,142 @@ func (p *PackageLayout) ContainsSBOM() bool {
 	return !helpers.InvalidPath(filepath.Join(p.dirPath, SBOMTar))
 }
 
-// SignPackage signs the zarf package
-func (p *PackageLayout) SignPackage(signingKeyPath, signingKeyPassword string) error {
-	return signPackage(p.dirPath, signingKeyPath, signingKeyPassword)
+// SignPackage signs the zarf package using cosign with the provided options.
+// If the options do not indicate signing should be performed (no key material configured),
+// this is a no-op and returns nil.
+func (p *PackageLayout) SignPackage(ctx context.Context, opts zarfCosign.SignBlobOptions) (err error) {
```
AustinAbro321: I think this function should be callable by SDK users. Since cosign is an internal package atm, users won't be able to create the necessary struct to call it (sorry I didn't think of that when I made the internal recommendation). 

This means we could either create another struct with a subset of options exposed, or make the cosign package public. I would lean towards the latter option.

src/internal/cosign/cosign.go
```diff
@@ -0,0 +1,259 @@
...
+// For new code, use CosignSignBlobWithOptions for full control.
+func CosignSignBlob(blobPath, outputSigPath, keyPath string, passFn cosign.PassFunc) ([]byte, error) {
+	ctx := context.Background()
+	opts := DefaultSignBlobOptions()
+	opts.KeyRef = keyPath
+	opts.PassFunc = passFn
+	opts.OutputSignature = outputSigPath
+	return CosignSignBlobWithOptions(ctx, blobPath, opts)
+}
+
```
AustinAbro321: I think we can delete these functions now, even if we make this package public since any potential users would have to change function signatures anyway, they may as well change to the new function signatures

src/cmd/package.go
```diff
@@ -1567,6 +1569,197 @@ func (o *packagePullOptions) run(cmd *cobra.Command, args []string) error {
...
+	defer func() {
+		if cleanupErr := pkgLayout.Cleanup(); cleanupErr != nil {
+			l.Warn("failed to cleanup package layout", "error", cleanupErr)
+		}
+	}()
+
+	// Check for existing signature and handle overwrite logic
+	sigPath := filepath.Join(pkgLayout.DirPath(), layout.Signature)
+	_, err = os.Stat(sigPath)
+	sigExists := err == nil
```
AustinAbro321: I believe we can use pkgLayout.IsSigned here

src/cmd/package.go
AustinAbro321: I believe this block can be deleted now

AustinAbro321 (changes requested):
Another iteration. This is getting close

src/pkg/packager/layout/package.go
brandtkeller: ah great point - I keep waffling on remembering to support that requirement.

src/internal/cosign/cosign.go
brandtkeller: agreed - now that we have parity I think this is fair.

src/cmd/package.go
brandtkeller: I believe this is still required - when output destination has not been specified the first block processes package source and if it is an OCI source it sets the output destination so that this block can then process the publishing whether output destination is defined or defaulted to based on oci source.

src/internal/cosign/cosign.go
brandtkeller: I'm good with removing them in favor of not maintaining them unless needed.

src/cmd/package.go
brandtkeller: ah there is duplicate -removing

AustinAbro321 (dismissed):
Nice work, I'm glad you evaluated the ecosystem as you looked at adding this feature

brandtkeller:
Merge conflicts resolved.

AustinAbro321 (approved):

PR #4306, opened 2025-10-23
fix: support remote valuesFiles

### LINKED ISSUE #3984: Pull valuesFiles from git
We have all `value.yaml` files that belong to helm charts in a separate git repository.
The reason for this is that applications are also installed by argocd and this improves clarity and is also recommended by argocd.

Because we also want to use `zarf dev find-images` to always load the appropriate docker images for our environments, the values.yaml files are required here. The reason for this is that individual features of applications are sometimes enabled or disabled, which in turn also affects the docker images used.

Because we would like to keep managing the value files only in one central location (config git repository), it would be good to extend the functionality of `valuesFiles` so that git repositories can also be specified.

An example of this is how you can define it in an Argocd application
```
  sources:
    - repoURL: 'zarf-docker-registry.zarf.svc.cluster.local:5000'
      chart: prometheus/prometheus
      targetRevision: 27.13.0
      helm:
        valueFiles:
        - $values/helm-values/prometheus/values.yaml
    - repoURL: 'https://gitlab.example.com/example-repo/cloud-config.git'
      targetRevision: HEAD
      ref: values
```

If I see it correctly, then I can also specify a git repo as url for manifest, for example.
```
    manifests:
      - name: flux-install
        namespace: flux
        files:
          - https://github.com/fluxcd/flux2/releases/download/v2.4.0/install.yaml
```

The current status seems to be that no url can be specified in the `valuesFiles`
```
- name: prometheus
  version: 27.23.0
  namespace: monitoring
  url: oci://registry.example.com/helm-charts/oci-images/prometheus/prometheus
  valuesFiles:
    - 'http://gitlab.example.com/my-org/-/blob/main/helm-values/prometheus/values.yaml?ref_type=heads'
 ```

ERR failed to create package: unable to download the published chart "prometheus": unable to process the values for the package: unable to copy chart values file https:/gitlab.example.com/my-org/-/blob/main/helm-values/prometheus/values.yaml?ref_type=heads: lstat https:/gitlab.example.com/my-org/-/blob/main/helm-values/prometheus/values.yaml?ref_type=heads: no such file or directory

- **Given** a git repository url (with a tag and including subdirectories) in the `valuesFiles` in the `zarf.yaml`
- **When** using `zarf dev find-images` or `zarf package create` the defined values.yaml should be taken from that git repo / url
- **Then** the `values.yaml` file from the git/url should be used to find the appropriate docker images, for example

### PR DESCRIPTION
**Pull request by michalszynkiewicz**
## Description

Fixed support for remote valuesFiles in zarf.yaml.

Before the change, the code assumed that if a path doesn't start from `/`, it is a relative path and has to be preprocessed.

### Automatic
Added a dedicated tests `TestPackageInspectRemoteValuesFiles` that exposes an http server that serves a values file, and builds the package with these values.

### Manual
Used keycloak package for tests. Exposed the values yaml with `python3 -m http.server 8000`. Altered keycloak Zarf package locally to use the values from http. Built the package.

## Related Issue

Fixes #3984

## Checklist before merging

### DIFF
--- a/src/cmd/package_test.go
+++ b/src/cmd/package_test.go
@@ -320,11 +320,7 @@ func TestPackageInspectRemoteValuesFiles(t *testing.T) {
 		definitionDir:  srcDir,
 		expectedOutput: filepath.Join("testdata", "inspect-values-files", "chart-remote", "expected.yaml"),
 		kubeVersion:    "1.25",
-		setVariables: map[string]string{
-			"REPLICAS":    "2",
-			"DESCRIPTION": ".chart.variables takes priority",
-			"PORT":        "8888",
-		},
+		setVariables:   map[string]string{},
 	}
 
 	checkPackageValuesInspectFiles(t, test)
--- a/src/cmd/testdata/inspect-values-files/chart-remote/zarf-template.yaml
+++ b/src/cmd/testdata/inspect-values-files/chart-remote/zarf-template.yaml
@@ -14,21 +14,4 @@ components:
         localPath: chart
         valuesFiles:
           - VALUES_YAML_URL
-        variables:
-          - name: REPLICAS
-            path: replicaCount
-            description: "Override the number of pod replicas"
-          - name: DESCRIPTION
-            path: description
-            description: ".chart.variables takes priority"
 
-  - name: different-values-set
-    charts:
-      - name: htppd-local
-        version: 1.0.0
-        namespace: httpd
-        localPath: chart
-        variables:
-          - name: REPLICAS
-            path: replicaCount
-            description: "Override the number of pod replicas"
--- a/src/cmd/testdata/inspect-values-files/chart/remote-values/values.yaml
+++ b/src/cmd/testdata/inspect-values-files/chart/remote-values/values.yaml
@@ -1,4 +1,0 @@
-image:
-  pullPolicy: IfNotPresent
-description: "this should be overwritten by .chart.variables"
-port: 8888
--- a/src/pkg/packager/layout/assemble.go
+++ b/src/pkg/packager/layout/assemble.go
@@ -548,7 +548,7 @@ func assembleSkeletonComponent(ctx context.Context, component v1alpha1.ZarfCompo
 				path = filepath.Join(packagePath, path)
 			}
 			if err := helpers.CreatePathAndCopy(path, filepath.Join(compBuildPath, rel)); err != nil {
-				return fmt.Errorf("unable to copy chart values file for assembling skeleton component %s: %w", path, err)
+				return fmt.Errorf("unable to copy chart values file %s: %w", path, err)
 			}
 		}
 	}
--- a/src/cmd/testdata/inspect-values-files/chart-remote/chart/values.yaml
+++ b/src/cmd/testdata/inspect-values-files/chart-remote/chart/values.yaml
@@ -1,1 +1,0 @@
-description: "values in this file will not show on zarf package inspect values-files since they are packaged and become part of the main chart"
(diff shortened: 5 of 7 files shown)

### CODE REVIEW

File: src/pkg/packager/layout/assemble.go [hunk 1]
```diff
@@ -548,7 +548,7 @@ func assembleSkeletonComponent(ctx context.Context, component v1alpha1.ZarfCompo
 				path = filepath.Join(packagePath, path)
 			}
 			if err := helpers.CreatePathAndCopy(path, filepath.Join(compBuildPath, rel)); err != nil {
-				return fmt.Errorf("unable to copy chart values file %s: %w", path, err)
+				return fmt.Errorf("unable to copy chart values file for assembling skeleton component %s: %w", path, err)
```
AustinAbro321: nit: Feels a bit redundant to add this to the error message. `AssembleSkeleton()` will already include `unable to create skeleton`

src/cmd/testdata/inspect-values-files/chart/remote-values/values.yaml
AustinAbro321, line 1: This file is not used

src/cmd/testdata/inspect-values-files/chart-remote/zarf-template.yaml
AustinAbro321, line 1: Lets make this test simpler to follow. The other tests wanted to test interactions between variables and multiple components, but this test only wants to check that remote values files work. 

I would recommend, deleting the variable, deleting the chart.variables, and deleting the second component

AustinAbro321:
In case anyone is wondering if you are pulling a file from git, you can use a commit sha. It looks like so https://raw.githubusercontent.com/stefanprodan/podinfo/6a069aed1abfa22077de0e780e58258ed6317351/charts/podinfo/values.yaml

src/cmd/testdata/inspect-values-files/chart/remote-values/values.yaml
michalszynkiewicz: it is the file that is exposed by http server here: https://github.com/zarf-dev/zarf/pull/4306/files#diff-9de8fbbff009c2f029b6b49167c0f0a341462f85d18090488d045eef05c6ce0eR305

AustinAbro321: Looks like that's referencing the path `/testdata/inspect-values-files/chart-remote/remote-values/values.yaml`, this file is `testdata/inspect-values-files/chart/remote-values/values.yaml`. Likely a duplicate when you copied over the chart.

michalszynkiewicz: oh, good catch, sorry :)

src/cmd/package_test.go
```diff
@@ -280,43 +293,78 @@ func TestPackageInspectValuesFiles(t *testing.T) {
...
+		name:           "chart inspect with remote values URL",
+		packageName:    "chart",
+		definitionDir:  srcDir,
+		expectedOutput: filepath.Join("testdata", "inspect-values-files", "chart-remote", "expected.yaml"),
+		kubeVersion:    "1.25",
+		setVariables: map[string]string{
+			"REPLICAS":    "2",
+			"DESCRIPTION": ".chart.variables takes priority",
+			"PORT":        "8888",
+		},
```
AustinAbro321: Now that we aren't using chart.variables, you can delete this

AustinAbro321:
@michalszynkiewicz One more small comment, and also you'll need to follow [DCO instructions](https://github.com/zarf-dev/zarf/pull/4306/checks?check_run_id=53616087056). Otherwise looks good

michalszynkiewicz:
added the missing signature to the rework commit

PR #4308, opened 2025-10-23
feat: add check version requirement

### LINKED ISSUE #4256: Add VersionRequirements at build time to prevent operations on incompatible versions

I believe Zarf should add a field `build.VersionRequirements` which will contain requirements for a package to run certain operations. 

This will work for new fields that change deploy, and changes to the package structure https://github.com/zarf-dev/zarf/issues/3813

This will not work on versions of Zarf created before this feature is shipped.

- **Given** A new feature `values` is released in version 0.65.0. I create a package with the values key.
- **When** I deploy my package on version 0.64.0
- **Then** I receive an error that I must use at least version 0.65.0 to deploy my package. 

### Details

The field will look like so and will be a list on Zarf package build data : 

```go
type OperationRequirement struct {
	// the minimum version of the Zarf CLI that can deploy the package
	Version string 
	// Reasons for why the package can't be deployed
	// EX: "values was not introduced until v0.64.0, package structure changed in v0.65.0"
	Reasons string
}
```

This field will apply to the following commands
- deploy
- remove
- create (when importing a skeleton package with operationalRequirements set)

There will be a hidden flag introduced `--bypass-version-requirement`, which a user is prompted with in the error message, if they'd like to continue at their own risk. 

#### Requirements specific to operations

There could be a case where a command will break on deploy, but not break on remove or publish. We could add an operations field to operationRequirements and only error on specific operations, however more often than not the command will break for all commands. Generally we should encourage users to upgrade when their packages are incompatible. It's simpler to error everywhere

#### .build.LastNonBreakingVersion
There is already a field `.build.LastNonBreakingVersion`, it currently does not do anything, so this should be easy to deprecate and replace with another name. 

This is in a similar problem space as https://github.com/zarf-dev/zarf/issues/2728, but in this situation we would error instead of warn. However, a new field does not necessarily imply that the package is not deployable, a new field in `.build` is not very consequential. Likewise a breaking change in how a field behaves, may stop a package from being deployable on previous versions.

AustinAbro321:
Commands that do not have an impact on the cluster such as inspect do not implement the version check. Additionally, Publish is very generic and not effected by the actual layout.

This removed `.build.LastNonBreakingChange` which was set but not used 

This is what the error message will look like to the user. I created got this by manually setting `.build.VersionRequirements` then creating and deploying that package

image

## Related Issue

Fixes #4256

## Checklist before merging

### DIFF
--- a/src/api/v1alpha1/package.go
+++ b/src/api/v1alpha1/package.go
@@ -248,9 +248,9 @@ type ZarfMetadata struct {
 // ZarfBuildData is written during the packager.Create() operation to track details of the created package.
 type ZarfBuildData struct {
 	// The machine name that created this package.
-	Terminal string `json:"terminal"`
+	Terminal string `json:"terminal,omitempty"`
 	// The username who created this package.
-	User string `json:"user"`
+	User string `json:"user,omitempty"`
 	// The architecture this package was created on.
 	Architecture string `json:"architecture"`
 	// The timestamp when this package was created.
--- a/site/src/content/docs/commands/zarf_package_create.md
+++ b/site/src/content/docs/commands/zarf_package_create.md
@@ -37,6 +37,7 @@ zarf package create [ DIRECTORY ] [flags]
       --signing-key string          Private key for signing packages. Accepts either a local file path or a Cosign-supported key provider
       --signing-key-pass string     Password to the private key used for signing packages
       --skip-sbom                   Skip generating SBOM for this package
+      --with-build-machine-info     Include build machine information (hostname and username) in the package metadata
 ```
 
 ### Options inherited from parent commands
--- a/site/src/content/docs/commands/zarf_package_publish.md
+++ b/site/src/content/docs/commands/zarf_package_publish.md
@@ -38,6 +38,7 @@ $ zarf package publish ./path/to/dir oci://my-registry.com/my-namespace
       --signing-key string          Private key for signing or re-signing packages with a new key. Accepts either a local file path or a Cosign-supported key provider
       --signing-key-pass string     Password to the private key used for publishing packages
       --skip-signature-validation   Skip validating the signature of the Zarf package
+      --with-build-machine-info     Include build machine information (hostname and username) in the package metadata
 ```
 
 ### Options inherited from parent commands
--- a/src/api/internal/v1beta1/package.go
+++ b/src/api/internal/v1beta1/package.go
@@ -154,9 +154,9 @@ type ZarfBuildData struct {
 	// Checksum of a checksums.txt file that contains checksums all the layers within the package.
 	AggregateChecksum string `json:"aggregateChecksum,omitempty"`
 	// The machine name that created this package.
-	Terminal string `json:"terminal"`
+	Terminal string `json:"terminal,omitempty"`
 	// The username who created this package.
-	User string `json:"user"`
+	User string `json:"user,omitempty"`
 	// The architecture this package was created on.
 	Architecture string `json:"architecture"`
 	// The timestamp when this package was created.
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -79,6 +79,7 @@ type packageCreateOptions struct {
 	flavor                  string
 	ociConcurrency          int
 	skipVersionCheck        bool
+	withBuildMachineInfo    bool
 }
 
 func newPackageCreateCommand(v *viper.Viper) *cobra.Command {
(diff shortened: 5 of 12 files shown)


src/internal/pkgcfg/pkgcfg.go
```diff
@@ -31,8 +31,6 @@ func Parse(ctx context.Context, b []byte) (v1alpha1.ZarfPackage, error) {
 
 // List of migrations tracked in the zarf.yaml build data.
 const (
-	// This should be updated when a breaking change is introduced to the Zarf package structure.  See: https://github.com/zarf-dev/zarf/releases/tag/v0.27.0
-	LastNonBreakingVersion   = "v0.27.0"
```
mkcp, line 35: 👍

src/pkg/packager/load/load.go
```diff
@@ -31,6 +31,8 @@ type DefinitionOptions struct {
 	CachePath string
 	// IsInteractive decides if Zarf can interactively prompt users through the CLI
 	IsInteractive bool
+	// BypassVersionCheck skips version requirement validation
+	BypassVersionCheck bool
```
mkcp: Thoughts on using "SkipVersionCheck" as the name for this field in various places? It offers some consistency with existing language.

src/internal/packager/requirements/requirements.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/zarf-dev/zarf/src/config"
+)
+
+// VersionRequirementsError is returned when operational requirements are not met
+type VersionRequirementsError struct {
+	Requirements   []v1alpha1.VersionRequirement
+	CurrentVersion string
+}
+
+func (e *VersionRequirementsError) Error() string {
```
mkcp, line 22: Non-blocking, more of a discussion item. There's an alternative path for this where we use a `NewVersionRequirementsError(...) VersionRequirementsError` constructor function, and calculate all of the necessary fields upfront. Then the error method can assemble the message string from the pre-calculated fields. I expect that this approach would be a bit more readable, but it is otherwise equivalent.

src/pkg/packager/load/load.go
AustinAbro321: Yeah I like that better

src/internal/packager/requirements/requirements.go
AustinAbro321: Not sure if this is what you mean but I moved the logic to determine the highest required version out of the error message, which I agree is better

src/api/v1alpha1/package.go
```diff
@@ -267,10 +267,10 @@ type ZarfBuildData struct {
 	DifferentialPackageVersion string `json:"differentialPackageVersion,omitempty"`
 	// List of components that were not included in this package due to differential packaging.
 	DifferentialMissing []string `json:"differentialMissing,omitempty"`
-	// The minimum version of Zarf that does not have breaking package structure changes.
-	LastNonBreakingVersion string `json:"lastNonBreakingVersion,omitempty"`
 	// The flavor of Zarf used to build this package.
 	Flavor string `json:"flavor,omitempty"`
+	// Requirements for specific package operations.
+	VersionRequirements []VersionRequirement `json:"versionRequirements,omitempty"`
```
brandtkeller, line 273: not a blocker - do we want to delineate between "how to build the package" and "How the package was built?" 

I see build data as uniquely serving the latter. So I wonder if this is best placed in the build data or if we should consider metadata as another suitable location.

AustinAbro321: I personally see build data as data that is gathered / determined / recorded at build time, but which the user should not set directly. I see this new field as similar to `.Build.Version`

brandtkeller: I would agree with that assessment - but I understood this change to be defined by the package author - how are we expecting to perform this without a user setting this in the build data of a package definition?

AustinAbro321: In the future, when we do add breaking changes, we'll need to see it. Likely in the [recordPackageMetadata](https://github.com/zarf-dev/zarf/blob/5c3453feb38ef77540262a39dcb4c5f35502c90b/src/pkg/packager/layout/assemble.go#L699) function.

For example, lets say this was an existing feature when the values feature was created. That function could check if a package has values, and if so add an entry to versionRequirements for v0.64.0

Merging this PR before we have any use cases, makes it so it'll work on v0.65.0 when in the future we do have use cases

brandtkeller: Ah I believe I misunderstood this being a calculated field versus author-defined. 

If I understand correctly - this field is intended to be used internally to generate these requirements to guard users :+1:

src/internal/packager/requirements/requirements.go
```diff
@@ -0,0 +1,103 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+// Package requirements validates that Zarf meets the version requirements defined by the package
+package requirements
```
brandtkeller, line 5: Couple Questions:

1. Do we have a expected example of how to utilize these when necessary? 
2. Are we using build data as state to manage collecting these requirements throughout the create lifecycle?
3. Should we add documentation to support more of a user-facing warning "skipping the version check may introduce...." .

AustinAbro321: 1. I see the setting of the data as something that will be done at create time in the [recordPackageMetadata](https://github.com/zarf-dev/zarf/blob/44bdaea2df8b44d364e9b1ceb97c2bf359f74f8f/src/pkg/packager/layout/assemble.go#L699) function. For instance, you may have a block like this:
```go
	var versionRequirements []v1alpha1.VersionRequirement
	for _, comp := range pkg.Components {
		for _, v := range comp.Charts {
			if v.Version == "" {
				versionRequirements = append(versionRequirements, v1alpha1.VersionRequirement{
					Version: "v0.65.0",
					Reason:  "the internal file structure of charts has changed in the v1beta1 schema. Zarf is backwards compatible with these changes from v0.65.0+",
				})
				break
			}
		}
	}
	pkg.Build.VersionRequirements = versionRequirements
```
2. I think this will be defined by whats in the package / general changes to Zarf, so we can always set it towards the end of create
3. I think so, especially when we have real use cases in the defined. Interested in your thoughts on how that might be structured.

AustinAbro321: Let me know if there's a better place to document this example, for now, I linked this block to issue #2245. As that's the first use case I can think of

brandtkeller: I think that all of the above is appropriate. For item 3 - my thought was that we want to surface each potential violation when skipped in an auditable fashion.

PR #4309, opened 2025-10-24
docs: remove warnings that Flux and ArgoCD webhooks are not stable

### LINKED ISSUE #2981: Track stability - Flux OCIRepository and HelmRepository mutation
### Feature to track
The agent mutation functionality for OCIRepositories and HelmRepositories is in alpha according to the [agent documentation](https://docs.zarf.dev/ref/init-package/#zarf-agent). This feature was introduced in [v0.36.0](https://github.com/zarf-dev/zarf/releases/tag/v0.36.0). It was placed in alpha, because of a bug in upstream helm. HelmRepositories with an insecure registry must use IP address instead of a hostname. Which causes Flux HelmRepositories to break if Zarf is using an external http registry with a hostname.

This was fixed in Helm, and a new flux release with the fix will be released once #4947 is implemented. At that point, we should move this feature into stable.

### LINKED ISSUE #2982: Track stability - ArgoCD applications & ArgoCD Repositories mutation
### Feature to track
ArgoCD applications and ArgoCD repositories are in beta according to the [agent documentation.](https://docs.zarf.dev/ref/init-package/#zarf-agent) This feature was released in [v0.29.0](https://github.com/zarf-dev/zarf/releases/tag/v0.29.0), and has been used by many in the community since.

AustinAbro321:
## Description

These features have been around for a long time and should be considered stable

## Related Issue

Fixes  #2981
Fixes #2982 

## Checklist before merging


AustinAbro321:
@brandtkeller Yeah we will need to keep up with new API versions for both our tests and examples. Not sure how we track that. We do import Flux however different api versions will be entirely different packages.  For argo we don't import the API at all. IIRC before they said to copy their objects out of their codebase, looks like they have different docs now, but I would not want to add all of those directives  https://argo-cd.readthedocs.io/en/stable/user-guide/import/

PR #4312, opened 2025-10-24
docs: mark differential as stable

AustinAbro321:
## Description

There is no issue for this, but differential has been around for well over a year and has seen at least some use. It is generally a simple feature to maintain, I vote that it should be considered stable

## Checklist before merging


brandtkeller:
Merging #4313 seems to have closed this PR? interesting... reopening.

PR #4318, opened 2025-10-27
chore: expand the LogWriter implementation

Racer159:
## Description

Changes the LogWriter to a more robust writer that handles log levels as well.

## Related Issue

Fixes #N/A

## Checklist before merging

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -254,7 +254,7 @@ require (
 	github.com/Azure/go-autorest/logger v0.2.1 // indirect
 	github.com/Azure/go-autorest/tracing v0.6.0 // indirect
 	github.com/AzureAD/microsoft-authentication-library-for-go v1.5.0 // indirect
-	github.com/BurntSushi/toml v1.5.0 // indirect
+	github.com/BurntSushi/toml v1.5.0
 	github.com/CycloneDX/cyclonedx-go v0.9.3 // indirect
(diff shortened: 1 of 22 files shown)


Racer159:
Note there is a "Break" if someone were reaching all the way into the `DebugWriter` (though I think that would be rare and is easy to fix).

src/types/runtime.go
```diff
@@ -16,4 +16,6 @@ type ZarfCommonOptions struct {
 	CachePath string
 	// Location Zarf should use as a staging ground when managing files and images for package creation and deployment
 	TempDirectory string
+	// Whether to prefer using the structured logger over printing to stdout/stderr (i.e. in actions or git repo pulls)
+	PreferLogger bool
```
AustinAbro321, line 20: While generally we've moved away from global config state, I am okay with it in this instance given that any user who is modifying this, will likely want this modified across all commands.

AustinAbro321:
Giving an example of what this will look like: 
commands like this: 
```yaml
              echo "multiline!"
              sleep 0.5
              echo "updates!"
              sleep 1
              echo "in!"
              sleep 0.5
              echo "realtime!"
              sleep 0.5

```
Will appear in terminal like this
image

brandtkeller (dismissed):
This is a nice iteration on a space where I've found logging lacking - LGTM

brandtkeller (approved):

PR #4330, opened 2025-10-30
docs: encourage unit tests in contribute section

AustinAbro321:
## Description

The current docs dis-encourage unit tests. I believe we should flip this approach and instead encourage them

## Checklist before merging

### DIFF
--- a/site/src/content/docs/contribute/testing.mdx
+++ b/site/src/content/docs/contribute/testing.mdx
@@ -142,11 +142,6 @@ func TestFooBarBaz(t *testing.T) {
 The end-to-end tests are run sequentially and the naming convention is set intentionally:
 
 - 00-19 tests run prior to `zarf init` (cluster not initialized).
-
-:::note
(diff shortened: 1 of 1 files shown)


site/src/content/docs/contribute/testing.mdx
```diff
@@ -115,41 +142,7 @@ func TestFooBarBaz(t *testing.T) {
 The end-to-end tests are run sequentially and the naming convention is set intentionally:
 
 - 00-19 tests run prior to `zarf init` (cluster not initialized).
-
-:::note
-Due to resource constraints in public GitHub runners, K8s tests are only performed on Linux.
```
AustinAbro321, line 120: Deleting this as I feel the intent of this was to say that we don't run on arm64. We do run k8s tests on arm64 now. Running on mac specifically isn't really relevant to Zarf, and running k8s tests on windows for Zarf is not possible

PR #4334, opened 2025-10-31
feat: support importing image archives into a package

### LINKED ISSUE #2181: Using local image archives no longer works

AustinAbro321:
## Description

Image archives will only be importable using the OCI-layout format.

Unless a user manually tags their OCI layout with annotations the only tools I found that will work with this natively are Docker save and podman save.

oras backup can export to tar, but does not give the repo name, only the ref / tag. ORAS seems amendable to changing this https://github.com/oras-project/oras/issues/1893. Skopeo does not add annotations, once oras has a strong stance, I plan on making an issue in skopeo.

The imageArchive object allows users to specify Images they want imported. If none are selected we will import every image in the archive. 

Image archives will error when used with skeleton packages. Given that skeleton packages imply the existence of a registry that users can push to, it should generally make more sense to push the image to that registry. Should users request image tars in skeleton packages, this can be follow up work. 

This sets the version requirement to v0.67.0, as this will not be backwards compatible with previous packages, since previous packages will not know to check the image archives field for the images during push. While we could add the images found during image archives scans to the `component.Images` field to make new packages compatible with older versions of Zarf, given that this is a net new feature, I believe this solution is cleaner and encourages user's to upgrade.

### replace image tar in images block

We could accept a tar file in the `.components[x].images` block and replace it to the image(s) that are inside of it. This was how Zarf originally worked with image tars. However this leads to poor tracability for a built package. Having an explicit field makes it clear how the package was built. 

### Only allow single image archives

One option to avoid having to check the annotation by only allowing a single image in each archive. However since the Docker Sha annotation issue is already solved in the new image store (https://github.com/moby/moby/issues/51389) and ORAS backups are likely to add more annotations (https://github.com/oras-project/oras/issues/1893), it's better to take full advantage of the layout spec. 

## Related Issue

Fixes #2181

## Checklist before merging

### DIFF
--- a/src/pkg/images/unpack.go
+++ b/src/pkg/images/unpack.go
@@ -39,7 +39,7 @@ const (
 // Unpack extracts an image tar and loads it into an OCI layout directory.
 // It returns a list of ImageWithManifest for all images in the tar.
 func Unpack(ctx context.Context, imageArchive v1alpha1.ImageArchive, destDir string, arch string) (_ []ImageWithManifest, err error) {
-	if imageArchive.Images == nil {
+	if len(imageArchive.Images) == 0 {
 		return nil, fmt.Errorf("images must be defined")
 	}
 	// Create a temporary directory for extraction


AustinAbro321:
What happens when an archive has the same image name, as an existing image in the package

AustinAbro321:
Need to make image archives work when importing a component

a1994sc:
Hey Austin, quick question, would this be able to import helm charts? Or is this just limited to OCI tar balls?

AustinAbro321:
Great question @a1994sc, it would work if you ran `docker save my-helm-oci-image:1.0.0`, however it would not work on a .tgz pulled with `helm pull` as that file is not formatted in an OCI layout directory

a1994sc:
> Great question @a1994sc, it would work if you ran `docker save my-helm-oci-image:1.0.0`, however it would not work on a .tgz pulled with `helm pull` as that file is not formatted in an OCI layout directory

Hm, gotcha! Thank you very much for the clarification. I like the direction this PR is going, so I wanted to asked about some of the "edge" cases

brandtkeller:
For posterity - we discussed the balance between UX and Package Definition principles and landed on the enforcement of requiring the image names to be specified in the list of images associated to an image archive - as opposed to the model where zarf imports everything if not specified. 

We believe this is a more secure and traceable behavior between packaging intent and what is packaged during execution.

AustinAbro321:
@brandtkeller Yup, updated PR to reflect this

src/pkg/images/unpack.go
```diff
@@ -0,0 +1,202 @@
...
+	// This is the default docker annotation for the image name
+	dockerRefAnnotation = "io.containerd.image.name"
+	// When the Docker engine containerd image store is used only this annotation is used for sha referenced images
+	dockerContainerdImageStoreAnnotation = "containerd.io/distribution.source.docker.io"
+)
+
+// Unpack extracts an image tar and loads it into an OCI layout directory.
+// It returns a list of ImageWithManifest for all images in the tar.
+func Unpack(ctx context.Context, imageArchive v1alpha1.ImageArchive, destDir string, arch string) (_ []ImageWithManifest, err error) {
+	if imageArchive.Images == nil {
```
brandtkeller: Current conditional allows an empty list to proceed

```suggestion
	if len(imageArchive.Images) == 0 {
```

brandtkeller (approved):
Besides continuing to reinforce that we need documentation which specifies artifacts from the local filesystem (and access to any location on the filesystem) as outside of the threat model for package creation - then I don't see any issues. 

Hoping this creates some interesting packaging use-cases. LGTM.

PR #4335, opened 2025-11-01
feat: add flag to update images in zarf.yaml

### LINKED ISSUE #2865: Allow `zarf dev find-images` update `zarf.yaml` with discovered images

robinlieb:
## Description

This PR adds a `--update` / `-u` flag the `zarf dev find-images` command to update the images for the components in the respective zarf.yaml file.

Due to the limitations of of yaml handling, this preserves the comments in the yaml file, but could lead to a change of spacing like e.g. change something like 

```
kind: ZarfPackageConfig # ZarfPackageConfig is the package kind for most normal zarf packages
metadata:
  name: wordpress       # specifies the name of our package and should be unique and unchanging through updates
...
```

To

```
kind: ZarfPackageConfig # ZarfPackageConfig is the package kind for most normal zarf packages
metadata:
  name: wordpress # specifies the name of our package and should be unique and unchanging through updates
...
```

Because of that behaviour, this changes checks before evaluating the change if the write is needed, to not only change spacing if no images are updated. 

## Related Issue

Fixes #2865

## Checklist before merging

### DIFF
--- a/src/pkg/packager/update.go
+++ b/src/pkg/packager/update.go
@@ -1,0 +1,149 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+package packager
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"slices"
+
+	"github.com/defenseunicorns/pkg/helpers/v2"
+	"github.com/goccy/go-yaml"
+	"github.com/goccy/go-yaml/ast"
+	"github.com/goccy/go-yaml/parser"
+	"github.com/zarf-dev/zarf/src/api/v1alpha1"
+	"github.com/zarf-dev/zarf/src/pkg/logger"
+	"github.com/zarf-dev/zarf/src/pkg/packager/layout"
+)
+
+// UpdateImages updates the images field for components in a zarf.yaml
+func UpdateImages(ctx context.Context, baseDir string, imagesScans []ComponentImageScan) error {
+	l := logger.From(ctx)
+	packageConfigFile := filepath.Join(baseDir, layout.ZarfYAML)
+
+	packageConfigBytes, err := os.ReadFile(packageConfigFile)
+	if err != nil {
+		return fmt.Errorf("failed to read %s: %w", packageConfigFile, err)
+	}
+
+	zarfPackage := v1alpha1.ZarfPackage{}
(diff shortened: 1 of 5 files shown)


src/pkg/packager/update.go
```diff
@@ -0,0 +1,126 @@
...
+		scanMap[scan.ComponentName] = imageSet
+	}
+
+	for _, component := range zarfPackage.Components {
+		imageSet, found := scanMap[component.Name]
+		if !found {
+			return true
+		}
+
+		for _, img := range component.Images {
```
AustinAbro321, line 118: It seems the logic here is reversed, we are checking if the images already defined in the component exist in the image set, we should instead check if the images found in the scan already exist in the component.

robinlieb: Good catch! But I think we need to check on both directions. Scans can contain new images, but also images could not be included anymore, which should also trigger an update. Added tests to make this more clear.

src/pkg/packager/update.go
```diff
@@ -0,0 +1,149 @@
...
+		componentMap[component.Name] = imageSet
+	}
+
+	for _, scan := range imageScans {
+		componentImages, found := componentMap[scan.ComponentName]
+		if !found {
+			return true
+		}
+
+		combined := slices.Concat(scan.Matches, scan.CosignArtifacts)
```
AustinAbro321: Not sure if leaving out `potentialMatches` was intentional, but I think we should keep them in. Feels more consistent since we always add them when we do an update

src/pkg/packager/update.go
AustinAbro321: Ah yeah you're right

src/pkg/packager/update.go
robinlieb: Added.

AustinAbro321 (approved):
Thanks for the PR, I believe many users will appreciate this addition

mkcp (approved):
This is great, thanks for taking it on @robinlieb!

PR #4344, opened 2025-11-05
fix(actions): make go-templates opt-in by default

mkcp:
## Description
This PR makes two fixes to go-templating in Zarf actions. First, we swap the `template: bool` field to be _false_ and disabled by default. This prevents breaking changes for actions that pass go-templates to CLI tooling. On the initial pass we underestimated how many packages would use this pattern, and we weren't aware of the negative impact this would have on some users' existing packages. We also ensure that the values feature must be enabled for this situation to occur - which should have been done in the first place to avoid alpha changes affecting users who haven't opted-in to testing it.

## Related Issue
Relates to https://github.com/zarf-dev/zarf/issues/4282

## Checklist before merging

### DIFF
--- a/src/pkg/packager/actions/actions.go
+++ b/src/pkg/packager/actions/actions.go
@@ -17,7 +17,6 @@ import (
 	"github.com/defenseunicorns/pkg/helpers/v2"
 	"github.com/goccy/go-yaml"
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
-	"github.com/zarf-dev/zarf/src/internal/feature"
 	ptmpl "github.com/zarf-dev/zarf/src/internal/packager/template"
 	"github.com/zarf-dev/zarf/src/internal/template"
 	"github.com/zarf-dev/zarf/src/internal/value"
@@ -88,12 +87,10 @@ func runAction(ctx context.Context, basePath string, defaultCfg v1alpha1.ZarfCom
 	}
 
 	// Apply go-templates in cmds if templating is enabled
-	if feature.IsEnabled(feature.Values) {
-		if action.ShouldTemplate() {
-			cmd, err = template.Apply(ctx, cmd, tmplObjs)
-			if err != nil {
-				return fmt.Errorf("could not template cmd %s: %w", cmdEscaped, err)
-			}
+	if action.ShouldTemplate() {
+		cmd, err = template.Apply(ctx, cmd, tmplObjs)
+		if err != nil {
+			return fmt.Errorf("could not template cmd %s: %w", cmdEscaped, err)
 		}
 	}
 


src/pkg/packager/actions/actions.go
```diff
@@ -87,10 +88,12 @@ func runAction(ctx context.Context, basePath string, defaultCfg v1alpha1.ZarfCom
 	}
 
 	// Apply go-templates in cmds if templating is enabled
-	if action.ShouldTemplate() {
-		cmd, err = template.Apply(ctx, cmd, tmplObjs)
-		if err != nil {
-			return fmt.Errorf("could not template cmd %s: %w", cmdEscaped, err)
+	if feature.IsEnabled(feature.Values) {
```
AustinAbro321: IMO now that the package creator is opting into having templates, we shouldn't put this behind a feature flag. Similar to how when the creator specifies values files and sourcepath/targetpath we don't require that the deployer to set enable values unless they are using the `--set-values` or `--values` flags. In my head, the current user choosing to opt-into the feature is the line where we require feature's be enabled. 

The scenario I'm thinking about is someone deploying a package with templated actions, and not realizing that they need to enables values

brandtkeller: You raise a good point and that got me spinning on the web of requirements. As you mentioned - We don't require the values feature to still allow creation just specifying a charts values sourcepath/targetpath - but the operation is largely a no-op. 

In this case (no features check) setting to true without otherwise using values will result in errors due to missing keys during templating. 

The current proposed workflow skips over templating entirely. 

Is there an alternative where - if templating is set to true we _then_ check for the feature to be enabled and others return a more-clear error?

AustinAbro321: Not necessarily, the package creator could add defaults for each of the values, in this case sourcepath/targetpath would work as intended even with values not enabled.

If we were to add that check I believe we should also error if templating is set to true for manifests or sourcepath/targetpath is used without values enabled. I think this is a reasonable strategy, though I personally would be in favor of only checking the feature flag when the user is consciously choosing to use the feature. 

Worth discussing as we'll want to setup a pattern of how this should be done when we have other feature flagged items in the schema that are set on create.

mkcp: These are all good considerations -- it doesn't make sense to only run these actions if the deployer opts-in to running them with the feature flag, because they've already been declared on the create side. Therefore this PR has been updated to _not_ check if the flag is enabled apply templates when `template: true`.

Out of band, we've discussed ways to do more effective create-time feature checks when using schema keys relating to alpha features. We've also discussed more of the nuance of features at deploy-time, which should remain a seamless experience. Each of these has its own solutions, like annotations and automatic checks on the PackageDefinition schema, and listing enabled features in the build metadata so these could automatically be enabled at deploy-time.

brandtkeller (approved):
LGTM

Issue #2955, opened 2024-08-30
Add Support for ApplicationSets in the zarf agent to handle mutation of git repourl


rdecuir:
### Is your feature request related to a problem? Please describe.
We deploy both Applications and ApplicationSets on our clusters that we are attempting to migrate to manage via Zarf. After testing locally and locking down our test clusters we found that the zarf mutating webhook does not redirect ApplicationSet's  git repourl fields to the internal gitea, as it does for the Applications

### Describe the solution you'd like
We would like to see ApplicationSets be supported in the zarf agent code.

mkcp:
Related to #3046, #2982

AustinAbro321:
This was implemented by https://github.com/zarf-dev/zarf/pull/4348

PR #4348, opened 2025-11-06
feat: add support for Argo CD ApplicationSets

robinlieb:
## Description

Adds functionality to patch repoURL for Argo CD ApplicationSets.

This only adds patching of the git URL for the generations section, not in the template section of the ApplicationSet. This will be patches by the Argo CD Application itself.

## Related Issue

Relates to #2955 

## Checklist before merging

### DIFF
--- a/examples/argocd/applicationset/values.yaml
+++ b/examples/argocd/applicationset/values.yaml
@@ -1,25 +1,23 @@
 applicationsets:
-  - name: cluster-addons
+  - name: guestbook
     namespace: argocd
     goTemplate: true
     goTemplateOptions:
       - missingkey=error
     generators:
       - git:
-          repoURL: https://github.com/argoproj/argo-cd.git
-          revision: v3.2.2
+          repoURL: https://github.com/argoproj/argocd-example-apps.git
+          revision: 0d521c6e049889134f3122eb32d7ed342f43ca0d
           directories:
-            - path: applicationset/examples/git-generator-directory/cluster-addons/*
-            - path: applicationset/examples/git-generator-directory/cluster-addons/prometheus-operator
-              exclude: true
+            - path: guestbook
     template:
       metadata:
         name: '{{.path.basename}}'
       spec:
         project: default
         source:
-          repoURL: https://github.com/argoproj/argo-cd.git
-          targetRevision: HEAD
+          repoURL: https://github.com/argoproj/argocd-example-apps.git
+          targetRevision: 0d521c6e049889134f3122eb32d7ed342f43ca0d
           path: '{{.path.path}}'
         destination:
           server: https://kubernetes.default.svc
--- a/examples/argocd/zarf.yaml
+++ b/examples/argocd/zarf.yaml
@@ -50,25 +50,24 @@ components:
     charts:
       - name: argocd-apps
         version: 1.6.1
-        namespace: argo-workflows
+        namespace: guestbook
         url: https://argoproj.github.io/argo-helm
         releaseName: argocd-apps
         valuesFiles:
           - applicationset/values.yaml
     repos:
-      - https://github.com/argoproj/argo-cd.git@v3.2.2
+      - https://github.com/argoproj/argocd-example-apps.git@0d521c6e049889134f3122eb32d7ed342f43ca0d
     images:
-      - quay.io/argoproj/argocli:v3.4.0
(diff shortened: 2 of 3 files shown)


AustinAbro321:
Hi @robinlieb thanks for the PR, and sorry for the delays in reviewing this. Generally we like to see the Zarf agent functionality tested in an end to end way. It helps us ensure that there are no regressions in the mutating web hook, and that the mutated resources work in a cluster as expected. Could you add an application set to the argocd package in examples? With how our test suite is currently structured, this will sufficiently test the webhook. 

Also we should add a quick line modification to the argocd line in the [init package docs](https://docs.zarf.dev/ref/init-package/) that adds Argocd Application Sets

robinlieb:
Thanks for the feedback, will update the PR with a useful example for the Argo CD Application set in the Argo CD example and update the init package docs.

robinlieb:
Example and docs added.

examples/argocd/zarf.yaml
```diff
@@ -45,6 +45,31 @@ components:
...
+    charts:
+      - name: argocd-apps
+        version: 1.6.1
+        namespace: argo-workflows
+        url: https://argoproj.github.io/argo-helm
+        releaseName: argocd-apps
+        valuesFiles:
+          - applicationset/values.yaml
+    repos:
+      - https://github.com/argoproj/argo-cd.git
```
AustinAbro321:
```suggestion
      - https://github.com/argoproj/argo-cd.git@v3.2.2
```
I noticed that this takes quite a while to download. It's a big repo, but it's taking me about 3 minutes to pull it, much longer than I'd expect with >500mbs. Likely something we'll want to look into. For now, I think it's pragmatic to pull a specific tag, as we want this to be faster for users trying out examples, and for our test suite

examples/argocd/zarf.yaml
```diff
@@ -45,6 +45,31 @@ components:
...
+    actions:
+      onDeploy:
+        after:
+          - description: Wait for Argo CD Application argo-workflows to be spawned
+            wait:
+              cluster:
+                kind: Application
+                name: argo-workflows
+                namespace: argocd
+                condition: exists
```
AustinAbro321: We should instead wait for the deployment or pod as we do in the other argo chart. The argo workflows are part of the helm chart, so those will already be waited on by Zarf / Helm normally. Since the pods / deployments happen outside of the scope of the Helm chart, Zarf won't wait for them normally and we want custom waits.

robinlieb: Changed it to wait for workflow controller pod.

examples/argocd/applicationset/values.yaml
```diff
@@ -0,0 +1,31 @@
...
+    namespace: argocd
+    goTemplate: true
+    goTemplateOptions:
+      - missingkey=error
+    generators:
+      - git:
+          repoURL: https://github.com/argoproj/argo-cd.git
+          revision: v3.2.2
+          directories:
+            - path: applicationset/examples/git-generator-directory/cluster-addons/*
```
brandtkeller: In a disconnected test - this package will not reconcile currently. This is due to kustomize trying to fetch the required resources indicated in this path. We may need to find/build another example. 

```bash
Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = `kustomize build <path to cached source>/applicationset/examples/git-generator-directory/cluster-addons/argo-workf │
│ lows` failed exit status 1: Error: accumulating resources: accumulation err='accumulating resources from 'https://github.com/argoproj/argo-workflows/releases/download/v3.4.0/namespace-install.yaml': Get "https://github.com/argoproj/argo-workflows/rel │
│ eases/download/v3.4.0/namespace-install.yaml": dial tcp: lookup github.com on 10.96.0.10:53: no such host': failed to run '/usr/bin/git fetch --depth=1 https://github.com/argoproj/argo-workflows HEAD': fatal: unable to access 'https://github.com/argo │
│ proj/argo-workflows/': Could not resolve host: github.com  
```

robinlieb: Good catch, sorry for missing that. I changed the example to use guestbook from argocd-example-apps.

brandtkeller (approved):
lgtm - appreciate the patience in review.

PR #4350, opened 2025-11-07
feat!: upgrade to Helm 4

### LINKED ISSUE #2756: Helm tool ignores errors
### Environment
Device and OS:  N/A
App version: v0.33.0
Other: N/A

### Steps to reproduce
1. Create a file called `config.json` with invalid json content
2. Export the registry config path `HELM_REGISTRY_CONFIG=config.json`
3. Run `zarf --version`

### Expected result

The Zarf version should be printed.

### Actual Result

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x3d61d4a]

goroutine 1 [running]:
github.com/zarf-dev/zarf/src/cmd/tools.init.2()
	/home/philip/go/src/github.com/zarf-dev/zarf/src/cmd/tools/helm.go:30 +0xea
```

### Severity/Priority

Medium

### Additional Context

### LINKED ISSUE #2818: Refactor helm embedding when/if a better solution becomes available
Within [src/cmd/tools/helm](https://github.com/zarf-dev/zarf/tree/b903e262f16424307e056a6c85102e2294cabe1b/src/cmd/tools/helm) there is a copy of several helm source code files. This duplication is not ideal as if the code ever changes we don't have a good way to know and update the code ourselves. A PR was opened to enable native cobra embedded https://github.com/helm/helm/pull/12725, however it was unfinished and closed. The creator of the PR mentioned that the Helm team was open to a PR like this if someone is willing to put in the time to get it in a merge-able state

### LINKED ISSUE #4437: Upgrade to Helm 4
Helm 4 is released. With that comes many breaking changes as library consumes, as well as some breaking changes for user functionality. There are also new features that Zarf should consider exposing to users / setting reasonable defaults. 

It's impossible for Zarf to test every way Helm 4 might break things for all the different charts that package using Zarf. There should be an RC release with Helm 4 merged so that we can give the ability for the community to test it out.

AustinAbro321:
## Description

User facing changes: 
- Helm 4 brings [Server Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/). If it is a first time chart install then server side apply will be used by default. Upgrades default to the apply method that was used in the latest release. To control the apply method for a specific packaged chart, a new field .Charts[x].ServerSideApply has been introduced with possible values true, false, auto . Auto will follow the default Helm behavior. 
  - A good example of how this could effect some charts can be seen in the [init package registry deployment](https://github.com/zarf-dev/zarf/pull/4350/changes#diff-0d526af7c8098cac8b11ac168902b1df32f24893064838264be5b3f20ab2f624R16-R18). Server side apply caught that replicas were set even when they were controlled by the autoscaler, which led to a conflict error. To solve this we conditionally set replicas when the autoscaler is not enabled.  
  - SSA enforces stricter field validation, if a field does not pass schema validation, you will receive an error instead of a warning.  
- Helm now uses slog style logging. 
- `zarf tools helm` is now a true copy of the Helm CLI. Previously, this command was a quasi fork of a subset of the Helm CLI and was not updated automatically. Zarf maintainer contributed to make this functionality possible [#13617](https://github.com/helm/helm/pull/13617)
- Helm's now uses kstatus as a backend for it's wait logic, since Zarf already used kstatus post Helm install the wait logic will be near identical to earlier Zarf versions. Zarf maintainers contributed this functionality to Helm in [#13604](https://github.com/helm/helm/pull/13604)

---

We will have to be very careful about merging this as there may be impacts to charts that we do not test for in the migration from Helm 3 to Helm 4. We want the experience to be seamless for users. I intend to make a R.C. release from this branch before merging into main.

Note that vendoring in the entirety of Helm instead of the current quasi fork bring binary size up about 5MB from 173M on main to 178M on this branch

Something interesting is that server side apply default to true during new installs, but "auto" during upgrades and rollbacks meaning that existing charts will use the existing strategy. I decided to add a new addition to the schema, serverSideApply, which accepts the field "true", "false",  and "auto". 

## Related Issue

Fixes #2818
Fixes #2756
Fixes #4437

## Checklist before merging


src/internal/packager/helm/chart.go
```diff
@@ -144,22 +153,6 @@ func InstallOrUpgradeChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, ch
...
-	}
-
-	runtimeObjs := []runtime.Object{}
-	for _, resource := range resourceList {
-		runtimeObjs = append(runtimeObjs, resource.Object)
-	}
-	if !zarfChart.NoWait {
-		// Ensure we don't go past the timeout by using a context initialized with the helm timeout
-		l.Info("running health checks", "chart", zarfChart.Name)
-		if err := healthchecks.WaitForReadyRuntime(helmCtx, opts.Cluster.Watcher, runtimeObjs); err != nil {
```
AustinAbro321, line 159: Now that Helm has implemented kstatus, there's no reason for Zarf to do it's own healthchecks anymore

packages/zarf-registry/chart/templates/deployment.yaml
```diff
@@ -13,7 +13,9 @@ spec:
     matchLabels:
       app: {{ template "docker-registry.name" . }}
       release: {{ .Release.Name }}
+  {{ if not .Values.autoscaling.enabled }}
   replicas: {{ .Values.replicaCount }}
+  {{ end }}
```
AustinAbro321, line 18: The reason this was added is because with server side apply if autoscaling is enabled then it owns the replicas field, so we will run into an error if we re-apply the chart with a set amount of replicas and auto-scaling.

site/src/content/docs/commands/zarf_tools_helm_version.md
```diff
@@ -8,7 +8,34 @@ tableOfContents: false
...
+
+### Synopsis
+
+
+Show the version for Helm.
+
+This will print a representation the version of Helm.
+The output will look something like this:
+
+version.BuildInfo{Version:"v3.2.1", GitCommit:"fe51cd1e31e6a202cba7dead9552a6d418ded79a", GitTreeState:"clean", GoVersion:"go1.13.10"}
```
brandtkeller: Current output reflects:
```bash
./build/zarf-mac-apple tools helm version
version.BuildInfo{Version:"v4.0", GitCommit:"", GitTreeState:"", GoVersion:"go1.25.5", KubeClientVersion:"v."}
```

appears incomplete - cursory glance shows the helm version package is internal - so we might need to evaluate options for this information. 

Important part being that the output will help establish provenance on the included version.

src/cmd/helm.go
```diff
@@ -5,34 +5,28 @@
...
 	helmArgs := []string{}
 	if len(os.Args) > 2 {
 		helmArgs = os.Args[3:]
 	}
-	// The inclusion of Helm in this manner should be changed once https://github.com/helm/helm/pull/13617 is merged
-	cmd, err := helm.NewRootCmd(actionConfig, os.Stdout, helmArgs)
+
+	kube.ManagedFieldsManager = "helm"
+
+	cmd, err := helmcmd.NewRootCmd(os.Stdout, helmArgs, helmcmd.SetupLogging)
```
brandtkeller, line 24: Would it make sense to use Zarfs default logger instead here?

```go
cmd, err := helmcmd.NewRootCmd(os.Stdout, helmArgs, func(debug bool) {
		l := logger.Default()
		slog.SetDefault(l)
	})
	if err != nil {
		logger.Default().Error("helm command initialization failed", "error", err)
		os.Exit(1)
	}
```

brandtkeller (changes requested):
Looking over helm generically in this pass - continuing to dig into the integration. Couple minor questions for helm is isolation. 

Excited to see helm vendored fully.

site/src/content/docs/commands/zarf_tools_helm_version.md
AustinAbro321: We can set these variables through build flags such as `helm.sh/helm/v4/internal/version.gitCommit`. I'm not sure there is a good way to get them besides using the tag to make an API request to GitHub. I'm amendable to it, but I'd personally rather avoid introducing that into our build process.

I did update the build flags to set kubeClientVersion. We could also set gitTreeState to clean, though maybe it makes more sense to leave empty given that we're not setting the commit.l

src/cmd/helm.go
AustinAbro321: This is a tough decision, mostly because Helm is setting a global logger when they should have a more SDK friendly approach. We could do something like this 
```go
		cfg := logger.ConfigDefault()
		if debug {
			cfg.Level = logger.Debug
		}
		l, err := logger.New(cfg)
		if err != nil {
			logger.Default().Error("logger setup failed", slog.Any("error", err))
			return
		}
		slog.SetDefault(l)
```

and get debug logs, however we won't be able to account for flags like `--color`. I'd personally advocate to continue using the Helm logging and creating an issue upstream for them to avoid using the global logger. I actually ran into this problem in a different space and created this issue https://github.com/helm/helm/pull/31717

Tested things out with the command `zarf tools helm install test-chart-2 oci://ghcr.io/stefanprodan/charts/podinfo --debug --wait`

site/src/content/docs/commands/zarf_tools_helm_version.md
brandtkeller: This sounds fair to me. As long as we have evaluated what is possible and/or available we can iterate - I think this information ultimately builds trust in the Zarf inventory of tooling.

src/cmd/helm.go
brandtkeller: I'm good with that direction - thank you for creating context upstream.

AustinAbro321: Created a broader issue about this as well https://github.com/helm/helm/issues/31724

packages/zarf-registry/chart/templates/deployment.yaml
brandtkeller: While not a zarf issue directly (more k8s migration to SSA issue) - this will be exposed to through zarf to consumers. 

Are there other known instances of this ownership issue and do we need to document any of them for users to discover?

src/internal/packager/helm/chart.go
```diff
@@ -261,26 +261,48 @@ func installChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, chart *char
 	// Post-processing our manifests to apply vars and run zarf helm logic in cluster
 	client.PostRenderer = postRender
 
+	useSSA := zarfChart.ServerSideApply != "false"
+	client.ServerSideApply = useSSA
+	// Only force conflicts if SSA is enabled and the option is set
+	client.ForceConflicts = useSSA && opts.ForceConflicts
```
brandtkeller: I like this distinction for compatibility - given backwards CSA potential.

src/internal/packager/helm/chart.go
```diff
@@ -261,26 +261,48 @@ func installChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, chart *char
...
+	} else {
+		client.WaitStrategy = kube.StatusWatcherStrategy
+	}
+
+	if zarfChart.ServerSideApply == "" {
+		zarfChart.ServerSideApply = "auto"
+	}
+	client.ServerSideApply = zarfChart.ServerSideApply
+	// Only force conflicts if SSA is enabled and the option is set
+	useSSA := zarfChart.ServerSideApply != "false" && zarfChart.ServerSideApply != ""
```
brandtkeller: Is the latter conditional possible given the preceding lines?

```suggestion
	useSSA := zarfChart.ServerSideApply != "false"
```

src/internal/packager/helm/chart.go
```diff
@@ -295,30 +317,37 @@ func upgradeChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, chart *char
...
-	return client.RunWithContext(ctx, zarfChart.ReleaseName, chart, chartValues)
+	_, err = client.RunWithContext(ctx, zarfChart.ReleaseName, chart, chartValues)
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
-func rollbackChart(name string, version int, actionConfig *action.Configuration, timeout time.Duration) error {
+func rollbackChart(zarfChart v1alpha1.ZarfChart, version int, actionConfig *action.Configuration, timeout time.Duration) error {
```
brandtkeller: Is there a permutation of scenarios where we need to set `client.ForceConflicts` on rollback?

src/internal/packager/helm/chart.go
```diff
@@ -333,7 +362,13 @@ func LoadChartData(zarfChart v1alpha1.ZarfChart, chartPath string, valuesPath st
 
 // migrateDeprecatedAPIs searches through all the objects from the latest release and migrates any deprecated APIs to the latest version.
 // If any deprecated fields are found, the release will be updated and saved back to the cluster.
-func migrateDeprecatedAPIs(ctx context.Context, c *cluster.Cluster, actionConfig *action.Configuration, latestRelease *release.Release) error {
+func migrateDeprecatedAPIs(ctx context.Context, c *cluster.Cluster, actionConfig *action.Configuration, latestReleaser release.Releaser) error {
+	// We can re-evaluate handling this functionality for chart v3 / release v2 when available.
+	// Potentially we stop doing this functionality for future chart APIs, as it could be outside of the intended scope of Zarf.
```
brandtkeller, line 366: I would agree - I believe this workflow also incurs a decent dependency footprint if I recall correctly.

src/internal/packager/helm/repo.go
```diff
@@ -236,19 +236,7 @@ func DownloadPublishedChart(ctx context.Context, chart v1alpha1.ZarfChart, chart
...
-	}
-	defer func(l *slog.Logger) {
-		err := os.RemoveAll(temp)
-		if err != nil {
-			l.Error(err.Error())
-		}
-	}(l)
-
-	saved, _, err := chartDownloader.DownloadTo(chartURL, pull.Version, temp)
+	saved, _, err := chartDownloader.DownloadToCache(chartURL, pull.Version)
```
brandtkeller: neat improvement under-the-hood for zarf (caching).

src/internal/packager/helm/chart.go
AustinAbro321: Yeah Helm will actually fail if force conflicts is true and csa is set which I think is good

src/internal/packager/helm/chart.go
AustinAbro321: I considered always forcing rollback to avoid users ending up in a bad state, but I think the simplest and clearest thing to do is to just follow the CLI force conflicts flag. I went ahead and implemented that

packages/zarf-registry/chart/templates/deployment.yaml
AustinAbro321: Vertical pod autoscalers is another example, I'm having trouble finding others though. I do plan on mentioning this example in the announcement

src/internal/packager/helm/chart.go
```diff
@@ -261,26 +260,52 @@ func installChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, chart *char
...
+	client.ForceConflicts = useSSA && opts.ForceConflicts
+
 	// Perform the loadedChart installation.
-	return client.RunWithContext(ctx, chart, chartValues)
+	_, err := client.RunWithContext(ctx, chart, chartValues)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
soltysh: You can easily replace this entire block with a simple:
```golang
	_, err := client.RunWithContext(ctx, chart, chartValues)
	return err
```
Since you're not doing anything special with that error.

src/internal/packager/helm/chart.go
```diff
@@ -295,30 +320,39 @@ func upgradeChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, chart *char
 	client.MaxHistory = maxHelmHistory
 
 	// Perform the loadedChart upgrade.
-	return client.RunWithContext(ctx, zarfChart.ReleaseName, chart, chartValues)
+	_, err = client.RunWithContext(ctx, zarfChart.ReleaseName, chart, chartValues)
+	if err != nil {
+		return err
+	}
+	return nil
```
soltysh: Same comment.

soltysh (dismissed):
Minor nits, but overall this looks great :)

src/internal/packager/helm/template.go
```diff
@@ -103,49 +113,53 @@ func newTemplateRenderer(actionConfig *action.Configuration, vc *variables.Varia
 // Run satisfies the Helm post-renderer interface and templates the Zarf vars in the rendered manifests.
 func (tr *templateRenderer) Run(renderedManifests *bytes.Buffer) (*bytes.Buffer, error) {
 	// This is very low cost and consistent for how we replace elsewhere, also good for debugging
-	resources, err := getTemplatedManifests(renderedManifests, tr.variableConfig, tr.actionConfig)
+	hooks, resources, err := getTemplatedManifests(renderedManifests, tr.variableConfig, tr.actionConfig)
```
AustinAbro321, line 118: Previously, Helm added hooks after the post renderer, now that they don't we need to render them in the post renderer

AustinAbro321:
This PR is blocked until a new release of Helm with https://github.com/helm/helm/pull/31772 merged is created

AustinAbro321:
Helm 4.1.1 was released and this is no longer blocked

src/cmd/helm.go
```diff
@@ -5,34 +5,28 @@
...
 	if err != nil {
-		logger.Default().Debug("failed to initialize helm command", "error", err)
+		logger.Default().Error("Helm command initialization", slog.Any("error", err))
 	}
-	cmd.Short = lang.CmdToolsHelmShort
-	cmd.Long = lang.CmdToolsHelmLong
-	cmd.AddCommand(newToolsVersionCmd("helm", helmVersion))
 
 	return cmd
```
a1994sc: Would it be possible to replace 

```golang
	return cmd
```

with

```golang
	return ReplaceCommandName("helm", "zarf tools helm", cmd)
```

AustinAbro321: Good call, I also decided to remove all the helm docs page, ultimately users are better served by going to the helm docs

a1994sc: valid, I think that makes sense.

src/test/e2e/25_helm_test.go
```diff
@@ -239,3 +280,22 @@ func testHelmAdoption(t *testing.T, tmpdir string) {
 	stdOut, stdErr, err = e2e.Zarf(t, "package", "remove", "dos-games", "--confirm")
 	require.NoError(t, err, stdOut, stdErr)
 }
+
+func testHelmHooks(t *testing.T) {
```
brandtkeller, line 284: Is parallel available here?

src/api/v1alpha1/component.go
```diff
@@ -206,6 +206,9 @@ type ZarfChart struct {
 	Values []ZarfChartValue `json:"values,omitempty"`
 	// Whether or not to validate the values.yaml schema, defaults to true. Necessary in the air-gap when the JSON Schema references resources on the internet.
 	SchemaValidation *bool `json:"schemaValidation,omitempty"`
+	// whether server side apply should be used during upgrades or installs. Defaults to true during installs and auto during upgrades.
```
brandtkeller: nit: this comment seems to describe behavior instead of semantics which makes understanding its use a little ambiguous without digging through the related utilization. 

Something like:
```go
// Controls whether Helm uses Server-Side Apply (SSA) or client-side apply (CSA) when deploying this chart.
//   - "true":  always use SSA
//   - "false": always use CSA
//   - "auto":  use SSA for fresh installs; for upgrades, match whichever strategy
//              was used when the chart was first installed
// Defaults to "auto" when omitted.
```

src/internal/packager/helm/template.go
brandtkeller: Tracking the update- considering what the implications might be for this release. Will anyone diff-ing package manifests expect to see a different result? is it deterministic run-over-run?

brandtkeller (changes requested):
Review nearly complete - working through the internals

src/test/e2e/25_helm_test.go
AustinAbro321: Yes

src/internal/packager/helm/template.go
AustinAbro321: There is a difference in that variables in hook files are now templated, but otherwise it should be the same as before

Issue #3046, opened 2024-09-30
Support OCI references for ArgoCD in Mutating Webhook


kdward:
### Is your feature request related to a problem? Please describe.
ArgoCD supports the deployment of Helm OCI charts: https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#declarative

This is quite useful for a gitops pattern where the chart is built as an OCI artifact and the git repository is only referenced for the values.yaml files for the various environments.

Currently when this pattern is used with Zarf the ArgoCD Application object that has a definition with:
```
  source:
    chart: cert-manager  
    repoURL: registry-1.docker.io/bitnamicharts
    targetRevision: 1.3.18
```

Is mutated by the webhook to refer to a gitea repository that won't exist rather than the zarf registry where the OCI chart would have been pushed when listed in the images section of the zarf component:
```
  source:
    chart: cert-manager
    repoURL: http://zarf-gitea-http.zarf.svc.cluster.local:3000/zarf-git-user/bitnamicharts-3383370025
    targetRevision: 1.3.18
```

### Describe the solution you'd like

- **Given** an OCI chart reference in an ArgoCD Application object and it's inclusion in the images section of the zarf component
- **When** The mutating webhook modifies the Application object
- **Then** The resulting Application object should reference the OCI chart in the Zarf registry

### Describe alternatives you've considered
Bundling the chart directly into the git repository as a self-managed helm dependency rather than an referenced artifact which is not ideal

### Additional context
I think this would additionally require the creation of a new repository secret for ArgoCD with `enableOCI: "true"` set in the data section.

Issue #4327, opened 2025-10-29
Using ArgoCD with values from the air-gapped gitea instance


al-jeyapal:
Hi team, thanks for all your work on this great project.

Just trying to get some clarity on a use case we're trying to implement.

It looks conceptually simlar to https://github.com/zarf-dev/zarf/issues/1830

I would like to do the following:

1. Deploy ArgoCD to the cluster using a zarf package.
    * No issues.
2. Take a `localPath` Helm chart and `mirror-resources` it to zarf registry.
    * A bit stuck here, see below.
3. Create a repo in the zarf `gitea` instance containing a Helm `values.yaml` for the chart.
    * No issues.
4. Setup an ArgoCD Application (done manually within the air-gapped environment, not via zarf) using the [Multiple Sources for an Applicaton](https://argo-cd.readthedocs.io/en/stable/user-guide/multiple_sources/) pattern. I.e. something like:

```
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: some-app
  namespace: argocd
spec:
  project: default
  destination:
    server: https://kubernetes.default.svc
    namespace: default
  sources:
    - repoURL: 'zarf-docker-registry.zarf.svc.cluster.local:5000'
      chart: some-app/some-app
      targetRevision: 1.2.3
      helm:
        valueFiles:
        - $values/helm-values/values.yaml
    - repoURL: 'zarf-gitea-http/some-user/some-argo-app-values.git'
      targetRevision: HEAD
      ref: values
```

The aim here being we deploy all the capability in terms of images/charts highside via `mirror-resouces`.
Then the users highside can configure ArgoCD to deploy their apps using their highside configuration (independent of zarf).

Is that pattern supported?

The issues I am having is step: *Take a `localPath` Helm chart and `mirror-resources` it to zarf registry.*
 
* From reading https://github.com/zarf-dev/zarf/issues/1830 it looks like I can use an `oci://...` path in `images` to reference remote Helm charts which get mirrored, but not something local.
* As a workaround, I can manually `helm registry login` and push the chart to the zarf image registry. Similar to Step 5 in https://github.com/zarf-dev/zarf/issues/2114#issuecomment-2039633718
* *Is that just something we need to work around? I.e. publish the local chart to some OCI repo so it can get bundled and mirrored?*

---

Useful tickets as background:

- https://github.com/zarf-dev/zarf/issues/1830
- https://github.com/zarf-dev/zarf/issues/4143
[...]

PR #4354, opened 2025-11-10
feat: add OCI support for Argo CD sources

robinlieb:
## Description

This PR adds the capability to mutate OCI references in Argo CD resources and replaces the given OCI reference to the Zarf registry URL.

This change adds the possibility to use Manifest and Helm Charts in the repoURL in the Application, like explained in the [OCI Docs in Argo CD](https://argo-cd.readthedocs.io/en/stable/user-guide/oci/) with the first two examples. To archive that it also changes the patching of the URL in repository secrets to use the registry instead of the git server if oci is used.

Additionally, added the patching of the repo url for Argo CD [repo-creds](https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#repository-credentials) secret type.

## Related Issue

Relates to #3046 

## Checklist before merging

### DIFF
--- a/examples/argocd/baseline/values.yaml
+++ b/examples/argocd/baseline/values.yaml
@@ -19,6 +19,19 @@ configs:
       url: https://github.com/argoproj/argocd-example-apps.git
       username: ""
       password: ""
+    oci-helm:
+      name: podinfo-helm
+      url: oci://ghcr.io/stefanprodan/charts/podinfo
+      username: ""
+      password: ""
+      type: helm
+      enableOCI: "true"
+    oci-manifests:
+      name: oci-repo
+      url: oci://registry-1.docker.io/dhpup/oci-edge
+      username: ""
+      password: ""
+      type: oci
 
 redis:
   image:
--- a/examples/argocd/zarf.yaml
+++ b/examples/argocd/zarf.yaml
@@ -2,6 +2,9 @@ kind: ZarfPackageConfig
 metadata:
   name: argocd
   description: Example showcasing installing ArgoCD
+values:
+  files:
+    - values/values.yaml
 
 components:
   - name: argocd-helm-chart
--- a/examples/argocd/zarf.yaml
+++ b/examples/argocd/zarf.yaml
@@ -8,18 +8,24 @@ components:
     required: true
     charts:
       - name: argo-cd
-        version: 5.54.0
+        version: 9.0.5
         namespace: argocd
         url: https://argoproj.github.io/argo-helm
         releaseName: argocd-baseline
         valuesFiles:
           - baseline/values.yaml
+        # Required for 'src/test/external' to validate external git server.
+        values:
+          - sourcePath: ".extraObjects"
+            targetPath: ".extraObjects"
+          - sourcePath: ".repoServer"
+            targetPath: ".repoServer"
     images:
-      - docker.io/library/redis:7.0.15-alpine
-      - quay.io/argoproj/argocd:v2.9.6
-      # Cosign artifacts for images - argocd - argocd-helm-chart
-      - quay.io/argoproj/argocd:sha256-2dafd800fb617ba5b16ae429e388ca140f66f88171463d23d158b372bb2fae08.sig
-      - quay.io/argoproj/argocd:sha256-2dafd800fb617ba5b16ae429e388ca140f66f88171463d23d158b372bb2fae08.att
+      - docker.io/library/redis:7.2.11-alpine
+      - quay.io/argoproj/argocd:v3.1.9
+      # Cosign artifacts for images - argocd-helm-chart
+      - quay.io/argoproj/argocd:sha256-9c33285c0b86a66e9cf9fcafc448e48e0ab7c4678b20b3723556bb6cce8ea0b9.sig
+      - quay.io/argoproj/argocd:sha256-9c33285c0b86a66e9cf9fcafc448e48e0ab7c4678b20b3723556bb6cce8ea0b9.att
   - name: argocd-apps
     required: true
     charts:
@@ -69,6 +75,32 @@ components:
                 name: app=guestbook-ui
                 namespace: guestbook
                 condition: ready
+  - name: oci-helm
+    required: true
+    charts:
+      - name: argocd-apps
+        version: 1.6.1
+        namespace: podinfo-oci
+        url: https://argoproj.github.io/argo-helm
+        releaseName: argocd-apps
+        valuesFiles:
+          - oci-helm/values.yaml
+    images:
+      - ghcr.io/stefanprodan/charts/podinfo:6.9.2
+      - ghcr.io/stefanprodan/podinfo:6.9.2
+  - name: oci-manifests
+    required: true
+    charts:
+      - name: argocd-apps
+        version: 1.6.1
+        namespace: oci-manifests
+        url: https://argoproj.github.io/argo-helm
+        releaseName: argocd-apps
+        valuesFiles:
+          - oci-manifests/values.yaml
+    images:
+      - registry-1.docker.io/dhpup/oci-edge:1.0.6
+      - ghcr.io/dhpup/guestbook:0.1.40
 
 documentation:
   readme: readme.md
--- a/src/internal/agent/hooks/argocd-application.go
+++ b/src/internal/agent/hooks/argocd-application.go
@@ -80,9 +80,15 @@ func mutateApplication(ctx context.Context, r *v1.AdmissionRequest, cluster *clu
 		"operation", r.Operation,
 		"gitServer", s.GitServer.Address)
 
+	// Get the registry service info if this is a NodePort service to use the internal kube-dns
+	registryAddress, clusterIP, err := cluster.GetServiceInfoFromRegistryAddress(ctx, s.RegistryInfo)
+	if err != nil {
+		return nil, err
+	}
+
 	patches := make([]operations.PatchOperation, 0)
 	if app.Spec.Source != nil {
-		patchedURL, err := getPatchedRepoURL(ctx, app.Spec.Source.RepoURL, s.GitServer)
+		patchedURL, err := getPatchedRepoURL(ctx, app.Spec.Source.RepoURL, registryAddress, clusterIP, s.GitServer, r)
 		if err != nil {
 			return nil, err
 		}
@@ -91,7 +97,7 @@ func mutateApplication(ctx context.Context, r *v1.AdmissionRequest, cluster *clu
 
 	if len(app.Spec.Sources) > 0 {
 		for idx, source := range app.Spec.Sources {
-			patchedURL, err := getPatchedRepoURL(ctx, source.RepoURL, s.GitServer)
+			patchedURL, err := getPatchedRepoURL(ctx, source.RepoURL, registryAddress, clusterIP, s.GitServer, r)
 			if err != nil {
 				return nil, err
 			}
@@ -107,7 +113,12 @@ func mutateApplication(ctx context.Context, r *v1.AdmissionRequest, cluster *clu
 	}, nil
 }
 
-func getPatchedRepoURL(ctx context.Context, repoURL string, gs state.GitServerInfo) (string, error) {
+func getPatchedRepoURL(
+	ctx context.Context,
+	repoURL, registryAddress, clusterIP string,
+	gs state.GitServerInfo,
+	r *v1.AdmissionRequest,
+) (string, error) {
 	l := logger.From(ctx)
 
 	// Skip mutation if the URL already points to the Zarf git server to prevent double-hashing
@@ -122,6 +133,82 @@ func getPatchedRepoURL(ctx context.Context, repoURL string, gs state.GitServerIn
 		return repoURL, nil
 	}
 
+	isOCIURL := helpers.IsOCIURL(repoURL)
+
+	shouldMutate, isPatchedClusterIP, err := shouldMutateURL(r.Operation, isOCIURL, repoURL, registryAddress, clusterIP, gs)
+	if err != nil {
+		return "", fmt.Errorf(lang.AgentErrHostnameMatch, err)
+	}
+
+	if !shouldMutate {
+		return repoURL, nil
+	}
+
+	if isOCIURL {
+		return mutateOCIURL(ctx, repoURL, registryAddress, isPatchedClusterIP)
+	}
+	return mutateGitURL(ctx, repoURL, gs)
+}
+
+func shouldMutateURL(operation v1.Operation, isOCIURL bool, repoURL, registryAddress, clusterIP string, gs state.GitServerInfo) (bool, bool, error) {
+	isCreate := operation == v1.Create
+	isUpdate := operation == v1.Update
+	if isCreate {
+		return true, false, nil
+	}
+
+	var isPatched bool
+	var isPatchedClusterIP bool
+	var err error
+	if isOCIURL {
+		zarfStateAddress := helpers.OCIURLPrefix + registryAddress
+		isPatched, err = helpers.DoHostnamesMatch(zarfStateAddress, repoURL)
+		if err != nil {
+			return false, false, err
+		}
+		if clusterIP != "" {
+			zarfStateClusterIPAddress := helpers.OCIURLPrefix + clusterIP
+			isPatchedClusterIP, err = helpers.DoHostnamesMatch(zarfStateClusterIPAddress, repoURL)
+			if err != nil {
+				return false, false, err
+			}
+		}
+	} else {
+		isPatched, err = helpers.DoHostnamesMatch(gs.Address, repoURL)
+	}
+
+	return (isUpdate && !isPatched), isPatchedClusterIP, err
+}
+
+func mutateOCIURL(ctx context.Context, repoURL, registryAddress string, isPatchedClusterIP bool) (string, error) {
+	l := logger.From(ctx)
+	var patchedSrc string
+	var err error
+
+	if isPatchedClusterIP {
+		patchedSrc, err = transform.ImageTransformHostWithoutChecksum(registryAddress, repoURL)
+		if err != nil {
+			return "", fmt.Errorf("%s: %w", AgentErrTransformOCIURL, err)
+		}
+	} else {
+		patchedSrc, err = transform.ImageTransformHost(registryAddress, repoURL)
+		if err != nil {
+			return "", fmt.Errorf("%s: %w", AgentErrTransformOCIURL, err)
+		}
+	}
+
+	patchedRefInfo, err := transform.ParseImageRef(patchedSrc)
+	if err != nil {
+		return "", fmt.Errorf("%s: %w", AgentErrTransformOCIURL, err)
+	}
+
+	patchedURL := helpers.OCIURLPrefix + patchedRefInfo.Name
+	l.Debug("mutated ArgoCD application OCI repoURL to the Zarf Registry URL", "original", repoURL, "mutated", patchedURL)
+	return patchedURL, nil
+}
+
+func mutateGitURL(ctx context.Context, repoURL string, gs state.GitServerInfo) (string, error) {
+	l := logger.From(ctx)
 	transformedURL, err := transform.GitURL(gs.Address, repoURL, gs.PushUsername)
 	if err != nil {
 		return "", fmt.Errorf("%s: %w", AgentErrTransformGitURL, err)
--- a/src/internal/agent/hooks/argocd-application.go
+++ b/src/internal/agent/hooks/argocd-application.go
@@ -58,16 +58,12 @@ func NewApplicationMutationHook(ctx context.Context, cluster *cluster.Cluster) o
 	}
 }
 
-// mutateApplication mutates the git repository url to point to the repository URL defined in the ZarfState.
+// mutateApplication mutates the repository url to point to the repository URL defined in the ZarfState.
 func mutateApplication(ctx context.Context, r *v1.AdmissionRequest, cluster *cluster.Cluster) (*operations.Result, error) {
 	l := logger.From(ctx)
 	s, err := cluster.LoadState(ctx)
 	if err != nil {
 		return nil, err
-	}
-	if !s.GitServer.IsConfigured() {
-		l.Debug("no Zarf git server configured, skipping ArgoCD Application mutation")
-		return &operations.Result{Allowed: true}, nil
 	}
 
 	app := Application{}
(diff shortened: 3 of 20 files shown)


src/internal/agent/hooks/argocd-repository.go
```diff
@@ -107,11 +135,20 @@ func mutateRepositorySecret(ctx context.Context, r *v1.AdmissionRequest, cluster
...
 
 // Patch updates of the Argo Repository Secret.
-func populateArgoRepositoryPatchOperations(repoURL string, gitServer state.GitServerInfo) []operations.PatchOperation {
+func populateArgoRepositoryPatchOperations(repoURL string, gitServer state.GitServerInfo, registryInfo state.RegistryInfo, isOCI bool) []operations.PatchOperation {
 	var patches []operations.PatchOperation
 	patches = append(patches, operations.ReplacePatchOperation("/data/url", base64.StdEncoding.EncodeToString([]byte(repoURL))))
-	patches = append(patches, operations.ReplacePatchOperation("/data/username", base64.StdEncoding.EncodeToString([]byte(gitServer.PullUsername))))
-	patches = append(patches, operations.ReplacePatchOperation("/data/password", base64.StdEncoding.EncodeToString([]byte(gitServer.PullPassword))))
+
+	if isOCI {
```
brandtkeller: Nit: for code patterns used across zarf (of those replaced thus far) we have been upgrading to use [the happy path is left-aligned](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88) concept whereby - when possible we process conditions and return early and avoid else returns.

brandtkeller (changes requested):
Made a first pass - spent a healthy amount of time educating myself on argo further. 

Left a nit that I think we can cleanup in a few places. Separately want to consider some additional scope for this change:
1. E2E testing - unit tests get us some good feedback loops on the internal dependency and behaviors but this is largely an external integration. I would love to think about how we test this is operational. 
2. Documentation - right now this can be appending to the existing tutorial. There may be more opportunity for a more granular pass at argocd orchestration docs. 

Two minor requests:
1. Can you comment on an issue if/when you pick it up for work (or at least with the PR) - In doing so we can assign you and help prevent any colliding submissions (as well as remember to track for reviews on the project board. 
2. Add context to the PR description. It took me a fair bit of reading to tell that the associated issue was asking for a feature around oci registries when you actually solved for oci images. The former isn't actually as valid  and you solved the problem correctly here IMO - but I had to hunt down the delta and process. 

Thank you for the continued support and contributions - I very much appreciate it!

robinlieb:
@brandtkeller thanks for the initial review. I added changes which addresses all of the mentioned points.
Added two applications in the Argo CD examples showcasing the the use of Helm Charts and manifests from OCI.
Added E2E test on base of this example.
Also refactored the patch logic for Argo CD resources since this contained some duplication.

For the other two points, will definitely keep that in mind for future work.

Update: For the sake of proper documentation, should I create a new issue specifying the OCI support for manifests?

al-jeyapal:
The issue we might have with this is that ArgoCD will try to hit the registry over HTTPS. We were getting this in the Argo logs:

```
Failed to load target state:
 failed to generate manifest for source 1 of 2:
   rpc error: code = Unknown desc = failed to resolve revision "0.0.0":
     cannot get digest for revision 0.0.0:
       Head "https://zarf-docker-registry.zarf.svc.cluster.local:5000/v2/helm-charts/manifests/0.0.0":
         http: server gave HTTP response to HTTPS client
```

Helm allows you to use http via:

```
 helm pull --plain-http oci://...
```

However it doesn't look like Argo exposes an option for that. It exposes a `--insecure-skip-server-verification` but that only allows it to trust unverified certs.

We're currently trying to work around this by putting an internal `caddy` service to act as a HTTPS proxy in front of the zarf registry (on #4327). Still working through the details.

The easier option would be to have a HTTPS endpoint exposed on the zarf registry (just internally as a service not as an ingress/GatewayAPI route), even if it just uses a self-signed certificate. Not sure if that's a route we want to take though.

robinlieb:
The [failed pipeline](https://github.com/zarf-dev/zarf/actions/runs/20178960749) Is probably related to the version bump of Argo CD, which throws an error with the gitea.localhost setup in the test. Not sure yet what exactly changed in Argo CD but will investigate further on how to resolve that.

robinlieb:
External test fails because `gitea.localhost` gets resolved to 127.0.0.1 when calling git / libcurl in Argo CD Repo Server. This behavior changed with Helm 7.4.0 / Argo CD v2.12.0, where Argo CD moved to ubuntu:24.04 base image, which is stricter / RFC 6761 conform in terms of localhost resolution.
Added `curloptResolve` option for git config in repo sever.
Alternative approach would move from `gitea.localhost` to e.g. `gitea.local` in tests, but that would require to add this entry in `/etc/hosts` on local machines running external tests and in pipeline.

al-jeyapal:
Thanks for the work on this @robinlieb. Out of curiosity, how did you work around ArgoCD forcing Helm to use HTTPS when interacting with OCI?  We had to put a HTTPS proxy in place to work around that.

robinlieb:
@al-jeyapal setting `insecureOCIForceHttp` to `true` for the repository secret should make that possible. Have you tried that? Had you also the chance to build and test the branch and could provide feedback if that world work on your end?

al-jeyapal:
> @al-jeyapal setting `insecureOCIForceHttp` to `true` for the repository secret should make that possible. Have you tried that? Had you also the chance to build and test the branch and could provide feedback if that world work on your end?

Thanks @robinlieb I hadn't seen any references to that in the doco, but can see it [in the code](https://github.com/argoproj/argo-cd/blob/c4f3bb8be401e3feb2bf8a85334bd4998eebbbc3/pkg/apis/application/v1alpha1/repository_types.go#L117). Very helpful.

I'm hoping to be rotated back onto the project later this month, so will aim to give it a go when next on.

src/internal/agent/hooks/argocd-repository.go
```diff
@@ -107,11 +101,29 @@ func mutateRepositorySecret(ctx context.Context, r *v1.AdmissionRequest, cluster
...
-	patches = append(patches, operations.ReplacePatchOperation("/data/password", base64.StdEncoding.EncodeToString([]byte(gitServer.PullPassword))))
+	patches = append(patches, operations.ReplacePatchOperation("/data/username", base64.StdEncoding.EncodeToString([]byte(username))))
+	patches = append(patches, operations.ReplacePatchOperation("/data/password", base64.StdEncoding.EncodeToString([]byte(password))))
+
+	if isOCIURL && registryInfo.IsInternal() && !useMTLS {
+		patches = append(patches, operations.ReplacePatchOperation("/data/insecureOCIForceHttp", base64.StdEncoding.EncodeToString([]byte("true"))))
+	}
+
+	if useMTLS && isOCIURL {
+		patches = append(patches, operations.ReplacePatchOperation("/data/tlsClientCertData", base64.StdEncoding.EncodeToString([]byte(cluster.RegistryClientTLSSecret))))
```
brandtkeller: If I am following correctly - this is patching the `secret` tlsClientCertData with the constant `"zarf-registry-client-tls"` value. But I wonder if this is instead supposed to be the actual certificate data.... 

Documentation is a little ambiguous but the statement:
> tlsClientCertData and tlsClientCertKey refer to secrets where a TLS client certificate (tlsClientCertData) and the corresponding private key tlsClientCertKey are stored for accessing the repositories

from the `repository` perspective makes me think that from the `secret` perspective that we are mutating here that the data is not by reference/name but rather direct. 

I need to validate this further.

brandtkeller (changes requested):
Making my way through each file - only one item stood out as a blocker thus far.

examples/argocd/baseline/values.yaml
```diff
@@ -19,7 +19,45 @@ configs:
...
+# /etc/hosts entries.
+extraObjects:
+  - apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: git-config-localhost-resolve
+    data:
+      gitconfig: |
+        [http]
+            curloptResolve = gitea.localhost:3000:172.31.0.99
```
brandtkeller: appreciate the callout on this change. Thinking out-loud: this has no effect on standard `zarf init` installations with in-cluster gitea as state will point to the gitea kubernetes service dns (which doesn't match gitea.localhost). 

I'm not opposed to keeping this here but I would be interested in any options for keeping it within the scope of tests.

brandtkeller: @robinlieb How would you feel about trying to implement this addition as a deploy-time values using the `values` feature?

robinlieb: Changed it to use deploy-time values.

src/internal/agent/hooks/argocd-repository.go
robinlieb: Updated it that both cert and key is used in the Argo CD Repo secret. Updated the test accordingly.

src/internal/agent/hooks/argocd-application.go
```diff
@@ -118,6 +129,82 @@ func getPatchedRepoURL(ctx context.Context, repoURL string, gs state.GitServerIn
...
+		return repoURL, nil
+	}
+
+	if isOCIURL {
+		return mutateOCIURL(ctx, repoURL, registryAddress, isPatchedClusterIP)
+	}
+	return mutateGitURL(ctx, repoURL, gs)
+}
+
+func shouldMutateURL(operation v1.Operation, isOCIURL bool, repoURL, registryAddress, clusterIP string, gs state.GitServerInfo) (bool, bool, error) {
```
brandtkeller: Should we unit test and and doccomments for this function? three return values is slightly confusing without any supporting context. I also want to ensure we capture edge cases.

src/internal/agent/hooks/argocd-repository.go
```diff
@@ -107,11 +103,30 @@ func mutateRepositorySecret(ctx context.Context, r *v1.AdmissionRequest, cluster
...
+
+	if useMTLS && isOCIURL {
+		patches = append(patches, operations.ReplacePatchOperation("/data/tlsClientCertData", cert.Cert))
+		patches = append(patches, operations.ReplacePatchOperation("/data/tlsClientCertKey", cert.Key))
+	}
 
 	return patches
 }
+
+// Helper for getting eiher git server of registry creds
```
brandtkeller:
```suggestion
// Helper for getting either git server of registry creds
```

brandtkeller (changes requested):
Minor changes - running this through local testing now to understand the implications.

src/internal/agent/hooks/argocd-application.go
robinlieb: Added docs and unit tests.

src/internal/agent/hooks/argocd-repository.go
robinlieb: Added change.

examples/argocd/zarf.yaml
```diff
@@ -8,18 +8,24 @@ components:
...
-        version: 5.54.0
+        version: 9.0.5
         namespace: argocd
         url: https://argoproj.github.io/argo-helm
         releaseName: argocd-baseline
         valuesFiles:
           - baseline/values.yaml
+        # Required for 'src/test/external' to validate external git server.
+        values:
+          - sourcePath: ".extraObjects"
```
brandtkeller, line 19: Deployment of the package will fail if you simply `zarf package create examples/argocd` now unless you provide the required values files on deploy. 

I think the strategy as it stands is sound for adding the testing information at-deploy - but we should include a default values file in the package which has these values specified (empty or defualt) so that the package can still deploy without deploy-time values.

brandtkeller (changes requested):
I believe tests may be failing related to the comment I made on the argocd package manifest itself. 

Otherwise I'd request one other change (once the argocd package is updated) which would be to run a package create/deploy under the `src/test/proxy/proxy_test.go` in `Test_3_OCIOpsPackage()` such that we continue to test argo for the proxy suite. 

I tested locally - it looks good to me and I can approve if all is good after updates.

brandtkeller:
Recent PR's touched cluster state - if you can update with main then I'll evaluate any additional testing on airgapped clusters and advocate to get this approved/merged. Thanks for the patience.

robinlieb:
> Recent PR's touched cluster state - if you can update with main then I'll evaluate any additional testing on airgapped clusters and advocate to get this approved/merged. Thanks for the patience.

Rebased.

src/internal/agent/hooks/argocd-application.go
brandtkeller: Looking at the new components - I believe we're enabling the orchestration of applications with argocd using only OCI - is that correct?

We made this change recently for an early return when the git server is not configured (for git repo mutations). Now that we have OCI do we need to check for scheme earlier and then determine which service is required?

brandtkeller (commented):
Recent change I want to surface - apologies for the churn but this shouldn't be too difficult of a change.

src/internal/agent/hooks/argocd-application.go
robinlieb: Good catch, thanks!
Updated that to check for both to have an early return, then check the repoURL for each individual source in the Application, in case there are mixed sources.

zeljkobekcic:
Hey, I would really like to use this feature. Can I expect this feature to be in the next release?

brandtkeller:
> Hey, I would really like to use this feature. Can I expect this feature to be in the next release?

In review! Next release is tomorrow - so if not that release then the following. 

@robinlieb Can you update with main before I run CI? We had some changes that will likely make them fail if I run now. 

Review nearly complete.

robinlieb:
> > Hey, I would really like to use this feature. Can I expect this feature to be in the next release?
> 
> In review! Next release is tomorrow - so if not that release then the following.
> 
> @robinlieb Can you update with main before I run CI? We had some changes that will likely make them fail if I run now.
> 
> Review nearly complete.

@brandtkeller rebased.

src/internal/agent/hooks/argocd-repository.go
brandtkeller: Doing some testing - Do we need to handle registry-only (OCI) for this resource?

src/internal/agent/hooks/argocd-appproject.go
brandtkeller: Are there OCI implications that we need to consider here as well?

brandtkeller (changes requested):
I think there are tests implicitly passing with the argo package because we also have the git server configured (required for the package to fully deploy) and the corresponding logic can handle OCI - but when you disable the git server and remove the git-server related elements from the package it does not deploy the applications as expected.

src/internal/agent/hooks/argocd-repository.go
robinlieb: Unified the check to conform with applications. Added tests for that.

src/internal/agent/hooks/argocd-appproject.go
robinlieb: Unified the check to conform with applications. Added tests for that.

src/internal/agent/hooks/argocd-application.go
```diff
@@ -122,6 +142,95 @@ func getPatchedRepoURL(ctx context.Context, repoURL string, gs state.GitServerIn
...
+// shouldMutateURL reports whether the given repoURL should be mutated for the admission operation.
+//
+// shouldMutate is true for Create operations and for Update operations where repoURL has not
+// already been patched to the Zarf-managed registry or git server address.
+//
+// isPatchedClusterIP is true when repoURL already matches the registry cluster-IP address
+func shouldMutateURL(operation v1.Operation, isOCIURL bool, repoURL, registryAddress, clusterIP string, gs state.GitServerInfo) (shouldMutate bool, isPatchedClusterIP bool, err error) {
+	isCreate := operation == v1.Create
+	isUpdate := operation == v1.Update
+	if isCreate {
```
brandtkeller: We removed checks for the create operation #4691, instead reaching for idempotent behaviors. There were edgecases regarding pre-mutated urls on create. I would look to verify if this is required or if the opreation is idempotent regardless.

src/internal/agent/hooks/common.go
```diff
@@ -20,6 +21,13 @@ import (
 	"oras.land/oras-go/v2/registry/remote/auth"
 )
 
+const (
+	// AgentErrTransformGitURL is thrown when the agent fails to make the git url a Zarf compatible url
+	AgentErrTransformGitURL = "unable to transform the git url"
```
brandtkeller, line 26: nit: if we have these constants then they likeky qualify as good sentinel errors. I won't block on this.

src/internal/agent/hooks/argocd-appproject.go
```diff
@@ -53,24 +53,32 @@ func mutateAppProject(ctx context.Context, r *v1.AdmissionRequest, cluster *clus
...
+	if err != nil {
+		return nil, err
+	}
 
 	patches := make([]operations.PatchOperation, 0)
 
 	for idx, repo := range proj.Spec.SourceRepos {
-		patchedURL, err := getPatchedRepoURL(ctx, repo, s.GitServer)
+		patchedURL, err := getPatchedRepoURL(ctx, repo, registryAddress, clusterIP, s.GitServer, r)
 		// The AppProject can also include source repositories like '*' (as in the default project),
```
brandtkeller, line 84: Something we should address in the near future is detecting wildcards before the patch operation. This current flow could open up quite a few avenues for unintended consequences on misconfiguration.

src/internal/agent/hooks/argocd-application.go
```diff
@@ -58,31 +58,46 @@ func NewApplicationMutationHook(ctx context.Context, cluster *cluster.Cluster) o
...
+	registryAddress, clusterIP, err := cluster.GetServiceInfoFromRegistryAddress(ctx, s.RegistryInfo)
+	if err != nil {
+		return nil, err
+	}
+
+	l.Info("mutating the ArgoCD Application",
 		"name", app.Name,
-		"operation", r.Operation,
-		"gitServer", s.GitServer.Address)
+		"operation", r.Operation)
```
brandtkeller: Minor nit - we lose the ability to track what the intended mutation was going to try - this has been useful in debugging historically.

src/internal/agent/hooks/argocd-application.go
brandtkeller: This check is missing the asymmetry for both git server address and registry address. shouldMutateURL also performs this check in a sense. So I think we just need to resolve how and where to best perform that validation.

src/internal/agent/hooks/argocd-applicationset.go
```diff
@@ -64,25 +64,40 @@ func mutateApplicationSet(ctx context.Context, r *v1.AdmissionRequest, cluster *
...
+
+	if !anyZarfServiceUsable(requiresGit, requiresRegistry, s) {
+		l.Debug("no Zarf services configured for source URL schemes, skipping ArgoCD ApplicationSet mutation")
+		return &operations.Result{Allowed: true}, nil
+	}
+
+	l.Info("mutating the ArgoCD ApplicationSet",
 		"name", appSet.Name,
-		"gitServer", s.GitServer.Address)
+		"operation", r.Operation)
```
brandtkeller: Same comment for losing address in the logs

brandtkeller (changes requested):
Some of my last comments (hopefully) - only other thing I was going to look at is where we landed for testing with the proxy init model and mTLS.

src/internal/agent/hooks/argocd-applicationset.go
robinlieb: Added both, git server and registry.

src/internal/agent/hooks/argocd-application.go
robinlieb: Added both, git server and registry.

src/internal/agent/hooks/argocd-application.go
robinlieb: Removed the operation checks.

src/test/e2e/22_git_and_gitops_test.go
```diff
@@ -192,15 +192,25 @@ func waitFluxPodInfoDeployment(t *testing.T) {
 func waitArgoDeployment(t *testing.T) {
 	// Deploy the argocd example and verify that it works
 	tmpdir := t.TempDir()
-	stdOut, stdErr, err := e2e.Zarf(t, "package", "create", "examples/argocd", "-o", tmpdir, "--skip-sbom")
+	stdOut, stdErr, err := e2e.Zarf(t, "package", "create", "examples/argocd", "-o", tmpdir, "--skip-sbom", "--features=\"values=true\"")
```
brandtkeller: For sake of consistency and not introducing any weirdness with cobras stringtostring parser let's be consistent with the features arguments we use elsewhere
```suggestion
	stdOut, stdErr, err := e2e.Zarf(t, "package", "create", "examples/argocd", "-o", tmpdir, "--skip-sbom", "--features", "values=true")
```

src/test/e2e/22_git_and_gitops_test.go
```diff
@@ -192,15 +192,25 @@ func waitFluxPodInfoDeployment(t *testing.T) {
...
 	require.NoError(t, err, stdOut, stdErr)
 	packageName := fmt.Sprintf("zarf-package-argocd-%s.tar.zst", e2e.Arch)
 	path := filepath.Join(tmpdir, packageName)
-	stdOut, stdErr, err = e2e.Zarf(t, "package", "deploy", path, "--components=argocd-apps", "--confirm")
+	stdOut, stdErr, err = e2e.Zarf(t, "package", "deploy", path, "--components=argocd-apps", "--confirm", "--features=\"values=true\"")
```
brandtkeller:
```suggestion
	stdOut, stdErr, err = e2e.Zarf(t, "package", "deploy", path, "--components=argocd-apps", "--confirm", "--features", "values=true")
```

src/test/external/ext_in_cluster_test.go
```diff
@@ -107,7 +107,7 @@ func (suite *ExtInClusterTestSuite) Test_0_Mirror() {
 	// Use Zarf to mirror a package to the services (do this as test 0 so that the registry is unpolluted)
 	t := suite.T()
 	tmpdir := t.TempDir()
-	err := exec.CmdWithPrint(zarfBinPath, "package", "create", "../../../examples/argocd", "-o", tmpdir, "--skip-sbom")
+	err := exec.CmdWithPrint(zarfBinPath, "package", "create", "../../../examples/argocd", "-o", tmpdir, "--skip-sbom", "--features=\"values=true\"")
```
brandtkeller:
```suggestion
	err := exec.CmdWithPrint(zarfBinPath, "package", "create", "../../../examples/argocd", "-o", tmpdir, "--skip-sbom", "--features", "values=true")
```

src/test/external/ext_out_cluster_test.go
```diff
@@ -110,7 +136,7 @@ func (suite *ExtOutClusterTestSuite) Test_0_Mirror() {
 	// Use Zarf to mirror a package to the services (do this as test 0 so that the registry is unpolluted)
 	t := suite.T()
 	tmpdir := t.TempDir()
-	err := exec.CmdWithPrint(zarfBinPath, "package", "create", "../../../examples/argocd", "-o", tmpdir, "--skip-sbom")
+	err := exec.CmdWithPrint(zarfBinPath, "package", "create", "../../../examples/argocd", "-o", tmpdir, "--skip-sbom", "--features=\"values=true\"")
```
brandtkeller:
```suggestion
	err := exec.CmdWithPrint(zarfBinPath, "package", "create", "../../../examples/argocd", "-o", tmpdir, "--skip-sbom", "--features", "values=true")
```

brandtkeller (changes requested):
Minor changes - please rebase with main and I'll kickoff CI and approve/advocate for second review. Appreciate the patience again.

robinlieb:
> Minor changes - please rebase with main and I'll kickoff CI and approve/advocate for second review. Appreciate the patience again.

Rebased and proposed changes applied, thanks!

brandtkeller (approved):

AustinAbro321 (approved):
LGTM, nice work

PR #4358, opened 2025-11-12
docs: update custom init tutorial

### LINKED ISSUE #4353: Breaking change introduced without doc updates
[PR 3956](https://github.com/zarf-dev/zarf/pull/3956) introduced a breaking change that made it so you could no longer build custom init packages without certain proxy values set, but there is nothing about it in the docs. A search for "proxy_image" doesn't return any results anyway. Why have a "Checklist before merging" on the PR if you're just going to ignore it?

AustinAbro321:
## Description

Update custom init tutorial to include the new proxy image, so that users creating custom init packages can handle the breakages. The general init package page still needs updates, these will happen in a separate PR as the registry proxy feature matures

Fixes #4353 

## Checklist before merging


mkcp (approved):
LGTM

PR #4365, opened 2025-11-13
refactor: remove init from src/cmd

### LINKED ISSUE #2773: Rethink Cobra code structure to remove use of global variables

AustinAbro321:
## Description

Simplifies the cmd layer by removing the call to the go builtin init()

## Related Issue

Fixes #2773 

## Checklist before merging

### DIFF
--- a/src/cmd/root.go
+++ b/src/cmd/root.go
@@ -258,8 +258,7 @@ func Execute(ctx context.Context) {
 		os.Exit(1)
 	}
 
-	// NOTE(mkcp): This line must be run with the unconfigured default logger because user flags are set downstream
-	// in rootCmd's preRun func.
+	// Use default logger in case there was an error prior to the logger being setup
 	logger.Default().Error(err.Error())
 	os.Exit(1)
 }


mkcp (approved):
LGTM

PR #4367, opened 2025-11-14
ci: cleanup pre-installed tools more aggressively and always log cluster info in test-upgrade

AustinAbro321:
## Description

Upgrade is failing because of disk-pressure, likely other jobs are not far off. I added some more tools to our cleanup step. I used this article as a reference https://mathio28.medium.com/squeezing-disk-space-from-github-actions-runners-an-engineers-guide-d5fbe4443692

I noticed that cluster logs are not actually being given on failure during test-upgrade. Also we're only giving logs for a specific deployment despite running the full e2e suite.

## Checklist before merging

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -183,7 +183,7 @@ jobs:
           GORELEASER_CURRENT_TAG: ${{ github.ref_name }}
 
       - name: Report disk space on failure
-        if: always()
+        if: failure()
         run: |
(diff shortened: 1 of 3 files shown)


mkcp (dismissed):
LGTM

mkcp (approved):

a1994sc:
It is cases like this I wish you could bring your own build image....

PR #4368, opened 2025-11-14
docs(values,templating): add reference and best practices pages

mkcp:
## Description
Add reference and best practices documentation for package values and templating.

## Related Issue
Relates to #3946 

## Checklist before merging


a1994sc (dismissed):
LGTM

site/src/content/docs/best-practices/package-values.mdx
```diff
@@ -0,0 +1,240 @@
...
+```
+
+:::note
+Conditionals are powerful but can quickly become confusing. If you find yourself needing complex conditionals, see if
+you can instead express the problem with data.
+:::
+
+See the [Templating Reference](/ref/templating/) for complete function documentation.
+
+## Avoid Sensitive Data in Values Files
```
brandtkeller: Generally this is good advice but I do think it is worth calling out that values files will be the primary mechanism for orchestrating sensitive values. Given the declarative delivery model - all secrets have to originate from _somewhere_. 

It will be important for sunsetting zarf variables as it will provide a way to break these out of zarf config files.

We may want to delineate between sensitive values defaults getting "baked" into the package - for which you had a few good examples of different environment overrides that would align to that story telling.

site/src/content/docs/best-practices/templating.mdx
```diff
@@ -0,0 +1,390 @@
...
+    {{ .Values.app.config | toYaml }}
+```
+
+**Choose the right function:**
+- `indent N` - Indent every line by N spaces
+- `nindent N` - Add newline, then indent by N spaces
+
+## Common Errors and Solutions
+
+### Template Parse Errors
```
brandtkeller, line 188: ++ I like that we are making errors in docs discoverable.

site/src/content/docs/ref/package-values.mdx
```diff
@@ -0,0 +1,427 @@
...
+            targetPath: ".config.database.host"
+
+          # Multiple mappings to the same target - last one wins
+          - sourcePath: ".database.host"
+            targetPath: ".config.database.host"
+          - sourcePath: ".database.host_override"
+            targetPath: ".config.database.host"  # Takes priority
+```
+
+#### Mapping All Values
```
brandtkeller, line 223: nit: we may want to refine this section at some point to help users understand the best practices for building a package with values files that aligns to broader zarf hardening goals.

said another way - I think we should encourage using sourcePath/targetPath as a best practice for package hardening.

site/src/content/docs/ref/templating.mdx
```diff
@@ -0,0 +1,448 @@
...
+  order: 37
+---
+
+:::caution
+
+Templating with Package Values is currently an **alpha feature** and requires the `--features="values=true"` flag when creating, deploying, inspecting, or removing packages.
+
+:::
+
+Zarf supports Go template syntax and `text/template` engine with powerful features including control flow and functions. Zarf includes the full library of [Sprig functions](http://masterminds.github.io/sprig/) and some builtins for dynamic configuration in manifests, files, and actions. This powerful templating system enables complex transformations, conditional logic, and data manipulation beyond simple string substitution.
```
brandtkeller: This may be a good place to identify what is different between zarf templating and helm templating.

brandtkeller (changes requested):

site/src/content/docs/best-practices/package-values.mdx
AustinAbro321, line 1: Across this PR we reference `zarf package inspect`, `zarf dev inspect` and using the values schema however none of these features are merged into Zarf as of yet. IMO we should wait until a feature is merged before documenting it

site/src/content/docs/best-practices/templating.mdx
```diff
@@ -0,0 +1,390 @@
...
+
+```yaml
+# Without quote, this breaks YAML parsing
+apiVersion: {{ .Values.version }}  # If version is "1.0", breaks
+apiVersion: {{ .Values.version | quote }}  # "1.0" works correctly
+```
+
+## Provide Defaults for Optional Values
+
+Use the `default` function to provide fallback values:
```
AustinAbro321: While defaults are generally a good practice, I don't think they really make sense in the context of Zarf since we always validate that there is a value for template. I could be missing scenarios but AFAIK there is no way for the default to come into play

site/src/content/docs/best-practices/package-values.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+**Benefits:**
+- Catch configuration errors before deployment
+- Provide clear error messages to users
+- Document valid values and constraints
+- Enable IDE autocomplete and validation
+
+## Leverage Templating Functions
+
+Use [functions](/ref/templating/) for dynamic configuration:
```
AustinAbro321: This should link to ref/templating/#template-functions

site/src/content/docs/best-practices/templating.mdx
```diff
@@ -0,0 +1,390 @@
...
+
+    app: nginx
+
+# With whitespace control:
+metadata:
+  labels:
+    app: nginx
+```
+
+## Comment Complex Template Logic
```
AustinAbro321: I feel that the advice to comment complex template logic is not relevant at this level of documentation. If I'm a user coming to this page, I'm looking for specifically templating advice, rather than general coding advice. I feel that this also gives the page a AI generated feel.

site/src/content/docs/best-practices/templating.mdx
```diff
@@ -0,0 +1,390 @@
...
+  # Numbers must be strings in ConfigMaps
+  port: {{ .Values.app.port | toString | quote }}
+  replicas: {{ .Values.app.replicas | toString | quote }}
+
+  # Booleans to strings
+  debug: {{ .Values.app.debug | toString | quote }}
+
+# ❌ RISKY: Type mismatch errors
+data:
+  port: {{ .Values.app.port }}  # May fail if port is a number
```
AustinAbro321: Numbers don't have to be strings in configmaps, Kubernetes will store either

site/src/content/docs/best-practices/templating.mdx
```diff
@@ -0,0 +1,390 @@
...
+**What to check:**
+- Templates render without errors
+- Output looks as expected
+- No unexpected whitespace
+- Values are correctly interpolated
+- Conditionals evaluate correctly
+
+## Handle Missing Values Gracefully
+
+Protect against missing or nil values:
```
AustinAbro321: I don't think this advice makes sense in the context of Zarf, the "good" example will still fail if .values.monitoring.enabled is not set

site/src/content/docs/best-practices/templating.mdx
```diff
@@ -0,0 +1,390 @@
...
+**Solution:**
+```yaml
+# Use default for optional values
+value: {{ .Values.app.missing | default "fallback" }}
+
+# Or check if exists first
+{{- if .Values.app.missing }}
+value: {{ .Values.app.missing }}
+{{- end }}
+```
```
AustinAbro321, line 199: Both of these solutions would not solve the error.

site/src/content/docs/best-practices/templating.mdx
```diff
@@ -0,0 +1,390 @@
...
+```yaml
+# Use whitespace control operators
+{{- if .Values.app.debug -}}
+debug: true
+{{- end -}}
+```
+
+### Quoting Problems
+
+**Problem:** YAML parsing fails with boolean-like strings
```
AustinAbro321, line 228: I think I understand why we would use the quote function at times, we wouldn't want a type to be interpreted as the incorrect type. Are there situations where YAML parsing would fail?

AustinAbro321 (changes requested):
Leaving comments on the review so far, I'll aim to do a review on the ref section at another point

site/src/content/docs/best-practices/templating.mdx
mkcp: Yeah I agree with that - I thought about taking it out during editing and left it in for whatever reason. Removed in latest

site/src/content/docs/best-practices/templating.mdx
mkcp: Thoughts on generalizing the example a bit then, like "explicit type conversion may be necessary?"

site/src/content/docs/best-practices/templating.mdx
mkcp: Yeah I agree, I don't like encouraging conditional blocks like this either. Removed in latest

site/src/content/docs/best-practices/templating.mdx
mkcp: I think there's a specific yaml spec version where this was a problem, I don't know if this specifically affects Zarf though. This is probably covered under the more general guidance to quote strings.

site/src/content/docs/best-practices/package-values.mdx
mkcp: I believe that this is addressed now

site/src/content/docs/best-practices/package-values.mdx
mkcp: This is a good callout - I rewrote this section to highlight not committing secrets to package defaults, and gave examples of how they can be passed in.

site/src/content/docs/best-practices/templating.mdx
mkcp: Trimmed this section down substantially in latest

site/src/content/docs/ref/package-values.mdx
mkcp: I added a note to the beginning that passing only the necessary values provides better security

site/src/content/docs/ref/templating.mdx
mkcp: I added some notes about the similarities and differences here

site/src/content/docs/best-practices/package-values.mdx
```diff
@@ -0,0 +1,252 @@
...
+:::note
+Conditionals are powerful but can quickly become confusing. If you find yourself needing complex conditionals, see if
+you can instead express the problem with data.
+:::
+
+See the [Templating Reference](/ref/templating/) for complete function documentation.
+
+## Do Not Commit Sensitive Values in Package
+
+Values are the best way to pass secrets into your Zarf package, but those secrets should not be baked into the values files a package is created with:
```
brandtkeller, line 171: ++

site/src/content/docs/ref/package-values.mdx
```diff
@@ -0,0 +1,424 @@
...
+
+import Properties from '@components/SchemaItemProperties.astro';
+
+:::caution
+
+Package Values is currently an **alpha feature** and requires the `--features="values=true"` flag when creating, deploying, inspecting, or removing packages. The API may change in future releases.
+
+:::
+
+Package Values provide a familiar way to define and reusable configuration data in your Zarf packages. Unlike the legacy [Variables and Constants](/ref/values/) system that uses string substitution with `###ZARF_VAR_###` syntax, Package Values uses Go templates with access to structured data and types, [go's templating engine](https://pkg.go.dev/text/template), and custom functions like those offered by [Sprig](/ref/templating/).
```
brandtkeller: Minor grammatical item: 

> a familiar way to define and reusable configuration data

site/src/content/docs/ref/package-values.mdx
```diff
@@ -0,0 +1,424 @@
...
+            setValues:
+              - name: runtime.settings
+                type: yaml
+```
+
+The `setValues` field accepts:
+- `name`: The path where the value should be stored (e.g., `database.password`)
+- `type`: The format of the output - `string` (default), `json`, or `yaml`
+
+Values set via `setValues` are available in subsequent actions and manifests within the same deployment.
```
brandtkeller, line 189: Making a note in-passing that an example here may be valuable (is that a pun?) - will resolve if included elsewhere.

site/src/content/docs/best-practices/package-values.mdx
```diff
@@ -0,0 +1,252 @@
+---
+title: Package Values
+sidebar:
+  order: 10
+---
+
+This guide provides best practices for using [Package Values](/ref/package-values/) in your Zarf packages.
```
brandtkeller, line 7: It's briefly touched on in the `ref` document - but If I think of the `ref` document as the "how" and best practices as the "why" - I wonder if a good addition here could be the mention of why the use of granular sourcePath/targetPath. Security considerations and treating it as an allow list as well as traceability when we think about potential collisions with on-deployment values files and the future of imports.

site/src/content/docs/ref/package-values.mdx
```diff
@@ -0,0 +1,424 @@
...
+
+Provide entire values files with the `-f` or `--values` flag:
+
+```bash
+# Deploy with custom values file
+zarf package deploy my-package.tar.zst \
+  --features="values=true" \
+  -f custom-values.yaml
+
+# Multiple values files (later files override earlier ones)
```
brandtkeller, line 297: Is the shorthand for this flag `-v`? (doesn't exist in dev deploy yet but `-f` is used already in dev deploy)

brandtkeller (changes requested):
Minor comments as I iterate through some tests - otherwise I like how clean and comprehensive this is.

site/src/content/docs/ref/templating.mdx
```diff
@@ -0,0 +1,448 @@
...
+- List operations (join, len, first, sortAlpha)
+- Default values and conditionals
+- Math operations (add, mul, max, min)
+- Encoding and hashing (b64enc, sha256sum)
+- Multi-line values with indentation
+- Real-world ConfigMap and Deployment examples
+
+## Using Helm charts within Zarf:
+- Helm templating happens inside chart templates (standard Helm behavior)
+- Zarf templating happens at the package level
```
AustinAbro321: The sentence Zarf templating happens a the package level, feels like it doesn't quite fit. In the flow of deploying a Helm chart, Zarf templating never happens, instead we forward values to the Helm chart for templating to occur

site/src/content/docs/ref/package-values.mdx
mkcp: Yep, thanks for catching that - fixed.

site/src/content/docs/ref/package-values.mdx
mkcp: Fixed

site/src/content/docs/ref/templating.mdx
mkcp: Clarified in latest

brandtkeller:
Closing as completed in #5090

PR #4369, opened 2025-11-15
feat(sign): support for sigstore bundle format

### LINKED ISSUE #4276: docs: clarify asymmetric KMS key needed for signing
Docs remain vague on using `--signing-key`. These two sites (https://docs.zarf.dev/commands/zarf_package_create/ and https://docs.zarf.dev/commands/zarf_package_publish/) state:
`--signing-key string          Private key for signing packages. Accepts either a local file path or a Cosign-supported key provider`

Cosign supports KMS keys, but they must be an asymmetric (vs symmetric) key (which was also never really found in the Cosign documentation explicitly). Could we clarify the need for asymmetric keys in the documentation somewhere?

- **Given** existing Zarf docs 
- **When** looking at `zarf package create --signing-key` and `zarf package publish --signing-key`
- **Then** we clarify in the docs that if using a KMS key, it needs to be asymmetric

No change to documentation

### Additional context
Would love to see a tutorial on the website specifically focusing on signing Zarf packages with a

### LINKED ISSUE #4296: Support for cosign bundle format

Investigate and propose supporting the [sigstore bundle format](https://docs.sigstore.dev/about/bundle/) for signature composition.  

- **Given** a package exists
- **When** signing that package occurs
- **Then** a bundle is created for verification purposes

Continue to stick to the current signature strategy.

### Additional context

This should offer more verification opportunities for offline environments. 

> A Sigstore bundle is everything required to verify a signature on an artifact. This is satisfied by the Verification Material and signature Content.

brandtkeller:
## Description

Migrates Zarf package signing to the Sigstore bundle format, aligning with Cosign v3 standards and improving signature portability and verification capabilities.

## Changes

The signing implementation now defaults to `NewBundleFormat: true` and generates both legacy signature files and the new bundle format (saved as `zarf.bundle.sig`). The `SignPackage()` function automatically configures the bundle path through the BundlePath option, ensuring all signing operations produce both formats for backward compatibility. 

Verification logic implements a fallback strategy that prefers the bundle format over legacy signatures. When verifying packages, the system first checks for a bundle file, and if not found, falls back to the legacy signature format with a deprecation warning to inform users of the transition. Validation Package integrity checks now exclude bundle files from validation, treating them similarly to how legacy signature files are handled to prevent false validation errors. 

The implementation maintains dual format support by generating both bundle and legacy signatures during signing operations. This enables a graceful migration path where verification prefers the modern bundle format but seamlessly falls back to legacy signatures with appropriate warnings. The bundle format is now enabled by default, aligning Zarf with Cosign v3 standards.

The bundle format significantly improves verification capabilities by including timestamps and transparency log entries, which enables verification of packages signed with short-lived certificates even after expiration. Existing packages with legacy signatures remain fully functional while new signatures automatically use the modern format. This sets the foundation for eventually deprecating legacy signature support.

## Related Issue

Fixes #4296
Fixes #4276 

## Checklist before merging


src/pkg/packager/layout/package.go
```diff
@@ -269,16 +283,31 @@ func (p *PackageLayout) VerifyPackageSignature(ctx context.Context, opts utils.V
...
+		if errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("signature not found: neither bundle nor legacy signature exists")
+		}
+		return fmt.Errorf("error checking legacy signature: %w", err)
 	}
 
-	// Note: this is the backwards compatible behavior
-	// this will change in the future
+	// Legacy signature found
+	l.Warn("non-bundle format signature is being deprecated in favor of the sigstore bundle format")
```
brandtkeller, line 341: Note the intent to deprecate the standard signature file at some point in the future. Allowing backwards compatibility while shifting all new package creations to using the bundle format.

brandtkeller:
Closing as a prototype ahead of the proposal and implementation in #4891

PR #4382, opened 2025-11-18
feat!: allow setting static injector nodeport 

### LINKED ISSUE #4376: The ability to assign a static NodePort to the zarf-injector
I'm looking to set up a POC with Zarf in air-gap environments.  We host rke2 clusters running on rhel 8/9 with disa stig profile and selinux enforcing.  The initial zarf injector is deployed with NodePort and random port chosen.  In the end, I'd like to add to automation:  deploy rke2 -> deploy zarf -> pull charts/images from hosted harbor registry (OR Zarf packages) -> deploy apps via zarf.

Since we configure what ports are allowed access in our hardened Linux systems, I'd like to have the selected nodeport pre-configured so that there are no failures in the zarf init deployment (zarf-registry).  

An alternative could be the ability to disable the zarf-injector feature after initial deployment.  Meaning, I tried to do a live upgrade of Zarf in our cluster, and it fails on not being able to pull images from zarf-injector due to the random nodeport assigned. Since we already host a harbor registry, I'd like to be able to use Zarf within the same RKE2 cluster as Harbor (for different purposes).  This would be during a transition phase if we decide to adopt Zarf.

### Additional context
In our particular environment, the end result would be for us to have a hdd with all packages/apps required for a full deployment upon inserting hdd on a host. The deployment of the apps could be using zarf package deploy or using rke2's built-in helmchartconfigs; or a combination of both.

AustinAbro321:
## Description

This changes the `--injector-hostport` flag to `--injector-port` and allows users to set the nodeport that the injector expose. 

This also fixes a bug with the injector-port flag in proxy mode, where we wouldn't save the port between init's so if a user re-ran init and didn't include the `--injector-port` flag they're daemonset would move back to the default. This differs from the rest of our state data, which is saved on subsequent init's

## Related Issue

Fixes #4376

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/injector.go
+++ b/src/pkg/cluster/injector.go
@@ -55,6 +55,11 @@ func (c *Cluster) StartInjection(ctx context.Context, tmpDir, imagesDir string, 
 
 	l.Info("creating Zarf injector resources")
 
+	svc, err := c.createInjectorNodeportService(ctx, injectorNodePort, registryNodePort, pkgName)
+	if err != nil {
+		return 0, err
+	}
+
 	payloadCmNames, shasum, err := c.CreateInjectorConfigMaps(ctx, tmpDir, imagesDir, injectorSeedSrcs, pkgName)
 	if err != nil {
 		return 0, err
@@ -70,11 +75,6 @@ func (c *Cluster) StartInjection(ctx context.Context, tmpDir, imagesDir string, 
 			corev1.ResourceMemory: resource.MustParse("256Mi"),
 		})
 	injectorImage, injectorNodeName, err := c.getInjectorImageAndNode(ctx, resReq)
-	if err != nil {
-		return 0, err
-	}
-
-	svc, err := c.createInjectorNodeportService(ctx, injectorNodePort, registryNodePort, pkgName)
 	if err != nil {
 		return 0, err
 	}


src/cmd/initialize.go
```diff
@@ -178,7 +177,7 @@ func (o *initOptions) run(cmd *cobra.Command, _ []string) error {
 		OCIConcurrency:         o.ociConcurrency,
 		SetVariables:           o.setVariables,
 		StorageClass:           o.storageClass,
-		InjectorHostPort:       o.injectorHostPort,
+		InjectorPort:           o.injectorPort,
```
brandtkeller, line 180: Do we want to perform any early validation of the port? say if a port outside the nodeport range was specified for the nodeport solution - we have to wait until post configmap injection to get the error.

AustinAbro321: Given that the nodeport range can be changed from the default, and I'm not sure there is a foolproof method of getting that setting outside of the cluster, I think it's better to let the Kube API server handle it.

brandtkeller: fair point - I'm good with keeping as-is - with the only other alternative to evaluate order of operations.

AustinAbro321: Good point, moved to the start given that we now put users in a situation where this could error

Issue #4384, opened 2025-11-19
Generate schema within Zarf rather than using go:embed


AustinAbro321:
### Describe what should be investigated or refactored
The schema is currently generated in hack/schema. We save the output of hack/schema to zarf.schema.json. This causes friction for SDK users who have to embed the schema before calling relevant functions in Zarf. It is not intuitive that they would need to set this. If the schema is upset they'll receive a panic during functions in the create flow. It would be better to automatically generate the schema directly in Zarf to avoid this situation. 

go:embed is also inconvenient during unit testing as we always have to set it.

Additionally, Zarf plans on introducing other schema versions #3433 . We don't want to be in a situation where we ask SDK users to embed multiple schemas. 

The downside of this is that it will require adding back the invopp/jsonschema to Zarf, however this is a small dependency and it comes with the advantage of Zarf gaining the ability to use methods on it's structs for more detailed jsonschema fields

PR #4385, opened 2025-11-19
refactor: generate schema purely through go

AustinAbro321:
## Description

This refactors the schema generation process so that it is done 100% through go code

Relates to #4384 

## Checklist before merging

### DIFF
--- a/hack/schema/main.go
+++ b/hack/schema/main.go
@@ -85,6 +85,5 @@ func main() {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-
 	fmt.Println(schema)
 }


zarf.schema.json
AustinAbro321, line 1: The order of fields is changed in this file, but it is equivalent json. I verified this through https://jsondiff.com/

mkcp: I got the same result when testing the file in this branch against the version in main.

hack/schema/main.go
```diff
@@ -12,6 +12,33 @@ import (
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
 )
 
+// addYAMLExtensions walks through the JSON schema and adds patternProperties
+// for "x-" prefixed fields to any object that has "properties".
+// This allows YAML extensions (custom fields starting with x-) to be valid.
+func addYAMLExtensions(data map[string]any) {
+	if _, hasProperties := data["properties"]; hasProperties {
```
mkcp, line 19: nits, optional: extract "properties", "patternProperties", and "^x-" to package consts with semantic names.

PR #4392, opened 2025-11-20
refactor!: generate schema within Zarf

### LINKED ISSUE #4384: Generate schema within Zarf rather than using go:embed

AustinAbro321:
## Description

This deletes the hack/schema folder and instead directly embeds the schema within Zarf. This makes it easier for SDK users to consume as they will no longer have to set lint.ZarfSchema, it's now automatically embedded for them. 

Important to note that this creates a file called src/pkg/schema/zarf-v1alpha1-schema.json, and keeps the existing zarf-schema.json at the root of the repo. We refer to the root zarf-schema.json in our docs and many users use it for IDE integration. Likely in the future this generate code will generate each individual schema, and the zarf-schema.json will be a schema file that works with every schema, in accordance with  https://github.com/zarf-dev/proposals/pull/49. 

Breaking change as lint.ZarfSchema is deleted, but the fix is very simple, you can just delete it. 

## Related Issue

Fixes #4384

## Checklist before merging


src/pkg/schema/schema.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	_ "embed"
+)
+
+//go:embed zarf-v1alpha1-schema.json
+var v1Alpha1Schema []byte
+
+// GetV1Alpha1Schema returns the embedded JSON schema for the v1alpha1 Zarf package config
+func GetV1Alpha1Schema() []byte {
```
mkcp, line 17: much better 👍

PR #4393, opened 2025-11-20
feat: add ability to store package documentation 

### LINKED ISSUE #4346: Support for zarf package reports/documentation

Currently, there is no standardized method for retrieving release notes, readmes, and other documents associated with a package other than tracing that package back to its source repository to view those documents.  This feature request would add the ability to package creators to attach additional documentation to packages, which would then be carried with the package through to the destination environment.

This would assist with the following two Zarf user personas (at a minimum)
1. Developers working in airgapped environments looking to gain additional context on packages without needing to leave that environment for an internet connection
1. non-technical external registry users, such as UDS Registry users, could view documentation in a more user friendly manner

**During Package Creation**
- **Given** a repository with a zarf.yaml, a readme/release notes/some other documentation
- **When** a package developer runs `zarf package create [--some-new-flag]`
- **Then** the package is created and documentation is embedded in the produced package.  The location of the documentation within an OCI artifact is parsable from package metadata.

**During Package Usage**
- **Given** an operator receives a Zarf package
- **When** a package developer runs `zarf package inspect *reports* *release-notes*`
- **Then** the requested documentation is retrieved from the package and displayed

(optional) A clear and concise description of any alternative solutions or features you've considered.

AustinAbro321:
## Description

Adds the ability to store documentation under the documentation key which is a map[string]string. Stores documentation files in the package by their base filename.

I decided not to include documentation from imported packages. If a user wants documentation from a package they are importing, it should be simple enough to copy that file and import the file separately, and I believe there are situations  where documentation for an imported package, wouldn't make sense in the parent package. 

Currently this feature does not allow pulling in remote files, I'd be open to adding this in the future if someone presents a use case.

One neat thing I learned is that if a user tries to define the same key twice in a yaml map then we will error, so no additional validation was needed for that. 

## Related Issue

Fixes #4346

## Checklist before merging


examples/dos-games/zarf.yaml
```diff
@@ -4,6 +4,11 @@ metadata:
   description: Simple example to load classic DOS games into K8s in the airgap
   version: 1.2.0
 
+# The documentation key can be utilized to pass information to package consumers
```
brandtkeller, line 7: Does this addition warrant bumping the package version (and possibly publishing)?

examples/dos-games/zarf.yaml
```diff
@@ -4,6 +4,11 @@ metadata:
...
+# Commonly used to detail package configuration, or as in this instance, application specific documentation.
+documentation:
+  secrets: "secrets.md"
```
brandtkeller: We could consider making these example key/value different from one-another so that users don't think there is a strict naming requirement here.

src/pkg/lint/validate.go
```diff
@@ -334,3 +339,17 @@ func validateManifest(manifest v1alpha1.ZarfManifest) error {
 
 	return err
 }
+
+func validateDocumentation(documentation map[string]string) error {
```
brandtkeller: Reasonable stance to take - though I can imagine we will see few examples in the future. 

This stance is based on the storage mechanism in the package based on the assemble logic... Is it worth considering a broader copy strategy that prepends the key in order to establish more uniqueness?

brandtkeller (commented):
Couple passing thoughts but otherwise clean and straight-forward.

examples/dos-games/zarf.yaml
AustinAbro321: If we had a smoother workflow I think potentially, but currently it requires running a workflow to update the new version then updating our e2e tests that pull this package to use the new version, which can be cumbersome.

examples/dos-games/zarf.yaml
AustinAbro321: Good idea, updated

src/pkg/lint/validate.go
AustinAbro321: My consideration was more so for the user when they get their documents. Keeping them unique makes it so their files are the same name when they read them. Though it is still probably best to prepend the filename with the key.

AustinAbro321: Implemented this. Note that now when a user gets a file it is called <key>-<original-file-basename>. The filename feels a bit awkward, however because keys are unique there is no longer any possibility of conflicts. Ultimately, I think it's a good change.

AustinAbro321: Actually it is totally possible to have duplicates, it's just a more rare case. Added validation back and tests for this

AustinAbro321:
@brandtkeller Do you think we should limit users to only adding files, or allow them to recursively add directories of documentation?

AustinAbro321:
One change I just made that will make it easier to pull in documentation directories in the future should we go that route, and also is, in my opinion, a cleaner solution is to create a documentation.tar file in the package, rather than keeping a documentation folder in the package.

brandtkeller:
> @brandtkeller Do you think we should limit users to only adding files, or allow them to recursively add directories of documentation?

Unless we have some prior art to pull from here - I would advocate for 1:1 for key:file and then evaluate enhancement requests for directories.

AustinAbro321:
@brandtkeller Sounds good, I was torn on the decision but that was my inclination as well

examples/dos-games/zarf.yaml
```diff
@@ -4,6 +4,11 @@ metadata:
   description: Simple example to load classic DOS games into K8s in the airgap
   version: 1.2.0
 
+# The documentation key can be utilized to pass information to package consumers
+# Commonly used to detail package configuration, or as in this instance, application specific documentation.
+documentation:
+  walkthrough: "secrets.md"
```
brandtkeller, line 10: not a blocker - but there is a drift here between existing consistency for how we specify files and this implementation. The use of a map[string]string is captured in the PR description but nowhere else as far as I can tell. 

I believe we could use more supporting documentation here - but the other note was just validating that we want to continue with this approach (I believe we do). 

said another way - we typically use `files` as a list and this approach deviates from that - I believe we want documentation to describe the current format and expectations?

src/pkg/packager/layout/package.go
```diff
@@ -319,6 +319,70 @@ func (p *PackageLayout) GetSBOM(ctx context.Context, destPath string) error {
...
+		for _, key := range keys {
+			if filePath, ok := p.Pkg.Documentation[key]; ok {
+				keysToExtract[key] = filePath
+			} else {
+				return fmt.Errorf("key %s not found in package documentation", key)
+			}
+		}
+	} else {
+		keysToExtract = p.Pkg.Documentation
+	}
```
brandtkeller: nit: I think we could remove this else block in favor of the default being `p.Pkg.Documentation` and then only performing the make in the event `len(keys) > 0`

src/pkg/packager/layout/package.go
```diff
@@ -319,6 +319,70 @@ func (p *PackageLayout) GetSBOM(ctx context.Context, destPath string) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to extract documentation.tar: %w", err)
+	}
+
+	if err := os.MkdirAll(destPath, helpers.ReadWriteExecuteUser); err != nil {
+		return fmt.Errorf("failed to create output directory %s: %w", destPath, err)
+	}
+
+	for key, file := range keysToExtract {
+		docFileName := FormatDocumentFileName(key, file)
```
brandtkeller: `docFileName` is the formatted name `key-filename.ext`. This may create some confusion if the manifest is consuming a `secrets.md` file and upon retrieval creates a `walkthrough-secrets.md` file. 

Should we consider a to/from format?

src/pkg/zoci/common.go
```diff
@@ -35,6 +35,8 @@ const (
 	AllLayers LayersSelector = ""
 	//SbomLayers is the selector for SBOM layers including metadata
 	SbomLayers LayersSelector = "sbom"
+	//DocLayers is the selector for package documentation
+	DocLayers LayersSelector = "documentation"
```
brandtkeller, line 39: Happy to see this included in the implementation btw.

brandtkeller (changes requested):
Minor considerations - we also probably want to pull in the latest changes and generate the schema with the new schema generation process.

src/pkg/packager/layout/package.go
AustinAbro321: Yeah, also realized I should be using maps.Clone here

src/pkg/packager/layout/package.go
AustinAbro321: Yeah ultimately there is a decision point in this PR. Do we want to allow multiple files with the same basename? If not, I think we can get rid of the key-name format.  I'd be okay with either way, though another reason I like the key-name format is that if we added remote files in the future then <key>-remote-1 seems like it'd be a natural addition, though I suppose we could do that either way. 

What would a to/from format look like?

examples/dos-games/zarf.yaml
AustinAbro321: Good point, keys might not be intuitive for users. I added a line to the docs explaining that these are descriptive keys.

src/pkg/packager/layout/package.go
brandtkeller: I think we'd need validation on the same basename discussion - personal opinion is that we should allow for it - my only thought is that pulling multiple files with the same basename _from different directories_ is an expected behavior. IE changelogs have common naming and a package reasonably has multiple pieces of software that could need many changelogs. I don't imagine that pattern is present in many other ways. 

the formatting concepts adding that bit of awareness - on assembly we add the key to the filename for uniqueness - on retrieval we remove the key from the filename so as to match the original manifest inputs.

AustinAbro321: The problem with removing the key when extracting the documentation to the user is that if two files have the same basename, then when they're extracted to the user they can't go inside of the same <my-package-name>-documentation directory.

We could add the prefix only when there are name collisions. The reason I didn't do that was so that it was always the same and there wouldn't be surprises in the event of automation. I think it's a reasonable alternative though as matching basenames is probably relatively niche situation.

brandtkeller: Great point. It would be present in the example I previously made. I'll defer to you but I believe the best UX is consistent naming - and then we handle collisions with the prefix as needed?

src/pkg/packager/layout/layout.go
```diff
@@ -16,12 +16,16 @@ const (
 	ValuesYAML   = "values.yaml"
 	ValuesSchema = "values.schema.json"
 
-	ImagesDir     = "images"
-	ComponentsDir = "components"
+	ImagesDir        = "images"
+	ComponentsDir    = "components"
+	ValuesDir        = "values"
```
brandtkeller: `ValuesDir` is removed in main correct?

AustinAbro321: Yup, removed nice catch

src/pkg/packager/layout/package.go
AustinAbro321: Updated to handle collisions only for duplicate basenames

PR #4402, opened 2025-11-21
fix: publish breaks when Zarf package name and existing directory nam…

### LINKED ISSUE #4148: Publish breaks when Zarf package name and existing directory name collide
### Steps to reproduce
1. Create a new dir called `bar`. In `bar`, create a new directory called `foo` and a `zarf.yaml` with `metadata.name` set to `foo`:

```
kind: ZarfPackageConfig
metadata:
  name: foo
  version: 0.1.0
components:
  - name: hello-world
    required: true
    actions:
      onDeploy:
        before:
          - cmd: echo "hello world"
```
2. Run `zarf p publish ./zarf-package-foo-arm64-0.1.0.tar.zst <valid OCI repo>`

3. error: `2025-09-04 10:21:07 ERR failed to push manifest: failed to create file /tmp/foo: open /tmp/foo: is a directory`

### Expected result

`zarf p publish` should work

### Actual Result

Zarf throws the following error: `2025-09-04 10:21:07 ERR failed to push manifest: failed to create file /tmp/foo: open /tmp/foo: is a directory`

Image

ericwyles:
## Description

Fixes issue where publish fails if executed from a directory containing a subdirectory name that matches the package name. To avoid the issue this PR introduces a tmp directory for the local OCI store instead of making it in the current working directory.

## Related Issue

Fixes #4148 

## Checklist before merging

### DIFF
--- a/src/pkg/packager/publish_test.go
+++ b/src/pkg/packager/publish_test.go
@@ -311,7 +311,7 @@ func TestPublishPackageDirectoryNameCollision(t *testing.T) {
 
 			// Ensure we restore the original directory
 			t.Cleanup(func() {
-				_ = os.Chdir(origDir)
+				require.NoError(t, os.Chdir(origDir))
 			})
 
 			// Publish test package
--- a/src/pkg/zoci/push.go
+++ b/src/pkg/zoci/push.go
@@ -51,7 +51,11 @@ func (r *Remote) PushPackage(ctx context.Context, pkgLayout *layout.PackageLayou
 	if err != nil {
 		return ocispec.Descriptor{}, fmt.Errorf("failed to create temp directory: %w", err)
 	}
-	defer os.RemoveAll(tempDir)
+	defer func() {
+		if err := os.RemoveAll(tempDir); err != nil {
+			l.Warn("failed to remove temp directory", "path", tempDir, "error", err)
+		}
+	}()
 
 	src, err := file.New(tempDir)
 	if err != nil {


src/test/e2e/50_oci_publish_deploy_test.go
```diff
@@ -120,6 +123,68 @@ func (suite *PublishDeploySuiteTestSuite) Test_2_Pull_And_Deploy() {
 	suite.NoError(err, stdOut, stdErr)
 }
 
+func (suite *PublishDeploySuiteTestSuite) Test_3_Publish_With_Directory_Name_Collision() {
```
AustinAbro321: Generally we like to avoid having e2e tests cover our general user flows rather than more specific bugs. I think a good place to move this test would be as a unit test in this [src/pkg/packager/publish_test.go](https://github.com/zarf-dev/zarf/blob/main/src/pkg/packager/publish_test.go)

ericwyles: @AustinAbro321 sure yeah, I'll look at changing that up! I originally looked in src/pkg/zoci/push_test.go and it doesn't really test anything there and I missed that these were being tested through packager/publish. I avoided putting anything in zoci/push_test.go because it seemed like I would have to bootstrap a registry which was definitely not happening in there, but this makes more sense. Thank you!

ericwyles: @AustinAbro321 converted this to a unit test if you want to take another look! 

I do want to point out that the unit test uses os.Chdir(). I think that does potentially affect the whole process, but I'm not sure how else to reliably test this other than potentially creating a directory in whatever the cwd happens to be which also doesn't seem great. 

`TestValidatePackageSchema` in `src/pkg/lint/schema_test.go` also follows a similar pattern so maybe it is fine: https://github.com/ericwyles/zarf/blob/main/src/pkg/lint/schema_test.go#L197-L200

AustinAbro321: Yup I think you choose a good strategy for the test. Seems to be some conflicts on the branch. Clear that up and I will approve

ericwyles: Resolved the conflicts!

AustinAbro321 (approved):
Nice work

PR #4403, opened 2025-11-21
fix!: allow values files from any directory on Zarf package create

### LINKED ISSUE #4302: remove expectation for values files defined in a package config to be within a directory called values
Values files in package definitions currently expect values files to be in a directory called values.  Zarf should not expect users to place their values files in a directory called values, but this directory should be created in a  built Zarf package.

For example in a package with the following definition:
```yaml
kind: ZarfPackageConfig
metadata:
  name: test
values:
  files:
    - values.yaml
components:
  - name: test-component
 ```
The user's directory with their zarf.yaml would look like this
```
.
├── values.yaml
└── zarf.yaml
```
Then if we decompressed the package, it would look like this:
```
.
├── checksums.txt
├── values
│   └── values.yaml
└── zarf.yaml
```
This way the values directory is invisible to the user.

### PR DESCRIPTION
**Pull request by Racer159**
## Description

Changes the behavior around Zarf values files to merge on create instead of on deploy.

> [!CAUTION]
> **Breaking Change**
> This change is breaking for packages that use the `alpha` Zarf Values feature - these packages will need to be recreated with a new version of Zarf to be deployed by that version of Zarf. (other packages not using this feature will be unaffected)

## Related Issue

Fixes #4302

## Checklist before merging

### DIFF
--- a/src/test/packages/42-values/schema-deploy-invalid/zarf.yaml
+++ b/src/test/packages/42-values/schema-deploy-invalid/zarf.yaml
@@ -5,7 +5,7 @@ metadata:
 
 values:
   files:
-    - values.yaml
+    - values/values.yaml
   schema: values.schema.json
 
 components:
--- a/src/test/packages/42-values/schema-invalid/zarf.yaml
+++ b/src/test/packages/42-values/schema-invalid/zarf.yaml
@@ -5,7 +5,7 @@ metadata:
 
 values:
   files:
-    - values.yaml
+    - values/values.yaml
   schema: values.schema.json
 
 components:
--- a/src/test/packages/42-values/schema-valid/zarf.yaml
+++ b/src/test/packages/42-values/schema-valid/zarf.yaml
@@ -5,7 +5,7 @@ metadata:
 
 values:
   files:
-    - values.yaml
+    - values/values.yaml
   schema: values.schema.json
 
 components:

### CODE REVIEW

File: src/pkg/packager/deploy.go [hunk 1]
```diff
@@ -142,15 +142,15 @@ func Deploy(ctx context.Context, pkgLayout *layout.PackageLayout, opts DeployOpt
...
-	for i, vf := range pkgLayout.Pkg.Values.Files {
-		valueFilePaths[i] = filepath.Join(pkgLayout.DirPath(), layout.ValuesDir, vf)
-	}
-	vals, err := value.ParseFiles(ctx, valueFilePaths, value.ParseFilesOptions{})
-	if err != nil {
-		return DeployResult{}, err
+	// Read the package values from values.yaml if it exists
+	vals := make(value.Values)
+	valuesPath := filepath.Join(pkgLayout.DirPath(), layout.ValuesYAML)
+	if _, err := os.Stat(valuesPath); err == nil {
```
AustinAbro321: nit: I think it's better to return errors here besides `errors.Is(err, fs.ErrNotExist)`

src/pkg/packager/layout/assemble.go
```diff
@@ -154,11 +155,9 @@ func AssemblePackage(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath
 		}
 	}
 
-	l.Debug("copying values files to package", "files", pkg.Values.Files)
-	for _, file := range pkg.Values.Files {
-		if err = copyValuesFile(ctx, file, packagePath, buildPath); err != nil {
-			return nil, err
-		}
+	l.Debug("merging values files to package", "files", pkg.Values.Files)
+	if err = mergeAndWriteValuesFile(ctx, pkg.Values.Files, packagePath, buildPath); err != nil {
```
AustinAbro321, line 159: Mostly an implementation detail, but I'm 50/50 on whether or not we should include the values.yaml file in the package always or if we should skip writing the file when the package does not set values in the definition. Curious to hear your reasoning behind adding it to the package.

Racer159: I debated this too but decided to just add it to the package so it could be more "expected" to be a part of the package definition in the future (even if for backwards compat reasons that may never happen).  It also simplifies the writing a little since you just iterate through the array (even if it is empty) and write the map without needing additional checks.  The backward compat check on deploy would be needed regardless because of existing packages.

AustinAbro321: Yeah, I think an argument against adding the values.yaml in this situation is that it isn't in line with the current paradigm where the SBOM.tar / zarf.yaml.sig files are not included when those features are not used in the package. However, those files do not have a clear 0 value, whereas the values.yaml does and I think that makes it make sense for me.

mkcp:
Reading back through for final review this afternoon, will update soon.

src/pkg/packager/layout/assemble.go
mkcp: This is definitely more of a nit but I believe we're better off not writing the file if no values are declared, but empty declared values file(s) are written out as an empty merged file. I think we just need to exit early in mergeAndWriteValuesFile if files == 0. I added a nit for it below.

src/internal/value/value.go
```diff
@@ -96,7 +96,8 @@ func ParseFiles(ctx context.Context, paths []string, _ ParseFilesOptions) (_ Val
 	return m, nil
 }
 
-func parseLocalFile(ctx context.Context, path string) (Values, error) {
+// ParseLocalFile reads and parses a single local YAML file into a Values map.
+func ParseLocalFile(ctx context.Context, path string) (Values, error) {
 	m := make(Values)
```
mkcp, line 105: nit: public fn should guard against empty path ""

```suggestion
if strings.TrimSpace(path) == "" {
  return make(Values), nil
}
```

src/internal/value/value.go
mkcp: nit: not rly in scope of this PR, but let's be consistent with returning not returning nil Values.


```suggestion
		return make(Values), err
```

src/pkg/packager/layout/assemble.go
```diff
@@ -874,35 +873,34 @@ func createReproducibleTarballFromDir(dirPath, dirPrefix, tarballPath string, ov
 	})
 }
 
-func copyValuesFile(ctx context.Context, file, packagePath, buildPath string) error {
+func mergeAndWriteValuesFile(ctx context.Context, files []string, packagePath, buildPath string) error {
 	l := logger.From(ctx)
```
mkcp, line 885: nit: return early if no files to read. 
```suggestion
if len(files) == 0 {
	return nil
}
```

src/pkg/packager/deploy.go
```diff
@@ -142,15 +142,19 @@ func Deploy(ctx context.Context, pkgLayout *layout.PackageLayout, opts DeployOpt
...
-	if err != nil {
-		return DeployResult{}, err
+	// Read the package values from values.yaml if it exists
+	vals := make(value.Values)
+	valuesPath := filepath.Join(pkgLayout.DirPath(), layout.ValuesYAML)
+	if _, err := os.Stat(valuesPath); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return DeployResult{}, err
+		}
+	} else {
```
mkcp: nit: typically we avoid elses in favor of guard clauses and semantic empty values. Maybe we can move the os.Stat into `ParseLocalFile` and ignore os.ErrNotExist like we do on :149 above? Alternatively os.Open will return this same error, so we can just check for it below in value.ParseLocalFile's err check.

src/pkg/packager/layout/assemble.go
Racer159: Yeah I could go either way - checked helm package too and it actually omits the values.yaml so that is probably the way to go - will make the change

src/internal/value/value.go
Racer159: added (and moved some things around)

src/internal/value/value.go
Racer159: swapped everything to have make(Values)

src/pkg/packager/layout/assemble.go
Racer159: added

src/pkg/packager/deploy.go
Racer159: Moved it over to ParseLocalFile

PR #4405, opened 2025-11-23
feat!: update injector to respect arch of package 

a1994sc:
## Description

This is PR adds checks during the scheduling of the injector pod to match the architecture of the package; e.i. if you have an `amd64` zarf init package, then the injector will only be scheduled on a node of `amd64`.

A chunk of the work for this PR came from PR #3935 but I felt it was better to focus on the injector part in this PR and open a second PR for scheduling the agent/registry.

## Related Issue

Relates to #3817

## Checklist before merging


a1994sc:
This PR does also include a new workflow, provided by Austin, to test some level of multi-arch workflows.

src/pkg/cluster/injector.go
```diff
@@ -39,13 +39,23 @@ import (
 
 var zarfImageRegex = regexp.MustCompile(`(?m)^(127\.0\.0\.1|\[::1\]):`)
 
+// InjectorOpts represents the config for the injector pod
+type InjectorOpts struct {
```
brandtkeller: Layout of the structure looks good. Quick passing thought of ensuring these items are indeed optional? 

Ideally required items are function parameters and optional are in the options struct. 

Another overlay may be deciding if we can set sane defaults.

brandtkeller (commented):
Passing comment for thoughts and opinions.

src/pkg/cluster/injector.go
a1994sc: Hm, gotcha, I think that makes sense too me... I will have to look up some examples on how to enforce certain fields are populated verses optional...

Thank you Brandt! 

(P.S. If you have a more meaningful name for this struct feel free to drop it, I'm crap at naming stuff)

src/pkg/cluster/injector.go
```diff
@@ -39,28 +39,69 @@ import (
 
 var zarfImageRegex = regexp.MustCompile(`(?m)^(127\.0\.0\.1|\[::1\]):`)
 
+// ZarfInjectorOptions represents the options used by injector pod
+type ZarfInjectorOptions struct {
```
AustinAbro321, line 43: We're not 100% consistent with this in the codebase, but I think the only fields in the options struct should be optional fields. In this case the RegistryNodePort and InjectorNodePort are the optional fields

.github/workflows/nightly-eks-multi-arch.yml
AustinAbro321, line 1: This test is pretty expensive in terms of resources and maintainer time since aws tends to flake. I'd be okay with deleting this test especially since the injector is unit tested, and the code to get multi arch working will be much simpler than it was when I initially shared this code with you.

Also I believe this won't work yet, as we will need to expose `nodeSelector` as a variable in the registry, agent, and gitea packages. I think I'd prefer that being done in this PR, also okay with opening a second PR for this.

AustinAbro321 (changes requested):
Great start

src/pkg/cluster/injector.go
a1994sc: So for this specific case, are you suggesting that I leave `RegistryNodePort` and `InjectorNodePort` public, but change the rest to private and write getters/setters?

a1994sc: And/or do you have an example in the current code base you would suggest that I look at as an example?

.github/workflows/nightly-eks-multi-arch.yml
a1994sc: I'm fine with removing the test workflow, as I have mixed feels with adding workflows that I did not test ha. I trust you that it works, but still.

I can also at least expose `nodeSelector` with our charts, I think that gitea already it so I will sanity check that.

src/pkg/cluster/injector.go
AustinAbro321: Yup, If you take a look at packager.Deploy, the package layout is required, however deploy can still work with every other parameter not set, making them optional  https://github.com/zarf-dev/zarf/blob/main/src/pkg/packager/deploy.go#L98. In the case of the ZarfInjectorOptions, every field except for registryNodePort and InjectorNodePort is a required field. Those two fields should remain in the struct, while the rest of the fields should stay as parameters on the function.

a1994sc: Hmmmm, gotcha, yea I will take a look and update the PR

src/pkg/packager/deploy.go
```diff
@@ -366,14 +366,18 @@ func (d *deployer) deployInitComponent(ctx context.Context, pkgLayout *layout.Pa
 			if err != nil {
 				return nil, err
 			}
-			payloadCMs, shasum, err := d.c.CreateInjectorConfigMaps(ctx, pkgLayout.DirPath(), pkgLayout.GetImageDirPath(), component.GetImages(), pkgLayout.Pkg.Metadata.Name)
+
+			payloadCMs, shasum, err := d.c.CreateInjectorConfigMaps(ctx, pkgLayout.DirPath(), pkgLayout.GetImageDirPath(), component.Images, pkgLayout.Pkg.Metadata.Name)
```
AustinAbro321:
```suggestion
			payloadCMs, shasum, err := d.c.CreateInjectorConfigMaps(ctx, pkgLayout.DirPath(), pkgLayout.GetImageDirPath(), component.GetImages(), pkgLayout.Pkg.Metadata.Name)
```
Looks like when resolving conflicts this was moved back to Images instead of GetImages()

AustinAbro321: Same with block below

src/pkg/cluster/injector.go
```diff
@@ -548,6 +560,9 @@ func buildInjectionPod(nodeName, image string, payloadCmNames []string, shasum s
 								WithType(corev1.SeccompProfileTypeRuntimeDefault),
 						),
 				).
+				WithNodeSelector(map[string]string{
+					"kubernetes.io/arch": architecture,
+				}).
```
AustinAbro321: Given that we specify the node name already, this doesn't have any effect. I think it'd be more clear to delete this

.github/workflows/nightly-eks-multi-arch.yml
a1994sc: @AustinAbro321 I think that I will circle back to adding support for only scheduling the agent/registry on the appropriate arch in a follow up PR. I know that for now, at least making sure the inject is scheduled correctly is a big step.

AustinAbro321 (approved):
LGTM

PR #4417, opened 2025-11-25
feat: add external data injection and fix registry prune resilience

### LINKED ISSUE #2121: Allow data outside of a zarf package to be injected on deploy time (DataInjection feature)

Currently zarf packages can be created and packaged with custom data to be injected using the [DataInjection](https://docs.zarf.dev/docs/create-a-zarf-package/zarf-components#data-injections) feature. This is a already a great feature as we can package additional scripts and files to be run at deploy time.

This feature request is about extending this functionality so that on-premises data that is not known at package creation can be injected from outside the zarf package itself. This will allow users to deploy data that is only available at the on-premises / air-gapped site. For example, data that is not available for privacy/security/ownership reasons.

Since a package creator cannot pre-pack all the data needed for a zarf deployment, we should add an option to `zarf package deploy' to define the data we want to inject via a CUSTOM environment variable or similar. I imagine that this env variable points to a directory relative to the zarf binary or the zarf package file location. In the zarf package itself, this env is defined as the source of the data. Optionally, we could limit the files and directories injected using file/dir pattern matching, similar to .gitignore matching.

If possible, the injected data should be available before the main container process starts (as it is now with the existing DataInjection functionality). 

As a workaround to this solution I've recently created a zarf package which deploys a complete zarf manifest including a datainjection job to the air-gapped system. On the air-gapped system I then do a `zarf package create` in order to inject the data we want. Once we have the newly created package we deploy it using `zarf package deploy`. It works but this workaround is rather tedious to setup and maintain. 

### Additional context

This feature has already been briefly discussed on the zarf slack channel [here](https://kubernetes.slack.com/archives/C03B6BJAUJ3/p1698386142051179)

### LINKED ISSUE #2641: request: Focus prune command to only worry about pruning
I would like `zarf tools registry prune` to not error out when it cannot find a manifest for an image that it thinks should be in the registry. When I use the `zarf tools registry prune` command I only care about pruning images that I am no longer referencing out of my registry. I am not using the command to verify that all the images my packages reference are in my registry.

There are times when I am deploying a Zarf package with several large images when I realize I might be brushing up against the size limit of my container registry. Sometimes I am partially through deploying a package/component when I realize this. When I realize this, I often try to prune my registry of unnecessary images. This prune command will fail if I'm in the middle of deploying a component that has a large image. Because my images are horrifically large I want to maintain whatever layers have been pushed to the registry to save myself potential hours (yes I know..) of time re-pushing them.

I realize that there are several things I can do to make my situation better, but I think this request will make the prune command better overall.

- I could remove packages (more specifically their components) that have been partially deployed and continue re-running the prune command until I remove all offending components.
- I could push a temporary image with a matching tag to the registry so that the prune command will continue 🤮 

### Additional context

Screenshot 2024-06-20 at 11 29 15 AM
> Screenshot of an image push failing midway through (most likely because of disk pressure within the registry).

<br><br>

Screenshot 2024-06-20 at 11 30 35 AM
> Screenshot of the prune command failing because it was unable to find a manifest for the image that was not fully pushed.

ljluestc:
This PR introduces a new feature for external data injection and improves the resilience of the registry prune command.

### 1. External Data Injection (Resolves #2121)
This feature allows users to inject data from the local filesystem at **deploy time**, rather than requiring the data to be bundled into the Zarf package during creation. This is particularly useful for large datasets, sensitive data, or dynamic data that is only generated at deploy time.

**Changes:**
- Updated `ZarfDataInjection` schema to include a `type` field (`embedded` | `external`).
- Updated `zarf package create` to skip bundling sources for `external` data injections.
- Updated `zarf package deploy` to resolve external sources using Zarf variables and pass the local path to the injection logic.
- Fixed SBOM generation to handle components that may not have a tarball (if they only contain external data).

### 2. Resilient Registry Prune (Resolves #2641)
The `zarf tools registry prune` command has been updated to handle missing image manifests gracefully. Previously, if the command encountered an image digest that was expected but missing from the registry (e.g., due to a failed push), it would error out.

**Changes:**
- The command now logs a warning when an image digest cannot be found and continues processing the remaining images.

### DIFF
--- a/src/api/internal/v1beta1/component.go
+++ b/src/api/internal/v1beta1/component.go
@@ -315,6 +315,8 @@ type ZarfDataInjection struct {
 	Target ZarfContainerTarget `json:"target"`
 	// Compress the data before transmitting using gzip. Note: this requires support for tar/gzip locally and in the target image.
 	Compress bool `json:"compress,omitempty"`
+	// [alpha] The type of data injection (default 'embedded' which bundles the data at create time).
+	Type string `json:"type,omitempty" jsonschema:"enum=embedded,enum=external"`
 }
 
 // ZarfComponentImport structure for including imported Zarf components.
--- a/src/api/v1alpha1/component.go
+++ b/src/api/v1alpha1/component.go
@@ -4,6 +4,13 @@
 // Package v1alpha1 holds the definition of the v1alpha1 Zarf Package
 package v1alpha1
 
+const (
+	// DataInjectionEmbedded is the default data injection type
+	DataInjectionEmbedded = "embedded"
+	// DataInjectionExternal is for external data injection
+	DataInjectionExternal = "external"
+)
+
 // ZarfComponent is the primary functional grouping of assets to deploy by Zarf.
 type ZarfComponent struct {
 	// The name of the component.
@@ -373,6 +380,8 @@ type ZarfDataInjection struct {
 	Target ZarfContainerTarget `json:"target"`
 	// Compress the data before transmitting using gzip. Note: this requires support for tar/gzip locally and in the target image.
 	Compress bool `json:"compress,omitempty"`
+	// [alpha] The type of data injection (default 'embedded' which bundles the data at create time).
+	Type string `json:"type,omitempty" jsonschema:"enum=embedded,enum=external"`
 }
 
 // ZarfComponentImport structure for including imported Zarf components.
--- a/src/cmd/crane.go
+++ b/src/cmd/crane.go
@@ -371,7 +371,8 @@ func doPruneImagesForPackages(ctx context.Context, options []crane.Option, s *st
 
 					digest, err := crane.Digest(transformedImageNoCheck, options...)
 					if err != nil {
-						return err
+						l.Warn("unable to get digest for image, skipping prune check", "image", transformedImageNoCheck, "error", err.Error())
+						continue
 					}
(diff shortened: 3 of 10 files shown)


AustinAbro321:
Hello @ljluestc, I appreciate the work you put into this PR! We are planning on deprecating data injections soon. Some of the reasons why are in PR #4068 and we are going to formally announce the intent to deprecate after #4334 is merged as that will give users an easier migration path. 

Even without deprecating data injections, I do think there are some issues with solving this problem in this way. Two of the principals of Zarf packages are that they should be simple to deploy and deployable on any machine. These are not always true, but by adding the requirement of having a specific directory we move away from these principals. 

I do believe that  is worth a separate issue outside of the context of data injections and https://github.com/zarf-dev/zarf/issues/2121. It's a common enough problem that Zarf should solve or assist in somehow. 

One way that a user solve this now, without data injections today is by following a similar guide to what's laid out in #4068. However, instead of adding the data image to the package, a user would create a separate Zarf package in their deploy environment which solely pushes images with their data. This package would be deployed before the main package. Still not a perfect solution, your new package would have to assume that there are existing images in the registry, but I do think it's cleaner solution.

Also generally a PR should solve one item. I am closing this out, but I encourage you to create a separate PR for issue https://github.com/zarf-dev/zarf/issues/2641

Issue #4400, opened 2025-11-21
Add linter to avoid accepting internal types in public functions


AustinAbro321:
### Describe what should be investigated or refactored
packager.Create currently accepts an internal type as a public function. This breaks SDK users and stops them from calling the function. We should have a linter that stops prevents this

AustinAbro321:
Looked through the linters on https://golangci-lint.run/docs/linters/configuration/, and I could not find any that would solve this issue. Closing this out. If anyone does know of any linters please comment

PR #4424, opened 2025-12-01
feat(sdk): export images package

AustinAbro321:
## Description

Currently images.RegistryOverrides is a field on the struct packager.CreateOptions. It's impossible for SDK users to set this field. I'd like to keep the struct definition in the images package, since that's where the data is used, so this PR exports the images package. 

As a part of making the images package public, this PR changes the config structs to optional structs, and makes it so all required fields are direct parameters on the functions. 

I was hoping to find a linter that would help avoid problems like this in the future (#4400), but was unable to find any

## Checklist before merging

### DIFF
--- a/src/pkg/images/push_test.go
+++ b/src/pkg/images/push_test.go
@@ -93,11 +93,11 @@ func TestPush(t *testing.T) {
 			}
 
 			// push images to registry
-			cfg := PushOptions{
+			opts := PushOptions{
 				PlainHTTP: true,
 				Arch:      "amd64",
 			}
-			err = Push(ctx, imageList, tc.SourceDirectory, regInfo, cfg)
+			err = Push(ctx, imageList, tc.SourceDirectory, regInfo, opts)
 
 			if tc.expectErr {
 				require.Error(t, err, tc.expectErr)
--- a/src/pkg/packager/deploy.go
+++ b/src/pkg/packager/deploy.go
@@ -460,7 +460,7 @@ func (d *deployer) deployComponent(ctx context.Context, pkgLayout *layout.Packag
 			}
 			refs = append(refs, ref)
 		}
-		pushConfig := images.PushOptions{
+		pushOpts := images.PushOptions{
 			OCIConcurrency:        opts.OCIConcurrency,
 			PlainHTTP:             opts.PlainHTTP,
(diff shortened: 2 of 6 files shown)


src/pkg/packager/layout/assemble.go
```diff
@@ -118,17 +118,15 @@ func AssemblePackage(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath
 	}
 	sbomImageList := []transform.Image{}
 	if len(componentImages) > 0 {
-		pullCfg := images.PullConfig{
+		pullCfg := images.PullOptions{
```
mkcp: Nit: update var name `pullCfg`

src/pkg/packager/deploy.go
```diff
@@ -460,19 +460,16 @@ func (d *deployer) deployComponent(ctx context.Context, pkgLayout *layout.Packag
 			}
 			refs = append(refs, ref)
 		}
-		pushConfig := images.PushConfig{
+		pushConfig := images.PushOptions{
```
mkcp: Nit: update var name `pushConfig`

src/pkg/packager/mirror.go
```diff
@@ -59,19 +59,16 @@ func PushImagesToRegistry(ctx context.Context, pkgLayout *layout.PackageLayout,
 	if len(refs) == 0 {
 		return nil
 	}
-	pushConfig := images.PushConfig{
+	pushConfig := images.PushOptions{
```
mkcp: Nit: update var name `pushConfig`

src/pkg/images/push_test.go
```diff
@@ -93,14 +93,11 @@ func TestPush(t *testing.T) {
 			}
 
 			// push images to registry
-			cfg := PushConfig{
-				SourceDirectory: tc.SourceDirectory,
-				RegistryInfo:    regInfo,
-				PlainHTTP:       true,
-				Arch:            "amd64",
-				ImageList:       imageList,
+			cfg := PushOptions{
```
mkcp: Nit: update var name `cfg`

mkcp (changes requested):
A few nits from refactor then good to go

PR #4426, opened 2025-12-02
feat: move to flux cli utils

AustinAbro321:
## Description

Unsure if we want to take this direction. 

Helm 4 uses the Flux version of CLI utils. We are going to move to Helm 4 soon #4350. Rather than having two versions of the same dependency in Zarf, this would keep us with in line with similar tools as Zarf in Helm / Flux. Additionally, if there is a bug in cli-utils it's likely to be caught by one of these popular projects given that they depend on it for a critical part of their deployment. 

## Checklist before merging


AustinAbro321:
Going to close this for now, maybe still worth considering, but I think the ideal future would be if kstatus is removed from cli-utils and everyone started using that

PR #4427, opened 2025-12-03
feat: implement import logic for Zarf Values

### LINKED ISSUE #4584: Support transient package values during component import

The values feature does not include performing any additional actions on package create when a component import is specified. Ideally onyl the component values would be included on import - but there is not a native specification for component imports and many packages may have common imports for package values. 

Rather for  packages values - zarf should include all values with the component - preventing collisions where possible.

- **Given** a child package exists with components
- **When** a child component is imported into a parent package
- **Then** the values of the child are merged with the values of the parent

Do not include any values/schema from the child.

### Additional context

There is additional optionality here to either merge all values outright on create or namespace the values on some pre-determined key to help isolate collision. In either scenario a command on inspect should exist to populate the package values - preferrably in a exportable-to-yaml format that can then be used on deployments.

Racer159:
## Description

This PR implements import logic for Zarf Values.

## Related Issue

Fixes #N/A

## Checklist before merging


AustinAbro321:
Overall logic looks good, but worth noting that this PR depends on #4403 as without that change imports will not work since Zarf will look for a values directory.

site/src/content/docs/ref/components.mdx
```diff
@@ -255,10 +255,18 @@ When merging components together Zarf will adopt the following strategies depend
...
+
+Some package level fields from imported components will also be merged with the importing package.  These fields will be processed from the first component to the last component in a Zarf package definition.
+
+| Kind                       | Field(s)                 | Description |
+|----------------------------|--------------------------|-------------|
+| 'name'd Globals            | `constants`, `variables` | These fields will match on `name` building up a map as components are processed (starting with the importing package definition).  Any names that already exist will be skipped so the parent or an earlier component import will take precedence. |
+| Un'name'd Primitive Arrays | `files.values`           | The importing package definition's array will be appended to the end of the array from the imported component's package definition.  Elements will be reappended regardless of whether they already exist in the array (since merge order impacts the final values). |
+| Global Behavior            | `files.schema`           | This field will always take the value of the importing component's package definition (even if it is empty) |
```
AustinAbro321:
```suggestion
| Un'name'd Primitive Arrays | `values.files`           | The importing package definition's array will be appended to the end of the array from the imported component's package definition.  Elements will be reappended regardless of whether they already exist in the array (since merge order impacts the final values). |
| Global Behavior            | `values.schema`           | This field will always keep the value of the importing component's package definition (even if it is empty) |
```

AustinAbro321: I think there is an important decision point here around merging or not merging schemas. It seems like the [allOf](https://json-schema.org/understanding-json-schema/reference/combining#introduction) key could be used here to manage this. 

There are a few non-ideal aspects of this strategy:
- There wouldn't be precedence, the parent wouldn't take priority.
- The value would have to be valid against every schema. This could lead to values that are impossible to define.

AnyOf, won't work since if any schemas don't define a certain property then it will pass. 

Still AllOf could be the way to go. It's a more intuitive experience. Curious if you considered schema merging and ran into any hard blockers

AustinAbro321 (changes requested):

site/src/content/docs/ref/components.mdx
Racer159: I did briefly consider schema merging - my thoughts though were to basically follow a similar tact to how values files are merged where parent keys would override child keys within the JSON.

This would need to be handled specific to the jsonschema format but for example:

`$defs`, `patternProperties` and `properties` would replace by key so if a parent redefined `Constant` here it would replace the whole definition under that key: https://github.com/zarf-dev/zarf/blob/092b72117457b4777d3a879831dcd39cfde75af0/zarf.schema.json#L3 but if it was a new key it would just be merged into the map.

Strings in the schema (like `$id`, `additionalProperties`, or `description`) could just take whatever is in the parent.

There could be some other things to think about in there too (like what happens if someone mixed jsonschema versions) but that shouldn't happen since there is really only one non-draft currently published and folks should just use `2020-12` - Zarf maybe could also just hard-require a specific version.

src/pkg/packager/load/testdata/import/variables/zarf.yaml
```diff
@@ -7,11 +7,17 @@ constants:
 variables:
   - name: PARENT_VAR
     default: "default from parent"
+values:
+  files:
+    - parent-values.yaml
+  schema: "parent-values.schema.json"
```
brandtkeller: I don't yet see these files existing for testing purposes. Also should we perhaps instrument the testing separate from existing variable tests?

brandtkeller (changes requested):
Testdata looks incomplete - or I am missing context.

src/pkg/packager/load/testdata/import/variables/zarf.yaml
brandtkeller: considering scope of the unit testing - this seems intentional but maybe less than ideal.

site/src/content/docs/ref/components.mdx
brandtkeller: considering the component-to-package relationship - have we considered namespacing values and nesting the schema?

There is a lot to consider here - so if it has been covered then the current discussion should continue. Otherwise I am curious if it could be a valid alternative to isolate merge collision issues.

src/pkg/packager/load/import.go
AustinAbro321, line 1: I believe we'll also want sourcePath/targetPath to work with imports with the parents sourcePath/targetPath taking priority

site/src/content/docs/ref/components.mdx
brandtkeller: Proposal PR for review: https://github.com/zarf-dev/proposals/pull/55

brandtkeller: Namespacing introduced more complexity than we want to pursue so I am not recommending that strategy.

brandtkeller:
@Racer159 Can you rebase with main to bring this up-to-date? I want to compare available options but I intend to evaluate any edgecases and otherwise advocate for getting this work merged as-is or similar and then pushing progress thereafter for values maturation.

src/pkg/packager/load/import.go
brandtkeller: This should be handled now - implemented in https://github.com/zarf-dev/zarf/pull/4581

brandtkeller:
Updated tests to isolate testing values import and merge behaviors. 

I think we should consider UX next:
- `zarf dev inspect` ignores imports entirely
- `zarf package inspect` lists the files as they would exist on import instead of the aggregate `values.yaml` that now lives within the package. Should we mutate this for accuracy around the built package or keep it for backwards transparency?

I think any schema decisions should be part of a separate PR. The docs include mention of the behavior now and we can decide if we want to pursue schema import behaviors.

brandtkeller:
Adding a thought - Looking at package values at the top level you specify values.files[] and this is a list of paths to package values files for the package. Imports will add items to this list. 

Once packaged - this list is slightly less representative - we parse all values files and consolidate into a single values.yaml file in the package. 

Options that stand out immediately:
Retain them as they are defined or resolved - zarf package inspect will include the declarative definition but the underlying package will only have a values.yaml
mutate the definition so that zarf package create specifies the packaged file values.yaml and not the original definition. 

1 feels more correct while 2 feels most accurate (regarding a built package). I do wonder if we should revisit the original design.... I think there is a third route where we don't consolidate internally to a values.yaml that could be most representative of both scenarios.

AustinAbro321:
I would stay with option 1 after considering the most likely workflows of users. 

Users are unlikely to unarchive a package to figure out what is in the values.yaml, and are more likely to run `zarf package inspect manifests|values-files`. If they get something unexpected, they may run `zarf package inspect definition` to see which files were included. By removing what the creator wrote, we make it more difficult to view what was used at create time. 

Similarly, I would rather compile the files at create time. It's simpler to implement, and while keeping them separate might give higher fidelity to someone viewing an unarchived package, I wouldn't necessarily want to encourage this, and would rather introduce inspect commands if there is that need.

src/pkg/packager/load/testdata/import/values/basic/zarf.yaml
AustinAbro321, line 1: We can delete this test, it's exactly the same as schema parent wins

src/pkg/packager/load/import.go
AustinAbro321, line 1: Values files are not implemented with skeleton packages, if you try to create a skeleton package you get something like below

```bash
../../build/zarf package create  --features="values=true"
2026-05-01 16:48:47 INF assembling package path=.
2026-05-01 16:48:47 INF fetching info for images count=1 destination=/tmp/zarf-1133454106/images
2026-05-01 16:48:47 INF pulling images count=1
2026-05-01 16:48:47 INF saving image name=docker.io/library/nginx:1.29.2 size=59.76 MBs
2026-05-01 16:48:48 INF done pulling images count=1 duration=1.5s
2026-05-01 16:48:48 INF composed components successfully
2026-05-01 16:48:48 INF generating SBOM
2026-05-01 16:48:48 INF creating image SBOM reference=docker.io/library/nginx:1.29.2
2026-05-01 16:48:51 ERR failed to create package: unable to access values file ../../../../.zarf-cache/oci/dirs/7bfd933bf94a51a6f7f1495990f04bd8ec220a0d0251f51c2ef4e4c2cbfc5e14/values/values.yaml: stat ../../../../.zarf-cache/oci/dirs/7bfd933bf94a51a6f7f1495990f04bd8ec220a0d0251f51c2ef4e4c2cbfc5e14/values/values.yaml: no such file or directory
```

src/pkg/packager/load/import.go
```diff
@@ -182,8 +183,20 @@ func resolveImports(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath,
 		components = append(components, composed)
 		variables = append(variables, importedPkg.Variables...)
 		constants = append(constants, importedPkg.Constants...)
+		for _, v := range importedPkg.Values.Files {
+			valuesFiles = append(valuesFiles, makePathRelativeTo(v, importPath))
+		}
 	}
 
+	valuesFiles = append(valuesFiles, pkg.Values.Files...)
```
AustinAbro321, line 196: We append the parent values files without cleaning them which could lead to a scenario such as "./foo.yaml" and "foo.yaml" colliding

site/src/content/docs/ref/components.mdx
```diff
@@ -277,10 +277,18 @@ When merging components together Zarf will adopt the following strategies depend
...
+
+Some package level fields from imported components will also be merged with the importing package.  These fields will be processed from the first component to the last component in a Zarf package definition.
+
+| Kind                       | Field(s)                 | Description |
+|----------------------------|--------------------------|-------------|
+| 'name'd Globals            | `constants`, `variables` | These fields will match on `name` building up a map as components are processed (starting with the importing package definition).  Any names that already exist will be skipped so the parent or an earlier component import will take precedence. |
+| Un'name'd Primitive Arrays | `values.files`           | The importing package definition's array will be appended to the end of the array from the imported component's package definition.  Duplicate paths are collapsed, with the first occurrence retained so the earliest position in the merge order is preserved. |
+| Global Behavior            | `values.schema`           | This field will always keep the value of the importing component's package definition (even if it is empty) |
```
AustinAbro321, line 291: Should we hold off on documenting this until we're sure this is the direction that we will take? I'd like to see us merge values schemas.

AustinAbro321 (changes requested):

src/pkg/packager/load/testdata/import/values/basic/zarf.yaml
brandtkeller: Fair point - these were once different but now don't instrument any difference.

site/src/content/docs/ref/components.mdx
brandtkeller: I'm good either way. It captures point-in-time accuracy and can be updated as the code changes. Depends on whether our philosophy for documentation is accuracy or consistenct-over-time.

src/pkg/packager/load/import.go
brandtkeller: Agreed - we handled standard packages and imports from a filesystem perspective. Skeletons is a significant gap for the values feature. I'm going to start drafting skeleton support but if it becomes cumbersome will split into another issue/PR.

site/src/content/docs/ref/components.mdx
brandtkeller: As a point of context - this is going to be a good challenge to potentially solve. A couple options available but merging schemas turns the import lifecycle into a process that now has to store or hold intermediary state.

Today resolving imports is updating the in-memory ZarfPackage to hand to package assembly. Because schema is a single string (not a list) merging them now requires storing each or storing the merged state somehow. 

This will be interesting to solve and worth some isolation IMO.

src/pkg/packager/load/import.go
AustinAbro321: Sounds good, in that case lets add an explicit error "importing a skeleton with values is not yet implemented" to this PR, so users who come across this situation don't file a bug.

src/pkg/packager/load/import.go
brandtkeller: Good catch! My uniqueness block wasn't effectively grabbing this line. Should now be resolved.

site/src/content/docs/ref/components.mdx
AustinAbro321: Yeah definitely a separate PR to implement this. I'm fine with leaving as is for now.

src/pkg/packager/load/import.go
brandtkeller: Updated for publishing/importing skeletons

AustinAbro321 (approved):
LGTM, glad to see this in

brandtkeller (approved):

PR #4432, opened 2025-12-03
fix(archive): clear attributes and sort files on archive creation

### LINKED ISSUE #4431: Successive pulls of a zarf package from a registry have different hashes
### Environment
Device and OS: Mac and Linux
App version: 0.66.0

### Steps to reproduce
1. Pull package from registry
2. sha sum package.tar.zst
3. pull same package from registry again
4. sha sum package.tar.zst

### Expected result

sha sums are the same

### Actual Result

sha sums are different

brandtkeller:
## Description

This PR fixes non-deterministic archive checksums when pulling OCI packages. 

Two items in particular contributed to fixing the problem:
1. `ClearAttributes: true` calls to strip filesystem timestamps, UIDs, GIDs, and other non-deterministic metadata during compression.
2.  (main issue) when OCI layers are pulled concurrently, files are written to disk in random order. The code creates a map from file paths to archive names, but Go's map iteration order is non-deterministic. When archives.FilesFromDisk() iterates over this map, it processes files in different orders across runs, resulting in tar archives with different entry sequences.

## Related Issue

Fixes #4431

## Checklist before merging

### DIFF
--- a/src/pkg/archive/archive.go
+++ b/src/pkg/archive/archive.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/mholt/archives"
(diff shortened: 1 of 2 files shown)


src/pkg/archive/archive.go
```diff
@@ -133,11 +134,25 @@ func Compress(ctx context.Context, sources []string, dest string, _ CompressOpts
...
+		ClearAttributes: true,
+	}, mapping)
 	if err != nil {
 		return fmt.Errorf("failed to stat sources: %w", err)
 	}
 
+	// Sort files by NameInArchive to ensure deterministic tar creation
+	// FilesFromDisk iterates over a map which has non-deterministic ordering
+	slices.SortFunc(files, func(a, b archives.FileInfo) int {
+		if a.NameInArchive < b.NameInArchive {
```
mkcp, line 147: The only edge case I could think of here is what if there are duplicated filenames (which would repeat the original non-deterministic order problem). That's disallowed by nearly every fs (and should be upper/lowercase unique), though tars can have multiple versions of the same file even though they overwrite to the latest on extract. I'm fairly sure this won't cause issues in OCI pulls though, just wanted to raise it as a possibility.

brandtkeller: Given the [preceeding map](https://github.com/zarf-dev/zarf/blob/main/src/pkg/archive/archive.go#L132-L135)  I don't believe we could have duplicates - but that isn't to say there isn't existing behavior potentially here that should/could handle that earlier in the function.

AustinAbro321 (approved):

PR #4434, opened 2025-12-04
feat: allow custom security context

### LINKED ISSUE #4429: Allow customization of the security context for the zarf-docker-registry
When trying to run `zarf init` in a restricted environment like OpenShift the registry pod will not schedule due to the hardcoded user and group IDs defined in the security context [here](https://github.com/zarf-dev/zarf/blob/main/packages/zarf-registry/chart/templates/deployment.yaml#L35)

The user can override the default user and group IDs in the security context via Zarf env vars

No alternatives as the values are not configurable

### Additional context
Happy to provide this as a pull request if deemed a feature worth having.

robertmarsal:
## Description

Adds 3 new registry variables, with defaults, to allow users to control the security context definition

## Related Issue

Fixes #4429

## Checklist before merging

### DIFF
--- a/packages/zarf-registry/zarf.yaml
+++ b/packages/zarf-registry/zarf.yaml
@@ -106,6 +106,18 @@ variables:
     description: Decides if host port or host network should be used with the registry proxy
     default: "false"
 
+  - name: REGISTRY_SECURITY_CONTEXT_RUN_AS_USER
+    description: The user ID to run the registry as
+    default: "1000"
+
+  - name: REGISTRY_SECURITY_CONTEXT_FS_GROUP
+    description: The filesystem group ID to run the registry as
+    default: "2000"
+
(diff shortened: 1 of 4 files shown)


packages/zarf-registry/zarf.yaml
```diff
@@ -106,6 +106,18 @@ variables:
     description: Decides if host port or host network should be used with the registry proxy
     default: "false"
 
+  - name: REGISTRY_SECURITY_CONTEXT_RUN_AS_USER
```
brandtkeller, line 109: nit: it would be good to establish a `pattern` on these variables for input validation - but the standard doesn't exist in our use of variables here today so I am not blocking on it (Can be future work to evaluate).

brandtkeller (approved):
Meets parity for backwards compatibility - appreciate the PR! LGTM.

PR #4435, opened 2025-12-04
chore(dependencies): update k3s version to 1.34.3

AustinAbro321:
## Description

Updates k3s to 1.34.3 which avoids the issue in #4443. 

This PR also updates the error message when sha's don't match so that it's include the source as it was hard to always tell which file the error message was referencing in the upstream error message since the files are moved to a temporary directory. 

Additionally, it makes it easier to see what the status is when health checks are failing

## Checklist before merging


brandtkeller (approved):
Deviation from the `debug-cluster` action makes sense in this instance given privilege requirements - lgtm.

PR #4441, opened 2025-12-05
fix(values): write and read values.schema.json from a fixed location

mkcp:
## Description
Just before merging #4252, I split the behavior for copying a schema file into its own function, and did so without ensuring it gets written to a fixed destination in the package. This PR ensure that we can copy a declared values.schema.json from anywhere, but that it always gets written to the same destination. This fixes some potential issues with file collisions during the assemble step, and will simplify package imports so that parent and child schemas can be merged and read from a deterministic location.

## Related Issue

Fixes https://github.com/zarf-dev/zarf/issues/4440

## Checklist before merging

### DIFF
--- a/src/pkg/packager/layout/assemble.go
+++ b/src/pkg/packager/layout/assemble.go
@@ -935,12 +935,6 @@ func copyValuesSchema(ctx context.Context, schema, packagePath, buildPath string
 		return fmt.Errorf("values schema validation failed: %w", err)
 	}
 
-	// Ensure relative paths don't escape the package root
-	cleanSchema := filepath.Clean(schema)
-	if strings.HasPrefix(cleanSchema, "..") {
-		return fmt.Errorf("values schema path %s escapes package root", schema)
-	}
-
 	// Copy schema file to package root
 	schemaDst := filepath.Join(buildPath, ValuesSchema)
 	l.Debug("copying values schema file", "src", schemaSrc, "dst", schemaDst)


src/internal/value/value.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/zarf-dev/zarf/src/pkg/logger"
 )
 
+const SchemaFileName = "values.schema.json"
```
brandtkeller: Should this be defined in layout such that SDK consumers have a consistent experience for how to target the schema file (should it exist)?

brandtkeller (changes requested):

src/internal/value/value.go
AustinAbro321: I would second it being placed in the layout package. The value package shouldn't need to know what a Zarf package is

mkcp: Yep! No need for `value` to know about assembled packages either. Updated

AustinAbro321: Could you delete the constant here

mkcp: oh that's weird, must've gotten undone in my editor. fixed

src/pkg/packager/layout/assemble.go
```diff
@@ -942,7 +942,7 @@ func copyValuesSchema(ctx context.Context, schema, packagePath, buildPath string
 	}
 
 	// Copy schema file to package root
-	schemaDst := filepath.Join(buildPath, cleanSchema)
+	schemaDst := filepath.Join(buildPath, ValuesSchema)
```
AustinAbro321, line 939: Now that we're changing this to not use cleanSchema, I think we should get rid of the block above this which verifies that the schema is not in a parent directory.

mkcp: Yeah that's fair, it doesn't really matter where users choose to store and reference their build values now that there's fixed destinations on assemble.

PR #4453, opened 2025-12-08
fix(tools,wait): add an http client with a timeout slightly shorter than waitInterval to wait.go

### LINKED ISSUE #4452: zarf tools wait-for http client does not have a configured
### Environment
First observed in github actions runners

### Steps to reproduce
1. Run `zarf tools wait-for` against an http server that does not respond.

### Expected result
Zarf will retry according to the configured waitInterval 

### Actual Result
Zarf hangs indefinitely until the process is cancelled or reaped.

Image
Timeout of 300s appears to be enforced by the containing runner. Upon testing this locally, wait-for will hang indefinitely on a server that accepts connections but isn't available to reply to the client.

### Severity/Priority
Moderate

### Additional Context
This is easily fixed by configuring a go http client with a timeout rather than using the default client without a timeout.

mkcp:
## Description
A user submitted this bug that appeared in CI. Initially it seemed that wait-for was only checking service health once, but looking through the code, it appears that the request will not timeout until the process is cancelled or reaped in some way. This PR adds an http client timeout similar to the waitInterval rather than prescribing a default.

## Related Issue
Fixes #4452 

## Checklist before merging

### DIFF
--- a/src/pkg/utils/wait.go
+++ b/src/pkg/utils/wait.go
@@ -208,7 +208,10 @@ func waitForNetworkEndpoint(ctx context.Context, resource, name, condition strin
 						l.Debug(err.Error())
 						continue
 					}
-					resp.Body.Close()
+					err = resp.Body.Close()
+					if err != nil {
+						l.Debug(err.Error())
+					}
 
 					// If the status code is not in the 2xx range, try again.
 					if resp.StatusCode < 200 || resp.StatusCode > 299 {
@@ -235,7 +238,10 @@ func waitForNetworkEndpoint(ctx context.Context, resource, name, condition strin
 					l.Debug(err.Error())
 					continue
 				}
-				resp.Body.Close()
+				err = resp.Body.Close()
+				if err != nil {
+					l.Debug(err.Error())
+				}
 
 				if resp.StatusCode != code {
 					l.Debug("did not receive expected status code", "expected", code, "actual", resp.StatusCode)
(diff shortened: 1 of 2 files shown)


src/pkg/utils/wait.go
```diff
@@ -197,11 +203,12 @@ func waitForNetworkEndpoint(ctx context.Context, resource, name, condition strin
 				// Default to checking for a 2xx response.
 				if condition == "success" {
 					// Try to get the URL and check the status code.
-					resp, err := http.Get(url)
+					resp, err := httpClient.Get(url)
 					if err != nil {
 						l.Debug(err.Error())
 						continue
 					}
+					resp.Body.Close()
```
mkcp: Defer in loops _can_ but don't necessarily always leak resources. In order to avoid stacking up a bunch of open request bodies until the end of the loop, I opted to close the response body immediately. This is to avoid degenerate cases where a short waitInterval combined with a long timeout keep a bunch of unnecessary memory allocated. How much memory I'm not sure.

src/pkg/utils/wait.go
```diff
@@ -197,11 +203,15 @@ func waitForNetworkEndpoint(ctx context.Context, resource, name, condition strin
 				// Default to checking for a 2xx response.
 				if condition == "success" {
 					// Try to get the URL and check the status code.
-					resp, err := http.Get(url)
+					resp, err := httpClient.Get(url)
```
brandtkeller, line 206: nit: Non-blocking - the only other outlier I see is context. I think respecting context would augment this strategy well here.

mkcp: This is a good idea that we should do but will change enough in how this code is arranged that I'd want to scope it out as a follow up. I'll write up an issue for it.

mkcp: https://github.com/zarf-dev/zarf/issues/4458

PR #4457, opened 2025-12-09
feat!(verify): deprecate --skip-signature-validation in favor of --verify

### LINKED ISSUE #4048: Make signature validation logic opt-in

As a Zarf package creator, adding a signature to my package will be a breaking change since end users must supply my package's public key when they deploy (otherwise the deploy will fail).

As an example of this behavior:
```console
❯ zarf p pull oci://ghcr.io/zarf-dev/packages/dos-games:1.2.0
2025-08-05 17:14:53 INF using config file location=/Users/mjnagel/defenseunicorns/uds-core/zarf-config.yaml
2025-08-05 17:14:55 INF pulling package name=ghcr.io/zarf-dev/packages/dos-games:1.2.0 size=3.75 MBs
2025-08-05 17:14:55 INF finished pulling package layers duration=100ms
2025-08-05 17:14:55 ERR package is signed but no key was provided
```

This validation can be skipped with the `--skip-signature-validation` flag, however the default behavior is to validate signatures. The opt-out nature of signature validation actually introduces more friction in my opinion and discourages adoption of package signing. As a case in point, the zarf init package is not currently signed, maybe partially due to this flow - it makes it hard to `zarf init` if a public key is always required. 

When a user deploys a signed zarf package without supplying the public key, then:
- The user is prompted to confirm they want to continue (if this is an interactive deployment)
- The user is presented with a warning that the package is signed but a public key was not provided

Signature validation will only be performed/fail deployment when a `--signing-key` is specified.

Better distribution/automation around pulling public keys at deploy time. If there were a better system around distributing public keys that might present an alternative option here. Essentially if the end user did not have to supply a public key on the deploy command, and Zarf were aware of the public key location - this would remove the same obstacle, while maintaining signature validation. The biggest question here is how to decide what public key to trust if the user has not explicitly provided a reference to one - which makes this feel a bit shaky.

Continuing to require public keys for signed packages, but documenting/providing conventions on where/how to store public keys to make it easier for end users to find them (log hints). This doesn't really solve the main painpoint, but could be helpful regardless.

### Additional context

Thinking about other "language"/"package" ecosystems - specifically where anyone can create and distribute a package - I am not aware of one that requires signature validation at "deploy"/usage time (it is typically optional):
- https://helm.sh/docs/topics/provenance/#overview
- https://docs.sigstore.dev/cosign/verifying/verify/
- https://docs.npmjs.com/verifying-registry-signatures

These were also interesting reads on the Python ecosystem: 
- https://caremad.io/posts/2013/07/packaging-signing-not-holy-grail/
- https://peps.python.org/pep-0480/

brandtkeller:
## Description

This change modifies the stance zarf takes when performing package verification. 

Verification will be attempted on any operations that loads a package - but only enforced when `--verify` or the `Verify` option is specified.

> [!WARNING]  
> **CLI Breaking Change**: `--skip-signature-validation` has been deprecated. Users can use `--verify` or their configuration files `package.verify: true` to establish parity with previous CLI execution.
> Additionally verifying an unsigned package now results in an error. 
>
> **SDK Breaking Change**: `LoadOptions|PackageLayoutOptions.SkipSignatureValidation` has been removed. Use `Verify` to enforce verification in LoadOptions or `VerificationStrategy` in `PackageLayoutOptions` to configure verification strategies.

Ideally we retain the CLI flag so as to not intentionally break automated workflows for end-users. We are intentionally making a breaking change to the SDK due to the implications of this change to the default stance and the ability that the change could impact SDK consumers without them being aware if no action were taken to review the updates.

Additionally #4481  was created to track removal of `PublicKeyPath` from options and instead using `VerifyBlobOptions` which wraps verification options more broadly. 

## Related Issue

Fixes #4048

## Checklist before merging

### DIFF
--- a/src/pkg/packager/layout/package.go
+++ b/src/pkg/packager/layout/package.go
@@ -36,8 +36,6 @@ type PackageLayout struct {
 // PackageLayoutOptions are the options used when loading a package.
 type PackageLayoutOptions struct {
 	PublicKeyPath string
-	// Deprecated: SkipSignatureValidation is no longer used and will be removed in a future version.
-	SkipSignatureValidation bool
 	// VerificationStrategy specifies whether verification is enforced
 	VerificationStrategy VerificationStrategy
 	IsPartial            bool
--- a/src/pkg/packager/load.go
+++ b/src/pkg/packager/load.go
@@ -32,11 +32,9 @@ type LoadOptions struct {
 	Shasum        string
 	Architecture  string
 	PublicKeyPath string
-	// Deprecated: SkipSignatureValidation is no longer used and will be removed in a future version.
-	SkipSignatureValidation bool
-	Verify                  bool
-	Filter                  filters.ComponentFilterStrategy
-	Output                  string
+	Verify        bool
+	Filter        filters.ComponentFilterStrategy
+	Output        string
 	// number of layers to pull in parallel
 	OCIConcurrency int
 	// Layers to pull during OCI pull


src/cmd/initialize.go
```diff
@@ -153,11 +156,11 @@ func (o *initOptions) run(cmd *cobra.Command, _ []string) error {
 	}
 
 	loadOpt := packager.LoadOptions{
-		PublicKeyPath:           o.publicKeyPath,
-		SkipSignatureValidation: o.skipSignatureValidation,
```
mkcp, line 157: For the places we've removed opts.SkipSignatureValidation, thoughts on keeping the field but providing a Warn or an Error that it's being removed soon and to use opts.Verify instead?

brandtkeller: I wrestled with that one... error would impact anyone who is currently using that flag/field - but a warn could be feasible. We don't have much for prior art (that I am aware of) for deprecation in the SDK. 

i'd be open to retaining it if agreed upon as a pattern we want to support (not that we'll have much choice post v1 for any other patterns).

mkcp: I'm definitely in favor of warning in this case, and doing so in future cases. One of the strengths of the options structs is that we can keep the deprecated option field around as long as is needed. Though I'd say removing it around the same time around the CLI flag.

AustinAbro321: I would be against warning and would rather see the option deleted. Imaging myself as an SDK user, if there was not a hard breaking change at compile time, it would be difficult for me to know that the option I'm using is no longer functional. I feel that a likely scenario would be that I release my product without noticing a change, users report that they're seeing the warning, but not the expected behavior. I then have to re-release to start using `verify` over `skipSignatureValidation`. 

Given that there is a clear path to equivalent behavior, I would rather have a clear signal to know what I must update.

brandtkeller: Exact issue I am reviewing currently - A warning is going to surface logs to a user in most scenarios. Feedback loops here would not be wonderful.

Trying to think with a "post-v1" hat on for how we will address scenarios such as this.

src/pkg/packager/layout/package.go
```diff
@@ -91,14 +93,41 @@ func LoadFromDir(ctx context.Context, dirPath string, opts PackageLayoutOptions)
...
+	verifyOptions := utils.DefaultVerifyBlobOptions()
+	verifyOptions.KeyRef = opts.PublicKeyPath
+
+	err = pkgLayout.VerifyPackageSignature(ctx, verifyOptions)
+	if err != nil {
+		if !opts.Verify {
+			l.Warn("package signature could not be verified:", "error", err.Error())
+			return pkgLayout, nil
 		}
+		return nil, fmt.Errorf("signature verification failed: %w", err)
```
AustinAbro321:
```suggestion
		if !opts.Verify {
			l.Warn("package signature could not be verified:", "error", err.Error())
		} else {
				return nil, fmt.Errorf("signature verification failed: %w", err)
		}
```
nit: rather than earlier return here, I believe conditionally returning and continuing on here would make for more maintainable code. It would be very easy to add a block underneath this call, and during review without the diff of the above code with the early return, it'd be hard to track that there are situations where we return early.

src/cmd/initialize.go
```diff
@@ -103,8 +104,10 @@ func newInitCommand() *cobra.Command {
 
 	cmd.Flags().IntVar(&o.retries, "retries", v.GetInt(VPkgRetries), lang.CmdPackageFlagRetries)
 	cmd.Flags().StringVarP(&o.publicKeyPath, "key", "k", v.GetString(VPkgPublicKey), lang.CmdPackageFlagFlagPublicKey)
-	cmd.Flags().BoolVar(&o.skipSignatureValidation, "skip-signature-validation", false, lang.CmdPackageFlagSkipSignatureValidation)
+	cmd.Flags().BoolVar(&o.verify, "verify", v.GetBool(VPkgVerify), lang.CmdPackageFlagVerify)
 	cmd.Flags().IntVar(&o.ociConcurrency, "oci-concurrency", v.GetInt(VPkgOCIConcurrency), lang.CmdPackageFlagConcurrency)
+	cmd.Flags().BoolVar(&o.skipSignatureValidation, "skip-signature-validation", false, lang.CmdPackageFlagSkipSignatureValidation)
+	_ = cmd.Flags().MarkDeprecated("skip-signature-validation", "Signature validation now occurs on every execution. Use --verify to enforce validation. This flag will be removed in Zarf v1.0.0.")
```
AustinAbro321: This message feels a bit misleading. Signature validation currently happen on every execution already. Maybe something like this. Open to other wording
```
Signature validation is attempted on every execution, but validation is not enforced by default. To enforce validation use the `--verify` flag. 
```

src/cmd/initialize.go
```diff
@@ -103,8 +104,10 @@ func newInitCommand() *cobra.Command {
 
 	cmd.Flags().IntVar(&o.retries, "retries", v.GetInt(VPkgRetries), lang.CmdPackageFlagRetries)
 	cmd.Flags().StringVarP(&o.publicKeyPath, "key", "k", v.GetString(VPkgPublicKey), lang.CmdPackageFlagFlagPublicKey)
-	cmd.Flags().BoolVar(&o.skipSignatureValidation, "skip-signature-validation", false, lang.CmdPackageFlagSkipSignatureValidation)
+	cmd.Flags().BoolVar(&o.verify, "verify", v.GetBool(VPkgVerify), lang.CmdPackageFlagVerify)
```
AustinAbro321, line 108: Glad to see this in here, one thing we could highlight in release notes, is that if the users want the old behavior of defaulting to enforced validation, then they should set this in their zarf config file

src/cmd/initialize.go
AustinAbro321: Thoughts on removing the flag after some time rather than at v1 https://github.com/zarf-dev/zarf/issues/3420. I think this is a strategy we could use post v1 as well. A year was what I went with before, but it honestly feels like it's too long when there is a 1:1 alternative. IMO 6 months would have been plenty.

src/pkg/packager/layout/package.go
```diff
@@ -91,14 +93,41 @@ func LoadFromDir(ctx context.Context, dirPath string, opts PackageLayoutOptions)
 		return nil, err
 	}
 
-	if pkgLayout.IsSigned() && !opts.SkipSignatureValidation {
-		verifyOptions := utils.DefaultVerifyBlobOptions()
-		verifyOptions.KeyRef = opts.PublicKeyPath
+	// Handle unsigned packages
+	if !pkgLayout.IsSigned() {
```
AustinAbro321: Nit: there are some early returns from signing in the general function `LoadFromDir()`. This doesn't cause any issues because signing is the last thing done in this function, however it would be easy to add more to this function without noticing the early returns in the future. To lower complexity, thoughts on adding another private function that encapsulates all the logic related to signing, that is called here.

src/cmd/initialize.go
brandtkeller: I don't feel strongly either way but am open. The balance here is post-v1 we may have to re-evaluate the timeline-driven language on a case-by-case basis. A breaking change has larger implications in a post-v1 world. 

But while we're v0.x.x then a 6 month timeline may very well be before v1. No one is going to stress about the timeline if they don't match.

src/cmd/initialize.go
brandtkeller: We don't have much for SDK documentation today - but I am leaning towards a middle ground here. Retaining the field - marking it as [deprecated](https://go.dev/doc/comment#deprecations) - and evaluating no-op (as the field is the default now).

brandtkeller: https://github.com/zarf-dev/zarf/issues/4482 - to track deprecation recommendations to downstream consumers.

src/pkg/packager/layout/package.go
brandtkeller: I moved validation to the source of truth in `VerifyPackageSignature` and now `LoadFromDir()` only has to orchestrate the baseline call and verify logic handling.

src/cmd/initialize.go
brandtkeller: Deprecated tag added to the structs. Feel free to unresolve if we need a different behavior.

src/cmd/initialize.go
brandtkeller: Added to the description - we will do our best to get it included into the release notes

src/pkg/packager/layout/package.go
```diff
@@ -35,12 +35,30 @@ type PackageLayout struct {
...
+	VerificationStrategy VerificationStrategy
+	IsPartial            bool
+	Filter               filters.ComponentFilterStrategy
+	VerifyBlobOptions    utils.VerifyBlobOptions
 }
 
+// VerificationStrategy describes a strategy for determining whether to verify a package.
+type VerificationStrategy int
+
+const (
```
AustinAbro321, line 49: I believe we don't want to encourage SDK users running VerifyNever, given that this is currently the default. I believe we introduced this field for packages that do not yet have the ability to be signed, such as a package during assemble. Thoughts on introducing a new public function `LoadFromDirNoSigVerification`. `LoadFromDir` could call this function. This would avoid a breaking change, while keeping the verify API consistent between functions

src/pkg/packager/load.go
```diff
@@ -29,10 +29,12 @@ import (
 
 // LoadOptions are the options for LoadPackage.
 type LoadOptions struct {
-	Shasum                  string
-	Architecture            string
-	PublicKeyPath           string
+	Shasum        string
+	Architecture  string
+	PublicKeyPath string
+	// Deprecated: SkipSignatureValidation is no longer used and will be removed in a future version.
```
AustinAbro321: I'm having some trouble seeing the advantages of the soft removal strategy of skip signature verification. Looking at how a consumer might use this such as uds-cli, they have it [explicitly set to false](https://github.com/defenseunicorns/uds-cli/blob/29a01c8fe3d71d32ea39b51677bc52129726c08d/src/pkg/bundle/inspect.go#L213). As an SDK user, I would rather have a breaking change so that I have a signal that the security posture I've opted-into is changing to be less strict. 

I believe deprecations can be very useful, but a deprecated field should still be functional. Deprecating in this way still makes it a breaking change, but leaves consumers with a much smaller signal that a breaking change occurred. 

Keeping the field functional is harder given that we are using booleans and it's hard to know when it's set as the user inputted value could be the 0 value. If we made verify an enum we could make it's value skipSignatureValidation when empty. However, I would still advocate for removing skipSignatureValidation, given that we are intending to make a breaking change here.

src/cmd/initialize.go
```diff
@@ -103,8 +104,13 @@ func newInitCommand() *cobra.Command {
...
 	cmd.Flags().IntVar(&o.ociConcurrency, "oci-concurrency", v.GetInt(VPkgOCIConcurrency), lang.CmdPackageFlagConcurrency)
+	cmd.Flags().BoolVar(&o.skipSignatureValidation, "skip-signature-validation", false, lang.CmdPackageFlagSkipSignatureValidation)
```
AustinAbro321, line 110: As I'm thinking about it more, given that we're leaning into a breaking change already, should we keep this flag around that doesn't have any effect? Removing it will break some automated workflows, but the fix is simple across the board. I think it will bring more awareness to the breaking change that we are making, and ultimately leave users with more clarity on how Zarf works. One situation I'm thinking about that's unlikely but possible is someone running `--skip-signature-validation=false`. In this case we are leaving them with a less strict security posture, and they will have to notice the warning to figure this out. 

Totally fine with keeping it as is. Just wanted to throw this line of thinking out here.

src/pkg/packager/load.go
brandtkeller: i've come to agree with this stance given the current posture of the changes and the project. I was experimenting with those explicit mentions yesterday to inform what the permutations look like for impacts. 

In the future we'll have a lot less optionality for breaking changes without major version implications - so this was meant to drive some more discussion around how we manage this across the SDK. Each scenario is nuanced and this change is a breaking change regardless of how we handle the field deprecation given that the we can really only handle the behavior of the new functionality without more complexity.

src/cmd/initialize.go
brandtkeller: If at all possible I want to control breaking end-user workflows as those can be tedious to repair and costly if not done in a centralized manner. Let me look into this further.

src/pkg/packager/layout/package.go
brandtkeller: You raise a good point - which is a consistent stance across CLI/DevEx. 

> we don't want to encourage SDK users running VerifyNever

I agree - could we invert the location of `VerifyNever` to be the last option? I'm not seeing a limitation that current requires the new public function but I am open to it.

src/pkg/packager/load.go
brandtkeller: Field removed from options.

src/cmd/initialize.go
brandtkeller: New approach provides a backwards compatible change for the CLI (SDK is still a breaking change as covered in the other discussion). Would love your thoughts - working to test each permutation but it does work well with `zarf package pull oci://ghcr.io/zarf-dev/packages/dos-games:1.2.0`.

AustinAbro321: I like this approach

src/pkg/packager/layout/package.go
AustinAbro321: Moving the enum so the default is `verifyIfpossible` is an approach I'd be fine with. Worth noting that since we are only using `verifyNever` within the layout package we could introduce a private function that assemble calls, which could eliminate the verify enum / simplify the API. 

Let me know which you prefer, otherwise pr looks good

brandtkeller: Given a private function and elimination of the enum - would this imply that there is a single strategy that SDK users can perform (unless we create multiple exported functions)?

I want to ensure we can support the require states (enforced, attempt, disregard) - my current stance being that the enum can support these well and reduce too much in required SDK variation/duplication.

AustinAbro321: Yes it would. Your reasoning makes sense, thanks for the explanation.

AustinAbro321 (approved):

PR #4463, opened 2025-12-10
chore(schema): synchronize v1alpha1 schema with generation

brandtkeller:
## Description

Noticed that the schema generation with `make docs-and-schema` looks to have changes in the v1alpha1 schema that have already been synchronized with the root `zarf.schema.json` but not the source schema. 

## Related Issue

No associated issue

## Checklist before merging

### DIFF
--- a/hack/check-zarf-docs-and-schema.sh
+++ b/hack/check-zarf-docs-and-schema.sh
@@ -2,10 +2,10 @@
 
 set -euo pipefail
 
-if [ -z "$(git status -s ./site/src/content/docs/commands/ ./zarf.schema.json)" ]; then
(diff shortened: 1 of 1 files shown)


AustinAbro321:
@brandtkeller I believe this was a regression due to merge timing, but could you also add the new v1alpha1 schema file to check-zarf-docs-and-schema.sh, just to make sure this doesn't happen again for some reason

brandtkeller:
> @brandtkeller I believe this was a regression due to merge timing, but could you also add the new v1alpha1 schema file to check-zarf-docs-and-schema.sh, just to make sure this doesn't happen again for some reason

@AustinAbro321 I'm good with that - Added!

PR #4466, opened 2025-12-10
feat: add --set-values flags and migrate --set to alias new --set-variables

### LINKED ISSUE #4224: Add --set-value flag for granular values key overrides

mkcp:
## Description
This PR adds the `--set-values` flag so users can directly override values via the CLI, and begins the path to migrating `--set`. To do so, we add the `--set-variables` flag (or versions of it like `--deploy-set-variables`) and change the description of `--set` to indicate that it now aliases --set-variables. This alias is just a documentation change in the CLI help text: the flags function identically. 

## Related Issue
Fixes #4224 

## Checklist before merging

### DIFF
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -824,6 +824,7 @@ func newPackageInspectValuesFilesCommand(v *viper.Viper) *cobra.Command {
 	cmd.Flags().StringVar(&o.components, "components", "", "comma separated list of components to show values files for")
 	cmd.Flags().StringVar(&o.kubeVersion, "kube-version", "", lang.CmdDevFlagKubeVersion)
 	cmd.Flags().StringToStringVar(&o.setVariables, "set", v.GetStringMapString(VPkgDeploySet), "Alias for --set-variables")
+	_ = cmd.Flags().MarkDeprecated("set", "use --set-variables instead")
 	cmd.Flags().StringToStringVar(&o.setVariables, "set-variables", v.GetStringMapString(VPkgDeploySet), lang.CmdPackageDeployFlagSetVariables)
 	errSig := cmd.Flags().MarkDeprecated("skip-signature-validation", "Signature verification now occurs on every execution, but is not enforced by default. Use --verify to enforce validation. This flag will be removed in Zarf v1.0.0.")
 	if errSig != nil {
@@ -937,6 +938,7 @@ func newPackageInspectManifestsCommand(v *viper.Viper) *cobra.Command {
 	cmd.Flags().StringVar(&o.components, "components", "", "comma separated list of components to show manifests for")
 	cmd.Flags().StringVar(&o.kubeVersion, "kube-version", "", lang.CmdDevFlagKubeVersion)
 	cmd.Flags().StringToStringVar(&o.setVariables, "set", v.GetStringMapString(VPkgDeploySet), "Alias for --set-variables")
+	_ = cmd.Flags().MarkDeprecated("set", "use --set-variables instead")
 	cmd.Flags().StringToStringVar(&o.setVariables, "set-variables", v.GetStringMapString(VPkgDeploySet), lang.CmdPackageDeployFlagSetVariables)
 	errSig := cmd.Flags().MarkDeprecated("skip-signature-validation", "Signature verification now occurs on every execution, but is not enforced by default. Use --verify to enforce validation. This flag will be removed in Zarf v1.0.0.")
 	if errSig != nil {
--- a/site/src/content/docs/commands/zarf_package_inspect_manifests.md
+++ b/site/src/content/docs/commands/zarf_package_inspect_manifests.md
@@ -22,7 +22,6 @@ zarf package inspect manifests [ PACKAGE ] [flags]
   -k, --key string                     Path to public key file for validating signed packages
       --kube-version string            Override the default helm template KubeVersion when performing a package chart template
       --oci-concurrency int            Number of concurrent layer operations when pulling or pushing images or packages to/from OCI registries. (default 6)
-      --set stringToString             Alias for --set-variables (default [])
       --set-variables stringToString   Specify deployment variables to set on the command line (KEY=value) (default [])
       --verify                         Verify the Zarf package signature
 ```
--- a/site/src/content/docs/commands/zarf_package_inspect_values-files.md
+++ b/site/src/content/docs/commands/zarf_package_inspect_values-files.md
@@ -26,7 +26,6 @@ zarf package inspect values-files [ PACKAGE ] [flags]
   -k, --key string                     Path to public key file for validating signed packages
       --kube-version string            Override the default helm template KubeVersion when performing a package chart template
       --oci-concurrency int            Number of concurrent layer operations when pulling or pushing images or packages to/from OCI registries. (default 6)
-      --set stringToString             Alias for --set-variables (default [])
       --set-variables stringToString   Specify deployment variables to set on the command line (KEY=value) (default [])
       --verify                         Verify the Zarf package signature
 ```


src/cmd/initialize.go
```diff
@@ -61,7 +61,8 @@ func newInitCommand() *cobra.Command {
 	v := getViper()
 
 	// Init package set variable flags
-	cmd.Flags().StringToStringVar(&o.setVariables, "set", v.GetStringMapString(VPkgDeploySet), lang.CmdInitFlagSet)
+	cmd.Flags().StringToStringVar(&o.setVariables, "set", v.GetStringMapString(VPkgDeploySet), "Alias for --set-variables")
```
brandtkeller, line 66: Should we explicitly mark these as deprecated? 

```golang
cmd.Flags().MarkDeprecated("set", "use --set-variables instead")
```

src/cmd/initialize.go
```diff
@@ -61,7 +61,8 @@ func newInitCommand() *cobra.Command {
...
+	cmd.Flags().StringToStringVar(&o.setVariables, "set-variables", v.GetStringMapString(VPkgDeploySet), lang.CmdInitFlagSetVariables)
```
brandtkeller, line 68: Given this alias - do we need to mark these flags as mutually exclusive? I understand this to create a last-write-wins scenario if there were to be both used?

brandtkeller (changes requested):
migration makes sense - added some thoughts around guardrails natively available in cobra that we might consider.

src/cmd/initialize.go
AustinAbro321: Rather than having a separate command as an alias that we mark as deprecated or hide, I think it would be better to use the builtin cobra functionality for aliasing a command. This leads to a more intuitive experience around viper config, and will lead to an easier time aliasing `--set-values` to `--set` in the future in the event that variables is fully replaced

For instance 
`cmd.Flags().StringToStringVarP(&o.setVariables, "set-variables", "set", v.GetStringMapString(VPkgCreateSet), lang.CmdPackageCreateFlagSetVariables)`

src/cmd/dev.go
```diff
@@ -697,8 +710,10 @@ func newDevFindImagesCommand(v *viper.Viper) *cobra.Command {
 		logger.Default().Debug("unable to mark dev-find-images flag as hidden", "error", err)
 	}
 	cmd.Flags().StringVarP(&o.flavor, "flavor", "f", v.GetString(VPkgCreateFlavor), lang.CmdPackageCreateFlagFlavor)
-	cmd.Flags().StringToStringVar(&o.createSetVariables, "create-set", v.GetStringMapString(VPkgCreateSet), lang.CmdDevFlagSet)
-	cmd.Flags().StringToStringVar(&o.deploySetVariables, "deploy-set", v.GetStringMapString(VPkgDeploySet), lang.CmdPackageDeployFlagSet)
+	cmd.Flags().StringToStringVar(&o.createSetVariables, "create-set", v.GetStringMapString(VPkgCreateSet), "Alias for --create-set-variables")
+	cmd.Flags().StringToStringVar(&o.createSetVariables, "create-set-variables", v.GetStringMapString(VPkgCreateSet), lang.CmdDevFlagSetVariables)
```
AustinAbro321: IMO we shouldn't add the variables verbiage to create time templates. Generally I think we should refer to these as package templates. I just noticed that Zarf is inconsistent right now since we call them package variables in this flag, and package templates in our docs.  Still, I don't think we'll ever have two types of package templates, at least not in the same API version, so I'm not sure extra verbosity is helpful.

src/cmd/viper.go
```diff
@@ -100,6 +100,7 @@ const (
 	VPkgDeployNamespace  = "package.deploy.namespace"
 	VPkgRetries          = "package.deploy.retries"
 	VPkgDeployValues     = "package.deploy.values"
+	VPkgDeploySetValues  = "package.deploy.set_values"
```
AustinAbro321, line 103: Currently both `package.deploy.values` and `package.deploy.set_values` are in use. Not sure which I prefer, but we should only use one.

src/cmd/package.go
```diff
@@ -306,14 +310,26 @@ func (o *packageDeployOptions) run(cmd *cobra.Command, args []string) (err error
 	v := getViper()
 	o.setVariables = helpers.TransformAndMergeMap(
 		v.GetStringMapString(VPkgDeploySet), o.setVariables, strings.ToUpper)
+	o.setValues = helpers.TransformAndMergeMap(
```
AustinAbro321: When there is no transform happening, this function is simply a `maps.Copy`, lets just use that instead.

AustinAbro321 (dismissed):

src/cmd/initialize.go
mkcp: I don't believe making them mutually exclusive is necessary. Multiple invocations of --set (soon --set-variables) already merge entries cleanly together, adding --set in between will be the same behavior.

src/cmd/viper.go
mkcp: --values accepts values files and --set-values accepts keys/values. Users will have valid use cases for either one, or both at the same time, so we should keep them.

src/cmd/initialize.go
mkcp: I understand the argument for making --set a flag shorthand, but I don't think I agree with doing so. Typically the shorthand for flags are one letter and intended to be stable for the life of the flag. Implementing `--set` as a separate command allows us to notify users when the behavior is deprecated.

michalszynkiewicz:
Do I understand correctly that we only support overriding values at deployment time and not the packaging time? Or is there a way to override when packaging too?

src/cmd/viper.go
AustinAbro321: Ah whoops, my mistake

src/cmd/initialize.go
AustinAbro321: Ah okay I was recommending this with the intent that `--set` would one day be the alias to `--set-values` once `--set-variables` is removed. Given that we plan on removing `--set` at some point, I would agree with @brandtkeller's assessment that it should be deprecated

mkcp:
@michalszynkiewicz Following up here and from other threads - the values files included in the package definition are intended to be the package defaults, and the set of values the package is created with. Allowing those defaults to be overrided at create-time in a way changes the declarative nature of the package and means you're building a different package than the definition, in a way.

After discussion, we'd be more likely to meet this usecase with more powerful package templates resolved in a pre-processing step. Then you'd be able to programmatically generate different packages from a base package definition e.g `zarf.yaml.tmpl`, and then run `zarf package create` on the result once all templates are resolved.

src/cmd/dev.go
mkcp: I get what you mean now. I updated the cli string verbiage and config structs to package templates / PkgTmpl, but only in the `cmd` pkg to keep the scope of the PR constrained. This should mean that users will now see `"package templates"` in the CLI layer, but internally there are still references to CreateSetVariables in dev. This should give us a step towards more consistent handling without trying to solve all of it here.

src/cmd/initialize.go
brandtkeller: Ah given the underlying type this appears to work as expected 👍

src/cmd/initialize.go
brandtkeller: I see `set` marked as deprecated in the `package` commands but not here in `init` so I figured id follow up here - do we intend to set this to deprecated as well?

mkcp: Ah, yes this has an alternative flag so it's meant to be deprecated. Updating

Issue #4419, opened 2025-11-25
Add ability to invert the zarf control of namespaces


kgalloway-bp:
### Is your feature request related to a problem? Please describe.
When zarf is deployed to a cluster, namespaces must "opt-out" of zarf control via adding the `zarf.dev/agent: ignore` label. This allows zarf agent to ignore everything in the namespace. 

### Describe the behavior you'd like

- **Given** an existing cluster ecosystem
- **When** adding zarf to a cluster
- **Then** a flag can be set to have namespaces be "opt-in" to zarf agent instead of "opt-out"

### Additional context
The purpose of this is to allow zarf to be implemented/used in clusters that are already in-use without conflicting with anything there (such as flux/argo/etc). It allows teams to also use a mixed-deployment style when sharing cluster resources where one team may deploy with zarf and others with some other tooling.

brandtkeller:
I agree with this use case and have also proposed similar in discussions. It might require some state for decision making but given the decision to opt-in explicitly - it would alleviate the concerns of using tooling (gitops, operators, etc) that unknowingly creates additional resources in other namespaces and then is incompatible with Zarf.

We'd want to think about the implications - such as ephemeral containers and other zarf expectations for mutation.

I'll bring this up at the next community meeting for discussion - we'd love to hear from you if interested.

vyaronvstavitzky:
We actually implemented exactly this and it’s working flawlessly.
We do however label all our namespaces with a specific label that the mutator will match.

We did this since we are operating in a semi-airgapped environment and there are cases where zarf is being deployed along with our packages to an existing cluster.

We can share a PR.
We indeed faced an issue where an operator in our package created a namespace and therefore lacked the label, so we ended up with different approaches such as an accompanying kyverno to watch for our namespaces and add the label to them on creation.

We are happy to hear your thoughts and recommendations.

brandtkeller:
I would love to hear more about your experience. My concern is less about the switch of this behavior in isolation and more about the implications - such as assumed behaviors given the Zarf ownership model today and what would be impacted. 

Given that the default stance remains the same and the inverted-stance is a configuration users can opt-in to, then I think this would be reasonable to pursue.

vyaronvstavitzky:
@brandtkeller  I can ask my colleague to create a simple PR, or would you rather we share the git diff?

brandtkeller:
@vyaronvstavitzky PR's are wonderful.

yuvalsimon:
[...]

PR #4467, opened 2025-12-10
feat: Add opt-in option for zarf init

yuvalsimon:
## Description
This PR adds flags `--agent-namespaces-mode=[active|passive]` and `--agent-objects-mode=[active|passive]` to `zarf init` command. By default, both are set to _active_. When using either of them to _passive_,  _MutatingWebhookConfiguration_ will be deployed with rules that enforce namespaces/objects to be labeled with `zarf.dev/agent: mutate`, or else the images won't be mutated. An exception that was added is _zarf_ namespace, there effectively modes are always _active_ (which enables us deploy there _Gitea_ for instance)

examples:

- **agent-namespaces-mode=active**; **agent-objects-mode=active**: Current zarf behaviour
- **agent-namespaces-mode=passive**; **agent-objects-mode=active**: If namespace is labeled with `zarf.dev/agent: mutate` then like current zarf behaviour on this whole namespace. Otherwise no images are mutated at all. If the pod is labeled with `zarf.dev/agent: ignore`, then it won't be mutated neither.
- **agent-namespaces-mode=active**; **agent-objects-mode=passive**: If a pod is labeled with `zarf.dev/agent: mutate` then image is mutated pod. Otherwise no mutation is applied.
- **agent-namespaces-mode=passive**; **agent-objects-mode=passive**: Only if both namespace and pod in the namespace are labeled `zarf.dev/agent: mutate` then image mutation occurs.

## Related Issue

Relates to #4419

## Checklist before merging

### DIFF
--- a/packages/zarf-agent/agent-values.yaml
+++ b/packages/zarf-agent/agent-values.yaml
@@ -1,0 +1,3 @@
+mode:
+  namespaces: "###ZARF_VAR_AGENT_NAMESPACES_MODE###"
+  objects: "###ZARF_VAR_AGENT_OBJECTS_MODE###"
--- a/packages/zarf-agent/chart/templates/_helpers.tpl
+++ b/packages/zarf-agent/chart/templates/_helpers.tpl
@@ -1,0 +1,31 @@
+{{/* vim: set filetype=mustache: */}}
+
+{{/*
+Generate selector expression based on mode.
+Active mode: exclude resources with zarf.dev/agent: skip|ignore
+Passive mode: only include resources with zarf.dev/agent: mutate, exclude resources with zarf.dev/agent: skip|ignore
+Usage: {{ include "zarf-agent.webhook.selectorExpression" "active" }}
+*/}}
+{{- define "zarf-agent.webhook.selectorExpression" -}}
+- key: zarf.dev/agent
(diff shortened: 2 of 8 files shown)


AustinAbro321:
Hello @yuvalsimon, really appreciate you making this PR. This is definitely a workflow that I feel Zarf should support. I am curious about your use cases. Could you tell me situations where you tend to use agent-objects-mode=active. Do you tend to only want mutations on certain kinds of resources, for instance, you want to mutate pods, but not Flux Git Repositories? Or is it the case where you want one package mutated in a namespace, but not the other? 

From the issue I see that you have kyverno adding this label. I think for the best user experience, we would want these labels to be automatically added to namespaces Helm post-renderer in [AdoptZarfManagedLabels](https://github.com/zarf-dev/zarf/blob/93623493ba4d39175543e271cb8ff78a3ebb7474/src/pkg/cluster/namespace.go#L74). This way any Zarf deployments automatically opt into mutation. Something similar could be done for pods and argo / flux resources. Do you have use cases, where you would not want this behavior?

AustinAbro321:
Thanks for the PR, closing this as we went with a different solution described here - https://docs.zarf.dev/ref/init-package/#agent-mutation-rules

PR #4486, opened 2025-12-18
fix: properly mutate flux ocirepo and helmrepo objects when using the registry proxy

### LINKED ISSUE #4474: When using [alpha] feature `registry-mode`, OCIRepo resource not properly mutated to intra-cluster endpoint
### Environment
OS: Rocky Linux 9.6
App version: `v0.67.0`
Kubernetes distro being used: `k0s`
Other: `--features=registry-proxy=true --registry-mode=proxy`

### Steps to reproduce
1. `zarf init --features=registry-proxy=true --registry-mode=proxy --confirm`
2. Deploy fluxcd
3. Deploy fluxcd OCI repo resource

### Expected result

Webhook mutates to intra-cluster docker registry

### Actual Result

Webhook mutates to "original address", or `127.0.0.1:5000`

```text
2025-12-14 02:24:05 DBG registry appears to not be a nodeport service, using original address address=127.0.0.1:5000
```

### Code links

https://github.com/zarf-dev/zarf/blob/81e5becb15fda68f2759a5e5fbb274fde9e870ec/src/internal/agent/hooks/flux-ocirepo.go#L74

https://github.com/zarf-dev/zarf/blob/81e5becb15fda68f2759a5e5fbb274fde9e870ec/src/pkg/cluster/secrets.go#L165-L179

AustinAbro321:
## Description

This fixes a bug with the registry proxy solution, where it does not work with flux and helm repo objects since the clusterip / hostname of the service is not added to the secret and was not returned. 

## Related Issue

Fixes #4474

## Checklist before merging

### DIFF
--- a/src/pkg/lint/rules_test.go
+++ b/src/pkg/lint/rules_test.go
@@ -117,7 +117,7 @@ func TestIsImagePinned(t *testing.T) {
 		{
 			input:    "busybox:bad/image",
 			expected: false,
-			err:      errors.New("invalid reference format"),
+			err:      errors.New("unable to parse ref busybox:bad/image: invalid reference format"),
 		},
 		{
 			input:    "busybox:###ZARF_PKG_TMPL_BUSYBOX_IMAGE###",


src/pkg/cluster/secrets.go
```diff
@@ -162,18 +181,29 @@ func (c *Cluster) UpdateZarfManagedGitSecrets(ctx context.Context, s *state.Stat
...
+	svc, port, err := serviceInfoFromNodePortURL(serviceList.Items, registryInfo.Address)
 	if err != nil {
-		logger.From(ctx).Debug("registry appears to not be a nodeport service, using original address", "address", stateRegistryAddress)
-		return stateRegistryAddress, nil
+		logger.From(ctx).Debug("registry appears to not be a nodeport service, using original address", "address", registryInfo.Address)
+		return registryInfo.Address, nil
 	}
 
-	return fmt.Sprintf("%s:%d", svc.Spec.ClusterIP, port), nil
+	return fmt.Sprintf("%s.%s.svc.cluster.local:%d", svc.Name, svc.Namespace, port), nil
```
AustinAbro321: Changing behavior of the nodeport service as well. I cannot think of any reason we initially used clusterip. Seems like hostname is a more consistent alternative.

a1994sc: Also a little un-educated guess, but I wonder if it is because not every cluster uses `cluster.local` as its FQDN?

AustinAbro321: Looks like cluster.local is available everywhere https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/.

a1994sc: > DNS names also need domains. You configure the local domain in the kubelet with the flag --cluster-domain=\<default-local-domain>.

Ref: https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/

It possible to set the internal DNS record to something other than `cluster.local`

AustinAbro321: Ah just tested it out, you are absolutely correct. Thanks for the catch!

src/pkg/cluster/secrets.go
```diff
@@ -33,6 +34,19 @@ type DockerConfigEntryWithAuth struct {
...
 }
 
+// addRegistryAuthEntries adds registry authentication entries for a service's ClusterIP and DNS hostname.
+func addRegistryAuthEntries(auths DockerConfigEntry, svc *corev1.Service, port int32, authValue string) {
+	kubeDNSRegistryURL := net.JoinHostPort(svc.Spec.ClusterIP, fmt.Sprintf("%d", port))
+	auths[kubeDNSRegistryURL] = DockerConfigEntryWithAuth{
+		Auth: authValue,
+	}
+
+	kubeDNSRegistryHostname := fmt.Sprintf("%s.%s.svc.cluster.local:%d", svc.Name, svc.Namespace, port)
```
a1994sc, line 44: I wonder if in the future we should have the ability to override what the cluster domain is? Beyond the scope of this specific PR, but wanted to throw that out there

AustinAbro321: Yeah, it's an interesting question, we do hardcode the git server as a cluster address with `.cluster.local` currently so there are definitely a few important paths where it's used. I'm not sure what the use cases are, but if someone opened up an issue I'd be open to it. Ideally there would be a cluster agnostic way to find this information automatically, but I'm not sure there is in this case.

PR #4488, opened 2025-12-18
test: use arch of local machine in external tests

robinlieb:
## Description

Running external tests on arm machines currently fails because of the hardcoded "amd64" in the zarf package archive names in the arguments. 

This PR adds dynamically construction of the name bases on the machine these tests are running on.

## Checklist before merging


brandtkeller:
Appreciate you identifying this! I was also setting up a linux arm testing environment recently and intended to work through these accordingly! I will give this a test.

PR #4489, opened 2025-12-18
fix: use fully qualified reference during image archive duplication check

AustinAbro321:
## Description

Fixes a bug in the image archive duplication logic where something like this wouldn't be flagged as a duplication, since the component image logic was using the image archive 
```yaml
    images:
     - rancher/k3s:v1.33.6-k3s1
    imageArchives:
      - path: regular.tar
        images: 
          - ghcr.io/zarf-dev/doom-game:0.0.1
          - rancher/k3s:v1.33.6-k3s1
```

## Checklist before merging

### DIFF
--- a/src/pkg/packager/layout/assemble_test.go
+++ b/src/pkg/packager/layout/assemble_test.go
@@ -56,4 +56,3 @@ func TestCreateReproducibleTarballFromDir(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, "c09d17f612f241cdf549e5fb97c9e063a8ad18ae7a9f3af066332ed6b38556ad", shaSum)
 }
-


src/pkg/lint/validate.go
```diff
@@ -334,3 +340,43 @@ func validateManifest(manifest v1alpha1.ZarfManifest) error {
...
+	for _, comp := range components {
+		for _, archive := range comp.ImageArchives {
+			for _, image := range archive.Images {
+				refInfo, err := transform.ParseImageRef(image)
+				if err != nil {
+					return fmt.Errorf("failed to parse image ref %s in archive %s: %w", image, archive.Path, err)
+				}
+
+				if existingArchivePath, exists := imageToArchive[refInfo.Reference]; exists {
+					// A user may want to represent the same tar twice across components if both components need the same image
```
brandtkeller, line 358: ++ this is a good distinction for consideration with components and component-imports.

Issue #4487, opened 2025-12-18
Allow different filenames to be built.


Dale-Sample:
### It's difficult to create many deployments of similar but not identical collections of zarf files
I'd like to be able to do something like:
root 
small.zarf.yaml - import x,y
big.zarf.yaml -import x,y,z,a,b,c,...
    x
          x-deploy.yaml
   ...
...
and use the command
`zarf package create small.zarf.yaml ` to create the small package. 
and 
`zarf package create big.zarf.yaml ` to create the big package.

### Describe the behavior you'd like

- **Given** a file path exists with a zarf file that can be built.
- **When** `zarf package create ${the file}.yaml `
- **Then** a zarf package is created using that file. No change for the default behavior of package create.

### Describe alternatives you've considered
I'm fine with any number of possible solutions, including 
`zarf package create --file theFile.yaml`
or 
`zarf package create /the/path/to/the/folder/with/the/file/`
as long as I can point multiple files to the same root setup. 
I prefer the complete path to a file option over a directory, but supporting both would be nice (though the directory option is not one that I'm super excited about, if it were dropped I wouldn't care).

brandtkeller:
completed in #4493

PR #4493, opened 2025-12-20
feat(create): allow creation with specified and arbitrary manifest file

brandtkeller:
## Description

Allow the ability to specify arbitrarily named definition files while retaining the current behaviors as default. 

This PR allows users to specify the Zarf manifest file by name during create and/or in import paths in the manifest - to include manifest files not named `zarf.yaml`

## Related Issue

Relates to #4487

## Checklist before merging

### DIFF
--- a/src/pkg/packager/load/import.go
+++ b/src/pkg/packager/load/import.go
@@ -156,12 +156,16 @@ func resolveImports(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath, 
 			return v1alpha1.ZarfPackage{}, err
 		}
 
-		importPackagePath, err := layout.ResolvePackagePath(filepath.Join(pkgPath.BaseDir, importPath))
+		// this is a special case for paths and imports where we do not want to join BaseDir and importPath
+		// we check that the path is valid but ensure the value remains relative for fixing
+		fileInfo, err := os.Stat(filepath.Join(pkgPath.BaseDir, importPath))
 		if err != nil {
 			return v1alpha1.ZarfPackage{}, fmt.Errorf("unable to access import path %q: %w", importPath, err)
 		}
-
-		importedComponent = fixPaths(importedComponent, importPackagePath.BaseDir, pkgPath.BaseDir)
+		if !fileInfo.IsDir() {
+			importPath = filepath.Dir(importPath)
+		}
+		importedComponent = fixPaths(importedComponent, importPath, pkgPath.BaseDir)
 		composed, err := overrideMetadata(importedComponent, component)
 		if err != nil {
 			return v1alpha1.ZarfPackage{}, err


src/pkg/packager/create.go
```diff
@@ -60,6 +61,16 @@ func Create(ctx context.Context, packagePath string, output string, opts CreateO
...
 
+	// Determine base directory for assembly
+	basePath := packagePath
+	fileInfo, err := os.Stat(packagePath)
+	if err != nil {
+		return "", fmt.Errorf("unable to access package path %q: %w", packagePath, err)
+	}
+	if !fileInfo.IsDir() {
+		basePath = filepath.Dir(packagePath)
+	}
```
AustinAbro321: I see this logic in a few places, I think it'd be helpful from a DRY and readability standpoint to have a function such as `getPackageDefinitionParentDir()`. Perhaps in the layout or util package.

src/pkg/lint/schema.go
```diff
@@ -16,10 +17,23 @@ import (
...
+
+	// Determine config path: if path is directory, append layout.ZarfYAML; otherwise use as-is
+	configPath := path
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return nil, fmt.Errorf("unable to access path %q: %w", path, err)
+	}
+	if fileInfo.IsDir() {
+		configPath = filepath.Join(path, layout.ZarfYAML)
+	}
```
AustinAbro321: Similarly, I think it'd be helpful for readability to have a helper function such as `layout.GetPackageDefinitionPath()`

src/pkg/packager/update.go
```diff
@@ -20,9 +20,17 @@ import (
 )
 
 // UpdateImages updates the images field for components in a zarf.yaml
-func UpdateImages(ctx context.Context, baseDir string, imagesScans []ComponentImageScan) error {
+func UpdateImages(ctx context.Context, packagePath string, imagesScans []ComponentImageScan) error {
```
AustinAbro321, line 22: PR description says there is a breaking change in this function, but it looks like this is non-breaking

brandtkeller: Great point - I was playing very conservatively with regards to any potential changes. Intent changes for the function parameter but it is handled gracefully so as to not be breaking. Thanks for calling that out.

src/pkg/packager/create.go
brandtkeller: I had contemplated moving `resolvePackagePath` to another location and/or exporting it for public use. Thoughts?

AustinAbro321: That seems like a reasonable path, and I believe would allow all the same solution to be used across all approaches

src/pkg/lint/schema.go
brandtkeller: updated

src/pkg/packager/create.go
brandtkeller: updated

PR #4503, opened 2026-01-07
fix: remove values log

AustinAbro321:
## Description

I noticed a few places where we log user submitted values. This is generally inadvisable as users may submit sensitive values as such as passwords 

## Checklist before merging


mkcp (approved):
This makes sense - IIRC these were originally used for development and I kept them in as debug prints. Better not to risk it.

PR #4506, opened 2026-01-08
CI: image archive test flake

AustinAbro321:
## Description

Test 44 image archives has been failing lately as seen by https://github.com/zarf-dev/zarf/actions/runs/20819481889/job/59803742229. It seems like the pod is crashing and starting, I believe because of the sleep 1 job. Looks like in some scenarios kubernetes never gets the chance to see the pod as healthy. I updated this to sleep infinity to avoid this in the future

## Checklist before merging


a1994sc:
LGTM

a1994sc (approved):

PR #4511, opened 2026-01-09
WIP: port init packages to value templating

### LINKED ISSUE #4328: Replace variables in init package with values
Zarf's package values are intended to replace variables and constants. We currently use variable templating extensively in Zarf's init package, so phasing this out in favor of package values is necessary for deprecating vars.

Once we're reasonably happy with the state of package values, we should swap over from using variables in the init package - dogfooding the feature on the way to general availability for users. 

### Additional context
We should be wary of anywhere values' go-templates might collide with go-templates that are meant to be consumed by downstream systems.

mkcp:
## Description
Experimenting with how much of the init packages we can migrate over to values and go-templating.

Package templates we don't yet have a good answer for.

## Related Issue
#3946 

## Checklist before merging

### DIFF
--- a/packages/distros/eks/eks.yaml
+++ b/packages/distros/eks/eks.yaml
@@ -2,22 +2,22 @@
 apiVersion: eksctl.io/v1alpha5
 kind: ClusterConfig
 metadata:
-  name: ###ZARF_VAR_EKS_CLUSTER_NAME###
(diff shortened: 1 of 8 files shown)


AustinAbro321:
We've discussed this and decided that we are not going to add values to the v1alpha1 init package, but instead move to values when we create the v1beta1 package. This way the breaking change will be clear and likely coincide with 1.0

PR #4519, opened 2026-01-14
feat(sign): implement support for sigstore bundle format

### LINKED ISSUE #4276: docs: clarify asymmetric KMS key needed for signing

### LINKED ISSUE #4296: Support for cosign bundle format

brandtkeller:
## Description

This is a backwards compatible change to update Zarf to using the sigstore bundle format for signatures when signing. 

This sets the default signing options to empty strings for infrastructure endpoints - cementing zarfs stance on ensuring no unintentional data or network calls are made. 

This is the first iteration towards [ZEP-0053: Bundle format and migration and verification strategy](https://github.com/zarf-dev/proposals/blob/main/0053-bundle-format/README.md). Intentionally establishing a baseline for compatibility with the bundle format before introducing more complexity in verification. 

Packages will now include both a `zarf.yaml.sig` and a `zarf.bundle.sig` until we fully deprecate the legacy signature format. Given intent to support verification of the legacy format - it could be as little as multiple releases. 

Validated that a package without the legacy signature performs verification as intended. 

Additionally this adds a tutorial for package sign/verify for the current baseline.

## Related Issue

Fixes #4296 
Fixes #4276
Relates to https://github.com/zarf-dev/proposals/issues/53 

## Checklist before merging


site/src/content/docs/tutorials/5-package-signing-and-verification.mdx
AustinAbro321, line 1: IMO this page should be broken up into two different pages. A reference page explaining more of the intricacies of signing, and a tutorial that goes through a tight example. I like how this framework thinks of tutorials https://diataxis.fr/tutorials/. 

The sections "Verifying Unsigned Packages" and "Signature Format Migration" both feel like they belong more so in a reference document rather than this tutorial. I believe the tighter tutorial structure would also make it easy to delete the "Complete Example Workflow" section as well.

src/pkg/packager/layout/package.go
```diff
@@ -234,13 +235,20 @@ func (p *PackageLayout) SignPackage(ctx context.Context, opts utils.SignBlobOpti
...
 	// Check if signature already exists in actual layout and warn
 	actualSignaturePath := filepath.Join(p.dirPath, Signature)
 	if _, err := os.Stat(actualSignaturePath); err == nil {
 		l.Warn("overwriting existing package signature", "path", actualSignaturePath)
 	}
 
+	// duplicate warning for overwriting the bundle signature
+	actualBundlePath := filepath.Join(p.dirPath, Bundle)
+	if _, err := os.Stat(actualBundlePath); err != nil {
+		l.Warn("overwriting existing package bundle signature", "path", actualBundlePath)
```
AustinAbro321: Do we want to expose the overwrite flag you've introduced at the SDK level? By doing this, we could avoid a warning message (maybe info or debug instead), and return an error when overwrite is not set.

brandtkeller: I think that is reasonable - changing the behavior to error otherwise will mean we need to adjust any commands that can potentially overwrite the signature and ensure that expose it accordingly (thinking publish in particular).

site/src/content/docs/tutorials/5-package-signing-and-verification.mdx
brandtkeller: I appreciate the distinction and went ahead with separating the content accordingly.

src/pkg/packager/layout/package.go
brandtkeller: Added to the cosign utils due to where the `SignBlobOptions` reside. I decided to pause on supporting overwrite in publish - as it feels like we might be at an intersection for deciding whether sign/verify are embedded in workflows or whether a developer using the SDK instruments those behaviors explicitly.

site/src/content/docs/ref/package-signing.mdx
```diff
@@ -0,0 +1,331 @@
...
+
+- **Authenticity**: Verify that a package comes from a trusted source
+- **Integrity**: Ensure the package has not been modified or corrupted
+- **Non-repudiation**: Prove who signed the package and when
+
+Zarf signs the `zarf.yaml` file within a package, which contains metadata and checksums for all package contents. This allows verification of the entire package through a single signature.
+
+## Signature Formats
+
+### Sigstore Bundle Format (Recommended)
```
AustinAbro321: I feel that the word "recommended" doesn't fit here as it implies that user's have a choice on which signature format that they can use. However, it's always chosen for them depending on the Zarf version.

site/src/content/docs/ref/package-signing.mdx
```diff
@@ -0,0 +1,331 @@
...
+
+Re-sign packages with legacy signatures using the `zarf package sign` command to upgrade to the bundle format. The `--overwrite` flag allows replacing existing signatures.
+
+:::
+
+## Signing Methods
+
+### Local Key Files
+
+The most common approach uses local private/public key pairs:
```
AustinAbro321: For now this is the only approach correct? We'll have to update it in the future, but I'd be confused reading this verbiage currently.

site/src/content/docs/ref/package-signing.mdx
```diff
@@ -0,0 +1,331 @@
...
+- **Standardized**: Based on the widely-adopted Sigstore specification
+- **Better interoperability**: Compatible with other Sigstore tooling and services
+
+The bundle format follows the [Sigstore Bundle Specification](https://github.com/sigstore/cosign/blob/main/specs/BUNDLE_SPEC.md).
+
+### Legacy Signature Format
+
+:::caution[Deprecated]
+
+The legacy signature format (`zarf.yaml.sig`) is deprecated in favor of the Sigstore bundle format. While Zarf currently supports both formats for backward compatibility, the legacy format will soon no longer be generated for new signatures. Zarf will retain the ability to verify legacy signatures as a fallback for when a package is signed but no bundle signature is present.
```
AustinAbro321, line 33: As of this PR being merged it will no longer be generated for any new signatures correct?

site/src/content/docs/ref/package-signing.mdx
```diff
@@ -0,0 +1,331 @@
...
+1. **Verify before deployment**: Always verify package signatures before deploying to critical environments
+2. **Automate verification**: Integrate signature verification into CI/CD pipelines
+3. **Enforce verification**: Use `--verify` flag during deployment to make verification mandatory
+4. **Distribute public keys securely**: Establish a trusted channel for distributing public keys
+5. **Maintain key inventory**: Keep records of which keys were used to sign which packages
+
+### Operational Security
+
+1. **Never commit private keys**: Exclude private keys from version control
+2. **Audit signature operations**: Log all signing and verification operations
```
AustinAbro321, line 230: What does the recommendation to log signing operations mean to users given that the logging is done by Zarf?

site/src/content/docs/tutorials/5-package-signing-and-verification.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+:::caution
+
+Keep your private key secure and never commit it to version control. Anyone with access to your private key can sign packages as you.
+
+:::
+
+## Step 2: Create and Sign a Package
+
+Now we'll create a package and sign it in one step. Navigate to your package directory (for example, the WordPress package from tutorial 0):
```
AustinAbro321: I think the tutorial should take the choice away from users, and tell them to go to a specific package. Right now it's 50/50 where the wordpress package is an example, but in the verify step it's assumed that it's the one chosen. 

Also instead of saying tutorial 0, I would link to the page. More intuitive that way.

site/src/content/docs/tutorials/5-package-signing-and-verification.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+You should see the package being verified before deployment begins:
+
+```
+Verifying package signature...
+✔ Package signature verified successfully
+Deploying package...
+```
+
+## What You Learned
```
AustinAbro321: This is personal preference, but it feels a bit redundant to summarize what we did in the conclusion here, given how short this tutorial is and given that there is also a summary of the tutorial contents in the introduction paragraph

src/pkg/utils/cosign.go
```diff
@@ -77,19 +81,35 @@ func (opts SignBlobOptions) ShouldSign() bool {
...
+// checkOverwrite validates that output files can be written, returning an error if they exist and Overwrite is false.
+func (opts SignBlobOptions) checkOverwrite(ctx context.Context) error {
+	for _, path := range []string{opts.BundlePath, opts.OutputCertificate} {
+		if path == "" {
+			continue
+		}
+		if _, err := os.Stat(path); err == nil {
+			if !opts.Overwrite {
+				return fmt.Errorf("file at path %s already exists", path)
+			}
```
AustinAbro321, line 93: The overwrite check will never be hit here, because we're passing in a tmp directory that doesn't yet have the bundle files, rather than the current package layout. Currently it works on `zarf package sign` because there's a check on the CLI layer. 

`zarf package publish` doesn't set overwrite to true, but still allows overwriting by default. I believe this in the behavior we want, but it's behaving this way coincidentally.

site/src/content/docs/ref/package-signing.mdx
brandtkeller: I agree - changing language to be less influenced by state of proposal.

site/src/content/docs/ref/package-signing.mdx
brandtkeller: negative - the legacy format is still being generated as of this PR - but I am proposing a shorter timeline to removal (~2 releases).

site/src/content/docs/ref/package-signing.mdx
brandtkeller: This is more of a generic recommendation external to the client. It varies by signing implementation (connected vs disconnected) and what the user may know about the transparency log - but overall is meant in practice to document knowing where signing occurs (execution of Zarf) and how to potentially cross-reference a specific signing event from a given artifact.

site/src/content/docs/tutorials/5-package-signing-and-verification.mdx
brandtkeller: That is fair - moving to wordpress explicitly.

site/src/content/docs/tutorials/5-package-signing-and-verification.mdx
brandtkeller: that is fair - this is more of a personal bias/habit and I realize that the other tutorials do not follow this format.

src/pkg/utils/cosign.go
brandtkeller: good catch - I believe this is still valid utility to the cosign implementation here but have since orchestrated it with the actual filepaths in the `layout` package.

src/pkg/packager/layout/package.go
brandtkeller: Given the other comment - went ahead with enabling overwrite on publish as an expected implicit behavior.

site/src/content/docs/ref/package-signing.mdx
```diff
@@ -0,0 +1,331 @@
...
+
+## Signature Formats
+
+### Sigstore Bundle Format
+
+Packages are signed using the [Sigstore bundle format](https://docs.sigstore.dev/about/bundle/), stored as `zarf.bundle.sig` within the package. This format provides:
+
+- **Self-contained**: Bundles include all verification materials in a single file
+- **Standardized**: Based on the widely-adopted Sigstore specification
+- **Better interoperability**: Compatible with other Sigstore tooling and services
```
AustinAbro321, line 25: The bullet points here have some inconsistency, the first two are adjectives, and read as "this format provides self-contained" and "this format format provides standardized". Maybe a more complete sentence such as "self-contained verification" and "standardized format" would make it read clearer.

site/src/content/docs/ref/package-signing.mdx
AustinAbro321: Ah yeah that makes sense, I think it'd advocate for closer to six releases, that way newer packages are still verifiyable on older versions for around 3 months. Either way once this is merged lets make an issue to track the timeline of it being removed

site/src/content/docs/ref/package-signing.mdx
```diff
@@ -0,0 +1,331 @@
...
+✖ package is already signed. Use --overwrite to replace the existing signature
+```
+
+**Solution**: Use the `--overwrite` flag to replace the existing signature:
+
+```bash
+zarf package sign zarf-package-example-amd64.tar.zst \
+  --signing-key cosign.key \
+  --overwrite
+```
```
AustinAbro321: This message for resolving this error feels intuitive enough where I think it should be cut from this document for brevity

site/src/content/docs/ref/package-signing.mdx
AustinAbro321: What do you think of a line like this, which gives a more actionable ask of users

```
Audit signature operations: Monitor logs to enable traceability to specific package signing events
```

site/src/content/docs/ref/package-signing.mdx
brandtkeller: It does read a little off - updating

site/src/content/docs/ref/package-signing.mdx
brandtkeller: That is reasonable - I was iterating through the error cases and the error itself here is quite self-explanatory.

PR #4520, opened 2026-01-15
docs: update kiwix example and migration guide to use read-only oci volumes

AustinAbro321:
## Description

update kiwix example and migration guide to use read-only oci volumes

## Related Issue

Relates to #3926

## Checklist before merging

### DIFF
--- a/examples/kiwix/zarf.yaml
+++ b/examples/kiwix/zarf.yaml
@@ -41,4 +41,4 @@ x-mdx: |
 
   By utilizing read-only OCI volumes the filesystem of the kiwix-data:local image is directly mounted into the pod.
 
-  To test this example read-only OCI volumes must be enabled in your Kubernetes cluster, and the cluster must be initialized with Zarf version 0.69.1 or greater. OCI volumes are generally available in Kubernetes 1.35.
+  To test this example read-only OCI volumes must be enabled in your Kubernetes cluster, and the cluster must be initialized with Zarf version 0.70.0 or greater. OCI volumes are generally available in Kubernetes 1.35.


examples/kiwix/Dockerfile
```diff
@@ -1,4 +1,4 @@
 # simple base image with shell for copying within the init container
 FROM alpine:3.18
```
a1994sc: If we are mounting this inside the pod, we don't need the shell anymore... should we move this to

```docker
FROM scratch
```

?

AustinAbro321: Yes, Nice catch Allen!

examples/kiwix/zarf.yaml
```diff
@@ -39,4 +39,6 @@ components:
 x-mdx: |
   This example shows how to save data during package create so that it can be loaded into a container running in a pod, in this case to initialize a [Kiwix server](https://www.kiwix.org/en/) to allow offline viewing of documentation and wiki pages.
 
-  By saving the Kiwix data to a container image, we can spin up an init container that will copy the data to the main container running Kiwix.
+  By utilizing read-only OCI volumes the filesystem of the kiwix-data:local image is directly mounted into the pod.
+
+  To test this example read-only OCI volumes must be enabled in your Kubernetes cluster, and the cluster must be initialized with Zarf version 0.69.1 or greater. OCI volumes are generally available in Kubernetes 1.35.
```
brandtkeller: Given the features being shipped in the next release - I'd expect `v0.70.0` to be the entrypoint.

```suggestion
  To test this example read-only OCI volumes must be enabled in your Kubernetes cluster, and the cluster must be initialized with Zarf version 0.70.0 or greater. OCI volumes are generally available in Kubernetes 1.35.
```

brandtkeller: I see the intent for v0.70.0 elsewhere now.

PR #4531, opened 2026-01-20
fix(deploy)!: override actions wait commands

### LINKED ISSUE #4388: bug: namespace override is not applied to component "wait" actions
### Environment
Device and OS: MacOS 26.1
App version: 0.66.0
Kubernetes distro being used: k3d v1.32.5+k3s1
Other:

### Steps to reproduce
1. Deploy zarf package using `-n` to override namespace with a `wait` action in the component

```
kind: ZarfPackageConfig
metadata:
  name: podinfo
  description: Install podinfo
  version: 6.7.0

variables:
  - name: REPLICAS
    default: '1'

components:
  - name: podinfo
    required: true
    charts:
      - name: podinfo
        version: 6.7.0
        namespace: podinfo
        # In this case `url` will load the helm chart located in the podinfo git repository
        url: https://github.com/stefanprodan/podinfo.git
        # By default git will look in the root of the git repository but you can define a sub directory with `gitPath`
        gitPath: charts/podinfo
        valuesFiles:
          - ./values.yaml
    images:
      - curlimages/curl:8.16.0
      - ghcr.io/stefanprodan/podinfo:6.7.0
      - ghcr.io/stefanprodan/podinfo:6.7.0
      - giantswarm/tiny-tools
      - stefanprodan/grpc_health_probe:v0.3.0
      # Cosign artifacts for images - helm-charts - demo-helm-charts
      - ghcr.io/stefanprodan/podinfo:sha256-57a654ace69ec02ba8973093b6a786faa15640575fbf0dbb603db55aca2ccec8.sig
    actions:
      onDeploy:
        before:
          - cmd: "sleep 3"
        after:
          - wait:
              cluster:
                kind: deployment
                name: podinfo
                namespace: podinfo
                condition: "{.status.readyReplicas}=${ZARF_VAR_REPLICAS}"
          - cmd: "sleep 3"

```

### Expected result
package should install in overriden namespace and wait should validate it exists

### Actual Result
package deploys in namespace but wait still looks for it in the namespace in the zarf.yaml not the override namespace

### Additional Context

brandtkeller:
## Description

Currently namespace overrides only detects and/or modifies charts and manifests. This PR adds support for actions wait definitions.

## Related Issue

Fixes #4388

## Checklist before merging

### DIFF
--- a/src/api/v1alpha1/package.go
+++ b/src/api/v1alpha1/package.go
@@ -111,14 +111,10 @@ func (pkg ZarfPackage) UniqueNamespaces() []string {
 	uniqueNamespaces := make(map[string]struct{})
 	for _, component := range pkg.Components {
 		for _, chart := range component.Charts {
-			if chart.Namespace != "" {
-				uniqueNamespaces[chart.Namespace] = struct{}{}
-			}
+			uniqueNamespaces[chart.Namespace] = struct{}{}
 		}
 		for _, manifest := range component.Manifests {
-			if manifest.Namespace != "" {
-				uniqueNamespaces[manifest.Namespace] = struct{}{}
-			}
+			uniqueNamespaces[manifest.Namespace] = struct{}{}
 		}
 	}
 	return slices.Collect(maps.Keys(uniqueNamespaces))
--- a/src/pkg/packager/packager.go
+++ b/src/pkg/packager/packager.go
@@ -120,13 +120,13 @@ func overrideComponentNamespaces(pkg *v1alpha1.ZarfPackage, original, target str
 				pkg.Components[i].Manifests[k].Namespace = target
 			}
 		}
-		overrideActionSetWaitNamespaces(pkg.Components[i].Actions.OnCreate, original, target)
-		overrideActionSetWaitNamespaces(pkg.Components[i].Actions.OnDeploy, original, target)
-		overrideActionSetWaitNamespaces(pkg.Components[i].Actions.OnRemove, original, target)
+		overrideActionSetWaitNamespaces(&pkg.Components[i].Actions.OnCreate, original, target)
+		overrideActionSetWaitNamespaces(&pkg.Components[i].Actions.OnDeploy, original, target)
+		overrideActionSetWaitNamespaces(&pkg.Components[i].Actions.OnRemove, original, target)
 	}
 }
 
-func overrideActionSetWaitNamespaces(set v1alpha1.ZarfComponentActionSet, original, target string) {
+func overrideActionSetWaitNamespaces(set *v1alpha1.ZarfComponentActionSet, original, target string) {
 	overrideActionWaitNamespaces(set.Before, original, target)
 	overrideActionWaitNamespaces(set.After, original, target)
 	overrideActionWaitNamespaces(set.OnSuccess, original, target)
--- a/src/pkg/packager/packager_test.go
+++ b/src/pkg/packager/packager_test.go
@@ -190,7 +190,8 @@ func TestOverridePackageNamespace(t *testing.T) {
 					},
 				},
 			},
-			namespace: "test-override",
+			namespace:   "test-override",
+			expectedErr: "package contains 2 unique namespaces, cannot override namespace",
 		},
 		{
 			name: "init package namespace override",
@@ -252,14 +253,10 @@ func validateNamespaceUpdates(t *testing.T, pkg v1alpha1.ZarfPackage, targetName
 	actualWaitNamespaces := make([]string, 0)
 	for _, component := range pkg.Components {
 		for _, chart := range component.Charts {
-			if chart.Namespace != "" {
-				require.Equal(t, targetNamespace, chart.Namespace)
-			}
+			require.Equal(t, targetNamespace, chart.Namespace)
 		}
 		for _, manifest := range component.Manifests {
-			if manifest.Namespace != "" {
-				require.Equal(t, targetNamespace, manifest.Namespace)
-			}
+			require.Equal(t, targetNamespace, manifest.Namespace)
 		}
 		actualWaitNamespaces = append(actualWaitNamespaces, collectWaitNamespaces(component.Actions)...)
 	}


src/api/v1alpha1/package.go
```diff
@@ -109,6 +109,11 @@ func (pkg ZarfPackage) UniqueNamespaceCount() int {
 		for _, manifest := range component.Manifests {
 			uniqueNamespaces[manifest.Namespace] = struct{}{}
 		}
+		for _, action := range component.Actions.GetAll() {
+			if action.Wait != nil && action.Wait.Cluster != nil && action.Wait.Cluster.Namespace != "" {
+				uniqueNamespaces[action.Wait.Cluster.Namespace] = struct{}{}
+			}
+		}
```
AustinAbro321: I don't think wait namespaces should count as unique namespaces. I think we should update the namespace when it matches the original package namespace, but not count the package as having multiple namespaces when the wait condition has a new namespace.  

I'm picturing a package that can be deployed into any namespace, but it has to check for the existence of a different resource before it deploys, maybe flux-system. We don't want to mutate that namespace check, but we would want that package deployable anywhere.

brandtkeller: That is a great point and I don't have any hard opinions against that option - let me give that a run.

brandtkeller: quick prototype - the only downside here is that the original solution did not need to be namespace aware. The associated methods for updating namespaces do so without any awareness of the previous value. In order to support this we need to determine the namespace we want to overwrite. 

I'm going to pose some slightly duplicative logic for sake of feeling this out further.

src/api/v1alpha1/package.go
```diff
@@ -113,7 +115,21 @@ func (pkg ZarfPackage) UniqueNamespaceCount() int {
...
+			uniqueNamespaces[chart.Namespace] = struct{}{}
+		}
+		for _, manifest := range component.Manifests {
+			uniqueNamespaces[manifest.Namespace] = struct{}{}
+		}
+	}
+	return slices.Collect(maps.Keys(uniqueNamespaces))
+}
+
+// UpdateAllComponentNamespaces updates all existing chart/manifest namespaces to the provided one
```
AustinAbro321: Given that we're taking the stance that it's a bug that wait actions are not updated. I'd be in favor of either removing it or adding "original" as a parameter instead of a separate "UpdateAllComponentNamespacesByName". 

Another option to avoid a breaking change would to deprecate this function and point to UpdateAllComponentNamespacesByName

src/api/v1alpha1/package.go
```diff
@@ -126,6 +142,25 @@ func (pkg ZarfPackage) UpdateAllComponentNamespaces(namespace string) {
 	}
 }
 
+// UpdateAllComponentNamespacesByName updates all matching namespaces to the provided one
+func (pkg ZarfPackage) UpdateAllComponentNamespacesByName(original, target string) {
```
AustinAbro321: I feel that pkg/packager may be a better place for this function. We're de-coupling the function from the logic of namespace overrides. For instance, charts or manifest namespaces won't be updated if they don't equal the original. In practice this won't ever happen, though I can see why this might be added to this function given that it's in the data layer. 

Still, this feels like a code smell to me, I think we might be better served by moving this function to the logic layer (pkg/packager), given that it's use case is tightly coupled to the namespace overrides logic. Having the function at this layer will also reduce duplication as we introduce new API versions.

AustinAbro321: @brandtkeller curious on your thoughts here. I don't think there's a flow where we'd want to encourage users to run this function outside of deploy with namespace overrides set. For instance, an SDK user, might see this function, call it on their package, then call `packager.Deploy` without setting namespace overrides. In this case, they'd miss the validation, and the secret features of namespace overrides.  

Ultimately, I think it's best to keep the schema object as a pure data object, instead of also having functions that mutate it. Looking through this and `UpdateAllComponentNamespaces` are the only mutating functions.  I feel that mutations will be better served being near the surrounding logic for their mutation.

brandtkeller: You make a good point - I need to revisit the model overall. I had some hesitations about moving mutation outside of the data object with regards to potential deviations across schemas. Let me consider further.

src/api/v1alpha1/package.go
```diff
@@ -101,29 +103,25 @@ func (pkg ZarfPackage) IsSBOMAble() bool {
...
+	return len(pkg.UniqueNamespaces())
+}
+
+// UniqueNamespaces returns a slice of all unique namespaces in the package
+func (pkg ZarfPackage) UniqueNamespaces() []string {
 	uniqueNamespaces := make(map[string]struct{})
 	for _, component := range pkg.Components {
 		for _, chart := range component.Charts {
-			uniqueNamespaces[chart.Namespace] = struct{}{}
+			if chart.Namespace != "" {
```
brandtkeller: note: given a package with one distinct namespace defined and then other undefined namespaces on a chart or manifest - we will not update elements with empty namespaces to the new target namespace (at least how this is currently). Worth considering this edge case and whether we want to update empty to the target.

src/pkg/packager/packager.go
```diff
@@ -86,17 +86,57 @@ func generateValuesOverrides(_ context.Context, chart v1alpha1.ZarfChart, compon
 }
 
 // OverridePackageNamespace overrides the package namespace if the package contains only one unique namespace
-func OverridePackageNamespace(pkg v1alpha1.ZarfPackage, namespace string) error {
+func OverridePackageNamespace(pkg *v1alpha1.ZarfPackage, namespace string) error {
```
brandtkeller, line 89: I think it might be worth updating this signature (as is currently proposed). It is currently working as expected because the struct fields are slices (which share their backing arrays). Making this behavior explicit leaves us less likely to continue the pattern.

src/api/v1alpha1/package.go
brandtkeller: Updated.

src/api/v1alpha1/package.go
AustinAbro321: Having one namespace defined and one namespace empty blocks namespace overrides, but this PR starts to allow that. Is this intentional? I'd lean towards keeping all namespaces as the same a strict requirement including empty, but maybe there is a use case I'm not thinking of. 

You cannot have an empty namespace on charts because of validation, but you can have an empty namespace for manifests

src/pkg/packager/packager.go
AustinAbro321: Agreed, you'll also want to change this on the other mutation functions `overrideActionSetWaitNamespaces`

src/api/v1alpha1/package.go
brandtkeller: Yes intentional, I see two distinct paths for opinionation here and both have pros/cons. 

If we treat empty namespace (which is permitted across the zarf manifest) as a distinct/singular namespace in the namespace count - then the override rule is more straight-forward and blocks the empty/defined scenario you mentioned. 

If we ignore empty then we have a slightly less strict stance but potentially more optionality. I have concerns about cluster-scoped resources that won't have a namespace and thus a package (say an operator) that reads those objects (and contains them in the package) would be blocked from the use of overrides.

AustinAbro321: To me, the main reason to avoid specifying the namespace are: 
- if your manifests have explicit differing namespaces, so specifying namespace doesn't have an effect
- If you only have cluster scoped resources where a namespace is misleading.

These would not work with namespace overrides as you'd have duplicate resources across packages. Barring a solid use case, I'd prefer to keep empty namespaces in the distinct namespace count. I feel that changing this is scope creep for this PR also.

brandtkeller: I think exposing the change in behavior is appropriate to considering this out of scope. I have concerns about what this will mean for cluster scoped resources but the use case is built more on assumptions. Let me revert this behavior.

src/pkg/packager/packager.go
brandtkeller: good call - updated

PR #4540, opened 2026-01-21
feat(load): support explicit VerificationStrategy

### LINKED ISSUE #4539: Support VerificationStrategy for packager.LoadOptions

If SDK consumers want to orchestrate signature verification under their own parameters - then they currently have to set `Verify: false` and then perform their own verification. 

Given Zarfs default `VerificationStrategy` of `VerifyIfPossible` - this creates a redundant check whereby verification is tried - warnings logged - and then the sdk consumer can orchestrate their intent. 

- **Given** The SDK is being used for loading packages
- **When** `packager.LoadPackage` is executed
- **Then** options are available to explicitly set VerificationStrategy to `VerifyNever`

Allow unsigned packages during verify to log a warning and not an error.

brandtkeller:
## Description

This adds `VerificationStrategy` passthrough for `layout.PackageLayoutOptions` such that a Zarf SDK consumer can explicitly opt-in to setting `Verify: false` and `VerificationStrategy: layout.VerifyNever` in the event they want to implement their own verification workflows without redundant processing and logging. 

## Related Issue

Fixes #4539 

## Checklist before merging

### DIFF
--- a/src/cmd/initialize.go
+++ b/src/cmd/initialize.go
@@ -21,7 +21,6 @@ import (
 	"github.com/zarf-dev/zarf/src/pkg/logger"
 	"github.com/zarf-dev/zarf/src/pkg/packager"
 	"github.com/zarf-dev/zarf/src/pkg/packager/filters"
-	"github.com/zarf-dev/zarf/src/pkg/packager/layout"
 	"github.com/zarf-dev/zarf/src/pkg/state"
 	"github.com/zarf-dev/zarf/src/pkg/utils"
 	"github.com/zarf-dev/zarf/src/pkg/zoci"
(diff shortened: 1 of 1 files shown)


src/pkg/packager/load.go
```diff
@@ -43,6 +43,8 @@ type LoadOptions struct {
 	CachePath string
 	// Only applicable to OCI + HTTP
 	RemoteOptions
+	// VerificationStrategy for explicit definition
+	layout.VerificationStrategy
```
AustinAbro321, line 46: IMO if we add this then we should take away the boolean option, it is a breaking change, but I think it is confusing to have both

brandtkeller: I agree - I was tempted to head that direction but always tend to default to playing conservative. Let me draft that now.

brandtkeller: Updated

src/cmd/package.go
```diff
@@ -330,15 +330,21 @@ func (o *packageDeployOptions) run(cmd *cobra.Command, args []string) (err error
 		return err
 	}
 
+	// Establish default stance
+	verificationStrategy := layout.VerifyIfPossible
```
AustinAbro321: Could we make a common function `getVerfificationStrategy(verify bool)` to encapsulate this logic across src/cmd

brandtkeller: added - good call

PR #4541, opened 2026-01-22
fix(helm): preserve block scalar semantics

### LINKED ISSUE #4537: Configmaps Missing Trailing Newlines
### Environment
App version: v0.63.0+

### Steps to reproduce
1. Add a configmap in a helm chart
2. Deploy this helm chart through zarf
3. Inspect the deployed configmap (i.e. kubectl get cm my-configmap -o json)

### Expected result
Configmap values should end in a newline

### Actual Result
Configmap value has the trailing newline chomped

Consider the following configmap in a helm chart:
```
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: needs-new-line
data:
  mykey: |
    this value should have a newline at the end

```

When deployed directly via helm v3.19.4, I see the following when inspecting:

```
kubectl get cm -n default needs-new-line -o json
...
    "data": {
        "mykey": "this value should have a newline at the end\n"
    },
...
```

However, if this helm chart is deployed via zarf, inspecting shows this cm is stored differently:

```
kubectl get cm -n default needs-new-line -o json
...
        "mykey": "this value should have a newline at the end"
...
```

### Additional Context
This seems to be the case for any configmap in any helm chart. A simple zarf.yaml I've been using to debug has been:
```
kind: ZarfPackageConfig
metadata:
  name: newline-repro
components:
  - name: chart-test
    required: true
    charts:
      - name: my-chart
        version: 1.0.0
        namespace: default
        localPath: ./my-chart
```

The behavior starts in zarf v0.63.0 (that is, zarf v0.62.0 is the last one where the new line is present in the json output). As far as I can tell, v0.63.0 and v0.62.0 use the same helm version underneath (zarf tools helm version shows v3.18.4). 
This is pure speculation, but is there a zarf mechanism that renders helm templates before deploying that might be dropping the trailing newline?

brandtkeller:
## Description

v0.63.0 introduced new behaviors to the post-render process for helm charts whereby:
- Helm's releaseutil.SortManifests strips trailing newlines from resource.Content (this has always been the case)
- When YAML with a block scalar (data: |) is parsed without a trailing document newline, the parser doesn't preserve the trailing newline in the scalar's value
- When re-marshaled, the library outputs data: |- (with chomp indicator), which explicitly removes the trailing newline

This change ensures we preserve the trailing newline. 

## Related Issue

Fixes #4537

## Checklist before merging

### DIFF
--- a/site/src/content/docs/roadmap.mdx
+++ b/site/src/content/docs/roadmap.mdx
@@ -1,57 +1,42 @@
 ---
-title: Zarf 2025 Roadmap
+title: Zarf 2026 Roadmap
 ---
 
 ## Introduction
 
-2025 is about building resilience, empowering extensibility, and preparing the system for broader integrations through SDKs.
-
-## Core Themes for 2025
-
-- **SDK Support & Decoupled Architecture**
-  Enable integration and extensibility by refactoring the core to cleanly separate CLI operations from library logic. This will unlock use cases for tooling integration, programmatic workflows, and community enhancements.
-
-- **Resilient-by-Design Infrastructure**
-  Invest in the internal consistency, recovery, and upgrade mechanisms of Zarf to ensure performance across edge and connected environments.
-
-- **Operational Enhancements**
-  Improve package lifecycle operations, diagnostics, and user observability to make Zarf more usable and reliable in production.
-
-- **Workflow Stability & Expansion**
-  Mature the package lifecycle, extend schema-based definitions, and reduce the learning curve for new users through simplification and documentation.
-
----
+2026 is about building long-term sustainability, provenance, and enhancing the package lifecycle from creation through deployment with enhanced state.
 
 ## Development Story
 
-### SDK Readiness (Q1 - Q2)
+### Sustainability and Provenance (Q1 - Q2)
 
-We began the year with a deliberate push to lay the groundwork for SDK support. This involves:
+Support the advancement of the project with schema updates, new features, and required deprecations/removals. Additionally establishing parity with the signing and verification ecosystem that aligns natively to the airgapped environments through offline verification:
 
-- Refactor internal logic to split the CLI from the shared library.
-- Establish SDK entry points for Zarf functionality.
-- Standardize on logging patterns for package lifecycle execution.
+- Introduce next schema version, improve structure, and identify deprecations removals
+- Zarf Package Signing and Verification Enhancements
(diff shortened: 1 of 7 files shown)


AustinAbro321:
@brandtkeller Did you find the PR or dependency update that changed this behavior?

brandtkeller:
> Did you find the PR or dependency update that changed this behavior?

#4194 introduced an additional marshal into the `editHelmResources` which didn't happen previously and leads to the sequence in the description.

AustinAbro321:
Could you add a test for this? E2E or unit if you can make that work

AustinAbro321:
What do you think of using kyaml "sigs.k8s.io/kustomize/kyaml/yaml" here instead. I tested it out and it seems to solve this problem in a cleaner way. Though I have not run the full e2e test suite with it https://kubernetes.io/docs/reference/encodings/kyaml/

brandtkeller:
> What do you think of using kyaml "sigs.k8s.io/kustomize/kyaml/yaml" here instead. I tested it out and it seems to solve this problem in a cleaner way. Though I have not run the full e2e test suite with it https://kubernetes.io/docs/reference/encodings/kyaml/

I'll get a test written to serve as a baseline and then look at swapping that dependency - I should have put this in draft as it was end-of-day and not complete (mainly wanted CI).

AustinAbro321:
Great! I rescind my recommendation to switch to  "sigs.k8s.io/kustomize/kyaml/yaml", I was conflating kustomize yaml with kyaml. The latter seems to be more of a tool for printing clearly rather than marshalling and unmarshalling

brandtkeller:
> Great! I rescind my recommendation to switch to "sigs.k8s.io/kustomize/kyaml/yaml", I was conflating kustomize yaml with kyaml. The latter seems to be more of a tool for printing clearly rather than marshalling and unmarshalling

Composed the logic into a new private function to make it easier to test (initial and regression). You can see the test fail or pass based on the existence of  re-adding the newline. 

I don't see any changes in behavior across yaml parsers but do largely believe that this is a combination of the newline being stripped prior to unmarshall/marshall. I see kyaml as an example of having [similarly](https://github.com/kubernetes-sigs/kustomize/pull/3431/changes#diff-798261e2e8abbbcadd63221b66ebc03f95221bbbffb0c6d1c1a3a9d583b12b4bR154-R156) implemented a check for re-adding the newline.

PR #4543, opened 2026-01-22
feat: move value pkg out of internal

mkcp:
## Description
This PR moves the `internal/value` package to `pkg/value` to make it publicly available to users. Because the API and datatypes in the `value` pkg have been stable for most of its development, and it is intended as a public API on release, it's time to make it available to SDK users. There's still the possibility for breaking changes, but the impact of those changes should be minor overall e.g. renames or additions rather than massive changes to the `value.Values` type and methods or `value.Path`.

This PR also updates some imports for style and consistency. 

## Related Issue
Relates to #3946 

## Checklist before merging


brandtkeller (approved):
lgtm

mkcp:
cc @michalszynkiewicz

PR #4547, opened 2026-01-22
feat: error early during healthchecks when status is terminal

AustinAbro321:
## Description

Helm recently made this improvement which made me realize this was also sub-optimal behavior in Zarf. https://github.com/helm/helm/pull/31730. 

## Checklist before merging

### DIFF
--- a/src/internal/healthchecks/healthchecks.go
+++ b/src/internal/healthchecks/healthchecks.go
@@ -97,7 +97,7 @@ func WaitForReady(ctx context.Context, sw watcher.StatusWatcher, objs []object.O
 	for _, id := range objs {
 		rs := statusCollector.ResourceStatuses[id]
 		if rs.Status != status.CurrentStatus {
-			errs = append(errs, fmt.Errorf("%s: %s not ready, status is %s", rs.Identifier.Name, rs.Identifier.GroupKind.Kind, rs.Status))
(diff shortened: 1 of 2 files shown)


src/internal/healthchecks/healthchecks.go
```diff
@@ -87,20 +93,18 @@ func WaitForReady(ctx context.Context, sw watcher.StatusWatcher, objs []object.O
...
-		for _, id := range objs {
-			rs := statusCollector.ResourceStatuses[id]
-			if rs.Status != status.CurrentStatus {
-				errs = append(errs, fmt.Errorf("%s: %s not ready, status is %s", rs.Identifier.Name, rs.Identifier.GroupKind.Kind, rs.Status))
-			}
+	errs := []error{}
+	for _, id := range objs {
+		rs := statusCollector.ResourceStatuses[id]
+		if rs.Status != status.CurrentStatus {
+			errs = append(errs, fmt.Errorf("%s: %s not ready, status is %s", rs.Identifier.Name, rs.Identifier.GroupKind.Kind, rs.Status))
```
brandtkeller: Thinking about the FailedStatus and looking at the `ResourceStatus` I wonder if there would be a benefit to also including `rs.Message` in this log?

AustinAbro321: Great idea! updated

PR #4549, opened 2026-01-23
refactor(wait)!: create wait package and call it directly within actions

AustinAbro321:
## Breaking changes:
- This deletes utils.ExecuteWait, in favor of two new functions in the new package src/pkg/wait. `wait.ForResource` and `wait.ForNetwork`. These functions no longer have both the parameter `waitTimeout string` and `timeout time.Duration` and instead only accept `timeout time.Duration`

## Description

`.actions.wait` currently calls shells out to `zarf tools wait-for`. It makes more sense to call the wait functions directly. This makes us less reliant on shelling out (though it's still necessary for now when `zarf tools kubectl wait` is called by `utils.ExecuteWait()`. For now...

New e2e test takes ~6 seconds 

Relates to #4550 

## Checklist before merging

### DIFF
--- a/.github/workflows/scan-codeql.yml
+++ b/.github/workflows/scan-codeql.yml
@@ -45,7 +45,7 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@19b2f06db2b6f5108140aeb04014ef02b648f789 # v4.31.11
+        uses: github/codeql-action/init@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v4.32.0
         with:
           languages: ${{ matrix.language }}
           config-file: ./.github/codeql.yaml
@@ -54,6 +54,6 @@ jobs:
         run: make build-cli-linux-amd
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@19b2f06db2b6f5108140aeb04014ef02b648f789 # v4.31.11
+        uses: github/codeql-action/analyze@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v4.32.0
         with:
           category: "/language:${{matrix.language}}"
--- a/.github/workflows/scorecard.yaml
+++ b/.github/workflows/scorecard.yaml
@@ -44,6 +44,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@19b2f06db2b6f5108140aeb04014ef02b648f789 # v4.31.11
+        uses: github/codeql-action/upload-sarif@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v4.32.0
         with:
           sarif_file: results.sarif
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,6 +86,10 @@ The Technical Steering Committee (the "TSC") will be responsible for all technic
 Affiliation: Defense Unicorns
 GitHub: @AustinAbro321
 
+#### Brandt Keller
+Affiliation: Defense Unicorns
+GitHub: @brandtkeller
+
 #### Danny Gershman
 Affiliation: Radius Method
 GitHub: @dgershman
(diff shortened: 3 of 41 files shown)


src/pkg/packager/actions/actions.go
```diff
@@ -191,40 +174,58 @@ retryCmd:
...
+	timeout := 5 * time.Minute
+	if action.MaxTotalSeconds != nil && *action.MaxTotalSeconds > 0 {
+		timeout = time.Duration(*action.MaxTotalSeconds) * time.Second
 	}
 
-	network := wait.Network
-	if network != nil {
-		// Make sure the protocol is lower case.
-		network.Protocol = strings.ToLower(network.Protocol)
+	if waitCfg.Cluster != nil {
```
brandtkeller: nit: given these are mutually exclusive I would typically reach for a switch.

src/pkg/wait/wait.go
brandtkeller: Not included in your changes but noticed this when tracing the full path - `errors.New()` does not accept formatting.

src/test/e2e/46_tools_wait_for_test.go
```diff
@@ -0,0 +1,126 @@
...
+		t.Cleanup(func() {
+			_, _, err := e2e.Kubectl(t, "delete", "pod", podName, "-n", namespace, "--force=true", "--grace-period=0")
+			require.NoError(t, err)
+		})
+
+		stdOut, stdErr, err := e2e.Zarf(t, "tools", "wait-for", "pod", podName, "{.status.phase}=Running", "-n", namespace, "--timeout", "60s")
+		require.NoError(t, err, stdOut, stdErr)
+	})
+
+	t.Run("wait for resource by by kind", func(t *testing.T) {
```
brandtkeller:
```suggestion
	t.Run("wait for resource by kind", func(t *testing.T) {
```

src/test/e2e/46_tools_wait_for_test.go
```diff
@@ -0,0 +1,126 @@
...
+		stdOut, stdErr, err := e2e.Zarf(t, "tools", "wait-for", "storageclass", "--timeout", "10s")
+		require.NoError(t, err, stdOut, stdErr)
+	})
+
+	t.Run("wait for pod created after wait starts", func(t *testing.T) {
+		podName := "delayed-pod"
+
+		// Start waiting for the pod in a goroutine before it exists
```
brandtkeller, line 103: observation: sometimes sleep-based synchronization can be flaky and this test also doesn't have a cancellation mechanism. 

Both are pretty minor given timeouts but there may be room to signal that the goroutine has started (channel) and maybe passing context to e2e.Zarf (or adding another function with context).

AustinAbro321: Are you thinking about something like this?
```go
  errCh := make(chan error, 1)
  started := make(chan struct{})
  go func() {                                                                                                                                                     
      close(started) 
      _, _, err := e2e.Zarf(t, "tools", "wait-for", "pod", podName, "ready", "-n", namespace, "--timeout", "60s")                                                 
      errCh <- err
  }()                                                                                                                                              
  <-started
```
The issue with this strategy, is that I would still want a wait because I wouldn't have confidence that a full wait cycle has run before the pod exists in the cluster, given that we do other calls prior to `zarf tools kubectl wait-for`. 

Technically we could make another function to pass context to e2e.Zarf, but since exec doesn't accept context, I don't know if there would be any benefit to doing so

brandtkeller: certainly not blockers here:

For sleep synchronization - I retract my statement - there area a number of edges here and custom wait logic in testing to test the wait-fair is a little tough and introduces more complexity that would be useful. 

For `context` - the existing zarf `exec` package has `CmdWithContext()` that accepts a context already. Currently `CmdWithtesting` orchestrates the context and cancel when the test ends but not mid-test.  

meaning I see two potential updates of minor value:
1. `CmdWithTesting` should use t.Context()` instead of `context.Background()`
2. a `ZarfWithContext` command that accepts a custom context would still be reasonable for context cancellation.

AustinAbro321: Made a follow up issue https://github.com/zarf-dev/zarf/issues/4583

Issue #4509, opened 2026-01-09
Images within Image Archives should be found when using `zarf dev find-images`


AustinAbro321:
### Is your feature request related to a problem? Please describe.
Currently images in [Image Archives](https://docs.zarf.dev/ref/components/#container-image-archives) do not show up in `zarf dev find-images`. Assuming there is a valid path to a tar file, Zarf should show all of the images in it in the output of `zarf dev find-images`.

### Describe the behavior you'd like

- **Given** a component with an image archive. The image archive was made with the image kiwix-data:local
```yaml
  - name: kiwix-serve
    imageArchives:
      - path: kiwix-data.tar
```
- **When** I run `zarf dev find-images`
- **Then** I receive an output that looks like this:
``` 
components:
 - name: kiwix-serve
    imageArchives:
      - path: kiwix-data.tar
        images:
          - kiwix-data:local
```

### Additional context
The implementor of this will run into a problem where `imageArchives.images` is required by the schema. In order to achieve this we may have to arbitrarily set the images field if it empty before loading the definition in   `packager.FindImages()`

chaospuppy:
In my reading of how things currently work, we will need to update `PackageDefinition` to set `images: []` before we get to validation.  Is this what you had in mind, or is there somewhere earlier that I can set `imageArchives.images` if empty?

AustinAbro321:
@chaospuppy find-images is really the only scenario where we'd want this field to be empty (because a user might still be figuring out using this command). We'd want to localize setting the field in `packager.FindImages()`. In that function when imageArchives.Images is empty then we could add a single entry in the list called `placeholder` to allow it to pass the schema check.

chaospuppy:
@AustinAbro321 Would that then necessitate loading the `zarf.yaml` twice?  Once to check for and set-if-missing `imageArchives.images` in `FindImages`, and again in `load.PackageDefinition`?

AustinAbro321:
Ah yeah, I forgot that loading and validating happens at the same time. While we could make a temporary zarf.yaml and change that, probably the best way to go about this would be to introduce a new public function `load.PackageDefinitionWithoutValidate()` and then also make the currently private function `load.validate()` public. Then update the struct between those calls.

chaospuppy:
Perfect, that makes sense to me.  Thanks!

chaospuppy:
[...]

PR #4551, opened 2026-01-23
feat(find-images): 4509 include archives in find image

### LINKED ISSUE #4577: zarf dev find-images should place images within Image Archives when archives hold relevant images
When you run `zarf dev find-images`, it doesn't take into account that some images might be from image archives. For instance, if `zarf dev find-images` is run on the examples/kiwix package, then a user will receive the following output:

```bash
components:
  - name: kiwix-serve
    images:
      - ghcr.io/kiwix/kiwix-serve:3.5.0-2
      - kiwix-data:local
```

However, given that there is an image archive, if you were to copy the images block into your component, or if you were to run `zarf dev find-images --update`, you would place images into your images block, that you were already bringing in through your image archive block. 

Instead, `zarf dev find-images` should have knowledge of which images are in an archive defined in the package. If an image is found in a manifest that's within an image archive, instead of outputting the image in the images block, it should be outputted in the image archive block like so: 

```bash
components:
  - name: kiwix-serve
    images:
      - ghcr.io/kiwix/kiwix-serve:3.5.0-2
    imageArchives:
      - path: kiwix-data.tar
        images:
          - kiwix-data:local
```

Notably, we should not include all images in the image archives block unless they are in a manifest

This should also work for `zarf dev find-images --update`.

Initially I created the issue https://github.com/zarf-dev/zarf/issues/4509, however I realized that the strategy in this issue didn't fit well with the `zarf dev find-images` command. Given that it looks for images through manifests, flags like `--why` and `--update` didn't fit well with the proposed behavior in that issue

chaospuppy:
## Description
This MR resolves #4577 by doing the following:
- Allow schema validation to succeed by removing findings related to an empty image list under `imageArchives` keys.  Currently this finding is only removed when calling PackageDefinition from FindImages
- Add FindDefinitionImages to deduplicate `images` that are provided by `imageArchives`
- Update unit tests

## Related Issue
Fixes #4577

## Checklist before merging


chaospuppy:
@AustinAbro321 when you have a moment, please take a look.  I am still going to update a few unit tests.

src/pkg/images/unpack.go
```diff
@@ -36,6 +36,67 @@ const (
 	dockerContainerdImageStoreAnnotation = "containerd.io/distribution.source.docker.io"
 )
 
+// FindImagesInArchive takes an image tar and extracts it into an OCI layout directory.
+// It returns a list of images found in the extracted OCI layout.
+func FindImagesInArchive(ctx context.Context, imageArchive, destDir string) ([]string, error) {
+	// Create a temporary directory for extraction
+	tmpdir, err := utils.MakeTempDir("")
```
AustinAbro321: The extraction logic in this block is mostly duplicated. Lets make a common function to de-duplicate it

src/pkg/packager/load/load.go
```diff
@@ -37,6 +39,8 @@ type DefinitionOptions struct {
 	IsInteractive bool
 	// SkipVersionCheck skips version requirement validation
 	SkipVersionCheck bool
+	// SkipImageArchivesImages ignores schema validation errors when imageArchives does not include an images list
+	SkipEmptyImageArchivesImages bool
```
AustinAbro321: I'd prefer a strategy where we create a function `load.PackageDefinitionNoValidate()` then also make a public function `load.Validate()`. Keep the load.PackageDefinition function this way validation happens by default.

 I believe a public validate function could be generally useful for SDK users who might be creating packages without zarf.yaml files, additionally it'll leave more room in the future for more generic solutions if we need to skip validation for a similar reason

src/pkg/packager/find_images.go
```diff
@@ -225,6 +226,22 @@ func FindImages(ctx context.Context, packagePath string, opts FindImagesOptions)
...
+			if err != nil {
+				return nil, err
+			}
+			archivePath := pkgPath.BaseDir + archive.Path
+			archiveImages, err := images.FindImagesInArchive(ctx, archivePath, tmpArchivePath)
+			if err != nil {
+				return nil, fmt.Errorf("failed to unpack image archive %s: %w", archive.Path, err)
+			}
+			for _, image := range archiveImages {
+				matchedImages[image] = true
```
AustinAbro321: I think matched images isn't the right construct. Maybe a new field should be added such as `archiveImages`. One of the goals of find-images is to be able to copy and paste the output into your zarf.yaml. Currently you'd see this output, which would result in trying to pull the image from the internet if you copied and pasted it
```yaml
2026-01-27 16:38:42 INF looking up cosign artifacts for discovered images count=3

components:
  - name: kiwix-serve
    images:
      - docker.io/library/kiwix-data:local
      - ghcr.io/kiwix/kiwix-serve:3.5.0-2
      - kiwix-data:local
```

src/pkg/packager/find_images.go
AustinAbro321, line 1: Image archives functionality should also work with packager.UpdateImages

AustinAbro321 (changes requested):
Good start! Thanks for taking this on

src/pkg/packager/load/load.go
chaospuppy: I'm happy to make that split still, but our failure was while validating the file itself [here](https://github.com/zarf-dev/zarf/blob/eba0280d293c1a1ba99b299504550a543ae6132e/src/pkg/packager/load/load.go#L106), rather than during the validation of `v1alpha1.ZarfPackage`.  Therefore, it wasn't useful for us to delay validation after adding a `placeholder` image as we previously discussed to the `v1alpha1.ZarfPackage`.

I could split them and then add some work to change the `zarf.ymal` on disk to add an empty array there, but that didn't seem like desirable behavior.

src/pkg/packager/find_images.go
chaospuppy: I see, good point, I missed the desire to pull images from the image archive on disk after `zarf.yaml` has been updated with the output of `zarf dev find-images`.  I'll fix that up!

src/pkg/packager/load/load.go
chaospuppy: Alternatively, I could also split out a `Validate` and `ValidateWithoutSchemaValidation` (not literally that name, but something like that) and use the later

AustinAbro321:
Hello @chaospuppy, I reazlied that the behavior I proposed in #4509, didn't really fit the `zarf dev find-images` command. However, there is still work to be done to get image archives to behave properly for `zarf dev find-images`. I've created https://github.com/zarf-dev/zarf/issues/4577 to track this work. If you're interested, I believe this PR could be adjusted  for that issue. LMK if you have any questions. 

If you do adjust this PR for https://github.com/zarf-dev/zarf/issues/4577, don't worry about validation, assume the given zarf.yaml file is valid

src/pkg/packager/find_images.go
chaospuppy: @AustinAbro321 to clarify, are you hoping to see `images` elements expressed in a such as way as to direct Zarf to pull them from the archive, rather than from the internet?

src/pkg/images/unpack.go
chaospuppy: Made a few changes with this in mind

src/pkg/packager/find_images.go
chaospuppy: Ah I missed your comment before asking, I think I see the answer in #4577

AustinAbro321: Just went back and clarified #4577 a bit, but just to write here as well, we should no longer include images because they're in an image archive. If they're in both a manifest we should check if the image archive contains that image, then in the output to the user structure it so it's in the image archive, and the user can simply copy and paste.

chaospuppy:
@AustinAbro321 Are you thinking about handling validation differently in the near future?  Trying to get an understanding for which changes I should back out.  I am also happy to keep the validation changes in, including making Validation public and creating PackageDefinitionWithoutValidation.

AustinAbro321:
@chaospuppy Very possibly handling validation in the near future yes. I suggest backing out of the validation changes in this PR

src/pkg/packager/find_images.go
```diff
@@ -71,17 +73,21 @@ type FindImagesOptions struct {
...
 	// ComponentName is the name of the component where the images were found
 	ComponentName string
+	// ArchivePath is the path to the imageArchive
+	Path string
 	// Matches contains definitively identified container images, such as those in an image: field
 	Matches []string
 	// PotentialMatches contains potential container images found by a regex
 	PotentialMatches []string
+	// ArchiveImages contains images that are included in Matches, but are provided with an imageArchive
+	ArchiveImages []string
```
AustinAbro321: Not yet doing a full review, as this is in draft, but thinking about the architecture here. I believe we shouldn't change anything in the `packager.FindImages` function in this PR. This function should be focused on finding images from manifests and charts. The logic for determining which of those images are in an image archive so we can customize the print out should be outside of this function.

chaospuppy: Okay that makes sense, I'll move imageArchives logic into a separate function that can be used for post-processing the images located in FindImages for find-images.

src/config/lang/english.go
```diff
@@ -383,8 +383,8 @@ $ zarf package pull oci://ghcr.io/zarf-dev/packages/dos-games:1.3.0 -a skeleton`
 	CmdDevSha256sumShort         = "Generates a SHA256SUM for the given file"
 	CmdDevSha256sumRemoteWarning = "This is a remote source. If a published checksum is available you should use that rather than calculating it directly from the remote link."
 
-	CmdDevFindImagesShort = "Evaluates components in a Zarf file to identify images specified in their helm charts and manifests"
-	CmdDevFindImagesLong  = "Evaluates components in a Zarf file to identify images specified in their helm charts and manifests.\n\n" +
+	CmdDevFindImagesShort = "Evaluates components in a Zarf file to identify images specified in their helm charts, manifests, and image archives"
+	CmdDevFindImagesLong  = "Evaluates components in a Zarf file to identify images specified in their helm charts, manifests, and image archives.\n\n" +
```
AustinAbro321: We can revert this actually, we still only look through helm charts and manifests, we just display where to put the images with image archives in mind

src/pkg/packager/find_images.go
```diff
@@ -81,7 +83,17 @@ type ComponentImageScan struct {
...
 
-// FindImages iterates over the manifests and charts within each component to find any container images
+type ImageArchive struct {
+	Images []string `yaml:"images"`
+	Path   string   `yaml:"path"`
+}
+
+type ImageArchivesScan struct {
+	ComponentName string
+	ImageArchives []ImageArchive
```
chaospuppy: I considered using `v1alpha1.ZarfComponent.ImageArchive` here, but thought perhaps there could be some divergence in the future.  Let me know if you'd prefer the re-use of `v1alpha1.ZarfComponent.ImageArchive`

src/pkg/packager/update.go
```diff
@@ -79,42 +88,114 @@ func createUpdate(zarfPackage v1alpha1.ZarfPackage, imagesScans []ComponentImage
...
+	for _, component := range zarfPackage.Components {
+		imageSet := make(map[string]struct{}, len(component.Images))
+		for _, img := range component.Images {
+			imageSet[img] = struct{}{}
+		}
+		componentMap[component.Name] = imageSet
+	}
+
+	// Update needed if:
+	// 1. A component is found in the package definition that is not included in the scan
```
chaospuppy: I can remove these comments if desired, they helped me as I was making sense of this function.

src/cmd/dev.go
```diff
@@ -768,6 +768,11 @@ func (o *devFindImagesOptions) run(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("unable to find images: %w", err)
 	}
 
+	archiveImageScans, imagesScans, err := packager.FilterImagesFoundInArchives(ctx, basePath, imagesScans)
+	if err != nil {
+		return fmt.Errorf("unable to filter images included in imageArchives: %w", err)
+	}
+
 	if o.why != "" {
```
AustinAbro321, line 783: The why block should be moved to before we filter the images in archives, since it is a separate print, only concerned with manifests, rather than what's in the zarf.yaml

src/pkg/packager/update.go
AustinAbro321: They're fine to keep

src/pkg/packager/find_images.go
AustinAbro321: I think they are coupled in that if image archives were to change, we'd also want to change this block since ultimately we'd want to use the information here to update / print a package definition

AustinAbro321 (changes requested):
Good progress. When running this locally on examples/kiwix I get

```yaml
components:
  - name: kiwix-serve
    images:
      - ghcr.io/kiwix/kiwix-serve:3.5.0-2
      - kiwix-data:local
  # Archive images - kiwix-serve
  - name: kiwix-serve
    imageArchives:
      - path: kiwix-data.tar
        images: []
```

Whereas I'd expect
```yaml
  - name: kiwix-serve
    images:
      - ghcr.io/kiwix/kiwix-serve:3.5.0-2
    imageArchives:
      - path: kiwix-data.tar
        images: 
         - kiwix-data:local
```
The kiwix-data:local is in the wrong spot, and they are printed as different components.

chaospuppy:
@AustinAbro321 I've resolved the duplicate components issue, thanks for spotting that.  To your point about `kiwix-data:local` missing from imageArchives, I'm wondering how we want to handle that exactly.  Since creating a tarball of an image tagged without a registry with `docker save` always results in an index.json `io.containerd.image.name` that is pre-pended with `docker.io/library/`, there is some ambiguity over if `kiwix-data:local` really does correspond to `docker.io/library/kiwix-data:local` found in the archive.  The user could also be looking for `ghcr.io/kiwix/kiwix-data:local`, for example.  Are do we also want to assume that images without registries in their references refer to docker.io/library images?

AustinAbro321:
So a fun thing about containers is that if the registry is not specified then it's always assumed to be `docker.io`. Docker was the registry that popularized containers, so they made themselves the default. Generally the community wants to move away from this. IMO it would be a positive change if the find images command started returning the full path, so instead of kiwix-data:local, it'd return `docker.io/library/kiwix-data:local`. This should simplify things for you as well. 

You should be able to use the `transform.ParseImageRef` library to get the qualified name.

chaospuppy:
@AustinAbro321 updated to return qualified names for scan `Matches`.  I did not do the same for `PotentialMatches` since if I understand the use case for `PotentialMatches`, we wouldn't want to push an image to the zarf registry from the archive that isn't actually used.

src/pkg/images/unpack.go
```diff
@@ -37,64 +36,88 @@ const (
...
+	}
+	imageDir, err := determineImageDirectory(extractionDir)
+	if err != nil {
+		return nil, fmt.Errorf("failed to determine image directory: %w", err)
+	}
+
+	manifests, err := getManifestsFromOCILayout(imageDir)
+
+	return manifests, err
+}
```
AustinAbro321: nit: just return getManifestsFromOCILayout

src/pkg/images/unpack.go
```diff
@@ -37,64 +36,88 @@ const (
 	dockerContainerdImageStoreAnnotation = "containerd.io/distribution.source.docker.io"
 )
 
+// GetManifestsFromArchive take an image archive and returns a list of image descriptiors
+func GetManifestsFromArchive(ctx context.Context, imageArchive string) ([]ocispec.Descriptor, error) {
+	// Create a temporary directory for extraction
+	extractionDir, err := utils.MakeTempDir("")
```
AustinAbro321:
```suggestion
	extractionDir, err := utils.MakeTempDir(config.CommonOptions.TempDirectory)
```

src/pkg/images/unpack.go
```diff
@@ -37,64 +36,88 @@ const (
 	dockerContainerdImageStoreAnnotation = "containerd.io/distribution.source.docker.io"
 )
 
+// GetManifestsFromArchive take an image archive and returns a list of image descriptiors
+func GetManifestsFromArchive(ctx context.Context, imageArchive string) ([]ocispec.Descriptor, error) {
```
AustinAbro321:
```suggestion
func GetManifestsFromArchive(ctx context.Context, imageArchive string) (_ []ocispec.Descriptor, err error) {
```

In order for the defer to forward the err, the err must be a named return

src/pkg/packager/find_images.go
```diff
@@ -306,6 +322,81 @@ func FindImages(ctx context.Context, packagePath string, opts FindImagesOptions)
...
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to retrieve image manifests from archive %s: %w", archive.Path, err)
+			}
+			archiveImages, err := images.FindImagesInOCIManifests(imageManifests)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to unpack image archive %s: %w", archive.Path, err)
+			}
+			imageArchive := v1alpha1.ImageArchive{
+				Images: archiveImages,
+				Path:   archivePath,
```
AustinAbro321: This should stay as the original archive.Path, when interacting with the archive you may have to use the base directory, however when we give this back to the user to put in their zarf.yaml it's relative to the zarf.yaml file not relative to their local path
```suggestion
				Path:   archive.Path,
```

src/pkg/packager/find_images.go
```diff
@@ -306,6 +322,81 @@ func FindImages(ctx context.Context, packagePath string, opts FindImagesOptions)
...
+		if len(component.ImageArchives) < 1 {
+			// Skip if there are no imageArchives
+			continue
+		}
+		scan := ImageArchivesScan{ComponentName: component.Name}
+		for _, archive := range component.ImageArchives {
+			pkgPath, err := layout.ResolvePackagePath(packagePath)
+			if err != nil {
+				return nil, nil, err
+			}
```
AustinAbro321: nit: you can move this outside of the loop

src/cmd/dev.go
```diff
@@ -805,11 +816,48 @@ func (o *devFindImagesOptions) run(cmd *cobra.Command, args []string) error {
 				componentDefinition += fmt.Sprintf("      - %s\n", cosignArtifact)
 			}
 		}
+		for _, archiveFinding := range imageArchivesScans {
+			if archiveFinding.ComponentName == finding.ComponentName {
+				componentDefinition += fmt.Sprintf("  # Archive images - %s\n", archiveFinding.ComponentName)
+				for _, imageArchive := range archiveFinding.ImageArchives {
+					componentDefinition += "    imageArchives:\n"
```
AustinAbro321: This needs to be done only once per component, otherwise you get to image archive blocks in the same component

src/pkg/packager/find_images.go
```diff
@@ -238,7 +247,14 @@ func FindImages(ctx context.Context, packagePath string, opts FindImagesOptions)
 		}
 
 		sortedMatchedImages, sortedExpectedImages := getSortedImages(matchedImages, maybeImages)
-		scan.Matches = sortedMatchedImages
+
+		for _, image := range sortedMatchedImages {
+			imageReference, err := transform.ParseImageRef(image)
+			if err != nil {
+				return nil, fmt.Errorf("could not parse image reference for matched image %s", image)
```
AustinAbro321:
```suggestion
				return nil, fmt.Errorf("could not parse image reference for matched image %s: %w", image, err)
```
nit: better to pass error

AustinAbro321 (changes requested):
Good progress, some more requests as I go through

src/pkg/packager/find_images.go
```diff
@@ -319,6 +335,82 @@ func FindImages(ctx context.Context, packagePath string, opts FindImagesOptions)
 	return componentImageScans, nil
 }
 
+// FilterImagesFoundInArchives recieves imageScans and discovers if any of its elements are duplicated by
+// images provided in imageArchives parsed out of the package located at packagePath.  If duplciates are found, they
+// are removed from imageScans and added to archiveImageScans.
+func FilterImagesFoundInArchives(ctx context.Context, packagePath string, imageScans []ComponentImageScan) (_ []ImageArchivesScan, _ []ComponentImageScan, err error) {
```
AustinAbro321: I'm thinking more about the architecture we're going with here. I'd like to see this function private and another abstract function introduced `FindDefinitionImages`. This function would return a struct like this, and would do the logic of filtering the images.  
```go
  type DefinitionImageResult struct {                                                                                                                           
      ComponentImageScan                                                                                                                                                
      ImageArchives []v1alpha1.ImageArchive
  }  
```

This should solve a few problems: 
1. dev.go and UpdateImages will be simplified since all data will already be in one struct
2. when --why is used, we can just call the original Packager.FindImages, it's no longer order dependent in dev.go
3. We can introduce a private function packager.findImages which accepts a package definition.  This way we only have to load the package once.

AustinAbro321 (changes requested):
Requesting an architecture change. Appreciate the iterations

src/pkg/packager/find_images.go
chaospuppy: I'm a little confused around `FindDefinitionImages` returning *a* struct like `DefinitionImageResult`.  I'm thinking `FindDefinitionImages` would need to return a `[]DefinitionImageResult` to that the rest of the logic in `dev.go` and `UpdateImages` can remain relatively unchanged.  Is that right?

Aside from that, I'm also not totally certain I understand the situations in which we call `packager.FindImages`, `packer.findImages`, and `packager.FindDefintionImages`.  My current idea is this:

- `FindImages` is only called if `--why` is set, in `dev.go`
- `FindDefinitionImages` is called if `--why` is not set, in `dev.go`
-  `findImages` is called by both `FindImages` and `FindDefinitionImages`
  - Both `FindImages` and `FindDefintitionImages` handle loading the package before passing it as an argument to `findImages`

Is that right?

AustinAbro321: Yes, I misspoke, it would return the struct. Your understanding and flow is correct

chaospuppy: Made updates accordingly!

src/pkg/packager/find_images.go
```diff
@@ -107,6 +131,80 @@ func FindImages(ctx context.Context, packagePath string, opts FindImagesOptions)
...
+			}
+			imageArchive := v1alpha1.ImageArchive{
+				Images: archiveImages,
+				Path:   archive.Path,
+			}
+			result.ImageArchives = append(result.ImageArchives, imageArchive)
+			allArchiveImages = append(allArchiveImages, archiveImages...)
+		}
+
+		definitionImageResults = append(definitionImageResults, result)
```
chaospuppy, line 191: I considered excluding results that were only populated with `ComponentName`.

src/pkg/packager/find_images.go
```diff
@@ -84,11 +87,32 @@ type ComponentImageScan struct {
...
+type DefinitionImageResult struct {
+	ComponentImageScan
+	ImageArchives []v1alpha1.ImageArchive
+}
+
+// FindDefinitionImages finds all images contained in a component and filters them according to images discovered in
+// imageArchives.
+// It returns []DefinitionImageResult
+func FindDefinitionImages(ctx context.Context, packagePath string, opts FindImagesOptions) ([]DefinitionImageResult, error) {
+	loadOpts := load.DefinitionOptions{}
```
AustinAbro321: You need to pass all of the relevant loadOptions here

src/cmd/dev.go
```diff
@@ -810,11 +830,25 @@ func (o *devFindImagesOptions) run(cmd *cobra.Command, args []string) error {
 				componentDefinition += fmt.Sprintf("      - %s\n", cosignArtifact)
 			}
 		}
+		for _, archive := range finding.ImageArchives {
+			componentDefinition += fmt.Sprintf("  # Archive images - %s\n", finding.ComponentName)
+			componentDefinition += "    imageArchives:\n"
```
AustinAbro321, line 844: This part needs to be outside of the loop so it works for multiple image archives in the same component

src/pkg/packager/find_images.go
```diff
@@ -107,6 +131,80 @@ func FindImages(ctx context.Context, packagePath string, opts FindImagesOptions)
...
+	var allArchiveImages []string
+	for _, component := range pkg.Components {
+		result := DefinitionImageResult{}
+		result.ComponentName = component.Name
+		if scan, ok := componentNameScanMap[component.Name]; ok {
+			result.ComponentImageScan = scan
+		}
+
+		for _, archive := range component.ImageArchives {
+			archivePath := path.Join(pkgPath.BaseDir, archive.Path)
```
AustinAbro321:
```suggestion
			archivePath := filepath.Join(pkgPath.BaseDir, archive.Path)
```
path is used for consistent separators, where filepath changes the separator depending on the OS so we want the latter here

src/pkg/packager/find_images.go
AustinAbro321: I think it's fine to assume findImages won't return empty scans

AustinAbro321 (changes requested):
The architecture is shaping up, a few more comments

src/pkg/packager/find_images.go
chaospuppy: Thanks for the callout, I forgot to fill in over making the placeholder.

src/cmd/dev.go
chaospuppy: Fixed

src/pkg/packager/find_images.go
AustinAbro321: This was never resolved

src/pkg/packager/find_images.go
```diff
@@ -257,7 +354,14 @@ func FindImages(ctx context.Context, packagePath string, opts FindImagesOptions)
...
+
+		for _, image := range sortedMatchedImages {
+			imageReference, err := transform.ParseImageRef(image)
+			if err != nil {
+				return nil, fmt.Errorf("could not parse image reference for matched image %s: %w", image, err)
+			}
+			scan.Matches = append(scan.Matches, imageReference.Reference)
+		}
 
 		// Handle the "maybes"
```
AustinAbro321, line 376: We should also normalize the maybe images with `transform.parseImageRef` (and the cosign artifacts)

src/cmd/dev.go
```diff
@@ -803,11 +803,31 @@ func (o *devFindImagesOptions) run(cmd *cobra.Command, args []string) error {
...
+	}
+	if err != nil {
+		return fmt.Errorf("unable to filter images included in imageArchives: %w", err)
+	}
+
+	imageScanComponentNames := make(map[string]bool)
+
+	for _, scan := range definitionImageResults {
+		imageScanComponentNames[scan.ComponentName] = true
+	}
```
AustinAbro321: This is dead code

src/pkg/packager/find_images.go
```diff
@@ -107,6 +130,80 @@ func FindImages(ctx context.Context, packagePath string, opts FindImagesOptions)
 		return nil, err
 	}
 
+	return findImages(ctx, pkg, packagePath, opts)
+}
+
+// FilterImagesFoundInArchives recieves imageScans and discovers if any of its elements are duplicated by
+// images provided in imageArchives parsed out of the package located at packagePath.  If duplciates are found, they
+// are removed from the final []DefinitionImageResult.ComponentImageScans.
```
AustinAbro321: Nit: this had some spelling mistakes, I think we could make it briefer as well
```suggestion
// filterImagesFoundInArchives merges scan results with each component's imageArchives.
// An image present in both surfaces only on the archive side
```

AustinAbro321 (changes requested):
Nice work so far, this is super close

src/pkg/packager/find_images.go
chaospuppy: I could be mistaken, but drilling down for cosign artifacts, it looks like we're receiving [`Repository.Name()`](https://github.com/google/go-containerregistry/blob/68a569e4a0eb927d36ccb0fcdf4578425c03b5a2/pkg/name/repository.go#L55) at the bottom of the call stack, which will always be a full ref.

AustinAbro321 (dismissed):
LGTM, nice work!

src/pkg/packager/find_images.go
```diff
@@ -107,6 +137,79 @@ func FindImages(ctx context.Context, packagePath string, opts FindImagesOptions)
...
+
+		for _, archive := range component.ImageArchives {
+			archivePath := filepath.Join(pkgPath.BaseDir, archive.Path)
+			imageManifests, err := images.GetManifestsFromArchive(ctx, archivePath)
+			if err != nil {
+				return nil, fmt.Errorf("failed to retrieve image manifests from archive %s: %w", archive.Path, err)
+			}
+			archiveImages, err := images.FindImagesInOCIManifests(imageManifests)
+			if err != nil {
+				return nil, fmt.Errorf("failed to unpack image archive %s: %w", archive.Path, err)
```
soltysh: Nit: maybe `failed to find images in archive...` would be more appropriate?

src/pkg/packager/find_images.go
```diff
@@ -84,11 +86,39 @@ type ComponentImageScan struct {
...
+}
+
+// FindDefinitionImages finds all images contained in a component and filters them according to images discovered in
+// imageArchives.
+// It returns []DefinitionImageResult
+func FindDefinitionImages(ctx context.Context, packagePath string, opts FindImagesOptions) ([]DefinitionImageResult, error) {
+	loadOpts := load.DefinitionOptions{
+		Flavor:           opts.Flavor,
+		SetVariables:     opts.CreateSetVariables,
+		CachePath:        opts.CachePath,
```
soltysh: The cache is never initiated similarly to how we're doing that in `FindImages`, although I doubt we're gonna need it. Still, just to double check my understanding, was it skipped intentionally?

src/pkg/images/unpack.go
```diff
@@ -37,64 +36,86 @@ const (
 	dockerContainerdImageStoreAnnotation = "containerd.io/distribution.source.docker.io"
 )
 
+// GetManifestsFromArchive take an image archive and returns a list of image descriptiors
```
soltysh:
```suggestion
// GetManifestsFromArchive take an image archive and returns a list of image descriptors
```

soltysh (commented):
A few nits, but overall lgtm

src/pkg/packager/find_images.go
chaospuppy: That's an oversight on my part, I'll update

soltysh (approved):
lgtm

AustinAbro321 (approved):

PR #4562, opened 2026-01-28
feat: add ability to supply custom init package

### LINKED ISSUE #4560: Allow custom package to be passed to `zarf init`

At present the only way to test custom versioned init packages is like the following:

```shell
mv zarf-init-amd64-3.5.1.tar.zst zarf-init-amd64-$(zarf version).tar.zst
```

- Keep the default behavior for looking for a zarf package that matches the current cli version
- Like with `zarf package deploy` allow the passing of a file path and use that path as the zarf package getting deploying.

### Additional context

This is not a major blocking, seeing as we get (mostly) the same functionality from doing just a `zarf package deploy`... However for those of use that use custom package versions, unless we do the `mv` command above, we are unable to enable the proxy-registry (I mean seeing as it is an alpha feature, it makes sense we need to go through some hops to get it enabled)

a1994sc:
## Description

Allow for passing custom init packages to `zarf init`

## Related Issue

Fixes #4560

## Checklist before merging

### DIFF
--- a/.github/workflows/test-upgrade.yml
+++ b/.github/workflows/test-upgrade.yml
@@ -43,43 +43,6 @@ jobs:
           path: build/
           retention-days: 1
 
-  validate-custom-upgrade:
-    runs-on: ubuntu-latest
-    needs:
-      - build-upgrade
-    steps:
-      - name: Checkout
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-
-      - name: Download build artifacts
-        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
-        with:
-          name: build-artifacts
-          path: build/
-
-      - name: Setup golang
-        uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
-        with:
-          go-version-file: go.mod
-
-      - name: Make Zarf executable
-        run: |
-          chmod +x build/zarf
-
-      - name: Install release version of Zarf
-        uses: zarf-dev/setup-zarf@10e539efed02f75ec39eb8823e22a5c795f492ae #v1.0.1
-        with:
-          download-init-package: true
-
-      - name: Initialize the cluster with the release version
-        # Test will initialize the cluster with the latest released version package, with the current built binary
-        run: |
-          sudo env CI=true ./build/zarf init --components k3s,git-server --nodeport 31337 --confirm ${HOME}/.zarf-cache/zarf-init-amd64-$(zarf version).tar.zst
-
-      - name: Upgrade the cluster with the build version
-        run: |
-          sudo env CI=true ./build/zarf init --components k3s,git-server --nodeport 31337 --confirm ./build/zarf-init-amd64-$(./build/zarf version).tar.zst
-
   validate-upgrade:
     runs-on: ubuntu-latest
     needs:
--- a/.github/workflows/test-version-delta.yml
+++ b/.github/workflows/test-version-delta.yml
@@ -1,0 +1,43 @@
+name: Test Version Delta
+on:
+  pull_request:
+    paths-ignore:
+      - "**.md"
+      - "**.jpg"
+      - "**.png"
+      - "**.gif"
+      - "**.svg"
+      - "adr/**"
+      - "docs/**"
+      - "CODEOWNERS"
+  merge_group:
+
+# Abort prior jobs in the same workflow / PR
+concurrency:
+  group: e2e-version-delta-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  version-delta:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Setup golang
+        uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
+        with:
+          go-version-file: go.mod
+
+      - name: Build PR binary and zarf init package
+        uses: ./.github/actions/packages
+
+      - name: Install release version of Zarf
+        uses: zarf-dev/setup-zarf@10e539efed02f75ec39eb8823e22a5c795f492ae #v1.0.1
+        with:
+          download-init-package: true
+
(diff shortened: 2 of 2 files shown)


a1994sc:
So I wanted to update `src/test/e2e/20_zarf_init_test.go` to reflect the changes done in this PR...

However I wanted to ask what would be the best way to go about testing it?

Should I first install an older version of zarf, say `v0.70.1` with :

```shell
zarf init oci://ghcr.io/zarf-dev/packages/init:v0.70.1
```

Then try to just a:

```shell
zarf init
```

?

site/src/content/docs/commands/zarf_init.md
```diff
@@ -35,6 +35,9 @@ $ zarf init
 # Initializing w/ Zarfs internal git server:
 $ zarf init --components=git-server
 
+# Initializing w/ Zarfs with a custom init package:
+$ zarf init oci://ghcr.io/cool-project/zarf-init:v1.2.3
```
AustinAbro321: This is a funny one, but I'm going to ask you either point to a local package, or a package starting with the path zarf-dev. We  wouldn't want the owner of https://github.com/cool-project adding in a malicious init package

src/cmd/initialize.go
AustinAbro321, line 1: I can see the argument against this given that we're simply calling deploy, and technically any package will work, however, I still think it would be prudent to check if the package is a ZarfInitConfig after it's loaded, and error if not.

AustinAbro321:
@a1994sc I actually have a different request for how you should test this. Could you create a new workflow that tests the previouses release of the init package, deployed with the Zarf binary in the PR? This feature makes it easier for older init package versions to be deployed, so I want to make sure we have a proper test for that.

site/src/content/docs/commands/zarf_init.md
a1994sc: That did give me a good laugh, but I addressed this comment

src/cmd/initialize.go
a1994sc: Added a check, please let me know if that was the correct way to tackle that

a1994sc:
> @a1994sc I actually have a different request for how you should test this. Could you create a new workflow that tests the previouses release of the init package, deployed with the Zarf binary in the PR? This feature makes it easier for older init package versions to be deployed, so I want to make sure we have a proper test for that.

I do think that would be a good idea. Should I create a new one, or update [`.github/workflows/test-upgrade.yml`](https://github.com/zarf-dev/zarf/blob/main/.github/workflows/test-upgrade.yml)

To install the latest release, but with the PR binary, and then upgrade?

src/cmd/initialize.go
AustinAbro321: Looks correct, though I'd recommend changing the error message to "zarf init can only deploy packages of kind "ZarfInitConfig"", this way it's more clear to the user what the issue is

AustinAbro321:
hmm I can see the argument both ways. I think test upgrade flow would work if we did: 
1. Deploy old release init package with old release version of Zarf
2. Deploy old release init package with build version of Zarf from PR
3. continue with test

The argument against adding it is that it makes our longest job longer, but given that a secondary init is only about a ~45 second increase, I think it'd be fine in test-upgrade.

a1994sc:
@AustinAbro321 can you let me know if the changes to `.github/workflows/test-upgrade.yml` are good for the updated test logic or if you would like me to change anything. Thank you.

.github/workflows/test-upgrade.yml
```diff
@@ -80,10 +71,13 @@ jobs:
...
-        # NOTE: "PATH=$PATH" preserves the default user $PATH. This is needed to maintain the version of zarf installed
-        #       in a previous step. This test run will the current release to create a K3s cluster.
+        # Test will initialize the cluster with the latest released version package, with the current built binary
+        run: |
+          sudo env CI=true ./build/zarf init --components k3s,git-server --nodeport 31337 --confirm ${HOME}/.zarf-cache/zarf-init-amd64-$(zarf version).tar.zst
+
+      - name: Upgrade the cluster with the build version
         run: |
-          sudo env "PATH=$PATH" CI=true zarf init --components k3s,git-server --nodeport 31337 --confirm
+          sudo env CI=true ./build/zarf init --components k3s,git-server --nodeport 31337 --confirm ./build/zarf-init-amd64-$(./build/zarf version).tar.zst
```
AustinAbro321: After some more consideration, I actually am going to advocate for a separate workflow to test using the build version of Zarf with the release init package

I don't want to lose the core case we're testing here, which is taking a zarf cluster initialized with version X to version Y. This is the most common use case, so I want to keep it first class within this test, and not risk a situation where it ends up passing, but only because the cluster was first initialized with the build version of Zarf. 

From a code clarity perspective should help keep it more clear what each workflow is testing too.

a1994sc: So just for my own understanding, do you like the current flow in `.github/workflows/test-upgrade.yml` in general? If so, I can work on creating a new workflow based off it

AustinAbro321: The test upgrade workflow can be a bit confusing I think, but overall it's fine. I don't think we'll need to duplicate the workflow though. I think just proving we can run a `build/zarf init` using the latest release version of Zarf is enough.

a1994sc: So do you want it to look like this?

```bash
sudo env CI=true ./build/zarf init --components k3s,git-server --nodeport 31337 --confirm
```

That should use the init package in the same folder as `build/zarf` right?

AustinAbro321: Yeah the command should mostly look like this, but I'd have it use the last release version of Zarf. Essentially, I'm thinking the command should look like this. We use the build Zarf, but the last release version 

```bash
sudo env CI=true ./build/zarf init --components k3s,git-server --nodeport 31337 --confirm ${HOME}/.zarf-cache/zarf-init-amd64-$(zarf version).tar.zst
```

a1994sc: Isn't that what I have on line 76?

AustinAbro321: Yes, but I'd like it to be outside of the test upgrade workflow. 

Right now we're not preforming the test case of release zarf + release zarf init. I don't want to risk there being a difference, between that and build zarf + release zarf init, especially given that using release Zarf to the next version is the the most common flow users will preform when upgrading. I believe having it in a separate workflow also decreases the complexity of this workflow, and makes what we're testing more obvious in both scenarios

PR #4565, opened 2026-01-28
feat: add nodeSelector to zarf agent and registry charts

### LINKED ISSUE #2176: Tolerations/NodeSelector for Zarf registry

We'd like to be able to add a NodeSelector and tolerations to the Zarf registry via deployment variables for the `zarf init` command with `--set`. The immediate need driving this request is to work around https://github.com/defenseunicorns/zarf/issues/2104, but I think other users would benefit by being able to fine-tune their registry placement or enable the use of Zarf on nodes already tainted for different reasons.

- Modify the Zarf registry charts to allow templating of the `nodeSelector` and `tolerations` keys in the spec.
- Add Zarf variables the Zarf registry zarf.yaml to enable setting these w/ `--set` during `zarf init`.

Manual modification and/or deployment of custom zarf init package with these set, or patching the registry manifests post-deployment.

rodneyxr:
## Description

Adds the ability to specify nodeSelector in the zarf-agent and zarf-registry helm charts.

## Related Issue

Fixes #2176.

## Checklist before merging

### DIFF
--- a/packages/zarf-agent/agent-values.yaml
+++ b/packages/zarf-agent/agent-values.yaml
@@ -1,0 +1,8 @@
+affinity:
+  ###ZARF_VAR_AGENT_AFFINITY###
+
+tolerations:
+  ###ZARF_VAR_AGENT_TOLERATIONS###
+
+nodeSelector:
+  ###ZARF_VAR_AGENT_NODE_SELECTOR###
--- a/packages/zarf-agent/zarf.yaml
+++ b/packages/zarf-agent/zarf.yaml
@@ -3,6 +3,22 @@ metadata:
   name: init-package-zarf-agent
   description: Install the zarf agent mutating webhook on a new cluster
 
+variables:
+  - name: AGENT_AFFINITY
+    description: Custom affinity map for the zarf-agent deployment
+    default: ""
+    autoIndent: true
+
+  - name: AGENT_TOLERATIONS
+    description: Custom tolerations array for the zarf-agent deployment
+    default: ""
+    autoIndent: true
+
+  - name: AGENT_NODE_SELECTOR
+    description: Custom node selector map for the zarf-agent deployment
+    default: ""
+    autoIndent: true
+
 constants:
   - name: AGENT_IMAGE
     value: "###ZARF_PKG_TMPL_AGENT_IMAGE###"
@@ -25,6 +41,8 @@ components:
         localPath: chart
         version: 0.1.0
         namespace: zarf
+        valuesFiles:
+          - agent-values.yaml
     actions:
       onCreate:
         before:


a1994sc (approved):

packages/zarf-agent/chart/values.yaml
AustinAbro321, line 1: I believe we'll have to add a Zarf variable to agent zarf.yaml, 'agent_node_selector' in order for users to be able to pass through a value here.

rodneyxr: I can definitely add that if that is something we want. I was just following how the packages were already structured.

The zarf-registry zarf.yaml has many passthrough variables but the zarf-agent zarf.yaml only has 2 constants defined.

In our current use case, we're using overrides in a uds bundles so this wasn't an issue for us.

AustinAbro321: Ah gotcha, yeah please add that. We can't assume that bundles exist when working on Zarf. The registry has many configuration options as a generic registry tool, whereas the agent has a specific role within Zarf, which naturally lead to that difference.

rodneyxr: Do you think I should add passthrough vars for affinity and tolerations as well then?

AustinAbro321: Yes

rodneyxr: So when staging `packages/zarf-agent/zarf.yaml`, I'm getting a failure in the pre-commit hook for `detect-aws-credentials`

```sh
detect aws credentials...................................................Failed
- hook id: detect-aws-credentials
- exit code: 1

AWS secret found in packages/zarf-agent/zarf.yaml: test************************
```

Looks like its happening [on this line](https://github.com/zarf-dev/zarf/blob/main/packages/zarf-agent/zarf.yaml#L31) which I'm not touching...

```yaml
    actions:
      onCreate:
        before:
          - cmd: "test \"###ZARF_PKG_TMPL_AGENT_IMAGE_TAG###\" != \"local\" || make build-local-agent-image AGENT_IMAGE_TAG=\"###ZARF_PKG_TMPL_AGENT_IMAGE_TAG###\" ARCH=\"###ZARF_PKG_ARCH###\""
```

Since this seems to be a false positive, what is your recommendation to address this or am I just doing something wrong? 😬

AustinAbro321: Interesting, never seen that before. I would say add an exclude to the pre-commit file for now. I'm not able to reproduce, but maybe I will be able to with your updated file.

rodneyxr: I've added the 3 zarf variables (`AGENT_AFFINITY`, `AGENT_TOLERATIONS`, `AGENT_NODE_SELECTOR`) to the zarf.yaml and had to create an `agent-values.yaml`.

PR #4566, opened 2026-01-29
feat(prune): allow for pruning to continue on manifest unknown

### LINKED ISSUE #2641: request: Focus prune command to only worry about pruning

### LINKED ISSUE #4563: Zarf prune no space left on device
### Environment
Device and OS: bottlerocket
App version: latest

### Steps to reproduce
1. Deploy a package that has images too large for your repo disk.
2. Exception is thrown during package deploy stating "no space left on device"
3. Run `zarf tools registry prune`

### Expected result

Prune runs successfully even if an error occurs on a single image.

### Actual Result

Prune fails with the following error
`MANIFEST_UNKNOWN: manifest unknown; unknown tag=7.1.62`

Image

### Severity/Priority

High due to unable to use the cluster if you get into this state.

brandtkeller:
## Description

This update allows prune to continue when encountering the `MANIFEST_UNKNOWN` error from `crane.Digest()`.  This one is tricky to test - I tried interrupting the image push and was not able to see the same result as when storage space is depleted - but given the prevalence of the issue in issues and slack messages I believe this is a reasonable path. 

## Related Issue

Fixes #2641
Fixes #4563 

## Checklist before merging


AustinAbro321:
I was able to test this with the following: 

1. zarf init --confirm
2. zarf package mirror-resources zarf-package-podinfo-flux-amd64.tar.zst --images
3. exec into the registry pod
4. rm /var/lib/registry/docker/registry/v2/repositories/stefanprodan/manifests/podinfo/_manifests/revisions/sha256/fc60d367cc05bedae04d6030e270daa89c3d82fa18b1a15
5. You get a warning instead of an error on this branch

I'm thinking about a case when this command is run with automation. Not knowing that this isn't working could cause issues where in a worse case, no images are being pruned, and there is no signal that somethings gone awry.

What are your thoughts on adding a flag such as `--ignore-missing`? When we do run into a transport problem, we could recommend re-running with the flag. It's probably best to add the flag to `zarf tools registry prune`, I don't know if we'd want to add it to `zarf tools registry catalog` too given that the list is the output rather than another action, but I'm not necessarily against it.

brandtkeller:
> What are your thoughts on adding a flag such as `--ignore-missing`? When we do run into a transport problem, we could recommend re-running with the flag. It's probably best to add the flag to `zarf tools registry prune`, I don't know if we'd want to add it to `zarf tools registry catalog` too given that the list is the output rather than another action, but I'm not necessarily against it.

I can appreciate that stance. Adding to prune for scope of these changes. With catalog we did not see the same issue originally (when debugging in support channels).

PR #4567, opened 2026-01-29
refactor(wait):  avoid shelling out to kubectl during wait

AustinAbro321:
## Description

This entirely eliminates any shelling out in the wait command. Instead, we first get the kind directly from Kubernetes to replace the kubectl get calls, then call the equivalent of `kubectl wait` directly through the kube go dependency.

Went for  parity with the current behavior of `zarf tools wait-for` which includes waiting for a cluster to be up, even if it currently doesn't exist. This may be behavior we want to re-evaluate when we introduce `zarf tools wait-for resource` https://github.com/zarf-dev/zarf/issues/4550

This does change the logging strategy, previously wait would log each second it waits like so

```bash
zarf tools wait-for -n kube-system pods metrics-server-5985cbc9d7-92mwh1
2026-02-11 13:17:07 INF using config file location=/home/austin/code/zarf/zarf-config.toml
2026-02-11 13:17:07 INF waiting for pods/metrics-server-5985cbc9d7-92mwh1 in namespace kube-system to exist.
2026-02-11 13:17:08 INF resource error error=exit status 1
2026-02-11 13:17:09 INF resource error error=exit status 1
```

Now the message will not continuously update, and instead look like, unless the `-l debug` flag is used. This is more consistent with the logging strategy used elsewhere in Zarf

```bash
build/zarf tools wait-for -n kube-system pods metrics-server-5985cbc9d7-92mwh1 ready 
2026-02-11 13:19:34 INF using config file location=/home/austin/code/zarf/zarf-config.toml
2026-02-11 13:19:34 INF waiting for resource kind=pods identifier=metrics-server-5985cbc9d7-92mwh1 condition=condition=ready namespace=kube-system
```

## Checklist before merging

### DIFF
--- a/src/test/e2e/46_tools_wait_for_test.go
+++ b/src/test/e2e/46_tools_wait_for_test.go
@@ -33,9 +33,10 @@ func TestWaitFor(t *testing.T) {
 		require.Error(t, err)
 	})
 
-	t.Run("wait for resource without specified namespace only looks in default namespace", func(t *testing.T) {
+	t.Run("wait for resource without specifying the namespace only looks in default namespace", func(t *testing.T) {
 		t.Parallel()
-		_, _, err := e2e.Zarf(t, "tools", "wait-for", "pod", "--timeout", "3s")
+		// There are never any jobs by default
+		_, _, err := e2e.Zarf(t, "tools", "wait-for", "jobs", "--timeout", "3s")
 		require.Error(t, err)
 	})
 


src/pkg/wait/wait.go
```diff
@@ -8,152 +8,238 @@ import (
...
+	var restConfig *rest.Config
+	var configFlags *genericclioptions.ConfigFlags
+	var mapping *meta.RESTMapping
+	deadline := time.Now().Add(timeout)
+	waitInterval := time.Second
+	for {
+		configFlags = genericclioptions.NewConfigFlags(true)
+		if namespace != "" {
+			configFlags.Namespace = ptr.To(namespace)
+		}
```
soltysh: I'd suggest not to set namespace here, but rather explicitly set it with every REST invocation below you're doing with your config. This way you'll always know where the namespace came from rather than rely on some weird behavior.

src/pkg/wait/wait.go
```diff
@@ -8,152 +8,238 @@ import (
...
 
-// unsafeShellCharsRegex matches any character that is NOT a letter, digit, underscore (/w) or shell safe special characters
-var unsafeShellCharsRegex = regexp.MustCompile(`[^\w@%+=:,./-]`)
+	// Fill these out in the Retry loop, which handles the cluster not yet being available
+	var restConfig *rest.Config
+	var configFlags *genericclioptions.ConfigFlags
+	var mapping *meta.RESTMapping
+	deadline := time.Now().Add(timeout)
+	waitInterval := time.Second
+	for {
```
soltysh: Two things wrt this for-loop:
1. Why do you need it? It seems this is only waiting for the kind resolution? Is that needed when you wait after applying CRD?
2. I'd suggest using something like `wait.PollUntilContextTimeout` [see example here](https://github.com/kubernetes/kubernetes/blob/dc1ec1211e4f54064ba6dafd8aac46ac3d4379b4/staging/src/k8s.io/kubectl/pkg/cmd/wait/wait.go#L331). Move all client setting up before the actual poll.

src/pkg/wait/wait.go
```diff
@@ -8,152 +8,238 @@ import (
...
-func ForResource(ctx context.Context, kind, identifier, condition, namespace string, timeout time.Duration) error {
+// waitForAnyResource waits for at least one resource of the given kind to exist.
+func waitForAnyResource(ctx context.Context, dynamicClient dynamic.Interface, resource schema.GroupVersionResource, namespace string, deadline time.Time) error {
 	l := logger.From(ctx)
 	waitInterval := time.Second
+	l.Info("waiting for at least one resource of kind to exist", "kind", resource.Resource, "namespace", namespace)
 
-	// Type of wait, condition or JSONPath
-	var waitType string
+	for {
```
soltysh: Same argument about using `wait.PollUntilContextTimeout`, it'll be easier to read too.

src/pkg/wait/wait.go
```diff
@@ -8,152 +8,238 @@ import (
...
-		select {
-		case <-expired:
-			return errors.New("wait timed out")
-		case <-ctx.Done():
-			return errors.New("received interrupt")
+func isJSONPathWaitType(condition string) bool {
+	if len(condition) == 0 || condition[0] != '{' || !strings.Contains(condition, "=") || !strings.Contains(condition, "}") {
+		return false
+	}
+	return true
```
soltysh: You can just do:
```golang
	return len(condition) > 0 && condition[0] = '{' && strings.Contains(condition, "=") && strings.Contains(condition, "}")
```

src/pkg/wait/wait.go
```diff
@@ -8,152 +8,238 @@ import (
...
-			if err != nil {
-				l.Debug("wait error", "error", err)
-				continue
-			}
+	waitInterval := time.Second
+	deadline := time.Now().Add(timeout)
 
-			// And just like that, success!
-			l.Info(completedMsg)
+	for {
```
soltysh: Again `wait.PollUntilContextTimeout` 😉

src/pkg/wait/wait.go
```diff
@@ -8,152 +8,238 @@ import (
...
+			ErrOut: io.Discard,
+		}
+		flags := cmdwait.NewWaitFlags(configFlags, streams)
+		flags.Timeout = timeout
+		flags.ForCondition = forCondition
+		if labelSelector != "" {
+			flags.ResourceBuilderFlags.LabelSelector = &labelSelector
+		}
+
+		opts, err := flags.ToOptions(args)
```
soltysh: You don't need to use the flags (those are only for wiring the cobra with options), but rather directly interact with options struct like here: https://github.com/kubernetes/kubernetes/blob/dc1ec1211e4f54064ba6dafd8aac46ac3d4379b4/staging/src/k8s.io/kubectl/pkg/cmd/delete/delete.go#L459-L468

So in your case, directly set `WaitOptions` struct members.

src/pkg/wait/wait.go
AustinAbro321: I think I'll try to make it more clear by splitting out the two steps, but essentially I'm trying to keep parity with the current `zarf tools wait-for` command. `zarf tools wait-for` will continue to wait even when the cluster doesn't exist, I think this has become a use case for some people. When we introduce the command `zarf tools wait-for resource`, I'm considering changing it in that command so an existing cluster is required. 

But yeah one use case is waiting for a CRD  to exist.

src/pkg/wait/wait.go
AustinAbro321: I don't see a way to mirror the behavior of the default ConditionFn outside of calling `flags.ToOptions` since it's a private function

src/pkg/wait/wait.go
AustinAbro321: I believe we have to use `genericclioptions.NewConfigFlags(true)` for the wait command, and the only way to set namespace is through this struct, however that is only for `opts.RunWait()`. I edited the code so that configFlags are only used for the wait command and there's no chance of namespace weirdness outside of that

src/pkg/wait/wait.go
```diff
@@ -8,156 +8,248 @@ import (
...
 	"net"
 	"net/http"
-	"path"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
 
-	"github.com/zarf-dev/zarf/src/api/v1alpha1"
+	"github.com/gpustack/gguf-parser-go/util/ptr"
```
brandtkeller: I feel less confident on the use of this as a direct dependency versus say `k8s.io/utils/ptr` - are there any hard requirements for this selection?

src/pkg/wait/wait.go
```diff
@@ -8,156 +8,248 @@ import (
...
-				if strings.Contains(stderr, "connect: connection refused") {
-					l.Info("api server unavailable")
-					continue
-				}
-				// otherwise just log and retry
-				l.Info("resource error", "error", err)
-				continue
-			}
+	forCondition := "create" // default: wait for existence
+	if condition != "" && !isExistsCondition(condition) {
```
brandtkeller: nit: a `delete` condition will evaluate true on this first conditional - setting `forCondition= "condition=delete"` - before then immediately be set to just `delete`

src/pkg/wait/wait.go
```diff
@@ -8,156 +8,248 @@ import (
...
+	}
+	return false
+}
 
-		select {
-		case <-expired:
-			return errors.New("wait timed out")
-		case <-ctx.Done():
-			return errors.New("received interrupt")
+// forResource is the internal implementation that can be tested with fake clients.
```
brandtkeller: unit testing would be a good addition for this internal function.

src/pkg/wait/wait.go
```diff
@@ -8,156 +8,248 @@ import (
...
-			return nil
+	opts, err := flags.ToOptions(args)
+	if err != nil {
+		return fmt.Errorf("failed to create wait options: %w", err)
+	}
+	opts.DynamicClient = dynamicClient
+
+	waitInterval := time.Second
+	err = wait.PollUntilContextTimeout(ctx, waitInterval, timeout, true, func(_ context.Context) (bool, error) {
+		err = opts.RunWait()
```
brandtkeller, line 253: Is my understanding here that we have two loops waiting with the same timeout? Loop structure looks correct but I wonder if having the same timeout could potentially lead to a bit of a competing timeout scenario?

I wonder if we set the inner wait timeout a shorter duration to allow the outer to retry if needed? unsure this is a hard requirement.

src/pkg/wait/wait.go
```diff
@@ -8,156 +8,248 @@ import (
...
 
-// ForResource waits for a Kubernetes resource to meet the specified condition using kubectl wait.
-func ForResource(ctx context.Context, kind, identifier, condition, namespace string, timeout time.Duration) error {
+// waitForAnyResource waits for at least one resource of the given kind to exist.
+func waitForAnyResource(ctx context.Context, dynamicClient dynamic.Interface, resource schema.GroupVersionResource, namespace string, timeout time.Duration) error {
 	l := logger.From(ctx)
 	waitInterval := time.Second
+	l.Info("waiting for any resource of kind to exist", "kind", resource.Resource, "namespace", namespace)
+
+	resourceClient := dynamicClient.Resource(resource).Namespace(namespace)
```
brandtkeller: I need to test this - I am not sure this can delineate between cluster-scoped resources and a lack of a namespace for a namespaced resource. 

if someone forgets to add a `-n` to a namespaced resource I believe this will unintuitively return on the existence of that resource in the cluster. 

I think we could check for resource being cluster scoped and then otherwise get the default from the kube context?

src/pkg/wait/wait.go
AustinAbro321: Added a comment, but the reason why we wrap `opts.RunWait` at all is because unless "create" is the condition, then `opts.RunWait()` will fail instantly if the resource doesn't exist. Instead, we want to retry when the resource doesn't exist.  

There is a clear benefit to giving opts.RunWait() a shorter timeout while we wait for https://github.com/kubernetes/kubernetes/pull/136781 to be merged. Right now if context is cancelled by the user, it won't be propagated since `opts.RunWait` doesn't accept context. Updated to give opts.RunWait() a timeout of 10 seconds.

src/pkg/wait/wait.go
AustinAbro321: yeah, great catch, definitely intended to use the k8s library

src/pkg/wait/wait.go
AustinAbro321: Sorry, that's an old comment I meant to remove. I spent a lot of cycles trying to get unit testing going here. I did get a mostly functional system, however with the way opts.RunWait was written, I had to write a lot of mocks, and it got quite complicated. I ended up feeling like e2e tests were more reliable, and simpler. It's possible there is a better unit testing strategy I didn't find.

brandtkeller: ah good context 👍

src/pkg/wait/wait.go
AustinAbro321: Really subtle and nice catch. Yeah release `zarf tools wait-for pod` will only pass if there is a pod in the default namespace, whereas this branchs `zarf tools wait-for pod` will pass if a pod is in the cluster. Will add a test and change the behavior

AustinAbro321: Added in logic to get the users default namespace. An added bonus here is that even when a user doesn't give a namespace, their default will show up in the log

src/test/e2e/46_tools_wait_for_test.go
```diff
@@ -26,11 +28,26 @@ func TestWaitFor(t *testing.T) {
 	})
 
 	t.Run("wait for non-existent resource times out", func(t *testing.T) {
+		t.Parallel()
```
AustinAbro321, line 31: Decided to make these run in parallel, this brings the entire test to 6 seconds from 26. I think I'm fine with the risks here, essentially you would need to make sure another test isn't creating a resource with the same name. Definitely open to objections though

PR #4582, opened 2026-02-03
feat: support helm charts in kustomizations

vincent-onebrief:
## Description

Currently kustomizations with helm chart references fail, even with plugins enabled.

```yaml
# zarf.yaml
components:
  - name: keycloak-server
    default: true
    manifests:
      - name: keycloak-server
        namespace: keycloak
        kustomizations:
          - .
        enableKustomizePlugins: true
```

```yaml
# kustomization.yaml
namespace: keycloak
helmCharts:
  - releaseName: keycloak
    name: keycloak
    repo: oci://registry-1.docker.io/bitnamicharts
    version: "23.3.3"
    valuesFile: values.yaml
```

```console
$ zarf dev inspect manifests .

2026-02-03 13:54:02 ERR unable to build kustomization .: trouble configuring builtin HelmChartInflationGenerator with config: `
name: keycloak
namespace: keycloak
releaseName: keycloak
repo: oci://registry-1.docker.io/bitnamicharts
valuesFile: values.yaml
version: 26.3.3
`: must specify --enable-helm
```

This can be worked around with an onCreate action that renders the manifests, but that's kinda jank and breaks `zarf dev inspect` and `zarf dev find-images`.

```yaml
# zarf.yaml
components:
  - name: keycloak-server
    default: true
    actions:
      onCreate:
        before:
          - description: kustomize build
            cmd: kubectl kustomize --enable-helm > generated.yaml
    manifests:
      - name: keycloak-server
        namespace: keycloak
        files:
          - generated.yaml
```

This PR enables helm inside kustomizations when `enableKustomizePlugins` is set, and adds tests to validate functionality

## Checklist before merging

### DIFF
--- a/src/internal/packager/kustomize/build.go
+++ b/src/internal/packager/kustomize/build.go
@@ -28,6 +28,8 @@ func Build(path string, destination string, kustomizeAllowAnyDirectory bool, ena
 
 	if enableKustomizePlugins {
 		buildOptions.PluginConfig = krustytypes.MakePluginConfig(krustytypes.PluginRestrictionsNone, krustytypes.BploUseStaticallyLinked)
+		buildOptions.PluginConfig.HelmConfig.Enabled = true
+		buildOptions.PluginConfig.HelmConfig.Command = "helm"
 	}
 
 	kustomizer := krusty.MakeKustomizer(buildOptions)
--- a/src/internal/packager/kustomize/build_test.go
+++ b/src/internal/packager/kustomize/build_test.go
@@ -1,0 +1,48 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+// Package template provides functions for templating yaml files.
+package kustomize
+
+import (
+	"os"
+	"path"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestBuild(t *testing.T) {
+	tests := []struct {
+		path                       string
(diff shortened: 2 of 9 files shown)


vincent-onebrief:
Added tests fail as expected with `build.go` changes commented out

src/internal/packager/kustomize/build.go
```diff
@@ -28,6 +28,8 @@ func Build(path string, destination string, kustomizeAllowAnyDirectory bool, ena
 
 	if enableKustomizePlugins {
 		buildOptions.PluginConfig = krustytypes.MakePluginConfig(krustytypes.PluginRestrictionsNone, krustytypes.BploUseStaticallyLinked)
+		buildOptions.PluginConfig.HelmConfig.Enabled = true
+		buildOptions.PluginConfig.HelmConfig.Command = "helm"
```
a1994sc, line 32: I half wonder if after #4350 is merged we should update this to use `zarf tools helm`.

Not a blocker, but a potential follow-up pr.

AustinAbro321: Yeah, #4350 is close but blocked by a bug in Helm that will be fixed in the next Helm release. 

I do think we should block this PR on that, the quasi Helm fork currently in Zarf is outdated, and I would like to avoid relying on the host binary, so that package builds are the same across devices

AustinAbro321:
Marking this as blocked by https://github.com/zarf-dev/zarf/pull/4350 which will fully embed Helm into the Zarf binary and should allow this to be done looking for the host binary

PR #4588, opened 2026-02-05
fix(actions): shell quote action wait bug

### LINKED ISSUE #4587: `.actions.wait.cluster` fails when jsonpath condition is surrounding by quotes on 0.71.0
### Environment
App version: 0.71.0

### Steps to reproduce
1. Have a deployment in kube-system called metrics-server with 1 replica available
2. Deploy a package with a jsonpath wait condition
```yaml
    actions:
      onDeploy:
        after:
          - description: Validate
            maxTotalSeconds: 60
            wait:
              cluster:
                kind: Deployments
                name: metrics-server
                namespace: kube-system
                condition: "'{.status.availableReplicas}'=1"
```

### Expected result

This will succeed assuming the condition is true

### Actual Result

Wait until timeout

### Severity/Priority

High

The simple fix would be to remove the quote from the condition. However, we should still fix this
```bash
  condition: "'{.status.availableReplicas}=1"
```

### Additional Context

Now that wait is no longer sent through a bash command, the quotes no longer gets taken away which leads to the logic in [isJSONPathWaitType](https://github.com/zarf-dev/zarf/blob/main/src/pkg/wait/wait.go#L26-L31), returning false now. This breaks deploy of uds-core, and probably requires a patch release

AustinAbro321:
## Related Issue

Fixes #4587 

This also fixes a bug in the oci-readonly-volume test and deletes the minikube e2e test. Minikube lets us test the docker driver, however I do not consider this necessary anymore as docker as a driver has been deprecated for over five years https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker/.

## Checklist before merging


AustinAbro321:
uds-cli tests are now passing on this branch https://github.com/defenseunicorns/uds-cli/actions/runs/21729841903/job/62681913634?pr=1312, which is how I originally found this bug

src/pkg/packager/testdata/find-images/pod-volume-image/deployment.yaml
```diff
@@ -29,7 +29,6 @@ spec:
           volumeMounts:
             - name: volume
               mountPath: /volume
-              subPath: dir
```
AustinAbro321, line 32: Seeing some failing tests / flakes here. Indeed there is no dir path in the zarf agent

Got this from the get-cluster-info step https://github.com/zarf-dev/zarf/actions/runs/21729739249/job/62684009437

```bash
          message: 'failed to mount image volume: ImageVolumeMountFailed: failed to
            ensure image subpath "dir" in "/run/containerd/io.containerd.grpc.v1.cri/image-volumes/43558b5af3071fb3b1640de71cce86afffea9c6647152e2536a3611823ffdd35/d21f3bade8e109aa748b36f4b90cca9a7d34002b6d3b989d93cf692583f4fc37":
            openat dir: no such file or directory'
          reason: CreateContainerError
```

PR #4592, opened 2026-02-06
feat(values): support values imports with namespacing

### LINKED ISSUE #4584: Support transient package values during component import

brandtkeller:
## Description

This adds support for namespacing an imported component artifacts templates and values/schema files on import to prevent silent collisions. 

## Related Issue

Fixes #4584

## Checklist before merging

### DIFF
--- a/src/api/v1alpha1/component.go
+++ b/src/api/v1alpha1/component.go
@@ -56,9 +56,6 @@ type ZarfComponent struct {
 
 	// List of resources to health check after deployment
 	HealthChecks []NamespacedObjectKindReference `json:"healthChecks,omitempty"`
-
-	// Prefix for namespacing all component
-	ValuesPrefix string `json:"valuesPrefix,omitempty"`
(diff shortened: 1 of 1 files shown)


brandtkeller:
Closing as an option we're not pursuing for values merge.

PR #4600, opened 2026-02-11
fix(bundle)!: bundle feature flag and version requirement

### LINKED ISSUE #4591: [bug] when trying to pull signed package created with 0.71.0 on 0.70.1 fails
### Environment
OS: Fedora
App version: cli: `0.70.1`, pkg: `0.71.0`

### Steps to reproduce
1. have Zarf version `0.70.1` installed
2. run:
    -  `zarf package pull oci://ghcr.io/colonel-byte/zarf/distro-k0s:v1.34.3-k0s.0-5-gd75e7fb`

### Expected result

Pull the above package locally

### Actual Result

Errors

```text
$ zarf package pull oci://ghcr.io/colonel-byte/zarf/distro-k0s:v1.34.3-k0s.0-5-gd75e7fb --verify=false --output-directory $TMPDIR
2026-02-06 12:17:40 INF pulling package name=ghcr.io/colonel-byte/zarf/distro-k0s:v1.34.3-k0s.0-5-gd75e7fb size=615.95 MBs
2026-02-06 12:17:40 INF finished pulling package layers duration=900ms
2026-02-06 12:17:42 WRN package signature could not be verified: error=package is signed but no verification material was provided (Public Key, etc.)
2026-02-06 12:17:42 INF writing package to disk path=/run/user/1000/tmp/zarf-package-distro-k0s-amd64-v1.34.3-k0s.0-5-gd75e7fb.tar.zst
2026-02-06 12:17:43 INF package downloaded successful path=/run/user/1000/tmp/zarf-package-distro-k0s-amd64-v1.34.3-k0s.0-5-gd75e7fb.tar.zst
```

brandtkeller:
## Description

This PR introduces a few breaking changes to largely resolve some of the symptoms of the problem of introducing new files into the package layout and backwards incompatibility. 

Given the scope of changes released for the bundle format signature - the compatibility and errors involved could have been improved by placing a feature flag in front of the bundle format process. This would allow the default workflows with new releases to still be backwards compatible. This allows for multiple releases to be introduced that contain the updated verification logic for bundles before we remove generation of the legacy signature entirely. 

### Future Proofing 

The root of the original issue is that new files outside of the checksums integrity check cannot exist in the package and be backwards compatible. This is a very niche case because in most cases the file should always be included in the checksums list. 

This PR introduces `ProvenanceFiles` to the build data for two purposes:
- It introduces optionality going forward to include a new file that all versions hereon forward would be able to handle files outside of the checksums (typically post-signing)
- It also provides an opportunity to create observability around these invisible files that are included in the package but not transparent in the definition

There is merit on concern on security of a MITM injection of an untracked file - but this is ever present and should be mitigated by a package being signed whereby the zarf.yaml cannot be modified. 

## Summary

- Move the bundle format behind a feature flag - not enabled by default for backwards compatibility
- Introduce provenance files for forward compatibility on the files tracked outside of the checksums file

## Related Issue

Fixes #4591

Relates to #

## Checklist before merging

### DIFF
--- a/src/pkg/packager/layout/package.go
+++ b/src/pkg/packager/layout/package.go
@@ -224,8 +224,11 @@ func (p *PackageLayout) SignPackage(ctx context.Context, opts utils.SignBlobOpti
 	// These are created after checksum generation and cannot be in checksums.txt.
 	// Listing them here allows integrity validation to dynamically exclude them
 	// without hardcoded knowledge of every possible signature file.
-	p.Pkg.Build.ProvenanceFiles = append(p.Pkg.Build.ProvenanceFiles, Signature)
-	if feature.IsEnabled(feature.BundleSignature) {
+	if !slices.Contains(p.Pkg.Build.ProvenanceFiles, Signature) {
+		p.Pkg.Build.ProvenanceFiles = append(p.Pkg.Build.ProvenanceFiles, Signature)
+	}
+
+	if feature.IsEnabled(feature.BundleSignature) && !slices.Contains(p.Pkg.Build.ProvenanceFiles, Bundle) {
 		p.Pkg.Build.ProvenanceFiles = append(p.Pkg.Build.ProvenanceFiles, Bundle)
 	}
 
@@ -361,7 +364,7 @@ func (p *PackageLayout) VerifyPackageSignature(ctx context.Context, opts utils.V
 	}
 
 	// Legacy signature found
-	l.Warn("bundle format signature not found: legacy signature is being deprecated. consider resigning this zarf package.")
+	l.Warn("bundle format signature not found: legacy signature is being deprecated. consider resigning this zarf package with the --features='bundle-signature=true' flag.")
 	opts.SigRef = signaturePath
 
 	opts.NewBundleFormat = false


Makefile
```diff
@@ -39,7 +39,7 @@ else
 	endif
 endif
 
-CLI_VERSION ?= $(if $(shell git describe --tags),$(shell git describe --tags),"UnknownVersion")
+CLI_VERSION ?= $(if $(shell git describe --tags),$(shell git describe --tags),"unset-development-only")
```
brandtkeller, line 42: This feels like the appropriate default - given that we have logic that checks for this constant?

src/pkg/packager/layout/package.go
```diff
@@ -626,11 +665,22 @@ func validatePackageIntegrity(pkgLayout *PackageLayout, isPartial bool) error {
 	if err != nil {
 		return err
 	}
-	// Remove files which are not in the checksums.
+	// zarf.yaml is the root of trust and is always excluded from checksums.
```
brandtkeller, line 657: The following logic has been updated - it now includes a strict check when unsigned and allowing for more dynamic SupplementFiles use when signed to help reduce any security concerns.

AustinAbro321: I like this, what do you think of a more specific name around supplementalfiles given that we are now explicit about it's use case? 

IntegrityFiles? VerificationFiles? You know the terminology in this space best, so I'll leave the specifics up to you

brandtkeller: That is a great thought and always the hardest to resolve 😆 - I like `VerificationFiles` - `ProvenanceFiles` offering some flexibility in the addition of other attestations in the event those also happened post-signing.

src/pkg/feature/feature.go
```diff
@@ -243,6 +244,13 @@ func init() {
 			Since:   "v0.64.0",
 			Stage:   Alpha,
 		},
+		{
+			Name:        BundleSignature,
```
brandtkeller, line 248: Bundle format will have been available for one release. Acknowledging that this reverts that to disabled by default. 

I don't believe replacing the layout constant for the legacy signature with the bundle format signature actually resolves anything - as the verification of a bundle signature won't function with the cosign utilities before bundles were integrated. 

The feature flag is a nice way for those doing signing today but maybe deploying with n-4 version of Zarf to have some flexibility to understand version constraints. When we fully remove the legacy signature and re-enable this by default - we'll need to mention that the minimum version for verification compatibility of packages signed with the bundle format is `v0.71.0`. 

It may be worth having a broader conversation about VersionRequirement optionality on Load as this is a great example of a minimum version requirement.

src/pkg/packager/layout/package.go
brandtkeller: acknowledging that the less-specific naming is a potential escape hatch that we may or may not need in the future.

AustinAbro321: provenance seems like a good choice. Always the chance that I'm proven wrong in hindsight, but seems like non provenance related files should always be able to go into the checksums.txt.

src/pkg/feature/feature.go
AustinAbro321: My argument against version requirements in load is that the intent of commands like pull, publish, and inspect is that they work generically. We did stumble upon a case where that's not true here, I'm not sure if that's something we're likely to see again. 

For something like deploy or remove which actively interact with the package definition, we know that in the future we will introduce fields that will change how those commands work on different versions hence the need for version requirements. These commands are more destructive, a deployment where the package deployed incorrectly can be quite expensive for users, whereas it failing to load or inspect will not be. Hence the higher friction for the former. 

Don't want to shut down the discussion though, happy to discuss this more, especially if there are situations I'm not considering around load.

brandtkeller: Agreed - I think this is fair and something I am comfortable removing as a requirement in this set of changes. I would like to keep this in mind as we encounter forward-compatible features and how we might set version requirements appropriately.

src/pkg/packager/layout/package.go
brandtkeller: Updated to `ProvenanceFiles` with a check for uniqueness in the list so that overwrite doesn't introduce duplicates).

Makefile
AustinAbro321: Yeah, it's pretty unlikely to be used given that it's only used when `git describe --tags` fails, but it's fine to be changed

AustinAbro321 (approved):

Makefile
brandtkeller: To elaborate - we have [version compare logic](https://github.com/zarf-dev/zarf/blob/main/src/internal/packager/requirements/requirements.go#L65-L68) that checks for the constant  but we have tests that run where version will still resolve to `UnknownVersion` which is not handled. Aligning on an expected non-tag default should close the gap here some.

PR #4606, opened 2026-02-16
feat: let zarf packages define requirements

### LINKED ISSUE #4605: Support Zarf Packages providing a REQUIREMENTS file
Many package systems have some sort of dependency tracking. When managing multiple zarf packages and deploying from/to many systems, it would be nice to allow developers to specify what is needed from their package. 
1) It would make it easier to track if two zarf packages have compatibility issues. 
2) While the zarf cmd action is very powerful, it is very easy to assume that certain tooling that exists on a developer's machine might not be present on all deploying systems.
3) This would make auditing much easier to track.

If a developer of a zarf package provides a REQUIREMENTS file and there is a known incompatibility during the deployment phase, the operator will be alerted. 

An example REQUIREMENTS file for a zarf package could be:
```
agent:
  tools:
    - name: yq
      version: ">= 4.40.5"
    - name: helm
      version: ">= 3.14.0"
  env:
    - name: HTTPS_PROXY
      required: false

cluster:
  crds:
    - name: gateways.gateway.networking.k8s.io
      version: ">= 1.0.0"
    - name: certificates.cert-manager.io
  resources:
    - apiVersion: v1
      kind: Namespace
      name: cert-manager
    - apiVersion: apps/v1
      kind: Deployment
      namespace: zarf
      name: zarf-injector
  packages:
    - name: iffy-package
      version: ">=1.4.0, !=1.6.2, !=1.6.3"
      reason: "our package is incompatible with iffy-package 1.6.2–1.6.3"
```

When an operator performs a `zarf package deploy`, the appropriate conditions should be checked for before proceeding. If an operator wants to proceed, then they may pass a `--skip-requirements-check` flag, with the understanding they do so at their own risk.

Every package we develop we add a "pre-flight" component which either involves running a bunch of onBefore cmds or some other script we design to check for compatibility. It would probably make sense for something like this at the zarf cli level.

Shasties:
## Description
Full disclosure upfront that Claude did all of the heavy lifting for building the tests. 

Currently, there is no visibility into what a zarf package needs to successfully deploy. This is burdensome on developers who have to use some sort of onBefore action or pre-req component to validate things, and it is burdensome to admins+operators who only learn partially through a deployment if a dependency is missing. This PR seeks to help bridge that gap and provide the option to insert REQUIREMENTS files into their zarf package.
 
I built and tested this on a windows machine, seems to function good enough in my local environment but obviously needs more eyes. Should probably be extended in the future with features like:
1) `zarf package list-requirements x` to display requirements of a package ahead of time.
2) Currently, developers specify the path to the tooling. This obviously isn't the right paradigm. Something like `zarf package deploy --alternate-req=custom-binary=/some/other/path` should be introduced so the developers focus on what tool is required for a package to work, while the operator can focus on how to make that tooling available. 
...

## Related Issue

Fixes #4605

### DIFF
--- a/examples/REQUIREMENTS/requirements.yaml
+++ b/examples/REQUIREMENTS/requirements.yaml
@@ -1,4 +1,4 @@
-# YAML file named REQUIREMENTS that specifies prereqs for package to have a clean deploy
+# YAML file that specifies prereqs for package to have a clean deploy
 # This is a very minimal & functional example
 # agent - specify utils & vars that must exist on the agent doing the zarf package deploy
 agent:
--- a/src/pkg/packager/layout/layout.go
+++ b/src/pkg/packager/layout/layout.go
@@ -16,6 +16,7 @@ const (
 	Checksums    = "checksums.txt"
 	ValuesYAML   = "values.yaml"
 	ValuesSchema = "values.schema.json"
+	Requirements = "REQUIREMENTS"
 
 	ImagesDir     = "images"
 	ComponentsDir = "components"
--- a/src/pkg/packager/layout/layout.go
+++ b/src/pkg/packager/layout/layout.go
@@ -16,7 +16,7 @@ const (
 	Checksums    = "checksums.txt"
 	ValuesYAML   = "values.yaml"
 	ValuesSchema = "values.schema.json"
-	Requirements = "REQUIREMENTS"
+	Requirements = "requirements.yaml"
 
 	ImagesDir     = "images"
 	ComponentsDir = "components"
--- a/examples/REQUIREMENTS/REQUIREMENTS
+++ b/examples/REQUIREMENTS/REQUIREMENTS
@@ -1,0 +1,34 @@
+# YAML file named REQUIREMENTS that specifies prereqs for package to have a clean deploy
+# Tooling that must exist on the agent doing the `zarf package deploy`
+agent:
+  tools:
+    - name: yq
+      version: ">= 4.40.5"
+    - name: custom_binary
+      version: ">= 3.14.0"
+      versionCommand: /path/to/custom_binary --flag-for-version
+      reason: "Used in our create onBefore scripts for X"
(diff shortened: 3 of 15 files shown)


src/pkg/packager/layout/layout.go
```diff
@@ -16,6 +16,7 @@ const (
 	Checksums    = "checksums.txt"
 	ValuesYAML   = "values.yaml"
 	ValuesSchema = "values.schema.json"
+	Requirements = "REQUIREMENTS"
```
a1994sc: I think that this file should be named `requirements.yaml` for 2 reasons:

- if this is going need to be a yaml file, the file extension should reflect that.
- we currently don't have any other files will all upper case

examples/REQUIREMENTS/requirements.yaml
```diff
@@ -0,0 +1,49 @@
...
+#    - name: certificates.cert-manager.io
+#  resources:
+#    - apiVersion: v1
+#      kind: Namespace
+#      name: cert-manager
+#    - apiVersion: apps/v1
+#      kind: Deployment
+#      namespace: zarf
+#      name: zarf-injector
+#  packages:
```
a1994sc, line 45: As an admin that has some custom packages, this check would be super nice to have

a1994sc (commented):
Two little comments.

I think that this is the right direction to go if we want to truly be the "Package Manager for Kubernetes", then we should package version validation before deploying the packages.

src/pkg/packager/layout/layout.go
Shasties: That's a fair point. When I was originally thinking about what the requirements file looked like, I wasn't sure if it should be json/yaml/other, so I left it as REQUIREMENTS. I can update that to requirements.yaml.

brandtkeller:
Closing as stale/not-planned - see #4605

PR #4608, opened 2026-02-17
feat(init): enable switching between nodeport and proxy mode

AustinAbro321:
## Description

This allows switching between nodeport and proxy mode. When a user runs a secondary `zarf init`, if no mode is specified, then the mode that this currently in Zarf state will be used. If a mode is specified, then that mode will be used instead

## Related Issue

Fixes https://github.com/zarf-dev/zarf/issues/4586

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/cluster_test.go
+++ b/src/pkg/cluster/cluster_test.go
@@ -355,7 +355,7 @@ func TestInitStateRegistryModeSwitch(t *testing.T) {
 			expected: state.State{
 				RegistryInfo: state.RegistryInfo{
 					RegistryMode: state.RegistryModeNodePort,
-					MTLSStrategy: state.MTLSStrategyZarfManaged,
+					MTLSStrategy: state.MTLSStrategyNone,
 					NodePort:     state.ZarfInClusterContainerRegistryNodePort,
 				},
 				InjectorInfo: state.InjectorInfo{Port: 0},
@@ -379,7 +379,7 @@ func TestInitStateRegistryModeSwitch(t *testing.T) {
 			expected: state.State{
 				RegistryInfo: state.RegistryInfo{
 					RegistryMode: state.RegistryModeNodePort,
-					MTLSStrategy: state.MTLSStrategyZarfManaged,
+					MTLSStrategy: state.MTLSStrategyNone,
 					NodePort:     30500,
 				},
 				InjectorInfo: state.InjectorInfo{Port: 0},


src/pkg/state/state.go
```diff
@@ -284,6 +284,29 @@ func (ri RegistryInfo) ShouldUseMTLS() bool {
 	return ri.MTLSStrategy != "" && ri.MTLSStrategy != MTLSStrategyNone
 }
 
+// CheckIfCredsChanged compares two RegistryInfo structs and returns true if any non-empty fields have changed
+func CheckIfCredsChanged(existing, given RegistryInfo) bool {
```
brandtkeller: This does not check NodePort - whereas it would have before. Which is intentional in the context of running zarf init and changing the mode from nodeport to proxy. 

But does/should a nodeport to nodeport upgrade with a nodeport change silently pass here now? Should we establish a check back in the caller for this case?

brandtkeller (changes requested):
One observation - otherwise documentation for switching from one to another as a supported operation could be helpful if not currently planned.

src/pkg/state/state.go
AustinAbro321: Yeah it does pass where it didn't before, but I think that's a benefit. I don't think there's an advantage to stopping the update of the nodeport. I believe the historical reason we did this is because it was more convenient to use `helpers.IsNotZeroAndNotEqual` instead of a custom function I created here

brandtkeller: That sounds fair. If the nodeport _is_ updated - do we ever hit a scenario where an artifact can not be pulled due to a change in the registry service? Everything GitOps pulls from the internal service (so we're fine there). Pods should be maintained by a cache on the node until another mutation request is sent to update the host.

AustinAbro321: Zarf and the kubelet won't because of the situation you described. If the user has something else pulling from the nodeport (discouraged outside of the kubelet case) it's possible, but since the user is opting into the nodeport change, I think that's reasonable.

src/pkg/state/state.go
```diff
@@ -284,6 +284,29 @@ func (ri RegistryInfo) ShouldUseMTLS() bool {
 	return ri.MTLSStrategy != "" && ri.MTLSStrategy != MTLSStrategyNone
 }
 
+// CheckIfRegistryCredsChanged compares two RegistryInfo structs and returns true if any non-empty fields have changed
+func CheckIfRegistryCredsChanged(existing, given RegistryInfo) bool {
```
brandtkeller: Slightly misleading function name. We include `Address` in the comparison but not `NodePort` or `Mode`. So it is not strictly creds nor is it a comparison for potential drift between two RegistryInfo Structs (See doccomment).

src/cmd/initialize.go
```diff
@@ -152,12 +150,12 @@ func (o *initOptions) run(cmd *cobra.Command, args []string) error {
...
-		if o.registryInfo.Address == "" {
-			o.registryInfo.RegistryMode = state.RegistryModeNodePort
-		} else {
-			o.registryInfo.RegistryMode = state.RegistryModeExternal
-		}
+	if o.registryInfo.RegistryMode == "" && o.registryInfo.Address != "" {
+		o.registryInfo.RegistryMode = state.RegistryModeExternal
+	}
+
+	if o.registryInfo.NodePort == 0 && o.registryInfo.RegistryMode == state.RegistryModeNodePort {
```
brandtkeller: Initially this feels duplicative of something that could occur in `FillInEmptyValues` but in tracing I see that this is needed for InitState's opts.RegistryInfo.NodePort != 0 guard to work on mode switches. 

This works in the given state but feels a little confusing. My only thought is that being able to discern empty from go default could be handy - but I don't know if we want to change the surface of those fields now... (thinking *int).

AustinAbro321: Yeah I think you are hitting on something important, it might have been a mistake to have nodeport and hostport have the same field, given that the default switches. Probably a bit confusing also. Going to look at adding a hostport field in a backwards compatible manner

src/pkg/state/state.go
AustinAbro321: Renamed

AustinAbro321:
Yeah, I hadn't noticed that it stays as MTLS, good catch. Took this away for now, as even if we support this in the future, I don't want users to start doing flows of initializing with registry proxy, then back to nodeport. 

Also definitely do need documentation. I have a separate docs PR #4622, which I think is close, but still needs some refinement

src/cmd/initialize.go
AustinAbro321: I actually thought about it more and decided to do keep nodeport for now, but make a separate issue to rename the field to port #4711 in a follow up.

Your point about the flow being confusing was still a great callout though. I reworked init state to make it more clear how things are handled when registry modes are changing, removing the need for this line in cmd.

PR #4610, opened 2026-02-18
fix(make): always run unit tests with -race flag

soltysh:
## Description

Whenever running unit tests we should be using `-race` flag, to detect any race problems early on. 

## Related Issue

None

## Checklist before merging


soltysh:
The 2nd commit (not mine) lgtm

Issue #4077, opened 2025-08-13
Introduce kstatus as an engine behind `zarf tools wait-for`


AustinAbro321:
### Is your feature request related to a problem? Please describe.
`zarf tools wait-for` currently works by shelling out  `kubectl wait`. `kubetl wait` is often good enough, however the UX leaves some room to be desired. Users have to know what the readiness condition or magic word is for certain resources. For instance deployments should be "available", while pods should be "ready". Other resources such as DaemonSets don't have a magic word and require waiting for a condition. With kstatus, we can assume readiness as a condition, which is likely what users would want 90+% of the time. 

Additionally, kstatus waits for resources to be fully reconciled which has advantages. For instance, kstatus will wait for all pods in a deployment to be ready, and all previous pods to be terminated. `kubectl wait` will only wait for all the new pods to be ready. 

Right now `zarf tools wait-for` does not accept an API version which is required by kstatus. `zarf tools wait-for` will be split into `zarf tools wait-for-resource` and `zarf tools wait-for-network` (#4550). `zarf tools wait-for` will always stay using `kubectl wait`,  while `zarf tools wait-for-resource` will default to kstatus, but use `kubectl wait` when a specific condition is set. 

### Describe the behavior you'd like

- **Given** I have a deployment podinfo -n podinfo
- -**When** I run a wait-for-resource command with the API version and without a condition `zarf tools wait-for-resource apps/v1 Deployment podinfo -n podinfo`
- **Then** The deployment is waited for with kstatus as the engine, and it waits for the resource to be ready
---
- **Given** I have a statefulset podinfo -n podinfo
- -**When** I run a wait-for-resource command with the API version, but ask for a specific condition `zarf tools wait-for-resource apps/v1 statefulset podinfo -n podinfo '{.status.availableReplicas}'=23 `
- **Then** The deployment is waited for with `kubectl wait` as the engine and it waits for the specific condition.
---
- **Given** I have two pods matching the selector app=podinfo
- -**When** I run a wait command with the API version, no condition set, and a selector `zarf tools wait-for-resource v1 pod app=podinfo ready -n podinfo`
- **Then** The deployment is waited for with `kstatus` as the engine. While kstatus does not natively handle selectors, Zarf will preform the necessary processing. 

### Additional context
This will also help the next schema version, which will require that API version is set under `.wait.cluster`, and only use `kubectl wait` as the engine when condition is set.

PR #4614, opened 2026-02-19
feat: split wait-for command

### LINKED ISSUE #4550: Separate `zarf tools wait-for` into two separate commands: `zarf tools wait-for-resource` and `zarf tools wait-for-network`
Deprecate the command `zarf tools wait-for` and create two new commands to replace it `zarf tools wait-for resource` and `zarf tools wait-for network`. It's confusing to have both in one command. Flags such as `--namespace` are only applicable to one. Additional flags that may be desired could add further confusion. 

Having separate command would also make for a more natural transition when defaulting to kstatus #4077. wait-for could always use kubectl wait, while wait-for-resource could require the APIVersion and only use kubectl wait when a specific condition is requested.

AustinAbro321:
## Description

This deprecates the `wait-for` command in favor of sub commands `zarf tools wait-for resource` and `zarf tools wait-for network`

## Related Issue

Fixes #4550
Fixes: #4077

## Checklist before merging

### DIFF
--- a/site/src/content/docs/commands/zarf_tools_wait-for_resource.md
+++ b/site/src/content/docs/commands/zarf_tools_wait-for_resource.md
@@ -24,7 +24,7 @@ zarf tools wait-for resource KIND [NAME | SELECTOR] [CONDITION] [flags]
 
 ```
 
-$ zarf tools wait-for resource deployment my-pod-name -n default                 #  wait for deployment to be reconciled (default condition)
+$ zarf tools wait-for resource deployment my-deployment -n default               #  wait for deployment to be reconciled (default condition)
 $ zarf tools wait-for resource po cool-pod-name -n cool                          #  wait for pod (using po alias) cool-pod-name in namespace cool to be reconciled
 $ zarf tools wait-for resource pod my-pod-name ready -n default                  #  wait for pod my-pod-name in namespace default to have the ready condition
 $ zarf tools wait-for resource pod app=podinfo -n podinfo                        #  wait for pod(s) with label app=podinfo in namespace podinfo to be reconciled
--- a/src/pkg/wait/wait.go
+++ b/src/pkg/wait/wait.go
@@ -149,7 +149,7 @@ func waitFor(ctx context.Context, restConfig *rest.Config, clientCfg clientcmd.C
 	}
 
 	if identifier == "" {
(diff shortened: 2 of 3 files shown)


site/src/content/docs/commands/zarf_tools_wait-for_resource.md
```diff
@@ -0,0 +1,64 @@
...
+
+```
+zarf tools wait-for resource KIND [NAME | SELECTOR] [CONDITION] [flags]
+```
+
+### Examples
+
+```
+
+$ zarf tools wait-for resource deployment my-pod-name -n default                 #  wait for deployment to be reconciled (default condition)
```
brandtkeller: nit: possibly confusing to use `my-pod-name` as the example name for a deployment

src/pkg/wait/wait.go
```diff
@@ -107,16 +147,20 @@ func ForResource(ctx context.Context, kind, identifier, condition, namespace str
...
+		return waitForReconciled(ctx, restConfig, dynamicClient, mapping, identifier, namespace, deadline)
+	}
+
+	return waitForResourceCondition(ctx, dynamicClient, condition, mapping.Resource.Resource, identifier, namespace, deadline)
 }
 
-// waitForAnyResource waits for at least one resource of the given kind to exist.
-func waitForAnyResource(ctx context.Context, dynamicClient dynamic.Interface, resource schema.GroupVersionResource, namespace string, deadline time.Time) error {
+// waitForAnyResourceInNamespace waits for at least one resource of the given kind to exist.
+func waitForAnyResourceInNamespace(ctx context.Context, dynamicClient dynamic.Interface, resource schema.GroupVersionResource, namespace string, deadline time.Time) error {
```
brandtkeller: nit: The name change feels slightly misleading given it works for cluster-scoped resources but I don't feel strongly it needs to change.

site/src/content/docs/commands/zarf_tools_wait-for_resource.md
AustinAbro321: Nice catch

src/pkg/wait/wait.go
AustinAbro321: I agree a bit misleading

PR #4616, opened 2026-02-19
fix(wait): ensure cluster is connectable in loop

AustinAbro321:
## Description

In #4567 we added a cluster to check to mimic the kubectl get on a loop behavior that would wait if the cluster was not available, however since the loop continues when a rest config is found, and not when a cluster connection is successful, it's possible to continue on a bad rest config then be in an unrecoverable state.

## Checklist before merging

### DIFF
--- a/src/pkg/wait/wait.go
+++ b/src/pkg/wait/wait.go
@@ -81,7 +81,7 @@ func ForResource(ctx context.Context, kind, identifier, condition, namespace str
 	err = wait.PollUntilContextTimeout(ctx, waitInterval, time.Until(deadline), true, func(_ context.Context) (bool, error) {
 		groupResources, err := restmapper.GetAPIGroupResources(discoveryClient)
 		if err != nil {
-			return true, fmt.Errorf("failed to get API group resources, retrying: %w", err)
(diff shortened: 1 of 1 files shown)


src/pkg/wait/wait.go
```diff
@@ -59,24 +60,28 @@ func ForResource(ctx context.Context, kind, identifier, condition, namespace str
...
-	}
-
 	// Wait for the resource kind to be resolvable (e.g. CRDs may not be registered yet).
 	var mapping *meta.RESTMapping
 	err = wait.PollUntilContextTimeout(ctx, waitInterval, time.Until(deadline), true, func(_ context.Context) (bool, error) {
 		groupResources, err := restmapper.GetAPIGroupResources(discoveryClient)
 		if err != nil {
-			l.Debug("failed to get API group resources, retrying", "error", err)
-			return false, nil
+			return true, fmt.Errorf("failed to get API group resources, retrying: %w", err)
```
brandtkeller: nit: given the distinction of the `true` boolean and the non-nil error which subsequently stops polling - this will intentionally stop retrying now correct? 


```suggestion
			return true, fmt.Errorf("failed to get API group resources: %w", err)
```

PR #4628, opened 2026-02-20
fix(wait): properly resolve kind when group conflicts between resources

AustinAbro321:
## Description

There is a bug in Zarf where running a wait command with a kind where there exists another kind of that name does not get properly resolved. For instance, `zarf tools wait-for gateways.networking.istio.io admin-gateway` would not work as it would try to resolve the default from Kubernetes gateway resource. 

Versions don't matter in this context as they're considered the same object by Kubernetes, I tested this out with kubectl wait

## Checklist before merging


brandtkeller (approved):
lgtm

PR #4641, opened 2026-02-24
feat(publish): support for tag specification

### LINKED ISSUE #4609: Non 1:1 tag-to-version relationship
Our team is promoting zarf packages that have been built and tested on feature branches rather than rebuilding them on main.  To avoid tag collisions when multiple feature branch pipelines are run at once, we tag packages built from feature branches with the commit sha that resulted in their creation.  Then, if the feature branch is merged, we figure out what the last commit was before the merge to main  that kicked off the pipeline was, and we promote that commit sha tag to the release version (e.g. aaaacccc5555bbb -> 1.0.0).  However, this results in zarf package list  versions showing aaaacccc5555bbb , which is unhelpful without doing some digging.

The hope was we could --set VERSION to the real version during feature branch build, but still tag according to commit sha, thus resulting in the promoted package containing the desired version.  However, because the tag is always sourced from metadata.version , we have no ability to tag independently.

- **Given** a zarf.yaml exists with `metadata.version: 1.0.0`
- **When** `zarf package create --tag aaaacccc5555bbb` and `zarf package publish ./zarf-package-example-amd64-aaaacccc5555bbb.tar.zst` (or `zarf package publish ./zarf-package-example-amd64-1.0.0.tar.zst`, the name of the package on disk doesn't matter so much) are run
- **Then** I see `myregistry.com/myproject/mypackage:aaaacccc5555bbb` in my registry.  When I deploy `mypackage`, I see `1.0.0` as the deployed version of mypackage.

1. `oras copy` to create the desired commit sha tag.  This has the consequence of opening the door to tag collisions, as multiple feature branches may push at once, and its unknown which will be copied from to create a commit sha tag.

1. local build of package on disk, and then `oras push` to the desired tag.  This might work but may have consequences downstream and also goes outside of zarf to accomplish the ask.

brandtkeller:
## Description

Adds support for a `--tag | -t` flag on `package publish` to override the package metadata.version tag that is statically defined. 

This purposefully does not add the tag to create at this time. 

Consideration could be made for a list of tags but that adds more complexity than should be needed to a publish activity. Additionally "flavor" could be an option on the zoci options but I deferred to keeping that explicit. 

## Related Issue

Fixes #4609

## Checklist before merging

### DIFF
--- a/src/pkg/zoci/utils.go
+++ b/src/pkg/zoci/utils.go
@@ -36,16 +36,18 @@ func ReferenceFromMetadataWithOptions(registryLocation string, pkg v1alpha1.Zarf
 	}
 	registryLocation = strings.TrimPrefix(registryLocation, helpers.OCIURLPrefix)
 
-	// Use the explicit tag if provided, otherwise fall back to the package version.
+	// Use the explicit tag if provided
+	// do not include flavor if provided
 	tag := pkg.Metadata.Version
 	if opts.Tag != "" {
 		tag = opts.Tag
+	} else {
+		if pkg.Build.Flavor != "" {
+			tag = fmt.Sprintf("%s-%s", tag, pkg.Build.Flavor)
+		}
 	}
 
 	raw := fmt.Sprintf("%s%s:%s", registryLocation, pkg.Metadata.Name, tag)
-	if pkg.Build.Flavor != "" {
-		raw = fmt.Sprintf("%s-%s", raw, pkg.Build.Flavor)
-	}
 
 	ref, err := registry.ParseReference(raw)
 	if err != nil {
--- a/src/pkg/packager/publish.go
+++ b/src/pkg/packager/publish.go
@@ -89,6 +89,10 @@ func PublishFromOCI(ctx context.Context, src registry.Reference, dst registry.Re
 		Retries:        opts.Retries,
 	}
 
+	if src.Reference != dst.Reference {
+		publishOptions.Tag = dst.Reference
+	}
+
 	// Execute copy
 	err = zoci.CopyPackage(ctx, srcRemote, dstRemote, publishOptions)
 	if err != nil {
--- a/src/pkg/packager/publish_test.go
+++ b/src/pkg/packager/publish_test.go
@@ -471,6 +471,7 @@ func TestPublishCopyTag(t *testing.T) {
 		name             string
 		packageToPublish string
 		opts             PublishPackageOptions
+		dstTag           string
 	}{
 		{
 			name:             "Publish package",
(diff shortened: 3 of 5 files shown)


src/pkg/zoci/utils.go
```diff
@@ -16,6 +16,18 @@ import (
...
+type ReferenceFromMetadataOptions struct {
+	Tag string
+}
+
+// ReferenceFromMetadataWithOptions returns a reference for the given metadata with optional overrides
+func ReferenceFromMetadataWithOptions(registryLocation string, pkg v1alpha1.ZarfPackage, opts ReferenceFromMetadataOptions) (registry.Reference, error) {
+	// Explicit requirement for version in order to publish
 	if len(pkg.Metadata.Version) == 0 {
 		return registry.Reference{}, errors.New("version is required for publishing")
 	}
```
AustinAbro321, line 33: Should we change this requirement to version or tag?

brandtkeller: I left Version as a strict requirement. Anytime we are publishing we have required a version so I believe that maintains the stance. Tag is optional.

AustinAbro321: I'm fine with keeping the requirement, I'd rather encourage users tag their packages through the metadata version. I called attention to this since the reason we added this requirement was because otherwise we wouldn't have a tag. Now that we can have an explicit tag it's no longer a technical requirement

brandtkeller: fair point - it is no longer a technical requirement - should it still be a publishing requirement? 

Or said another way - do we want to allow a package to be published that does not have a `metadata.version`?

src/pkg/zoci/utils.go
AustinAbro321, line 1: in zoci/copier.go we still set the tag to the src repo, meaning `--tag` isn't respected for oci to oci.

src/pkg/zoci/utils.go
AustinAbro321: Yeah, I wanted to raise it for consideration, but thinking it over I'd probably still lean towards keeping it as a requirement. 

Thinking through the workflow, I'd rather have people set version with `.metdata.Version` than with `--tag` as I feel it's a bit easier to track. However, if we give an error message telling users, that they have to set version or give a tag, then they're more likely to use a tag since it's least resistance, so I'm cool with keeping the requirement.

src/pkg/zoci/utils.go
brandtkeller: Good catch! updated to have packager be the glue between CLI override and updating zoci in a way that doesn't break the behavior.

PR #4655, opened 2026-02-28
fix: prevent panic on double call to `Tracker.StopReporting`

joonas:
## Description

`StopReporting` guarded the `close(tt.stopReports)` call with a nil check, but a closed channel is not nil — so a second call passed the guard and panicked. Every current call site uses `defer`, making a double call unlikely today, but the type's public API was unsafe.

Wrap the close in `sync.Once` so only the first call closes the channel. Subsequent calls skip the close and proceed to `wg.Wait`, which is safe to call repeatedly — it returns immediately once the counter is zero.

## Checklist before merging

### DIFF
--- a/src/pkg/images/progress.go
+++ b/src/pkg/images/progress.go
@@ -87,7 +87,7 @@ func (tt *Tracker) StartReporting(ctx context.Context) {
 	}()
 }
 
-// StopReporting stops the reporting goroutine. It is safe to call multiple times.
+// StopReporting stops the reporting goroutine.
 func (tt *Tracker) StopReporting() {
 	tt.stopOnce.Do(func() {
 		if tt.stopReports != nil {


src/pkg/images/progress.go
```diff
@@ -87,11 +87,13 @@ func (tt *Tracker) StartReporting(ctx context.Context) {
 	}()
 }
 
-// StopReporting stops the reporting goroutine
+// StopReporting stops the reporting goroutine. It is safe to call multiple times.
```
AustinAbro321:
```suggestion
// StopReporting stops the reporting goroutine.
```
nit, but I think it's generally the expectation that multiple calls will be safe, would prefer not to mention that.

joonas: That's fair, I agree!

PR #4656, opened 2026-02-28
fix: close chunk file descriptors per iteration in `SplitFile`

joonas:
## Description

`SplitFile` used `defer dstFile.Close()` inside a `for` loop, but defers only execute on function return — not at the end of each iteration. When splitting a large file into many chunks, every destination file handle stayed open simultaneously until the function exited.

This meant that a 10 GB file split into 100 MB chunks would hold 100 file descriptors open at once, risking EMFILE ("too many open files") on systems with lower FD limits.

The change wraps the loop body in an immediately-invoked closure so that the deferred close runs at the end of each iteration. This matches the pattern already used in `ReassembleFile` (line 185) and reduces the steady-state FD usage from O(chunks) to O(1) - only the source file and a single destination chunk are open at any time.

## Checklist before merging

### DIFF
--- a/src/internal/split/split.go
+++ b/src/internal/split/split.go
@@ -63,47 +63,60 @@ func SplitFile(ctx context.Context, srcPath string, chunkSize int) (_ string, er
 
 	hash := sha256.New()
 	fileCount := 0
-	// TODO(mkcp): The inside of this loop should be wrapped in a closure so we can close the destination file each
-	//   iteration as soon as we're done writing.
 	for {
 		path := fmt.Sprintf("%s.part%03d", srcPath, fileCount+1)
-		dstFile, err := os.OpenFile(path, os.O_CREATE|os.O_RDWR|os.O_TRUNC, 0644)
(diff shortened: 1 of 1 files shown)


src/internal/split/split_fd_leak_test.go
AustinAbro321, line 1: I think I'd lean towards deleting this test. It's nice that we can run it against `src/internal/split/split.go` on main to prove that there was a problem, however it does feel like we're testing how the go programming language works rather than our function. I'm not particularly worried that we would introduce code that would regress the function in a way that this would catch.

joonas: Totally fair, the goal is definitely not to test the language, but rather ensure that we're not leaking FDs in the new implementation. Admittedly that's a little bit difficult (as evidenced by the test) to demonstrate, so I suppose as long as the pre-existing tests don't break we should consider the issue addressed.

I removed this file.

PR #4657, opened 2026-02-28
fix: close leaked file handles in `pull_test.go` HTTP handlers

joonas:
## Description

The test HTTP handlers in `TestPull`, `TestPullUncompressed`, and `TestPullUnsupported` each call `os.Open` to serve a package file but never close the resulting handle. The file descriptor leaks for the lifetime of the test process — one per request to the `httptest.Server`.

Add `defer file.Close()` after the open-error check in all three handlers. Since each handler invocation is its own function call, the `defer` fires at exactly the right time — when the handler returns after `io.Copy` completes.

## Checklist before merging

### DIFF
--- a/src/pkg/packager/pull_test.go
+++ b/src/pkg/packager/pull_test.go
@@ -29,7 +29,7 @@ func TestPull(t *testing.T) {
 			rw.WriteHeader(http.StatusInternalServerError)
 			return
 		}
-		defer file.Close()
+		defer file.Close() //nolint:errcheck // best-effort cleanup in test handler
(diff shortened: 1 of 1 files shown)


joonas:
> LGTM, was curious if it'd make sense to check the error here, but these test servers run as gorountines so we don't want to use `require.NoError` in them, and I don't think it's worth passing a errors through a channel to check.

Yeah, I initially used `require.NoError` inside an anonymous function block for the `defer`, but that felt unnecessary given the context these are for.

PR #4658, opened 2026-02-28
fix: preserve error chains by using `%w` instead of `%s`

joonas:
## Description

Multiple `fmt.Errorf` call sites wrapped errors using `%s` with `err.Error()`, which converts the error to a plain string and severs the error chain. This means `errors.Is` and `errors.As` cannot match on any cause in the chain — callers lose the ability to programmatically inspect underlying errors.

The root cause is twofold: the `lang` package format constants (e.g., `ErrDownloading`, `ErrWritingFile`, `ErrCreatingDir`) were defined with a `%s` placeholder for the error argument, and every call site passed `err.Error()` to match. Several inline format strings outside `lang` had the same pattern.

Update all 8 affected `lang` constants in `english.go` to use `%w` as the error placeholder, and change every call site (20 occurrences across 8 files) from `err.Error()` to `err`. Also convert 4 inline format strings that used `%s` + `err.Error()` directly — in `repository.go`, `tunnel.go`, `remove.go`, and `assemble.go` — to use `%w` + `err`.

The `lang` constants that already used `%w` (e.g., `ErrUnmarshal`) and call sites that already wrapped correctly (e.g., `helpers.CreatePathAndCopy` callers using `%w`) are left unchanged.

## Checklist before merging

### DIFF
--- a/src/cmd/zarf_tools.go
+++ b/src/cmd/zarf_tools.go
@@ -608,7 +608,7 @@ func (o *genKeyOptions) run(cmd *cobra.Command, _ []string) error {
 			Message: lang.CmdToolsGenKeyPrompt,
 		}
 		if err := survey.AskOne(prompt, &password); err != nil {
-			return nil, fmt.Errorf(lang.CmdToolsGenKeyErrUnableGetPassword, err.Error())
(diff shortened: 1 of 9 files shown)


brandtkeller (approved):
lgtm

PR #4664, opened 2026-03-02
feat: add retries on create operations

mjnagel:
## Description

Addresses the linked issue by using `retry-go` (already in codebase) to add retries for:
- HTTP downloads for manifests, files, data injections, and chart values
- Helm chart downloads
- Image pulls (specifically `oras.Copy` where I've seen failures before)

This PR also:
- Honors `Retry-After` header on 429 responses for HTTP downloads
- Uses shared retry defaults from `config` (3 attempts, 500ms base delay, 8s max, exponential backoff)

## Related Issue

Fixes https://github.com/zarf-dev/zarf/issues/4470

## Checklist before merging

### DIFF
--- a/src/pkg/utils/network.go
+++ b/src/pkg/utils/network.go
@@ -24,6 +24,14 @@ import (
 	"github.com/zarf-dev/zarf/src/pkg/logger"
 )
 
+// retryAfterDuration is returned on a 429 so the custom DelayType can use it
+// instead of stacking on top of the normal backoff.
+type retryAfterDuration time.Duration
+
+func (d retryAfterDuration) Error() string {
+	return fmt.Sprintf("rate limited (HTTP 429), retry after %s", time.Duration(d))
+}
+
 func parseChecksum(src string) (string, string, error) {
 	atSymbolCount := strings.Count(src, "@")
 	var checksum string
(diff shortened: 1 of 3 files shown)


src/pkg/utils/network.go
```diff
@@ -104,7 +122,21 @@ func httpGetFile(ctx context.Context, url string, destinationFile *os.File) (err
 
 	// Check server response
 	if resp.StatusCode != http.StatusOK {
-		return fmt.Errorf("bad HTTP status: %s", resp.Status)
+		if resp.StatusCode == http.StatusTooManyRequests {
+			if d := parseRetryAfter(resp.Header.Get("Retry-After")); d > 0 {
+				l.Info("rate limited, waiting before retry", "url", url, "retry_after", d)
```
brandtkeller: Not blocking - I wonder if there is a reasonable cutoff we want to apply for UX considerations. If the retry-after is some silly long value or if the value was purposefully malicious.

brandtkeller (changes requested):
Minor comment otherwise I don't have too much other feedback to provide. Adds retry functionality on create which feels appropriate for CI systems building large packages or dealing with intermittent connectivity.

src/pkg/utils/network.go
mjnagel: That's a good callout, let me think about what a sane default might be here to check (in alignment with the rest of the code and other timeouts).

mjnagel: Added a 60s max here, if that is surpassed we just bail out with an unrecoverable error on the retry. Open to other values for sure - this seems like a reasonable value to handle some shorter rate limiting while not impacting UX? If someone is hitting a longer rate limiting period it might be nice to surface that to them since it could be an indication of something else going on (missing auth to bypass limits, etc).

AustinAbro321: I think we'd want to do something similar to this example, https://github.com/avast/retry-go/blob/5bccbfa9340dfe6609f4ecfce30c971e2756c796/examples/delay_based_on_error_test.go#L5. This way we don't stack the retry after wait and the regular delay wait.

src/pkg/utils/network.go
```diff
@@ -54,18 +57,33 @@ func DownloadToFile(ctx context.Context, src, dst string) (err error) {
...
+		retry.DelayType(retry.BackOffDelay),
+		retry.LastErrorOnly(true),
+		retry.Context(ctx),
+		retry.OnRetry(func(n uint, err error) {
+			l.Warn("retrying download",
+				"attempt", n+1,
+				"max_attempts", config.ZarfDefaultRetries,
+				"url", src,
+				"error", err,
+			)
```
AustinAbro321: A bit verbose, but we should do something like this in these blocks so users don't get a message that they're retrying after we've already done our last retry
```go
if opts.Retries > 1 && n+1 < uint(opts.Retries) {
```

mjnagel: Added - good call.

src/pkg/utils/network.go
mjnagel: Great callout - I think I addressed this with a pretty lightweight approach and aligned with that example?

PR #4666, opened 2026-03-02
feat!: give OCI package layers appropriate media type

### LINKED ISSUE #3009: Preserve appropriate `mediaType` on package layers

Currently Zarf uses a generic `application/vnd.zarf.layer.v1.blob` to describe every layer in the OCI manifest. This is confusing because many of these layers point to blobs with well-defined `mediaTypes` already.

#### The Easy Stuff

- [ ] `images/oci-layout`: `application/vnd.oci.layout.header.v1+json`
- [ ] `images/index.json`: `application/vnd.oci.image.index.v1+json`
- [ ] `images/blobs/sha256/*` (as appropriate):
  - `application/vnd.oci.image.manifest.v1+json`
  - `application/vnd.oci.image.config.v1+json`
  - `application/vnd.oci.image.layer.v1.tar+gzip`
- [ ] `zarf.yaml`: => (deprecate?) alternatives:
  - rename `zarf.json` using `application/vnd.zarf.package.config.v1+json`
  - alternatively, the full `ZarfPackageConfig` could be included in the image `config` (we already have a custom `application/vnd.zarf.config.v1+json`)
- [ ] `checksums.txt`: => (deprecate?)
  - this just seems like a plaintext version of the image manifest embedded into the image manifest?
  - if we keep it, define something like `application/vnd.zarf.package.checksums.v1+json`

#### The Harder Stuff

- [ ] `sboms.tar`: => (attestation(s) using sigstore bundle format) `application/vnd.dev.sigstore.bundle.v0.3+json`
  - uses `dev.sigstore.bundle.predicateType` annotation for disambiguating SBOM formats (ex. `sbom.spdx.json` or `https://cyclonedx.org/schema`)
  - Full spec (https://github.com/sigstore/cosign/pull/3622): https://github.com/sigstore/cosign/blob/main/specs/BUNDLE_SPEC.md
  - Trivy support: https://aquasecurity.github.io/trivy/v0.40/docs/attestation/sbom/
- [ ] `components/<component>.tar`: => (separate layer per chart, use Helm-native OCI format):
  - `application/vnd.cncf.helm.config.v1+json`
  - `application/vnd.cncf.helm.chart.content.v1.tar+gzip`
  - `application/vnd.cncf.helm.chart.provenance.v1.prov`
  - `application/tar+gzip` (legacy fallback)

### Additional context

Overloading `application/vnd.zarf.layer.v1.blob` makes the existing OCI manifests difficult to consume by other tooling. Preserving the `mediaTypes` for these layers makes it much easier to disambiguate the blobs.

For example, when implementing container image scanning, it would be great if we could just look for `application/vnd.oci.image.manifest.v1+json` layers as a starting point.

Along the same lines, the "harder stuff" would bring us inline with other CNCF/OpenSSF tooling and ultimately reduce the plumbing necessary to use things like `trivy` with Zarf packages.

AustinAbro321:
## Description

This gives layers will well known names an appropriate media type

## Related Issue

Fixes #3009

## Checklist before merging

### DIFF
--- a/src/pkg/zoci/push.go
+++ b/src/pkg/zoci/push.go
@@ -200,11 +200,11 @@ func buildBlobMediaTypes(pkgDir string) (map[string]string, error) {
 		}
 		manifestBytes, err := os.ReadFile(filepath.Join(pkgDir, layout.ImagesBlobsDir, hex))
 		if err != nil {
-			continue
+			return nil, err
 		}
 		var mf ocispec.Manifest
 		if err := json.Unmarshal(manifestBytes, &mf); err != nil {
-			continue
+			return nil, err
 		}
 		if mf.Config.MediaType != "" {
 			result[layout.ImagesBlobsDir+"/"+mf.Config.Digest.Encoded()] = mf.Config.MediaType
--- a/src/pkg/zoci/push.go
+++ b/src/pkg/zoci/push.go
@@ -6,10 +6,12 @@ package zoci
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"maps"
 	"os"
+	"path/filepath"
 	"sort"
 	"time"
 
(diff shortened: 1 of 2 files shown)


AustinAbro321:
Closing this as it is actually not possible to set the media type to application/vnd.oci.image.manifest.v1+json since oras-go will push layers with a manifest layer media type differently. oras-go will assume that they should be actual manifest images in the registry rather than blobs as they are within Zarf.

This can be seen in https://github.com/zarf-dev/zarf/pull/4666 where you get the following error.

--- FAIL: TestAssembleLayers (1.62s)
    --- FAIL: TestAssembleLayers/Assemble_layers_from_a_package (1.62s)
        pull_test.go:76: 
                Error Trace:    /home/austin/code/zarf/src/pkg/zoci/pull_test.go:76
                Error:          Received unexpected error:
                                could not push package: publish failed: failed to perform "Push" on destination: PUT "http://localhost:33761/v2/my-namespace/basic-pod/manifests/sha256:5547d2d19f7dc1845873ddef0624ba101a2ffaf1e76828c56b37c45ef99c1040": response status code 400: manifest blob unknown: blob unknown to registry: sha256:1c4eef651f65e2f7daee7ee785882ac164b02b78fb74503052a26dc061c90474; manifest blob unknown: blob unknown to registry: sha256:da324ac903c3287a9ab7f12d10fea0177251ca5d1aae156b293f042a722c414d
I don't think we should alter other errors, as it'd likely be confusing to have some image layers with media types and others without it. Open to other thoughts though.

I would still agree that the full ZarfPackageConfig could be included in the image config instead of as a layer

PR #4670, opened 2026-03-03
feat(agent): add mutate-if-exists conditional image mutation

### LINKED ISSUE #4640: zarf agent automatic redirect according to image existence
When running Knative Services on a cluster managed by Zarf, image mutation currently applies uniformly to all containers in a Pod.

A Knative Service Pod always contains:
* A platform sidecar container (queue-proxy) owned and versioned by Knative
* One or more user application containers, owned by the application developer

In our environment:
* The queue-proxy image is bundled and hosted in the Zarf registry
* The user application image must be pulled from an on-prem/private registry (not Zarf-managed)

However, Zarf’s current mutation behavior rewrites all container image references in the Pod, causing one of the following failures:
* The user container image is rewritten to the Zarf registry where it does not exist
* Or Zarf must be disabled for the namespace, preventing the queue-proxy from being pulled from the Zarf bundle

**Given** Opt-in Mutation Mode to "only existing images"
**And** the namespace is managed by zarf
**When** the Pod is admitted
**Then** the Zarf agent should:
* Rewrite the image reference only if the image exists in the Zarf registry
* Leave the image unchanged if it does not exist in the Zarf registry
* the agent can (optionally) cache the results using a configurable TTL to minimize API calls to the registry.
* images with latest tags should always pull from the containers registry.

### Additional context
* This feature will also work well to reduce friction in hybrid envs. where the initial services/platform are deployed with zarf and app level images should be pulled from other private/public registries.

orinvast:
Add support for conditional image mutation using the zarf.dev/agent=mutate-if-exists namespace label. When enabled, only images that exist in the Zarf registry are mutated, while others keep their original references.

Changes:
- Add checkNamespaceMutationBehavior to detect mutate-if-exists label
- Implement imageExistsInRegistry to check registry before mutation
- Add namespace read permissions to ClusterRole
- Fix in-cluster registry address for agent pod
- Add user documentation to init-package.mdx
- Fix linting issues in pods_mutate_if_image_exists.go

## Description

...

## Related Issue

Fixes #4640

Relates to #4640

## Checklist before merging

### DIFF
--- a/packages/zarf-agent/chart/templates/clusterrole.yaml
+++ b/packages/zarf-agent/chart/templates/clusterrole.yaml
@@ -10,3 +10,10 @@ rules:
     verbs:
       - get
       - list
+  - apiGroups:
+      - ""
+    resources:
+      - namespaces
+    verbs:
+      - get
+      - list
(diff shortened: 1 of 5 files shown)


AustinAbro321:
We've decided to close this, reasoning in https://github.com/zarf-dev/zarf/issues/4640

PR #4673, opened 2026-03-03
fix: retry on error for waitForSingleResourceMatchingCriteria

wcrum:
## Description

Retry on single resource fail to list.

## Related Issue

Fixes #

Relates to #

## Checklist before merging


AustinAbro321:
I believe we determined that this was not the root cause of your error, as far as I can tell, after we validate that the resource type exists, failing to list is a terminal error. Feel free to re-open otherwise

PR #4675, opened 2026-03-03
feat: use Zarf Package Config as image config

AustinAbro321:
## Description

This adds a Zarf Package Config as the image config item. This is more in line with the image spec and tools like Helm. We will still grab the zarf.yaml from the oci layer for backwards compatibility, but this should make it more convenient for consumers of Zarf OCI objects to grab the package config. 

Also fixes and tests for a subtle bug in `PushPackage` where it was always returning an empty descriptor. No Zarf callers used this, but some sdk consumers might.

## Related Issue

Relates to #3009

## Checklist before merging

### DIFF
--- a/src/pkg/zoci/push.go
+++ b/src/pkg/zoci/push.go
@@ -137,9 +137,10 @@ func (r *Remote) PushPackage(ctx context.Context, pkgLayout *layout.PackageLayou
 			trackedRemote.StartReporting(ctx)
 			defer trackedRemote.StopReporting()
 
-			publishedDesc, err = oras.Copy(ctx, src, root.Digest.String(), trackedRemote, "", copyOpts)
-			if err != nil {
-				return err
+			var copyErr error
+			publishedDesc, copyErr = oras.Copy(ctx, src, root.Digest.String(), trackedRemote, "", copyOpts)
+			if copyErr != nil {
+				return copyErr
 			}
 
 			return r.OrasRemote.UpdateIndex(ctx, r.Repo().Reference.Reference, publishedDesc)


src/pkg/zoci/push.go
```diff
@@ -107,9 +108,16 @@ func (r *Remote) PushPackage(ctx context.Context, pkgLayout *layout.PackageLayou
 		func() error {
 			l.Info("pushing package to registry", "destination", r.Repo().Reference.String(),
 				"architecture", pkgLayout.Pkg.Build.Architecture, "size", utils.ByteFormat(float64(totalSize), 2))
-			manifestConfigDesc, cfgErr := r.OrasRemote.CreateAndPushManifestConfig(ctx, annotations, ZarfConfigMediaType)
-			if cfgErr != nil {
-				return cfgErr
+			if annotations[ocispec.AnnotationTitle] == "" {
+				return fmt.Errorf("invalid annotations: please include value for %q", ocispec.AnnotationTitle)
+			}
```
brandtkeller: should this empty title check by moved outside of the retry logic?

src/pkg/zoci/push.go
```diff
@@ -107,9 +108,16 @@ func (r *Remote) PushPackage(ctx context.Context, pkgLayout *layout.PackageLayou
...
+			manifestConfigBytes, err := json.Marshal(pkgLayout.Pkg)
```
brandtkeller, line 115: nit: `err` here is going to shadow the named return. Perhaps `manifestErr` or similar?

brandtkeller:
Do you have any thoughts for when we might implement this in our baseline? 

IE check and see if the config blob contains the zarf.yaml - use it - otherwise fall back to the appropriate layer?

src/pkg/zoci/push.go
AustinAbro321: Yes

src/pkg/zoci/push.go
AustinAbro321: IMO shadowing is fine for the named return error when the error is immediately returned. At least that's the pattern we've stuck to so far across the code base.

AustinAbro321:
@brandtkeller Not planning on grabbing the config within Zarf at any point. By keeping the zarf.yaml as a separate file older Zarf versions can stay backwards compatiable when a new schema is introduced using the strategies laid out by https://github.com/zarf-dev/proposals/pull/49.

src/pkg/zoci/push.go
brandtkeller: fair - it's an implicit behavior with no side effects in this specific case. The logic that follows this is more explicit with regards to the closure but I wouldn't block on this.

PR #4683, opened 2026-03-04
feat(connect): create zarf connect resource sub-command

### LINKED ISSUE #2716: Untangle Zarf connect command
The Zarf connect command is confusing because flags can be set that conflict with the arguments. 

For example, `zarf connect doom --name=dbz --remote-port=9001 --type=svc --namespace=other-world` is a valid command even if the service labeled with `zarf.dev/connect-name: doom` has nothing to do with the flags given because the options are ignored when a target is given after `zarf connect`. 

Zarf connect should still have these flags, for now as to not cause a breaking change, but they should be deprecated and hidden.

Zarf should introduce a new command `zarf connect resource` and it should have required flags name, namespace and remote-port. As well as an optional flags type which defaults to svc and local-port which defaults to a random available port.

- **Given** I have a svc called doom in the dos-games namespace serving on port 8000
- **When** I run `zarf connect resource --name doom --namespace dos-games --remote-port 8000`
- **Then** my browser is opened to a random local port on my local machine.

AustinAbro321:
## Related Issue

Fixes #2716

## Checklist before merging

### DIFF
--- a/src/cmd/connect.go
+++ b/src/cmd/connect.go
@@ -96,8 +96,6 @@ func (o *connectOptions) run(cmd *cobra.Command, args []string) error {
 	return waitForTunnel(ctx, tunnel, o.open)
 }
 
-// waitForTunnel handles the shared logic of logging the tunnel URLs, optionally opening a browser,
-// and waiting for user interrupt or tunnel error.
 func waitForTunnel(ctx context.Context, tunnel *cluster.Tunnel, openBrowser bool) error {
 	l := logger.From(ctx)
 	urls := tunnel.FullURLs()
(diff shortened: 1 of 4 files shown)


src/cmd/connect.go
```diff
@@ -85,34 +93,78 @@ func (o *connectOptions) run(cmd *cobra.Command, args []string) error {
...
 		return fmt.Errorf("lost connection to the service: %w", err)
 	}
 }
 
+type connectResourceOptions struct {
+	open bool
+	zt   cluster.TunnelInfo
+}
+
+func newConnectResourceCommand() *cobra.Command {
```
brandtkeller, line 128: I think documenting the `Long` message with examples will be appreciated by those unfamiliar with trying a portforward with zarf. I was trying it blind (before reviewing) and still caught a few rough edges.

brandtkeller (changes requested):
Besides some supporting docs I think this is ready to ship.

src/cmd/connect.go
AustinAbro321: Added some examples

PR #4684, opened 2026-03-04
fix(cmd): honor --insecure-skip-tls-verify for registry tools

willswire:
## Description

`zarf tools registry` commands are wrapped around crane and previously only honored the local `--insecure` flag. The global root flag `--insecure-skip-tls-verify` was not being applied on that code path.

This change makes registry tools honor either flag by:
- adding a small helper to unify insecure flag behavior
- using that helper in `tools registry` pre-run (covers `pull`, `push`, etc.)
- using that helper in `tools registry prune`
- adding unit tests for all flag combinations

Validation:
- `GOCACHE=/tmp/go-build-cache go test ./src/cmd -run TestUseInsecureRegistryTransport -count=1`

## Related Issue

Relates to zarf-dev/zarf#4671

## Checklist before merging

### DIFF
--- a/src/cmd/crane.go
+++ b/src/cmd/crane.go
@@ -35,6 +35,10 @@ type registryOptions struct {
 	platform string
 }
 
+func useInsecureRegistryTransport(commandFlagInsecure bool, globalFlagInsecureSkipTLSVerify bool) bool {
+	return commandFlagInsecure || globalFlagInsecureSkipTLSVerify
+}
+
 func newRegistryCommand() *cobra.Command {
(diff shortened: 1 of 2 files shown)


AustinAbro321:
We shouldn't do this because Crane's insecure flag affects both plain-http and tls verification. Someone could expect to only allow insecure tls verification, but not plain-http requests if using the flag `--insecure-skip-tls-verify`. FWIW we do hide the `--insecure-skip-tls-verify` flag from `zarf tools registry` commands.

PR #4685, opened 2026-03-04
feat(deploy)!: introduce connected mode

### LINKED ISSUE #4580: Add "connected" mode to package deploy
In order to create a Zarf package that is deployable in connected environments, users are encouraged to use "yolo" mode, which is a create time construct that blocks packages from being created with images or repositories, and allows packages to be deployable without having an initialized cluster.  

A big problem with "yolo" mode (aside from the name) is that it prevents users from having a single package that is deployable inside and outside of the air-gap. Creators have to maintain two separate packages.  The `--connected` flag would eliminate this need, it would also enable deleting "yolo" from the the v1beta1 schema (#3433).

This would also entail adding `zarf.dev/agent: ignore` to resources mutated by the agent so that connected packages can be deployed to an initialized cluster. Similar to "yolo" packages today, `--connected` will allow deployments even when the cluster is uninitialized. 

- **Given** I have a Zarf package <my-package> which contains images, repositories, deployments, and FluxGitRepositories. 
- **When** I run zarf package deploy <my-package> --connected
- **Then** The package is deployed to the cluster without images and repositories being mirror'd. The pods and flux git repositories are given the label `zarf.dev/agent: ignore`.

Introduce a new key `connected` into the next Zarf schema version (v1beta1 https://github.com/zarf-dev/proposals/pull/52). When `connected: true` the package would be built without images or repositories  and the name would become `zarf-package-my-package-connected.tar.zst` similar to if a flavor was used. However, often required components or actions will change between connected and dis-connected environments, which would require creators to use both use flavors and this construct. This may get confusing. I believe it'll be simpler to introduce better templating features for creators  

### Additional context
This issue was inspired by [this comment](https://github.com/zarf-dev/proposals/pull/52#discussion_r2593760271) on https://github.com/zarf-dev/proposals/pull/52

AustinAbro321:
## Description

This introduces a connected mode, allowing users to deploy to packages created with images and git repos, without actually pushing those images or git repos or mutating resources to point to Zarf infrastructure. This should let users have a single zarf.yaml that works in both connected and disconnected environments. 

This doesn't introduce "connected" packages would wouldn't include images and git repos, however I think this can be a future enhancement. 

This replaces the `zarf dev deploy --no-yolo` flag with `--connected=false` (connected defaults to true in dev deploy mode). I bit awkward for a boolean flag, but I prefer the consistency

The breaking change here is that yolo mode and dev mode now by default label resources to not be mutated by the agent. I can't think of a strong case for this use case, as the only behavior of yolo mode previously was that it could deploy resources without Zarf existing, and yolo mode packages cannot have images or repos. 

Something to note that's stayed consistent with yolo mode is that connected deploys set values for the Zarf state struct, and pass state through the application, but they don't persist any state to the cluster. 

## Related Issue

Fixes #4580 

## Checklist before merging

### DIFF
--- a/src/internal/packager/helm/post-render.go
+++ b/src/internal/packager/helm/post-render.go
@@ -37,7 +37,6 @@ type renderer struct {
 	adoptExistingResources bool
 	cluster                *cluster.Cluster
 	connectedDeploy        bool
-	skipSecretUpdates      bool
 	state                  *state.State
 	actionConfig           *action.Configuration
 	variableConfig         *variables.VariableConfig
@@ -59,13 +58,11 @@ func newRenderer(ctx context.Context, chart v1alpha1.ZarfChart, adoptExistingRes
 		return nil, fmt.Errorf("package name required to run post renderer")
 	}
 	// Update secrets when not in connected mode, as connected packages in hybrid / air-gap clusters could rely on pulling from the registry with ###ZARF_REGISTRY###
-	skipSecretUpdates := s.GetClusterConnectivity() == state.ClusterConnectivityConnected
 	rend := &renderer{
 		chart:                  chart,
 		adoptExistingResources: adoptExistingResources,
 		cluster:                c,
 		connectedDeploy:        connectedDeploy,
-		skipSecretUpdates:      skipSecretUpdates,
 		state:                  s,
 		actionConfig:           actionConfig,
 		variableConfig:         variableConfig,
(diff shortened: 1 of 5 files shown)


src/internal/packager/helm/post-render.go
```diff
@@ -328,6 +338,55 @@ func (r *renderer) addLabelsToNestedPath(obj *unstructured.Unstructured, path []
 	return unstructured.SetNestedStringMap(obj.Object, templateLabels, path...)
 }
 
+// agentMutatedKinds maps resources mutated by the Zarf agent webhook to the
+// label paths where the ignore label should be applied.
+// These come from the webhook configuration in packages/zarf-agent/chart/templates/webhook.yaml.
```
soltysh, line 337: Can you add a test-case ensuring that the wehbook.yaml and this list are in sync?

AustinAbro321: Added

src/pkg/state/state.go
```diff
@@ -108,6 +119,8 @@ type State struct {
 	ZarfAppliance bool `json:"zarfAppliance"`
 	// K8s distribution of the cluster Zarf was deployed to
 	Distro string `json:"distro"`
+	// The connectivity mode of the cluster
+	ClusterConnectivity ClusterConnectivity `json:"clusterConnectivity"`
```
brandtkeller: I think the semantics around the package connected vs airgap state make a lot of sense. I am struggling a little with a stored cluster connectivity state versus other optionality around checking if state exists. 

Is there a future extension that makes this something that will be utilized more broadly? Otherwise this is never "flipped" and is only ever a constant when initialized.

AustinAbro321: It is currently set in dev and connected mode to "connected", but as you note we never actually store a "connected" state to the cluster. I have no immediate use case in mind for "connected" Zarf state to be stored in cluster.

While we could change connected / yolo mode to pass around a nil state when state is not initialized I feel it's semantically easier to work with the cluster connectivity field

brandtkeller: I don't disagree with that stance. I mainly just want to think through what this field conveys and whether it remains sustainable under the state API. 

For the former - does someone look at state and understand how this setting propagates through the system.

For the context we've received around connected deployments I very much see these capabilities co-existing and that is where I wonder if this is misleading or not. 

I wouldn't block on this alone though.

AustinAbro321: Yeah, that's a good point, it would be confusing for onlookers. I added functions IsConfigured for registryinfo and gitserver structs, which gets us away from a distro == "YOLO" check which is what I was going for, and I think sets us up better for future refactors. 

LMK what you think, good callout

brandtkeller: I like that pattern - creation of secrets pertaining to a specific service should only happen when that service exists.

brandtkeller (approved):
lgtm - the issue outlines some interesting constraints we'll want to keep an eye on regarding "Connected Package creation/pull".

PR #4694, opened 2026-03-06
feat(init): clarify --registry-secret or --registry-url

### LINKED ISSUE #3736: Improve UX of `--registry-secret` flag
The `--registry-secret` flag is available on `zarf init` with the help text `Registry secret value`. It should be clear to the user that this flag is only used for the internal registry. Additionally, if a user tries to set both `--registry-secret` and `--registry-url` then `zarf init` should fail as the `--registry-secret` flag will not have any effect

bm54cloud:
## Description

- Clarifies that `--registry-secret` is used for internal registry only and is a valid flag only if `--registry-url` is not set
- Prevents users from setting both `--registry-url `and `--registry-secret`

## Related Issue

Fixes #3736 

Relates to #3736 

## How was this tested?
- Built zarf binary locally and confirmed `--registry-secret` command description is now updated
Screenshot 2026-03-11 at 2 47 08 PM
Screenshot 2026-03-11 at 2 49 09 PM
- Confirmed setting both `--registry-url` and `--registry-secret` does in fact give an error
Screenshot 2026-03-11 at 2 52 10 PM

## Checklist before merging


bm54cloud:
@AustinAbro321 I think it's ready. LMK if more testing is needed. Thanks!

PR #4699, opened 2026-03-09
fix!: only pull required layers during OCI deploy

### LINKED ISSUE #4698: zarf package deploy oci://<oci-source> pull optional components, even when not required
### Environment
App version: v0.73.2

### Steps to reproduce
1. Publish a package with optional components
2. Run zarf package deploy oci://<my-package>

### Expected result

Only required components are pulled

### Actual Result

Optional components are also pulled

AustinAbro321:
## Description

I noticed this issue while working on #4580. There are two issues here. 

The first is that in cmd/package we weren't filtering components on load during deploy. We shouldn't filter early for interactive deploys, however we should filter early when we are confirming.

The second issue is that is that assemble layers was pulling every layer when AllLayers was given even when requested components were only a subset of the components. Also the previous API didn't have a nice way to ask for multiple layer types. For instance for skeleton packages, and the soon to be introduced "connected" packages, we won't want to pull images, so we had a isSkeleton check that pulled everything except images. Now we can request all the layers besides images. 

## Related Issue

Fixes #4698

## Checklist before merging

### DIFF
--- a/src/pkg/packager/pull.go
+++ b/src/pkg/packager/pull.go
@@ -136,7 +136,7 @@ func pullOCI(ctx context.Context, opts pullOCIOptions) (*layout.PackageLayout, e
 		}
 	}
 
-	// Get all the layers for relevant components, optionally include images if it's a skeleton package
+	// Get all the layers for relevant components, exclude images if it's a skeleton package
 	layerTypes := opts.LayerTypes
 	if isSkeleton(desc.Platform) {
 		if len(layerTypes) == 0 {
--- a/src/pkg/zoci/pull.go
+++ b/src/pkg/zoci/pull.go
@@ -26,7 +26,7 @@ import (
 
 var (
 	// PackageAlwaysPull is a list of paths that will always be pulled from the remote repository.
-	PackageAlwaysPull = []string{layout.ZarfYAML, layout.Checksums, layout.Signature}
+	PackageAlwaysPull = []string{layout.ZarfYAML, layout.Checksums, layout.Signature, layout.Bundle}
 )
 
 // PullPackage pulls the package from the remote repository and saves it to the given path.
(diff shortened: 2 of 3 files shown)


src/pkg/packager/pull.go
```diff
@@ -136,9 +136,17 @@ func pullOCI(ctx context.Context, opts pullOCIOptions) (*layout.PackageLayout, e
 		}
 	}
 
-	// zarf creates layers around the contents of component primarily
-	// this assembles the layers for the components - whether filtered above or not
-	layersToPull, err := remote.AssembleLayers(ctx, pkg.Components, isSkeleton(desc.Platform), opts.LayersSelector)
+	// Get all the layers for relevant components, optionally include images if it's a skeleton package
```
brandtkeller: nit: not entirely wrong but the comment leads me to think that images ARE being included if it is a skeleton package.

src/pkg/zoci/pull.go
```diff
@@ -67,66 +68,66 @@ func (r *Remote) PullPackage(ctx context.Context, destinationDir string, concurr
...
-	layerMap[AllLayers] = root.Layers
-
-	// We always pull the metadata layers provided we can locate them
-	alwaysPull := make([]ocispec.Descriptor, 0)
+	if len(include) == 0 {
+		include = GetAllLayerTypes()
+	}
+	// Metadata layers are always included
+	layers := make([]ocispec.Descriptor, 0)
 	for _, path := range PackageAlwaysPull {
```
brandtkeller, line 85: Reminds me that we might need to update `PackageAlwaysPull` to include the bundle signature.

brandtkeller (changes requested):
Generally have no issues with the approach and appreciate the update. 

Noticed a minor gap outside of the scope of these changes that we could include and otherwise just need to resolve CI.

src/pkg/packager/pull.go
AustinAbro321: Ah yeah is confusing, updated

src/pkg/zoci/pull.go
AustinAbro321: Yeah, we should. Added it to this PR

PR #4704, opened 2026-03-11
feat(schema): update v1beta1 types

AustinAbro321:
## Description

This brings the v1beta1 schema in line with what is proposed in https://github.com/zarf-dev/proposals/pull/52

Relates to #3433 

## Checklist before merging


src/pkg/schema/generate.go
```diff
@@ -15,27 +15,40 @@ import (
...
+	}
+	if err := writeSchema("v1beta1", "zarf-v1beta1-schema.json", &v1beta1.Package{}); err != nil {
+		fmt.Println(err.Error())
 		os.Exit(1)
 	}
+	if err := writeSchema("v1beta1", "zarf-v1beta1-component-schema.json", &v1beta1.ComponentConfig{}); err != nil {
+		fmt.Println(err.Error())
+		os.Exit(1)
+	}
+}
```
soltysh, line 34: Any particular reason for generating two files for v1beta1? I think from a user perspective, it's easier to import a single file containing all the elements of the schema, rather than multiple.

src/api/v1beta1/component.go
```diff
@@ -0,0 +1,408 @@
...
+type Component struct {
+	// The name of the component.
+	Name string `json:"name" jsonschema:"pattern=^[a-z0-9][a-z0-9\\-]*$"`
+	// Message to include during package deploy describing the purpose of this component.
+	Description string `json:"description,omitempty"`
+	// Do not install this component unless explicitly requested. Defaults to false, meaning the component is required.
+	Optional bool `json:"optional,omitempty"`
+	ComponentSpec
+	// Data injections removed from the v1beta1 schema; kept as a v1alpha1 backwards-compatibility shim.
+	dataInjections []v1alpha1.ZarfDataInjection
```
soltysh: I think, ideally we should not include v1alpha1 here, but rather copy the struct in-place. Then when needed, you'll want to convert from one to the other.

src/api/v1beta1/component.go
```diff
@@ -0,0 +1,408 @@
...
+	// group removed from the v1beta1 schema; kept as a v1alpha1 backwards-compatibility shim.
+	group string
+}
+
+// GetDeprecatedDataInjections returns the v1alpha1 data injections carried as a backwards-compatibility shim.
+func (c Component) GetDeprecatedDataInjections() []v1alpha1.ZarfDataInjection {
```
soltysh: I know what you're trying to achieve, but I don't think this is the right way to do it. Currently, you're hiding the field as private, but you're providing public method for setting/getting it. This might mean, external consumers will be able to change it, and that's not what we want. This field should only be settable by our conversion logic, right?

soltysh: This comment applies to all private fields here and in other files below.

src/api/v1beta1/component.go
```diff
@@ -0,0 +1,408 @@
...
+	return c.group
+}
+
+// SetDeprecatedGroup sets the v1alpha1 group carried as a backwards-compatibility shim.
+func (c *Component) SetDeprecatedGroup(group string) {
+	c.group = group
+}
+
+// GetImages returns all image names specified in the component, including those from ImageArchives.
+func (c Component) GetImages() []string {
```
soltysh, line 37: I wonder about these struct methods vs just regular helpers, I don't have any strong preferences, but I like having clean API types only 😅

src/api/v1beta1/component_config.go
```diff
@@ -0,0 +1,66 @@
+// SPDX-License-Identifier: Apache-2.0
```
soltysh, line 1: Nit: regarding file naming, majority of our files use lower-casing for file names. there are some rare cases where we have a CamelCase naming, but I think sticking with lower-case for file names is the best.

src/api/v1beta1/component.go
soltysh: The idea is that a consumer of v1beta1, should not be forced into importing deprecated API.

src/api/v1beta1/package.go
```diff
@@ -0,0 +1,171 @@
...
+	// Zarf-generated package build data.
+	Build BuildData `json:"build,omitempty"`
+	// List of components to deploy in this package.
+	Components []Component `json:"components" jsonschema:"minItems=1"`
+	// Values imports Zarf values files for templating and overriding Helm values.
+	Values Values `json:"values,omitempty"`
+	// Documentation files included in the package.
+	Documentation map[string]string `json:"documentation,omitempty"`
+	// Variables removed from the v1beta1 schema; kept as a v1alpha1 backwards-compatibility shim.
+	variables []v1alpha1.InteractiveVariable
```
soltysh: Same comment about not relying on v1alpha1, especially that those structs are not that big to copy them over.

soltysh: Then, during conversion testing we'll be able to uncover drifts and ensure those are properly moved in both directions.

soltysh (commented):
Some high-level comments, but I'll do another pass regarding the API as a whole.

src/api/v1beta1/component.go
```diff
@@ -0,0 +1,408 @@
...
+// ServerSideApplyMode controls when server-side apply is used during deploy.
+type ServerSideApplyMode string
+
+const (
+	// ServerSideApplyEnabled always uses server-side apply.
+	ServerSideApplyEnabled ServerSideApplyMode = "true"
+	// ServerSideApplyDisabled always uses client-side apply.
+	ServerSideApplyDisabled ServerSideApplyMode = "false"
+	// ServerSideApplyAuto uses server-side apply for fresh installs and matches the prior strategy on upgrade.
+	ServerSideApplyAuto ServerSideApplyMode = "auto"
```
soltysh, line 108: Suggestion, the above values, especially `true` and `false` might cause problems during yaml conversions. Should we consider turning all constants into upper cases values? Here, I'd specifically use `Enabled`, `Disabled` and `Auto` (alternatively `Always`, `Never` and `Auto`).

soltysh: Similarly, above we'd use `Registry`, `SeedRegistry`, `Injector`, etc.

src/api/v1beta1/component.go
```diff
@@ -0,0 +1,408 @@
...
+	// SetValueString sets the raw value.
+	SetValueString SetValueType = "string"
+)
+
+// SetValue declares a value that can be set during a package deploy.
+type SetValue struct {
+	// Key represents which value to assign to.
+	Key string `json:"key,omitempty"`
+	// Value is the current value at the key.
+	Value any `json:"value,omitempty"`
```
soltysh: The use of `any` here might be problematic in the conversion layer (yaml to go and vice-versa). Why not just use `string` always here, since this will match one of the above types (`yaml`, `json` and `string`). Alternatively, introduce `StringValue`, JSONValue` and `YAMLValue` instead as pointers?

src/api/v1beta1/component.go
```diff
@@ -0,0 +1,408 @@
...
+}
+
+// ComponentActionDefaults sets the default configs for child actions.
+type ComponentActionDefaults struct {
+	// Hide the output of commands during execution (default false).
+	Silent bool `json:"silent,omitempty"`
+	// Default timeout in seconds for commands (default 0, no timeout).
+	MaxTotalSeconds int32 `json:"maxTotalSeconds,omitempty"`
+	// Retry commands a given number of times if they fail (default 0).
+	Retries int32 `json:"retries,omitempty"`
```
soltysh, line 307: I think, for both of these fields, I'm inclined to drop the `omitempty`, such that those will always show up in the yaml after zarf will process these. This will ensure users have those values explicit, especially when we default them to 0.

src/api/v1beta1/package.go
```diff
@@ -0,0 +1,171 @@
...
+	// The machine name that created this package.
+	Hostname string `json:"hostname,omitempty"`
+	// The username who created this package.
+	User string `json:"user,omitempty"`
+	// The architecture this package was created on.
+	Architecture string `json:"architecture"`
+	// The timestamp when this package was created.
+	Timestamp string `json:"timestamp"`
+	// The version of Zarf used to build this package.
+	Version string `json:"version"`
```
soltysh, line 114: Since `BuildData` field is `omitempty` inside `Package`, so there are some edge-cases when these values might show up as `""`, when at least one of the other fields is set.

src/pkg/schema/generate.go
AustinAbro321: There are two files for the two different kinds. One is for the component config kind, while the other is for the package kind. To your point, perhaps we should call one of the package schema and one the component schema

src/api/v1beta1/component.go
AustinAbro321: Removed the setter, I believe we'll be able to use a different strategy in #4823 that we discussed

src/api/v1beta1/component_config.go
AustinAbro321: sure updated for snake case

src/api/v1beta1/component.go
AustinAbro321: Since they will be in quotes it should be fine, we haven't run into any issues yet, and this is an existing field in the v1alpha1 schema. I want to keep it as "true", "false", "auto" since this is essentially a pass through of the options on the "helm install --server-side` flag.

src/api/v1beta1/component.go
AustinAbro321: I would disagree there, since removing `omitempty` will make it so the fields are required by our jsonschema when `.actions.defaults` is defined

src/api/v1beta1/package.go
AustinAbro321: I believe this is okay. Zarf populates these fields, and we treat `omitEmpty` as a way to say these fields are "required" and Zarf will always populate them out when creating a package.

src/api/v1beta1/component.go
AustinAbro321: Copied that over from v1alpha1, I hadn't realized that this field is unused. I believe it was mistakenly added and we can delete. As far as I can tell there are no other fields like that.

src/api/v1beta1/component.go
AustinAbro321: I like the ergonomics of pulling data directly from the type, but I see where you're coming from

src/api/v1beta1/component.go
AustinAbro321: https://github.com/zarf-dev/zarf/pull/4823 is in draft, but I'm happy with the strategy there. It exposes a function `SetDeprecatedFromGeneric(g types.Package, pkg Package) Package` but it's only callable from Zarf since types.Package is internal

src/pkg/schema/generate.go
soltysh: I mean, we should have a single schema for all kinds. This way, we'll make it similar to `zarf-v1alpha1-schema.json` which has all of them. From a dev pov, you want to import a single schema for your `zarf.yaml`, rather many, depending on which types you're using.

src/api/v1alpha1/package.go
```diff
@@ -273,6 +273,18 @@ type ZarfBuildData struct {
 	// These are files added after checksum generation (e.g., signature files).
 	// This list is authenticated through the signed zarf.yaml.
 	ProvenanceFiles []string `json:"provenanceFiles,omitempty"`
+	// originalAPIVersion records the apiVersion the package was read from before any conversion.
+	originalAPIVersion string
+}
+
+// OriginalAPIVersion returns the apiVersion the package was read from before any conversion.
+func (b ZarfBuildData) OriginalAPIVersion() string {
```
soltysh, line 281: Private field vs public method applies here as well.

src/api/v1beta1/deprecated.go
```diff
@@ -0,0 +1,89 @@
...
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+package v1beta1
+
+// These types are duplicated from v1alpha1 and carried as backwards-compatibility
+// shims so v1alpha1 packages can be converted to v1beta1 without v1beta1 importing
+// v1alpha1. They are not part of the v1beta1 schema.
+
+// VariableType represents a type of a Zarf package variable.
+type VariableType string
```
soltysh, line 11: All of those types should be private and accessible only within this package for backwards compatibility.

soltysh (commented):
Two remaining things: 
1. Private fields in `deprecated.go`
2. Single schema for v1beta1.

src/api/v1beta1/deprecated.go
AustinAbro321: We'll need these types to getttable when using them in packager for things like deploy, however I did add an explicit deprecated mark so linters will catch them.

src/pkg/schema/generate.go
AustinAbro321: Let's take care of the schema combining in #4949

src/api/v1alpha1/package.go
AustinAbro321: Not sure if there is a change requested for this field

soltysh: What I mean by that is do we need external users of zarf to be able to set that field? To my understanding this field is internal and only used by zarf. So nobody except for zarf should be setting this field. So I'm proposing to remove both the getter and setter. The same comment applies to v1beta1, which has that same field.

src/api/v1beta1/component.go
```diff
@@ -0,0 +1,409 @@
...
+	Name string `json:"name" jsonschema:"pattern=^[a-z0-9][a-z0-9\\-]*$"`
+	// Message to include during package deploy describing the purpose of this component.
+	Description string `json:"description,omitempty"`
+	// Do not install this component unless explicitly requested. Defaults to false, meaning the component is required.
+	Optional      bool `json:"optional,omitempty"`
+	ComponentSpec `json:",inline"`
+
+	// Data injections removed from the v1beta1 schema; kept as a v1alpha1 backwards-compatibility shim.
+	dataInjections []ZarfDataInjection
+	// group removed from the v1beta1 schema; kept as a v1alpha1 backwards-compatibility shim.
```
soltysh, line 18: For ease of finding these fields in this file, I'm inclined to propose some kind of big comment and move all the depreateced fields under that. Something like:
```golang
type Component struct {
...
	ComponentSpec `json:",inline"`

	// *** REMOVED FIELDS ***
    // Below you'll find a set of fields which are kept here for v1alpha1 backwards-compatiblity.
	dataInjections []ZarfDataInjection
	group string
}
```
wdyt?

soltysh: Same comment applies to all sub-structs and package struct as well.

src/api/v1beta1/component.go
```diff
@@ -0,0 +1,409 @@
...
+	group string
+}
+
+// GetDeprecatedDataInjections returns the v1alpha1 data injections carried as a backwards-compatibility shim.
+//
+// Deprecated: only used to convert v1alpha1 packages; will be removed once v1alpha1 support is dropped.
+func (c Component) GetDeprecatedDataInjections() []ZarfDataInjection {
```
soltysh, line 25: Same comment here regarding exposing the fields. Can't you just drop them in https://github.com/zarf-dev/zarf/pull/4823? Where you're introducing a mechanism, to handle the conversions, which I remember was the only reason for those setters and getters.

src/api/v1beta1/component.go
```diff
@@ -0,0 +1,409 @@
...
+	// SetValueString sets the raw value.
+	SetValueString SetValueType = "string"
+)
+
+// SetValue declares a value that can be set during a package deploy.
+type SetValue struct {
+	// Key represents which value to assign to.
+	Key string `json:"key,omitempty"`
+	// Type declares the kind of data being stored in the value.
+	Type SetValueType `json:"type,omitempty"`
```
soltysh, line 368: Question, do we default it if not specified? If so it'd be good to document that defaulting here.

soltysh: It'd be good to go through all fields and add the defaulting information where appropriate. I see we have that in some places, but not all.

src/api/v1beta1/deprecated.go
soltysh: Fair, although that deprecated mark isn't safe enough from my experience 😓 how about creating an internal package that will stop users from accessing that?

soltysh (approved):
From previous review:

> Private fields in deprecated.go

Can be handled in followup. Also there's another topic regarding setters and getters which I hope we'll be able to handle with conversions in https://github.com/zarf-dev/zarf/pull/4823

> Single schema for v1beta1.

Being worked on in https://github.com/zarf-dev/zarf/pull/4949

With all that, this is good to go, the remaining comments are non-blocking.

src/api/v1beta1/component.go
AustinAbro321: The fields will still be used by the rest of Zarf though, for instance in `packager.Deploy()` we'll call `pkgLayout.Pkg.Component[x].GetDeprecatedDataInjections()` when going through the data injection logic

src/api/v1beta1/component.go
AustinAbro321: I like that

src/api/v1beta1/component.go
AustinAbro321: Yeah this is a good idea

src/api/v1alpha1/package.go
AustinAbro321: We don't need them to set that field, but we will want it available to other packages within Zarf so that when we parse the zarf.yaml we can immediately set it.

brandtkeller (approved):
lgtm - matches expectations and sets us up to begin iteration

Issue #4681, opened 2026-03-04
expand shell completion for vendor-ed tools


a1994sc:
## Driving Questions
What are we hoping to do to help improve the end user experience?

- Shell completion for the sub-commands:
  - [ ] `kubectl`
  - [ ] `helm`

brandtkeller:
No disagreement here. Issues like this in the future might be best separated - as for instance we'd want to potentially block the helm improvements here on the pending helm 4 PR (but not block kubectl).

AustinAbro321:
Feel free to open if I'm missing something, but I believe all tools have shell completion now with #4676

PR #4705, opened 2026-03-12
feat: enable plugin support for vender-ed kubectl

a1994sc:
## Description

This update enables `zarf tools kubectl` to use `kubectl` plugins, tested with the following:

[`kubectl-cnpg`](https://github.com/cloudnative-pg/cloudnative-pg/tree/main/cmd/kubectl-cnpg)
[`kubectl-get_all`](https://github.com/stackitcloud/kubectl-get-all)

## Related Issue

Relates to #4681

## Checklist before merging


src/cmd/kubectl.go
```diff
@@ -6,29 +6,78 @@ package cmd
...
-			Out:    os.Stdout,
-			ErrOut: os.Stderr,
-		},
-	})
-
-	kubectlCmd.Use = "kubectl"
-	kubectlCmd.Aliases = []string{"k"}
+	var kubectlCmd *cobra.Command
+	// because of some weirdness around how the kubectl command works we only want to pass it the args when it is being called via `zarf tools kubectl` and not, as an example, `zarf tools sbom`
+	if (len(os.Args) > 1 && (os.Args[2] == kubeLong || os.Args[2] == kubeShort)) || len(os.Args) > 2 && (os.Args[2] == cobra.ShellCompRequestCmd) && (os.Args[3] == kubeLong || os.Args[3] == kubeShort) {
```
AustinAbro321: I will probably need to come back to this another time and really dig into it, but I found this as a potentially alternative solution which feels a bit cleaner. It also allows other cli consumers to work such as uds-cli since it doesn't rely on hardcoded args
```
	kubectlCmd.Run = nil
	kubectlCmd.RunE = func(cmd *cobra.Command, args []string) error {
		if len(args) > 0 {
			pluginHandler := kubecmd.NewDefaultPluginHandler(plugin.ValidPluginFilenamePrefixes)
			if err := kubecmd.HandlePluginCommand(pluginHandler, args, 0); err != nil {
				return err
			}
			// HandlePluginCommand replaces the process via syscall.Exec when a
			// plugin is found. If we reach here, no plugin matched.
		}
		return cmd.Help()
	}
```

a1994sc: I will take a look, I will say that I'm not a fan of that if statement. I did notice that if I don't have that check a command like:

```
zarf tools sbom --parallelism 10 registry:ghcr.io/zarf-dev/zarf/agent:v0.71.0
```

would fail... which is annoying

a1994sc: Also this section may need to get reworked to support zarf being vendor-ed:

https://github.com/zarf-dev/zarf/blob/2d16d15e483f91439f7fc3f2e806e2e628cb8434/src/cmd/vendor.go#L66-L72

a1994sc: @AustinAbro321 can you let me know if this is cleaner?

https://github.com/zarf-dev/zarf/pull/4705/changes#diff-b2f2d3931dfb07d633f20207848a76126bad08dec09f577f92a78454ac9fa919R28-R33

src/cmd/kubectl.go
```diff
@@ -6,29 +6,80 @@ package cmd
...
+
+	if argOffset > -1 {
+		kubectlCmd = kubecmd.NewDefaultKubectlCommandWithArgs(kubecmd.KubectlOptions{
+			IOStreams: genericiooptions.IOStreams{
+				In:     os.Stdin,
+				Out:    os.Stdout,
+				ErrOut: os.Stderr,
+			},
+			Arguments: os.Args[argOffset:],
+			PluginHandler: &kubecmd.DefaultPluginHandler{
```
soltysh: Let's use `kubecmd.NewDefaultPluginHandler(plugin.ValidPluginFilenamePrefixes)` this will make this code less prone to changes happening in kubectl.

src/cmd/kubectl.go
```diff
@@ -6,29 +6,80 @@ package cmd
...
+// code credit, k0sproject/k0s
+// https://github.com/k0sproject/k0s/blob/df88db5f317bb84dcda797ff6a68956bc2e49683/cmd/kubectl/kubectl.go#L202
+func patchPluginListSubcommand(kubectlCmd *cobra.Command) {
+	cmd, _, err := kubectlCmd.Find([]string{"plugin", "list"})
+	cmdutil.CheckErr(err)
+
+	originalRun := cmd.Run
+	cmd.Run = func(_ *cobra.Command, args []string) {
+		root := kubecmd.NewKubectlCommand(kubecmd.KubectlOptions{})
+		root.Use = strings.ReplaceAll(kubectlCmd.CommandPath(), " ", "-")
```
soltysh, line 78: As one of the maintainers of kubectl I will openly admit this is not the best path forward 😭 . Honestly, I'd be very keen to see a PR opened against kubectl allowing for this to be tweaked in more reasonable way.

src/cmd/kubectl.go
```diff
@@ -6,29 +6,80 @@ package cmd
...
+		})
+	} else {
+		kubectlCmd = kubecmd.NewKubectlCommand(kubecmd.KubectlOptions{
+			IOStreams: genericiooptions.IOStreams{
+				In:     os.Stdin,
+				Out:    os.Stdout,
+				ErrOut: os.Stderr,
+			},
+		})
+	}
```
soltysh, line 50: This entire block will be easier to read if we write it like so:
```golang
opts := kubecmd.KubectlOptions{
	IOStreams: genericiooptions.IOStreams{
		In:     os.Stdin,
		Out:    os.Stdout,
		ErrOut: os.Stderr,
	},
}
if argOffset > -1 {
	opts.Arguments = os.Args[argOffset:]
	opts.PluginHandler = kubecmd.NewDefaultPluginHandler(plugin.ValidPluginFilenamePrefixes)
}

kubectlCmd = kubecmd.NewDefaultKubectlCommandWithArgs(opts)
kubectlCmd.Use = kubeLong
kubectlCmd.Aliases = []string{kubeShort}
kubectlCmd.SilenceErrors = true
```

src/cmd/kubectl.go
```diff
@@ -6,29 +6,80 @@ package cmd
...
-	})
-
-	kubectlCmd.Use = "kubectl"
-	kubectlCmd.Aliases = []string{"k"}
+	var kubectlCmd *cobra.Command
+
+	argOffset := slices.Index(os.Args, kubeLong)
+	if argOffset < 0 {
+		argOffset = slices.Index(os.Args, kubeShort)
+	}
```
soltysh, line 45: Matching a string inside the slice is error-prone. As long as we're matching at least `kubectlLong` it's not that bad, but just `k` might get picked up faster. This is not ideal either, but let's hard code at least the requirement for `tools kubectl` ordering:
```golang
argOffset := -1
for i, arg := range os.Args {
	if arg == "tools" || arg == "t" {
		if i+1 < len(os.Args) && (os.Args[i+1] == kubeLong || os.Args[i+1] == kubeShort) {
			argOffset = i + 1
			break
		}
	}
}
```

src/cmd/kubectl.go
a1994sc: In a similar vein, I was a bit frustrated at `SetupPluginCompletion` because it seemed to me that it would only work when `kubectl` is the root command.

https://github.com/kubernetes/kubectl/blob/82a3c054228a186e09d62e606a4028b10da87250/pkg/cmd/plugin/plugin_completion.go#L43-L46

soltysh: It was written with kubectl in mind for sure, but again I'm happy to review PRs :smile:

src/cmd/kubectl.go
a1994sc: Tested locally and this seems to work just as well, I will push the changes up this afternoon

src/cmd/kubectl.go
a1994sc: Tested locally and this seems to work just as well, I will push the changes up this afternoon

src/cmd/kubectl.go
a1994sc: Tested locally and this seems to work just as well, I will push the changes up this afternoon

src/cmd/kubectl.go
a1994sc: Ha, I may take you up on that offer.

a1994sc: I do think that I would like to at least get this PR and then we can update things once I have some more bandwidth to get the PR into kubectl

a1994sc:
Hey @soltysh, I was wondering if you would be happy with reviewing this PR as is? I would like like submit a PR for the kubectl updates at a later point when I have a bit more bandwidth.

soltysh:
> Hey @soltysh, I was wondering if you would be happy with reviewing this PR as is? I would like like submit a PR for the kubectl updates at a later point when I have a bit more bandwidth.

Yeah, let's land this as is and deal with kubectl separately.

PR #4707, opened 2026-03-12
fix: set field manager once during pre-run to avoid data race

AustinAbro321:
## Description

We are setting the managed fields manager within a function that's called multiple times. When tests run in parallel they run into a race condition of both setting it. See https://github.com/zarf-dev/zarf/actions/runs/23006049045/job/66803088360?pr=4704

## Checklist before merging

### DIFF
--- a/main.go
+++ b/main.go
@@ -12,13 +12,9 @@ import (
 
 	"github.com/zarf-dev/zarf/src/cmd"
 	"github.com/zarf-dev/zarf/src/config"
-	"github.com/zarf-dev/zarf/src/pkg/cluster"
-	"helm.sh/helm/v4/pkg/kube"
 )
 
 func main() {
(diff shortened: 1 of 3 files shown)


src/internal/packager/helm/common.go
```diff
@@ -32,6 +32,10 @@ import (
 
 var contentCachePath = filepath.Join("helm", "content")
 
+func init() {
+	kube.ManagedFieldsManager = cluster.FieldManagerName
+}
```
soltysh: I don't think this is the best path forward. Using inits have weird consequences long-term for our consumers. How about doing some sort of a singleton to do that?

AustinAbro321: Fair point, updated to set this in main. It won't be set automatically for SDK users anymore, but I think that's fine as they'll get the choice between setting it to their app specific name or to Zarf.

PR #4720, opened 2026-03-17
feat: use legacy Helm wait + reconciliation Healthchecks

### LINKED ISSUE #4716: podinfo-flux deployment flake in CI
We've been seeing a flake in deploys of Flux from Helm where Flux doesn't finish the deploy. Seemingly it's failing on the wait logic. Seems to be mostly a problem in CI, perhaps it's missing events? I'd like to add debug logging to confirm this. 

Here is an example run, https://github.com/zarf-dev/zarf/actions/runs/23148577968/job/67243522831

One option is to go back to the old waiting system and continue with healthchecks. 

Future options could be to either change Helm so the resync period is 5 minutes (this seems like a more reasonable default anyway), or wait until Helm introduces a poller instead of a watcher https://github.com/helm/helm/issues/31777. 

I am making an assumption that missed is the actual problem versus something else. I'd like to confirm this with additional debug logging on that job

AustinAbro321:
## Description

In #4350 we upgraded to Helm 4 and used the newest Helm wait logic, which mirror'd the logic we use in Zarf. However we keep running into flakes in CI as seen by #4716. I believe it's possible that the system is getting overloaded with events and flaking because of that. Whereas in this system we only run healthchecks after most items are already ready, limiting events to process. In any case, until we have a clear idea of the solution, this will revert us to the same waiting behavior we had in Helm 3

## Related Issue

Fixes #4716

## Checklist before merging

### DIFF
--- a/src/internal/packager/helm/chart.go
+++ b/src/internal/packager/helm/chart.go
@@ -5,6 +5,7 @@
 package helm
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


brandtkeller (approved):
lgtm - straight-forward and should allow for additional scrutiny under the associated flake issue.

PR #4723, opened 2026-03-18
fix: change scale down policy back to csa

AustinAbro321:
## Description

In #4437 we changed the scale down policy to ssa. Originally, the only option was to set the ssa field manager to Zarf, however since sdk users now have the ability to set the ssa manager to whatever they want, modifying the scale down policy using ssa will end in errors, as we either won't find the items on the resource, or we'll update it to a different field manager and the tool used to deploy the init package through the Zarf SDK would fail on their next deploy unless force conflicts was set.  

## Checklist before merging


AustinAbro321:
This solution actually doesn't work because ssa will still conflict with csa, it's just that csa won't have conflicts. Fixed in #4725 instead

PR #4725, opened 2026-03-18
feat: stop managing scale down policy in CLI

AustinAbro321:
## Description

We were setting the scale down policy in the cli and in the chart, which goes against ssa, and would cause failures if the field manager was not Zarf. This complexity is unnecessary, as we already have retry logic on image push that will trigger if the pod that is being pushed to is deleted through a HPA scale down

## Checklist before merging


brandtkeller (approved):
Lgtm - I think we need to call out the implications. I don't imagine many people deploy older versions of the init package with newer versions of the CLI - but it could have a minor impact here.

Issue #4739, opened 2026-03-21
Feature: Add password flag for zarf tools gen-key


natrufthr:
### Is your feature request related to a problem? Please describe.

I need to be able to pass a password to zarf tools gen-key to generate keys without being prompted.

### Describe the behavior you'd like

- **Given** a state
   Given user runs zarf tools gen-key and provide a password flag with a password
- **When** an action is taken
-  When Zarf is installed
- **Then** something happens
-  command succeeds successfully

### Describe alternatives you've considered
(optional) A clear and concise description of any alternative solutions or features you've considered.
Running cosign and not running zarf.

### Additional context

I would prefer to stay as zarf specific as possible and use as little other tools I can. This would allow me to do that in line when scripting things out. Thank you!

brandtkeller:
I think this is a valid request. In order to keep it finely-scoped I would recommend replicating the password flags from `zarf tools registry login`. 

We could follow the cosign pattern as well (reading from the env variable) - but it lacks some consistency we might expect elsewhere.

nicholas-rees:
This looks like an issue that might be good for me to give a go on. I looked at some of the other issues that are documentation, but this looks like it has a little bit of anything. I would like to give this a go.

I'm following the [contributor guide](https://docs.zarf.dev/contribute/contributor-guide/) but I don't know if this issue is good with any of the [release milestones](https://github.com/zarf-dev/zarf/milestones) or even if that part of the contributor guide is even relevant. It looks like the release milestones might not be kept up to date.

Let me know if I'm mistaken on anything so far (i.e. I've picked an incorrect issue or something else). I'll try to get a PR as soon as I can.

nicholas-rees:
Thanks for assigning me.

I'm working on this, but obviously I have to get familar with the code. Seems pretty straightforward to setup my dev environment, and it seems to be build ok. I'll post in slack if I have any questions.

EDIT:
It does seem like running `pre-commit run -a` does result in a couple of linting issues, but I don't think that will be an issue since it seems like the ci only tests file that I touch. Again, I'll post in slack if I need advisement.

PR #4747, opened 2026-03-25
feat: add password flag for zarf tools gen-key “Relates to #4739”

nicholas-rees:
## Description

Adds password command line flags, and an interactive flag to allow easier scripting for `zarf tools gen-key`.

As suggested in #4739, I tried to mimic the behavior as close as possible to `zarf registry login`. However, by default, the registry login command runs interactively. So as to not change the current behavior and api, the tools gen-key command runs as interactive by default unless you pass `--interactive=false`. This behavior can be modified if it is desired to change the behavior. Just let me know.

## Related Issue

Relates to #4739

## Checklist before merging

P.S. Let me know if you want me to fill out this checklist. I ran the pre-commit hooks, but obviously the ci in the actions is authoritative, so I don't want to check things before they're done. I also read and tried best to follow the Contributor guide. Let me know if anything is unsatisfactory, and I will do what I can to fix it.

### DIFF
--- a/site/src/content/docs/commands/zarf_tools_gen-key.md
+++ b/site/src/content/docs/commands/zarf_tools_gen-key.md
@@ -17,10 +17,11 @@ zarf tools gen-key [flags]
 ### Options
 
 ```
-  -h, --help              help for gen-key
-      --interactive       Interactively prompt for registry server, username, and password if not provided
-      --password string   Password
-  -p, --password-stdin    Take the password from stdin
+      --force                       Force overwrite of keyfiles -- undefined behavior when running interactive
+  -h, --help                        help for gen-key
+      --interactive                 Interactively prompt for password
+      --password string             Private key password -- undefined behavior when running interactive
+      --password-stdin --password   Take the password from stdin -- undefined behavior with --password
 ```
 
 ### Options inherited from parent commands
--- a/site/src/content/docs/commands/zarf_tools_gen-key.md
+++ b/site/src/content/docs/commands/zarf_tools_gen-key.md
@@ -17,11 +17,11 @@ zarf tools gen-key [flags]
 ### Options
 
 ```
-      --force             Force overwrite of keyfiles -- cannot be used with --interactive
+      --force             Force overwrite of keyfiles
   -h, --help              help for gen-key
       --interactive       Interactively prompt for password
-      --password string   Private key password -- cannot be used with --password-stdin or --interactive
-      --password-stdin    Take the password from stdin -- cannot be used with --password or --interactive
+      --password string   Private key password
+      --password-stdin    Take the password from stdin
 ```
 
 ### Options inherited from parent commands
--- a/site/src/content/docs/commands/zarf_tools_gen-key.md
+++ b/site/src/content/docs/commands/zarf_tools_gen-key.md
@@ -17,11 +17,11 @@ zarf tools gen-key [flags]
 ### Options
 
 ```
-      --force                       Force overwrite of keyfiles -- undefined behavior when running interactive
-  -h, --help                        help for gen-key
-      --interactive                 Interactively prompt for password
-      --password string             Private key password -- undefined behavior when running interactive
-      --password-stdin --password   Take the password from stdin -- undefined behavior with --password
+      --force             Force overwrite of keyfiles -- cannot be used with --interactive
+  -h, --help              help for gen-key
+      --interactive       Interactively prompt for password
+      --password string   Private key password -- cannot be used with --password-stdin or --interactive
+      --password-stdin    Take the password from stdin -- cannot be used with --password or --interactive
 ```
 
 ### Options inherited from parent commands
--- a/site/src/content/docs/commands/zarf_tools_gen-key.md
+++ b/site/src/content/docs/commands/zarf_tools_gen-key.md
@@ -17,7 +17,10 @@ zarf tools gen-key [flags]
 ### Options
 
 ```
-  -h, --help   help for gen-key
+  -h, --help              help for gen-key
+      --interactive       Interactively prompt for registry server, username, and password if not provided
+      --password string   Password
+  -p, --password-stdin    Take the password from stdin
 ```
 
 ### Options inherited from parent commands
--- a/src/cmd/zarf_tools.go
+++ b/src/cmd/zarf_tools.go
@@ -592,6 +592,7 @@ type genKeyOptions struct {
 	force         bool
 	password      string
 	passwordStdin bool
+	reader        io.Reader
 }
 
 func newGenKeyCommand() *cobra.Command {
@@ -616,6 +617,10 @@ func newGenKeyCommand() *cobra.Command {
 	cmd.Flags().BoolVar(&o.force, "force", false, lang.CmdToolsGenKeyShortFlagForce)
 	cmd.Flags().StringVar(&o.password, "password", "", lang.CmdToolsGenKeyShortFlagPassword)
 	cmd.Flags().BoolVar(&o.passwordStdin, "password-stdin", false, lang.CmdToolsGenKeyShortFlagPasswordStdin)
+
+	if o.passwordStdin {
+		o.reader = os.Stdin
+	}
 
 	return cmd
 }
@@ -680,7 +685,10 @@ func (o *genKeyOptions) genKey(prvKeyFileName string, pubKeyFileName string) err
 			return []byte(password), nil
 		}
 	} else if o.passwordStdin {
-		password, err := io.ReadAll(os.Stdin)
+		if o.reader == nil {
+			return fmt.Errorf("reader not set for genKeyOptions")
+		}
+		password, err := io.ReadAll(o.reader)
 		if err != nil {
 			return err
 		}
--- a/src/cmd/zarf_tools.go
+++ b/src/cmd/zarf_tools.go
@@ -684,6 +684,9 @@ func (o *genKeyOptions) genKey(prvKeyFileName string, pubKeyFileName string) err
 		if err != nil {
 			return err
 		}
+		if string(password) == "" {
+			return fmt.Errorf("empty password disallowed when using --password-stdin")
+		}
 		passwordFunc = func(bool) ([]byte, error) {
 			return []byte(password), nil
 		}
--- a/src/cmd/zarf_tools.go
+++ b/src/cmd/zarf_tools.go
@@ -609,6 +609,9 @@ func newGenKeyCommand() *cobra.Command {
 					return errors.New(lang.CmdToolsGenKeyErrNoPasswordProvided)
 				}
 			}
+			if o.passwordStdin {
+				o.reader = os.Stdin
+			}
 			return nil
 		},
 	}
@@ -620,10 +623,6 @@ func newGenKeyCommand() *cobra.Command {
 	cmd.MarkFlagsMutuallyExclusive("password", "password-stdin", "interactive")
 	cmd.MarkFlagsMutuallyExclusive("interactive", "force")
 
-	if o.passwordStdin {
-		o.reader = os.Stdin
-	}
-
 	return cmd
 }
 
@@ -634,7 +633,6 @@ func (o *genKeyOptions) run(cmd *cobra.Command, _ []string) error {
 	err := o.genKey(prvKeyFileName, pubKeyFileName)
 
 	if err != nil {
-		logger.From(cmd.Context()).Error("Failed to generate key pair")
 		return err
 	}
 
--- a/src/cmd/zarf_tools.go
+++ b/src/cmd/zarf_tools.go
@@ -589,6 +589,7 @@ func (o *genPKIOptions) run(cmd *cobra.Command, args []string) error {
 
 type genKeyOptions struct {
 	interactive   bool
+	force         bool
 	password      string
 	passwordStdin bool
 }
@@ -603,7 +604,8 @@ func newGenKeyCommand() *cobra.Command {
 		RunE:    o.run,
 	}
 
-	cmd.Flags().BoolVar(&o.interactive, "interactive", true, lang.CmdToolsGenKeyShortFlagInteractive)
+	cmd.Flags().BoolVar(&o.interactive, "interactive", false, lang.CmdToolsGenKeyShortFlagInteractive)
+	cmd.Flags().BoolVar(&o.force, "force", false, lang.CmdToolsGenKeyShortFlagForce)
 	cmd.Flags().StringVar(&o.password, "password", "", lang.CmdToolsGenKeyShortFlagPassword)
 	cmd.Flags().BoolVar(&o.passwordStdin, "password-stdin", false, lang.CmdToolsGenKeyShortFlagPasswordStdin)
 
@@ -611,6 +613,24 @@ func newGenKeyCommand() *cobra.Command {
 }
 
 func (o *genKeyOptions) run(cmd *cobra.Command, _ []string) error {
+	prvKeyFileName := "cosign.key"
+	pubKeyFileName := "cosign.pub"
+
+	err := o.genKey(prvKeyFileName, pubKeyFileName)
+
+	if err != nil {
+		logger.From(cmd.Context()).Error("Failed to generate key pair")
+		return err
+	}
+
+	logger.From(cmd.Context()).Info("Successfully generated key pair",
+		"private-key-path", prvKeyFileName,
+		"public-key-path", pubKeyFileName)
+
+	return nil
+}
+
+func (o *genKeyOptions) genKey(prvKeyFileName string, pubKeyFileName string) error {
 	var passwordFunc func(bool) ([]byte, error)
 	// Utility function to prompt the user for the password to the private key
 	if o.interactive {
@@ -660,13 +680,15 @@ func (o *genKeyOptions) run(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("unable to generate key pair: %w", err)
 	}
 
-	prvKeyFileName := "cosign.key"
-	pubKeyFileName := "cosign.pub"
-
 	// Check if we are about to overwrite existing key files
 	_, prvKeyExistsErr := os.Stat(prvKeyFileName)
 	_, pubKeyExistsErr := os.Stat(pubKeyFileName)
-	if (prvKeyExistsErr == nil || pubKeyExistsErr == nil) && o.interactive {
+	var keyfilesExist bool
+	if prvKeyExistsErr == nil || pubKeyExistsErr == nil {
+		keyfilesExist = true
+	}
+
+	if keyfilesExist && o.interactive {
 		var confirm bool
 		confirmOverwritePrompt := &survey.Confirm{
 			Message: fmt.Sprintf(lang.CmdToolsGenKeyPromptExists, prvKeyFileName),
@@ -678,19 +700,14 @@ func (o *genKeyOptions) run(cmd *cobra.Command, _ []string) error {
 		if !confirm {
 			return errors.New("did not receive confirmation for overwriting key file(s)")
 		}
+	} else if keyfilesExist && !o.force {
+		return errors.New("cosign key file(s) already exist: pass --force to overwrite")
 	}
 
 	// Write the key file contents to disk
 	if err := os.WriteFile(prvKeyFileName, keyBytes.PrivateBytes, helpers.ReadWriteUser); err != nil {
 		return err
 	}
-	if err := os.WriteFile(pubKeyFileName, keyBytes.PublicBytes, helpers.ReadAllWriteUser); err != nil {
-		return err
-	}
 
-	logger.From(cmd.Context()).Info("Successfully generated key pair",
-		"private-key-path", prvKeyFileName,
-		"public-key-path", pubKeyFileName)
-
-	return nil
+	return os.WriteFile(pubKeyFileName, keyBytes.PublicBytes, helpers.ReadAllWriteUser)
 }
--- a/src/cmd/zarf_tools.go
+++ b/src/cmd/zarf_tools.go
@@ -602,6 +602,14 @@ func newGenKeyCommand() *cobra.Command {
 		Aliases: []string{"key"},
 		Short:   lang.CmdToolsGenKeyShort,
 		RunE:    o.run,
+		PreRunE: func(cmd *cobra.Command, _ []string) error {
+			if !o.interactive {
+				if !cmd.Flags().Changed("password") && !o.passwordStdin {
+					return errors.New(lang.CmdToolsGenKeyErrNoPasswordProvided)
+				}
+			}
+			return nil
+		},
 	}
 
 	cmd.Flags().BoolVar(&o.interactive, "interactive", false, lang.CmdToolsGenKeyShortFlagInteractive)
--- a/src/cmd/zarf_tools.go
+++ b/src/cmd/zarf_tools.go
@@ -626,6 +626,15 @@ func newGenKeyCommand() *cobra.Command {
 	return cmd
 }
 
+func (o *genKeyOptions) validatePassword(s string) error {
+	for _, c := range s {
+		if c < 32 || c > 126 {
+			return fmt.Errorf("invalid character (ascii code 0x%02X) in password", c)
+		}
+	}
+	return nil
+}
+
 func (o *genKeyOptions) run(cmd *cobra.Command, _ []string) error {
 	prvKeyFileName := "cosign.key"
 	pubKeyFileName := "cosign.pub"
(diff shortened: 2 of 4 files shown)


src/cmd/zarf_tools.go
```diff
@@ -599,36 +603,55 @@ func newGenKeyCommand() *cobra.Command {
 		RunE:    o.run,
 	}
 
+	cmd.Flags().BoolVar(&o.interactive, "interactive", true, lang.CmdToolsGenKeyShortFlagInteractive)
```
brandtkeller: great callout in the original description for defaulting to true. You were entirely right to look for retaining the same API and I appreciate that approach.

Given this is an imperative process that would not have been easily automatable previously - it should not be used much in automated workflows. My opinion here is that we can default this to false and change the behavior.

src/cmd/zarf_tools.go
```diff
@@ -599,36 +603,55 @@ func newGenKeyCommand() *cobra.Command {
...
+	cmd.Flags().StringVar(&o.password, "password", "", lang.CmdToolsGenKeyShortFlagPassword)
+	cmd.Flags().BoolVarP(&o.passwordStdin, "password-stdin", "", false, lang.CmdToolsGenKeyShortFlagPasswordStdin)
```
brandtkeller: Empty shorthand here - could probably revert to `BoolVar`

```suggestion
	cmd.Flags().BoolVar(&o.passwordStdin, "password-stdin", false, lang.CmdToolsGenKeyShortFlagPasswordStdin)
```

site/src/content/docs/commands/zarf_tools_gen-key.md
```diff
@@ -17,7 +17,10 @@ zarf tools gen-key [flags]
 ### Options
 
 ```
-  -h, --help   help for gen-key
+  -h, --help              help for gen-key
+      --interactive       Interactively prompt for registry server, username, and password if not provided
```
brandtkeller: nit: This looks like it might have been generated before the current state of the code (doesn't match the flag?).

src/cmd/zarf_tools.go
```diff
@@ -643,7 +666,7 @@ func (o *genKeyOptions) run(cmd *cobra.Command, _ []string) error {
 	// Check if we are about to overwrite existing key files
 	_, prvKeyExistsErr := os.Stat(prvKeyFileName)
 	_, pubKeyExistsErr := os.Stat(pubKeyFileName)
-	if prvKeyExistsErr == nil || pubKeyExistsErr == nil {
+	if (prvKeyExistsErr == nil || pubKeyExistsErr == nil) && o.interactive {
```
brandtkeller: I'd be open to possibly adding a separate overwrite flow that includes a `--force` flag. If this is not interactive then I think it would silently overwrite the files? We'd like to try and protect against that.

brandtkeller (changes requested):
Appreciative of this first pass - a few minor changes otherwise this is great. Just for local testing purposes I might recommend using a branch other than main for your development - but I could work-around it easily enough.

site/src/content/docs/commands/zarf_tools_gen-key.md
nicholas-rees: Oops 😅

src/cmd/zarf_tools.go
nicholas-rees: My mistake, I missed that, I updated the docs to make `--password-stdin` to have a shorthand flag `-p` [in the site doc ](https://github.com/zarf-dev/zarf/pull/4747/changes#diff-f2cc9f37b02905026b4e1109798b89057a43d469aa482e522f87031ac778c2c8R23).

Do you *want* to not have the shorthand flag? I was thinking it might be nice, but I also think its a little confusing to have -p (which password flag does it go to?).

I just want to confirm, because it being empty was a mistake, but I want to confirm that we want no shorthand flag so I can be sure to update the site docs too.

src/cmd/zarf_tools.go
nicholas-rees: I was thinking this, but I feel, in an automated workflow you generally expect things to be overwritten? I understand the hesitation though. I think a --force flag is a good release valve. In either case, I want to clarify: do you want this to throw an error if the key will be overwritten AND `--force` is false?

site/src/content/docs/commands/zarf_tools_gen-key.md
nicholas-rees: Hold on. On second thought, I might be doing something wrong here. I was manually updating the site docs. Do these docs get auto-generated? Have I missed something?

nicholas-rees:
> Appreciative of this first pass - a few minor changes otherwise this is great. Just for local testing purposes I might recommend using a branch other than main for your development - but I could work-around it easily enough.

To be honest, I read that in the contributing guide that you have short lived development branches with trunk based development, but forgot because of the way github requires you to fork to make a PR. 

Since the PR is already made, I'm just going to sin boldly, and leave it be until this is finished--unless otherwise requested, for the purposes of reducing confusion at this point.

src/cmd/zarf_tools.go
brandtkeller: I would advocate to remove the shorthand for now. We tend to defer minor assumptions around what _could_ be useful to future improvements when there are any potential questions or concerns about the shape of that improvement (IE UX).

site/src/content/docs/commands/zarf_tools_gen-key.md
brandtkeller: They are _not_ automatically updated - rather we try to catch the drift in CI. (technically we do check in pre-commit but it has limitations). 

Rather you need to run `make docs-and-schema`.

src/cmd/zarf_tools.go
brandtkeller: I like that approach - throw an error if overwriting and `--force` is `false`.

site/src/content/docs/commands/zarf_tools_gen-key.md
nicholas-rees: Thanks for that. That make command is noted on the contribution guide as well, and I missed it.

nicholas-rees:
A few notes: 

The biggest thing I that I want to make sure the force flag and interactive flag logic with the "keyfilesExist" check is appropriate. 

I also rebased and force pushed, because I realized that when I started I didn't add my keyfile to my git commits. Sorry for that confusion. If you'd like me to squash to one commit, please let me know.

Also I looked into increasing the code coverage for this function, but I'd need a little more direction on how you would like that done. [According to the tests page](https://docs.zarf.dev/contribute/testing/), it seems you would prefer e2e tests? But I want to be sure, before I go adding extra files to the repo.

I'm happy to do that work, but I would like just a tiny bit of direction.

brandtkeller:
> A few notes:
> 
> The biggest thing I that I want to make sure the force flag and interactive flag logic with the "keyfilesExist" check is appropriate.

I'll review this shortly.

> 
> I also rebased and force pushed, because I realized that when I started I didn't add my keyfile to my git commits. Sorry for that confusion. If you'd like me to squash to one commit, please let me know.

If it's not blocked by our branch protections then no worries. 

> Also I looked into increasing the code coverage for this function, but I'd need a little more direction on how you would like that done. [According to the tests page](https://docs.zarf.dev/contribute/testing/), it seems you would prefer e2e tests? But I want to be sure, before I go adding extra files to the repo.

Testing is often dependent on some of the required infrastructure. Given this operation is isolated to the CLI _and_ there is an existing `zarf_tools_test.go` file - that would be my first thought for coverage around testing this logic.

nicholas-rees:
I wrote some tests for gen-key.

I anticipate the review.

If I've gone overboard on the testing let me know.

src/cmd/zarf_tools.go
```diff
@@ -599,36 +604,85 @@ func newGenKeyCommand() *cobra.Command {
...
+func (o *genKeyOptions) genKey(prvKeyFileName string, pubKeyFileName string) error {
+	var passwordFunc func(bool) ([]byte, error)
 	// Utility function to prompt the user for the password to the private key
-	passwordFunc := func(bool) ([]byte, error) {
-		// perform the first prompt
-		var password string
-		prompt := &survey.Password{
-			Message: lang.CmdToolsGenKeyPrompt,
+
+	if len(o.password) > 0 && o.passwordStdin {
```
brandtkeller: We can shift this validation up to the command flags using `cmd.MarkFlagsMutuallyExclusive("password", "password-stdin")`

src/cmd/zarf_tools.go
```diff
@@ -599,36 +604,85 @@ func newGenKeyCommand() *cobra.Command {
...
+		return fmt.Errorf("cannot use --password with --password-stdin")
+	}
+
+	if o.interactive {
+		if len(o.password) > 0 || o.passwordStdin {
```
brandtkeller: same comment as above - this validation can be encapsulated in cobra natively.

src/cmd/zarf_tools.go
```diff
@@ -599,36 +604,85 @@ func newGenKeyCommand() *cobra.Command {
 		RunE:    o.run,
 	}
 
+	cmd.Flags().BoolVar(&o.interactive, "interactive", false, lang.CmdToolsGenKeyShortFlagInteractive)
+	cmd.Flags().BoolVar(&o.force, "force", false, lang.CmdToolsGenKeyShortFlagForce)
+	cmd.Flags().StringVar(&o.password, "password", "", lang.CmdToolsGenKeyShortFlagPassword)
+	cmd.Flags().BoolVar(&o.passwordStdin, "password-stdin", false, lang.CmdToolsGenKeyShortFlagPasswordStdin)
+
 	return cmd
 }
```
brandtkeller, line 628: Thinking about the UX - I think it's fair that a user passes in an empty string for password but I am not sure I want to implicitly accept an empty password if not-interactive but no flags are provided. 

Perhaps a PreRun here provides an option for this UX:
```go
cmd.PreRunE = func(cmd *cobra.Command, args []string) error {
    if !o.interactive {
        if !cmd.Flags().Changed("password") && !o.passwordStdin {
            return errors.New(lang.CmdToolsGenKeyErrNoPasswordProvided)
        }
    }
    return nil
}
```

src/cmd/zarf_tools_test.go
```diff
@@ -103,3 +103,96 @@ func TestGetCreds(t *testing.T) {
...
+			options: genKeyOptions{},
+		},
+		{
+			name: "gen key password",
+			options: genKeyOptions{
+				password: "test-password",
+			},
+		},
+		{
+			name: "gen key password-stdin",
```
brandtkeller, line 126: stepping through the debugger on the password-stdin tests and am currently not positive this is testing as we expect. Need to dig a little further,

src/cmd/zarf_tools.go
```diff
@@ -599,36 +604,85 @@ func newGenKeyCommand() *cobra.Command {
...
 
 func (o *genKeyOptions) run(cmd *cobra.Command, _ []string) error {
+	prvKeyFileName := "cosign.key"
+	pubKeyFileName := "cosign.pub"
+
+	err := o.genKey(prvKeyFileName, pubKeyFileName)
+
+	if err != nil {
+		logger.From(cmd.Context()).Error("Failed to generate key pair")
+		return err
```
brandtkeller, line 644: Need to look at the errors being returned here - cobra will already print the returned error - which might make the previous error message redundant/duplicitive.

brandtkeller (changes requested):
This is taking shape! appreciate the iterations.

src/cmd/zarf_tools_test.go
nicholas-rees: oops, you're right, I missed this. Will get on that.

src/cmd/zarf_tools.go
nicholas-rees: ok, cool so with that, the user can pass --password="" and still have an empty password, but without --password at all, it errors.

nicholas-rees: After thinking about this, do you have a similar opinion about an automated script that passes into --passwordStdin.

I actually think this might be a bigger concern since shell errors can fail but still pass nothing to the next pipe . . . Or is this on the user to use `set -o pipefail` or otherwise?

I'm not sure there's as clean a solution for passwordStdin, but the user does have an option? Add it to the documentation maybe?
i.e. --password-std in recommended to be used with set -o pipefail?

The other option might be to just explicitly disallow an emptystring from std in which might be a more sane approach.

nicholas-rees: I went ahead and just made an error and a test if an emptystring is passed into stdin. There's an escape hatch using --password="".

Given your reviews so far I think you'll agree with the change.

src/cmd/zarf_tools.go
nicholas-rees: I removed this. It is a bit duplicative. I don't really know why I thought that might be a good idea 😅

nicholas-rees:
post got duplicated sorry.

nicholas-rees:
One thing to note, when I disabled the the empty string for --password-stdin you can actually still:
```zsh
❯ echo "" | build/zarf-mac-apple tools gen-key --password-stdin
2026-04-01 11:51:12 INF using config file location=/Users/nick/code/git_repos/github/zarf-dev/zarf/zarf-config.toml
2026-04-01 11:51:12 INF Successfully generated key pair private-key-path=cosign.key public-key-path=cosign.pub
```
but:
```zsh
❯ printf "" | build/zarf-mac-apple tools gen-key --password-stdin
2026-04-01 11:52:02 INF using config file location=/Users/nick/code/git_repos/github/zarf-dev/zarf/zarf-config.toml
2026-04-01 11:52:02 ERR empty password disallowed when using --password-stdin
```
This is obviously because echo passes a `\n` , and I'm just checking for the empty string and not all whitespace characters. ~~Obviously this is not a problem for the `--password` flag~~ (see bottom).

I'm not sure how deep this should go for checking the errors, and I think maybe its not that great of a solution for example, if a shell variable isn't set or is set to an empty string:

```zsh
❯ TESTVAR=''
❯ echo "${TESTVAR}" | build/zarf-mac-apple tools gen-key --password-stdin --force
2026-04-01 11:56:03 INF using config file location=/Users/nick/code/git_repos/github/zarf-dev/zarf/zarf-config.toml
2026-04-01 11:56:03 INF Successfully generated key pair private-key-path=cosign.key public-key-path=cosign.pub
❯ printf "${TESTVAR}" | build/zarf-mac-apple tools gen-key --password-stdin --force
2026-04-01 11:56:09 INF using config file location=/Users/nick/code/git_repos/github/zarf-dev/zarf/zarf-config.toml
2026-04-01 11:56:09 ERR empty password disallowed when using --password-stdin
```
if you set `TESTVAR="$(<some-command>)"`, and that command fails, then you run into the same problem I mentioned above . . .

I could test for all whitespace, but I'm actually discovering that it seems like it might be legit to pass `\t` `\n` etc into a legit cosign password:

```zsh
❯ printf "\t" | build/zarf-mac-apple tools gen-key --password-stdin --force
2026-04-01 12:02:36 INF using config file location=/Users/nick/code/git_repos/github/zarf-dev/zarf/zarf-config.toml
2026-04-01 12:02:36 INF Successfully generated key pair private-key-path=cosign.key public-key-path=cosign.pub
❯ cosign sign-blob --key cosign.key --bundle /dev/null test.txt > /dev/null
Enter password for private key:
Error: signing test.txt: getting keypair and token: getting signer: reading key: decrypt: encrypted: decryption failed
error during command execution: signing test.txt: getting keypair and token: getting signer: reading key: decrypt: encrypted: decryption failed
```

But I have no idea how you're supposed to enter the tab character in the prompt? XD. I have no idea how this should be handled? Is it a problem with cosign? Or something we should do about?

If cosign allows for it, it seems like its a problem with cosign? But it also doesn't feel right. Let me know your opinion.

Yeah, I think this is definitely a real issue even for `--password` . . .

```zsh
❯ build/zarf-mac-apple tools gen-key --password="$(printf '\n\t\b')" --force
2026-04-01 12:16:38 INF using config file location=/Users/nick/code/git_repos/github/zarf-dev/zarf/zarf-config.toml
2026-04-01 12:16:38 INF Successfully generated key pair private-key-path=cosign.key public-key-path=cosign.pub
❯ cosign sign-blob --key cosign.key --bundle /dev/null test.txt > /dev/null
Enter password for private key:
Error: signing test.txt: getting keypair and token: getting signer: reading key: decrypt: encrypted: decryption failed
error during command execution: signing test.txt: getting keypair and token: getting signer: reading key: decrypt: encrypted: decryption failed
```

To be clear it's also a problem with the cosign package, [they're doing even less checks than we are](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/generate/generate_key_pair.go#L130).

```zsh
❯ echo "" | cosign generate-key-pair
Private key written to cosign.key
Public key written to cosign.pub
❯ cosign sign-blob --key cosign.key --bundle /dev/null test.txt > /dev/null
Enter password for private key:
Error: signing test.txt: getting keypair and token: getting signer: reading key: decrypt: encrypted: decryption failed
error during command execution: signing test.txt: getting keypair and token: getting signer: reading key: decrypt: encrypted: decryption failed
❯ printf "" | cosign generate-key-pair
Private key written to cosign.key
Public key written to cosign.pub
❯ cosign sign-blob --key cosign.key --bundle /dev/null test.txt > /dev/null
Enter password for private key:
Using payload from: test.txt
Wrote bundle to file /dev/null
```

nicholas-rees:
I wrote a little password validation trying to whitelist only printable ascii characters. Definitely need a review on that to be sure its good.

It will also fail with echo every time, but not with printf.

I suppose I could just strip the newlines? I'm not sure what should be stripped and what shouldn't. I am neither going to confirm nor deny whether or not I use spaces in my passwords, but i definitely don't use newlines. . . 

Someone might want to cat a file, so what should happen? Should new lines just be stripped, replaced with spaces? Personally, I think at this point, at least tell the user that its invalid, and they can use whatever utility or logic, as long as it doesn't contain non printable ascii. printf is way better than echo anyway. . .

nicholas-rees:
I pulled in changes from main and repushed. Also, there was a slight issue with interactive mode that I fixed when I went back to make sure everything was working after the rebase.

src/cmd/zarf_tools.go
brandtkeller: agreed - this sounds sufficient.

src/cmd/zarf_tools.go
```diff
@@ -597,21 +603,68 @@ func newGenKeyCommand() *cobra.Command {
...
+		"public-key-path", pubKeyFileName)
+
+	return nil
+}
+
+func (o *genKeyOptions) genKey(prvKeyFileName string, pubKeyFileName string) error {
+	var password []byte
+	var err error
+	var passwordFunc func(bool) ([]byte, error)
 	// Utility function to prompt the user for the password to the private key
```
brandtkeller: Nit: This comment looks slightly out-of-place now. Can probably remove.

src/cmd/zarf_tools.go
```diff
@@ -597,21 +603,68 @@ func newGenKeyCommand() *cobra.Command {
...
+	}
+	return nil
+}
+
 func (o *genKeyOptions) run(cmd *cobra.Command, _ []string) error {
+	prvKeyFileName := "cosign.key"
+	pubKeyFileName := "cosign.pub"
+
+	err := o.genKey(prvKeyFileName, pubKeyFileName)
+
```
brandtkeller: nit: remove empty line

src/cmd/zarf_tools.go
```diff
@@ -620,15 +673,35 @@ func (o *genKeyOptions) run(cmd *cobra.Command, _ []string) error {
...
 		}
+		if len(password) == 0 {
+			return fmt.Errorf("empty password disallowed when using --password-stdin")
+		}
+	} else {
+		password = []byte(o.password)
+	}
 
-		return []byte(password), nil
+	passwordFunc = func(bool) ([]byte, error) {
```
brandtkeller: Given this is the only assignment of `passwordFunc` I think it can serve as the primary declaration and the  variable declared above can be removed.

src/cmd/zarf_tools.go
```diff
@@ -597,21 +603,68 @@ func newGenKeyCommand() *cobra.Command {
...
+	}
+
+	logger.From(cmd.Context()).Info("Successfully generated key pair",
+		"private-key-path", prvKeyFileName,
+		"public-key-path", pubKeyFileName)
+
+	return nil
+}
+
+func (o *genKeyOptions) genKey(prvKeyFileName string, pubKeyFileName string) error {
```
brandtkeller: nit: collapse parameter styling:

```suggestion
func (o *genKeyOptions) genKey(prvKeyFileName, pubKeyFileName string) error {
```

src/cmd/zarf_tools.go
```diff
@@ -620,15 +673,35 @@ func (o *genKeyOptions) run(cmd *cobra.Command, _ []string) error {
...
+		if len(password) == 0 {
+			return fmt.Errorf("empty password disallowed when using --password-stdin")
+		}
+	} else {
+		password = []byte(o.password)
+	}
 
-		return []byte(password), nil
+	passwordFunc = func(bool) ([]byte, error) {
+		err = o.validatePassword(password)
```
brandtkeller: I love the idea of validating the password - I do wonder about when we execute it. This won't validate until the cosign logic executes. I think it might be more deterministic to validate earlier.

src/cmd/zarf_tools.go
brandtkeller: If I am thinking about this correctly - we have errors that may be introduced below based in interactivity and whether or not the `force` flag is used. I think we could therefor place this logic _after_ the validation logic for overwrite in order to not perform any processing before all validation is complete?

src/cmd/zarf_tools.go
```diff
@@ -597,21 +603,68 @@ func newGenKeyCommand() *cobra.Command {
...
+	cmd.Flags().BoolVar(&o.force, "force", false, lang.CmdToolsGenKeyShortFlagForce)
+	cmd.Flags().StringVar(&o.password, "password", "", lang.CmdToolsGenKeyShortFlagPassword)
+	cmd.Flags().BoolVar(&o.passwordStdin, "password-stdin", false, lang.CmdToolsGenKeyShortFlagPasswordStdin)
+	cmd.MarkFlagsMutuallyExclusive("password", "password-stdin", "interactive")
+	cmd.MarkFlagsMutuallyExclusive("interactive", "force")
+
 	return cmd
 }
 
+func (o *genKeyOptions) validatePassword(bs []byte) error {
```
brandtkeller: This looks appropriate as a private function - I don't see any use of the `genKeyOptions` that makes it required as a method?

brandtkeller (changes requested):
Appreciate the patience! Adding some additional thoughts.

src/cmd/zarf_tools.go
nicholas-rees: I'm not sure that I'm following you here. I think I understand why you want the validation before any processing. But, I'm not following you on errors based in interactivity.

1. [interactive and force are mutually exlusive](https://github.com/zarf-dev/zarf/pull/4747/changes#diff-20c786fb844bc36af875137ebd5659fce7ce40e0a7a46214547729af9a26cf00R624), so it shouldn't even be possible for this situation to happen.
2.  if interactive is running, and there are files that are determined to exist, the force flag would be ignored (going back to why I made the flags mutually exclusive in the first place).

My logic was that if you choose to run in interactive, then you shouldn't really be using any cli options at all--which kind of defeats the purpose of running interactively...

The end result, I think, is the same. I still think it makes sense to move the validation logic before any cosign logic, I'd just like to better understand what you're meaning, because I feel like I'm missing something here.

brandtkeller: github is rendering this poorly but the idea was that we should call `cosign.GenerateKeyPair` after all the validation logic - to include checking for interactive/force (mutually exclusive) and overwrite logic. 

`cosign.GenerateKeyPair` should only be executed if all pre-validation has passed. If we have a deterministic issue such as "key files exist but the force flag wasn't used - then we shouldn't even execute any cosign logic.

nicholas-rees: makes sense, I think I might have just been making it more complicated.

src/cmd/zarf_tools.go
```diff
@@ -620,30 +670,43 @@ func (o *genKeyOptions) run(cmd *cobra.Command, _ []string) error {
...
-		return []byte(password), nil
+		password = []byte(tmpPassword)
+	} else if o.passwordStdin {
+		if o.reader == nil {
+			return fmt.Errorf("reader not set for genKeyOptions")
+		}
+		password, err = io.ReadAll(o.reader)
+		if err != nil {
+			return err
+		}
```
brandtkeller, line 688: You made a good callout previously for trimming newlines. 

IE 
```
echo "test" | ../build/zarf-mac-apple tools gen-key --password-stdin
2026-04-15 12:46:37 ERR invalid character (ascii code 0x0A) in password
```

I don't think this is entirely intuitive but I think we could keep this simple for now until other feedback is received. 


```suggestion
		}
		password = bytes.TrimSuffix(password, []byte("\n"))
        password = bytes.TrimSuffix(password, []byte("\r"))
```

src/cmd/zarf_tools_test.go
```diff
@@ -103,3 +103,96 @@ func TestGetCreds(t *testing.T) {
...
+func TestGenKey(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name       string
+		options    genKeyOptions
+		keysExist  bool
+		shouldFail bool
+	}{
+		{
+			name:    "gen key",
```
brandtkeller:
```suggestion
			name:    "gen key empty password",
```

brandtkeller (changes requested):
minor go-backs before approval. Testing locally with the suggested changes appears to be intuitive.

src/cmd/zarf_tools.go
brandtkeller: Might want to check formatting on this suggestion (or just replicate it yourself)

nicholas-rees: This concerns me. I'm happy to remove characters that shouldn't exist, but I think the correct approach is to simply take the validation script, and remove all bytes that shouldn't exist, basically printiable ascii+ space character (everything between 32-126, perhaps also rename it from validatePassword to sanitizePassword)
Like I said before, this is a cosign issue--not a zarf issue. Cosign allows you to assign any byte as a password. So I ask, why stop at new lines. Should a tab character be permitted?
My opinion, I'd leave it as it is. Its on the user to use cosign, and its how the actual cosign binary currently works. However, if you insist on modifying the behavior of cosign, I suggest giving me a list of all characters that should be allowed in a password, and I will remove them.
Your solution still leaves many characters being inputted into a password that will cause problems. I get that the validation script will catch them, but I'm unsure how its any less intutive. I think its less intuitive for echo to magically work than for it not to.

brandtkeller: Those are appropriate questions to ask and I think there _could_ be a fair argument for it needing to be fixed in Cosign upstream. 

I'd also wager this is some of the intention of the cosign `PassFunc` - but here we are validating eagerly before this is ever called. 

So less from an extensive list of "what is correct" to asking "what is intuitive?". 

I don't think `echo -n` is going to be the obvious choice but I also don't think we need an exhaustive list of what is permitted until we have signals to support their use. Replicating what [docker](https://github.com/docker/cli/blob/master/cli/command/registry/login.go#L104) does is where I was leaning towards for prior art.  

Thoughts?

nicholas-rees: I have trouble with this because, what is intuitive to me is generally not intuitive to other people. I think modifying any of the data without informing the user can be a bit of an issue, which is why I made the validation function in the first place.

What if, we logged a warning which will strip a single \r and a single \n, informing the user that they shouldn't do that, but that we are going to modify their input to "make it work"? At least then we're being explicit about what is happening?

brandtkeller: we can certainly give it a try. One of the zarf personas is a low level IT user using zarf. If they were to run `echo "mysecretpassword" | zarf tools gen-key --password-stdin` would they understand why we're logging a warning for the existence of stripped characters that they never provided?

nicholas-rees: This is what is confusing me. If they didn't provide the \r \n, then how did the \r \n get there? If they just don't understand they're being put there, isn't this a great learning opportunity for low level it users?

brandtkeller: Also a good question - That's the POSIX/bash default behavior. echo is defined to print its arguments followed by a line terminator. Part of why we see trimming these characters by default across CLI's.

nicholas-rees: If its standard behavior then why would they think they didn't provide the \n? They did that when they chose to use echo.

Also, I'm not suggesting anyone ever use echo -n. That's not posix compliant either. I don't want to get into the weeds of posix compliance here, but the general recommendation is to printf because then you do have direct control as a user over the formatting of the output.

I leave the decision to you.

brandtkeller: My only concern is that the user experience aligns to what already exists. I mentioned the `zarf tools registry login` command originally as having behaviors we would want to replicate - and we see [this pattern](https://github.com/google/go-containerregistry/blob/main/cmd/crane/cmd/auth.go#L243-L251) used here as well.

brandtkeller:
@nicholas-rees friendly nudge - are you still interested (and have time) to see this completed? I think it's nearly there.

brandtkeller:
Friendly nudge - this one is close but looking to determine whether this work is going to see completion or closing as stale to free up the issue.

Issue #3828, opened 2025-05-20
Zarf should respect XDG_CACHE_HOME


eddiezane:
### Is your feature request related to a problem? Please describe.

The [XDG Base Directory Specification](https://wiki.archlinux.org/title/XDG_Base_Directory) is used to direct programs where they should be storing files in a users home directory.

Currently zarf stores all downloaded init packages in `$HOME/.zarf-cache` and does not respect `XDG_CACHE_HOME` if it is set.

### Describe the behavior you'd like

- **Given** A user has `XDG_CACHE_HOME` set on their system
- **When** Zarf downloads an init package
- **Then** It should store it in `XDG_CACHE_HOME/zarf`

### Describe alternatives you've considered

I believe this can be configured with a viper config file, but we should still probably respect XDG.

### Additional context

This is low priority.

We might want to transition to this being the default. I've used https://github.com/adrg/xdg before for this.

brandtkeller:
See https://github.com/zarf-dev/zarf/pull/4765#discussion_r3074986128 for information around how we can introduce this prior to `v1.0.0`.

> One way to handle tasks like this, without introducing PRs that need to be worked right before we hit v1.0, would to add a feature flag, --new-cache-home, that we keep disabled until v1.0.

PR #4765, opened 2026-03-30
fix(cache): support for default sdk cache path

### LINKED ISSUE #4764: SDK Default Cache Path Resolution

Zarf's src/pkg/packager package exposes public functions (Create, LoadPackage, DevDeploy, Pull, FindImages, PublishSkeleton, Publish, Lint, Inspect) that each accept a CachePath string in their options. Today, every one of these functions passes CachePath directly to downstream consumers — images.Pull, zoci.GetOCICacheModifier, createImageSBOM, helm chart packaging — all of which assume the path is set and the directory exists.

This works for the CLI because getCachePath in src/cmd/cmd.go always resolves a path before calling into the library. But the library layer itself has no safety net. An SDK consumer who omits CachePath (the zero value "") hits failures deep in the call stack: zoci.GetOCICacheModifier tries to open ""/images as an OCI store, images.Pull calls helpers.CreateDirectory(""), and stereoscope's layer cache writes to a path that doesn't exist.

This matters because Zarf is actively being adopted as a library. The src/pkg/packager functions are the public API surface, and every one of them has an undocumented requirement that CachePath must be non-empty and point to a valid location. SDK consumers have no way to discover this except by hitting runtime errors. The cache is also not truly optional — it's wired into OCI pulls (orasCache.New wraps every remote registry fetch), SBOM generation (stereoscope uses it for uncompressed layer storage), and helm chart operations.

brandtkeller:
## Description

Zarf's public `src/pkg/packager/` functions all accept a `CachePath` string in their options. The CLI always sets this via `getCachePath()`, but SDK consumers who omit it get an empty string that causes failures deep in the call stack (`zoci.GetOCICacheModifier`, `images.Pull`, `createImageSBOM`, helm operations). This change adds a `GetCachePath` utility and applies it at every public entry point so SDK consumers get a working default (`os.UserCacheDir()/zarf`) without needing CLI knowledge. 

CLI behavior is unchanged — the CLI always passes an explicit path but worth acknowledging optionality here to change the CLI to support #3828 .

There is some redundancy across some transient load operations where this is called multiple times. Given the cases all are subsequently public functions and the operation is idempotent I think this is reasonable. 

## Related Issue

Fixes #4764

## Checklist before merging

### DIFF
--- a/src/pkg/utils/io.go
+++ b/src/pkg/utils/io.go
@@ -17,9 +17,9 @@ const (
 	tmpPathPrefix = "zarf-"
 )
 
-// GetCachePath returns cachePath if non-empty, otherwise falls back to
+// ResolveCachePath returns cachePath if non-empty, otherwise falls back to
 // filepath.Join(os.UserCacheDir(), "zarf") which respects XDG_CACHE_HOME on Linux.
-func GetCachePath(cachePath string) (string, error) {
+func ResolveCachePath(cachePath string) (string, error) {
 	if cachePath != "" {
 		return cachePath, nil
 	}
--- a/src/pkg/packager/create.go
+++ b/src/pkg/packager/create.go
@@ -49,7 +49,7 @@ func Create(ctx context.Context, packagePath string, output string, opts CreateO
 		return "", fmt.Errorf("cannot skip SBOM creation and specify an SBOM output directory")
 	}
 
-	opts.CachePath, err = utils.GetCachePath(opts.CachePath)
+	opts.CachePath, err = utils.ResolveCachePath(opts.CachePath)
 	if err != nil {
 		return "", err
 	}
--- a/src/pkg/packager/dev.go
+++ b/src/pkg/packager/dev.go
@@ -62,7 +62,7 @@ func DevDeploy(ctx context.Context, packagePath string, opts DevDeployOptions) (
 		opts.Timeout = config.ZarfDefaultTimeout
 	}
 
-	opts.CachePath, err = utils.GetCachePath(opts.CachePath)
+	opts.CachePath, err = utils.ResolveCachePath(opts.CachePath)
 	if err != nil {
 		return err
 	}
--- a/src/pkg/packager/find_images.go
+++ b/src/pkg/packager/find_images.go
@@ -86,7 +86,7 @@ type ComponentImageScan struct {
 func FindImages(ctx context.Context, packagePath string, opts FindImagesOptions) (_ []ComponentImageScan, err error) {
 	l := logger.From(ctx)
 
-	opts.CachePath, err = utils.GetCachePath(opts.CachePath)
+	opts.CachePath, err = utils.ResolveCachePath(opts.CachePath)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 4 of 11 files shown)


AustinAbro321:
@brandtkeller I am curious what your thoughts are on allowing users not to cache resources at all, there are some blockers here, see #3774. Right now Zarf assumes and relies on a cachepath existing in some spots, but there should be a future state where a user could preform operations without a cache at all. 

I think there are two paths for this: 
1. `--no-cache` flag / option, this would work with this PR
2. An empty cachepath to SDK calls or explicitly setting --zarf-cache="" would allow for no cache usage.

I think option 1 is a good approach. Wanted to get your thoughts on this first, then I can update #3774 with the decision.

brandtkeller:
> I am curious what your thoughts are on allowing users not to cache resources at all, there are some blockers here, see #3774. Right now Zarf assumes and relies on a cachepath existing in some spots, but there should be a future state where a user could preform operations without a cache at all.
> 
> I think there are two paths for this:
> 
> 1. `--no-cache` flag / option, this would work with this PR
> 2. An empty cachepath to SDK calls or explicitly setting --zarf-cache="" would allow for no cache usage.
> 
> I think option 1 is a good approach. Wanted to get your thoughts on this first, then I can update #3774 with the decision.

I'm in the same boat. I thought about this originally and decided there were a few constraints (as I understood them) where even a temporary cache is strictly required (I believe Helm errors and syft has a requirement) - but to where we could introduce an additional option that would essentially make the tempDir ephemeral per-cache-required function call.

src/pkg/utils/io.go
```diff
@@ -17,6 +17,19 @@ const (
 	tmpPathPrefix = "zarf-"
 )
 
+// ResolveCachePath returns cachePath if non-empty, otherwise falls back to
+// filepath.Join(os.UserCacheDir(), "zarf") which respects XDG_CACHE_HOME on Linux.
+func ResolveCachePath(cachePath string) (string, error) {
+	if cachePath != "" {
+		return cachePath, nil
+	}
+	cacheDir, err := os.UserCacheDir()
```
AustinAbro321, line 26: Curious what your thoughts are for the future default of the CLI? Do we see it changing to also `ResolveCachePath` maybe in v1.0 or will it always stay as `~/.zarf-cache`? 

One strategy we could use would be to create function similar to `os.UserCacheDir`, but only return a value when environment variables are set. Otherwise we would keep a default of `~/.zarf-cache` for consistency through the CLI and SDK.

PR looks good otherwise.

brandtkeller: That is definitely the question... My vote is to move towards `ResolveCachePath` in the CLI in `v1.0` so that CLI and SDK align and we can further evaluate the requirements of using the `config` package.

AustinAbro321: Seems like a good path to me as well.

One way to handle tasks like this, without introducing PRs that need to be worked right before we hit v1.0, would to add a feature flag, `--new-cache-home`, that we keep disabled until v1.0.

brandtkeller: That is a great callout - I'll link this to the outsanding issue

PR #4766, opened 2026-03-31
fix(docs): update broken acknowledgment link in README

willswire:
## Description

Updates the broken acknowledgement link in the README.

## Related Issue

Fixes #

Relates to #

## Checklist before merging

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ To dive deeper into the tech, you can read the [Nerd Notes](https://docs.zarf.de
 
 ## ⭐️ Special Thanks
 
-> Early Zarf research and prototypes were developed jointly with [United States Naval Postgraduate School](https://nps.edu/) research you can read [here](https://calhoun.nps.edu/handle/10945/68688).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,7 +94,7 @@ To dive deeper into the tech, you can read the [Nerd Notes](https://docs.zarf.de
 
 ## ⭐️ Special Thanks
 
-> Early Zarf research and prototypes were developed jointly with [United States Naval Postgraduate School](https://nps.edu/) research you can read [here](https://calhoun.nps.edu/handle/10945/68688).
+> Early Zarf research and prototypes were developed jointly with [United States Naval Postgraduate School](https://nps.edu/) research you can read [here](https://calhoun.nps.edu/entities/publication/d92c5122-6793-41df-b603-9c4faed1911f).
```
brandtkeller, line 97: Can you clarify how this link is broken? 

The original redirects me to your fixed link automatically - albeit I am seeing 500 errors for either periodically. 

I'm still open to changing this if one of the references is more correct.

brandtkeller: Reading around on the handle system - I'd wager the original URL is the most correct. It’s the repository’s persistent identifier, which is generally a better fit for long-term historical references than an application-specific page URL.

That is provided there is not some crucial error I am missing.

brandtkeller:
the redirect is still operational for me when tested this morning. Open to changing this if there is a more correct option - closing this for now.

PR #4769, opened 2026-03-31
feat(tls): support for tls generation with duration

### LINKED ISSUE #4767: Support Configurable Agent TLS Certificate Expiration

Currently, the Zarf Agent's TLS certificates are hardcoded with a validFor constant (approx. 375 days). For systems deployed in disconnected or Read-Only (RO) environments that may remain in the field for years without updates, this 1-year expiration creates a hard failure point.

Users need a way to configure the agent certificate lifetime (e.g., up to 10 years) during the initialization process to ensure long-term stability in air-gapped or immutable environments.

Given a user is deploying Zarf to a long-lived, disconnected environment.
When the user executes the zarf init command with a specified certificate duration flag (e.g., --agent-cert-duration=87600h)
Then the Zarf Agent TLS certificates and CA should be generated with the requested expiration, and the MutatingWebhookConfiguration and zarf-state secret should reflect this extended validity period.

- Generate certs externally — Create a CA + server cert/key with desired lifetime; server cert SAN must include agent-hook.zarf.svc
- Replace agent-hook-tls secret — Delete and recreate the TLS secret in the zarf namespace with the new cert and key
- Restart agent pods — kubectl -n zarf rollout restart deployment agent-hook so pods serve the new cert
- Verify agent health — Confirm the restarted pods are running and the liveness probe at /healthz:8443 passes
- Patch MutatingWebhookConfiguration zarf — Update caBundle on all 8 webhook entries to the base64-encoded new CA cert
- Update zarf-state secret — Extract the state JSON, replace agentTLS.ca, agentTLS.cert, and agentTLS.key with the new PEM values, write it back
- Avoid running zarf tools update-creds agent afterward — It will overwrite everything with a fresh 375-day cert generated from the hardcoded validFor constant

brandtkeller:
## Description

Adds support for `--duration` to `zarf tools gen-pki` in order to support a more composable model of custom certificate requirements. 

This is then made compatible with the zarf agent tls (as outlined in the issue) by allowing `zarf init` to accept agent tls ca,cert,key. 

```bash
# generate tls ca,cert,key with a 10 year duration
zarf tools gen-pki agent-hook.zarf.svc --duration 87600h
```

```bash
# pass provided tls information above to zarf init
zarf init --agent-tls-ca tlstest/tls.ca --agent-tls-cert tlstest/tls.crt --agent-tls-key tlstest/tls.key
```

## Related Issue

Fixes #4767

Relates to #

## Checklist before merging

### DIFF
--- a/.github/workflows/nightly-ecr.yml
+++ b/.github/workflows/nightly-ecr.yml
@@ -22,7 +22,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Auth with AWS
-        uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6.0.0
+        uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # v6.1.0
         with:
           role-to-assume: ${{ secrets.AWS_NIGHTLY_ECR_ROLE_ARN }}
           role-session-name: ${{ github.job || github.event.client_payload.pull_request.head.sha || github.sha }}
--- a/.github/workflows/nightly-eks.yml
+++ b/.github/workflows/nightly-eks.yml
@@ -39,7 +39,7 @@ jobs:
         uses: ./.github/actions/packages
 
       - name: Auth with AWS
-        uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6.0.0
+        uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # v6.1.0
         with:
           role-to-assume: ${{ secrets.AWS_NIGHTLY_EKS_ROLE_ARN }}
           role-session-name: ${{ github.job || github.event.client_payload.pull_request.head.sha || github.sha }}
--- a/.github/workflows/nightly-release.yaml
+++ b/.github/workflows/nightly-release.yaml
@@ -102,7 +102,7 @@ jobs:
           echo ZARF_AGENT_IMAGE_DIGEST=$(docker buildx imagetools inspect ghcr.io/zarf-dev/zarf/agent:$CLI_VERSION --format '{{ json . }}' | jq -r .manifest.digest) >> $GITHUB_ENV
 
       - name: Auth with AWS
-        uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6.0.0
+        uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # v6.1.0
         with:
           role-to-assume: ${{ secrets.AWS_KMS_ROLE_ARN }}
           role-session-name: ${{ github.job || github.event.client_payload.pull_request.head.sha || github.sha }}
@@ -129,7 +129,7 @@ jobs:
           CLI_VERSION=$CLI_VERSION make publish-init-package ARCH=arm64 REPOSITORY_URL=ghcr.io/zarf-dev/packages
 
       - name: Upload build artifacts
-        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: build-artifacts
           path: build/
--- a/.github/workflows/publish-application-packages.yml
+++ b/.github/workflows/publish-application-packages.yml
@@ -17,7 +17,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Auth with AWS
-        uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6.0.0
+        uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # v6.1.0
         with:
           role-to-assume: ${{ secrets.AWS_KMS_ROLE_ARN }}
           role-session-name: ${{ github.job || github.event.client_payload.pull_request.head.sha || github.sha }}
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - name: Get Release Please token
         id: app-token
-        uses: actions/create-github-app-token@f8d387b68d61c58ab83c6c016672934102569859 # v3.0.0
+        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
         with:
           app-id: ${{ secrets.RELEASE_PLEASE_GITHUB_APP_ID }}
           private-key: ${{ secrets.RELEASE_PLEASE_GITHUB_APP_SECRET }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,7 +51,7 @@ jobs:
           echo ZARF_AGENT_IMAGE_DIGEST=$(docker buildx imagetools inspect ghcr.io/zarf-dev/zarf/agent:$GITHUB_REF_NAME --format '{{ json . }}' | jq -r .manifest.digest) >> $GITHUB_ENV
 
       - name: Auth with AWS
-        uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # v6.0.0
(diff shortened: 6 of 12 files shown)


src/cmd/zarf_tools.go
```diff
@@ -559,17 +584,23 @@ func newGenPKICommand() *cobra.Command {
...
 
 	cmd.Flags().StringArrayVar(&o.subAltNames, "sub-alt-name", []string{}, lang.CmdToolsGenPkiFlagAltName)
+	cmd.Flags().DurationVar(&o.duration, "duration", time.Hour*24*375, "Duration for the generated certificates (e.g., 8760h for 1 year, 87600h for ~10 years)")
 
 	return cmd
 }
 
 func (o *genPKIOptions) run(cmd *cobra.Command, args []string) error {
-	pki, err := pki.GeneratePKI(args[0], o.subAltNames...)
+	pki, err := pki.GeneratePKIWithDuration(args[0], o.duration, o.subAltNames...)
```
AustinAbro321: Thoughts on `GeneratePKIWithOptions` to avoid needing another function in the future?

src/cmd/initialize.go
```diff
@@ -344,6 +368,29 @@ func validateExistingStateMatchesInput(ctx context.Context, registryInfo state.R
...
+	key, err := os.ReadFile(keyPath)
+	if err != nil {
+		return pki.GeneratedPKI{}, fmt.Errorf("unable to read agent TLS key: %w", err)
+	}
+	if _, err := tls.X509KeyPair(cert, key); err != nil {
+		return pki.GeneratedPKI{}, fmt.Errorf("agent TLS cert and key do not match: %w", err)
+	}
+	if err := pki.ValidateCertSAN(cert, state.ZarfAgentHost); err != nil {
+		return pki.GeneratedPKI{}, err
+	}
```
AustinAbro321: I believe we can replace this function with something like this
```go
	parsed, err := pki.ParseCertFromPEM(cert)
	if err != nil {
		return pki.GeneratedPKI{}, fmt.Errorf("failed to parse agent TLS certificate: %w", err)
	}
	caPool := x509.NewCertPool()
	if !caPool.AppendCertsFromPEM(ca) {
		return pki.GeneratedPKI{}, fmt.Errorf("failed to parse provided CA certificate")
	}
	if _, err := parsed.Verify(x509.VerifyOptions{
		Roots:     caPool,
		DNSName:   state.ZarfAgentHost,
	}); err != nil {
		return pki.GeneratedPKI{}, fmt.Errorf("agent TLS certificate failed validation: %w", err)
	}
```
An added benefit is that we can also use this to check that the root is accurate

src/cmd/zarf_tools.go
```diff
@@ -310,6 +320,19 @@ func (o *updateCredsOptions) run(cmd *cobra.Command, args []string) error {
...
+	if slices.Contains(args, state.AgentKey) {
+		if o.agentTLSCAPath != "" {
+			loadedTLS, err := loadAndValidateAgentTLS(o.agentTLSCAPath, o.agentTLSCertPath, o.agentTLSKeyPath)
+			if err != nil {
+				return fmt.Errorf("invalid agent TLS certificates: %w", err)
+			}
+			opts.AgentTLS = &loadedTLS
+		} else if oldState.AgentTLSUserProvided {
+			l.Warn("the current agent TLS certificates were user-provided; running update-creds agent without --agent-tls-* flags will replace them with auto-generated certificates")
+		}
```
AustinAbro321: This is an interesting situation. While replacing user submitted passwords is the current behavior for text passwords like `--registry-push-password`, I'm not sure this is something I'd like to propagate with certificates. My tenative idea for the registry proxy was only updating registrys certs when they are zarf-managed.  I could imagine someone running `zarf tools update-creds --confirm` without seeing the warning or seeing it too late. 

What are your thoughts on erroring here instead? We could also have a flag like `--replace-user-managed-agent-tls`, though that feels like a very unlikely use case. I imagine users with their own tls have a very specific reason for doing so that they'd not want to override.

src/cmd/zarf_tools.go
brandtkeller: I like that plan - I was too hyperfocused on a single path.

src/cmd/zarf_tools.go
brandtkeller: Given the flags are exposed here I am comfortable with erroring to prevent any unwarranted overwrite and requiring this set of users to regenerate/replace with the flags if needed.

src/cmd/initialize.go
brandtkeller: good call! removing.

src/cmd/zarf_tools.go
brandtkeller: removed from the existing conditional and left-aligned.

src/cmd/initialize.go
AustinAbro321, line 1: Agent flags will be silently ignored here on secondary init. We should add agent logic to `validateExistingStateMatchesInput`

src/cmd/zarf_tools.go
```diff
@@ -310,6 +320,20 @@ func (o *updateCredsOptions) run(cmd *cobra.Command, args []string) error {
 		ArtifactServer: o.artifactServer,
 		Services:       args,
 	}
+
+	if slices.Contains(args, state.AgentKey) {
+		if oldState.AgentTLSUserProvided && o.agentTLSCAPath == "" {
+			return fmt.Errorf("current agent TLS certificates are user-provided; provide --agent-tls-ca, --agent-tls-cert, and --agent-tls-key to update them, or remove 'agent' from the service list to skip")
```
AustinAbro321:
```suggestion
			return fmt.Errorf("current agent TLS certificates are user-provided; provide --agent-tls-ca, --agent-tls-cert, and --agent-tls-key to update them, or explicitly define service list without `agent`")
```
nit: adjust the error message since the default case includes all services.

AustinAbro321 (changes requested):
UX consideration on re-init

src/cmd/initialize.go
brandtkeller: good catch - added early validation.

src/cmd/initialize.go
```diff
@@ -104,6 +111,11 @@ func newInitCommand() *cobra.Command {
 	cmd.Flags().StringVar(&o.artifactServer.PushUsername, "artifact-push-username", v.GetString(VInitArtifactPushUser), lang.CmdInitFlagArtifactPushUser)
 	cmd.Flags().StringVar(&o.artifactServer.PushToken, "artifact-push-token", v.GetString(VInitArtifactPushToken), lang.CmdInitFlagArtifactPushToken)
 
+	// Flags for providing user-managed agent TLS certificates
+	cmd.Flags().StringVar(&o.agentTLSCAPath, "agent-tls-ca", "", "Path to a PEM-encoded CA certificate for the Zarf agent")
+	cmd.Flags().StringVar(&o.agentTLSCertPath, "agent-tls-cert", "", "Path to a PEM-encoded TLS certificate for the Zarf agent")
+	cmd.Flags().StringVar(&o.agentTLSKeyPath, "agent-tls-key", "", "Path to a PEM-encoded TLS private key for the Zarf agent")
```
AustinAbro321: Thoughts on introducing viper config for these? init config flags can get quite verbose

src/cmd/initialize.go
```diff
@@ -350,9 +375,46 @@ func validateExistingStateMatchesInput(ctx context.Context, registryInfo state.R
 	if helpers.IsNotZeroAndNotEqual(artifactServer, s.ArtifactServer) {
 		return fmt.Errorf("cannot change artifact server information after initial init, to update run `zarf tools update-creds artifact`")
 	}
+	if agentTLSProvided && s.AgentTLSUserProvided {
+		return fmt.Errorf("cannot change agent TLS certificates after initial init, to update run `zarf tools update-creds agent`")
+	}
```
AustinAbro321, line 385: This should error even when tls is not user provided, as it would also be skipped in that case. 

Another option would be checking if the certs are the same as the certs in cluster, this way if someone is re-running init with the same flags it would work

src/cmd/initialize.go
brandtkeller: Added!

src/cmd/initialize.go
brandtkeller: It wasn't something I thought about initially but between the viper config flags and the realistic scenario where someone is using the same command with the same certs to upgrade - I decided to implement the comparison of state to check for change.

src/cmd/initialize.go
```diff
@@ -350,9 +376,50 @@ func validateExistingStateMatchesInput(ctx context.Context, registryInfo state.R
 	if helpers.IsNotZeroAndNotEqual(artifactServer, s.ArtifactServer) {
 		return fmt.Errorf("cannot change artifact server information after initial init, to update run `zarf tools update-creds artifact`")
 	}
+	if agentTLS != nil && s.AgentTLSUserProvided {
```
AustinAbro321:
```suggestion
	if agentTLS != nil {
```
I would still expect an error here even if the original certs weren't user provided. For instance if I ran
```bash
zarf init 
```
Then ran 
```bash
zarf init --agent-tls-ca tls.ca --agent-tls-cert tls.crt --agent-tls-key tls.key
```
the certs I input will have no effect, but I won't receive any signal for this.

PR #4783, opened 2026-04-03
feat(lint): add --all-variants flag for lint

### LINKED ISSUE #4598: zarf dev lint fails if a flavor is not specified for a flavored package
### Environment
Device and OS:
App version:
Kubernetes distro being used:
Other:

### Steps to reproduce
1. Create a `zarf.yaml` that contains flavors
2. Run `zarf dev lint` without specifying a flavor
3. `zarf dev lint` will fail because a flavor is not specified

### Expected result
I would expect that `zarf dev lint` would be a schema check to ensure that `zarf.yaml` is well formed and free of any unrecognized keys, types, etc.  Flavor shouldn't be taken into account unless there is actually something invalid about how a `flavor` key is defined, such as yaml path or type, so you should not need to specify a flavor when linting.

### Actual Result
Linting fails

Image

### Additional Context
This issue is arising as our team attempts to run `zarf dev lint` in a CI process.  To fix this without changing `zarf dev lint` behavior, we would need to parse `zarf.yaml` to discover flavors before running `zarf dev lint --flavor=<flavor>`.  There are no default components for this zarf.yaml, which would presumably mask this problem, although I'm not sure if that means that only default components would be linted.

chaospuppy:
## Description
This PR adds an `--all-variants` flag to the `zarf dev lint` command that will result in all variants of a zarf package being linted, even if there are multiple flavors present in the package definition.  Previously, it was required that users specify a `--flavor` flag in order to pass validation.  This is done to make linting zarf package in CI or in situations where you would otherwise not want to/could not know which flavor you wanted to lint specifically, and wished to validate the entire package.

## Related Issue

Fixes #4598 

## Checklist before merging


src/pkg/packager/load/load.go
```diff
@@ -28,7 +28,9 @@ import (
 
 // DefinitionOptions are the optional parameters to load.PackageDefinition
 type DefinitionOptions struct {
-	Flavor       string
+	Flavor string
+	// All variants will ignore Flavor and will return all components, regardless of flavor
+	AllVariants  bool
```
AustinAbro321: Lets add the same validation as the CLI layer to ensure flavor and allVariants are mutually exclusive

src/pkg/packager/load/load.go
AustinAbro321, line 1: If you have a package without variants you get an error like this. Be sure to add a test case for using `--all variants` without variants in the package. 
```bash
2026-04-24 12:33:27 ERR error of type <nil> does not have method Unwrap(
```

src/internal/api/v1alpha1/validate.go
```diff
@@ -50,6 +50,15 @@ const (
...
+type ComponentNameNotUniqueErr struct {
+	component string
+}
+
+func (e *ComponentNameNotUniqueErr) Error() string {
+	return fmt.Sprintf(PkgValidateErrComponentNameNotUnique, e.component)
+}
+
 // ValidatePackage runs all validation checks on the package.
 func ValidatePackage(pkg v1alpha1.ZarfPackage) error {
```
AustinAbro321, line 55: I'd rather add `validateOpts` than the approach of comparing errors. It's simpler and the validate function is under `/internal`, so there's no breaking change to consumers.

AustinAbro321 (changes requested):
This should be a nice win for linting, thanks! A few comments

src/pkg/packager/load/load.go
chaospuppy: We shouldn't get this if we're no longer filtering errors based on type in favor of `validateOpts`, since that's emanating from the `FilterErr` utils function I added.  Still, I will make sure to add coverage for `--all-variants` being set on a package without flavors.

src/pkg/utils/error.go
```diff
@@ -0,0 +1,33 @@
...
+package utils
+
+import (
+	"errors"
+	"fmt"
+)
+
+// FilterErr error expects an error type and an error that implements the Unwrap() method
+// Any errors included in the wrapped error that match the error type T are removed from the final joined error
+func FilterErr[T error](joined error) (error, error) {
```
chaospuppy: I won't need this function since we're switching to `validateOpts`, should I remove it?  I don't know if you think it could be useful in the future or if filtering based on error type is an undesirable pattern.  If it is, I'd be curious to know more about why.

src/pkg/packager/load/load.go
chaospuppy: I'll add a check to PackageDefinitition to enforce exclusivity for now, but would you prefer a NewDefinitionOptions function to perform the validation everywhere we use DefintionOptions?

src/pkg/utils/error.go
AustinAbro321: Comes down to personal preference, but I generally prefer simplicity. The mental model of tracking and removing errors is harder to follow than stopping the errors from existing. 

And yeah any code that is unused shouldn't be in the repo. Easier to add later than figure out why it's there and if there are any consumers of it from the SDK.

src/pkg/packager/testdata/lint-with-imports/all-variants/zarf.yaml
```diff
@@ -0,0 +1,20 @@
+kind: ZarfPackageConfig
+metadata:
+  name: all-variants-lint
+
+components:
+  - name: duplicate
+    only:
+      flavor: first-flavor
```
AustinAbro321: Need to test importing a child components with flavors here

src/pkg/packager/load/import.go
```diff
@@ -259,9 +259,9 @@ func validateComponentCompose(c v1alpha1.ZarfComponent) error {
 	return errors.Join(errs...)
 }
 
-func compatibleComponent(c v1alpha1.ZarfComponent, arch, flavor string) bool {
+func compatibleComponent(c v1alpha1.ZarfComponent, arch, flavor string, allVariants bool) bool {
```
AustinAbro321: I believe we'll also want all variants of architectures.

src/internal/api/v1alpha1/validate.go
```diff
@@ -85,7 +90,7 @@ func ValidatePackage(pkg v1alpha1.ZarfPackage) error {
 	}
 	for _, component := range pkg.Components {
 		// ensure component name is unique
-		if _, ok := uniqueComponentNames[component.Name]; ok {
+		if _, ok := uniqueComponentNames[component.Name]; ok && !opts.SkipComponentNameUniquenessValidation {
```
AustinAbro321: This is too broad, let's change it so we only allow duplicate component names when the `only` blocks are not equal

AustinAbro321 (changes requested):

src/internal/api/v1alpha1/validate.go
chaospuppy: Since we can have any number of components with duplicate names when this validation option is true, I needed to map component names to all of the `only` blocks associated with those component names.  Let me know if I could simplify further, but I also didn't want to run `reflect.DeepEqual` unless we enabled `SkipComponentNameUniquenessValidation`.

src/pkg/packager/load/import.go
chaospuppy: I interpreted this to mean we want `allVariants` to b evaluated within `compatibleComponent` for both flavor and arch checks, but I also see how it might be better to leave the function unchanged and instead do something like `if !compatibleComponent && !allVariants`.

src/pkg/packager/load/load.go
AustinAbro321: No, adding it here were you did is correct

src/pkg/packager/load/load.go
```diff
@@ -86,30 +93,39 @@ func PackageDefinition(ctx context.Context, packagePath string, opts DefinitionO
...
+	if !hasFlavoredComponent(pkg, flavor) && !allVariants {
 		l.Warn("flavor not used in package", "flavor", flavor)
 	}
-	if err := internalv1alpha1.ValidatePackage(pkg); err != nil {
+
+	validationOpts := internalv1alpha1.ValidateOpts{}
+
+	if allVariants {
+		validationOpts.SkipComponentNameUniquenessValidation = true
+	}
```
AustinAbro321:
```suggestion
	validationOpts := internalv1alpha1.ValidateOpts{
		SkipComponentNameUniquenessValidation: allVariants,
	}
```

src/pkg/packager/load/import.go
AustinAbro321, line 160: We'll now need to handle multiple components with the same name being found. Add a test for this as well

src/internal/api/v1alpha1/validate.go
```diff
@@ -84,11 +91,21 @@ func ValidatePackage(pkg v1alpha1.ZarfPackage) error {
...
 	}
 	for _, component := range pkg.Components {
+		var duplicateOnly bool
 		// ensure component name is unique
 		if _, ok := uniqueComponentNames[component.Name]; ok {
-			err = errors.Join(err, fmt.Errorf(PkgValidateErrComponentNameNotUnique, component.Name))
+			// only check if only block is duplicated if we're skipping name uniqueness checks
+			if opts.SkipComponentNameUniquenessValidation {
+				duplicateOnly = slices.ContainsFunc(uniqueComponentNames[component.Name], func(o v1alpha1.ZarfComponentOnlyTarget) bool {
+					return reflect.DeepEqual(o, component.Only)
```
AustinAbro321: We generally avoid reflect, I'd do something like this instead
```go
func onlyTargetsEqual(a, b v1alpha1.ZarfComponentOnlyTarget) bool {
    return a.LocalOS == b.LocalOS && a.Flavor == b.Flavor &&
        a.Cluster.Architecture == b.Cluster.Architecture &&
        slices.Equal(a.Cluster.Distros, b.Cluster.Distros)
}
```

src/pkg/packager/load/load.go
```diff
@@ -28,7 +28,9 @@ import (
 
 // DefinitionOptions are the optional parameters to load.PackageDefinition
 type DefinitionOptions struct {
-	Flavor       string
+	Flavor string
+	// All variants will ignore Flavor and will return all components, regardless of flavor
```
AustinAbro321:
```suggestion
	// All variants will ignore Flavor & architecture and will return all components. Mutually exclusive with flavor.
```

AustinAbro321 (changes requested):
Apologies for the delay in review. A few comments

chaospuppy:
@AustinAbro321 sorry for the delay on my end here, I've addressed the feedback you provided above.  Setting `allVariants` will now result in multiple components of the same name but different `only` being imported.

src/pkg/packager/load/load.go
```diff
@@ -28,7 +28,9 @@ import (
 
 // DefinitionOptions are the optional parameters to load.PackageDefinition
 type DefinitionOptions struct {
-	Flavor       string
+	Flavor string
+	// All variants will ignore Flavor & architecture and will return all components. Mutually exclusive with flavor.
+	AllVariants  bool
```
AustinAbro321: There is similar work going on in #5020 src/pkg/packager/load/load.go. Instead of a boolean, let's copy over the variant Dimension construct, that way users can selectively add the filters that they want.

AustinAbro321 (commented):
I've not done a full review, but asking that this be in line with #5020

PR #4795, opened 2026-04-07
feat: expose sbom generation tools

a1994sc:
## Description

Exposes a handful of useful generic functions from `github.com/zarf-dev/zarf/src/pkg/packager/layout`

## Related Issue

Fixes #4794

## Checklist before merging


a1994sc:
A little unsure about the comments for the functions that I exposed, so if someone can double check them please and thank you.

src/pkg/packager/layout/assemble.go
```diff
@@ -844,7 +844,9 @@ func recordPackageMetadata(pkg v1alpha1.ZarfPackage, flavor string, registryOver
 	return pkg
 }
 
-func getChecksum(dirPath string) (string, string, error) {
+// GetChecksum takes a directory then creates a sha256 check sum for each files in the
```
brandtkeller: a small improvement might be explicitly mentioning the return values.

src/pkg/packager/layout/sbom.go
```diff
@@ -112,7 +112,8 @@ func generateSBOM(ctx context.Context, pkg v1alpha1.ZarfPackage, buildPath strin
 	return nil
 }
 
-func createImageSBOM(ctx context.Context, cachePath, outputPath string, img v1.Image, src string) ([]byte, error) {
+// CreateImageSBOM is used to generate a syft SBOM for a provided image reference.
```
brandtkeller: This function is a bit "leaky" with regards to how it operates - returning both the bytes _AND_ writing the file to the filesystem. Acceptable perhaps as a private function but supported in our public API's.... 

Nonetheless we should doccomment the behaviors.

brandtkeller (changes requested):
Slight changes to the doccomments. 

I have some reservations:

Things like the signatures. They were okay as non-exported functions but some of the behaviors as public API's are a little "leaky" IMO. 

I would also question the package origin. Do these really make quality exported functions for the `layout` package? 

Not inherently against making these public but do want to play devils advocate. Any additional surfaces we expose get locked in during v1.0.0 and create more surface areas for vulnerabilities. 

Tagging @zarf-dev/maintainers to weigh in as well.

a1994sc:
Going to close this for now and revisit post 1.0

PR #4823, opened 2026-04-13
feat: introduce conversions between API versions

AustinAbro321:
## Description

I had previously discussed the possibility of moving the deprecated types into their internal package, but I decided against this as I believe we want SDK users to be able to fully inspect the contents of their package. While variables are set for instance, we should let SDK users examine them.

relates to #3433

## Checklist before merging


src/api/convert/convert.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"github.com/zarf-dev/zarf/src/api/v1alpha1"
+	"github.com/zarf-dev/zarf/src/api/v1beta1"
+	internalv1alpha1 "github.com/zarf-dev/zarf/src/internal/api/v1alpha1"
+	internalv1beta1 "github.com/zarf-dev/zarf/src/internal/api/v1beta1"
+)
+
+// V1Alpha1PkgToV1Beta1 converts a v1alpha1 ZarfPackage to a v1beta1 Package.
+func V1Alpha1PkgToV1Beta1(pkg v1alpha1.ZarfPackage) v1beta1.Package {
```
soltysh: Maybe the function name should be `PackageV1Alpha1ToV1Beta1`, to clearly state the type and then the version conversion?

src/internal/api/v1beta1/convert.go
```diff
@@ -0,0 +1,776 @@
...
+		Documentation: g.Documentation,
+	}
+
+	if pkg.Kind == "" {
+		pkg.Kind = v1beta1.ZarfPackageConfig
+	}
+
+	// v1beta1 has no Kind ZarfInitConfig; collapse the v1alpha1 init kind into the normal package kind.
+	if string(pkg.Kind) == "ZarfInitConfig" {
+		pkg.Kind = v1beta1.ZarfPackageConfig
```
soltysh, line 313: Should we set `component.service` accordingly, not to use the information that a package was init?

soltysh: Similarly, I'd expect the inverse operation in the v1alpha1 conversion.

src/api/convert/convert_test.go
```diff
@@ -0,0 +1,1183 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/zarf-dev/zarf/src/api/v1alpha1"
+	"github.com/zarf-dev/zarf/src/api/v1beta1"
+	"github.com/zarf-dev/zarf/src/internal/api/types"
+)
+
+func TestV1Alpha1PkgToV1Beta1_Metadata(t *testing.T) {
```
soltysh, line 14: Manual testing are ok, but I'd imagine us writing something like a fuzzer-based testing which creates v1alpha1, then converts to v1beta1 and then back to v1alpha1. The tests is successful, when initial v1alpha1 matches the final one. Then one more, in the opposite direction.

This will:
1. Ensure that the conversions are not losing data.
2. Fuzzing will help strengthening the conversion to handle all allowed data.

soltysh: See https://github.com/kubernetes/kubernetes/blob/9e570c4124691bd0d765e780d0225655f99a905e/pkg/api/testing/serialization_test.go#L141-L150 for what this could look like. I'm happy to pick it up as a followup.

src/internal/api/v1beta1/convert.go
AustinAbro321: Good catch

src/api/convert/convert.go
AustinAbro321: Yeah I like that

src/api/convert/convert_test.go
AustinAbro321: Agree this makes sense as a follow up

src/internal/api/v1alpha1/convert.go
```diff
@@ -0,0 +1,871 @@
...
+
+	return g
+}
+
+func componentToGeneric(c v1alpha1.ZarfComponent) types.Component {
+	gc := types.Component{
+		Name:           c.Name,
+		Description:    c.Description,
+		Default:        c.Default,
+		Required:       c.Required,
```
soltysh: I guess you're missing:
```golang
		Optional:       !c.Required,
```

src/internal/api/v1alpha1/convert.go
```diff
@@ -0,0 +1,871 @@
...
+				ac.Version = ch.HelmRepository.Version
+			}
+		case ch.OCI != nil && ch.OCI.URL != "":
+			ac.URL = ch.OCI.URL
+			if ac.Version == "" {
+				ac.Version = ch.OCI.Version
+			}
+		case ch.Git != nil && ch.Git.URL != "":
+			gitURL := ch.Git.URL
+			if idx := strings.LastIndex(gitURL, "@"); idx > 0 {
```
soltysh: Do we support ssh-style based git url, like `git@host:repo.git` ?

src/api/v1beta1/convert.go
```diff
@@ -0,0 +1,124 @@
...
+	for i := range pkg.Components {
+		gc := g.Components[i]
+		pkg.Components[i].dataInjections = dataInjectionsFromGeneric(gc.DataInjections)
+		pkg.Components[i].group = gc.Group
+
+		for j := range pkg.Components[i].Charts {
+			gch := gc.Charts[j]
+			pkg.Components[i].Charts[j].version = gch.Version
+			pkg.Components[i].Charts[j].variables = chartVarsFromGeneric(gch.Variables)
+		}
```
soltysh: AI complains me here about missing `v1alpha1` fields copying (specifically about `TemplatedValuesFiles`, but I believe there might be more looking at `src/internal/api/types/package.go` and its `// v1alpha1-only flat source fields. Used during conversion to populate structured sources.` section: https://github.com/zarf-dev/zarf/pull/4823/changes#diff-18da185750076d92ff90961eb6b9d40d80fc7ea0438b146ee6189657e47202fdR179

src/internal/api/v1alpha1/roundtrip_test.go
```diff
@@ -0,0 +1,143 @@
...
+		Variables: []v1alpha1.InteractiveVariable{
+			{Variable: v1alpha1.Variable{Name: "VAR", Sensitive: true, Type: "raw"}, Description: "d", Default: "def", Prompt: true},
+		},
+		Values:        v1alpha1.ZarfValues{Files: []string{"vals.yaml"}, Schema: "schema.json"},
+		Documentation: map[string]string{"doc": "doc.md"},
+	}
+
+	roundTripped := ConvertFromGeneric(ConvertToGeneric(original))
+	require.Equal(t, original, roundTripped)
+}
```
soltysh, line 147: I assume this is the place where you're planning to introduce fuzzying, to ensure that we fill in all the fields automatically (through reflection), which should ensure that we catch all new fields being added, and conversion properly handled? Followup iirc, right?

src/internal/api/v1beta1/convert.go
```diff
@@ -0,0 +1,851 @@
...
+
+	return g
+}
+
+func componentToGeneric(c v1beta1.Component) types.Component {
+	gc := types.Component{
+		Name:         c.Name,
+		Description:  c.Description,
+		Optional:     c.Optional,
+		Service:      string(c.Service),
```
soltysh, line 79: Here, the opposite, so `Required: !c.Optional`, to ensure that the generic struct has all the fields filled in.

soltysh (commented):
A few comments/questions, the reflection-based round tripping tests  should help us uncover eventual missing bits. Those few I've pointed out are b/c AI told me 😅

src/internal/api/v1alpha1/convert.go
AustinAbro321: Yes, started using our transform library which natively handles this

src/api/v1beta1/convert.go
AustinAbro321: Yup we added templatedValuesFiles after this PR was initially opened and I hadn't included it. Looks like all the other fields are covered.

src/internal/api/v1alpha1/roundtrip_test.go
AustinAbro321: I decided to just add fuzzing in this PR. Was useful, caught a few other things

src/internal/api/v1alpha1/convert.go
AustinAbro321: Yes, when possible I've only included the newer, canonical field and let the generic API handle it, however in this case we need required so that we can do a successful round trip from v1alpha1->generic->v1alpha1 without losing that `required` is nil

src/internal/api/v1beta1/convert.go
AustinAbro321: I have not filled in every field in the generic struct, but rather let the conversion logic handle it so long as there is no information missing by filling out the newer, canonical field. In this case since required would always be nil, the conversion logic fallsback to using optional.

src/internal/api/v1alpha1/convert.go
```diff
@@ -0,0 +1,989 @@
...
+	gc := types.Component{
+		Name:              c.Name,
+		Description:       c.Description,
+		Default:           c.Default,
+		Optional:          !c.IsRequired(),
+		Required:          c.Required,
+		Group:             c.DeprecatedGroup,
+		DataInjections:    dataInjectionsToGeneric(c.DataInjections),
+		HealthChecks:      healthChecksToGeneric(c.HealthChecks),
+		DeprecatedScripts: scriptsToGeneric(c.DeprecatedScripts),
```
brandtkeller, line 87: Are we consuming `DeprecatedScripts` anywhere to perform the migration to actions?

src/internal/api/v1alpha1/convert.go
```diff
@@ -0,0 +1,989 @@
...
+		OnCreate: actionSetToGeneric(a.OnCreate),
+		OnDeploy: actionSetToGeneric(a.OnDeploy),
+		OnRemove: actionSetToGeneric(a.OnRemove),
+	}
+}
+
+func actionSetToGeneric(s v1alpha1.ZarfComponentActionSet) types.ComponentActionSet {
+	defaults := types.ComponentActionDefaults{
+		Silent:          s.Defaults.Mute,
+		MaxTotalSeconds: int32(s.Defaults.MaxTotalSeconds),
```
brandtkeller, line 229: seems negligible but do we need to add safe migrations from int to int32?

src/api/v1beta1/component.go
```diff
@@ -144,8 +125,6 @@ type Manifest struct {
 type Chart struct {
 	// The name of the chart within Zarf; note that this must be unique and does not need to be the same as the name in the chart repository.
 	Name string `json:"name"`
-	// The version of the chart. This field is removed from the schema, but kept as a backwards compatibility shim so v1alpha1 packages can be converted to v1beta1.
-	version string
 	// The Helm repository where the chart is stored.
 	HelmRepository *HelmRepositorySource `json:"helmRepository,omitempty"`
```
brandtkeller: comparing to the ZEP - should/could these sources utilize `oneof_required`?

brandtkeller (commented):
Adding some thoughts as I get time to test.

src/api/v1beta1/component.go
AustinAbro321: Good callout, added

src/internal/api/v1alpha1/convert.go
AustinAbro321: I'm not concerned about that honestly

src/internal/api/v1alpha1/convert.go
AustinAbro321: Yes in the pkgcfg package, It should always be migrated right as v1alpha1 is parsed, and so technically not be needed in the migration logic, but I think it's still better to have the converter convert everything.

src/cmd/dev_upgrade.go
```diff
@@ -0,0 +1,198 @@
...
+		return nil
+	}
+
+	switch {
+	case sourceVersion == v1alpha1.APIVersion && o.to == v1beta1.APIVersion:
+		var pkg v1alpha1.ZarfPackage
+		if err := goyaml.Unmarshal(b, &pkg); err != nil {
+			return err
+		}
+		if err := checkRemovedFields(pkg); err != nil {
```
brandtkeller, line 83: How strongly do we feel about this being an error return versus a process that outputs warnings? 

Considering the UX of the upgrade process - and that many packages may be migrating away from variables _at the same time_ - Is there any optionality to allow the upgrade to continue to produce the new schema package for use in iteration?

brandtkeller: I agree with the default behavior being an error - I wonder if there is an opt-in for lossy output.

AustinAbro321: Hmm good question, imo the possibility of a user (or agent) missing or ignoring the warning and trying to continue with a v1beta1 package without variables is a big UX risk. They would have their v1beta1 package config, but their manifests or charts wouldn't actually work, and they'd find this out at deploy time. 

I'd rather the progress be in discrete steps. 
- modify v1alpha1 package so it's compatiable with v1beta1
- switch from v1alpha1 to v1beta1

src/cmd/dev_upgrade.go
```diff
@@ -0,0 +1,198 @@
...
+			return fmt.Errorf("writing converted package: %w", err)
+		}
+	default:
+		return fmt.Errorf("unsupported conversion from %s to %s", sourceVersion, o.to)
+	}
+
+	return nil
+}
+
+func detectAPIVersion(b []byte) (string, error) {
```
brandtkeller, line 101: This makes a lot of sense - my brain wanted to reach for how we might start injecting this into other workflows (create, inspect, etc) - so that people begin seeing the v1alpha1 version on their packages - but there might be implications to this as well.

src/cmd/dev_upgrade.go
AustinAbro321: > I agree with the default behavior being an error - I wonder if there is an opt-in for lossy output.

I'm more open to this, I can see the argument that this could give users an idea of what the package will look like and a lower chance that this will lead users astray. I'm 50/50. If you'd like to see this I would add it.

brandtkeller: No change requested now. let's keep it on the radar as we feel out how users perceive the UX.

src/cmd/dev_upgrade.go
AustinAbro321: I'd be open to adding the api version to v1alpha1 packages, might be more natural to do this in #5048 alongside the other load work

src/api/v1beta1/component.go
```diff
@@ -368,8 +353,8 @@ const (
 type SetValue struct {
 	// Key represents which value to assign to.
 	Key string `json:"key,omitempty"`
-	// Type declares the kind of data being stored in the value.
-	Type SetValueType `json:"type,omitempty"`
+	// Type declares the kind of data being stored in the value. Defaults to "string" when omitted.
```
brandtkeller, line 356: Passing thought - not blocking - should this type have an option/capability to infer that same as `--set-values` does on the CLI?

AustinAbro321: Hmm it's interesting because we don't allow other types at all. I'd actually prefer not having an infer since we already have a convenient way to specify types, we'd just need to add int/float/whatever to the list

src/cmd/dev_upgrade.go
```diff
@@ -0,0 +1,198 @@
...
+type devUpgradeSchemaOptions struct {
+	to string
+}
+
+// This command will be unhidden once v1beta1 is ready for use
+func newDevUpgradeSchemaCommand() *cobra.Command {
+	o := &devUpgradeSchemaOptions{}
+
+	cmd := &cobra.Command{
+		Use:     "upgrade-schema [ DIRECTORY ]",
```
brandtkeller: Do we think users think of the package manifest as a schema? Given the existence of `zarf dev generate-schema` I wonder if intuitively this should just be `zarf dev upgrade` or similar?

src/cmd/dev_upgrade.go
```diff
@@ -0,0 +1,198 @@
...
+		RunE:    o.run,
+	}
+
+	cmd.Flags().StringVar(&o.to, "to", v1beta1.APIVersion, "Specify the API version to upgrade the package definition to.")
+
+	return cmd
+}
+
+func (o *devUpgradeSchemaOptions) run(cmd *cobra.Command, args []string) error {
+	dir := "."
```
brandtkeller: We get an ambiguous error if you mistakenly pass a built .tar.zst package to this command. use of `setBaseDirectory()` handles this.

src/cmd/dev_upgrade.go
```diff
@@ -0,0 +1,198 @@
...
+	if len(args) > 0 {
+		dir = args[0]
+	}
+
+	pkgPath, err := layout.ResolvePackagePath(dir)
+	if err != nil {
+		return err
+	}
+	b, err := os.ReadFile(pkgPath.ManifestFile)
```
brandtkeller, line 55: Unsure where best to put this - but how should the ergonomics around imports be handled and is there any preflight validation that can be done? This is pretty distinct for each schema version. 

Separately should we be providing a warning for running this on a `ZarfInitConfig`? It looks to be handling the upgrade gracefully with regards to to services (that is neat).

src/cmd/dev_upgrade.go
brandtkeller: nit: allows for specifying specific file

```suggestion
		Use:     "upgrade-schema [ PATH ]",
```

brandtkeller (changes requested):
Core conversion logic looks great. Few observations on the CLI surface.

src/cmd/dev_upgrade.go
AustinAbro321: Yeah good callout, added a log message when coming across a component import. I'm planning to have intuitive errors when resolving a v1beta1 package as well

AustinAbro321: Yeah given that init packages should work as well as any other package that is being upgraded, I don't think we need to any explicit warnings, though it will be worth spelling out in our documentation / guide

src/cmd/dev_upgrade.go
AustinAbro321: I do like zarf dev upgrade better yeah. The dev commands are assumed to run on a package definition

brandtkeller (approved):
lgtm

PR #1469, opened 2023-03-23
Feature: OCI Package Composability

Noxsios:
## Description

Allows for skeleton packages to be published and components imported via OCI:

```bash
$ zarf package publish examples/helm-local-chart oci://localhost:555 --insecure
```

```yaml
  - name: import-component-oci
    required: false
    import:
      # name is optional, if not provided the name of the component will be used to import
      name: demo-helm-local-chart
      url: oci://localhost:555/helm-local-chart:0.0.1
    actions:
      onDeploy:
        after:
          - wait:
              cluster:
                kind: deployment
                name: local-demo
                namespace: local-chart
                condition: available
```

This adds the following behavior to `zarf package publish`:

- `$version-skeleton`: If a directory is specified, the skeleton package will be published (and can be used for composability)
- `$version-$arch`: If a tarball is specified, that fully rendered package will be published

## Related Issue

Fixes https://github.com/defenseunicorns/zarf/issues/1097

Fixes https://github.com/defenseunicorns/zarf/issues/1060

Relates to #1426 

Fixes #1560 

Fixes #1215

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Other (security config, docs update, etc)

## Checklist before merging

- [x] Test, docs, adr added or updated as needed
- [x] [Contributor Guide Steps](https://github.com/defenseunicorns/zarf/blob/main/CONTRIBUTING.md#developer-workflow) followed

PR #4826, opened 2026-04-14
fix!: remove 0.27.0 layout shim

### LINKED ISSUE #4819: Use different user to create zarf package will have permission issue when deploy the package
The issue steps:
- In CICD, i use ubuntu os, `root` account to build zarf package `zarf-package-myapp-amd64-1.0.tar.zst`.   
Command is `zarf package create . -c`, it's output `zarf.yaml` just like this:
```yaml
- name: kubevirt-namespace
  required: true
  manifests:
  - name: kubevirt-namespace
    namespace: kubevirt
    files:
    - ../../../../../root/.zarf-cache/oci/dirs/03a3e106d7a419551e85e780e32fbd95023f264b562169f6f57ae3b62e20abcb/manifests/kubevirt-namespace-0.yaml
  actions:
    onRemove:
      after:
      - cmd: ./zarf tools kubectl delete ns kubevirt
```
- I install this package in my PC using root account, it's OK.
- Then, in another CICD, it use ubuntu os and `ubuntu` account to create upgrade package `zarf-package-myapp-amd64-1.0-differential-1.2.tar.zst`.  
Actually the  `kubevirt-namespace` not changed, keep same with first zarf package.  
Command is `zarf package create . --differential zarf-package-myapp-amd64-1.0.tar.zst -c`, it's output `zarf.yaml` like this:
```yaml
- name: kubevirt-namespace
  required: true
  manifests:
  - name: kubevirt-namespace
    namespace: kubevirt
    files:
    - ../../../../../home/ubuntu/.zarf-cache/oci/dirs/03a3e106d7a419551e85e780e32fbd95023f264b562169f6f57ae3b62e20abcb/manifests/kubevirt-namespace-0.yaml
  actions:
    onRemove:
      after:
      - cmd: ./zarf tools kubectl delete ns kubevirt
```
- Then when i download the upgrade zarf package, and install it using `dev` account with command `zarf package deploy zarf-package-myapp-amd64-1.0-differential-1.2.tar.zst -c`, it failed with error permission denied:
```bash
ERR failed to deploy package: unable to deploy component "kubevirt-namespace": unable to read manifest file /home/ubuntu/.zarf-cache/oci/dirs/03a3e106d7a419551e85e780e32fbd95023f264b562169f6f57ae3b62e20abcb/manifests/kubevirt-namespace-0.yaml: open /home/ubuntu/.zarf-cache/oci/dirs/03a3e106d7a419551e85e780e32fbd95023f264b562169f6f57ae3b62e20abcb/manifests/kubevirt-namespace-0.yaml: permission denied

[ERROR] Zarf package deployment failed.
```

Because the final user is `dev`, it's `zarf-cache` should be `/home/dev/.zarf-cache`, but it still used `/home/ubuntu` folder, and we don't have this account.

So i think this is the bug for `zarf package create` with `zarf package deploy` when using different account.

AustinAbro321:
## Description

#1469 changed how manifests were loaded into a package, but kept a shim around in case old packages were deployed on newer versions of Zarf. This shim can cause a bug under specific circumstances, see #4819. 

Some consideration for adding a minimumVersionRequirement here, but given that this package layout change happened in v0.27.0 from May 2023, I deemed this not practically necessary. 

The breaking change here is that we will no longer deploy packages from <v0.27.0. I doubt this is relevant to anyone. 

## Related Issue

Fixes #4819

## Checklist before merging

### DIFF
--- a/src/pkg/packager/deploy.go
+++ b/src/pkg/packager/deploy.go
@@ -628,12 +628,9 @@ func (d *deployer) installManifests(ctx context.Context, pkgLayout *layout.Packa
 	installedCharts := []state.InstalledChart{}
 	for _, manifest := range component.Manifests {
 		for idx := range manifest.Files {
+			manifest.Files[idx] = fmt.Sprintf("%s-%d.yaml", manifest.Name, idx)
 			if helpers.InvalidPath(filepath.Join(manifestDir, manifest.Files[idx])) {
-				// The path is likely invalid because of how we compose OCI components, add an index suffix to the filename
-				manifest.Files[idx] = fmt.Sprintf("%s-%d.yaml", manifest.Name, idx)
-				if helpers.InvalidPath(filepath.Join(manifestDir, manifest.Files[idx])) {
-					return installedCharts, fmt.Errorf("unable to find manifest file %s", manifest.Files[idx])
-				}
+				return installedCharts, fmt.Errorf("unable to find manifest file %s", manifest.Files[idx])
 			}
 			if manifest.IsTemplate() {
 				path := filepath.Join(manifestDir, manifest.Files[idx])


soltysh (commented):
Two things:
1. Can you add a unit-test to ensure we don't cause a regression.
2. I'm not sure I'm recreating the problem correctly, I've tried this with `zarf package create examples/yolo/ -c` the resulting `zarf.yaml` still points to the same files as https://github.com/zarf-dev/zarf/blob/bb4083d694ced66984ae0298cf8e2f7a5655dc3b/examples/yolo/zarf.yaml#L14-L15 But we package both of these files as `multi-games-0.yaml` and `multi-games-1.yaml` inside the `componentsyolo-games.tar` in `zarf-package-yolo-amd64.tar.zst`. So either I'm reproducing the problem the wrong way, or they fix is not correct.

AustinAbro321:
The problem is in the check that we currently have inside of Zarf skips permission errors with `helpers.InvalidPath` so we don't end up using the fallback. 

Here are reproduction steps:
- create a simple manifest with root permissions
```bash
sudo tee /root/secret-ns.yaml >/dev/null <<'EOF'                                                                                                                      
apiVersion: v1
kind: Namespace
metadata:
  name: demo
EOF
```
- Create a zarf.yaml to consume this manifest
```yaml
kind: ZarfPackageConfig
metadata:                                                                                                                                                             
  name: repro                                                                                                                                                       
  version: 1.0.0                                                                                                                                                      
components:                                                                                                                                                         
  - name: ns
    required: true
    manifests:                                                                                                                                                        
      - name: ns
        namespace: demo                                                                                                                                               
        files:                                                                                                                                                      
          - ../../../../../../../../../../root/secret-ns.yaml
```
- Create the package as root
- Deploy the package with latest Zarf, you'll get a permission error because we escape the filepath leaves the unarchived Zarf package and goes to the `/root/secret-ns.yaml` file on disk 
- Deploy with Zarf from this branch is successful because we removed the fallback

AustinAbro321:
> Can you add a unit-test to ensure we don't cause a regression

We don't currently have any unit tests for deploy.go. Stems partially from the fact that it's hard to unit test our calls to Helm. I feel confident in our e2e tests here, and I see this more as a feature removal (backwards compatibility shim removal) so this won't be a case that comes back. Thoughts on tabling this until we get a better grasp on a testing strategy for deploy?

soltysh (approved):
Confirmed, I was able to reproduce this w/o the fix and working with the fix. Although I will admit that the `zarf.yaml` contents inside the package, whether you manually extract it or use `zarf package inspect definition` could also be updated not to point to the source location, but rather package relative. But that's not blocking this PR. 

lgtm

PR #4827, opened 2026-04-15
feat: parse multi doc zarf.yaml files

AustinAbro321:
## Description

This implements parsing for multi doc Zarf packages, which will be used in built packages according to https://github.com/zarf-dev/proposals/issues/48. This also implements parsing for arbitrary API versions. Right now there is only v1alpha1, but when a new API version is added, it can be easily converted to the return type

Relates to #3433 

## Checklist before merging

### DIFF
--- a/src/internal/pkgcfg/pkgcfg.go
+++ b/src/internal/pkgcfg/pkgcfg.go
@@ -22,7 +22,7 @@ func Parse(ctx context.Context, b []byte) (v1alpha1.ZarfPackage, error) {
 		return v1alpha1.ZarfPackage{}, err
 	}
 	switch version {
-	case "", "zarf.dev/v1alpha1":
+	case "", v1alpha1.APIVersion:
 		return ParseV1Alpha1(ctx, b)
 	default:
 		return v1alpha1.ZarfPackage{}, fmt.Errorf("unknown apiVersion %q", version)
--- a/src/internal/pkgcfg/pkgcfg.go
+++ b/src/internal/pkgcfg/pkgcfg.go
@@ -31,8 +31,8 @@ var knownAPIVersions = []apiVersionHandler{
 	{version: v1alpha1.APIVersion, priority: 1, decode: decodeV1Alpha1},
 }
 
-// ParseDefinition parses a package definition
-func ParseDefinition(ctx context.Context, b []byte) (v1alpha1.ZarfPackage, error) {
+// Parse parses a package definition
+func Parse(ctx context.Context, b []byte) (v1alpha1.ZarfPackage, error) {
 	docs, err := parseZarfYAMLDocs(b)
 	if err != nil {
 		return v1alpha1.ZarfPackage{}, err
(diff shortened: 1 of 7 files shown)


src/internal/pkgcfg/pkgcfg.go
```diff
@@ -6,27 +6,156 @@ package pkgcfg
...
+	decode   func(ctx context.Context, node ast.Node) (v1alpha1.ZarfPackage, error)
+}
+
+// knownAPIVersions lists every apiVersion this Zarf version can decode. To add a
+// new version, append an entry with a higher priority than any existing one.
+var knownAPIVersions = []apiVersionHandler{
+	{version: v1alpha1.APIVersion, priority: 1, decode: decodeV1Alpha1},
+}
+
+// Parse parses a package definition
```
brandtkeller: nit: if someone is looking at the doccomments for `Parse` and `ParseMultiDoc` - do we have enough information here to help them understand the distinction of when to use each?

src/internal/pkgcfg/pkgcfg.go
```diff
@@ -6,27 +6,156 @@ package pkgcfg
...
+	seenVersions := map[string]bool{}
+
+	for i, doc := range docs {
+		version, err := apiVersionFromNode(doc.Body)
+		if err != nil {
+			return v1alpha1.ZarfPackage{}, fmt.Errorf("document %d: reading apiVersion: %w", i, err)
+		}
+		handler, known := handlerFor(version)
+		if !known {
+			l.Debug("found unsupported API version during parse", "API version", version)
```
brandtkeller: nit: standard go convention for slog keys (as specified but not enabled in our lint settings) is camelCase. 
```bash
golangci-lint run --enable sloglint
```

```suggestion
			l.Debug("found unsupported API version during parse", "apiVersion", version)
```

The UX is slightly different - humans might like the `API version` output but I'm hesitant to break something that we could start enforcing.

brandtkeller (changes requested):
minor nits otherwise tracing the logic currently in the debugger

src/internal/pkgcfg/pkgcfg.go
AustinAbro321: Added to the parse comment

src/internal/pkgcfg/pkgcfg.go
```diff
@@ -6,27 +6,156 @@ package pkgcfg
...
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
 	"github.com/zarf-dev/zarf/src/pkg/logger"
 )
 
-// Parse parses the yaml passed as a byte slice and applies schema migrations.
+// apiVersionHandler pairs a supported apiVersion with its decoder.
+type apiVersionHandler struct {
+	version  string
+	priority int
+	decode   func(ctx context.Context, node ast.Node) (v1alpha1.ZarfPackage, error)
```
brandtkeller, line 25: In the PR description you mention:
> Right now there is only v1alpha1, but when a new API version is added, it can be easily converted to the return type

And I want to validate my understanding-  as would be directly impacted here - would be that we treat latest as our canonical version correct? So when v1beta1 lands this is (and other decoder/parse/etc) return types are updated accordingly?

AustinAbro321: I think the ideal solution would be for decode to use the generic type in this PR https://github.com/zarf-dev/zarf/pull/4823/changes#diff-ce7753862b94966fcb901b6bff29cea161e9c35aba3a8c14f5b5ca11a7f59d28R1.

Then the public functions can call ConvertFromGeneric for whatever the return type is, and we need only update the type to return.

brandtkeller: Sounds aligned with the overall plan then. 👍

brandtkeller (approved):
lgtm - either purposefully or otherwise the errors for parsing yaml that isn't a zarf manifest begin to become more clear.

Issue #4289, opened 2025-10-18
Package Signing Enhancements - parent issue


brandtkeller:
## Description

This is a parent issue for consolidating feature enhancements for package signing operations. 

## Additional Context

This aligns implementation for the sub-issues to support a clear overview for work in progress and account for dependencies between issues.

brandtkeller:
```mermaid
flowchart TD
  N8_Sigstore_Bundles_0["8. Sigstore Bundles"]
  N1_Cosign_v3_x_1["1. Cosign v3.x"]
  N2_Package_Sign_2["2. Package Sign"]
  N4_Package_Verify_3["4. Package Verify"]
  N9_Keyless_4["9. Keyless"]
  N5_Deploy_opt_in_opt_out_5["5. Deploy opt-in / opt-out"]
  N7_Documentation_6["7. Documentation"]
  N6_State_Updates_7["6. State Updates"]
  N3_Build_metadata_8["3. Build metadata"]
  N10_Trusted_Roots_9["10. Trusted Roots"]
  N11_Verified_Resources_10["11. Verified Resources"]

  N8_Sigstore_Bundles_0 --> N9_Keyless_4
  N8_Sigstore_Bundles_0 --> N10_Trusted_Roots_9
  N1_Cosign_v3_x_1 --> N8_Sigstore_Bundles_0
  N2_Package_Sign_2 --> N9_Keyless_4
  N2_Package_Sign_2 --> N3_Build_metadata_8
  N4_Package_Verify_3 --> N5_Deploy_opt_in_opt_out_5
  N4_Package_Verify_3 --> N10_Trusted_Roots_9
  N9_Keyless_4 --> N10_Trusted_Roots_9
  N5_Deploy_opt_in_opt_out_5 --> N6_State_Updates_7
  N5_Deploy_opt_in_opt_out_5 --> N11_Verified_Resources_10

```

brandtkeller:
Objectives for phase 1 - #4290 , #4240 , #3959 are to establish explicit behaviors for zarf package sign and zarf package verify - modifying underlying processes where needed - while remaining backwards compatible. 

Objectives for phase 2 - Introduce breaking changes in the signature verification process ( #4048 ) and the format (Bundle)

brandtkeller:
Bundle format, keyless signing, and offline verification are included in the scope for [ZEP-0053](https://github.com/zarf-dev/proposals/pull/54)

brandtkeller:
[...]

PR #4829, opened 2026-04-16
feat(verfication): trusted root fetch command

### LINKED ISSUE #4570: Support natively retrieving a Trusted Root

Create a reproducible process to "create" a Trusted Root given the Sigstore public good instance. This should additionally support private sigstore infrastructure but that may not be initially tested. 

- **Given** connectivity to an instance of Sigstore
- **When** A command or process is invoked
- **Then** the current trusted root for that instance is written to a file locally

Create a `hack/` script to create the trusted root and do not expose this as a command to users.

### Additional context

This may be a command we want to hide initially - it should probably live under `zarf tools` but we will need to identify it's proper home. 

This will be used to reproducibly create the Trusted Root that is embedded in zarf.

brandtkeller:
## Description

This implements a `zarf tools trusted-root create` command to retrieve the Trusted Root via TUF (The Update Framework).

This mirrors cosign as opposed to embedding the cobra command directly as documented in the code. 

This will then allow us to reliably and reproducibly fetch the trusted root for either embedding in zarf AND/OR external users getting an update trusted root under the circumstances they cannot upgrade to a Zarf version containing the updated embedded certificate. 

More details can be found in #4289 - this trusted root will only be used for non-key signing verification. 

## Related Issue

Fixes #4570

## Checklist before merging


soltysh (approved):
lgtm

PR #4831, opened 2026-04-17
chore: add sloglint to golganci-lint configuration

AustinAbro321:
## Description

This adds sloglint to our golganci-lint configuration and fixes all non-compliant uses. 

There appears to be a bug in sloglint, so this may not catch everything for the time being https://github.com/go-simpler/sloglint/issues/132. 

## Checklist before merging

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -9,6 +9,7 @@ linters:
     - ineffassign
     - nolintlint
     - revive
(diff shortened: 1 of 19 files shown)


brandtkeller (approved):
beat me to it but I love it nonetheless - lgtm

PR #4832, opened 2026-04-17
feat: stop adding Zarf service default values to state when the service does not exist

### LINKED ISSUE #2740: The Zarf agent should not mutate git repos when there was no git repo specified on Zarf init
It doesn't make sense to mutate resources to the default Zarf git server if it wasn't created on Zarf init. 

- **Given** a git repo wasn't created or specified on Zarf init
- **When** an argo or flux git resource is created
- **Then** they are not mutated by the Zarf agent

### LINKED ISSUE #2947: Refactor Zarf state to accurately reflect existance of internal resources
By default when Zarf state is created it has the default internal values for all Zarf deployed servers and resources. This includes the registry, agent-tls, git-server, artifact-server. This leads to confusing code because it's impossible to tell if an internal resource exists and is used by looking at the Zarf state. It also makes the Zarf state inaccurate for anyone looking at the secret in the cluster. 

This leads to several issues in the codebase:
- Zarf ignores errors when connecting to an internal resource that doesn't exist. https://github.com/zarf-dev/zarf/blob/4b58792690f1c649391822ac1e87bc0b4d4d27b9/src/cmd/tools/zarf.go#L194
- Zarf has to check if a resource exists before it. https://github.com/zarf-dev/zarf/blob/4b58792690f1c649391822ac1e87bc0b4d4d27b9/src/cmd/tools/zarf.go#L162
- The Zarf agent preforms mutations as if the resource exists, this can be seen with argo #2740

AustinAbro321:
## Description

Zarf always adds default Git server values to state even when the Git server is not configured or deployed. This means anyone deploying GitOps resources has their resources configured to point at a non-existent Git server, guaranteeing failure. Additionally, we always add private-git-server secrets to every namespace even when their irrelevant.

## Related Issue

Fixes #2947
Fixes #2740

## Checklist before merging

### DIFF
--- a/src/internal/agent/hooks/argocd-application.go
+++ b/src/internal/agent/hooks/argocd-application.go
@@ -65,6 +65,10 @@ func mutateApplication(ctx context.Context, r *v1.AdmissionRequest, cluster *clu
 	if err != nil {
 		return nil, err
 	}
+	if !s.GitServer.IsConfigured() {
+		l.Debug("no Zarf git server configured, skipping ArgoCD Application mutation")
+		return &operations.Result{Allowed: true}, nil
+	}
 
 	app := Application{}
 	if err = json.Unmarshal(r.Object.Raw, &app); err != nil {
--- a/src/pkg/state/state.go
+++ b/src/pkg/state/state.go
@@ -40,16 +40,18 @@ const (
 )
 
 // AllServiceKeys is the canonical ordered list of every supported service key.
-var AllServiceKeys = []ServiceKey{RegistryKey, GitKey, ArtifactKey, AgentKey}
+func AllServiceKeys() []ServiceKey {
+	return []ServiceKey{RegistryKey, GitKey, ArtifactKey, AgentKey}
+}
 
 // ParseServiceKey returns the ServiceKey matching s, or an error if s is not recognized.
 func ParseServiceKey(s string) (ServiceKey, error) {
-	for _, k := range AllServiceKeys {
+	for _, k := range AllServiceKeys() {
 		if string(k) == s {
 			return k, nil
 		}
 	}
-	return "", fmt.Errorf("invalid service key %q, valid keys are: %v", s, AllServiceKeys)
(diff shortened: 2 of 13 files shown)


src/internal/agent/hooks/argocd-application.go
```diff
@@ -65,6 +65,10 @@ func mutateApplication(ctx context.Context, r *v1.AdmissionRequest, cluster *clu
 	if err != nil {
 		return nil, err
 	}
+	if !s.GitServer.IsConfigured() {
```
brandtkeller, line 68: This feels like the least amount of overhead but for sake of alternatives did we evaluate the complexity involved in configuring the webhook in any way? I imagine it would be more correct to not have the webhook configurations but possibly much more complex.

AustinAbro321: Yeah, decided against to forego complexity. Adding configuration to the chart is one level of complexity, but we'd also have to make sure that whenever we update state, for instance, during `zarf tools update-creds` that we re-deploy the agent with the proper values. Or if users wanted to update state themselves to zero out their Git server, they would have to update the agent.

src/pkg/state/state.go
```diff
@@ -28,17 +27,55 @@ const (
...
+}
+
+// Has reports whether k is in the set.
+func (s ServiceSet) Has(k ServiceKey) bool {
+	_, ok := s[k]
+	return ok
+}
+
+// Add inserts k into the set.
+func (s ServiceSet) Add(k ServiceKey) {
```
brandtkeller, line 65: If all callers go through `NewServiceSet` then this should be fine. Typically I'd guard against assigning to a nil map (validating as much) - but that shouldn't be a concern here.

brandtkeller (approved):
testing looks comprehensive and I'd lean on the other external tests we have for git to exercise any gaps around external optionality. 

Otherwise I like the ServiceSet abstraction and the current approach feels well grounded in the existing logic.

Issue #2477, opened 2024-05-06
E2E test happy path


phillebaba:
### Describe what should be investigated or refactored

The current e2e tests are large and require a certain amount of knowledge of Zarf and its code base to run. After spending some time getting the tests to run as a new Zarf user I have identified some friction points which could easily be addressed.

1. The dependencies documentation does not specify that Docker and buildx are required dependencies. While it could be assumed that most people will have Docker installed, they may not have buildx installed.

https://github.com/defenseunicorns/zarf/blob/0fa217351e6b75e4f5b18df6e3096797ee767b8c/site/src/content/docs/contribute/testing.mdx?plain=1#L13-L15

2. The tests assume, unless run in appliance mode that access to a Kubernetes cluster is properly configured. A check exists but it does not seem to do anything as it does not check the error. This results in errors much later, which wastes time and becomes harder to decipher. 

https://github.com/defenseunicorns/zarf/blob/0fa217351e6b75e4f5b18df6e3096797ee767b8c/src/test/common.go#L57-L62

3. It would be nice to have a minimal test run documented to make it easy to verify that the e2e tests can run, as it would take a long time for all the tests to run completely.

4. A lot of these tests seem to have little impact on each other, even those that are running against a cluster. While some test cases are marked to run in parallel some are not. This should probably be explored further to speed up test runs.

5. It could be good to move the test images from Docker Hub to GHCR to reduce the risk of rate limiting occurring when running the e2e locally multiple times in a day.

bm54cloud:
Please assign this to me 😄

brandtkeller:
To restate the intent and scope for this issue - there is a little bit of tribal knowledge currently on running the full test suite locally - to include the E2E tests. 

- What local tooling and dependencies are required
- What expectations for infrastructure exist (IE E2E testing assumes a cluster is available when not in appliance mode)
- What other gaps exist in the documentation

Points 4&5 are potential extensions of scope that we can explore further in follow-on issue(s). Some progress has been made here already and parallelizing tests is something we should change in isolation from other changes.

bm54cloud:
Started draft PR but since I'm only working on this about 4 hours per week, it will take me a while to get through all the docs and validate testing. Will remove draft status once ready for review. https://github.com/zarf-dev/zarf/pull/4834

PR #4834, opened 2026-04-17
docs: validate and update E2E test docs

bm54cloud:
## Description

This PR aims to:
- Update dependencies required in E2E testing docs
- Clarify Kubernetes cluster configuration needed for E2E testing
- Reduce tribal knowledge by validating the testing E2E testing docs are correct and identify/correct any knowledge gaps in the documentation

## Related Issue

Relates to #2477 

## Checklist before merging

### DIFF
--- a/site/src/content/docs/contribute/testing.mdx
+++ b/site/src/content/docs/contribute/testing.mdx
@@ -12,7 +12,9 @@ To run the end-to-end tests locally, you must meet the same prerequisites as tho
 
 1. GoLang version >= the version specified in the [go.mod](https://github.com/zarf-dev/zarf/blob/main/go.mod)
 2. Make
-3. Any clean K8s cluster (local or remote) or Linux with `root` if you want to use the Zarf-installed K3s cluster
+3. Docker
+4. Docker Buildx
+5. Any clean K8s cluster (local or remote) or Linux with `root` if you want to use the Zarf-installed K3s cluster
 
 ## Unit Tests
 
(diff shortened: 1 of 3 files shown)


bm54cloud:
@brandtkeller @AustinAbro321 Can I get some early feedback on the //Verify cluster connectivity section I added in 20_zarf_init_test.go. Originally was trying to work in main_test.go to verify cluster, but skipK8sEnvVar did not work as I expected it to. I left my original thought process in main_test.go commented out. This is what the error in 20_zarf_init_test.go would look like, so it would fail early. More so just wondering if putting it in 20_zarf_init_test.go is good or if you would truly prefer it in main_test.go.

Screenshot 2026-04-24 at 3 55 48 PM

brandtkeller:
> @brandtkeller @AustinAbro321 Can I get some early feedback on the //Verify cluster connectivity section I added in 20_zarf_init_test.go. Originally was trying to work in main_test.go to verify cluster, but skipK8sEnvVar did not work as I expected it to. I left my original thought process in main_test.go commented out. This is what the error in 20_zarf_init_test.go would look like, so it would fail early. More so just wondering if putting it in 20_zarf_init_test.go is good or if you would truly prefer it in main_test.go.

This is a great fallback but I would love to hear more about what limitations or constraints you ran into regarding `main_test.go` (this doesn't have to be the location/solution if we deem an appropriate solution is something along what you have currently - but I do want to capture notes).

I would point out that the current solution is limited to specific test commands. The out of the box `test-e2e` would likely run through all non-cluster e2e tests before failing and any filter of cluster e2e tests that run while _not_ using `20_` would also not be covered.

PR #4840, opened 2026-04-20
docs: add example of patching a helm chart

bburky:
Not sure if you're interested in this, or if this is the best place in the examples, but I thought I'd document this pattern we discovered that pretty cleanly can be used to carry patches on upstream Helm charts. This is useful when the upstream chart doesn't provide sufficient options in values overrides to make the required changes.

## Description

Adding an example of patching a helm chart

This is a simple method to apply modifications to a helm chart that is not under your control, if value overrides are not sufficient.

`*.patch` files are stored beside the `zarf.yaml` and applied in an `onCreate` action that clones the chart's git repository, applies the patches and then loads the patched chart from the local filesystem with `localPath` in the component definition.
## Checklist before merging


bburky:
Changes split into a new `demo-patched-helm-chart` component in the same Zarf package. This will prevent the example from showing at https://docs.zarf.dev/ref/components/#helm-charts

Instead, a description was added to `readme.md` which will be shown at https://docs.zarf.dev/ref/examples/helm-charts/.

Let me know if you want more detail in the README, but I don't think a full git tutorial there is appropriate? I included a link to online [`git format-patch`](https://git-scm.com/docs/git-format-patch) documentation.

bburky:
Also, not sure if I need to do anything to handle Windows? I saw one of the tests fail at one point. This should be POSIX-compatible `sh` which won't work on Windows (alternative Powershell code could do the same though).

examples/helm-charts/readme.md
```diff
@@ -1 +1,5 @@
...
+
+The `demo-patched-helm-chart` component shows how patches can be applied to a Helm chart. If possible, you should utilize Helm value overrides or consider upstreaming required modifications to Helm charts. But if these options are insufficient, you may need to make custom modifications to a Helm chart. Typically, you might fork the upstream Helm chart's git repository and periodically rebase or merge in new upstream changes.
+
+`demo-patched-helm-chart` demonstrates an alternative method using a [Zarf `onCreate` action](/ref/actions/), which performs a `git clone` of the upstream git repository and applies patch files with `git am`. This is conceptually equivalent to rebasing a set of changes onto the upstream repository, and is a useful technique to apply some minor changes to an upstream repository. If more complexity is required, or if you need to make changes to the patches frequently, maintaining a fork of the upstream repository may be more appropriate. To create patches, clone the upstream repository, commit your changes locally and use [`git format-patch`](https://git-scm.com/docs/git-format-patch) to generate patches.
```
bburky:
```suggestion
To customize Helm charts you should utilize Helm value overrides or consider upstreaming any required modifications to Helm charts. But if these options are insufficient, you may need to make custom modifications to a Helm chart. Typically, you might fork the upstream Helm chart's git repository and periodically rebase or merge in new upstream changes.

The `demo-patched-helm-chart` component an alternative method using a [Zarf `onCreate` action](/ref/actions/), which performs a `git clone` of the upstream git repository and applies patch files with `git am`. This is conceptually equivalent to rebasing a set of changes onto the upstream repository, and is a useful technique to apply some minor changes to an upstream repository. If more complexity is required, or if you need to make changes to the patches frequently, maintaining a fork of the upstream repository may be more appropriate. To create patches, clone the upstream repository, commit your changes locally and use [`git format-patch`](https://git-scm.com/docs/git-format-patch) to generate patches.
```

examples/helm-charts/zarf.yaml
```diff
@@ -55,10 +55,38 @@ components:
...
+        namespace: podinfo-from-git-patched
+        # In this case `localPath` will load the local podinfo chart which is created by the `onCreate` action
+        localPath: podinfo/charts/podinfo
+        valuesFiles:
+          - values.yaml
+    # git clone podinfo/ for podinfo-patched chart and apply patches (typically created with `git format-patch`)
+    # This is a simple method to apply modifications to a Helm chart that is not under your control, if value overrides are not sufficient
+    actions:
+      onCreate:
+        before:
```
brandtkeller, line 81: What are your thoughts on cleaning up the cloned repository? Could we have an after action for cleaning up? It removes some of the step-by-step subtlety but I also don't want an (even an ignored) directory remaining behind if possible.

brandtkeller (commented):
One observation otherwise this is a neat callout.

examples/helm-charts/zarf.yaml
bburky: If you get an error during the `onCreate` `before` action, I think the `after` action would be skipped. This action is expected to fail if the patches fail to apply cleanly (which may happen when bumping the git tag and the user would need to examine the repo and see what has changed).

[uds-k3d does an `rm` in their `onDeploy` `after`](https://github.com/defenseunicorns/uds-k3d/blob/v0.19.5/airgap/k3d/zarf.yaml#L48), I've seen it litter my current directory with files when deploy fails. This `rm` never gets run on error.

For `onCreate`, a gitignore is pretty easy and should work well. But I can add an `rm` in to `after` if you'd like.

Because the `after` can be skipped, it is not possible to remove the first `rm` command currently present in the `before`

brandtkeller: agree with the semantics around errors. I was thinking merely for testing that we have cleanup as that example is used in our testing. 

We are seeing the side-effect of that in the tests though - where we can't run a find images on the manifest as Zarf hasn't performed the clone yet and as such the local path doesn't exist for templating the chart....

examples/helm-charts/zarf.yaml
```diff
@@ -55,10 +55,38 @@ components:
...
+          - cmd: |
+              # Clone and apply patches to git repo
+              rm -rf podinfo
+              git clone -b 6.4.0 --depth 1 -c advice.detachedHead=false https://github.com/stefanprodan/podinfo.git
+              cd podinfo
+              export GIT_COMMITTER_NAME="zarf"
+              export GIT_COMMITTER_EMAIL="zarf@invalid"
+              git am --3way ../*.patch  # will error if the patch does not apply cleanly
+
```
bburky, line 90: Anyways, if you'd like an `rm` in after, you can apply this GitHub suggestion

```suggestion
        after:
           - cmd: rm -rf podinfo

```

I tested locally.

examples/helm-charts/zarf.yaml
```diff
@@ -55,10 +55,38 @@ components:
         releaseName: cool-release-name
         valuesFiles:
           - values.yaml
+
     images:
       - ghcr.io/stefanprodan/podinfo:6.4.0
       # This is the cosign signature for the podinfo image for image signature verification
       - ghcr.io/stefanprodan/podinfo:sha256-57a654ace69ec02ba8973093b6a786faa15640575fbf0dbb603db55aca2ccec8.sig
 
+  - name: demo-patched-helm-chart
```
brandtkeller, line 64: Existing tests that attempt to run `zarf dev find-images` on this package will now fail because the `localPath` doesn't exist until the action runs. 

We'll need to make a decision here around this being solely documentation or looking at the behaviors/testing for updates.

PR #4842, opened 2026-04-20
chore(development): harden the pre-commit configuration

### LINKED ISSUE #4841: pre-commit tooling drift and security updates

`.pre-commit-config.yaml` pulls and executes third-party code on every contributor's machine at `git commit` time. Two weaknesses in how it's configured are worth hardening. No known incident - this is a preventative pass.

#### Issues

1. **External hook repos are pinned by mutable tag** (`pre-commit/pre-commit-hooks@v5.0.0`, `sirosen/texthooks@0.6.4`). Git tags can be moved server-side, so a compromised maintainer account could retarget a tag and execute new code on the next `pre-commit autoupdate` or fresh hook install. We already pin GitHub Actions by commit SHA for this reason.

2. **Go toolchain is fetched from the internet at hook-install time.** The `goimports` and `lint` hooks declare `language: golang` with no `language_version`, so pre-commit downloads Go from `go.dev` into its own cache rather than using the Go on `PATH`. This introduces a second trust root for the Go toolchain and decouples lint/format behavior from the version pinned in `go.mod`.

#### Proposed changes

- Pin external hook repos by commit SHA with the tag preserved as a trailing comment.
- Set `language_version: system` on the `goimports` and `lint` hooks so the contributor's existing Go toolchain is used.
- Update `CONTRIBUTING.md` with prerequisites and a short rationale for both pins, so the intent is visible to future maintainers.

brandtkeller:
## Description

I noticed recently that the version of tooling used in my development (on my path) are not the same as the tooling downloaded and used for the pre-commit hooks. 

Additionally I noted that the hooks are not sha-pinned. 

Proposing that we outline tooling we require for development in our contributing docs and instead have pre-commit run using the tools managed by our development systems. 

Future updates might include development configuration for contributing to zarf with something like `mise`. 

## Related Issue

Fixes #4841 

## Checklist before merging

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
+    rev: cef0300fd0fc4d2a87a85fa2093c6b283ea36f4b  # v5.0.0
     hooks:
       - id: check-added-large-files
         args: ["--maxkb=1024"]
@@ -21,7 +21,7 @@ repos:
         args: [--markdown-linebreak-ext=md]
         exclude: site/src/content/docs/commands/.*
   - repo: https://github.com/sirosen/texthooks
-    rev: 0.6.4
+    rev: 72a68994497c827ced727b7c6b2567cfc5a96cb2  # 0.6.4
     hooks:
       - id: fix-smartquotes
   - repo: local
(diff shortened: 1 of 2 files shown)


a1994sc:
> Future updates might include development configuration for contributing to zarf with something like `mise`.

I'm a fan of `nix` personality as with tooling like "flakes" it is possible to create reproducible development environments.

a1994sc (approved):

AustinAbro321 (approved):

PR #4843, opened 2026-04-21
feat!: add multi arch support

AustinAbro321:
### Breaking changes

Breaking changes are only in the SDK. ImagesWithManifests->PulledImage. This removes manifests from the struct.  Instead of having to collect the manifest in several places (indexes, image archives, daemon) we check the manifest before we create the SBOM, this was our only use case for it  The name PulledImage should be more resistant to future breaking changes. `Unpack` and `Pull` return `PulledImage` now. 

## Description

This adds multi arch support thorough a special package architecture "multi"

We may want to change `zarf package pull` so that if a multi platform exists for a Zarf OCI package, when running `zarf package pull <package-oci-ref>` we fallback to pull multi. This PR is already large so I didn't change anything atm.

This adds a minimum version check for v0.76.0. Also worth noting that if a multi arch package is pulled from an earlier version of Zarf then Zarf will panic, since it assumed the index.json in an OCI layout will only have manifests. 

## Related Issue

Relates to #2425 (arguably fixes though we may want to keep it open or break it out into other issues such as packaging only x and y architecture).

## Checklist before merging


site/src/content/docs/ref/components.mdx
```diff
@@ -131,6 +131,10 @@ Kustomizations are handled a bit differently than normal manifests in that Zarf
 
 Images can either be discovered manually, or automatically by using [`zarf dev find-images`](/commands/zarf_dev_find-images/). The image list is not limited to containers, any OCI image following the [Image Manifest specification](https://github.com/opencontainers/image-spec/blob/main/manifest.md) can be pulled
 
+#### Multi-architecture images
+
+By default a Zarf package is built for one architecture and each image pulled matches that architecture. To build a single package that deploys on multi-arch clusters, set `metadata.architecture: multi`. Every image in a `multi` package must be pinned by digest. The digest should point to an OCI image index to pull multiple architectures, but can also point to a single-platform manifest.
+
```
a1994sc, line 137: So based off this comment it is currently not possible to say, `docker.io/library/nginx:1.29.6`, and have Zarf pull the various architectures?

src/api/v1alpha1/package.go
```diff
@@ -221,7 +224,7 @@ type ZarfMetadata struct {
 	// Disable compression of this package.
 	Uncompressed bool `json:"uncompressed,omitempty"`
 	// The target cluster architecture for this package.
-	Architecture string `json:"architecture,omitempty" jsonschema:"example=arm64,example=amd64"`
+	Architecture string `json:"architecture,omitempty" jsonschema:"example=arm64,example=amd64,example=multi"`
```
a1994sc: If it wasn't a breaking change, I would wonder if having this be an array of strings would be more ideal for multi-arch support:
e.i. if we had:

```yaml
metadata:
  architecture:
    - amd64
    - arm64
```

Would tell Zarf to pull both the amd64 and arm64 images if possible, maybe warn on images that do not have both.

a1994sc (commented):
Just a couple quick UX/package creator thoughts.

src/api/v1alpha1/package.go
a1994sc: I guess we could add `.metadata.architectures` to the package spec.

examples/wordpress/zarf.yaml
```diff
@@ -1,6 +1,7 @@
 kind: ZarfPackageConfig # ZarfPackageConfig is the package kind for most normal zarf packages
 metadata:
   name: wordpress       # specifies the name of our package and should be unique and unchanging through updates
+  architecture: multi   # a "multi" architecture package can pulls images by index sha, allowing use in multi-node clusters
```
brandtkeller:
```suggestion
  architecture: multi   # a "multi" architecture package pulls images by index sha, allowing use in multi-architecture clusters
```

site/src/content/docs/ref/components.mdx
brandtkeller: And to add to this question - is it a design decision or a limitation (or both).

src/pkg/images/pull.go
```diff
@@ -149,12 +150,13 @@ func Pull(ctx context.Context, imageList []transform.Image, destinationDirectory
 	}
 
 	l.Debug("gathering credentials from default Docker config file", "credentialsConfigured", credStore.IsAuthConfigured())
+	multiArch := opts.Arch == v1alpha1.MultiArch
```
brandtkeller: Noticing before now we never validate that `opts.Arch` is not an empty string. Digging a little this was not of consequence for a single platform but could be more multi-arch. 

Upstream code mitigates the problem but thinking about this from the exported function perspective.

brandtkeller (changes requested):
minor comments - review WIP

src/api/v1alpha1/package.go
AustinAbro321: Yeah, I had noted that as a future enhancement to https://github.com/zarf-dev/zarf/issues/2425, I think it'd be good follow up to have a comma separated list for now, and we can make schema changes in the future. I implemented the digest path first, so users always have the option of the more secure path (shasum).

site/src/content/docs/ref/components.mdx
AustinAbro321: Initially it was a design decision, but I changed my mind on this and decided to allow tags. This will pull every platform under that reference, and also log each platform.

src/api/v1alpha1/package.go
AustinAbro321: Hmm to your point, maybe we shouldn't introduce the concept of "multi" at all. We could add a boolean field in the v1beta1 schema such as "requireArchitectures" since images will become a struct. This could let users be explicit about which images they need, while introducing validation to prevent missing images. In this scenario index sha's would still work

AustinAbro321:
I've thought it over, and I'm planning on changing this PR to avoid introducing a the a "multi" architecture and instead  allow `architectures` as a list. When more than one architecture is introduced we will allow index sha's. 

I believe "Multi" will be vague for the package consumer, and there won't be a great way to validate if the package was built for  them. A user might ask "does this work arm64/v8 or just arm64? or maybe only amd64 and risc-v?". They wouldn't be able to validate without individually checking the upstream index sha's or unarchiving the package and examining the index sha. Even then, a package could coincidentally bring in all the images for arm64, when really risc-v support was intended and thus break on a future image addition.  Better to be intentional. 

Explicit architecture should allow for better validation behavior as well. We may want to validate that container images (those with the correct media type) have the listed architectures. 

This will make package naming a bit awkward with variants. I'm thinking we'll do something like `zarf-package-dos-games-amd64+arm64+arm64_v8-0.0.1.tar.zst` assuming the package was created with `-a arm64,arm64/v8,amd64`. Also since ocispec.Platform doesn't accept multiple architectures, we'll have to use a comma separated list, but this should be fine

src/api/v1alpha1/package.go
AustinAbro321: Yeah, added a comment on the PR, but I've decided we shouldn't introduce multi at all

Femure:
Thanks for putting this together. I’d like to better understand the direction you’re taking with this PR, and whether it’s intended to resolve the problem described in issue #2425.

The move toward explicit architecture lists makes sense from a clarity and validation standpoint. However, I want to make sure a specific use case is considered, as it’s currently a blocker for me and may not be covered by architecture selection alone.

In my workflow with FluxCD, images are pinned to an OCI index digest (not a platform-specific image). Because of that, the registry must contain the OCI index object itself, with the same digest.

From what I understand, this PR focuses on selecting and validating architectures, but:

* selecting a single platform breaks the digest
* pulling multiple architectures without preserving the index also breaks the digest

So even with this change, the index SHA wouldn’t exist in the air-gapped registry, and deployments would still fail.

The key requirement on my side is the ability to pull, package, and push an OCI index as-is, preserving its digest.

Can you confirm whether this PR is expected to support that, or if this would need to be handled separately?

AustinAbro321:
Good question @Femure. Index sha's will be supported for pulling and pushing index sha's when multiple architectures are selected. We definitely want to keep this case around, as it is the most secure route.

When a tag is used, then Zarf will pull only the architectures for that tag. When an index sha is used then Zarf will pull every platform to keep the index in sync

Femure:
Thanks for the clarification, that makes things much clearer.

Glad to hear that OCI index SHAs will be preserved when using them, that fully addresses my use case.

I’m looking forward to seeing this PR progress and hopefully get merged soon 🙂

AustinAbro321:
This PR has ballooned in size, and I feel it is hard to review. Going to break in into three parts: 
- introduce the test util and migrating unit tests over so they no longer call a registry on the internet 
- Introduce pulling and pushing indexes behind a feature flag
- Introduce defining multiple architectures and pulling those architectures by tag

AustinAbro321:
Image sha support was introduced in https://github.com/zarf-dev/zarf/pull/4879. I'm going to stop working on this for the time being to focus on getting Zarf to v1.0.

Adding in proper multi arch support raises a lot of questions, and probably needs a ZEP to design it. For instance:
- Would likely need to pass architecture through the system ? 
- Assuming a package has multiple architectures how should `.only.cluster.architecture` be handled ?
- How do we deal with variants? Do we go as far as to deal with OS features as well? 
  - Should each image get it's own list of platforms to pull?
- what should the platform be of the package in OCI?

PR #4848, opened 2026-04-24
fix(template): add to dissallowed functions

brandtkeller:
## Description

This PR removes the sprig function `getHostByName` as it creates additional avenues for unintentional network calls for which Zarf would want to protect against on all operations. 

## Related Issue

No related issue

## Checklist before merging


AustinAbro321 (dismissed):

AustinAbro321:
Not sure why test-site is only failing now, but we were using http:// instead of https:// which was causing it to not get filtered by our regex. We should be fine to change it to https://

AustinAbro321 (approved):

PR #4850, opened 2026-04-24
fix: ensure zarf say honors no-color

mjnagel:
## Description

Ensures that `zarf say` honors the `--no-color` flag.

## Related Issue

N/A

## Checklist before merging


brandtkeller (approved):
Given the scope and impact I don't have an issue accepting this. 

Given that message import - which checks whether color is enabled or disabled through pterm - I discovered pterm has a `RemoveColorFromString()` function. Using it or wrapping in a `message` function would also be acceptable.

AustinAbro321 (approved):

PR #4851, opened 2026-04-26
fix(init): skip namespace agent labels when agent is disabled

### LINKED ISSUE #4852: zarf init labels namespaces even when zarf-agent is not deployed
## Problem

When running `zarf init` with an init package that does not deploy `zarf-agent`, Zarf still labels existing namespaces with:

```yaml
zarf.dev/agent: ignore
```

rift0nix:
## Description

Updates `InitState` so existing namespaces are only labeled with `zarf.dev/agent=ignore` when the init package is actually deploying the Zarf agent.

Previously, `zarf init` always listed and updated existing namespaces during state initialization, even when the selected init components did not include `zarf-agent`. This caused registry-only init packages to make unnecessary cluster-wide namespace updates.

With this change, namespace labeling is gated by `InternalServices.Has(state.AgentKey)`, matching the existing source of truth used for agent TLS setup.

## Related Issue

Fixes #4852

## Testing

- [x] `go test ./src/pkg/cluster ./src/pkg/packager`

## Checklist before merging


src/pkg/cluster/cluster.go
```diff
@@ -246,27 +246,29 @@ func (c *Cluster) InitState(ctx context.Context, opts InitStateOptions) (*state.
...
 
-		namespaceList, err := c.Clientset.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
-		if err != nil {
-			return nil, fmt.Errorf("unable to get the Kubernetes namespaces: %w", err)
-		}
-		// Mark existing namespaces as ignored for the zarf agent to prevent mutating resources we don't own.
-		for _, namespace := range namespaceList.Items {
-			if namespace.Name == "zarf" {
-				continue
+		if opts.InternalServices.Has(state.AgentKey) {
```
AustinAbro321: nit: the block above has the same condition, you can merge these

src/pkg/cluster/cluster.go
AustinAbro321, line 1: Right now the codebase always assumes the agent is active. We'll need edits in a few places to make sure this change doesn't introduce future bugs. Rough outline: 

1. Introduce AgentIsConfigured() function 
```go
func (s *State) AgentIsConfigured() bool {
	return len(s.AgentTLS.Cert) > 0
}
``` 
2. when a secondary init occurs we should still initialize the agent if it's not yet configured and ignore the existing namespaces. See [this block](https://github.com/zarf-dev/zarf/blob/ea94cfeed259722292061ddba7d29db0f0c5171f/src/pkg/cluster/cluster.go#L307-L323). I recommend a private function such as `initAgent` to abstract the logic
```go
if opts.InternalServices.Has(state.AgentKey) && !s.AgentIsConfigured() {
  s.AgentTLS = initAgent(opts.AgentTls)
}
```
3. In [deploy.go](https://github.com/zarf-dev/zarf/blob/ea94cfeed259722292061ddba7d29db0f0c5171f/src/pkg/packager/deploy.go#L726) gate `pki.CheckForExpiredCert(ctx, s.AgentTLS)` to only run if the agent is configured
4. In [post-render.go](https://github.com/zarf-dev/zarf/blob/ea94cfeed259722292061ddba7d29db0f0c5171f/src/internal/packager/helm/post-render.go#L207-L211) only call `addAgentIgnoreLabels` if the agent is configured

AustinAbro321 (changes requested):
Thank you for this contribution! Increasing scope a bit to ensure this change will not introduce other bugs. Let me know if you have any questions

rift0nix:
Thanks for the review. I updated the PR to handle the optional agent state as required

rift0nix:
Looks like the missing agent ignore labels broke the deploy scenario with the agent enabled. I added a fix for that path and updated the unit tests based on the coverage report.

AustinAbro321:
@rift0nix rather than adding those conditions to deploy, I feel that it would make more sense to change [this block](https://github.com/zarf-dev/zarf/blob/c41c7b58f3aff4db09c3e885ee6b167555a58387/src/pkg/packager/dev.go#L125-L135) in dev.go to something like below. This way we use state when it exists in a dev deploy, and don't have to add any complexity to the deploy logic. 

```go
	if !opts.AirgapMode {
		// Helm charts may reference state values, so seed defaults first.
		d.s, err = state.Default()
		if err != nil {
			return err
		}

		requiresCluster := slices.ContainsFunc(pkg.Components, func(c v1alpha1.ZarfComponent) bool {
			return c.RequiresCluster()
		})
		if requiresCluster {
			d.c, err = cluster.NewWithWait(ctx)
			if err != nil {
				return err
			}
			clusterState, err := d.c.LoadState(ctx)
			if err != nil && !kerrors.IsNotFound(err) {
				return err
			}
			if clusterState != nil {
				d.s = clusterState
			}
		}

		if opts.RegistryURL != "" {
			d.s.RegistryInfo.Address = opts.RegistryURL
		}
	}
```

rift0nix:
+, moved the connected dev deploy state handling into dev.go as suggested

AustinAbro321 (approved):
LGTM, nice work

brandtkeller (approved):
lgtm - interested to see if there is other surface areas we can evaluate RBAC requirements around in order to isolate.

PR #4857, opened 2026-04-27
fix(create): built package path separators

### LINKED ISSUE #3275: Zarf package output file path separators
### Environment
Device and OS: Linux and Windows
App version: v0.43.0
Other: N/A

### Steps to reproduce

1. Build Zarf package on a Linux machine.
2. Build Zarf package on a Windows machine.
3. Compare zarf.yaml in the two different Zarf packages.

### Expected result

I expect the content apart from the build field to be identical.

### Actual Result

File paths in the zarf.yaml differ due to different file separators. These should be the same.

N/A

### Severity/Priority

Low

### Additional Context

This does not have any functional impact on Zarf as we currently normalize the separators in runtime,

brandtkeller:
## Description

I bumped this issue in priority after realizing the implications. 

Built Zarf packages should be byte-identical across operating systems, but filepath.Join was producing OS-native separators in path strings that get serialized into the published zarf package.

This really only impacts import/skeletons and how paths are resolved to my knowledge. Worth doing now as a pattern before we introduce any new import/component behaviors. 

This statically defines checksums in testing which can be a burden during development when the package shape changes but a worthwhile tradeoff for reproducibility. 

## Related Issue

Fixes #3275 

Relates to #

## Checklist before merging

### DIFF
--- a/src/test/testutil/paths.go
+++ b/src/test/testutil/paths.go
@@ -1,0 +1,82 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+package testutil
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"strconv"
+	"strings"
+	"testing"
+
+	goyaml "github.com/goccy/go-yaml"
+	"github.com/stretchr/testify/require"
+	"github.com/zarf-dev/zarf/src/api/v1alpha1"
+)
+
+// ChecksumZarfYAMLContent returns sha256(marshal(pkg)) with pkg.Build zeroed.
+// Build carries intrinsically non-reproducible values (timestamp, user,
+// hostname, CLI version) so it is excluded; the result pins everything else
+// and must be stable across build hosts.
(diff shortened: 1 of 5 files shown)


src/test/testutil/paths.go
```diff
@@ -0,0 +1,82 @@
...
+				check(prefix+".manifests["+strconv.Itoa(mi)+"].files["+strconv.Itoa(fi)+"]", f)
+			}
+			for ki, k := range m.Kustomizations {
+				check(prefix+".manifests["+strconv.Itoa(mi)+"].kustomizations["+strconv.Itoa(ki)+"]", k)
+			}
+		}
+		for di, d := range comp.DataInjections {
+			check(prefix+".dataInjections["+strconv.Itoa(di)+"].source", d.Source)
+		}
+		actionsByName := map[string][]v1alpha1.ZarfComponentAction{
```
AustinAbro321, line 67: I wasn't sure if this would cause a bug since windows users still use these paths, but it looks like windows is okay with a forward slash as a separator

AustinAbro321 (approved):
LGTM

PR #4862, opened 2026-04-30
ci: speed up windows tests

AustinAbro321:
## Description

It's common for windows unit tests to take 40+ minutes https://github.com/zarf-dev/zarf/actions/runs/25177060585/job/73811938191. Often 20+ minutes waiting on the post-setup step, I believe because of the cache. I'm hoping this will speed it up. 

From this PR, unit tests took 11 minutes, and there was no 20 minute shutdown time. If we move to a vendor system this will get even faster

## Checklist before merging


brandtkeller (approved):
Changes look correct and appropriate. There were some matching issues regarding similar problems reported upstream but I believe this is more nuanced and sufficient for our workflows. lgtm.

PR #4865, opened 2026-05-01
experimental: v1beta1 init package example

AustinAbro321:
## Description

This adds in a draft of what the v1beta1 package would look like given the proposal https://github.com/zarf-dev/proposals/pull/52 and with values instead of variables. This is not meant to merged, but to help evaluate the proposal. 

Some notes: 
- This has convinced me that each componentConfig should be able to define its own values schema, otherwise if the schema only lives in the root chart, then skeleton packages won't inherit it. 
- Skeleton or remote component don't have images so users have to define images themselves. 

## Checklist before merging


AustinAbro321:
We've made the decision for v1beta1 that this helped foster. Closing it out!

AustinAbro321:
Reopenign this so we can use it to evaluate https://github.com/zarf-dev/proposals/pull/64

PR #4866, opened 2026-05-01
fix(variables): file variable substitution alignment

### LINKED ISSUE #4863: ZARF_VAR template as sole entry under an empty mapping block fails on deploy but renders correctly in zarf dev inspect manifests
### Environment
Device and OS: macos
App version: Zarf 0.75.0
Kubernetes distro being used: v1.34.4+k3s1
Other: N/A

### Steps to reproduce
1. Create the following `zarf.yaml`
```yaml
kind: ZarfPackageConfig
metadata:
  name: test
  version: test
variables:
  - name: FOO
    default: |
      foo: bar
components:
  - name: test
    required: true
    manifests:
      - name: test
        files:
          - configmap.yaml
```
2. Create the following `configmap.yaml`
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  #app: test # <-- if you uncomment this, it will work
  ###ZARF_VAR_FOO###
```
3. Run `zarf package create . --confirm`
4. Run `zarf package deploy zarf-package-test-*.tar.zst --confirm`

### Expected result

The ConfigMap deploys successfully with `data.foo: bar`, consistent with what `zarf dev inspect manifests` renders.

### Actual Result

Deployment fails with an error similar to:

```
error=failed to create typed patch object (default/test; /v1, Kind=ConfigMap): .foo: field not declared in schema
```

`zarf dev inspect manifests` shows the manifest rendering correctly before deploy. The error only surfaces during `zarf package deploy`.

Image

*Workaround:* Adding any sibling field alongside `###ZARF_VAR_FOO###` under `data:` resolves the issue. The bug is isolated to the case where a mapping block contains only a `ZARF_VAR` template and no other keys.

Image

### Severity/Priority

Has a workaround, but creates a misleading inconsistency between zarf dev inspect manifests output and actual deploy behavior.

### Additional Context

- Reproduces with `autoIndent: true` and `autoIndent: false`
- Reproduces across multiple mapping fields — confirmed with `data:`, `annotations:`, and `labels:` blocks
- The variable renders correctly when inspected, suggesting the issue lies in how the templated YAML is handled during the typed patch apply step rather than during template rendering itself

brandtkeller:
## Description

The bug highlights an issue regarding the lifecycle of variable replacement. Inspect performs replacement instead of relying upon the helm wrapper postrender logic. There was a change in upstream postrender/kyaml workflows that strippedi indentation based on an empty field as a result of the content of the original resource only have "comments". 

This would propose templating files before the helm wrapper logic and reverses the order of operations such that variable replacement happens before go templating - the order used consistently for templating operations. 

Need to consider any edge cases or permutations of components that may be impacted by this change in sequence.

## Related Issue

Fixes #4863

## Checklist before merging


.github/workflows/scan-lint.yml
```diff
@@ -14,3 +14,5 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Run golangci-lint
         uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
+        with:
+          version: v2.12.0
```
brandtkeller, line 18: I had an inconsistency between local version and CI - added versioning pin here given current work to standardize.

AustinAbro321 (approved):

PR #4867, opened 2026-05-01
feat(values): package inspect values feature support

### LINKED ISSUE #4222: Add values and templating support for package remove actions and inspect

brandtkeller:
## Description

Adds support for `zarf package inspect manifests|values-files`  values templating.

Note: this does not include imports or skeletons as of yet - that will be covered by #4427 

This also includes a minor fix for detectign built packages passed to commands that expect a definition path. `zarf dev inspect <tarball>` gives confusing output. 

## Related Issue

Fixes #4222

## Checklist before merging


src/cmd/cmd.go
```diff
@@ -6,19 +6,33 @@ package cmd
...
 )
 
-// setBaseDirectory sets the base directory. This is a directory with a zarf.yaml.
-func setBaseDirectory(args []string) string {
+// setBaseDirectory returns the path to a package definition (a directory containing zarf.yaml,
+// or a path to a definition file). args[0] is used if provided, otherwise ".".
+//
+// Built-package artifacts (.tar.zst, .tar, .part000*) are rejected with a redirect to the
+// `zarf package` subcommands — passing a built package to a definition-side command would
+// otherwise feed a binary blob to the YAML parser.
```
AustinAbro321: Just my personal opinion, but I think comments this verbose are not valuable especially on private short functions. I've started instructed my AI tooling to have brief comments that avoid repeating the implementation details.

src/cmd/package.go
```diff
@@ -330,8 +332,12 @@ func (o *packageDeployOptions) run(cmd *cobra.Command, args []string) (err error
...
 		strings.ToUpper,
 	)
-	// Merge values
-	maps.Copy(o.setValues, v.GetStringMapString(VPkgDeploySetValues))
+	// Merge values; CLI --set-values overrides viper config, matching --set-variables.
+	o.setValues = helpers.TransformAndMergeMap(
+		v.GetStringMapString(VPkgDeploySetValues),
+		o.setValues,
+		func(s string) string { return s },
+	)
```
AustinAbro321: Maps.copy was using the wrong precedence, but I would rather avoid using `helpers.TransformAndMergeMap` when we aren't doing a transform. Could we introduce a function in src/cmd/common.go `mergeMap()`

src/pkg/packager/inspect.go
```diff
@@ -59,11 +62,30 @@ func InspectPackageResources(ctx context.Context, pkgLayout *layout.PackageLayou
...
+	valuesPath := filepath.Join(pkgLayout.DirPath(), layout.ValuesYAML)
+	vals, err := value.ParseLocalFile(ctx, valuesPath)
+	if err != nil {
+		return nil, err
+	}
+	vals.DeepMerge(opts.Values)
+
+	if pkgLayout.Pkg.Values.Schema != "" {
+		schemaPath := filepath.Join(pkgLayout.DirPath(), layout.ValuesSchema)
+		if err := vals.Validate(ctx, schemaPath, value.ValidateOptions{SkipRequired: true}); err != nil {
```
AustinAbro321, line 84: Want to call out that we don't do validation in the `dev inspect manifests|values-files` path. Up to you whether to solve that in this PR or a separate PR

AustinAbro321 (changes requested):
Overall looks good, minor suggestions

src/pkg/packager/inspect.go
brandtkeller: There is certainly a gap here. The schema IS validated against the package values files declared but not against any additional inputs from CLI/Viper which does feel like a gap.  I'll add the same validation as I believe inspect should be surfacing any errors.

src/cmd/cmd.go
brandtkeller: Fair point.

src/cmd/package.go
brandtkeller: That seems reasonable to me.

AustinAbro321 (approved):

PR #4876, opened 2026-05-05
test: improve OCI unit testing

AustinAbro321:
## Description

This is work that I've broken out of #4843. This introduces a test fixture for unit testing OCI objects so our tests don't need to rely on the upstream registry

## Checklist before merging


src/pkg/zoci/pull_test.go
```diff
@@ -82,44 +81,129 @@ func TestAllLayersRespectsRequestedComponents(t *testing.T) {
...
+			manifest: manifestDesc,
+		},
+	}
+}
+
 func TestAssembleLayers(t *testing.T) {
 	ctx := testutil.TestContext(t)
+	pkg := buildVirtualPackage(ctx, t)
 
-	remote, pkgLayout := publishAndConnect(ctx, t, "testdata/basic")
```
brandtkeller, line 88: Is this testdata still used or can we remove the test fixtures now?

src/test/testutil/ociregistry.go
```diff
@@ -0,0 +1,149 @@
...
+
+// PushMultiArchIndex pushes a flat multi-arch OCI image index with one single-platform manifest
+// per entry in platforms. Returns the index digest.
+func PushMultiArchIndex(ctx context.Context, t *testing.T, repoRef, tag string, platforms []ocispec.Platform) string {
+	t.Helper()
+	repo := NewRepo(t, repoRef)
+	children := make([]ocispec.Descriptor, 0, len(platforms))
+	for _, platform := range platforms {
+		desc := PushSinglePlatformImage(ctx, t, repo, platform.Architecture)
+		p := platform
```
brandtkeller: nit: we don't need the loop variable per-iteration assignment anymore correct?

src/pkg/zoci/pull_test.go
```diff
@@ -82,44 +81,129 @@ func TestAllLayersRespectsRequestedComponents(t *testing.T) {
...
+		include     []zoci.LayerType
+		expectedLen int
 	}{
 		{
 			name:        "all layers (default)",
 			include:     nil,
-			expectedLen: 10,
-		},
-		{
-			name:        "sbom layers",
```
brandtkeller, line 113: Noting we lost the sbom layers coverage here - likely as a limitation of the virtual package. We still have indirect coverage elsewhere but worth thinking about.

brandtkeller (changes requested):
Minor go backs but it also prompted a thought; I wonder if we need v1 documentation regarding exported APIs that are not included under the project umbrella of supported API's. We should be able to make breaking changes to API's under testing utilities?

src/pkg/zoci/pull_test.go
AustinAbro321: Yeah, I considered this, but I believe it's a worthwhile change for the unit tests to no longer have to reach out to the registry

src/pkg/zoci/pull_test.go
AustinAbro321: No longer in use, remove and forgot, thank you

src/test/testutil/ociregistry.go
AustinAbro321: Correct, fixed a few other spots as well

AustinAbro321:
I would consider test APIs outside of the stable API promise

brandtkeller (approved):
lgtm

PR #4879, opened 2026-05-07
feat!: allow pulling images by index sha

AustinAbro321:
## Breaking Changes
Breaking changes are limited to the SDK in `src/pkg/images`
  - ImageWithManifest type removed → replaced by PulledImage. Pull and UnPack now return this type.
    - Old: struct { Image transform.Image; Manifest ocispec.Manifest }
    - New: struct { Image transform.Image } — the Manifest field is gone. SDK users who relied on the returned manifest must fetch it themselves from the OCI store. 
  - PushOptions.Arch field removed. I believe this was never actually needed, we can just push the architectures that the layout includes. 

## Description

This implements pulling image indexes. It adds a version requirements so that Zarf can correctly pull the sha'd images. It also changes the deploy check to allow deploying to any node when an index sha image exists in the package. 

When there are multiple container images pointed to an index, Zarf will create an SBOM for each image. 

Pulling Indexes place a minimumVersionRequirement of v0.76.0. Also due to our pull logic not properly handling indexes, Zarf will panic if it tries to pull a package with indexes before this change.

I considered adding a metadata field, or requiring `.metadata.architecture: multi` for index pulls, but I think that will cause more confusion than improvements to the UX. We don't block users from pulling sha's of different digests. I'm hoping the log of the platform during pull will be enough of an indicator to those using index sha's mistakenly. 

## Related Issue

Relates to https://github.com/zarf-dev/zarf/issues/2425. I'll probably close this is merged, but I want to comment on how this doesn't implement tags or specific architectures and ask the issue thread to make an issue if this is a feature they desire.

## Checklist before merging

- [ ]


src/pkg/images/common.go
```diff
@@ -214,12 +217,80 @@ func WithPushAuth(ri state.RegistryInfo) crane.Option {
...
+	}
+	childSize, platforms, err := sumManifestsSize(ctx, fetcher, idx.Manifests)
+	if err != nil {
+		return 0, nil, err
+	}
+	return indexDesc.Size + childSize, platforms, nil
+}
+
+// sumManifestsSize walks each descriptor (recursing into nested indexes) and totals up the byte
+// size of every referenced blob plus one "arch[/variant]" string per leaf manifest.
```
brandtkeller: More for my education looking at the OCI spec -

1. do we need to consider OS? Variety of implications here
2. seems unlikely but will there ever be a case of a shared layer?

AustinAbro321: 1. For now no, we hardcode `os: linux` when pulling packages. 
2. It's technically possible, this function should have had unit tests anyhow. Added logic to handle this

src/test/e2e/48_multi_platform_image_test.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, entry := range sbomEntries {
+		name := entry.Name()
+		if strings.HasSuffix(name, ".json") && strings.Contains(name, "podinfo_6.4.0") {
+			count++
+		}
+	}
+	require.GreaterOrEqual(t, count, 2, "expected per-platform SBOMs for the digested multi-platform image")
+
+	stdOut, stdErr, err = e2e.Zarf(t, "package", "deploy", pulledPkgPath, "--confirm", "--skip-version-check")
+	require.NoError(t, err, stdOut, stdErr)
```
brandtkeller, line 74: Do we need fidelity in the state of the pod in the cluster to confirm any deployment behaviors or are we confident in it if the deployment occurs successfully?

AustinAbro321: I'm confident if the deploy occurs successfully.

src/pkg/packager/layout/assemble.go
```diff
@@ -819,17 +824,15 @@ func recordPackageMetadata(pkg v1alpha1.ZarfPackage, flavor string, registryOver
...
 
-	var versionRequirements []v1alpha1.VersionRequirement
-	for _, comp := range pkg.Components {
-		if len(comp.ImageArchives) > 0 {
-			versionRequirements = append(versionRequirements, v1alpha1.VersionRequirement{
-				Version: "v0.68.0",
-				Reason:  "This package contains image archives which will only be recognized on v0.68.0+",
-			})
-			break
+	hasIndex := false
```
brandtkeller, line 827: Nit: nonblocking - the split here between `buildPath`, `hasIndex` and `collectVersionRequirements` is entirely functional but feels a little awkward. in so much as I think `buildPath` could be a parameter and then `imageLayoutasIndex` could be internal to `collectVersionRequirements` ?

src/pkg/packager/deploy.go
```diff
@@ -761,12 +760,21 @@ func setupState(ctx context.Context, c *cluster.Cluster, connected bool) (*state
...
 	if !pkg.HasImages() {
 		return nil
 	}
 
+	hasImageIndex, err := pkgLayout.HasImageIndex()
+	if err != nil {
+		return fmt.Errorf("failed to inspect package image layout: %w", err)
+	}
+	if hasImageIndex {
+		return nil
```
brandtkeller, line 775: ++ short-circuit validation. mental note for you mention of potentially checking for the desired target arch if we deemed necessary (not required).

brandtkeller (approved):
thoughtfully implemented - no blocking comments but a few minor thoughts or observations. lgtm

src/pkg/packager/layout/assemble.go
AustinAbro321: The reason I did it this way was so I could unit test `collectVersionRequirements` without a image archive. Also `imageLayoutHasIndex` is used `packageLayout.HasImageIndex()`

PR #4880, opened 2026-05-07
feat(sign)!: align signing and verification flags to cosign

### LINKED ISSUE #4572: Expose support for additional signing and verification options

Zarf only supports asymmetrical key pairs as of this issue for signing and verification. We should expand support for signing and verification to include other workflows supported by Cosign.

- **Given** a package is present on the filesystem
- **When** signing and verification occur
- **Then** `zarf package sign|verify` have the required flags to support alternative workflows

Only retain the asymmetrical key pair support - this creates a system reliant on public key distribution that may not entirely be aligned with the zarf portability requirements. 

### Additional context

Zarf should evaluate the additional inputs exposed on create/deploy/publish workflows to align or streamline the processes.

brandtkeller:
## Description

Updates the SDK logic to align to cosign in the use of `Sign|VerifyBlobOptions`. These will be made available to CLI users of `Sign|Verify` in an opt-in basis of what the code supports. 

These wire up the flags to the options but actual use of them for packages is still primarily isolated to asymmetrical keys. A follow on PR is underway to provide the wiring in support of additional signing/verify workflows and an embedded trusted root. 

**NOTE:** this is breaking is as much as the sign/verify options now embed cosigns options structs directly. The primary-use `KeyRef` and `SigRef` are being deprecated in favor of eliminating aliases where possible. Originally the intent was to map flags manually before I realized the upstream methods existed. I think it is reasonable to accept changing the embedded options in the cosign `util` package options in favor of alignment.  

## Related Issue

Fixes #4572

Relates to #

## Checklist before merging


src/cmd/package.go
```diff
@@ -1872,15 +1852,55 @@ func newPackageSignCommand(v *viper.Viper) *cobra.Command {
...
+	mergeCosignSignFlags(cmd, &o.cosign)
+
 	return cmd
 }
 
+// mergeCosignSignFlags binds cosign's SignBlobOptions onto a throwaway command,
+// hides flags whose support is deferred or are deprecated upstream, applies
+// zarf's air-gap default overrides, and folds the result into cmd via AddFlagSet.
+// AddFlagSet skips flag names already registered, so zarf's pre-existing --key
+// and --output retain their semantics.
```
brandtkeller: A little verbose but explains the reasoning.

AustinAbro321:
While I agree with mirroring the the cosign options on the SDK side, I would prefer not to mirror the commands 1:1. While it is extra steps for us to copy the command, I'd rather have us explicitly opt into any flag we expose on the CLI side, this way it's more difficult for flags to sneak in on a version update of cosign. 

Side note, I wasn't sure if the we were actually able to verify Zarf packages through Kubernetes secrets as the new command text suggests, but it looks like this is actually possible which is neat.

site/src/content/docs/commands/zarf_package_verify.md
```diff
@@ -33,9 +33,12 @@ $ zarf package verify zarf-package-demo-amd64-1.0.0.tar.zst
...
 
 ```
-  -h, --help                  help for verify
-  -k, --key string            Public key for signature verification
-      --oci-concurrency int   Number of concurrent layer operations when pulling or pushing images or packages to/from OCI registries. (default 6)
+  -h, --help                                help for verify
+      --insecure-ignore-tlog                ignore transparency log verification, to be used when an artifact signature has not been uploaded to the transparency log. Artifacts cannot be publicly verified when not included in a log (default true)
+  -k, --key string                          path to the public key file, KMS URI or Kubernetes Secret
+      --oci-concurrency int                 Number of concurrent layer operations when pulling or pushing images or packages to/from OCI registries. (default 6)
+      --rekor-url string                    address of rekor STL server (default "https://rekor.sigstore.dev")
```
AustinAbro321: Does this have a use case yet?

brandtkeller:
> While I agree with mirroring the the cosign options on the SDK side, I would prefer not to mirror the commands 1:1. While it is extra steps for us to copy the command, I'd rather have us explicitly opt into any flag we expose on the CLI side, this way it's more difficult for flags to sneak in on a version update of cosign.

This was my original train of thought and the goal was discoverability in the update process. A new flag being introduced would be flagged in the docs checks giving us a chance to review changes further.  That is a slightly implicit behavior versus explicitly tested.

I think a test could exist to flag this and yet we could solidify the stance of what is explicitly adopted. it sure was fun to mess with those existing flags though ha.

site/src/content/docs/commands/zarf_package_verify.md
brandtkeller: I added them as they are functional with existing signing measures - but I'll likely remove them while migrating to explicit flags on the Zarf side to keep this PR minimal.

src/pkg/utils/cosign.go
```diff
@@ -97,62 +87,87 @@ func (opts SignBlobOptions) CheckOverwrite(ctx context.Context) error {
...
-		},
-		Offline:    true,
-		IgnoreTlog: true,
-		Timeout:    CosignDefaultTimeout,
-	}
+	var opts VerifyBlobOptions
+	opts.CommonVerifyOptions.IgnoreTlog = true
+	opts.CertVerify.IgnoreSCT = true
+	opts.CommonVerifyOptions.NewBundleFormat = true
+	opts.Timeout = CosignDefaultTimeout
```
AustinAbro321, line 111: We set this timeout, but it's never used

AustinAbro321 (commented):
small comment otherwise lgtm

src/pkg/utils/cosign.go
brandtkeller: Ah good catch - updated

AustinAbro321 (approved):

PR #4881, opened 2026-05-08
chore(deps): pin github.com/fsnotify/fsnotify to version 1.9.0

soltysh:
## Description

Pin github.com/fsnotify/fsnotify to version 1.9.0, which based on conversations from https://github.com/kubernetes/kubernetes/issues/138812 seems a safe option for now. The future versions are considered unwanted until we get an audit which was initiated by the kubernetes community. 

## Checklist before merging


soltysh:
Did some more reading and found we already rely on https://github.com/actions/dependency-review-action which has the ability to deny specific versions. Based on https://github.com/kubernetes/kubernetes/issues/138812 I've denied 1.10.0 and 1.10.1. 

For testing purposes, I've added 2nd commit for now, which actually bumps that library to one of the denied, to see what the CI will show. Once confirmed I'll drop this part.

soltysh:
I did some more testing and I'm quite disappointed with https://github.com/actions/dependency-review-action, see:
image
I have explicitly denied a direct dependency (you can't deny indirect, and probably we shouldn't) and yet the job is perfectly green:
image

If we want to do that, we'll need to create our own script, probably somewhere in `check-go-mode.yml` file.

PR #4882, opened 2026-05-08
fix: add cluster timeout cluster.New in dev deploy

AustinAbro321:
## Description

We recently add a call to `cluster.NewWithWait` to dev deploy, but we never set the context timeout, meaning it could wait forever. Unlikely practically, but might as well fix

## Checklist before merging


brandtkeller (approved):
Looks reasonable to me. There is a negligible consideration for the `defer` being that it is function-scoped - so the timer keeps running until `DevDeploy` returns. Should anyone accidentally re-use that there may be issues but that seems unlikely.

PR #4883, opened 2026-05-08
feat: verify dynamic path fields in package config are clean

AustinAbro321:
## Description

We've seen some previous potential security vulnerabilities in Zarf stemming from the fact some fields are used to build dynamic paths on the file system. There are no active vulnerabilities that I know of that use this, but this should give an extra layer of safety. Added checks to the schema to avoid any genuine mistakes on create; added checks at load time to verify that there is not a maliciously edited package to get around schema checks.

## Checklist before merging


brandtkeller (approved):
lgtm - very little likely hood that this should impact anyone and should be easily resolvable for good reasons.

PR #4891, opened 2026-05-11
feat(sign): support for keyless signing and offline verification

### LINKED ISSUE #2805: Integrate Sigstore for "keyless signing" of packages and their build provenence

Preserving supply chain integrity when moving artifacts downloaded from the Internet into an air-gapped environment is very challenging. Presently, there is no native way for zarf to verify that a downloaded package was built in a CI system using a "trusted workflow".

##### Image Signature

For a practical example, refer to the "Provenance" tab for an image in Chainguard's directory: https://images.chainguard.dev/directory/image/busybox/provenance

Notice that you can not only verify that the image is signed, you are also intrinsically verifying the image's "build provenance". Specifically, you can verify that it was built on GitHub Actions using a workflow maintained by Chainguard (i.e. `https://github.com/chainguard-images/images/.github/workflows/release.yaml@refs/heads/main`):

```bash
cosign verify \
  --certificate-oidc-issuer=https://token.actions.githubusercontent.com \
  --certificate-identity=https://github.com/chainguard-images/images/.github/workflows/release.yaml@refs/heads/main \
  cgr.dev/chainguard/busybox | jq
```

##### Additional Image Attestations

Chainguard provides additional, verifiable, guarantees about their images in the form of additional [attestations](https://slsa.dev/attestation-model):

- `https://slsa.dev/provenance/v1`: The [SLSA 1.0](https://slsa.dev/spec/v1.0/provenance) provenance attestation contains information about the image build environment.
- `https://apko.dev/image-configuration`: Contains the configuration used by that particular image build, including direct dependencies, user accounts, and entry point.
- `https://spdx.dev/Document`: Contains the image SBOM (Software Bill of Materials) in SPDX format.

This is important because, in addition to being able to download these attestations without downloading the entire image, their signatures can be verified in the same way:

```bash
# download SPDX SBOM
cosign download attestation \
  --platform=linux/amd64 \
  --predicate-type=https://spdx.dev/Document \
  cgr.dev/chainguard/busybox | jq -r .payload | base64 -d | jq .predicate

# verify its signatures
cosign verify-attestation \
  --type https://spdx.dev/Document \
  --certificate-oidc-issuer=https://token.actions.githubusercontent.com \
  --certificate-identity=https://github.com/chainguard-images/images/.github/workflows/release.yaml@refs/heads/main \
  cgr.dev/chainguard/busybox
```

- **Given** a package
- **When** the package is pushed
- **Then** the package should be signed with signed build provenance metadata (attestations) included

IMHO packages should be signed on `zarf package push` and verified on `zarf package pull` _by default_. This behavior should be opt-out via something like `--insecure` on respective commands.

We could sign packages with `cosign` after they are built and pushed to a registry. This limits our ability to provide "secure-by-default" behavior. It also leaves build provenance attestations as an exercise for the package authors, which is really annoying for things like SBOM generation. 

### Additional context

There is a lot to unpack in terms of what attestations we provide. Intentionally left open for discussion

### LINKED ISSUE #4571: Embed the Public Good Sigstore Trusted Root

As a key enabler for offline verification of bundles - the Zarf CLI/SDK could embed the current Trusted Root for the Sigstore public good instance to create an optimal path for automatic offline verification. 

packages with the public good instance could be verified offline without needing to bring any additional verification material. 

This is not meant to be the sole-provider of verification - but rather one optimal route that can be easily substituted with alternative trusted roots

- **Given** a package is signed with the public good sigstore instance
- **When** a package is deployed
- **Then** signature verification can be performed without any additional verification material

it was discussed that we can try and find a delivery mechanism that could bring the verification material along with the package (as it should not be baked into the package). This still succumbs to the issue of transport of additional required artifacts.

### Additional context

Processes should include a posture for update as well as processes for users who need new trusted roots without the ability to get a new CLI immediately. This will ensure users can still pass in a trusted root to override the embedded root.

brandtkeller:
## Description

**Note - Breaking Changes**
- The `cosign` functionality under `src/pkg/utils` has been moved to `src/pkg/signing
- The embedded structs in `signing.Sign|VerifyBlobOptions` have been updated to align with cosign.  

This PR enables the use of connected keyless signing workflows and offline-compatible verification for a keyless signed package via an embedded trusted root. 

This uses the previously implemented `zarf tools trusted-root create` command to refresh the embedded trusted root. Think of this as a mechanism to streamline the UX of enabling offline verification of keyless-signed packages so that users are not required to bring additional verification material by default - but of which they can overwrite as required. 

Additionally this PR enables the bundlesignature feature by default. Note that the `VersionRequirement` for a package with a bundle signature is `v0.71.0` which is the version that introduced the ability to load packages with the bundle file. The feature flag gates are retained such that users who cannot use the bundle still have an opportunity to opt-out until we fully remove. 

One early improvement is the UX of retrieving the keyless-signed identity and issuer. These are retrieved on `package sign` so that we can log the entries that package creators need to provide to those who want to verify - but additionally I believe we could add a `zarf package inspect signature <package>` command to make this discoverable. We can't place it directly in the build data because it is only accessible post-signing - whereby the zarf definition can no longer be modified without invalidating the signature. 

Best practices documentation to follow. 

## Try it Yourself

```bash
# Build the CLI
make build

# Pull a package for signing
zarf tools download-init

# Sign the package (this will conduct browser auth)
./build/zarf package sign zarf-init-arm64-v0.75.0.tar.zst --keyless

# (optional) turn off wifi

# Verify the package
./build/zarf package verify zarf-init-arm64-v0.75.0.tar.zst --certificate-identity brandt.keller@defenseunicorns.com --certificate-oidc-issuer https://github.com/login/oauth

```

## Related Issue

Fixes #2805
Fixes #4571

Relates to #

## Checklist before merging


hack/refresh-trusted-root.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+# Refresh the embedded Sigstore TrustedRoot used for keyless verification.
+# Run before each release. Commit the result.
+
```
brandtkeller, line 4: Making a mental note to schedule daily checks and notify the zarf channel using this process. A lag in releasing new trusted roots won't be critical but should be nice to have.

src/pkg/packager/layout/package.go
```diff
@@ -335,6 +366,10 @@ func (p *PackageLayout) VerifyPackageSignature(ctx context.Context, opts utils.V
 		return fmt.Errorf("invalid package layout: %s is not a directory", p.dirPath)
 	}
 
+	hasKey := opts.Key != ""
```
AustinAbro321, line 363: HasKey is computed before the opts.KeyRef migration, lets make a test to guard against this regression

src/cmd/package.go
```diff
@@ -1872,6 +1887,16 @@ func newPackageSignCommand(v *viper.Viper) *cobra.Command {
 	cmd.Flags().IntVar(&o.retries, "retries", v.GetInt(VPkgRetries), lang.CmdPackageFlagRetries)
 	cmd.Flags().BoolVar(&o.verify, "verify", v.GetBool(VPkgVerify), lang.CmdPackageFlagVerify)
 
+	cmd.Flags().BoolVar(&o.keyless, "keyless", false, lang.CmdPackageSignFlagKeyless)
```
AustinAbro321, line 1892: Lets make keyless and signing-key mutually exclusive

src/cmd/package.go
```diff
@@ -1872,6 +1887,16 @@ func newPackageSignCommand(v *viper.Viper) *cobra.Command {
...
+	cmd.Flags().StringVar(&o.identityToken, "identity-token", "", lang.CmdPackageSignFlagIdentityToken)
+	cmd.Flags().StringVar(&o.fulcioURL, "fulcio-url", "https://fulcio.sigstore.dev", lang.CmdPackageSignFlagFulcioURL)
+	cmd.Flags().StringVar(&o.fulcioAuthFlow, "fulcio-auth-flow", "normal", lang.CmdPackageSignFlagFulcioAuthFlow)
+	cmd.Flags().StringVar(&o.oidcIssuer, "oidc-issuer", "https://oauth2.sigstore.dev/auth", lang.CmdPackageSignFlagOIDCIssuer)
+	cmd.Flags().StringVar(&o.oidcClientID, "oidc-client-id", "sigstore", lang.CmdPackageSignFlagOIDCClientID)
+	cmd.Flags().StringVar(&o.rekorURL, "rekor-url", "https://rekor.sigstore.dev", lang.CmdPackageSignFlagRekorURL)
+	cmd.Flags().BoolVar(&o.tlogUpload, "tlog-upload", false, lang.CmdPackageSignFlagTlogUpload)
+	cmd.Flags().BoolVar(&o.yes, "yes", false, lang.CmdPackageSignFlagYes)
```
AustinAbro321: While we are basing this command off of cosign sign, I'd still lean towards making this `--confirm` to be consistent with the rest of Zarf CLI

src/pkg/signing/bundle.go
AustinAbro321, line 1: I believe all the files here justify their own signing package

src/api/v1alpha1/package.go
```diff
@@ -290,3 +292,14 @@ type VersionRequirement struct {
 	// Explanation for why this version is required
 	Reason string `json:"reason,omitempty"`
 }
+
+// ZarfSignatureMetadata records how a package was signed.
+// Populated by SignPackage and included in the signed zarf.yaml.
+// Fields are limited to values known before signing begins; identity and issuer
+// (available only after the Fulcio OIDC exchange) are stored in the Sigstore bundle.
+type ZarfSignatureMetadata struct {
```
AustinAbro321: I'm hesitant to add more information to the build struct as I believe we can get this information from the bundle.

src/pkg/packager/layout/package.go
```diff
@@ -381,6 +427,13 @@ func (p *PackageLayout) VerifyPackageSignature(ctx context.Context, opts utils.V
...
 
+	// Legacy signatures don't carry a certificate chain, so keyless identity
+	// verification has nothing to match against. Fail fast with a clear message
+	// rather than letting cosign emit a generic key/cert/bundle error.
+	if hasKeylessIdentity {
+		return errors.New("keyless verification requires bundle-format signatures, but this package has only a legacy .sig. Ask the publisher to re-sign with bundle format, or verify with --key")
+	}
+
 	// Legacy signature found
 	l.Warn("bundle format signature not found: legacy signature is being deprecated. consider resigning this zarf package with the --features='bundle-signature=true' flag.")
```
AustinAbro321: We can get rid of the feature flag mention

AustinAbro321 (changes requested):
Excited how signing is taking shape, no doubt this will improve the UX

src/pkg/signing/bundle.go
brandtkeller: Are you thinking along the lines of moving the cosign logic and everything else to a new package? My only constraint on locality was accessibility to the cosign logic - but I am not against promoting this to another package.

AustinAbro321: Yeah I think everything cosign and signing related deserves it's own package, rather than being in `utils`

brandtkeller: agreed!

brandtkeller: agreed!

src/pkg/packager/layout/package.go
brandtkeller: done!

src/api/v1alpha1/package.go
brandtkeller: I think that is fair - removed

src/cmd/package.go
brandtkeller: good call - I added a note in the help text to bridge the gap

src/cmd/package.go
brandtkeller: good call

src/pkg/packager/layout/package.go
brandtkeller: shifted the logic and added a test for good measure

src/pkg/utils/oci_artifacts.go
```diff
@@ -0,0 +1,56 @@
...
+// Package utils provides generic utility functions.
+package utils
+
+import (
+	"github.com/google/go-containerregistry/pkg/name"
+	ociremote "github.com/sigstore/cosign/v3/pkg/oci/remote"
+)
+
+// GetCosignArtifacts returns signatures and attestations for the given image.
+func GetCosignArtifacts(image string) ([]string, error) {
```
brandtkeller, line 13: I kept this function in utils - it is not soo much a signing function as it is a helper for signature/attestation retrieval.

hack/refresh-trusted-root.sh
```diff
@@ -0,0 +1,23 @@
...
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+EMBED_PATH="${REPO_ROOT}/src/pkg/utils/embedded_trusted_root.json"
```
AustinAbro321:
```suggestion
EMBED_PATH="${REPO_ROOT}/src/pkg/signing/embedded_trusted_root.json"
```

AustinAbro321 (changes requested):

hack/refresh-trusted-root.sh
brandtkeller: good eye

src/pkg/signing/bundle.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+	}
+
+	return identity, issuer
+}
+
+// BundleInfo contains parsed metadata from a Sigstore bundle file.
+type BundleInfo struct {
+	// Method is "keyless" for Fulcio-issued certificate bundles, "key" for public-key bundles.
+	Method   string
```
AustinAbro321: nit: can we make this an enum

src/pkg/packager/layout/package.go
```diff
@@ -223,12 +224,17 @@ func (p *PackageLayout) SignPackage(ctx context.Context, opts utils.SignBlobOpti
...
 	p.Pkg.Build.Signed = &signed
 
-	// Save original provenance files for rollback
+	// Save original fields for rollback
 	originalProvenanceFiles := slices.Clone(p.Pkg.Build.ProvenanceFiles)
+	originalVersionRequirements := slices.Clone(p.Pkg.Build.VersionRequirements)
+
+	// Keyless signatures require bundle format — the cert chain cannot be stored in the
+	// legacy .sig file. For key-based signing, respect the BundleSignature feature flag.
+	bundleEnabled := feature.IsEnabled(feature.BundleSignature) || opts.Keyless
```
AustinAbro321: nit: I think it makes more sense to error if the feature is not enabled and keyless is true. Probably doesn't matter much now that we're enabled by default, but that's the preferred pattern for the codebase imo

src/pkg/signing/bundle.go
```diff
@@ -0,0 +1,93 @@
...
+		return BundleInfo{Method: "key"}, nil
+	default:
+		return BundleInfo{}, fmt.Errorf("unrecognised verification content type %T", vc)
+	}
+}
+
+// ReadKeylessIdentityFromBundle parses a Sigstore bundle file and returns the
+// signer identity (cert SAN) and OIDC issuer claim. Returns an error if the
+// bundle does not contain a certificate (i.e. is not a keyless signature).
+func ReadKeylessIdentityFromBundle(bundlePath string) (identity, issuer string, err error) {
```
AustinAbro321: This function is not used

src/pkg/signing/bundle.go
```diff
@@ -0,0 +1,93 @@
...
+//   - sigstoreIssuerOIDV2: DER-encoded UTF8String, used by Fulcio v1+
+var (
+	sigstoreIssuerOIDLegacy = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 57264, 1, 1}
+	sigstoreIssuerOIDV2     = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 57264, 1, 8}
+)
+
+// extractIdentityFromCert extracts the signer identity (cert SAN) and OIDC issuer
+// from a Fulcio-issued X.509 certificate using Sigstore OID extensions.
+// SAN priority: email > URI > DNS. OID priority: V2 > legacy.
+func extractIdentityFromCert(cert *x509.Certificate) (identity, issuer string) {
```
AustinAbro321: I believe we can replace this function with `github.com/sigstore/sigstore-go/pkg/fulcio/certificate.SummarizeCertificate`

src/cmd/package.go
```diff
@@ -1974,20 +2005,57 @@ func (o *packageSignOptions) run(cmd *cobra.Command, args []string) error {
...
+	signOpts.Rekor.URL = o.rekorURL
+	signOpts.TlogUpload = o.tlogUpload
+	signOpts.SkipConfirmation = o.confirm
+
+	// Keyless certs are short-lived (~10 min). Without Rekor or a TSA timestamp
+	// the signature is unverifiable past expiry. Default --tlog-upload=true for
+	// keyless unless the user explicitly opted out.
+	if o.keyless && !cmd.Flags().Changed("tlog-upload") {
+		signOpts.TlogUpload = true
+	}
```
AustinAbro321: Is the current plan to support tsa using the same enable flag if keyless method? Cosign does use tsa by default and our default trusted root has the `timestampAuthorities` field

brandtkeller: Added the flag but not currently thinking about making it populated by default. The current flow needs one and I think the Rekor is a sane default for the "connected signing" scope. If we populate both then there are more to opt-out of versus opt-in and I think that creates more surface area for unwanted network calls.

src/pkg/signing/bundle.go
brandtkeller: great catch

src/pkg/signing/bundle.go
brandtkeller: cleaned up

src/pkg/packager/layout/package.go
brandtkeller: agreed

src/pkg/signing/bundle.go
brandtkeller: done!

brandtkeller:
Removed viper configuration in favor of additional issue / PR to track a few solutions.

src/cmd/package.go
```diff
@@ -1913,31 +1943,34 @@ func (o *packageSignOptions) run(cmd *cobra.Command, args []string) error {
...
-
-	// For local file output, use existing sign logic
 	cachePath, err := getCachePath(ctx)
 	if err != nil {
 		return err
 	}
 
-	// Load the package - do not verify
+	// Pull from OCI to a local temp dir before loading
+	if helpers.IsOCIURL(packageSource) {
```
AustinAbro321: I believe we can consolidate this with packager.LoadPackage which accepts an oci url

src/pkg/packager/layout/package.go
```diff
@@ -335,59 +355,87 @@ func (p *PackageLayout) VerifyPackageSignature(ctx context.Context, opts utils.V
...
+			}
+		}
 	}
 
-	// Check for bundle format signature (preferred)
-	bundlePath := filepath.Join(p.dirPath, Bundle)
-	_, err := os.Stat(bundlePath)
-	if err == nil {
+	if !hasVerificationMaterial {
+		return errors.New("package is signed but no verification material was provided (--key, --certificate-identity + --certificate-oidc-issuer, or --certificate)")
```
AustinAbro321:
```suggestion
		return errors.New("package is signed but no verification material was provided (--key, --certificate-identity + --certificate-oidc-issuer)")
```
there's no `--certificate` flag

src/pkg/signing/trustedroot.go
```diff
@@ -0,0 +1,41 @@
...
+// Refresh before each release with hack/refresh-trusted-root.sh.
+//
+//go:embed embedded_trusted_root.json
+var embeddedTrustedRoot []byte
+
+// writeEmbeddedTrustedRoot stages the embedded TrustedRoot JSON to a tempfile so
+// cosign's VerifyBlobCmd (which only accepts file paths) can consume it.
+// Caller must invoke cleanup when done; cleanup returns the os.Remove error.
+func writeEmbeddedTrustedRoot() (string, func() error, error) {
+	f, err := os.CreateTemp("", "zarf-trusted-root-*.json")
```
AustinAbro321: nit: use utils.MakeTempDir(config.CommonOptions.TempDirectory)

RELEASES.md
```diff
@@ -60,13 +60,27 @@ Once the prerelease artifacts are published, a Homebrew Tap PR is created. It is
...
 * [ ] Review and merge the open Release Please PR
 * [ ] The tag is automatically created and pushed, triggering the release workflow
 * [ ] Review the GitHub release:
   * [ ] Add a summary of release updates and any required documentation around updates or breaking changes
 * [ ] Ensure goreleaser workflows execute successfully and review the release assets
 * [ ] Review, approve, and merge the [homebrew-tap](https://github.com/defenseunicorns/homebrew-tap) PR for the zarf release
 
+### Embedded TrustedRoot
+
+Zarf binaries embed a Sigstore TrustedRoot JSON used by `zarf package verify` for keyless verification when `--trusted-root` is not supplied. The TUF-fetched copy is committed at `src/pkg/utils/embedded_trusted_root.json` and must be refreshed before each release so binaries ship with current Sigstore trust material.
```
AustinAbro321:
```suggestion
Zarf binaries embed a Sigstore TrustedRoot JSON used by `zarf package verify` for keyless verification when `--trusted-root` is not supplied. The TUF-fetched copy is committed at `src/pkg/signing/embedded_trusted_root.json` and must be refreshed before each release so binaries ship with current Sigstore trust material.
```

src/pkg/signing/trustedroot.go
brandtkeller: I think this is reasonable but it does make me wonder if it should be an option of the verifybloboptions struct? Keeps the signing logic decoupled? maybe something we come back to.

src/pkg/packager/layout/package.go
brandtkeller: updated

src/pkg/signing/trustedroot.go
brandtkeller: took a quick stab at this - willing to fallback to standard conventions but wanted to see if we could do so without requiring the common options in this package.

AustinAbro321: Yeah that's fine, it's a cleaner pattern we just haven't adopted it everywhere since it's generally more work to wire in a tmpdir everywhere throughout the codebase, and it's easy for us / callers to forget.

AustinAbro321 (approved):
LGTM, excited to see this in

PR #4900, opened 2026-05-14
chore(deps): vendor dependencies

soltysh:
## Description

Vendor dependencies into the repository, see [ZEP-0010](https://github.com/zarf-dev/proposals/pull/11) for reasoninig.

## Related Issue

Relates to https://github.com/zarf-dev/proposals/issues/10

## Checklist before merging


a1994sc:
Slightly off topic, but vendoring is one way cilium secures its CI CD

https://cilium.io/blog/2026/05/06/securing-cicd-open-source-lessons-from-cilium/

soltysh:
Looks like the failures might have been caused by our `.gitignore` configuration. I've updated it to skip some patterns. Let's see how CI reacts :crossed_fingers:

.gitignore
```diff
@@ -17,8 +17,8 @@
 *.vbox
 *.zst
 *.zim
-build
-data/
+/build
+/data/
```
AustinAbro321:
```suggestion
/src/test/external/data/
```
The intent of this was for this location to be excluded, we can just set it explicitly

Makefile
```diff
@@ -82,7 +82,6 @@ destroy: ## Run `zarf destroy` on the current cluster
 # Note: the path to the main.go file is not used due to https://github.com/golang/go/issues/51831#issuecomment-1074188363
 .PHONY: build
 build: ## Build the Zarf CLI for the machines OS and architecture
-	go mod tidy
```
AustinAbro321, line 85: Matter of preference, but I like having go mod tidy here so I don't have to remember to run it before I commit. Thoughts on adding back `go mod tidy` and `go mod vendor` here as well?

AustinAbro321 (commented):
minor comments, otherwise good to go

Makefile
soltysh: No preference either way, added `go mod vendor`.

.gitignore
soltysh: Updated.

Makefile
```diff
@@ -83,6 +83,7 @@ destroy: ## Run `zarf destroy` on the current cluster
 .PHONY: build
```
brandtkeller, line 83: When I run `go mod vendor` locally I see changes to the vendor directory due to our line-ending policy in `.gitattributes`. 

Should we ignore the vendor directory from this policy? 

```
vendor/** -text
```

There might be additional changes we still need to accept that. I just want to make sure we see as little drift on subsequent PR's unless necessary.

brandtkeller (changes requested):

Makefile
soltysh: After a slack discussion, I was able to reproduce the same problem locally. Applying `vendor/** -text` to `.gitattributes` solves it for me.

brandtkeller: Solved for me as well now.

brandtkeller (approved):

AustinAbro321 (approved):
LGTM

PR #4905, opened 2026-05-15
feat(agent): allow opting into agent mutation

### LINKED ISSUE #4419: Add ability to invert the zarf control of namespaces

AustinAbro321:
## Description

This introduces a new flag to init `--agent-mutation-mode` with values `opt-in` and `opt-out`. The default value is `opt-out` but we will likely change it to `opt-in` in v1.0. New namespaces created by Zarf during package deploy are labeled with `zarf.dev/agent: mutate`, so resources are still automaticaly pointed to the zarf agent. 

Resource labels always take priority so if the resource has `zarf.dev/agent: mutate` and the namespace has `zarf.dev/agent: ignore` the resource will be mutated. If the resource has `zarf.dev/agent: ignore` and the namespace has `zarf.dev/agent: mutate` the resource will be mutated.

To my surprise the podinfo-flux example works even when the agent is in opt-in mode. The reason for this is that even though Zarf doesn't deploy any resources to the intended deploy namespace (the flux resources are deployed to `flux-system) the `manifests.namespace` field is declared to `podinfo-git` and so the mutate label is added

```yaml
    manifests:
      - name: podinfo
        namespace: podinfo-git
        files:
          - git/podinfo-source.yaml
          - git/podinfo-kustomization.yaml
```

## Related Issue

Fixes  #4419

## Checklist before merging


src/pkg/state/state.go
```diff
@@ -16,6 +16,16 @@ import (
...
 
+// MutationMode controls whether the Zarf agent mutates resources by default or only on explicit opt-in.
+type MutationMode string
+
+const (
+	// MutationModeOptOut mutates all resources unless they carry zarf.dev/agent: ignore/skip.
+	MutationModeOptOut MutationMode = "opt-out"
+	// MutationModeOptIn mutates only resources (or resources in namespaces) labeled zarf.dev/agent: mutate.
+	MutationModeOptIn MutationMode = "opt-in"
+)
```
soltysh, line 27: The proposed values read weird. I'd suggest to either to directly with true/false, or if you prefer enums, which in most cases are better choice, why not `mutate` and `ignore` which is something that zarf users are already familiar with?

soltysh (commented):
One comment regarding the new flag values, otherwise this looks good

src/pkg/state/state.go
AustinAbro321: Going to continue to think about this, but yeah we can do better than opt-in and opt-out. 

We could have the states: 
- Never - don't even deploy the agent
- default-mutate - Mutate everything unless `zarf-dev/agent: ignore`
- default-ignore - ignore resources unless `zarf-dev/agent: mutate`

AustinAbro321: Thought on this some more, and I changed it to `--agent-mutation-policy` with the values `all` and `labeled`. I didn't include the value `never` as it feels too different. The other options change how the agent targets resources while `never` changes if the agent is deployed. 

I think it makes more sense to do a future `--skip-agent` flag for that, or just encourage custom init packages. It's a very niche use case, so I'd rather have the more intuitive flag experience in the base cases

AustinAbro321:
GitHub won't update this with the latest commit, I assume because of an outage earlier today. Moving this to a new PR https://github.com/zarf-dev/zarf/pull/4937

PR #4923, opened 2026-05-22
chore(examples): refresh longhorn example to v1.11.2

### LINKED ISSUE #4875: Upgrade longhorn example package to latest version

The `examples/longhorn` package is pinned to Longhorn `v1.4.0` (released
December 2022). Upstream is currently at `v1.11.1`. The example no longer
reflects a version a user would realistically deploy and the pinned image
list / chart values have drifted significantly from the current chart.

Bump the example to the latest stable Longhorn release. Concretely:

- `metadata.version` and `charts[0].version` → latest Longhorn chart version
- `longhorn-environment-check` file source URL + shasum → matching tag (is this stil accurate?)
- Regenerate the `images:` list to match the new chart's rendered manifests
  (image set has changed between 1.4 and 1.11 — new sidecars, dropped
  components)
- Refresh `values.yaml` against the new chart defaults, preserving the
  K3s-specific `kubeletRootDir` override that motivates the file

- https://github.com/zarf-dev/zarf/blob/main/examples/longhorn/zarf.yaml

### Additional context

Validation should include creating the package against an offline cluster.

acald-creator:
## Description

Updates the `examples/longhorn` package from the legacy `v1.4.0` setup to current Longhorn `v1.11.2`, and replaces the old environment check script flow with `longhornctl` preflight checks.

## Changes

- Bump Longhorn package/chart versions:
  - `metadata.version` -> `1.11.2`
  - `charts[].version` -> `1.11.2`
- Replace `longhorn-environment-check` component with `longhornctl`
  - Download `longhornctl` binary from the `v1.11.2` release
~  - Update `onDeploy` to run:~
~    - `./longhornctl check preflight`~
- Regenerate pinned `images:` list for the updated chart/image set
- Refresh `examples/longhorn/values.yaml` to previous values

## Validation

```bash
./build/zarf package create examples/longhorn -o build -a amd64 --confirm

build/zarf-package-longhorn-amd64-1.11.2.tar.zst
```

## Related Issue

Fixes #4875

## Checklist before merging


AustinAbro321:
Hey @acald-creator thanks for the PR! Is this ready for a review? If so move it out of draft and we will take a look

acald-creator:
@AustinAbro321 ready for review

AustinAbro321:
@acald-creator did you get this to deploy successfully? I am trying locally and running into problems

acald-creator:
@AustinAbro321 let me take a look

acald-creator:
@AustinAbro321 I tested deployment on kind and k3s, and it deployed successfully for me

examples/longhorn/zarf.yaml
```diff
@@ -2,31 +2,26 @@ kind: ZarfPackageConfig
...
     required: true
     files:
-      - source: https://raw.githubusercontent.com/longhorn/longhorn/v1.4.0/scripts/environment_check.sh
-        target: environment_check.sh
-        shasum: 4df400fa81e65e51d83db8ff8e4f513ddd32c124619bb36c94319e71ca4d0a4d
-        executable: true
-      - source: https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
-        target: jq
-        shasum: af986793a515d500ab2d35f8d2aecd656e764504b789b66d7e1a0b727a124c44
+      - source: https://github.com/longhorn/cli/releases/download/v1.11.2/longhornctl-linux-amd64
```
brandtkeller: This script is indeed no longer available but introducing a binary will add a fair bit of complexity (IE this cannot be used with arm64 as written and making this architecture aware could be complex). 

Two suggestions:
1. Look at any "portable" solutions we might be able to leverage
2. remove the environment check / longhornctl component entirely

For ease of update - I think it is valid to remove that component and document any prerequisites in the package documentation.

examples/longhorn/values.yaml
```diff
@@ -1,332 +1,10 @@
...
-    name: ~
-    dataSourceType: ~
-    dataSourceParameters: ~
-    expectedChecksum: ~
-  defaultNodeSelector:
-    enable: false # disable by default
-    selector: []
-  removeSnapshotsDuringFilesystemTrim: ignored # "enabled" or "disabled" otherwise
-
+# Keep this file minimal so we inherit Longhorn chart defaults.
```
brandtkeller: I tend to agree here in sentiment but the current capabiltiies of `zarf dev|package inspect values-files` do offer some capability to airgapped environments if we have re-specified the whole values file (or at least what is pertinent). 

I'd lean towards maintaining the copy but feel free to push back.

brandtkeller (changes requested):
Thanks for taking this on! A few thoughts but otherwise I think the only thing we need to do is validate functionality. 

- Package creates successfully
- Package deploys successfully (preferably offline but I can test that if you can't)
- Validate operational functionality (can we deploy a workload that has a persistent volume requirement)

examples/longhorn/zarf.yaml
acald-creator: I was thinking about that, I wanted to try it out and see how longhornctl would looks like, but I think that removing the component would be best and document any prerequisites.

I can look into your suggestions in the meantime.

examples/longhorn/values.yaml
acald-creator: Thank you for your feedback, I am still learning Zarf, so I will look more into zarf's current capabilities, and look more into this.

brandtkeller:
Running a test currently on an airgapped cluster. Deployment was successful. 

Digging some more I found that there is a model for running preflight checks via the CLI container image - see: https://github.com/longhorn/cli/blob/v1.12.0/deploy/check.yaml

We might consider adding this as an optional component (likely copying the manifest and using a version-pinned image). I won't block on this being a requirement - so it can be a future improvement if you feel inclined. 

Appreciate the patience - we've been on the move and are finally back to reviewing.

brandtkeller (dismissed):
Great work! This is functioning on my airgapped cluster over the weekend without any issues. 

I would like to file an issue for two improvements. One being to research an optional preflight daemonset component that the CLI has upstream. The other being an improved healthcheck given that the helm release succeeds and the CLI reports package deployment success before it inevitably begins rolling out more resources (which could fail). Let us know if you'd be interested in looking at those further. Appreciate the patience.

acald-creator:
> Great work! This is functioning on my airgapped cluster over the weekend without any issues.
> 
> I would like to file an issue for two improvements. One being to research an optional preflight daemonset component that the CLI has upstream. The other being an improved healthcheck given that the helm release succeeds and the CLI reports package deployment success before it inevitably begins rolling out more resources (which could fail). Let us know if you'd be interested in looking at those further. Appreciate the patience.

Sure, I would be interested! You're welcome, not a problem.

examples/longhorn/values.yaml
```diff
@@ -2,6 +2,7 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 global:
+  imageRegistry: "127.0.0.1:31999"
```
AustinAbro321:
```suggestion
  imageRegistry: "###ZARF_REGISTRY###"
```
This will keep this working regardless of the registry address

AustinAbro321 (changes requested):
Nice work, one change to make this more robust

brandtkeller (approved):

AustinAbro321 (approved):
LGTM, thanks!

PR #4925, opened 2026-05-22
fix(tools,wait): handle context propagation for networking calls

### LINKED ISSUE #4458: Tie zarf tools wait-for into parent context cancellation and deadlines
We fixed a gap in wait-for's timeout handling in #4452 on servers that hang indefinitely, but there's still some gaps. We have a parent ctx available but neither the HTTP client nor the net dialer use the context variants with cancels. 

https://github.com/zarf-dev/zarf/blob/main/src/pkg/utils/wait.go#L168

### Additional context
We could also refactor this to use go-retry like other places in Zarf we use retries.

Via #4452

knrc:
## Description

Ensures the context is propagated to http client and networking calls

## Related Issue

Fixes #4458 

## Checklist before merging

### DIFF
--- a/src/pkg/wait/wait.go
+++ b/src/pkg/wait/wait.go
@@ -399,7 +399,8 @@ func ForNetwork(ctx context.Context, protocol, address, condition string, timeou
 
 func forNetwork(ctx context.Context, protocol string, address string, condition string, timeout time.Duration, waitInterval time.Duration) error {
 	l := logger.From(ctx)
-	expired := time.After(timeout)
+	timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
 
 	condition = strings.ToLower(condition)
 	if condition == "" {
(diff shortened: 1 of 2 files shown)


src/pkg/wait/wait.go
soltysh: Your current fix wires context to the current `time.Sleep`, which is ok. But now we have:
```golang
select...
select...
    case...
```
which unnecessarily complicates the code. How about squashing both selects together and move the rest of the code after it? This way the code could look like the following:
```golang
	for {
		pauseTimer := time.NewTimer(delay)

		// handle expiration, context cancellation and sleep in a single select function
		select {
		case <-ctx.Done():
			pauseTimer.Stop()
			return errors.New("received interrupt")
		case <-expired:
			pauseTimer.Stop()
			return errors.New("wait timed out")
		case <-pauseTimer.C:
		}

		// now you're able to serve the actual request, which has context wired
		delay = waitInterval
		switch protocol {
		case "http", "https":
....
```
This would eliminate too many selects, and also eliminate code indentation, which means the code will be easier to read. Opinions?

knrc: That sounds good, I can do that.

knrc: @soltysh should be good to go now, thanks for the feedback

knrc:
FYI I've also just rebased on main

src/pkg/wait/wait.go
```diff
@@ -416,54 +416,31 @@ func forNetwork(ctx context.Context, protocol string, address string, condition
...
-		delay = waitInterval
-
+		pauseTimer := time.NewTimer(delay)
 		select {
 		case <-expired:
+			pauseTimer.Stop()
 			return errors.New("wait timed out")
 		case <-ctx.Done():
+			pauseTimer.Stop()
 			return errors.New("received interrupt")
```
brandtkeller: In order for callers to distinguish between different errors I think we should return the error directly:


```suggestion
			return fmt.Errorf("wait cancelled: %w", ctx.Err())
```

 I believe this would apply above  as well

brandtkeller (changes requested):
hey Kevin! Appreciate the PR. 

Minor comment on errors. There are some interesting pre-existng behaviors to sort through here that I am still trying to navigate. Particularly around the competing timers. 

Otherwise I think outward error propagation tied into the testing could capture some edge cases.

src/pkg/wait/wait.go
soltysh: Good idea.

soltysh (dismissed):
this lgtm modulo Brandt's comment regarding context's error.

src/pkg/wait/wait.go
knrc: Sounds good, I can work on it today and send an update to the PR.

knrc: There's only one `ctx.Done()`, so I've used your updated change there.  I also noticed the `expired` timer wasn't being stopped, so I've added that.  I'll add some tests to cover some of the edge cases.

knrc: I've finally updated the PR, there's now a table driven test for http/tcp timeouts which I believe covers them all.

knrc:
Sorry, one more push to rebase against main

soltysh (dismissed):
lgtm

brandtkeller:
Hey Kevin! Could I ask for you to resign the commits? we do enforce signed commits. Apologies for the overhead.

knrc:
> Hey Kevin! Could I ask for you to resign the commits? we do enforce signed commits. Apologies for the overhead.

Sorry, I thought I had already done that.  Doing it now.

knrc:
> > Hey Kevin! Could I ask for you to resign the commits? we do enforce signed commits. Apologies for the overhead.
> 
> Sorry, I thought I had already done that. Doing it now.

Should be good now

```
commit 9c4ace03ad76723d4cbcad361984f11bf859362a (HEAD -> 4458_context_propagation, knrc/4458_context_propagation)
Good "git" signature for kev.conner@gmail.com with RSA key SHA256:WWuI61ZRa8aKH2kTsegHAZBE7gTGxTl06uMquONnb1U
Author: Kevin Conner <kev.conner@gmail.com>
Date:   Fri May 22 15:20:58 2026 -0700

    fix(tools,wait): handle context propagation for networking calls

    Signed-off-by: Kevin Conner <kev.conner@gmail.com>
```

knrc:
Just a rebase

knrc:
just another rebase on main

src/pkg/wait/wait.go
```diff
@@ -399,7 +399,8 @@ func ForNetwork(ctx context.Context, protocol, address, condition string, timeou
 
 func forNetwork(ctx context.Context, protocol string, address string, condition string, timeout time.Duration, waitInterval time.Duration) error {
 	l := logger.From(ctx)
-	expired := time.After(timeout)
+	expired := time.NewTimer(timeout)
```
brandtkeller: testing my own understanding across this function more - how would you feel about deriving a single context from timeout here such that the http client and DialContext share a deadline? I get the impression we could then eliminate the expired timer perhaps?

Something to the effect of:

```golang
func forNetwork(ctx context.Context, protocol string, address string, condition string, timeout time.Duration, waitInterval time.Duration) error {
	l := logger.From(ctx)
	ctx, cancel := context.WithTimeout(ctx, timeout)
	defer cancel()

	// ...

	for {
		pauseTimer := time.NewTimer(delay)
		select {
		case <-ctx.Done():
			pauseTimer.Stop()
			if errors.Is(ctx.Err(), context.DeadlineExceeded) {
				return errors.New("wait timed out")
			}
			return fmt.Errorf("wait cancelled: %w", ctx.Err())
		case <-pauseTimer.C:
		}
```

brandtkeller (changes requested):
One thought otherwise this is a definite improvement. Feel free to push back!

src/pkg/wait/wait.go
knrc: That's certainly possible, I think the only issue we need to consider would be distinguishing between the deadline coming from the parent context or from this one when we report the time out.  That should be straight forward though.

knrc: okay, updated and rebased

brandtkeller (approved):
lgtm - thank you for the support and flexibility!

knrc:
> lgtm - thank you for the support and flexibility!

I'll look for more once this merges

soltysh (approved):
lgtm

Issue #4919, opened 2026-05-21
Zarf injector picks random image on init.


kylegalloway:
### Environment
Device and OS: Macbook Pro/macOS
App version: 0.76.0
Kubernetes distro being used: kind version 0.29.0, also replicated on eks, 1.34
Other:

### Steps to reproduce
Following instructions in [Getting Started](https://docs.zarf.dev/getting-started/)
1. `kind delete cluster && kind create cluster`
2. `zarf init --log-level=debug`. Select 'Y' to download the default init package, Select 'Y' to confirm deployment, Select 'N' to ignore gitea deployment. Note the line `DBG selected image for injector node=kind-control-plane image=registry.k8s.io/coredns/coredns:v1.12.0` (might also be etcd or some other default image that is part of kubernetes).
3. Deploy simple busybox container and label the namespace to be ignored by zarf agent: `kubectl run zarf-ignore-test --image=busybox --labels="zarf.dev/agent=ignore" -- sleep 3600`, `kubectl label namespace default zarf.dev/agent=ignore`
4. Re-run `zarf init --log-level=debug`. Note that the image selected by the injector is now different (usually busybox).

### Expected result

There should be a consistent image or way to influence the image choice by the injector.

### Actual Result

In practice, the injector chooses the first image that it finds on the node that isn't caught by [the simple regex filter.](https://github.com/zarf-dev/zarf/blob/main/src/pkg/cluster/injector.go#L40)
This is a problem when using a private registry, as the injector will not have credentials for the private registry and thus go into an ImagePullBackOff.
This is somewhat fixed in the Daemonset path for the injector, in that it tries to filter down to the pause image. However, even that simple filtering may not be correct. Really, there needs to be a way to set a variable/flag to pass the correct image.

### Visual Proof (screenshots, videos, text, etc)
```
> kubectl -n zarf describe pod injector | tail -n10
Events:
  Type     Reason   Age                From     Message
  ----     ------   ----               ----     -------
  Normal   Pulling  13s (x2 over 28s)  kubelet  spec.containers{injector}: Pulling image "my-private-registry.me.io/custom-image:v0.0.1"
[...]

PR #4935, opened 2026-05-27
feat(init): prefer injector image without imagePullSecrets

AustinAbro321:
## Description

This also shifts the logic to prefer images without imagePullSecrets, I assume 99% of the time this should make us land on kube-system. 

## Related Issue

Relates to #4919 

## Checklist before merging


src/pkg/cluster/injector.go
```diff
@@ -82,15 +85,19 @@ func (c *Cluster) StartInjection(ctx context.Context, tmpDir, imagesDir string,
 			corev1.ResourceCPU:    resource.MustParse("1"),
 			corev1.ResourceMemory: resource.MustParse("256Mi"),
 		})
-	injectorImage, injectorNodeName, err := c.getInjectorImageAndNode(ctx, resReq, architecture)
-	if err != nil {
-		return 0, err
+	injectorImage := opts.InjectorImage
+	var injectorNodeName string
+	if injectorImage == "" {
+		injectorImage, injectorNodeName, err = c.getInjectorImageAndNode(ctx, resReq, architecture)
```
soltysh: What is the expected node name to be used here when we are working with injected image?

soltysh: I believe we should either get the node explicitly injected or discover the node that has that image available, no?

AustinAbro321: This is a great point, and made me rethink adding the `--injector-image` as it would need a `--injector-node` flag as well since an image may be pullable by the node, but not yet pulled so .status.images is unreliable. I decided to remove the `--injector-image` flag to simplify the ux

soltysh (approved):
lgtm

PR #4937, opened 2026-05-27
feat(init): add init option for the agent to only mutate labeled resources

AustinAbro321:
## Description

This introduces a new flag to init `--agent-mutation-policy` with values `all` and `labeled`. The default value is `all` but we will likely change it to `labled` in v1.0. New namespaces created by Zarf during package deploy are labeled with `zarf.dev/agent: mutate`, so resources are still automaticaly pointed to the zarf agent. 

Resource labels always take priority so if the resource has `zarf.dev/agent: mutate` and the namespace has `zarf.dev/agent: ignore` the resource will be mutated. If the resource has `zarf.dev/agent: ignore` and the namespace has `zarf.dev/agent: mutate` the resource will not be mutated

To my surprise the podinfo-flux example works even when the agent is in labeled mode. The reason for this is that even though Zarf doesn't deploy any resources to the intended deploy namespace (the flux resources are deployed to `flux-system`) the `manifests.namespace` field is declared to `podinfo-git` and so the mutate label is added

```yaml
    manifests:
      - name: podinfo
        namespace: podinfo-git
        files:
          - git/podinfo-source.yaml
          - git/podinfo-kustomization.yaml
```

## Checklist before merging


src/internal/agent/hooks/pods.go
```diff
@@ -67,15 +63,11 @@ func getAnnotationKey(ctx context.Context, image string) string {
...
+func mutatePod(ctx context.Context, r *v1.AdmissionRequest, c *cluster.Cluster, pod *corev1.Pod) (*operations.Result, error) {
 	l := logger.From(ctx)
-	pod, err := parsePod(r.Object.Raw)
-	if err != nil {
-		return nil, fmt.Errorf(lang.AgentErrParsePod, err)
-	}
 
 	if r.SubResource != "" {
-		return mutatePodSubresource(ctx, r, cluster)
+		return mutatePodSubresource(ctx, r, c)
```
soltysh: Since you already have a pod resource here, you should be able to pass that here down, and safe the double unmarshalling which you're doing in https://github.com/zarf-dev/zarf/blob/b4d7f59fb87c3561f8f3c5c2ca751f2be1f11900/src/internal/agent/hooks/pods.go#L172

src/internal/agent/hooks/common.go
```diff
@@ -28,6 +32,45 @@ const (
...
+		}
+		if !operations.ShouldMutate(obj.GetLabels(), nsLabels, mode) {
+			return &operations.Result{Allowed: true, PatchOps: []operations.PatchOperation{}}, nil
+		}
+		return fn(ctx, r, obj)
+	}
+}
+
+func getNamespaceLabels(ctx context.Context, c *cluster.Cluster, name string) (map[string]string, error) {
+	ns, err := c.Clientset.CoreV1().Namespaces().Get(ctx, name, metav1.GetOptions{})
```
soltysh, line 67: Potential improvement, have you considered caching the labels here? For example up to a minute or so, the labels rarely change that fast, so 1min delay could buy us some time, and increase the throughput on subsequent requests.

src/cmd/initialize.go
```diff
@@ -459,3 +462,19 @@ func (o *initOptions) validateInitFlags() error {
 
 	return nil
 }
+
+type mutationPolicyFlag state.MutationPolicy
```
soltysh: Wouldn't just using a simple string flag and validating inside `validateinitFlags` be simpler than this?

src/internal/agent/hooks/common.go
```diff
@@ -28,6 +32,45 @@ const (
...
+func withMutationGuard[T any, PT interface {
+	*T
+	metav1.Object
+}](
+	ctx context.Context,
+	c *cluster.Cluster,
+	mode state.MutationPolicy,
+	fn func(ctx context.Context, r *admission.AdmissionRequest, obj PT) (*operations.Result, error),
+) operations.AdmitFunc {
+	return func(r *admission.AdmissionRequest) (*operations.Result, error) {
```
soltysh, line 46: It seems you're passing `ctx` twice here:
1. when you're invoking `withMutationGuard` which means you're getting the early context into this function
2. inside this function again to pass to `fn`, that one is fine.

I think, the right way would be to pass per-request context, thus expanding the `AdminFunc` signature with context like so:
```
type AdmitFunc func(ctx context.Context, request *admission.AdmissionRequest) (*Result, error)
```
especially that you already wrap each execution, so you could guard each request separately.

Although the more I dig into it, it'll likely require a bigger rewrite, to ensure the admission handler from https://github.com/zarf-dev/zarf/blob/b4d7f59fb87c3561f8f3c5c2ca751f2be1f11900/src/internal/agent/http/admission/handler.go#L69 properly passes a request-scope context.

soltysh: I'd suggest to fix that as followup, or I can pick that up.

soltysh (commented):
Dropping the double umarshaling and the flag validating would be my two blockers. The rest is just nice to have, or can be addressed in followup (especially that context one).

src/internal/agent/hooks/common.go
AustinAbro321: One thing we might be able to do if I understand the package correctly is to setup a [shared informer ](https://pkg.go.dev/k8s.io/client-go/informers) which would cache the item in memory and automatically update it for us

src/cmd/initialize.go
AustinAbro321: Yeah, I liked the idea of using the pflag construct, but I suppose it is more code, and adds a type to the help docs that users don't care about.

src/internal/agent/hooks/common.go
AustinAbro321: Makes sense, I was able to write the complete fix, I'll submit a quick follow up after this PR is in

src/internal/agent/hooks/common.go
soltysh: That'd be a good improvement as well.

soltysh (approved):
lgtm

PR #4939, opened 2026-05-27
fix(signing)!: resolve auth flow for CI environments

### LINKED ISSUE #4938: keyless signing in CI fails to detect CI environment
### Steps to reproduce
1. Attempt to sign a package in github CI only passing `--keyless --confirm` 

### Expected result

Package is signed using the `id-token: write` permission

### Actual Result

fails to open a browser

https://github.com/zarf-dev/zarf/actions/runs/26530532678/job/78145730626

### Severity/Priority

High

brandtkeller:
## Description

Signing the init package in the release workflows fails currently due to the explicitly defined `AuthFlow = normal` in the `DefaultSignBlobOptions`. This being set silently discards the OIDC token fetched from the provider and falls back to browser OAuth. 

Additionally this explicitly defines registering the providers in the event that ever changes in cosign proper.

See workflow tests for signing in CI:
- [Failing](https://github.com/zarf-dev/zarf/actions/runs/26541068336/job/78182368373)
- [Successful](https://github.com/zarf-dev/zarf/actions/runs/26541379700/job/78183381789)

Documentation PR to follow shortly. 

## Related Issue

Fixes #4938

## Checklist before merging


AustinAbro321:
From what I can tell these are already registered through transitive dependencies, but I do think it's a good idea to make them more explicit. 

To test that out I ran this unit test, which passes on main, but not a new go project
```go
func TestOIDCProvidersAlreadyRegistered(t *testing.T) {
	t.Parallel()
	registered := make(map[string]bool)
	for _, p := range providers.Providers() {
		registered[p.Name] = true
	}
	for _, name := range []string{"github-actions", "google-workload-identity", "google-impersonate", "envvar", "filesystem", "buildkite-agent", "spiffe"} {
		require.True(t, registered[name], "expected provider %q to be registered without an explicit providers/all import", name)
	}
}
```

so I will approve, but I recommend changing it to a refactor from a fix

AustinAbro321 (dismissed):

AustinAbro321:
I am also not sure what is actually causing the issue, maybe I'm misunderstanding and this is the problem

brandtkeller:
> I am also not sure what is actually causing the issue, maybe I'm misunderstanding and this is the problem

(this should have been draft initially - oops)

You nailed the original investigation - I was looking to see how providers were being registered  and adding them explicitly. I then wanted to test with an actual workflow to get some confidence.  Registration is inherently differnt from enabled though so I wanted to validate functionality.

If you look at commit [b73c425](https://github.com/zarf-dev/zarf/pull/4939/commits/b73c4255e70e0dfcae21c4fd3431fb2dbaaee8a4) I created a [workflow](https://github.com/zarf-dev/zarf/actions/runs/26541068336/job/78182368373) for keyless package signing that subsequently fails. 

Digging further - the `normal` `AuthFlow` setting that we set short circuits the ambient OIDC token that is fetched and silently falls through to the browser OAuth. See [here](https://github.com/sigstore/cosign/blob/main/internal/auth/auth.go#L111-L115). By removing our setting this and instead retaining an empty string we can see this same testing workflow [pass](https://github.com/zarf-dev/zarf/actions/runs/26541379700/job/78183381789).

I then removed the temporary workflow for signing. Should be ready for review now.

AustinAbro321 (approved):

Issue #4917, opened 2026-05-20
Support signature verification flags on all package commands


brandtkeller:
### Is your feature request related to a problem? Please describe.

Signature verification happens on all package load operations. As zarf expands the surface area of signing and verification options we need the ability to perform verification any time a package is loaded.

As such, the `zarf package` subcommand should have all required verification flags for use by children commands.

### Describe alternatives you've considered

Have the expectation that operators run package verify before all load operations.

knrc:
@brandtkeller Do you mind if I take this one?

brandtkeller:
@knrc This one has a minor blocker from #4946 that I would like to clear up first. Appreciate the interest!

PR #4940, opened 2026-05-27
chore(docs): add keyless signing documentation

### LINKED ISSUE #4906: Signing and Verification Documentation
## Summary

After we begin opening up additional pathways for signing and verification - it will be important that we document the capabiltiies and implied or otherwise configuration for use. 

This should include:
- Keyless Signing and Verification
- Trusted root (embedded vs provided)
- Connectivity requirements and Optionality
- Private infrastructure
- Transparency Log versus Timestamp Authority

brandtkeller:
## Description

Adds basic documentation for keyless signing and covers nuance such as when to use the transparency log versus signed timestamps. 

There is additional proposed work around trusted signers that could then expand upon verification recommendations - but we'll need to vet that implementation when the time comes. 

Adds a note about the gap in verfication options during package operations with `zarf package <subcommand>` and the associated issue #4917 

## Related Issue

Fixes #4906

## Checklist before merging


site/src/content/docs/ref/package-signing.mdx
```diff
@@ -187,6 +274,37 @@ Verified OK
...
+# Use it during verification
+zarf package verify zarf-package-example-amd64.tar.zst \
+  --trusted-root trusted-root.json \
+  --certificate-identity-regexp "https://github.com/my-org/my-repo/.github/workflows/release.yml@refs/tags/" \
+  --certificate-oidc-issuer "https://token.actions.githubusercontent.com"
+```
+
+Users running private Sigstore infrastructure can supply their own TrustedRoot the same way via `--trusted-root /path/to/custom.json`.
+
+**Air-gap environments**: when a keyless package was signed with `--tlog-upload` (the default for `--keyless`), the Rekor inclusion proof is embedded in the bundle. Combined with the embedded TrustedRoot (which includes Rekor's public key), tlog verification works fully offline — no extra flags required.
```
AustinAbro321: This line feels very similar to the first line in this section. What do you think of moving the "if you need fresher trust material" section after the tlog explainations, then combining the first line in this section with this part.

site/src/content/docs/ref/package-signing.mdx
```diff
@@ -302,12 +426,42 @@ zarf package sign zarf-package-example-amd64.tar.zst \
...
+```bash
+zarf package sign zarf-package-example-amd64.tar.zst \
+  --keyless \
+  --identity-token "$MY_OIDC_TOKEN" \
+  --confirm
+```
+
+For GitHub Actions specifically, ensure `id-token: write` is set in the job's `permissions` block — cosign will acquire the token automatically.
+
+### Keyless Verification: Certificate Identity Does Not Match
```
AustinAbro321: IMO the error message here is clear enough where it doesn't warrant a section

site/src/content/docs/ref/package-signing.mdx
```diff
@@ -112,6 +112,65 @@ All KMS providers require proper authentication and permissions configured on yo
...
+  --confirm
+```
+
+Ensure the token value is registered as a masked secret in your CI platform so it is redacted from log output.
+
+:::note
+
+Keyless signatures include a Rekor transparency log entry that extends verifiability past the ~10-minute Fulcio certificate validity window. There are two reasons you might use `--tsa-server-url` with `--tlog-upload=false` instead (since `--keyless` enables `--tlog-upload` by default):
+
+1. **Rekor is unavailable** in your environment and you still need long-term verifiability
```
AustinAbro321:
```suggestion
1. **Rekor is unavailable** in your environment
```
nit: I'd rather shorten this framing since there isn't really a real situation where you don't need long term verifiability

AustinAbro321 (commented):
Nice work, this will definitely help some folks understand keyless better. Some comments

site/src/content/docs/ref/package-signing.mdx
```diff
@@ -87,39 +72,92 @@ zarf package sign zarf-package-example-amd64.tar.zst \
...
+:::note
 
-### Sign an Existing Package
+By default, keyless signatures include a Rekor transparency-log entry. This entry extends verification beyond the Fulcio certificate's approximately 10-minute validity period.
 
-Sign a package after it has been created:
+Use a TSA timestamp instead when Rekor is unavailable or you do not want to publish the signing event in Rekor:
+
+1. Rekor is unavailable in your environment.
+2. Rekor is a public transparency log, and a TSA timestamp better meets your privacy requirements.
```
AustinAbro321: Some duplicated content here

AustinAbro321 (commented):
one comment otherwise lgtm

site/src/content/docs/ref/package-signing.mdx
brandtkeller: blah - great catch. I tried to give this a entire fresh write/read and my eyes still read straight over that.

AustinAbro321 (approved):

PR #4946, opened 2026-05-28
feat(viper): support viper configuration for sign/verify

### LINKED ISSUE #4907: Sign and Verify viper configuration support

Add support for `zarf package sign|verify` flags after keyless has landed in #4891 

### Additional context

There is a known constraint such that `package.verify` is currently used for operations where the package is "loaded" into memory and whether we enforce signature verification. 

We will need to handle this collision as standard convention should likely be that `package verify` and `package.verify` be aligned.

brandtkeller:
## Description

This adds viper configuration for `zarf package sign|verify`. 

Notably it adds the viper configurations for `verify` under the root of `package` instead of `package.verify`. This is for a few reasons:
1. We potentially want to introduce flags for the verification flags on all package commands. As such this should probably be common configuration as seen by the pre-existence of `package.public_key` which has always been a root package viper config. 
2. It allows us to continue using the existing `package.verify` as a boolean for the enforcement behaviors without creating any breaking changes

There is optionality here and is highly subjective. We could implement `package.verify.*` as the placeholder. It follows existing conventions for the commands. `verify` is a nuanced command / workflow in that we intend to continue to support its use on all package load operations. The same may not necessarily apply to signing as I believe we can encourage users to implement that in multiple steps. 

## Related Issue

Fixes #4907

## Checklist before merging


src/cmd/package.go
```diff
@@ -1868,7 +1869,7 @@ type packageSignOptions struct {
 }
 
 func newPackageSignCommand(v *viper.Viper) *cobra.Command {
-	o := &packageSignOptions{}
+	o := &packageSignOptions{v: v}
```
brandtkeller, line 1872: Note: inclusion of a viper instance allows us to isolate instances while testing. Our overriding behavior leaves some room to change something in the future without realizing the implications (see test cases).

AustinAbro321: Yeah we're also using it to check if a field is set, I'd be fine instead adding a boolean for that, but I think this is fine

AustinAbro321 (approved):

PR #4948, opened 2026-05-29
ci: improve codecov accuracy

AustinAbro321:
## Description

Our codecov was only tracking coverage within each package, this makes it so any test on a package will affect coverage.

## Checklist before merging


soltysh (approved):
lgtm

PR #4949, opened 2026-05-29
feat(schema): support multiple API versions from a single schema file

AustinAbro321:
## Description

This creates a new schema at the root level of the repository that supports multiple versions. It ends up being very long since it can't define definitions at the top otherwise they would clash. 

I opted to group the `zarf.schema.json` by API version because so that it works regardless of the user's current version. For instance users will be able to point all their zarf.yaml files at the same schema file, and it will work regardless of whether they are on v1alpha1 or v1beta1. We encourage users to point to the zarf.schema.json file in our docs https://docs.zarf.dev/ref/dev/#vscode. 

This PR also allows the website to serve multiple schema formats, but I've hidden the v1beta1 schema for now. 

## Related Issue

Relates to #3433

## Checklist before merging


site/src/content/docs/schema/v1beta1-package.mdx
```diff
@@ -0,0 +1,12 @@
+---
+title: v1beta1 Package
+sidebar:
+  order: 20
+  hidden: true
+---
+
+import SchemaReference from '@components/SchemaReference.astro';
+
+[JSON Schema source](https://github.com/zarf-dev/zarf/blob/main/src/pkg/schema/zarf-v1beta1-package-schema.json)
```
brandtkeller, line 10: This is the valid existing v1beta1 schema - is the intent to retain this and this PR to be purely additive? I'm good with that but just wanted to make sure given the `schema` package only adding a getter for the combined.

AustinAbro321: Yeah, this is still hidden, but basically wanted to make sure that this PR will allow us to introduce and access new schemas, hence the `SchemaReference` object

src/pkg/schema/generate.go
```diff
@@ -48,6 +60,77 @@ func writeSchema(apiVersion, filename string, rootType any) error {
...
+	modulePath := fmt.Sprintf("github.com/zarf-dev/zarf/src/api/%s", apiVersion)
+	if err := reflector.AddGoComments(modulePath, typePackagePath); err != nil {
+		return nil, fmt.Errorf("unable to add Go comments to %s schema: %w", apiVersion, err)
+	}
+
+	schema := reflector.Reflect(rootType)
+	// Drop the per-branch $schema so only the combined schema declares the dialect.
+	schema.Version = ""
+	return schema, nil
+}
```
soltysh, line 97: This method introduces a significant code duplication, why not rewrite `generateSchema` to accept a list of APIs to generate, and it will produce either single or combined schema?

soltysh (commented):
One comment regarding combining schema generation helpers

src/pkg/schema/generate.go
AustinAbro321: Good call, done

soltysh (approved):
lgtm

PR #4951, opened 2026-05-29
feat(values)!: support schema merge on imports

### LINKED ISSUE #4877: Merge values schema during component import
Zarf allows merging child values on component import, but will completely ignore the child values schema. Zarf should treat the schema the same way it does values, and merge the child schema with the parent schema, favoring the parent on conflicts 

  - **Given** a parent package that imports a child component, where both define a `values-schema.json`:

    Child component schema (`registry/values-schema.json`):
    ```json
    {
      "type": "object",
      "required": ["registry"],
      "properties": {
        "registry": {
          "type": "object",
          "required": ["image"],
          "properties": {
            "image": { "type": "string", "description": "child description" }
          }
        }
      }
    }
    ```

    Parent schema (`values-schema.json`):
    ```json
    {
      "type": "object",
      "properties": {
        "registry": {
          "type": "object",
          "properties": {
            "image": { "type": "string", "description": "parent description" },
            "tag":   { "type": "string" }
          }
        }
      }
    }
    ```

  - **When** the parent package is created (`zarf package create`)

  - **Then** the child's schema is merged into the parent's — with the parent winning on conflicts — mirroring how `values.yaml` files are already merged on import:

    ```json
    {
      "type": "object",
      "required": ["registry"],
      "properties": {
        "registry": {
          "type": "object",
          "required": ["image"],
          "properties": {
            "image": { "type": "string", "description": "parent description" },
            "tag":   { "type": "string" }
          }
        }
      }
    }
    ```

    `image.description` comes from the parent (conflict → parent wins), `tag` is contributed by the parent, and `required: ["image"]` is preserved from the child (no conflict → kept).

We could completely ignore the child values, however, this may be un-intuitive. It would also mean that skeleton packages have no way to require values from those importing them

brandtkeller:
## Description

Support importing child package values schemas and merging them during package create for local import paths only. Skeleton values support will follow. Given that we do not want to modify the source parent schema as we process imports and it not being a list - I've proposed the current strategy:

Updating the function parameters and adding an intermediary type `DefinedPackage` to support a proper handoff between the require packages.

I had originally created an exported field `ImportedSchemas` which is used to track schemas during the import resolution process for access during `AssemblePackage()`. I felt like this modified the public API in an inappropriate way.

This does take a stance on not allowing the use of `$refs` as they create concerns for availability of the referenced files and I figured an explicit error would be better than trying to handle them outside of the schema. This means the schemas used are required to be flat.  

## Related Issue

Fixes #4877

## Checklist before merging


src/pkg/packager/load/import.go
```diff
@@ -37,13 +37,13 @@ func getComponentToImportName(component v1alpha1.ZarfComponent) string {
 	return component.Name
 }
 
-func resolveImports(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath, arch, flavor string, importStack []string, cachePath string, skipVersionCheck bool, remoteOptions types.RemoteOptions) (v1alpha1.ZarfPackage, error) {
+func resolveImports(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath, arch, flavor string, importStack []string, cachePath string, skipVersionCheck bool, remoteOptions types.RemoteOptions) (v1alpha1.ZarfPackage, []string, error) {
```
AustinAbro321, line 40: Now that we merge values files from children we can have values files in the package but not set in the schema. We should add a function `pkgLayout.HasValuesSchema()` that checks the package and replace two instances of `pkgLayout.Pkg.Values.Schema != ""` currently in the codebase.

src/pkg/packager/layout/assemble.go
AustinAbro321: We should thread in definedPkg here so we can add  `|| len(importedSchemas) > 0` so that a skeleton importing a component with a values schema is also denied

src/pkg/value/schema.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	return out
+}
+
+// MergeSchemas merges child into parent with parent-wins semantics and returns a new map.
+// Neither parent nor child is modified. Rules:
+//   - "properties": recursively merged; parent wins on same key
+//   - "required": union of both arrays, deduplicated
+//   - all other keys: parent wins (child value used only when key absent from parent)
+func MergeSchemas(parent, child map[string]any) map[string]any {
```
AustinAbro321, line 90: What do you think about erroring here when schemas are not the same draft (I.E. `"$schema": "http://json-schema.org/draft-07/schema#"`)? If the child draft is not the same as the parent draft then the new schema may not be valid. Alternatively, we could require that everyone always use the latest draft?

src/pkg/value/schema.go
```diff
@@ -0,0 +1,145 @@
...
+			out[k] = cp
+		default:
+			out[k] = v
+		}
+	}
+	return out
+}
+
+// MergeSchemas merges child into parent with parent-wins semantics and returns a new map.
+// Neither parent nor child is modified. Rules:
```
AustinAbro321, line 84: There are instances were this is not true. I had an agent write unit tests to prove this
```go

func TestMergeSchemasDoesNotAliasInputs(t *testing.T) {
	t.Run("child-only top-level key is deep-copied", func(t *testing.T) {
		childElem := map[string]any{"type": "string"}
		parent := map[string]any{"type": "object"}
		child := map[string]any{"allOf": []any{childElem}}

		merged := MergeSchemas(parent, child)
		childElem["type"] = "mutated"

		allOf, ok := merged["allOf"].([]any)
		require.True(t, ok, "allOf should be a slice")
		elem, ok := allOf[0].(map[string]any)
		require.True(t, ok, "allOf[0] should be a map")
		assert.Equal(t, "string", elem["type"], "merged aliases child substructure stored by reference")
	})

	t.Run("child-only property is deep-copied", func(t *testing.T) {
		childProp := map[string]any{"type": "string"}
		parent := map[string]any{"properties": map[string]any{}}
		child := map[string]any{"properties": map[string]any{"extra": childProp}}

		merged := MergeSchemas(parent, child)
		childProp["type"] = "mutated"

		props, ok := merged["properties"].(map[string]any)
		require.True(t, ok, "properties should be a map")
		extra, ok := props["extra"].(map[string]any)
		require.True(t, ok, "extra should be a map")
		assert.Equal(t, "string", extra["type"], "merged aliases child property stored by reference")
	})

	t.Run("mutating result does not modify parent slice elements", func(t *testing.T) {
		parentElem := map[string]any{"type": "string"}
		parent := map[string]any{"allOf": []any{parentElem}}
		child := map[string]any{}

		merged := MergeSchemas(parent, child)
		allOf, ok := merged["allOf"].([]any)
		require.True(t, ok, "allOf should be a slice")
		elem, ok := allOf[0].(map[string]any)
		require.True(t, ok, "allOf[0] should be a map")
		elem["type"] = "mutated"

		assert.Equal(t, "string", parentElem["type"], "mutating result mutated parent: copyMap shallow-copies slices")
	})
}
```

src/pkg/packager/load/testdata/import/values/schema-parent-empty/expected-merged-schema.json
AustinAbro321, line 1: It seems none of the expected-merged-schema.json files are used

src/pkg/value/schema.go
```diff
@@ -0,0 +1,145 @@
...
+//   - "required": union of both arrays, deduplicated
+//   - all other keys: parent wins (child value used only when key absent from parent)
+func MergeSchemas(parent, child map[string]any) map[string]any {
+	result := copyMap(parent)
+	for key, childVal := range child {
+		switch key {
+		case "properties":
+			result["properties"] = mergeProperties(result["properties"], childVal)
+		case "required":
+			result["required"] = mergeRequired(result["required"], childVal)
```
AustinAbro321: This can add `required: "null"` to the jsonschema if the child or parent has an empty list has `"required": []`. This gives an invalid jsonschema. If we instead have merged required return `[]any` we can do this and avoid the issue. 
```suggestion
			if req := mergeRequired(result["required"], childVal); len(req) > 0 {
				result["required"] = req
			}
```

AustinAbro321 (changes requested):
Glad to see this moving along, poked at it and found some bugs / decision points

src/pkg/packager/load/testdata/import/values/schema-parent-empty/expected-merged-schema.json
brandtkeller: It was in the E2E test but I removed it to not pollute that test.

src/pkg/value/schema.go
brandtkeller: good catch!

src/pkg/value/schema.go
brandtkeller: Yeah that's a really good callout. I don't want us in the business of dictating if our underlying utility doesn't require it.

There is an interesting detail here in that technically we don't require a schema be explicitly defined. `gojsonschema` can handle absent `$schema`.

src/pkg/packager/layout/assemble.go
brandtkeller: agreed!

src/pkg/value/schema.go
brandtkeller: Good call - Updated to properly deepcopy recursively.

src/pkg/packager/layout/testdata/schema-merge/child-with-ref.schema.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "definitions": {
+    "name": { "type": "string", "minLength": 1 }
+  },
+  "properties": {
+    "appName": { "$ref": "#/definitions/name" }
```
AustinAbro321, line 8: I believe we should allow internal refs such as these.

src/pkg/value/schema.go
AustinAbro321: I'd rather require the draft be defined since if we move away from gojsonschema some day (seems likely considering the two latest drafts are not supported and the repo isn't updated) a future library might handle it differently

src/pkg/packager/load/testdata/import/values/schema-parent-empty/expected-merged-schema.json
AustinAbro321: I think we can delete the three expected-merged-schema.json files in that case

src/pkg/packager/layout/assemble.go
```diff
@@ -1031,36 +1028,120 @@ func mergeAndWriteValuesFile(ctx context.Context, files []string, packagePath, b
...
+			return nil, fmt.Errorf("reading %s schema: %w", label, err)
+		}
+		var s map[string]any
+		if err := json.Unmarshal(b, &s); err != nil {
+			return nil, fmt.Errorf("parsing %s schema: %w", label, err)
+		}
+		if err := value.CheckNoRefs(s); err != nil {
+			return nil, fmt.Errorf("%s schema %s: %w", label, relPath, err)
+		}
+		if err := value.ValidateSchemaDocument(s); err != nil {
```
AustinAbro321, line 1064: What would you think about also running this after the final merge is completed? Not strictly necessary, but in case we missed something and our merge logic can end up with a bad schema, it'd be better to fail fast.

AustinAbro321 (commented):
looks close, a few more comments

AustinAbro321:
Could you also mark this PR with a "!" and note the breaking changes in load in the description

src/pkg/packager/layout/testdata/schema-merge/child-with-ref.schema.json
brandtkeller: Fair! I also added some logic to handle definition merge so that the parent does not wipe out child definitions.

src/pkg/packager/layout/assemble.go
brandtkeller: Seems like a good catch-all for scenarios we are unaware of - and create time is certainly the better time for a failure - added.

src/pkg/value/schema.go
brandtkeller: Updated to disallow undefined and require matching

src/pkg/packager/load/testdata/import/values/schema-parent-empty/expected-merged-schema.json
brandtkeller: removed!

AustinAbro321 (approved):
LGTM, we should document the behavior (merge for "properties, "definitions", ...) and overwrite for others, but lets do that in a follow up

PR #4954, opened 2026-06-01
docs: version documentation site

### LINKED ISSUE #4056: Version Zarf website docs
We should version our Zarf docs, so that users still on older version can view the Zarf docs at a point in time. 

This is especially important as a v1.0.0 release for Zarf is being targeted. There will be breaking changes in v1.0.0 which may slow the time users take to upgrade to a newer version.

AustinAbro321:
## Description

Versions the Zarf website content. Show a new version for each of the last 10 minors

This also updated the starlight version so I could test that a new version wouldn't break older versions. 

Pros of this approach: 
- Simple, no additional framework or dependencies required
- Build everything with current versions of dependencies. 
  - This is possible because we are only switching out the mdx which should always work across astro versions. 
- We don't have to duplicate any content, it's all pulled from Git

Cons of this approach: 
- The javascript components are consistent throughout, only the content is changed, so components can be deleted or changed in breaking ways. 
  - I am okay with this tradeoff since components are generally simple, and we don't have many components

## Related Issue

Fixes #4056

## Checklist before merging


site/hack/build-versions.mjs
```diff
@@ -0,0 +1,211 @@
...
+  // Keep only the newest patch per minor.
+  const newestByMinor = new Map();
+  for (const tag of tags) {
+    if (!/^v?\d+\.\d+\.\d+$/.test(tag)) continue;
+    const minor = minorKey(tag);
+    const current = newestByMinor.get(minor);
+    if (!current || parseSemver(tag)[2] > parseSemver(current)[2]) {
+      newestByMinor.set(minor, tag);
+    }
+  }
```
soltysh, line 90: This entire block will return tags based on minor. But given that we're heading towards 1.0.0, and we'll most likely would want to make some of the older versions visible for a bit longer, wouldn't it be reasonable to explicitly put a limit per major version too? 

Example, with `0.81.0` being the last pre-1.0 version, and 1.0 being released for so long that we're currently at `1.10.1, I think it'd be reasonable to still show 0.81.0 for some time, and only show 3 (or other configurable number) under `1.x.x`, wdyt?

soltysh (commented):
One suggestion, but overall I like the approach

site/hack/build-versions.mjs
AustinAbro321: Yeah I like this that way we never have to move the version floor. I ended up changing it so we'll keep the latest 10 versions of the current major, and the latest 3 versions of the previous major

site/hack/copy-examples.js
```diff
@@ -6,19 +6,28 @@ import { fileURLToPath } from 'url';
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
-const examplesDir = path.join(__dirname, "../../examples");
-const dstDir = path.join(__dirname, "../src/content/docs/ref/Examples");
-
-async function preflight() {
+const repo = "https://github.com/zarf-dev/zarf";
+
+// Generate `ref/Examples` docs from a checkout's `examples/` directory. The
```
brandtkeller, line 11: I don't know that I would change the current stance but worth noting that this only works as far back as `v0.75.1` as that was when we started reading from the example readme's.

brandtkeller (approved):
This is really neat and effective. I left a comment for examples but I think it applies in the same for schema but wouldn't recommend changing anything.

PR #4957, opened 2026-06-03
feat(values)!: support state built-ins

### LINKED ISSUE #4226: Add a template.Objects key for Zarf builtins
Currently, Zarf's builtins from state are available via the {{ .Variables }} object, but this incidental from how we import _all_ variables and not ideal. Instead, it'd be better to have a dedicated key for these, either .Zarf, .Internal, .State, or some better alternative.

### Additional context
Similarly, there's some builtins that aren't included in this map, like registry username! This is a good opportunity to get the state builtins up to date.

brandtkeller:
## Description

Adds support for `{{ .State.<target>. }}` with an explicit field in the schema for components to opt-into sensitive state values.

**Breakng Change:** The actions `Run()` function signature has been modified. This only impact SDK users invoking that functionality. 

## Related Issue

Fixes #4226

## Checklist before merging

### DIFF
--- a/examples/values-templating/zarf.yaml
+++ b/examples/values-templating/zarf.yaml
@@ -78,5 +78,18 @@ components:
           - cmd: |
               echo "This {{ .wontBeProcessed }} stays as-is"
 
+  - name: state-templates
+    description: "Demonstrates .State fields in actions — public cluster state, no stateAccess declaration required"
+    required: true
+    actions:
+      onDeploy:
+        before:
+          - cmd: |
+              echo "Registry:      {{ .State.Registry.Address }}"
+              echo "Storage class: {{ .State.StorageClass }}"
+              echo "IP family:     {{ .State.IPFamily }}"
+              echo "Registry mode: {{ .State.Registry.Mode }}"
+            template: true
+
 documentation:
   readme: readme.md
--- a/examples/values-templating/zarf.yaml
+++ b/examples/values-templating/zarf.yaml
@@ -78,18 +78,5 @@ components:
           - cmd: |
               echo "This {{ .wontBeProcessed }} stays as-is"
 
-  - name: state-templates
-    description: "Demonstrates .State fields in actions — public cluster state, no stateAccess declaration required"
-    required: true
-    actions:
-      onDeploy:
-        before:
-          - cmd: |
-              echo "Registry:      {{ .State.Registry.Address }}"
-              echo "Storage class: {{ .State.StorageClass }}"
-              echo "IP family:     {{ .State.IPFamily }}"
-              echo "Registry mode: {{ .State.Registry.Mode }}"
-            template: true
-
 documentation:
   readme: readme.md
--- a/src/internal/template/template.go
+++ b/src/internal/template/template.go
@@ -109,9 +109,7 @@ func (o Objects) WithPackage(pkg v1alpha1.ZarfPackage) Objects {
 
 // WithState adds Zarf runtime state to the template Objects under the "State" key.
 // Non-sensitive fields (addresses, usernames, configuration) are always included.
-// Sensitive fields (passwords, tokens) are only included when allowSensitive is true —
-// accessing them without opt-in causes a template error (missingkey=error).
-// If s is nil, the Objects map is returned unchanged.
+// Sensitive fields (passwords, tokens) are only included when allowSensitive is true
 func (o Objects) WithState(s *state.State, allowSensitive bool) Objects {
 	if s == nil {
 		return o
--- a/src/internal/template/template.go
+++ b/src/internal/template/template.go
@@ -107,10 +107,22 @@ func (o Objects) WithPackage(pkg v1alpha1.ZarfPackage) Objects {
 	return o
 }
 
+// StateAccess bundles the runtime state with its access permissions for template rendering.
+// A zero value (nil State) is safe and causes WithState to be a no-op.
+type StateAccess struct {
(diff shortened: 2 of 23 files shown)


src/internal/template/template.go
```diff
@@ -105,6 +107,57 @@ func (o Objects) WithPackage(pkg v1alpha1.ZarfPackage) Objects {
...
+	if allowSensitive {
+		registry["PushPassword"] = s.RegistryInfo.PushPassword
+		registry["PullPassword"] = s.RegistryInfo.PullPassword
+		registry["Secret"] = s.RegistryInfo.Secret
+		git["PushPassword"] = s.GitServer.PushPassword
+		git["PullPassword"] = s.GitServer.PullPassword
+		artifact["PushToken"] = s.ArtifactServer.PushToken
+	}
+
+	o[objectKeyState] = map[string]any{
```
AustinAbro321: I believe we are missing each of these which will be required for the init package: 

 - ZARF_AGENT_CA — AgentTLS.CA
  - ZARF_AGENT_CRT — AgentTLS.Cert 
  - ZARF_AGENT_KEY — AgentTLS.Key
  - ZARF_INJECTOR_PAYLOAD_CONFIGMAPS 
  - ZARF_INJECTOR_SHASUM 
  - ZARF_HTPASSWD 
  - ZARF_SEED_REGISTRY 

One other note is that the for the current Zarf builtins some only exist when the component requires it, for instance, the agent certs are only available when the component is named "zarf-agent". I'm undecided if we should keep the old filtering, but wanted to call it out. We could force components to specifically request builtins, instead of a blanket bool.

src/pkg/packager/find_images.go
```diff
@@ -328,7 +328,7 @@ func findImages(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath strin
 			}
 		}
 		for _, manifest := range component.Manifests {
-			manifestResources, err := getTemplatedManifests(ctx, manifest, pkgPath.BaseDir, compBuildPath, variableConfig, vals, pkg)
+			manifestResources, err := getTemplatedManifests(ctx, manifest, pkgPath.BaseDir, compBuildPath, variableConfig, vals, pkg, nil, false)
```
AustinAbro321: This should send state (defined by state.Default()). I believe we should also send the component value for getTemplatedManifests so the package is either successful or breaks across commands.

src/internal/template/template.go
```diff
@@ -105,6 +107,57 @@ func (o Objects) WithPackage(pkg v1alpha1.ZarfPackage) Objects {
 	return o
 }
 
+// WithState adds Zarf runtime state to the template Objects under the "State" key.
+// Non-sensitive fields (addresses, usernames, configuration) are always included.
+// Sensitive fields (passwords, tokens) are only included when allowSensitive is true
+func (o Objects) WithState(s *state.State, allowSensitive bool) Objects {
```
AustinAbro321: nit: perhaps this should be a type, maybe `template.StateAccess` or `template.StateOptions`. We have a lot of functions that take both of these parameters as a group.

AustinAbro321 (changes requested):
Mostly clean implementation. A few things I noticed

src/pkg/packager/find_images.go
brandtkeller: Agreed

src/internal/template/template.go
brandtkeller: Updated - I think that is a good call as we explore the other comment (Above)

src/internal/template/template.go
brandtkeller: Let me think on this - I like the idea of granularity. Mty brain was in the boat of component versus package but I could definitely get on board with minimizing access to only those you need.

examples/values-templating/zarf.yaml
```diff
@@ -78,5 +78,18 @@ components:
           - cmd: |
               echo "This {{ .wontBeProcessed }} stays as-is"
 
+  - name: state-templates
```
AustinAbro321: nit: to keep our documentation brief I wouldn't add this here, someone viewing the example should infer that this will work

src/internal/template/template.go
```diff
@@ -105,6 +109,93 @@ func (o Objects) WithPackage(pkg v1alpha1.ZarfPackage) Objects {
...
+	git := map[string]any{
+		"Address":      s.GitServer.Address,
+		"PushUsername": s.GitServer.PushUsername,
+		"PullUsername": s.GitServer.PullUsername,
+		"IsInternal":   s.GitServer.IsInternal(),
+	}
+	artifact := map[string]any{
+		"Address":      s.ArtifactServer.Address,
+		"PushUsername": s.ArtifactServer.PushUsername,
+	}
```
AustinAbro321: I think there's a good chance we'll delete the artifact server before v1.0. IMO lets delete this for now, could be a signal if anyone asks about it, otherwise helps us avoid more breaking changes later

src/internal/template/template.go
AustinAbro321: I like the solution you landed on

examples/values-templating/zarf.yaml
brandtkeller: That's fair - removed

src/internal/template/template.go
brandtkeller: good call

AustinAbro321 (approved):

PR #4959, opened 2026-06-04
refactor: improve cosign signature/attestation lookup with oras

Racer159:
## Description

This improves the speed of cosign signature and attestation lookups using `zarf dev find-images` to encourage their use.

Testing against the [Neuvector UDS Package](https://github.com/uds-packages/neuvector) `registry1` variant more than halved the total `find-images` time:

image

This also aligns the authentication flow of this lookup with that of the image pull on create.

## Related Issue

Fixes #N/A

## Checklist before merging

### DIFF
--- a/src/pkg/images/common.go
+++ b/src/pkg/images/common.go
@@ -23,7 +23,9 @@ import (
 	"github.com/zarf-dev/zarf/src/pkg/logger"
 	"github.com/zarf-dev/zarf/src/pkg/state"
 	"oras.land/oras-go/v2/content"
+	orasRemote "oras.land/oras-go/v2/registry/remote"
 	"oras.land/oras-go/v2/registry/remote/auth"
+	"oras.land/oras-go/v2/registry/remote/credentials"
 	"oras.land/oras-go/v2/registry/remote/retry"
 )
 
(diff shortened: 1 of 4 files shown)


src/pkg/images/common.go
```diff
@@ -187,6 +189,48 @@ func saveIndexToOCILayout(dir string, idx ocispec.Index) error {
...
+
+	client := &auth.Client{
+		Client: &http.Client{
+			Transport: transport,
+		},
+		Cache:      auth.NewCache(),
+		Credential: credentials.Credential(credStore),
+	}
+
+	logger.From(ctx).Debug("gathering credentials from default Docker config file", "credentialsConfigured", authConfigured)
```
brandtkeller: Should this log be after we check `IsAuthConfigured()` such that it doesn't always report false?

src/pkg/images/common.go
```diff
@@ -187,6 +189,48 @@ func saveIndexToOCILayout(dir string, idx ocispec.Index) error {
...
+			registry, err := orasRemote.NewRegistry(host)
+			if err != nil {
+				return nil, false, fmt.Errorf("failed to create registry: %w", err)
+			}
+			registry.Client = client
+			_ = registry.Ping(ctx) //nolint: errcheck
+		}
+	}
+
+	return client, authConfigured, nil
```
brandtkeller: authConfigured is never set and would always return false - but the updated call sets also don't read the value.

brandtkeller (changes requested):
Minor requests otherwise this is neat to see.

src/pkg/images/common.go
Racer159: blep yep missed that - will fix

src/pkg/images/common.go
Racer159: yep moved things around and didn't double check this.

Racer159: fixed - that's what I get for rushing!

brandtkeller (approved):
lgtm. clean and effective - technically a breaking change (public function signature modification) but I believe we're intending to have the policy list `utils` as not within the boundary. 

I haven't seen any requests for this logic to support non-https registries and the current approach is still an improvement over what was here previously. That was the only thing that stood out between `pull.go` and the implementation here.

AustinAbro321 (approved):
LGTM, Thanks! I would love for find-images to rely entirely on oras-go instead of Crane. This is a good improvement for now 

 I am renaming this from chore-> refactor

PR #4961, opened 2026-06-05
feat(schema)!: zarf dev schema-generate from existing values

### LINKED ISSUE #4918: Generate Schema JSON from Zarf Values files

I would like a way to generate a JSON schema for the values within a Zarf package from the values files that are wired into the package once composed.  This will help deterministically update the schema when values change and make it easier to have a starting point to encourage the use of schemas when adopting Zarf Values as a feature.

- **Given** I have a Zarf package with `values` in it
> ```yaml
> values:
>   files:
>     - zarf-values.yaml
> ```
- **When** I run a new command `zarf dev generate-schema`
- **Then** A new schema file `zarf-values.schema.json` is created
- **And** That JSON schema follows the default values provided in a package to determine objects and types.

We could create a more complex implementation that is able to include schemas for common types that people may include but not fully flesh out (i.e. Kubernetes resources) but this could be confusing to implement and could introduce risk of becoming too strict as the K8s API changes into the future.

### Additional context

This came up in discussions of adopting Values in a standardized way across many packages in our [`uds-common`](https://github.com/defenseunicorns/uds-common/pull/679#discussion_r3277070726)

Racer159:
## Description

This adds a `zarf dev generate-schema` command to load the Zarf values defaults and then generate an inferred schema from those values.  It does this by:

1. Loading and merging the Zarf `values.files` to create an initial default list
2. Looping through the `components[x].charts` `values.sourcePath` and `values.targetPath` to fill in any missing values
   - These can either be from Zarf's `valuesFiles` or the Helm Chart's `values.yaml`
3. Deterministically map these values to inferred types in a JSON Schema and save the schema

This also will load an existing schema (if one exists) and reconcile that schema with the inferred types in case the Zarf package creator already has a schema and has added additional keys (such as `description`, `required`, etc).

> [!CAUTION]
> **BREAKING CHANGE**
> This includes a technically breaking change to how values merge together.  Instead of strictly replacing keys Values.Set now merges them so users may see more values than they would expect if they had already gotten used to how Zarf Values worked.  Likely though this new pattern will make more sense to Zarf users though as it is more consistent with how other primitives (like `valuesFiles`) merge in Zarf.

## Related Issue

Fixes #4918

## Checklist before merging


src/cmd/dev.go
```diff
@@ -73,11 +75,150 @@ func newDevCommand() *cobra.Command {
...
+					}
+				}
+			}
+		}
+	}
+
+	// Step 3: Generate JSON schema from the final map and save it
+	schema := value.GenerateJSONSchema(zarfValues)
+
+	outputFileName := "values.schema.json"
```
brandtkeller: Should we also consider the specific `defined.Pkg.Values.Schema` path?

src/cmd/dev.go
```diff
@@ -73,11 +75,150 @@ func newDevCommand() *cobra.Command {
...
+	loadOpts := load.DefinitionOptions{
+		Flavor:               o.flavor,
+		SetVariables:         o.setPkgTmpl,
+		IsInteractive:        true,
+		SkipVersionCheck:     true,
+		SkipSchemaValidation: true,
+		CachePath:            cachePath,
+		RemoteOptions:        defaultRemoteOptions(),
+	}
+	defined, err := load.PackageDefinition(ctx, basePath, loadOpts)
```
brandtkeller, line 134: `defined` is a `DefinedPackage`

```
type DefinedPackage struct {
	Pkg             v1alpha1.ZarfPackage
	ImportedSchemas []string
}
```

These `ImportedSchemas` likely need to be merged with the specified schema (or created if not specified) prior to the final resolve workflow.

brandtkeller (changes requested):
initial thoughts around the shape of existing schemas - review WIP

src/cmd/dev.go
Racer159: Yes we likely should - I'll update to do that (and fallback if there isn't one)

src/cmd/dev.go
Racer159: I didn't do it but would we want to merge the helm chart schemas in too? I think we would still need the final reconcile at the end based on what we found as "truth" - it is available in helmChart.schema and for those sections will become an enforcement mechanism (so likely best to reencode that in the Zarf schema to fail early?)

src/pkg/packager/load/load.go
```diff
@@ -33,6 +33,8 @@ type DefinitionOptions struct {
 	// SkipRequiredValues ignores values schema validation errors when a "required" field is empty. Used when a package
 	// value should be supplied at deploy-time and doesn't have a default set in the package values.
 	SkipRequiredValues bool
+	// SkipSchemaValidation skips schema validation for the package values entirely.
+	SkipSchemaValidation bool
```
AustinAbro321:
```suggestion
	SkipValuesSchemaValidation bool
```
The package schema is also validated during load so we should specify

src/cmd/dev.go
Racer159: done.

src/cmd/dev.go
Racer159: I thought about the helm chart thing and decided against it - there is a decent amount of complexity there that would warrant a deeper look.

With the `values.yaml` controlling just the `type` and existence of a key right now there is a pretty clear story that that is "ground truth" information that should replace what the Zarf schema says.  With a Helm chart schema, some things (like validation keys) are imperative and would ideally be enforced in the Zarf schema since they will be enforced on a Helm install later, but other keys (like `description`) are more nebulous - it would be nice to pull in description from the chart but you might not always want to take the description from the chart if it didn't explain things well enough and that creates some tension worth designing through before doing anything there I think.

src/cmd/dev.go
```diff
@@ -73,11 +75,164 @@ func newDevCommand() *cobra.Command {
...
+		}
+	}
+
+	// Step 3: Generate JSON generatedSchema from the final map
+	generatedSchema := value.GenerateJSONSchema(zarfValues)
+
+	// Step 4: Merge and reconcile any existing schema
+	existingSchema, mergeErr := value.MergeSchemaFiles(defined.Pkg.Values.Schema, defined.ImportedSchemas, basePath)
+	if mergeErr != nil {
+		l.Warn("unable to merge imported schemas for schema generation", "error", mergeErr.Error())
```
AustinAbro321: This should be an error. I'd recommend changing the earlier warns to errors too. Technically we could end up with a fine schema if we skip a chart, but the user will want to re-run this command after those charts are fixed, so I'd rather it be clear that an error occurred

src/pkg/value/generate.go
```diff
@@ -0,0 +1,124 @@
...
+	if !ok {
+		existingProps = make(map[string]any)
+		existing["properties"] = existingProps
+	}
+
+	for key := range existingProps {
+		if _, found := inferredProps[key]; !found {
+			delete(existingProps, key)
+		}
+	}
```
AustinAbro321: I believe we should not delete properties not found in the values.yaml documents. A user might intentionally not include values in the values.yaml so that the deployer is forced to explicitly declare them. This would then delete those properties. I think a flag like `--delete-not-found` would be a reasonable way to add this back

src/cmd/dev.go
```diff
@@ -73,11 +75,164 @@ func newDevCommand() *cobra.Command {
...
+	outputFileName := "values.schema.json"
+	if defined.Pkg.Values.Schema != "" {
+		if !filepath.IsAbs(defined.Pkg.Values.Schema) {
+			outputFileName = filepath.Join(basePath, defined.Pkg.Values.Schema)
+		} else {
+			outputFileName = defined.Pkg.Values.Schema
+		}
+	}
+
+	err = os.WriteFile(outputFileName, b, helpers.ReadAllWriteUser)
```
AustinAbro321: Lets instead output to stdout, this is consistent with `zarf dev find-images` and future planned commands such as `zarf dev convert`

src/cmd/dev.go
```diff
@@ -73,11 +75,164 @@ func newDevCommand() *cobra.Command {
...
+
+			err := layout.PackageChart(ctx, chart, basePath, chartPath, valuesFilePath, cachePath, defaultRemoteOptions())
+			if err != nil {
+				l.Warn("unable to package chart for schema generation", "chart", chart.Name, "error", err.Error())
+				continue
+			}
+
+			helmChart, valuesFilesValues, err := helm.LoadChartData(chart, chartPath, valuesFilePath, nil)
+			if err != nil {
+				l.Warn("unable to to load default values for chart", "chart", chart.Name, "error", err.Error())
```
AustinAbro321:
```suggestion
				l.Warn("unable to load default values for chart", "chart", chart.Name, "error", err.Error())
```

src/pkg/value/generate.go
```diff
@@ -0,0 +1,124 @@
...
+	for k, v := range vals {
+		props[k] = inferSchemaType(v)
+	}
+
+	return schema
+}
+
+// ReconcileJSONSchema updates structural fields in an existing schema from inferred values.
+// Non-structural fields (description/enum/required/etc.) are preserved by default.
+func ReconcileJSONSchema(existing, inferred map[string]any) map[string]any {
```
AustinAbro321: I would add a copyMap here on the existing so we avoid mutating the maps in public function

AustinAbro321 (changes requested):
Nice work, a few comments and requests

src/pkg/value/generate.go
Racer159: done

src/cmd/dev.go
Racer159: done

src/cmd/dev.go
Racer159: done (added `-u` like `zarf dev find-images` does)

src/pkg/value/generate.go
Racer159: added `--delete-not-found` and made your suggestion default

src/cmd/dev.go
Racer159: done -made them all errors

src/pkg/packager/load/load.go
Racer159: done

AustinAbro321 (dismissed):
LGTM

examples/values-templating/values/values.schema.json
```diff
@@ -1,115 +1,139 @@
 {
```
brandtkeller, line 1: broader question which I wanted to challenge my initial testing for accuracy - How is this command intended to be used to iteratively upgrade package value schemas as the package evolves? 

My initial expectation was that you would execute as expected and then redirect output to overwrite the existing schema file - this produces an error currently (trivial to resolve). 

Want to ensure I understand appropriately in order to have a mental model for future docs.

brandtkeller: ah the `--update` flag - I should read more carefully. Odd that this responds differently but it does have an expected use via the flag.

src/cmd/dev.go
```diff
@@ -73,11 +75,168 @@ func newDevCommand() *cobra.Command {
...
+		outputFileName := "values.schema.json"
+		if defined.Pkg.Values.Schema != "" {
+			if !filepath.IsAbs(defined.Pkg.Values.Schema) {
+				outputFileName = filepath.Join(basePath, defined.Pkg.Values.Schema)
+			} else {
+				outputFileName = defined.Pkg.Values.Schema
+			}
+		}
+
+		err = os.WriteFile(outputFileName, b, helpers.ReadAllWriteUser)
```
brandtkeller, line 234: Nit: what is the expected standard convention for `zarf dev generate*` commands? `zarf dev generate` uses CWD as it has nothing else to build upon - but the other commands vary. I wonder if this should just anchor on basepath so that the "-u" case without an existing file writes to the package directory?

brandtkeller (dismissed):
I generally only found nits - one that I wouldn't necessarily block on but we should consider as this is user facing.

Other point - I wonder if we should cater language of the "values" feature such that the values at the root of the package are "Package Values" and the values elsewhere are values for an artifact (really just chart values currently). `Zarf Values` feels like it lacks a target (which is actually the package versus other Zarf _things_). Something we should dissect further.

src/cmd/dev.go
Racer159: Hmm - I keep flip flopping thinking about it - just typing through it, the current two commands `generate` and `generate-config` don't reference an existing `zarf.yaml` so they just plop the file into the CWD - `zarf dev generate-schema` is actually kind of a hybrid of those generate commands and `zarf dev find-images` - but `find-images` won't create any new file it will just update the `zarf.yaml` if you request it to.  Maybe `-u` for the nonexist path should drop it next to the `zarf.yaml` but _also_ update the `zarf.yaml` to reference the new schema?  That may be more in line with what `zarf dev find-images -u` does and would be more "complete" from a wiring perspective...

AustinAbro321: Yeah, I can see the argument either way, but I think it's more consistent to have it go to the package directory.

Racer159: now it will take the base path and update the zarf.yaml under `-u`

AustinAbro321 (approved):

brandtkeller (approved):
lgtm

PR #4966, opened 2026-06-08
fix(agent): properly pass through context

AustinAbro321:
## Description

This changes the agent admission functions to use the context from the http request, rather than the context from the Zarf CLI 

## Checklist before merging


soltysh (approved):
lgtm

Issue #4680, opened 2026-03-04
Support `values.yaml.gotmpl` in charts valuesFiles


ion-elgreco:
### Is your feature request related to a problem? Please describe.
We can template things from `values.files` into charts[0].values with `sourcePath` and `targetPath`. However this makes everything quite verbose. I would rather want to provides a valuesFiles like this:

```yaml
components:
  - name: helm-charts
    charts:
      - name: gateway-helm
        version: 1.7.0
        url: oci://docker.io/envoyproxy
        namespace: envoy
        valuesFiles:
          - ./gateway/values.yaml.gotmpl
```

### Describe the behavior you'd like
`valuesFiles` in charts should support gotmpl versions of values.yaml so that we can simply have these `values.yaml.gotmpl`, and these values will come from `values.files` in the root

```yaml
global:
  images:
    ratelimit:
      image: "{{ .Values.containerRegistry }}/{{ .Values.envoy.ratelimit.repository }}/{{ .Values.envoy.ratelimit.tag }}"

deployment:
  envoyGateway:
    image:
      repository: "{{ .Values.containerRegistry }}/{{ .Values.envoy.repository }}"
      tag:  "{{ .Values.envoy.tag }}"
```

brandtkeller:
Closing as completed in #4967 with the `templateValuesFiles` addition.

PR #4967, opened 2026-06-08
feat(values): support for go templating in chart valuesFIles

brandtkeller:
## Description

Support resolving go templates in chart `valuesFiles` as is supported for variables for parity. This implements an additional field in the schema for `templatedValuesFiles` for v1alpha1. For v1beta1 we can update the values files to a list of objects with a template field. 

This takes the strictly declarative stance of requiring definition of these files in a list of files to be templated. It ensures we do not break anyone using go templating in values files today while also not creating any new systems for templating. 

If this is selected - I would recommend we close 4680 as not-planned. 

## Alternatives

As linked in #4680 - we could support extension detection. I chose not to go this route as it creates a new implicit system of how templating occurs that is not supported by other artifacts. Similarly I had concerns that this may create future issues around whether zarf handles artifacts in implicit ways that cannot be reversed - IE after this change someone could not transfer files with these extensions without opting-out in some way. 

## Related Issue

Relates to #4680

## Checklist before merging


src/internal/packager/helm/common.go
AustinAbro321, line 1: What would you think of adding a type and function like so: 
```go
type ChartValuesFile struct {
	Path     string
	Template bool
}

func GetChartValuesFiles(valuesFiles, templatedValuesFiles []string) []ChartValuesFile {
	chartValuesFiles := []ChartValuesFile{}
	for _, v := range valuesFiles {
		chartValuesFiles = append(chartValuesFiles, ChartValuesFile{Path: v})
	}
	for _, v := range templatedValuesFiles {
		chartValuesFiles = append(chartValuesFiles, ChartValuesFile{Path: v, Template: true})
	}
	return chartValuesFiles
}
```
I believe this would let us follow what we expect to be in the v1beta1 format, while removing duplication and the  `StandardTemplatedValuesName` concept.

AustinAbro321 (commented):
Overall looks good, a comment on how we might simplify this code for the future

src/internal/packager/helm/common.go
brandtkeller: This is clean - I like it.

AustinAbro321 (approved):

PR #4968, opened 2026-06-08
feat(values): fail early during deploy if values are not defined

### LINKED ISSUE #4693: Zarf should fail early if there are undefined values necessary for a deployment
During `zarf package deploy` if there are values that are not defined, for instance, during an on-deploy-after action, the package deployment fail. However, this will be after the component has already been successfully deployed. Leaving users in an awkward position. Zarf should instead check that all necessary values for the package deployment exist.  

Another alternative would be to stop requiring values be defined, and instead default to an empty string when they are not defined.

AustinAbro321:
## Description

This adds a pre-flight check to deploy to fail if required values are not defined. 

## Related Issue

Fixes #4693

## Checklist before merging


src/pkg/packager/values_preflight.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		err = errors.Join(err, os.RemoveAll(tmpDir))
+	}()
+
+	var sources []templateSource
+	if hasManifests {
+		manifestDir, err := pkgLayout.GetComponentDir(ctx, tmpDir, component.Name, layout.ManifestsComponentDir)
```
brandtkeller, line 199: Not blocking - wanted to note that an underlying detail here is that we now call `GetComponentDir()` multiple times for multiple targets across this preflight check and deploy. 

Concern is probably pretty minimal but It does stand out as a "smell" for optimizing.

src/pkg/packager/values_preflight.go
```diff
@@ -0,0 +1,257 @@
...
+	setValueKeys [][]string
+	setValueRoot bool
+}
+
+func buildDefinedValues(components []v1alpha1.ZarfComponent, vals value.Values) definedValues {
+	d := definedValues{vals: vals}
+	if d.vals == nil {
+		d.vals = value.Values{}
+	}
+	for _, component := range components {
```
brandtkeller: This iteration doesn't handle awareness around what `SetValues` are made available _when_ correct? So based on permutations of component filtering - an earlier component that requires a key that a later component happens to fill would be valid? 

Unsure on the restructuring required to solve this - but it came across my head while running through the code.

brandtkeller (commented):
Non-blocking observations - I think this is clean and entirely ready to ship but I had a few thoughts before stamping lgtm.

src/pkg/packager/values_preflight.go
AustinAbro321: Fixed this

src/pkg/packager/values_preflight.go
AustinAbro321: I think optimizing this would be complicated. Right now we get the directory for each component for  manifests and files assuming that at least one of the manifests and files are templatable. We could perhaps `getComponentDir` for all of these before we call this function, but this will get hard to track. Perhaps we could instead not delete the temporary directory and have packageLayout "cache" when the same directory is pulled twice, though this would rely on us not altering the file content.

brandtkeller (approved):

Issue #4878, opened 2026-05-06
Allow static Zarf values which cannot be changed at deploy time


AustinAbro321:
### Is your feature request related to a problem? Please describe.
As we evaluate the use cases for Zarf values, I've realized there are cases where a package creator wants to set values that cannot by overwritten at deploy time. An instance of this can be seen in the [experimental init package PR](https://github.com/zarf-dev/zarf/pull/4865). This PR uses the file `shared-template-values.yaml` to template the images in the init package AND provide the image references to the respective Helm charts at deploy time. The images will be locked in at create time, but the values are still overwriteable. 

Another use case is security sensitive values that package authors may not want to be altered.

If the deployer triers to set these values we should error before the deploy starts. 

This would be accomplished by a new field for "constant" values. The field could be could be called `staticFiles`, though I am open to other naming. 

### Describe the behavior you'd like

- **Given** I have a values block in my zarf.yaml with the new key `staticFiles` and a values.yaml that declares the key `.registry.image`  
```yaml
values:
  staticFiles:
    - values.yaml
```
- **When** I create and deploy the package with `zarf package deploy <my-package> --set-values=.registry.image=my-image`
- **Then** I get an error that the value `.registry.image` cannot be altered at deploy time. 

### Describe alternatives you've considered
You could not expose certain Zarf values to Helm chart values to keep security controls tight, however this does not work for manifests or actions. 

### Additional context
I believe this could be implemented after beta. 

This could get complicated with imports. For instance, if a child component declares static values that a parent component allows to be changeable, who would win? I think we would also favor static values, even if it came from a child. This would require documentation.

Racer159:
cc @mjnagel on this as well - I think we may want something like this or a `negative` pattern on values as well - going through a relatively simple but real example with [UDS Core Logging ](https://docs.defenseunicorns.com/core/concepts/core-features/logging/) exposing every value except those related to images or pod security gets loooooong:

```yaml
kind: ZarfPackageConfig
metadata:
  name: core-logging
  description: UDS Core (Logging)
  version: 1.3.1
  architecture: amd64
  authors: "Defense Unicorns - Product"
  aggregateChecksum: a617228adfa519aa50b6847d14bc781d75113e27c0fa67abe056e8ec5f0300a6
  annotations:
    dev.uds.categories: UDS-Core,Logging
    dev.uds.keywords: uds,logging,loki,vector
    dev.uds.tagline: Enables backend log storage and log shipping capabilities
[...]

PR #4974, opened 2026-06-10
feat(values): chart values exclude path

AustinAbro321:
## Description

This adds a new list `excludePath` to chartValues allowing creators to block certain fields that may be hard coded into the package for instance an image reference, or are security sensitive

## Related Issue

Relates to #4878 

## Checklist before merging

### DIFF
--- a/examples/values-templating/charts/example-chart/values.yaml
+++ b/examples/values-templating/charts/example-chart/values.yaml
@@ -3,9 +3,9 @@ appName: "default-app"
 replicaCount: 1
 
 image:
-  repository: nginx
-  tag: "latest"
-  pullPolicy: IfNotPresent
+  reference: "docker.io/library/nginx:1.29.2"
+  pullPolicy: "always"
+  name: IfNotPresent
 
 config:
   database:
--- a/examples/values-templating/values/values.yaml
+++ b/examples/values-templating/values/values.yaml
@@ -38,8 +38,8 @@ app:
   # Helm chart overrides
   name: "helm-overrides-app"  # Maps to .appName in helm chart
   image:
-    pullPolicy: "always"
-    name: IfNotPresent
+    pullPolicy: "Always"
+    name: nginx
 
 # Database configuration for helm chart
 database:
--- a/examples/values-templating/values/values.yaml
+++ b/examples/values-templating/values/values.yaml
@@ -38,10 +38,10 @@ app:
   # Helm chart overrides
   name: "helm-overrides-app"  # Maps to .appName in helm chart
   image:
-    tag: "v2.1.0"             # Maps to .image.tag in helm chart
+    pullPolicy: "always"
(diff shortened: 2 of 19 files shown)


examples/values-templating/values/values.yaml
```diff
@@ -38,10 +38,10 @@ app:
   # Helm chart overrides
   name: "helm-overrides-app"  # Maps to .appName in helm chart
   image:
-    tag: "v2.1.0"             # Maps to .image.tag in helm chart
+    pullPolicy: "Always"
+    name: nginx
```
brandtkeller, line 42: nit: if we are going to have the exclude example - should we add the value somewhere to represent that it has no effect? Or are we expecting that to be set elsewhere?

src/pkg/packager/packager.go
```diff
@@ -79,8 +79,20 @@ func generateValuesOverrides(_ context.Context, chart v1alpha1.ZarfChart, compon
 			return nil, fmt.Errorf("targetPath \"%s\" must start with a dot", chartValue.TargetPath)
 		}
 
+		// Drop any excluded sub-paths before extracting. Work on a copy so the
+		// shared source values are not mutated for other charts.
+		sourceValues := opts.values
```
brandtkeller, line 84: non-blocking - this makes me wonder whether there is an inherent reason we're not working on a copy for either path...

examples/values-templating/charts/example-chart/values.yaml
```diff
@@ -3,9 +3,9 @@ appName: "default-app"
 replicaCount: 1
 
 image:
-  repository: nginx
-  tag: "latest"
-  pullPolicy: IfNotPresent
+  reference: "docker.io/library/nginx:1.29.2"
+  pullPolicy: "always"
```
brandtkeller, line 7: nit: this isn't a valid pull policy but it's entirely cosmetic as they are not used in any resources.

brandtkeller (approved):
approving with a few minor thoughts - well executed.

examples/values-templating/values/values.yaml
AustinAbro321: I went back in forth on this, but it would be somewhat contrived to set it on create, the real case we're blocking against is the value being set on deploy. 

Maybe a line in the readme, or we change the example to have values it sets on deploy.

Either way lets clean up the example in a future PR.

src/pkg/packager/packager.go
AustinAbro321: We just didn't implement deep copy logic before, but I agree that'd be a good refactor

examples/values-templating/charts/example-chart/values.yaml
AustinAbro321: Good catch, lets clean up the example in a future pr

PR #4976, opened 2026-06-10
feat: add registry logout function

### LINKED ISSUE #4975: Add a registry logout functionality
We have automated some tasks using scripts, including logging in to and interacting with an external registry.
The `zarf tools registry login` creates an entry with the authentication information in `~/.docker/config.json`

After the tasks have completed, we would like to perform a logout so that the credentials are removed from `~/.docker/config.json`, similar to `docker registry logout registry.example.com`

Add a `zarf tools registry logout registry.example.com` function, which removes the authentication information from `~/.docker/config.json`

sepauli:
## Description

This PR adds the command:

```
zarf tools registry logout registry.example.com
```

which removes the corresponding credentials from `~/.docker/config.json`

## Related Issue

Fixes #4975

## Checklist before merging


src/cmd/crane.go
```diff
@@ -591,3 +592,31 @@ func zarfCraneInternalWrapper(commandToWrap func(*[]crane.Option) *cobra.Command
...
+	originalRunFn func(cmd *cobra.Command, args []string) error
+}
+
+func newRegistryLogoutCommand() *cobra.Command {
+	o := &registryLogoutOptions{}
+	craneCmd := craneCmd.NewCmdAuthLogout()
+	o.originalRunFn = craneCmd.RunE
+	craneCmd.RunE = o.run
+
+	craneCmd.Args = nil
```
brandtkeller: Minor: we lose fidelity of the original command argument validation `cobra.ExactArgs(1)` which allows us to run with multiple args without any error - could be misinterpreted. 

Do we need to set this to `nil`?

brandtkeller (changes requested):
One minor go-back otherwise this is ready.

src/cmd/crane.go
sepauli: i removed the line. We do not need this to set this to nil. It was a small mistake from my side

brandtkeller: No worries!

brandtkeller (dismissed):
lgtm. This is an entirely valid update - I will mention one item of recent discussion though; we are looking to transition from [crane to oras-go](https://github.com/zarf-dev/zarf/issues/5003). 

This update helps set the precedent on this having parity in the future when this work starts - but if you'd like to help we do intend to begin working towards this direction.

soltysh (dismissed):
lgtm

soltysh:
@sepauli after we landed #5032, you'll need to regenerate docs.

soltysh (approved):
lgtm

brandtkeller (approved):

PR #4980, opened 2026-06-11
feat(agent): add readiness probe

AustinAbro321:
## Description

I've seen the webhook timeout https://github.com/zarf-dev/zarf/actions/runs/27347872505/job/80801361476 recently in CI. It seems like it might be that the webhook has so little CPU that it's not able to complete the request in a normal amount of time. I'm not sure if the readinessprobe will help, but I don't think it will hurt

## Checklist before merging

### DIFF
--- a/packages/zarf-agent/chart/templates/deployment.yaml
+++ b/packages/zarf-agent/chart/templates/deployment.yaml
@@ -38,6 +38,14 @@ spec:
               path: /healthz
               port: 8443
               scheme: HTTPS
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: 8443
+              scheme: HTTPS
+            periodSeconds: 3
+            timeoutSeconds: 2
+            failureThreshold: 2
           ports:
             - containerPort: 8443
           securityContext:
--- a/packages/zarf-agent/chart/templates/webhook.yaml
+++ b/packages/zarf-agent/chart/templates/webhook.yaml
@@ -42,6 +42,7 @@ webhooks:
       - "v1"
       - "v1beta1"
     sideEffects: None
+    timeoutSeconds: 20
   - name: agent-flux-ocirepo.zarf.dev
     namespaceSelector:
       matchExpressions:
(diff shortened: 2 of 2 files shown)


packages/zarf-agent/chart/templates/webhook.yaml
```diff
@@ -42,6 +42,7 @@ webhooks:
       - "v1"
       - "v1beta1"
     sideEffects: None
+    timeoutSeconds: 20
```
soltysh: This is doubling the timeout from the default 10. Is it really something that we should be doing? Ideally, webhooks should be fast, b/c every slow webhooks significantly affects how responsive the cluster feels.

packages/zarf-agent/chart/templates/deployment.yaml
```diff
@@ -38,6 +38,14 @@ spec:
               path: /healthz
               port: 8443
               scheme: HTTPS
+          readinessProbe:
```
soltysh, line 41: Readiness probe is always a good idea to have. So 👍 this.

soltysh (commented):
From that failed CI run do we have an explicit information that the webhooks was that slow that we need to double its timeout? The readiness probe is always a good idea, we can also consider startup probe, if we see that the webhook startup is slow, but I don't think so.

packages/zarf-agent/chart/templates/webhook.yaml
AustinAbro321: Yeah it's hard to tell what went wrong in that job, I can't imagine a single namespace call should ever take 10+ seconds (though it is the second time I've seen that flake this week). 

Maybe since this test is running immediately after the agent is restarted there is some bugginess with the readiness probe. Alternatively, maybe we should be raising the cpu requests? Right now they default to only 100m. I've seen github runners have low cpu capacity before.

https://github.com/zarf-dev/zarf/blob/main/packages/zarf-agent/chart/values.yaml#L33-L36

AustinAbro321: I got rid of the timeout changes, lets add the readiness probe for now and continue to watch for flakes

soltysh (approved):
lgtm

PR #4981, opened 2026-06-11
feat(values)!: expose pkg built in

AustinAbro321:
## Description

This exposes a pkg built-in with the entire Zarf package object. This removes the `.Metadata` and `.Build` built-ins, in favor of `{{ .Pkg.Metadata }}` instead of `{{ .Metadata }}`. Because these features were never documented, and values is an alpha feature behind a feature flag, I believe this breaking change is okay.

I did not add {{ .Pkg }} to actions, {{ .Metadata }} and {{ .Build }} are also not available in actions, I believe this can be a follow up, I want us to consider which actions these should apply to. 

Relates to #4878

## Checklist before merging

### DIFF
--- a/src/internal/template/template.go
+++ b/src/internal/template/template.go
@@ -41,6 +41,7 @@ const (
 	objectKeyValues    = "Values"
 	objectKeyMetadata  = "Metadata"
 	objectKeyBuild     = "Build"
+	objectKeyPackage   = "Pkg"
 	objectKeyConstants = "Constants"
 	objectKeyVariables = "Variables"
 	objectKeyState     = "State"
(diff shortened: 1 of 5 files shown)


src/internal/template/template.go
```diff
@@ -41,6 +41,7 @@ const (
 	objectKeyValues    = "Values"
 	objectKeyMetadata  = "Metadata"
 	objectKeyBuild     = "Build"
+	objectKeyPackage   = "Pkg"
```
soltysh, line 42: With this in, should we start deprecating `.Metadata` and `.Build`?

soltysh (dismissed):
lgtm

brandtkeller (approved):
lgtm

src/internal/template/template.go
AustinAbro321: I ended up just deleting this since they never documented and values is an alpha feature

PR #4982, opened 2026-06-11
fix(docs): render markdown tables

AustinAbro321:
## Description

Markdown tables are not getting rendered since we updated from astro 5 to astro 6. We were using a deprecated field which astro is going to release in the next major, this updates us to use the non-deprecated field. 

## Checklist before merging


soltysh (approved):
lgtm

PR #4987, opened 2026-06-15
fix(pull): always pull values.yaml and values.schema.json

AustinAbro321:
## Description

Values layers are never getting pulled, so pulling a package with values is broken

There is some consideration to make a "Values" layer, but because they will almost always be a few kbs, or maybe a 1mb in a rare case, I opted to keep them as Metadata layers so the API would be more convenient to use. 

## Checklist before merging


brandtkeller (approved):
lgtm

Issue #3842, opened 2025-05-22
Enhancing CVE Identification & Reporting Process


brandtkeller:
## Background

Currently, our CVE triage relies on dependabot scans and manual reviews. This approach can lead to delays in vulnerability awareness, inconsistent reporting formats, and gaps in downstream processes (e.g., generating VEX data). To elevate our security posture and streamline vulnerability management, I propose an automated, nightly SBOM-based scanning & reporting pipeline on `main`.

---

## Proposed Workflow

1. **Nightly SBOM Generation & Scanning**

   * Use syft to generate an SPDX SBOM for the `main` branch every night.
   * Feed the resulting SBOM into grype, which produces:

     * **SARIF** output (for GitHub’s Code Scanning UI)
     * A machine-readable JSON report of discovered CVEs

2. **Locally Reproducible**

   * Ensure all steps can be reproduced locally - aiding in the review and mitigation process. 
   * This should likely use a new / existing `make` target(s)

3. **Artifact Upload**

   * Upload both the SBOM and the rendered CVE report as workflow artifacts for audit and offline review by any consumer without tooling available.

4. **High-Severity Alerting**

   * If any CVE with **severity ≥ High** is detected, fail the job (or optionally just raise a warning) and trigger a GitHub issue (via \[actions/create-issue]) tagging `@zarf-maintainers` (and maybe optionally a future`@security-team`).

---

## Example GitHub Actions Snippet

```yaml
# .github/workflows/cve-report.yml
name: "Nightly SBOM & CVE Report"

on:
  schedule:
    - cron: '0 2 * * *'  # Every day at 02:00 UTC
  workflow_dispatch:

jobs:
  sbom-and-scan:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout main
        uses: actions/checkout@v3
        with:
          ref: main

      - name: Install Syft
        run: |
          curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh \
            | sh -s -- -b /usr/local/bin

      - name: Install Grype
        run: |
          curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh \
            | sh -s -- -b /usr/local/bin

      - name: Generate SBOM with Syft
        id: sbom
        run: |
         make sbom

      - name: Scan SBOM with Grype
        id: scan
        run: |
          make sbom-scan

      - name: Upload SARIF to Code Scanning
        uses: github/codeql-action/upload-sarif@v2
        with:
[...]

PR #4988, opened 2026-06-16
chore: bump go version

brandtkeller:
## Description

go stdlib has a number of bugs that are reachable by Zarf code. Many are a low risk but given that they are reachable and that we are are on `1.26` I am advocating to bump this version acknowledging that this will require all downstream consumers to bump as well. 

Working on an update in #3842 to include stdlib scanning and that should be available shortly. 

## Related Issue

No associated issue

## Checklist before merging

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/zarf-dev/zarf
 
-go 1.26.0
+go 1.26.4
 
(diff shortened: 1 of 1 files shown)


soltysh (approved):
lgtm

PR #4993, opened 2026-06-19
feat!: add a digest to zarf package state and a digest inspect command

Racer159:
## Description

This PR adds support for a canonical digest for a package that can identify a unique Zarf package within the diverse ecosystem that Zarf packages can traverse through so that a package in cluster can be matched up with one from a tarball or one from an OCI registry.

This digest is not intended for trust applications but can serve reconciliation workflows to determine if a package is the same package even if the package's name and version happen to match (which can happen across a diverse ecosystem that may have inconsistent/lax practices in some areas).

## Related Issue

Fixes #N/A

## Checklist before merging


Racer159:
After further discussions going to add this to Zarf's internal package tracking in cluster so that tools can use this for Zarf packages wherever they are.

src/pkg/zoci/common.go
Racer159, line 1: realize this would be breaking for lib users but due to import cycles could not import from here into layout

src/pkg/packager/layout/oci.go
```diff
@@ -0,0 +1,275 @@
...
+		blobs[fileDigest] = filePath
+		totalLayerSize += fileSize
+	}
+
+	// Sort by digest for deterministic ordering.
+	sort.Slice(descs, func(i, j int) bool {
+		return descs[i].Digest.String() < descs[j].Digest.String()
+	})
+
+	configBytes, err := json.Marshal(p.Pkg)
```
AustinAbro321: I believe this should be from the actual zarf.yaml file given that p.Pkg can be filtered or mutated

src/pkg/packager/layout/oci.go
```diff
@@ -0,0 +1,275 @@
...
+// and sets p.digest.
+//
+// SHA256s for most files are read from checksums.txt (already computed at build
+// time), so only the small files excluded from that list (zarf.yaml, checksums.txt
+// itself, and post-signing provenance files) are read from disk.
+func (p *PackageLayout) computeManifest(ctx context.Context) error {
+	// Parse checksums.txt into relpath → sha256hex.
+	checksumsPath := filepath.Join(p.dirPath, Checksums)
+	checksumsBytes, err := os.ReadFile(checksumsPath)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
```
AustinAbro321: Shouldn't we error if the checksums file is doesn't exist?

src/cmd/package.go
```diff
@@ -708,9 +708,63 @@ func newPackageInspectCommand(v *viper.Viper) *cobra.Command {
...
+	ociConcurrency    int
+	namespaceOverride string
+}
+
+func newPackageInspectDigestCommand(v *viper.Viper) *cobra.Command {
+	o := &packageInspectDigestOptions{}
+	cmd := &cobra.Command{
+		Use:   "digest [ PACKAGE_SOURCE ]",
+		Short: "Outputs the SHA256 digest of the package's OCI manifest",
+		Long:  "Outputs the SHA256 digest of the package's OCI manifest. For OCI sources the digest is resolved directly from the registry. For local tarballs the manifest is computed deterministically from the package contents.",
```
brandtkeller: nit: semi-existing behavior but in isolation -  forgetting the `oci://` prefix (or not knowing it is required) leads to a generic error `ERR unknown source ...`. Maybe we need a few examples to set expectations and/or updating the error to potentially allude to needing a scheme?

AustinAbro321 (changes requested):
This is a cool feature, still in-review, but leaving some comments

src/pkg/packager/layout/oci.go
Racer159: we could - there are ancient Zarf packages that don't have one - not sure if we officially have moved on from those though?

src/cmd/package.go
Racer159: added examples to all the commands that take a [ PACKAGE SOURCE ] in package.go and added a suggestion to the error returned by identify source.

src/pkg/packager/layout/oci.go
AustinAbro321: Yes we have

src/pkg/zoci/push.go
AustinAbro321: Move this to Layout as well (and make it private)

src/pkg/packager/layout/oci.go
Racer159: done!

src/pkg/packager/layout/oci.go
Racer159: done!

src/cmd/package.go
```diff
@@ -708,9 +709,64 @@ func newPackageInspectCommand(v *viper.Viper) *cobra.Command {
 	cmd.AddCommand(newPackageInspectDefinitionCommand(v))
 	cmd.AddCommand(newPackageInspectValuesFilesCommand(v))
 	cmd.AddCommand(newPackageInspectDocumentationCommand(v))
+	cmd.AddCommand(newPackageInspectDigestCommand(v))
 	return cmd
 }
 
+type packageInspectDigestOptions struct {
+	ociConcurrency    int
```
brandtkeller: We register `ociConcurrency` value/flag but is it used anywhere?

src/pkg/zoci/push.go
```diff
@@ -48,97 +39,31 @@ func (r *Remote) PushPackage(ctx context.Context, pkgLayout *layout.PackageLayou
 		opts.Retries = DefaultRetries
 	}
 
-	tempDir, err := utils.MakeTempDir(config.CommonOptions.TempDirectory)
-	if err != nil {
-		return ocispec.Descriptor{}, fmt.Errorf("failed to create temp directory: %w", err)
+	if pkgLayout.Digest() == "" {
```
AustinAbro321: It'd be better to check the cache here since we could pull a partial package which would have a digest, but not be pushable, maybe a function `pkgLayout.IsPushable()` which checks for the cache?

AustinAbro321 (changes requested):
Two more small comments. Please change the PR title to begin with "feat!"

src/pkg/zoci/push.go
Racer159: done

src/pkg/zoci/push.go
Racer159: done

src/cmd/package.go
Racer159: no - and it isn't relevant - was vestigial from prior experiments - removed

AustinAbro321 (approved):
Nice work

brandtkeller (approved):
lgtm. I think there are a few negligible but present areas of improvement around the additional filesystem walks this introduces but any error surface I could find would have been prominent before any new logic - so risk is minimal (and the OS may optimize in some cases).

PR #4996, opened 2026-06-20
feat(package): support verification on all CLI package load operations

### LINKED ISSUE #4917: Support signature verification flags on all package commands

brandtkeller:
## Description

Given the core load operations for a built package have always accepted `VerifyBlobOptions`, SDK use has supported the newer keyless verification. Current CLI operations do not yet support the use of Keyless verification material and related options. 

This PR looks to enable CLI support for keyless verification options on all package load commands. This takes the stance of trying to reduce code duplcation where possible but ultimately configures the flags on each command. This feels like the more correct approach when compared to the alternative of persistent flags inherited given that not all commands use the globals similarly (IE package verify).  

## Related Issue

Fixes #4917

## Checklist before merging

### DIFF
--- a/site/src/content/docs/commands/zarf_init.md
+++ b/site/src/content/docs/commands/zarf_init.md
@@ -60,39 +60,46 @@ $ zarf init --artifact-push-password={PASSWORD} --artifact-push-username={USERNA
 ### Options
 
 ```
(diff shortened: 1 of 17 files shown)


AustinAbro321 (approved):
LGTM

PR #5001, opened 2026-06-22
feat: direct docker daemon pull

### LINKED ISSUE #4990: Change Docker Daemon fallback to pull directly using Docker SDK instead of Crane
The Docker Daemon fallback uses Crane, however this is bulky and slow. I believe it would be faster for users, and simpler in the codebase to pull direct form the daemon. The only issue is that Docker started using the OCI layout in Feb 2024 onward. It would be complex to support older versions. 

We should have this as a v1 breaking change. We can make a feature flag such as `PullDockerDaemonAsOCILayout` (ideally think of a better name) and not enable it by default.

AustinAbro321:
## Description

Direct pulling from Docker Daemon is far faster than the Crane SDK. In my experiments, a 1 GB image of random data took 44 seconds to pull with the Crane SDK and 11 with the Docker SDK, a 4x difference! 

Crane handles the legacy Docker formats which are complex, we already have most of the logic to handle an OCI format with Image archives, so we use Docker directly so long as it's using the newer format >v0.25.0.

Peak storage usage is higher with this PR 3x the image instead of 2x the image, but I believe this can be easily resolved in a follow up.

I added a feature flag "docker-daemon-direct-pull" so that if anything goes wrong user's have an escape hatch to go back to the Crane way of doing things, which we are keeping around until v1 so that users not up to date on their Docker Daemon can use the latest version of Zarf

## Related Issue

Fixes #4990 

## Checklist before merging

### DIFF
--- a/src/pkg/images/pull.go
+++ b/src/pkg/images/pull.go
@@ -26,6 +26,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/empty"
 	clayout "github.com/google/go-containerregistry/pkg/v1/layout"
 	"github.com/moby/moby/client"
+	"github.com/moby/moby/client/pkg/versions"
 	"github.com/zarf-dev/zarf/src/config"
 	"github.com/zarf-dev/zarf/src/pkg/logger"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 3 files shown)


src/pkg/images/pull.go
```diff
@@ -289,109 +293,201 @@ func pullFromDockerDaemon(ctx context.Context, daemonImages []imageWithOverride,
...
+
+	dockerImageSrc, err := oci.NewWithContext(ctx, dockerImageOCILayoutPath)
+	if err != nil {
+		return fmt.Errorf("failed to create OCI store: %w", err)
+	}
+	l.Info("pulling image from docker daemon", "name", daemonImage.overridden.Reference)
+	if _, err := copyImageFromOCILayout(ctx, dockerImageSrc, dst, manifests[0].Digest.String(), daemonImage.original, arch, concurrency); err != nil {
+		return err
+	}
+	return nil
```
soltysh, line 393: Nit: since you're not wrapping the error in any way, you could drop the if-check and just have:
```golang
 _, err := copyImageFromOCILayout(ctx, dockerImageSrc, dst, manifests[0].Digest.String(), daemonImage.original, arch, concurrency)
return err

soltysh (approved):
lgtm

src/pkg/images/pull.go
AustinAbro321: I think I'm going to make a follow up to this PR for storage efficiencies, I'll address this there

Issue #4986, opened 2026-06-12
Flake: mutating webhook timing out in e2e test labeled mode


AustinAbro321:
### Describe what should be investigated or refactored
In https://github.com/zarf-dev/zarf/pull/4937 we added `--init-agent-mode`, and test 98. This test is flaking. I've only seen it flake during arm64 runs. It's possible that we are not giving the pod enough CPU. We are doing an extra namespace check now, however this should never take 10+ seconds. 

We are creating the pod immediately after the init finishes, but that shouldn't be an issue, we even added a readiness check to the pod

### Links
https://github.com/zarf-dev/zarf/actions/runs/27431922672/job/81084166272?pr=4984

PR #5006, opened 2026-06-25
ci: additional agent logs

AustinAbro321:
## Description

I think this may have to do with endpoint slices being slow to create when the system is under CPU pressure, 
Relates to #4986

## Checklist before merging


brandtkeller (approved):
lgtm

Issue #4969, opened 2026-06-09
Expose effective applied package configuration for drift detection


emoskito:
### Is your feature request related to a problem? Please describe.

Zarf already tracks some deployment state for a package, such as installed components, but it does not currently expose the full effective configuration that was applied during deployment. This makes it difficult for orchestration tools to reliably detect package-level configuration drift.

For tools managing Zarf deployments declaratively, it is important to know not only what was intended to be applied, but what Zarf actually applied at deployment time. That configuration can then be stored in the tool’s own state and later compared against the current deployed package configuration.

Today, Zarf exposes some deployment metadata, but there does not appear to be a way to retrieve the full set of deploy-time inputs that influenced a deployment, such as values, variables, component selection, overrides, or other mutations

This becomes challenging when a package is later re-deployed with different inputs by another user, tool, automation, or workflow. An orchestration tool may no longer have a reliable way to determine whether the deployed package configuration still matches what it originally applied.

### Describe the behavior you'd like

- Given an orchestration tool deploys a Zarf package with a specific set of values, variables, selected components, or other deploy-time configuration  
- When the deployment completes  
- Then Zarf exposes the effective configuration that was applied so the orchestration tool can persist it in its own state

- Given a Zarf package has already been deployed to a cluster  
- When an orchestration tool queries the deployment through the SDK, CLI, or another supported interface  
- Then Zarf returns the current or last-applied deployment configuration for that package

- Given an orchestration tool has a previously saved copy of the applied configuration  
- When it compares that configuration to the current or last-applied configuration returned by Zarf  
- Then the tool can determine whether package-level configuration drift has occurred

Ideally, Zarf would persist and expose an applied package configuration record representing the effective configuration used during deployment. Examples of useful data could include:

- Package identity, version, and/or digest
- Installed components (already tracked today)
- Variables and values supplied at deploy time
- Effective Helm values or value overrides
- Package mutations or deployment overrides
- Other deploy-time inputs that influence the resulting deployment
- Optionally, a stable digest/hash of the effective configuration

For drift detection, the most useful representation would be the resolved/effective configuration after defaults, variables, overrides, and mutations have been applied, since that reflects what Zarf actually used during deployment.

[...]

PR #5007, opened 2026-06-26
feat: add source and package status to package state

### LINKED ISSUE #4182: Add overall package status to state
It would be nice to have a top level "status" field for a package to keep track of the package lifecycle. Currently the components have a "ComponentStatus" I would like a top level PackageStatus as well to track when the package is deploying, finished successfully, errored, or removing

- **Given** a package is not deployed
- **When** the package deployment is started
- **Then** the package state stores off the status as "Deploying"

---
- **Given** a package is deploying
- **When** the package deployment finishes successfully
- **Then** the package state stores off the status as "Deployed"
--
- **Given** a package is deploying
- **When** the package deployment fails
- **Then** the package state stores off the status as "Error"
--
- **Given** a package has successfully deployed
- **When** the package is requested to be removed
- **Then** the package state stores off the status as "Removing"

### Additional context
This allows the Zarf SDK to be used to better understand package lifecycle without needing to redefine the package status.

Racer159:
## Description

This adds the package's source and an overall package status to Zarf's DeployedPackage in-cluster state.

## Related Issue

Fixes #4182

## Checklist before merging

### DIFF
--- a/src/pkg/packager/deploy.go
+++ b/src/pkg/packager/deploy.go
@@ -79,6 +79,9 @@ type DeployOptions struct {
 
 	// [Library Only] A map of component names to chart names containing Helm Chart values to override values on deploy
 	ValuesOverridesMap ValuesOverrides
+	// Source is the original source string used to load the package (e.g. oci:// URL or tarball path).
+	// Recorded in the cluster's deployed package state.
+	Source string
 	// IsInteractive decides if Zarf can interactively prompt users through the CLI
 	IsInteractive bool
 	// SkipVersionCheck skips version requirement validation
@@ -265,7 +268,7 @@ func (d *deployer) deployComponents(ctx context.Context, pkgLayout *layout.Packa
 		deployedComponents = append(deployedComponents, deployedComponent)
 		idx := len(deployedComponents) - 1
 		if d.isConnectedToCluster() {
-			if _, err := d.c.RecordPackageDeployment(ctx, pkgLayout.Pkg, pkgLayout.Digest(), deployedComponents, packageGeneration, state.WithPackageConnectivity(opts.Connected), state.WithPackageNamespaceOverride(opts.NamespaceOverride)); err != nil {
+			if _, err := d.c.RecordPackageDeployment(ctx, pkgLayout.Pkg, pkgLayout.Digest(), deployedComponents, packageGeneration, state.WithPackageConnectivity(opts.Connected), state.WithPackageNamespaceOverride(opts.NamespaceOverride), state.WithPackageSource(opts.Source), state.WithPackageStatus(state.PackageStatusDeploying)); err != nil {
 				l.Debug("unable to record package deployment", "component", component.Name, "error", err.Error())
 			}
 		}
@@ -292,7 +295,7 @@ func (d *deployer) deployComponents(ctx context.Context, pkgLayout *layout.Packa
 				deployedComponents[idx].Status = state.ComponentStatusFailed
 				deployedComponents[idx].InstalledCharts = state.MergeInstalledChartsForComponent(deployedComponents[idx].InstalledCharts, charts, true)
 				if d.isConnectedToCluster() {
-					if _, err := d.c.RecordPackageDeployment(ctx, pkgLayout.Pkg, pkgLayout.Digest(), deployedComponents, packageGeneration, state.WithPackageConnectivity(opts.Connected), state.WithPackageNamespaceOverride(opts.NamespaceOverride)); err != nil {
+					if _, err := d.c.RecordPackageDeployment(ctx, pkgLayout.Pkg, pkgLayout.Digest(), deployedComponents, packageGeneration, state.WithPackageConnectivity(opts.Connected), state.WithPackageNamespaceOverride(opts.NamespaceOverride), state.WithPackageSource(opts.Source), state.WithPackageStatus(state.PackageStatusFailed)); err != nil {
 						l.Debug("unable to record package deployment", "component", component.Name, "error", err.Error())
 					}
 				}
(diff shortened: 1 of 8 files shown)


src/pkg/packager/deploy.go
```diff
@@ -79,6 +79,9 @@ type DeployOptions struct {
 
 	// [Library Only] A map of component names to chart names containing Helm Chart values to override values on deploy
 	ValuesOverridesMap ValuesOverrides
+	// Source is the original source string used to load the package (e.g. oci:// URL or tarball path).
+	// Recorded in the cluster's deployed package state.
+	Source string
```
AustinAbro321, line 84: From a consumer perspective, will you care about the actual source or only which type of source. If the latter, then an enum here might be better

src/pkg/state/state.go
```diff
@@ -87,15 +87,29 @@ func ParseServiceKey(s string) (ServiceKey, error) {
 	return "", fmt.Errorf("invalid service key %q, valid keys are: %v", s, AllServiceKeys())
 }
 
+// PackageStatus defines the overall deployment status of a Zarf package.
+type PackageStatus string
```
AustinAbro321, line 91: I'm curious what you'd think of splitting out the state of the lifecycle with the health of the package.  Something like below, this could also apply to components, though we'd want to keep their current status for some time for backward compatibility

```
// PackagePhase is the current lifecycle operation.
type PackagePhase string

const (
      PackagePhaseIdle      PackagePhase = "Idle" 
      PackagePhaseDeploying PackagePhase = "Deploying"
      PackagePhaseRemoving  PackagePhase = "Removing"
)

// PackageHealth is the outcome of the most recent completed operation.
type PackageOutcome string

const (
      PackageOutcomeUnknown  PackageOutcome = "Unknown" 
      PackageOutcomeHealthy    PackageOutcome = "Healthy" 
      PackageOutcomeDegraded PackageOutcome = "Degraded"
)
```

dgershman (changes requested):
## Code & Security Review

Adds `Source` and an overall package `Status` to the in-cluster `DeployedPackage` state, surfaces them in `zarf package list`, and threads status transitions through deploy/remove. Clean, well-scoped, and backwards-compatible (`omitempty` + `GetStatus()` → `Unknown` for pre-existing secrets, reflected in the updated testdata). e2e coverage for the new fields is solid.

> Verified locally on Go 1.26.4: `gofmt -l` clean, and `go build` + `go vet` pass on the changed packages (`src/pkg/packager/...`, `src/pkg/state/...`, `src/cmd/...`). The full test suite (incl. e2e, which needs a live cluster) is left to CI.

### Critical Issues
None.

### Security Review
**Strengths:**
- The recorded `Source` (oci:// URL or tarball path) is not credential-bearing for Zarf, and it is stored in the same `zarf`-namespace package secret that already holds deployment metadata — no new exposure surface.
- No injection / auth / crypto surface touched; changes are pure state plumbing.

**Concerns:** None.

### Code Quality

🟡 **`src/pkg/packager/remove.go:219-225` — partial-remove hardcodes `Status = Succeeded`, which can misrepresent package health.**
After a successful partial remove (some requested components removed, others left untouched), the package status is unconditionally set to `PackageStatusSucceeded`. The remaining components are never re-evaluated, so a leftover component that still carries `ComponentStatusFailed` (e.g. removing a healthy component out of a previously-`Failed` deploy) leaves the package reporting `Succeeded` while a child component reports `Failed`. That internal inconsistency is exactly what the new `Status` column is meant to surface in `zarf package list`. Consider deriving the reset value from the remaining `DeployedComponents` (or only resetting to `Succeeded` when all remaining components are `Succeeded`, else `Failed`/`Unknown`).

🟢 **`src/pkg/packager/remove.go` — status writes guard on `opts.Cluster != nil` rather than `requiresCluster`.**
In the `cluster != nil && !requiresCluster` path, the `else` branch (lines 98-105) builds a fresh `depPkg` with only component names — no `Source`, `Digest`, `Generation`, or `InstalledCharts`. The new `UpdateDeployedPackage` calls (lines 135-140, 198-202, 212-225) then apply that stripped object over the real stored secret. This is largely **pre-existing** (the inner chart-removal updates at the old 165/187 already did this), but the added writes propagate the clobber. Worth confirming whether these status writes should also gate on `requiresCluster` for consistency.

🟢 **`src/cmd/package.go:1330-1338, 1377` — `Digest` and `Source` are emitted in JSON/YAML list output but omitted from the table columns** (only `Status` was added to the table header). Presumably intentional to keep the table narrow; flagging the JSON-vs-table asymmetry in case it's unintended.

### Summary Table
| Color  | Meaning      | Verdict effect            |
|--------|--------------|---------------------------|
| Red    | Must fix     | Request changes           |
| Yellow | Should fix   | Request changes           |
| Green  | Consider     | Approve allowed           |

**Recommendation:** Request Changes — driven by [0 Red, 1 Yellow, 2 Green] findings. The Yellow (partial-remove status accuracy) directly undercuts the observability goal of this feature; addressing it in this round avoids a follow-up.

---

[🐦‍⬛ Reviewed by Crow via Claude Code](https://github.com/radiusmethod/crow)

AustinAbro321 (commented):
Commenting some ideas, but I think this should be hashed out further in a ZEP. Can be a short one, but I'd prefer that we have a document to gather feedback from others, especially since this data is largely targeted at those integrating with Zarf. 

For instance https://github.com/zarf-dev/zarf/issues/4969 was another recent request for additional data from state

src/pkg/packager/deploy.go
Racer159: Actual source - tarballs are hard to reconstruct but an `oci://` reference for example could be repulled and deployed again if you needed to.

Racer159:
> Commenting some ideas, but I think this should be hashed out further in a ZEP. Can be a short one, but I'd prefer that we have a document to gather feedback from others, especially since this data is largely targeted at those integrating with Zarf.
> 
> For instance #4969 was another recent request for additional data from state

Copy - thinking of resurrecting the old https://github.com/zarf-dev/proposals/pull/27 ZEP for this - would it be "ok" to consider Zarf behavior changes there as well? (i.e that if a component `name` is removed from one package version to another that it is internally reconciled (i.e. that component is removed prior to the deploy proceeding?) or should it stay higher level to library consumers only.

(I think that can be done without a full history system, just look at the old state if the package requires the cluster and reconcile and the more I think about it I don't know if we need a full history system either but we can discuss that too)

AustinAbro321:
@Racer159 Yeah I think it makes sense to resurrect that ZEP

src/pkg/state/state.go
Racer159: added this in https://github.com/zarf-dev/proposals/pull/27

Racer159:
> @Racer159 Yeah I think it makes sense to resurrect that ZEP

Resurrected here https://github.com/zarf-dev/proposals/pull/27

PR #5010, opened 2026-06-26
fix: change adoptExistingResources to takeOwnership

### LINKED ISSUE #4436: Replace custom adoptExistingResource logic with client.takeOwnership
Helm introduced an option on upgrade called take ownership. See https://helm.sh/docs/helm/helm_upgrade/. Zarf created custom logic for this functionality called adopt-existing-resources, we should be able to remove that logic now in way of using the Helm take ownership client option. 

Also to be considered is if we should rename adopt-existing-resources to take-ownership.

AustinAbro321:
## Description

This changes `--adoptExistingResources` to `--take-ownership` and the supporting backend logic to use the native Helm feature. 

This fixes a bug where Zarf did not properly handle cluster scoped resources. 

There could be an argument for not taking the Helm flag name since we do more; we also label the namespace `zarf.dev/agent: mutate`, but overall I think it's more intuitive to be in line with upstream.

Removes the parallel in TestHelm, but this was only actually saving a few seconds, and made the tests less convenient to run.

## Related Issue

Fixes #4436

## Checklist before merging


dgershman (dismissed):
## Code & Security Review

Solid, well-scoped change. Swapping Zarf's hand-rolled adoption (dynamic-client `GET`/`Update` in the post-renderer) for Helm v4's native `client.TakeOwnership` is the right call: I verified `TakeOwnership` exists on both `action.Install` and `action.Upgrade` in `helm.sh/helm/v4 v4.2.0`, and that its `requireAdoption` + `setMetadataVisitor` path stamps the same `app.kubernetes.io/managed-by: Helm` label and `meta.helm.sh/release-{name,namespace}` annotations the removed code did — but resolves the resource via its real REST mapping, so cluster-scoped resources are handled correctly. The old code issued a namespaced `GET` (`dc.Resource(...).Namespace(deployedNamespace)`) for every kind, which 404s on cluster-scoped resources and silently skipped them — exactly the bug being fixed. `go build` and `go vet` pass on the changed packages.

Backward-compat is handled well: `--adopt-existing-resources` remains as a deprecated alias bound to the same var, and `DeployOptions.AdoptExistingResources` is retained and OR'd in (`opts.TakeOwnership || opts.AdoptExistingResources`) for SDK callers.

### Security Review
**Strengths:**
- Net reduction in attack surface — removes a manual privileged dynamic-client update path in favor of well-tested upstream Helm logic.
- Destructive ownership-takeover remains opt-in behind the flag with unchanged warnings; the initial-namespace guard (`default`/`kube-system`/...) is preserved.
- No new secret handling, auth, or injection surface.

**Concerns:** none.

### Code Quality

- **Yellow — `--take-ownership` is a silent no-op on `zarf dev deploy`.** `src/cmd/dev.go:542-544` binds both flags to `o.takeOwnership`, but `DevDeploy` never threads it through: `DevDeployOptions` (`src/pkg/packager/dev.go:28`) has no take-ownership field, and the internal `DeployOptions{}` it builds (`src/pkg/packager/dev.go:160-167`) leaves `TakeOwnership` at its zero value. So `o.takeOwnership` is written by the flag and never read — `zarf dev deploy --take-ownership` does nothing. This was already true for `--adopt-existing-resources` on `main`, but the PR is reworking exactly this flag and now newly advertises `--take-ownership` here. Please either wire it through (`DevDeployOptions.TakeOwnership` → `DeployOptions.TakeOwnership`) or drop the flag from `dev deploy` so the help text stops promising behavior it doesn't deliver.

- **Yellow — no test for the headline cluster-scoped fix.** The PR's stated purpose is fixing adoption of cluster-scoped resources, but the updated e2e tests only exercise a namespaced `Deployment` (`src/test/e2e/25_helm_test.go:257` `testHelmAdoption`) and namespace agent-label handling (`src/test/e2e/49_helm_adopt_ns_test.go`). Nothing adopts a cluster-scoped object (e.g. `ClusterRole`/`ClusterRoleBinding`/`PV`), so the exact regression being fixed has zero coverage and could silently break again. The PR's own checklist item \"Test ... added or updated as needed\" is unchecked. A focused adoption test on a cluster-scoped resource would lock this in.

- **Green — ADR.** Given the user-facing flag rename plus the shift to Helm-native ownership semantics (and the naming tradeoff you flag in the description), consider whether an ADR is warranted. Maintainer call.

### Summary Table
| Color  | Meaning      | Verdict effect            |
|--------|--------------|---------------------------|
| Red    | Must fix     | Request changes           |
| Yellow | Should fix   | Request changes           |
| Green  | Consider     | Approve allowed           |

**Recommendation:** Request Changes — driven by [0 Red, 2 Yellow, 1 Green] findings. Both Yellows are small and fit naturally in this same change.

---

[🐦‍⬛ Reviewed by Crow via Claude Code](https://github.com/radiusmethod/crow)

src/internal/packager/helm/post-render.go
```diff
@@ -269,49 +257,6 @@ func (r *renderer) editHelmResources(ctx context.Context, resources []releaseuti
 			r.namespaces[namespace] = cluster.NewZarfManagedNamespace(namespace)
 		}
 
-		// If we have been asked to adopt existing resources, process those now as well
-		if r.adoptExistingResources {
```
brandtkeller, line 273: I'm in favor of the approach - but does this removal signal a behavioral change that is not inherently one-for-one with the deprecation language?

AustinAbro321: As far as I can tell from evaluating Helm behavior, no

brandtkeller (changes requested):
minor - I've found one stray mention in the docs for `--adopt-existing-resources` as well as uncovering a few mentions for `sget://` which was previously removed but in the same command and elsewhere. Worth including in this PR?

AustinAbro321:
@brandtkeller I changed the html file to use `--take-ownership`, I did not update it for sget. It's using the old logs and we already have an issue to remake the docs https://github.com/zarf-dev/zarf/issues/3969

brandtkeller (approved):

brandtkeller (approved):

PR #5011, opened 2026-06-26
feat(verification)!: behavior updates for verification logic

### LINKED ISSUE #4909: [Proposal] Update verification strategy propagation and evaluation
### Summary

Zarf has three verification strategies for package signatures: VerifyAlways, VerifyIfPossible, and VerifyNever. The default `VerifyIfPossible` attempts verification but swallows all failures with a warning and continues. This means a signed package whose signature does not match is treated identically to an unsigned package. For supply chain security purposes, these are meaningfully different conditions and should not be handled the same way.

Recommendation

VerifyIfPossible should distinguish between two failure modes: a package with no signature and no verification material provided (tolerate, warn, continue) versus any case where a verification claim exists or is asserted and fails (error). A signed package with an invalid signature should always fail regardless of strategy. An unsigned package presented alongside a public key should also fail as providing a key implies an expectation of a signature.

This potentially means evolving the original discussion from having the verification be a boolean to using the verification strategey enums directly. The escape hatch would still exist to perform deployments when the verification operation is failing for some reason.

The distinction from the original discussion being that a signed package without verification material would now pass by default as we handle 3 distinct default paths:
- Signed, valid signature — verification succeeds
- Signed, wrong or corrupt signature — tamper-evident failure; should always error
- Signed, no verification material available — operator has nothing to verify against

One of the primary constraints here is that we don't yet have a plan for exposing all of the verification flags to the commands that load packages. 

Otherwise the `--verify` flag should likely become a enum that allows for the mentioned escape hatch when necesary (previously --skip-signature-validation` using `never` and distinguish between `default` and `always` accordingly.

brandtkeller:
## Description

This PR fixes two behaviors around Zarf signature verification:
1. On the CLI - `--verify` is now an enum representing `Verification Strategy`. The allowed options are `always|if-possible|never` with `if-possible` being the default. We retained the use of booleans for backwards compatibility.
2. Changes the verification behavior - A signed package that is provided verification material but for which the verification fails is now (correctly) an error. Additionally providing key material to the verification workflow for an unsigned package likely indicates intent to verify and is now a failure - this protects against signature stripping attacks. 

Behaviorally this is a breaking change but more of a correction. The `--verify` CLI flag is also being updated from a boolean to an enum but currently has backwards compatibility. I think this can go either way for being a true breaking change or doing so when the boolean optionality is removed.

## Related Issue

Fixes #4909

## Checklist before merging

### DIFF
--- a/src/config/lang/english.go
+++ b/src/config/lang/english.go
@@ -224,7 +224,7 @@ $ zarf init --artifact-push-password={PASSWORD} --artifact-push-username={USERNA
 	CmdPackageShort                       = "Zarf package commands for creating, deploying, and inspecting packages"
 	CmdPackageFlagConcurrency             = "Number of concurrent layer operations when pulling or pushing images or packages to/from OCI registries."
 	CmdPackageFlagFlagPublicKey           = "Path to public key file for validating signed packages"
-	CmdPackageFlagVerify                  = "Signature verification mode: never, if-possible (default; tampered signatures always fail), or always. Bare --verify means always."
+	CmdPackageFlagVerify                  = "Signature verification mode (never|if-possible|always). Tampered signatures are always fatal except with never."
 	CmdPackageFlagSkipSignatureValidation = "[Deprecated] Use --verify=never instead. This flag will be removed in Zarf v1.0.0."
 	CmdPackageFlagRetries                 = "Number of retries to perform for Zarf operations like git/image pushes"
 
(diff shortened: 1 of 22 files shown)


src/config/lang/english.go
```diff
@@ -224,8 +224,8 @@ $ zarf init --artifact-push-password={PASSWORD} --artifact-push-username={USERNA
 	CmdPackageShort                       = "Zarf package commands for creating, deploying, and inspecting packages"
 	CmdPackageFlagConcurrency             = "Number of concurrent layer operations when pulling or pushing images or packages to/from OCI registries."
 	CmdPackageFlagFlagPublicKey           = "Path to public key file for validating signed packages"
-	CmdPackageFlagVerify                  = "Verify the Zarf package signature"
-	CmdPackageFlagSkipSignatureValidation = "[Deprecated] Skip validating the signature of the Zarf package"
+	CmdPackageFlagVerify                  = "Signature verification mode (never|if-possible|always)."
```
AustinAbro321: nit: I'd prefer having the user read "always" first. Maybe it will subtly influence some users to use always more often, or maybe I'm doing fake psychology. Anyway 
```suggestion
	CmdPackageFlagVerify                  = "Signature verification mode (always|if-possible|never)."
```

AustinAbro321 (commented):
Nice work, I have one extremely small nit, then lets get this in

AustinAbro321 (approved):

PR #5013, opened 2026-06-27
feat: support multiple package removal

### LINKED ISSUE #4992: Support passing multiple packages to zarf package remove

I would like to be able to pass multiple package names to `zarf package remove` to speed up iterating on a deployment that is made up of multiple packages.

I can of course work around this by using a bash for-loop, but it feels like this could be natively supported by Zarf.

- **Given** a list of packages
- **When** I feed said the list of packages to zarf package remove
- **Then** Zarf handles removing all of the packages for me.

A bash for-loop is my current workaround for this.

Avadhut03:
## Description
This PR adds support for removing multiple packages in a single zarf package remove command by allowing multiple package names or sources to be passed as arguments. The existing interactive package selection behavior is preserved when no arguments are provided, and the command now iterates through each specified package

## Related Issue

Fixes #4992 

## Checklist before merging

### DIFF
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -1490,7 +1490,6 @@ func (o *packageRemoveOptions) run(cmd *cobra.Command, args []string) error {
 			errs = append(errs, fmt.Errorf("package %q: %w", packageSource, err))
 			continue
 		}
-
 	}
 	return errors.Join(errs...)
 }
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -1420,7 +1420,6 @@ func newPackageRemoveCommand(v *viper.Viper) *cobra.Command {
 }
 
 func (o *packageRemoveOptions) run(cmd *cobra.Command, args []string) error {
-
 	ctx := cmd.Context()
 	if len(args) == 0 {
 		packageSource, err := choosePackage(ctx, nil)
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -1396,9 +1396,9 @@ func newPackageRemoveCommand(v *viper.Viper) *cobra.Command {
 	o := &packageRemoveOptions{}
 
 	cmd := &cobra.Command{
-		Use:               "remove { PACKAGE_SOURCE | PACKAGE_NAME } --confirm",
+		Use:               "remove { PACKAGE_SOURCE | PACKAGE_NAME }... --confirm",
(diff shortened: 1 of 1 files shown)


brandtkeller:
Thank you for the patience @Avadhut03 ! We'll get this into review shortly.

src/cmd/package.go
```diff
@@ -1451,37 +1454,44 @@ func (o *packageRemoveOptions) run(cmd *cobra.Command, args []string) error {
 		RemoteOptions:        defaultRemoteOptions(),
 		CachePath:            cachePath,
 	}
-	pkg, err := packager.GetPackageFromSourceOrCluster(ctx, c, packageSource, o.namespaceOverride, loadOpts)
-	if err != nil {
-		return fmt.Errorf("unable to load the package: %w", err)
-	}
+
 	removeOpt := packager.RemoveOptions{
```
brandtkeller, line 1458: If we want to pursue this change (leaving the option open to leave the command as-is) - I think we'll want additional validation on the options you can pass in. 

things like namespace override shouldn't be accepted on `remove` when multiple packages are being targeted. Same for values and potentially other inputs.

brandtkeller (changes requested):
Initial thought around inputs that will be acceptable. 

I do want to leave open the distinct possibility that we decide not to do this work. It's still very much appreciated and we should evaluate it through this PR - but I also don't want to make the command oddly-shaped by adding variance in how the command operates based on number of packages.

src/cmd/package.go
Avadhut03: Thanks for the feedback! That makes sense. I agree that the command shouldn't have ambiguous behavior depending on the number of packages.

If we decide to move forward with supporting multiple package removal, I can add validation to explicitly reject options that don't make sense in that context (for example, `--namespace`, `--values`, `--set`, and any other package-specific overrides). That way the behavior remains predictable and we avoid ambiguous semantics.

If, after discussion, we decide that supporting multiple package removal isn't the right direction for the CLI, I'm also happy to keep the command as-is. I'll wait for the design decision before investing further in the implementation.

brandtkeller:
Hi @Avadhut03 ! i've been taking a look at this further and thinking about this more and believe that the project is going to want to maintain the current stance. One package per remove operation from a CLI perspective. This maintains alignment with other commands and prevents mis-aligning command configuration with the intended target.

PR #5017, opened 2026-06-29
feat: reduce docker daemon pull storage footprint

AustinAbro321:
## Description

In #5001 we introduced a Docker daemon pull that is about 4x faster than the Crane pull, but peaks at 3x original image size during the pull form the Daemon as opposed to the 2x original image size with the Crane method.

1. The image.tar we pull from the Docker daemon
2. The temporary directory we unarchive the image.tar into
3. The final OCI layout of the package

By streaming directly from the Docker daemon buffer, we remove step 1. Locally, a package with a five gb image took 57 seconds to create and 10GB of peak storage usage whereas on main it took 1 minute and 15GB of peak storage usage. 

I looked for any other storage efficiencies to be gained by DecompressStream but didn't find any because it's only useful when you have the following conditions: 
- non-seekable export stream
- The stream is buffered only to extract
- The format is known ahead of time

Technically for the last point we could use archives.Identify, but we've run into some bugs with it in the past so I made the extract format required.

## Checklist before merging


src/pkg/archive/archive.go
```diff
@@ -230,53 +215,41 @@ func withArchive(path string, extractor archives.Extractor, fn func(ex archives.
...
 	}
-	root, err := os.OpenRoot(dst)
-	if err != nil {
-		return fmt.Errorf("opening root %q: %w", dst, err)
+	return decompressFrom(ctx, opts.Extractor, src, dst, opts)
+}
+
+// decompressFrom extracts input into dst per opts, then optionally recurses into nested archives.
+// It is the shared core of Decompress (file source) and DecompressStream (stream source).
+func decompressFrom(ctx context.Context, extractor archives.Extractor, input io.Reader, dst string, opts DecompressOpts) error {
```
brandtkeller: This signature requiring an extractor explicitly and then also consuming the `opts.Extractor` under a possible permutation looks odd to me. Maybe not avoidable but I am curious what happens if they are potentially different?

src/pkg/archive/archive.go
```diff
@@ -230,53 +215,41 @@ func withArchive(path string, extractor archives.Extractor, fn func(ex archives.
...
 
-// unarchiveWithStrip extracts all entries from src into dst, stripping the
-// first 'strip' path components and optionally overwriting existing files.
-func unarchiveWithStrip(ctx context.Context, extractor archives.Extractor, src, dst string, strip int, overwrite bool) (err error) {
-	if err := os.MkdirAll(dst, dirPerm); err != nil {
-		return fmt.Errorf("creating dest %q: %w", dst, err)
+// DecompressStream extracts an archive read from src into dst. Because the archive format cannot be detected
+// from a stream, opts.Extractor is required. The extractor must support sequential reads from a non-seekable stream.
+// Formats that require random access to the whole archive (e.g. zip) will error unless src also implements io.ReaderAt and io.Seeker.
+func DecompressStream(ctx context.Context, src io.Reader, dst string, opts DecompressOpts) error {
```
brandtkeller, line 230: thinking about input sanitization here - what would an empty/nil reader produce for runtime/output?

brandtkeller (commented):
Two questions to help inform my understanding and maybe potential changes - not inherently blocking but interested if there are any protective measures or tests we might add.

src/pkg/archive/archive.go
AustinAbro321: The cause of this is that if someone sets opts.Extractor then that extractor is used for all the nested unarchives as well. And so in order for this to work we can't set opts.Extractor on the first read of it. We could make a breaking SDK change and introduce a new field `NestedExtractor` which would be easier to follow and give users better flexibility. Thoughts?

src/pkg/archive/archive.go
AustinAbro321: I believe it will panic, added input validation

src/pkg/archive/archive.go
brandtkeller: My only hangup currently is that it feels like a pre-existing issue is being highlighted here in that that parameter on `nestedUnarchive` feels redundant - It only processes `.tar` extensions and thus a non-nil extractor has to be of type `archives.Tar{}`. 

Which means we consider removing the parameter and/or changing the nesting logic?

AustinAbro321: In Zarf `nestedUnarchive` only handles `archives.Tar{}`, but an SDK user could give `Decompress` a .gz file with .bz inner archives. So long as they keep `Extractor` nil and set `UnarchiveAll` then `nestedUnarchive` will auto select the .bz extractor for the inner archives

brandtkeller: in `nestedUnarchive` would:
```golang
if !d.Type().IsRegular() || !strings.HasSuffix(path, extensionTar) {
			return nil
		}
```
not inhibit anything but tar archives?

AustinAbro321: I totally missed that, thank you, fixed

brandtkeller (approved):
lgtm - this should be good context for build-time resource requirements. I've also secretly wondered for a while now what it would mean to stream contents of a package during a deploy... fun thought.

PR #5020, opened 2026-06-30
fix: skeleton packages drop non-host-arch components

### LINKED ISSUE #5019: Skeleton packages drop `only.cluster.architecture` components for non-host arches, making multi-arch composed packages un-importable
### Environment
Device and OS: N/A
App version: v0.80.0
Other:

### Steps to reproduce
1. zarf.yaml importing from the published init skeleton:
```
kind: ZarfInitConfig
metadata:
  name: init
components:
  - name: zarf-injector
    required: true
    import:
      url: oci://ghcr.io/zarf-dev/packages/init:v0.80.0
```
2. Build for a non-amd64 arch: `zarf package create . -a arm64 --confirm`

### Expected result
Skeletons are arch-agnostic, so all `only.cluster.architecture` variants are retained and the arm64 injector composes in.

### Actual Result
`ERR  failed to create package: no compatible component named zarf-injector found`
The skeleton only kept the publisher host-arch (amd64) components; the arm64 zarf-injector was filtered out at publish time.

Source defines both arches, but the published skeleton's `zarf.yaml` has a single zarf-injector gated `architecture: amd64`. `-a amd64` builds fine; `-a arm64` fails as above.

JeffResc:
## Description
Skeleton publishing resolved the package definition using the build host's architecture, so compose filtered out components gated to other arches via `only.cluster.architecture` before the package was marked arch-agnostic. The published `zarf init` skeleton therefore shipped only the amd64 `zarf-injector`/`k3s`, and importing it to build for arm64 failed with `no compatible component named zarf-injector found`.

This resolves the skeleton definition under `SkeletonArch` (match-all), so all architecture variants are retained, mirroring how `Only.Flavor` is already neutralized for skeletons. Consumers are unaffected: they import the (now complete) skeleton and their own load filters to the concrete target arch.

## Related Issue

Fixes #5019

## Checklist before merging

### DIFF
--- a/src/pkg/packager/load/import_test.go
+++ b/src/pkg/packager/load/import_test.go
@@ -521,3 +521,52 @@ func TestCompatibleComponent(t *testing.T) {
 		})
 	}
 }
+
+func TestCompatibleComponentSkeletonArch(t *testing.T) {
+	t.Parallel()
+
+	for _, arch := range []string{"amd64", "arm64"} {
+		c := v1alpha1.ZarfComponent{
+			Only: v1alpha1.ZarfComponentOnlyTarget{
+				Cluster: v1alpha1.ZarfComponentOnlyCluster{Architecture: arch},
+			},
+		}
+		require.True(t, compatibleComponent(c, v1alpha1.SkeletonArch, ""),
+			"skeleton compose must retain %s component", arch)
+	}
+}
+
+func TestResolveImportsSkeletonRetainsAllArches(t *testing.T) {
+	t.Parallel()
+	ctx := testutil.TestContext(t)
+
+	dir := t.TempDir()
+	zarfYAML := `kind: ZarfPackageConfig
+metadata:
+  name: arch-skeleton
+components:
+  - name: injector
+    only:
+      cluster:
+        architecture: amd64
+  - name: injector
+    only:
+      cluster:
+        architecture: arm64
+`
+	require.NoError(t, os.WriteFile(filepath.Join(dir, layout.ZarfYAML), []byte(zarfYAML), 0o644))
+	b, err := os.ReadFile(filepath.Join(dir, layout.ZarfYAML))
+	require.NoError(t, err)
+	pkg, err := pkgcfg.Parse(ctx, b)
+	require.NoError(t, err)
+
+	resolved, _, err := resolveImports(ctx, pkg, dir, v1alpha1.SkeletonArch, "", []string{}, "", false, types.RemoteOptions{})
+	require.NoError(t, err)
+
+	archs := []string{}
+	for _, c := range resolved.Components {
+		archs = append(archs, c.Only.Cluster.Architecture)
+	}
+	require.ElementsMatch(t, []string{"amd64", "arm64"}, archs,
+		"skeleton must retain all only.cluster.architecture variants")
+}
--- a/src/pkg/packager/load/load.go
+++ b/src/pkg/packager/load/load.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 	"time"
 
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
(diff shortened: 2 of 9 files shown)


src/pkg/packager/publish.go
```diff
@@ -231,6 +231,7 @@ func PublishSkeleton(ctx context.Context, path string, ref registry.Reference, o
 		Flavor:             opts.Flavor,
 		SkipVersionCheck:   opts.SkipVersionCheck,
 		SkipRequiredValues: true,
+		Skeleton:           true,
```
brandtkeller: You've got coverage on this through e2e but I think we could also add a test to publish?

brandtkeller (changes requested):

src/pkg/packager/publish.go
JeffResc: Just added `TestPublishSkeletonMultiArch()`

src/test/e2e/10_component_flavor_test.go
```diff
@@ -85,34 +85,30 @@ func TestPublishFlavor(t *testing.T) {
...
-	var config v1alpha1.ZarfPackage
-	err = yaml.Unmarshal([]byte(stdOut), &config)
-	require.NoError(t, err)
-
-	for i, component := range config.Components {
-		require.Equal(t, expectedIDs[i], component.Name)
+	// A skeleton is architecture-agnostic, so it retains every arch variant of the flavor's components.
+	assertSkeleton := func(t *testing.T, flavor string) {
+		t.Helper()
+		_, _, err := e2e.Zarf(t, "package", "publish", flavorTest, "--flavor", flavor, "--no-color", "oci://"+ref, "--plain-http")
```
AustinAbro321:
```suggestion
		_, _, err := e2e.Zarf(t, "package", "publish", flavorTest, "--flavor", flavor, "oci://"+ref, "--plain-http")
```
no color is built in to e2e Zarf

src/pkg/packager/load/load.go
```diff
@@ -41,6 +41,8 @@ type DefinitionOptions struct {
 	IsInteractive bool
 	// SkipVersionCheck skips version requirement validation
 	SkipVersionCheck bool
+	// Skeleton retains all architecture variants instead of filtering to the build host arch
+	Skeleton bool
```
AustinAbro321: I'd like to avoid tying `load.PackageDefinition` to skeletons.

I think something like below would be a better abstraction. It would also aid the work that @chaospuppy is doing in #4783.
```
type VariantDimension string

const (
    VariantArchitecture VariantDimension = "architecture"
    VariantFlavor       VariantDimension = "flavor"
)

// SkipVariantFilters lists dimensions to retain instead of filtering out.
SkipVariantFilters []VariantDimension

func AllVariantDimension() []VariantDimension
```

AustinAbro321 (changes requested):
Thanks for taking this on, requesting an abstraction

src/pkg/packager/load/load.go
JeffResc: Just pushed a commit to address this, let me know any feedback you have

src/pkg/packager/load/import_test.go
```diff
@@ -521,3 +521,52 @@ func TestCompatibleComponent(t *testing.T) {
...
+			Only: v1alpha1.ZarfComponentOnlyTarget{
+				Cluster: v1alpha1.ZarfComponentOnlyCluster{Architecture: arch},
+			},
+		}
+		require.True(t, compatibleComponent(c, v1alpha1.SkeletonArch, ""),
+			"skeleton compose must retain %s component", arch)
+	}
+}
+
+func TestResolveImportsSkeletonRetainsAllArches(t *testing.T) {
```
AustinAbro321, line 539: This test shouldn't be focused on skeletons (publish_test.go is handling that). Make this test ensure that the variant dimensions work separately and together

src/pkg/packager/load/load.go
```diff
@@ -75,7 +95,11 @@ func PackageDefinition(ctx context.Context, packagePath string, opts DefinitionO
 	if err != nil {
 		return DefinedPackage{}, err
 	}
-	pkg.Metadata.Architecture = config.GetArch(pkg.Metadata.Architecture)
+	if slices.Contains(opts.SkipVariantFilters, VariantArchitecture) {
+		pkg.Metadata.Architecture = v1alpha1.SkeletonArch
```
AustinAbro321, line 99: We still don't want the load package to know about skeletons, instead of setting the architecture, and checking for a skeleton architecture we should pass the variant dimensions into validate and import

AustinAbro321 (changes requested):
Few more separation of concern comments

PR #5022, opened 2026-06-30
chore(deps): introduce 7 day cooldown period for dependabot

soltysh:
## Description
As discussed in various places, we want to introduce a cooldown period to ensure we don't pull dependencies sooner than necessary. This also matches our previous cadence on dependency merges, so this only makes it more official.

## Related Issue

None

## Checklist before merging

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -5,6 +5,8 @@ updates:
       - "/"
     schedule:
       interval: daily
+    cooldown:
+      default-days: 7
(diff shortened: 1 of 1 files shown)


brandtkeller (approved):
lgtm - reviewed the [docs](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-)

a1994sc (approved):

PR #5027, opened 2026-06-30
feat: add component filtering by variable

jsmith1988:
## Description

Adds `only.variable` to component gating, allowing a component to be conditionally included at deploy time based on the resolved value of a package variable or constant. Closes the gap between `only.flavor` (create-time) and `only.localOS` / `only.cluster.architecture` (deploy-time, but environmental) by letting operators drive component selection from `--set` input or declared constants.

Key changes:
- New `only.variable map[string]string` field on `ZarfComponentOnlyTarget` in both `v1alpha1` and `v1beta1`, with schema updates.
- Deploy pipeline reshaped so variables resolve **before** the gating filter and optional-component picker (`packager.Deploy` / `DevDeploy` / init). `ForDeploy` moves out of the `cmd` layer and into `packager.Deploy`; `OptionalComponents` is plumbed through `DeployOptions`.
- New `ByVariable` filter (`src/pkg/packager/filters/variable.go`) with unit tests.
- Compose/import merge semantics: union of `only.variable` maps; conflicting values on the same key are an error (consistent with `only.localOS`).
- Lint: hard error on non-`^[A-Z0-9_]+$` keys; warning when a key isn't declared as a package variable or constant.
- ADR 0027 documenting the design, ordering change, and trade-offs (string-equality comparison, deploy-only scope, UX change around prompt ordering).
- E2E test (`51_component_only_variable_test.go`) and example package demonstrating the feature.

Scope: deploy-time only — not wired into create, mirror, pull, or remove, since deploy variables aren't bound in those contexts and `remove` operates from the recorded deployed-package secret.

## Related Issue

N/A

## Checklist before merging


brandtkeller:
Appreciative of your drawing this up @jsmith1988. After recent discussions - if you believe we're going to iterate on the idea and move forward with a differently-shaped solution could you close this out?

PR #5030, opened 2026-07-01
feat(sign)!: remove legacy signature from production

### LINKED ISSUE #4569: Deprecate legacy signature in favor of bundle

With #4519 completed - zarf now includes BOTH a standard signature (legacy) and a bundle signature (latest) in the zarf package when it is signed. 

These are duplicative and zarf should remove the legacy signature after a minimum of 6 releases while providing deprecation notice to users. 

This will not remove the ability to verify packages that only contain the legacy signature for backwards compatibility.

brandtkeller:
## Description

Originally this started as a bump to cosign to resolve transitive vulnerabilities. What it introduced was a new behavior from Cosign such that `sign_blob` no longer produces both the bundle and the legacy signature when provided paths as inputs. If the bundle path is provided (and the format use boolean is set) then it solely creates the bundle and returns. 

In v0.71.0 we introduced the bundle format and documented that the legacy signature was deprecated. We could have took an additional step to add a warning to the execution logs. 

This PR:
- Bumps Cosign to resolve transitive dependency vulnerabilities - `3.0.6` -> `3.1.1`
- Now only produces the bundle signature during signing
- Handle overwrite of legacy-only to bundle regarding files and metadata
- Updates testing for layout
- Adds testing for signing - existing SDK consumers can still opt to sign without the bundle format (we are aware of such use today).

## Alternatives

We could alternatively still deprecate more loudly and insert a compat shim such that we pull out the certificate from the bundle but it deviates from cosign and creates more code to support (ie risk). I believe we've met the expectation of the minimum 6 releases mentioned in the issue and additionally the signature isn't a file that is yet user (CLI) inspect-able - so reliance on specific formats should be minimal.

## Breaking Changes

CLI/SDK: Zarf package signing now-forward _only_ produces a bundle signature.

SDK: The `CosignSignBlobWithOptions()` command only produces the bundle signature by default - but can be overridden to continue producing the legacy signature _only_. 

## Related Issue

Fixes #4569 

## Checklist before merging


site/src/content/docs/ref/package-signing.mdx
```diff
@@ -26,22 +26,7 @@ Packages are signed using the [Sigstore bundle format](https://docs.sigstore.dev
...
-2. **Non-standard**: Uses a custom format not compatible with other Sigstore tooling
-3. **Reduced functionality**: Cannot support advanced features like keyless signing or transparency log integration
-4. **Maintenance burden**: Supporting multiple formats increases complexity and testing requirements
-
-**Migration:**
-
-Re-sign packages with legacy signatures using the `zarf package sign --features="bundle-signature=true" --overwrite` command to upgrade to the bundle format. The `--overwrite` flag allows replacing existing signatures.
+Re-sign packages with legacy signatures using the `zarf package sign --overwrite` command to upgrade to the bundle format. The `--overwrite` flag allows replacing existing signatures.
 
 :::
```
AustinAbro321:
```suggestion
```

src/pkg/packager/layout/package.go
```diff
@@ -292,30 +278,21 @@ func (p *PackageLayout) SignPackage(ctx context.Context, opts signing.SignBlobOp
 		return fmt.Errorf("failed to write temp %s: %w", ZarfYAML, err)
 	}
 
-	// Configure signing to write to temp directory
+	// Configure signing. cosign v3.1.1+ writes only the bundle when NewBundleFormat=true.
 	signOpts := opts
-	signOpts.NewBundleFormat = bundleEnabled
+	signOpts.NewBundleFormat = true
```
AustinAbro321, line 283: The description says that this PR allows SDK users to sign with the legacy format, but doesn't this mean that they'd have to sign with the new bundle format? 

(FWIW I'm not totally against completely deprecating the legacy format, but want to understand current behavior)

brandtkeller: Correct. Zarf package signing will now only sign/produce a bundle signature going forward. 

SDK users _can_ still set `NewBundleFormat` to `false` using the `signing` package if they are signing anything using those generic utilities.

AustinAbro321: Gotcha, I'm in favor of this strategy. Please clarify the description and denote this PR as a breaking change.

AustinAbro321 (approved):

Issue #5004, opened 2026-06-25
Deprecate and Remove `zarf tools registry export`


AustinAbro321:
### Is your feature request related to a problem? Please describe.
In #5003 we are planning to remove the reliance on Crane for `zarf tools registry`. The only command that I feel should be entirely removed is `zarf tools registry export`. This command is centers around working with container images rather than working with a registry. While it has it's uses, I believe these will be niche in the context of Zarf, and not worth the complexity.

It should be deprecated and removed in v1.0. 

If you are using this feature, and feel strongly that it should stay, please comment

PR #5031, opened 2026-07-01
feat(deprecate): zarf tools registy export

AustinAbro321:
## Description

Deprecates `zarf tools registry export`. This command will not be replaced

## Related Issue

Relates to #5004

## Checklist before merging


brandtkeller (approved):
Im not sure if `version` really has a role to play in a future migration but it's a negligible change nonetheless. lgtm.

src/cmd/crane.go
```diff
@@ -48,7 +48,7 @@ func newRegistryCommand() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:     "registry",
-		Aliases: []string{"r", "crane"},
```
soltysh, line 51: My only worry is that this might break some users, that did use that alias. Are we sure this is not being used?

soltysh (commented):
One non-blocking comment.

PR #5035, opened 2026-07-02
docs(search): load pages from currently selected version

AustinAbro321:
## Description
 
Right now there is a bug on the docs site where looking up any keyword shows results the same results for a bunch of different versions. We should instead only show the currently selected version

## Checklist before merging

### DIFF
--- a/site/src/components/Search.astro
+++ b/site/src/components/Search.astro
@@ -14,41 +14,27 @@ import Default from '@astrojs/starlight/components/Search.astro';
 
   const pageVersion = versionFromPath(window.location.pathname);
 
-  function activateVersionFilter(container: HTMLElement): boolean {
-    const input = container.querySelector(
-      `.pagefind-ui__filter-panel input[value="${pageVersion}"]`
-    ) as HTMLInputElement | null;
-    if (!input || input.checked) return false;
-    input.click();
-    return true;
+  // Pagefind lazy-loads its search UI and renders the version filter checkbox
+  // asynchronously, so we watch for it to appear. One click is enough: Pagefind
+  // holds the selection in its own state, which survives later searches and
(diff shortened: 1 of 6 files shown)


site/src/components/Search.astro
```diff
@@ -0,0 +1,40 @@
...
+  /* Version filter panel is hidden — version scoping is applied programmatically. */
+  .pagefind-ui__filter-panel { display: none !important; }
+</style>
+
+<script>
+  import { versionFromPath } from '../lib/version.ts';
+
+  const pageVersion = versionFromPath(window.location.pathname);
+
+  // Pagefind lazy-loads its search UI and renders the version filter checkbox
```
brandtkeller, line 17: Digging around - I wonder if we could use PagefindUI (the class from @pagefind/default-ui) as it exposes a public instance method, triggerFilters(filters)? 

May be negligible but rules out DOM query, simulated click, and waiting on render timing.

brandtkeller (changes requested):
One consideration otherwise this is entirely functional.

site/src/components/Search.astro
AustinAbro321: From what I can tell, Starlight discards the PagefindUI instance after constructing it (new PagefindUI(...) with no assignment), so we have no reference to call triggerFilters on.

brandtkeller: Im not married to this idea, but If we fork Search.astro instead of wrapping it with <Default />, we do the new PagefindUI(...) call ourselves and just don't discard it?

AustinAbro321: Yeah I think we could do that, but without a measurable benefit, I'd prefer to keep it simple.

brandtkeller (approved):
lgtm

PR #5037, opened 2026-07-02
fix: update oras-go and mitigation

### LINKED ISSUE #5044: oras upgrade exposes latent bug in PlainHTTP application
### Description

In an attempt to resolve and mitigate vulns in oras-go, updates promptly broke e2e testing due to a bleed between CLI-level intent (where to push a package once created) versus package artifact source. `--plain-http` applies globally to both and recent security updates across dependency projects are now returning errors in the case of sending an http request to an https endpoint. Originally this redirected gracefully but utlimately was susceptible to downgrade attacks and now returns an explicit 401 error.

### Steps to reproduce
1. Upgrade `oras-go` to `v2.6.1`
2. Run the E2E test suite

### Expected result

All tests pass

### Actual Result

Tests that perform composability or other local OCI operations from local registries using the `--plain-http` flag while also packaging upstream resources (ghcr.io etc) will fail now. 

### Severity/Priority

High - prevents upgrades to oras-go and other dependencies

### Additional Context

Evaluating changes in #5037 - expecting that dependency review continues to flag the mitigated but not-fixed finding still present in v2.6.1

brandtkeller:
## Description

There are currently five open advisories for oras-go - bumping to `v2.6.1` resolves the following four:

Advisory | Severity | Summary
-- | -- | --
GHSA-jxpm-75mh-9fp7 | High | Blob upload vulnerable to credential forwarding via unvalidated Location header
GHSA-8xwf-rjm4-xvhv | Medium | File store write outside workingDir via symlink traversal
GHSA-vh4v-2xq2-g5cg | Medium | Forwards registry credentials across registry redirects
GHSA-xf85-363p-868w | Low | Malicious registry can hijack Bearer token realm to exfiltrate credentials and refresh tokens

The fifth has not been included in a release: GHSA-fxhp-mv3v-67qp

In the normal zarf lifecycle, we are not susceptible to the use of the unpack annotation, as we never set it on push nor use on pull. 

[file.go:275](https://github.com/oras-project/oras-go/blob/v2.6.1/content/file/file.go#L275-L279) in push():

```golang
if needUnpack := expected.Annotations[AnnotationUnpack]; needUnpack == "true" && !s.SkipUnpack {
    err = s.pushDir(name, target, expected, content)   // ← vulnerable path
} else {
    err = s.pushFile(target, expected, content)         // ← safe path
}
```
With SkipUnpack = true, the && !s.SkipUnpack term is always false, so this branch can never select pushDir, regardless of what annotation a malicious registry sends on the descriptor.

I believe we can mitigate now and keep explicit use of `SkipUnpack` going forward. 

## Related Issue

Fixes #5044

## Checklist before merging


go.mod
```diff
@@ -58,7 +58,7 @@ require (
 	k8s.io/klog/v2 v2.140.0
 	k8s.io/kubectl v0.36.0
 	k8s.io/streaming v0.36.1
-	oras.land/oras-go/v2 v2.6.0
+	oras.land/oras-go/v2 v2.6.1
```
brandtkeller, line 61: expect the `Dependency Review` action to still fail here.

vendor/oras.land/oras-go/v2/registry/remote/auth/client.go
```diff
@@ -207,6 +297,13 @@ func (c *Client) Do(originalReq *http.Request) (*http.Response, error) {
 	if resp.StatusCode != http.StatusUnauthorized {
 		return resp, nil
 	}
+	// If the challenge came from a different origin than originally requested
+	// (e.g. the request was redirected to another host or port), do not resolve
+	// or send the registry credentials to that origin.
+	// Reference: https://github.com/oras-project/oras-go/security/advisories/GHSA-vh4v-2xq2-g5cg
+	if resp.Request != nil && !sameHTTPOrigin(originalReq.URL, resp.Request.URL) {
```
brandtkeller, line 304: note the root cause of our issue here - `sameHTTPOrigin` is going to return false as the original `http://` won't match the redirect for `https://` so we get the raw 401 directly.

src/pkg/transport/transport.go
```diff
@@ -0,0 +1,254 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+// Package transport decides whether a registry should be reached over plain HTTP or
```
brandtkeller, line 4: Leaving this commentary overly verbose to overlay details about the approach and decisions made. Open to consolidating if needed.

brandtkeller:
split into #5047 and #5049

PR #5040, opened 2026-07-07
docs: update Azure KeyVault URL for package signing

niklashempel:
Removed path 'keys' from Azure KeyVault URL

## Description

When following the docs to sign a package using Azure KeyVault, the specified URL is not correct. If followed, an error occurs:

```
Error: kms get: kms specification should be in the format azurekms://[VAULT_NAME][VAULT_URL]/[KEY_NAME]/[VERSION (optional)]
```

The solution is to not include `keys` after the vault URL.
...

## Checklist before merging

### DIFF
--- a/site/src/content/docs/ref/package-signing.mdx
+++ b/site/src/content/docs/ref/package-signing.mdx
@@ -75,7 +75,7 @@ Requires `gcloud` authentication and appropriate IAM roles. The key must be an a
 
 ```bash
 zarf package sign zarf-package-example-amd64.tar.zst \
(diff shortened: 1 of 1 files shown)


brandtkeller (approved):
lgtm - comparing against the upstream azure [readme](https://github.com/sigstore/sigstore/blob/main/pkg/signature/kms/azure/README.md?plain=1#L23) in sigstore. 

Thanks for fixing this - much of this has been guess work for the project as we don't have an azure environment for testing.

soltysh (approved):
lgtm

jonnyborbs (approved):
Thanks for the update, we appreciate contributions like these. LGTM

Issue #5005, opened 2026-06-25
Deprecate and Remove Zarf artifact Server


AustinAbro321:
### Is your feature request related to a problem? Please describe.
Zarf has an artifact server that natively integrates with Gitea and can be specified on `zarf init`. I believe it's likely that no current Zarf users are using this, or few enough are, that it is not worth the maintenance burden and extra CLI flags to the user.  

We should deprecate then remove in v1.0

If you are using this feature, and feel strongly that it should stay, please comment

PR #5041, opened 2026-07-07
feat: deprecate artifact server

AustinAbro321:
## Description

This deprecates the artifact server. It will not be replaced, it was built for a specific user group whom I believe no longer uses Zarf. I suspect it has little to no users currently. It also deprecates `zarf internal http-proxy`, which was created to push and pull to the artifact server

I considered also writing a deprecation message when users stand up the init package gitea server (since this automatically configures the artifact server) however because there would be a lot of false positives, and because all users taking advantage of the artifact server will run one of: 
- `zarf tools get-creds` 
- `zarf init with --artifact-url --artifact-user --artifact-pass`
- `zarf internal http-proxy`

they should all still see the deprecation message. 

## Related Issue

Relates to #5005

## Checklist before merging


brandtkeller (approved):
lgtm

PR #5047, opened 2026-07-08
feat(ocischeme)!: support per-host transport negotiation for plain http

### LINKED ISSUE #5044: oras upgrade exposes latent bug in PlainHTTP application

brandtkeller:
## Description

This replaces Zarf's --plain-http handling with a new src/pkg/transport package that decides, per registry host, whether to speak plain HTTP or HTTPS rather than applying one global flag to every registry a command happens to touch (third-party image refs, Helm chart URLs, chart OCI dependencies, cosign lookups — none of which are "the registry the flag was meant for").

This has implications for upgrading to oras-go >= v2.6.1 in that we need to upgrade for various security fixes as well as consider the impact across zarf workflows (of which create/deploy) are impacted. 

**NOTE:**: negotiation only happens when `--plain-http / PlainHTTP` is supplied as `true`. Otherwise the expectation is to honor the existing workflows and expect https. 

### CLI user impact
- --plain-http behaves the same for the registry it's most obviously "for" (internal Zarf registry via zarf package deploy) — that path still keeps its existing localhost auto-detection.
- For everything else (third-party images/charts/dependencies), --plain-http no longer forces the scheme directly — it now gates a per-host probe, and a registry that only speaks HTTPS is unaffected either way.
- A registry that only speaks plain HTTP now requires both --plain-http set and a passing probe; it's no longer possible to force plain HTTP onto something that fails the scheme check.
zarf package create/dev pulling images from a local insecure registry (e.g. localhost:5000) still works without the flag, unchanged from before.
### SDK user impact (breaking)
- New public package: github.com/zarf-dev/zarf/src/pkg/transport — the Negotiator API (New, UsePlainHTTP, Invalidate, WithNegotiator/From for context-based DI). Long-running SDK consumers should install their own Negotiator via transport.WithNegotiator (with a TTL matching their process lifetime); otherwise they transparently get a shared 5-minute-TTL singleton.
- src/pkg/utils.GetCosignArtifacts signature changed: (ctx, image, client, insecureSkipTLSVerify bool) → (ctx, image, client, remoteOptions types.RemoteOptions). Direct callers must update.
- src/pkg/images.PullOptions.PlainHTTP / PushOptions.PlainHTTP gates per-host negotiation instead of forcing the scheme unconditionally.

## Testing

This is a fix for upgrading oras-go to >= v2.6.1. If you did so on main and ran E2E testing it would fail. Upgrading on this branch resolves those failures - but I split out the upgrade for simplicity of review.

## Related Issue

Fixes #5044

Relates to #

## Checklist before merging


src/pkg/ocischeme/ocischeme.go
```diff
@@ -0,0 +1,265 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
```
brandtkeller, line 3: Leaving the overly verbose commentary here in the event it fills some context. Open to refining if needed.

src/internal/packager/helm/template_test.go
```diff
@@ -39,3 +42,42 @@ func TestChartTemplate(t *testing.T) {
 	require.NoError(t, err)
 	require.YAMLEq(t, string(b), manifest)
 }
+
+func TestChartTemplate_DoesNotNegotiateDeclaredOCIDependencies(t *testing.T) {
```
AustinAbro321: This test attempts to check a behavior of Helm, but I'm not sure it's valuable. Do we know that if Helm did probe for OCI resources it would hang and take longer than 2 seconds?

src/pkg/transport/transport.go
```diff
@@ -0,0 +1,265 @@
...
+// HTTPS (breaking the fetch), or leave a registry that only speaks plain HTTP
+// unreachable because the flag wasn't set for an unrelated reason.
+//
+// This package answers the question per host instead: it probes the host directly
+// and only ever falls back to plain HTTP when the host proves it speaks plaintext
+// HTTP on that port. It never falls back on a timeout, a refused connection, a TLS
+// certificate error, or an ordinary non-2xx status code — none of those prove
+// anything about the correct scheme, and treating them as license to downgrade would
+// make Zarf a protocol-downgrade oracle.
+package transport
```
AustinAbro321: I think we should find a more intuitive name than transport:
- we alias it to negotiate in every spot 
- This package is specifically meant to work with oci registries, but nothing makes that obvious. For instance, all the comments in `src/internal/packager/helm/repo.go` state that the scheme for chartURL is used in place of this package, but this code would never work with that anyway

Alternatively, we could add this to the images package

src/internal/packager/helm/repo.go
```diff
@@ -30,13 +30,73 @@ import (
...
+	if err != nil {
+		return false, fmt.Errorf("unable to reach chart registry for %q: %w", ociURL, err)
+	}
+	return plainHTTP, nil
+}
+
+// negotiateChartDependenciesPlainHTTP loads the chart from disk and negotiates its
+// OCI dependency hosts; see negotiateLoadedChartDependenciesPlainHTTP.
+func negotiateChartDependenciesPlainHTTP(ctx context.Context, chart v1alpha1.ZarfChart, remoteOptions types.RemoteOptions) (bool, error) {
+	_, parsed, err := loadAndValidateChart(chart.LocalPath)
```
AustinAbro321: We always load before this call, it would be cleaner to pass in the `parsed.Metadata.Dependencies` to this function

src/internal/agent/hooks/common.go
```diff
@@ -115,19 +123,23 @@ func getManifestConfigMediaType(ctx context.Context, zarfState *state.State, tra
...
 	if err != nil {
 		return "", err
 	}
 
-	registry := &orasRemote.Repository{
-		PlainHTTP: plainHTTP,
-		Reference: ref,
-		Client:    client,
+	b, err := fetchManifestBytes(ctx, ref, client, plainHTTP, imageAddress)
+	if err != nil && isTransportSchemeFailure(err) {
```
AustinAbro321: could we add a unit test for this functionality? cache result -> upgrade server to https -> call again

AustinAbro321 (changes requested):
We should mark this as a breaking change, we don't know if there are registries that don't respond to a plain-http call in the way this PR expects. I think the code / description that this negotiation is only for OCI calls as I was confused by that on first review.

src/pkg/transport/transport.go
brandtkeller: Fair! i'm terrible with naming. How do we feel about something like `ocitransport`? we can then remove aliases and it feels descriptive?

src/internal/packager/helm/template_test.go
brandtkeller: Good point - I updated it to test whether TemplateChart made any negotiation connections but rather feel it's not testing zarf logic - removing.

src/internal/packager/helm/repo.go
brandtkeller: great point!

src/internal/agent/hooks/common.go
brandtkeller: Added!

src/pkg/transport/transport.go
AustinAbro321: Naming is hard! How about ocischeme? I think ocitransport implies that the package provides a transport

brandtkeller: Updated!

src/internal/agent/hooks/common.go
```diff
@@ -9,22 +9,29 @@ import (
...
 	"oras.land/oras-go/v2/registry"
 	orasRemote "oras.land/oras-go/v2/registry/remote"
 	"oras.land/oras-go/v2/registry/remote/auth"
+	orasRetry "oras.land/oras-go/v2/registry/remote/retry"
 )
 
+// transportNegotiator decides plain-HTTP vs. HTTPS for the internal Zarf registry as
+// seen from in-cluster. Unlike a CLI invocation, the agent is a long-running process,
+// so decisions are cached with a bounded TTL rather than for the process lifetime.
+var transportNegotiator = ocischeme.New(ocischeme.Options{TTL: 5 * time.Minute})
```
AustinAbro321: What would you think of removing this global and making this a local variable created on every call? I think the complexity the cache adds is not worth the time saved. I'm also weighing that we retry on every error case, and any future registry calls in the agent would need to add that retry logic. 

Those pings already happened on main, so this PR wouldn't make things any slower. 

(sorry I should have commented this before the test ask)

brandtkeller: I saw the cache as a nice-to-have improvement for the long-lived process side, but I am not inherently against removal for sake of early iteration and reducing complexity now on something that can be improved in the future.

brandtkeller: This does leave `Invalidate()` as having no current utilization in-tree. I haven't removed it yet as it still provides SDK utility but I could be convinced to remove it and the associated logic.

src/internal/agent/hooks/common.go
```diff
@@ -115,19 +116,37 @@ func getManifestConfigMediaType(ctx context.Context, zarfState *state.State, tra
...
+		// Re-probe regardless of what kind of error this was: re-probing is cheap (a
+		// single fast request via the unwrapped transport) and doesn't depend on
+		// recognizing every shape a "the scheme changed" failure can take across
+		// registry implementations — some, e.g. Go's own http.Server, reject a
+		// plaintext request on a TLS-only port with a well-formed but ordinary-looking
+		// error response that's indistinguishable by error type from an unrelated
+		// failure. If the fresh probe agrees with what was already used, this error is
+		// unrelated to scheme and is reported as-is; if it disagrees, the scheme
+		// changed underneath us and it's worth one retry with the corrected value.
+		if fresh, negotiateErr := probe(); negotiateErr == nil && fresh != plainHTTP {
```
AustinAbro321: I believe we can delete this re-probe. The scheme changing between the first probe and the call to fetchManifestBytes feels unrealistic

AustinAbro321 (changes requested):

src/internal/agent/hooks/common.go
brandtkeller: I'm good with that - simplifies versus potentially unnecessary defense in depth.

AustinAbro321: With that, I think we can delete the tests introduced in common_test.go and the changes in src/test/testutil/registry.go

AustinAbro321 (approved):

PR #5048, opened 2026-07-08
feat(v1beta1): load v1beta1 packages

### LINKED ISSUE #5164: Load v1beta1 packages
### Description

User's should be able to load v1beta1 packages with `zarf dev inspect definition`. 

Other functionality such as importing remote components and `zarf dev *` commands are not included in this scope. 

Relates to #3433

Follows implementation from https://github.com/zarf-dev/proposals/tree/main/0048-schema-update-process and  https://github.com/zarf-dev/proposals/tree/main/0051-v1beta1-schema

AustinAbro321:
## Description

Implements loading v1beta1 packages. Does not implement remote components. Create a new `api` package to hold the `PackageAccessor` interface as well as the `PackageDefintion` concrete type. Most of the SDK still uses a v1alpha1 package. 

A fair amount of duplication / split paths between schema versions in load.PackageDefinition, but it should be unique since it's so concerned with imports, validation, and the actual definition, most other functions should use PackageDefinition / PackageLayout. 

## SDK breaking changes

 - load.PackageDefinition now returns load.ResolvedPackage instead of load.DefinedPackage.
   - SkipRequiredValues is no longer accepted. Required values should always be skipped when loading a package definition.
 - assemble.AssemblePackage now accepts load.ResolvedPackage instead of v1alpha1.ZarfPackage and []importedSchema.

## Related Issue

Fixes #5164

Relates to #3433

## Checklist before merging


src/pkg/packager/load/load.go
```diff
@@ -30,9 +35,6 @@ import (
 type DefinitionOptions struct {
 	Flavor       string
 	SetVariables map[string]string
-	// SkipRequiredValues ignores values schema validation errors when a "required" field is empty. Used when a package
-	// value should be supplied at deploy-time and doesn't have a default set in the package values.
-	SkipRequiredValues bool
```
brandtkeller, line 35: Is removing this intentional/required? I see the replacement now as a hard-coded below as `skipRequired: true` which would be a slight change in behavior for the operations which previously used the type default of `false`.

AustinAbro321: Yeah definitely intentional, and called out in the breaking changes. Sorry to bury it in such a large PR, but we always set this to true when calling the function. It doesn't make sense to require that values be set when loading the package IMO

src/pkg/packager/load/import_v1beta1.go
```diff
@@ -0,0 +1,234 @@
...
+			}
+		}
+		config, err := readComponentConfig(path)
+		if err != nil {
+			return loadedComponentConfig{}, err
+		}
+		loaded = append(loaded, loadedComponentConfig{config: config, entry: entry, dir: filepath.Dir(path), path: path})
+	}
+
+	if len(loaded) == 1 {
```
brandtkeller: this will short-circuit compatibility checks correct? so I have only 1 component - even if it's the wrong variant - it is still selected?

src/pkg/packager/load/import_v1beta1.go
```diff
@@ -0,0 +1,234 @@
...
+
+	baseDir := pkgPath.BaseDir
+
+	var components []v1beta1.Component
+	var vals importedValues
+	for _, component := range pkg.Components {
+		if !compatibleComponentV1Beta1(component.Selector, arch, flavor) {
+			continue
+		}
+		mergedSpec, compVals, err := resolveComponentSpecImports(ctx, component.ComponentSpec, baseDir, arch, flavor, []string{filepath.Clean(pkgPath.ManifestFile)})
```
brandtkeller, line 44: Probably not a requirement - but I was thinking about about what we lose when passing the spec when only `ComponentSpec` is passed. Really just comes down to name - do we care if the parent component name is not the same as the imported component?

src/pkg/packager/load/load.go
brandtkeller: Ah yes - there it is. 🤔 I see a number of `Load.PackageDefinition` calls that relied on the default `false` value but I don't know that I disagree with setting this explicitly to true either given the intent of those operations. So I think this is fair.

src/pkg/packager/load/load.go
```diff
@@ -64,48 +66,93 @@ func PackageDefinition(ctx context.Context, packagePath string, opts DefinitionO
...
+
+	pkg, importedSchemas, err := resolveImportsV1Beta1(ctx, pkg, pkgPath, pkg.Metadata.Architecture, opts.Flavor)
 	if err != nil {
-		return DefinedPackage{}, err
+		return ResolvedPackage{}, err
 	}
-	l.Debug("done layout.LoadPackage", "duration", time.Since(start))
-	return DefinedPackage{Pkg: pkg, ImportedSchemas: importedSchemas}, nil
+
+	if err := validateV1Beta1(ctx, pkg, pkgPath.ManifestFile, opts.Flavor, opts.SkipValuesSchemaValidation); err != nil {
```
brandtkeller: We have `importedSchemas` from the `resolveImportsV1Beta1` function above - but this function only evaluates the top-level schema (parent) - is that sufficient?

src/pkg/packager/load/import_v1beta1_merge.go
```diff
@@ -0,0 +1,212 @@
...
+		m := out[idx]
+		if h.Namespace != "" {
+			m.Namespace = h.Namespace
+		}
+		m.Files = append(m.Files, h.Files...)
+		if h.Kustomize != nil {
+			if m.Kustomize == nil {
+				m.Kustomize = h.Kustomize
+			} else {
+				m.Kustomize.Files = append(m.Kustomize.Files, h.Kustomize.Files...)
```
brandtkeller, line 127: we need to merge more than just the files correct? The `KustomizeManifest` type also has two boolean fields?

src/pkg/packager/load/import_v1beta1_merge.go
```diff
@@ -0,0 +1,212 @@
...
+func mergeActions(base, head v1beta1.ComponentActions) v1beta1.ComponentActions {
+	return v1beta1.ComponentActions{
+		OnCreate: mergeActionSet(base.OnCreate, head.OnCreate),
+		OnDeploy: mergeActionSet(base.OnDeploy, head.OnDeploy),
+		OnRemove: mergeActionSet(base.OnRemove, head.OnRemove),
+	}
+}
+
+func mergeActionSet(base, head v1beta1.ComponentActionSet) v1beta1.ComponentActionSet {
+	base.Defaults = head.Defaults
```
brandtkeller: Should this be conditional based on whether `head.Defaults != nil` ? Would an importing (head) component that does not declare defaults not then zero-out the importing (base) component?

src/pkg/packager/load/load.go
```diff
@@ -64,48 +66,93 @@ func PackageDefinition(ctx context.Context, packagePath string, opts DefinitionO
...
+	version, err := pkgcfg.SelectVersion(ctx, b)
 	if err != nil {
-		return DefinedPackage{}, err
+		return ResolvedPackage{}, err
 	}
+
+	var defined ResolvedPackage
+	switch version {
+	case v1beta1.APIVersion:
+		pkg, err := pkgcfg.ParseAs(ctx, b, pkgcfg.V1Beta1)
```
brandtkeller, line 86: nit: both `pkgcfg.ParseAs` and `pkgcfg.SelectVersion` call `parseZarfYAMLDocs - so we're parsing the root manifest twice.

src/internal/api/v1alpha1/migrate.go
```diff
@@ -0,0 +1,207 @@
...
+	// Record the migrations that have been run on the package.
+	pkg.Build.Migrations = []string{
+		ScriptsToActionsMigrated,
+		PluralizeSetVariable,
+	}
+
+	return pkg, warnings
+}
+
+// migrateScriptsToActions coverts the deprecated scripts to the new actions
```
brandtkeller:
```suggestion
// migrateScriptsToActions converts the deprecated scripts to the new actions
```

brandtkeller (changes requested):
This is huge (as in a ton of effort and progress). Appreciate the patience as I came up to speed. My focus is on evaluating risk to current releases and schema (low) first and then moving over to v1beta1 functionality which can very well evolve throughout the parent. 

I've got some comments - I don't see them as blocking if the intent is to evolve some of this as we intregrate the remaining lifecycle. I've reviewed all files so post any changes I am ready to approve.

src/pkg/packager/load/import_v1beta1.go
AustinAbro321: Great catch, yeah that was a bug

src/pkg/packager/load/import_v1beta1_merge.go
AustinAbro321: Yes

src/pkg/packager/load/import_v1beta1.go
AustinAbro321: I intentionally allowed differing names. Previously, names mattered since you could import one of several components in a zarf.yaml, but now there is only one component per config.

src/pkg/packager/load/load.go
AustinAbro321: Yeah good catch, I changed it so that we evaluate against all the imported schemas. Also added this for v1alpha1. Technically a breaking change, but I think a very reasonable one (as they should run into the breakage either way, unless the deployer carefully changes the values)

src/pkg/packager/load/load.go
AustinAbro321: Yeah, this is a tradeoff that I accept given how cheap it is. I think it makes for more logical interactions with the pkgcfg package. However, if you have an idea for how to restructure pkgcfg, happy to try it out.

src/api/api.go
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+// Package api supplies helpers for working generically with the different API versions
+package api
+
+import (
+	"github.com/zarf-dev/zarf/src/api/v1alpha1"
+	"github.com/zarf-dev/zarf/src/api/v1beta1"
+	internalTypes "github.com/zarf-dev/zarf/src/internal/api/types"
```
soltysh:
```suggestion
	internaltypes "github.com/zarf-dev/zarf/src/internal/api/types"
```
Using lowercase for import aliases is usually flagged by linters

src/api/api.go
```diff
@@ -0,0 +1,61 @@
...
+func (p PackageDefinition) AsV1beta1() v1beta1.Package {
+	return internalv1beta1.ConvertFromGeneric(p.pkg)
+}
+
+// OriginalAPIVersion returns the apiVersion the package was authored in before any conversion.
+func (p PackageDefinition) OriginalAPIVersion() string {
+	return p.pkg.Build.OriginalAPIVersion
+}
+
+// SetName sets the package metadata name.
```
soltysh, line 53: I'd suggest not to expose any additional methods here. I'm worried that these two will provide the excuse for other ones.

src/pkg/packager/load/import_v1beta1_merge.go
AustinAbro321: I went back and forth on this. Not considering child defaults is what v1alpha1 does. I see the argument for this strategy, but I think it does break down, while the child is the only one with actions. 

Another option is to make each field a pointer on defaults, and do a per-field merge. However, I think the risk of confusion is too high. 

Ended up making defaults a pointer as you suggested.

src/pkg/packager/load/load.go
AustinAbro321: Ah yeah that's true I forgot about the implicit default. But yeah still makes sense to remove.

src/internal/api/v1alpha1/migrate.go
```diff
@@ -0,0 +1,211 @@
...
+	defaults := v1alpha1.ZarfComponentActionDefaults{
+		// ShowOutput (default false) -> Mute (default false)
+		Mute: !c.DeprecatedScripts.ShowOutput,
+		// TimeoutSeconds -> MaxSeconds
+		MaxTotalSeconds: c.DeprecatedScripts.TimeoutSeconds,
+	}
+
+	// Retry is now an integer vs a boolean (implicit infinite retries), so set to an absurdly high number
+	if c.DeprecatedScripts.Retry {
+		defaults.MaxRetries = math.MaxInt
```
soltysh: Honestly, I'd prefer putting a cap, rather than using `MaxInt` here. Not sure how well this might go through json conversion.

src/internal/api/v1alpha1/migrate.go
```diff
@@ -0,0 +1,211 @@
...
+		return c, fmt.Sprintf("Component '%s' is using scripts which will be removed in the next schema version. Please migrate to actions.", c.Name)
+	}
+
+	return c, ""
+}
+
+func migrateSetVariableToSetVariables(c v1alpha1.ZarfComponent) (v1alpha1.ZarfComponent, string) {
+	hasSetVariable := false
+
+	migrate := func(actions []v1alpha1.ZarfComponentAction) []v1alpha1.ZarfComponentAction {
```
soltysh: I'm not a fan of embedded functions, why not just extract that into `migrateComponentAction`?

src/internal/api/v1beta1/validate.go
```diff
@@ -0,0 +1,227 @@
...
+
+	err = errors.Join(err, validateActionSet(a.OnDeploy))
+	err = errors.Join(err, validateActionSet(a.OnRemove))
+
+	return err
+}
+
+// hasSetValues returns true if any of the actions contain setValues.
+func hasSetValues(as v1beta1.ComponentActionSet) bool {
+	check := func(actions []v1beta1.ComponentAction) bool {
```
soltysh: Here and in the following functions you could use `slices.ContainsFunc` instead.

src/internal/api/v1beta1/validate.go
```diff
@@ -0,0 +1,227 @@
...
+	validate(as.OnFailure)
+	validate(as.OnSuccess)
+	return err
+}
+
+// validateAction runs all validation checks on an action.
+func validateAction(action v1beta1.ComponentAction) error {
+	var err error
+
+	if action.Wait != nil {
```
soltysh: If you flip the condition:
```golang
if action.Wait == nil {
    return nil
}
...
``` 
the code will be easier to read.

Further more, why the validate function have to return flat error, rather than a list of errors? This would make the entire thing easier to read.

src/pkg/packager/load/import_v1beta1.go
```diff
@@ -0,0 +1,229 @@
...
+	path   string
+}
+
+// selectImportVariant loads every local import entry and selects the single one compatible with the
+// active target. Entries are treated as variants: exactly one must be compatible with the target.
+func selectImportVariant(entries []v1beta1.ComponentImportLocal, specDir, arch, flavor string, importStack []string) (loadedComponentConfig, error) {
+	var loaded []loadedComponentConfig
+	for _, entry := range entries {
+		path := filepath.Clean(filepath.Join(specDir, entry.Path))
+		for _, seen := range importStack {
```
soltysh: You could replace this entire loop with `slices.Contains`

src/pkg/packager/load/import_v1beta1_merge.go
```diff
@@ -0,0 +1,223 @@
...
+	if head.Defaults != nil {
+		base.Defaults = head.Defaults
+	}
+	base.Before = append(base.Before, head.Before...)
+	base.OnSuccess = append(base.OnSuccess, head.OnSuccess...)
+	base.OnFailure = append(base.OnFailure, head.OnFailure...)
+	return base
+}
+
+func indexByName(n int, nameAt func(int) string, name string) int {
```
soltysh: You could replace this with `slices.IndexFunc`

src/pkg/packager/load/load.go
```diff
@@ -64,48 +66,93 @@ func PackageDefinition(ctx context.Context, packagePath string, opts DefinitionO
...
+	}
+
+	var defined ResolvedPackage
+	switch version {
+	case v1beta1.APIVersion:
+		pkg, err := pkgcfg.ParseAs(ctx, b, pkgcfg.V1Beta1)
+		if err != nil {
+			return ResolvedPackage{}, err
+		}
+		defined, err = v1beta1PackageDefinition(ctx, pkg, pkgPath, opts)
```
soltysh, line 93: Inside `v1beta1PackageDefinition` you're calling `validateV1Beta1` which again loads pkg, even though you already read them in line 72 above when invoking `os.ReadFile`, why?

src/api/api.go
AustinAbro321: This is part of the design actually. In #5173 we are going to change the PackageLayout to move from a v1alpha1 Package to a PackageDefinition. Right now, this object is fully mutable, and SDK users such as uds-cli take advantage of this mutability. Additionally there are certain fields that we change during/after build, for instance, the build fields during AssemblePackage or filtering components after a package is loaded. I've done an evaluation of common SDK callers and the set of mutated fields is rather small.

The idea being that these objects `load.ResolvedPackage` and `layout.PackageLayout`, which once held a fully mutable package object, now have a much smaller subset of mutability, and mutating fields is API agnostic.

src/internal/api/v1alpha1/migrate.go
AustinAbro321: Sure set to 10,000. Note that all this code was just verbatim moved from pkgcfg, as I felt this was a more sensible home.

src/pkg/packager/load/load.go
AustinAbro321: Excellent question, the reason to get the on disk package is so we can send it into the schema validator without marshalling it to a particular object. This way when someone adds an unsupported field (or more likely mispells a field, the schema reports an error)

For instance, without this check, in the below example `documenttaion` would just be dropped, and the user wouldn't notice, with this check, they error. 
```yaml
kind: ZarfInitConfig
metadata:
  name: init

documenttaion:
  zarf.cli.openvex.json: .vex/zarf.cli.openvex.json
```

src/pkg/packager/load/import_v1beta1.go
AustinAbro321: I now know many slices functions to use next time :smile:

src/internal/api/v1beta1/validate.go
AustinAbro321: Flipped the condition. 

The reason for the list of errors, is so if a user has multiple validation errors, then all are presented. For instance:

```bash
failed to create package: package validation failed: invalid chart definition: chart "podinfo-local" must include a chart version
chart "podinfo-local" excludePath ".podinfo.image" must be a descendant of sourcePath ".podinfo.replicaCount"
chart name "podinfo-local" is not unique
```

brandtkeller (dismissed):
lgtm

src/internal/api/v1beta1/validate.go
soltysh: Right, but I'm saying our API could be clearer if we expose the list of errors, either directly or through a alias type, which will handle the error list.

src/pkg/packager/load/load.go
soltysh: Hmmm... why can't we separate the loading errors which I consider that unsupported field as one, from actual validation which should only focus on the struct itself?

src/internal/api/v1beta1/validate.go
AustinAbro321: Ah yes, good callout

src/pkg/packager/load/load.go
AustinAbro321: Fair enough, moved the functions that validate directly on the schema down a level

soltysh (approved):
Two non-blocking comments:
1. using list of errors in validation
2. separating loading errors from validation.

Otherwise this lgtm

PR #5051, opened 2026-07-09
feat(cmd): add viper option for cache

AustinAbro321:
## Description

We deprecated `--zarf-cache` and re-named it to `--cache`, but never added an equivalent viper config

## Checklist before merging


brandtkeller (approved):
lgtm

PR #5054, opened 2026-07-09
fix(mTLS): prevent negotiation when portforward with mTLS

### LINKED ISSUE #5053: Flaky Test: k3d Proxy test and http negotiation

registry-mode=proxy uses mTLS, which already deterministically proves the internal registry speaks HTTPS. But push.go and flux-ocirepo.go (via getManifestConfigMediaType) ran the ocischeme plain-HTTP negotiation unconditionally anyway, adding a redundant HTTPS probe over the same kubectl port-forward tunnel the real request depends on. 

When that tunnel is transiently flaky (as shown in the raw k8s logs — the registry-proxy pod's port-forward hitting "broken pipe" / "connection refused"), the probe fails with a bare EOF, an error shape decide() correctly refuses to interpret as proof of anything.

 it hard-fails with "refusing to downgrade to plain HTTP" even though plain HTTP was never a real possibility and the scheme was never actually in question.

### Additional context

https://github.com/zarf-dev/zarf/actions/runs/29056159091/job/86248680350#step:7:696

brandtkeller:
## Description

post oras-go and negotation logic update we're seeing some flaky testing. I dont believe there is any signals indicating that the system is producing fatal errors where it previously was unable to recover. 

Rather the failing tests indicate probes that were failing in scenarios where we deterministically know the scheme and do not need to probe.

both call sites now skip the negotiation entirely when mTLS is in play, hardcoding plainHTTP = false instead. 

This matches the pattern already established by argocd-repository.go and flux-helmrepo.go, which treat useMTLS as a deterministic fact rather than something to probe for.

One outlier is whether the test failure mentioned in the issue is the result of a slow rollout after updating mtls certs and rolling the pods?

## Related Issue

Fixes #5053

## Checklist before merging

### DIFF
--- a/src/pkg/ocischeme/ocischeme.go
+++ b/src/pkg/ocischeme/ocischeme.go
@@ -105,6 +105,21 @@ func New(o Options) *Negotiator {
 	}
 }
 
+// KnownPlainHTTP reports whether a registry's scheme is already certain without
+// probing: mTLS implies HTTPS, and Zarf's own internal registry (when not fronted
+// by the mTLS proxy) never has a TLS listener of its own. ok is false when neither
+// applies, meaning the caller must negotiate via UsePlainHTTP instead.
+func KnownPlainHTTP(usingMTLS, isInternal bool) (plainHTTP bool, ok bool) {
+	switch {
+	case usingMTLS:
+		return false, true
+	case isInternal:
+		return true, true
+	default:
+		return false, false
+	}
+}
+
 // UsePlainHTTP returns true if host should be reached over plain HTTP rather than
 // HTTPS, deciding by probing the host directly (see decide). A cached decision is
 // reused until it expires; see Options.TTL. Failures are cached too, but only for
--- a/src/pkg/ocischeme/ocischeme.go
+++ b/src/pkg/ocischeme/ocischeme.go
@@ -105,21 +105,6 @@ func New(o Options) *Negotiator {
 	}
 }
 
-// KnownPlainHTTP reports whether a registry's scheme is already certain without
-// probing: mTLS implies HTTPS, and Zarf's own internal registry (when not fronted
-// by the mTLS proxy) never has a TLS listener of its own. ok is false when neither
-// applies, meaning the caller must negotiate via UsePlainHTTP instead.
-func KnownPlainHTTP(usingMTLS, isInternal bool) (plainHTTP bool, ok bool) {
-	switch {
-	case usingMTLS:
-		return false, true
-	case isInternal:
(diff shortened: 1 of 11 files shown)


src/pkg/ocischeme/ocischeme.go
```diff
@@ -105,6 +105,21 @@ func New(o Options) *Negotiator {
 	}
 }
 
+// KnownPlainHTTP reports whether a registry's scheme is already certain without
+// probing: mTLS implies HTTPS, and Zarf's own internal registry (when not fronted
+// by the mTLS proxy) never has a TLS listener of its own. ok is false when neither
+// applies, meaning the caller must negotiate via UsePlainHTTP instead.
+func KnownPlainHTTP(usingMTLS, isInternal bool) (plainHTTP bool, ok bool) {
```
AustinAbro321: The ocischeme package should be unaware of the Zarf registry, how about moving this function to registry struct? I think in that case we could use the internal functions `IsInternal` and `ShouldUseMTLS` (We check for the cluster cfg to not be nil, but this is mostly for safety so we never panic)

AustinAbro321 (changes requested):
Comment on package organization, otherwise looks good

src/pkg/ocischeme/ocischeme.go
brandtkeller: great point - I moved it to the struct but with the caveat that `usingMTLS` remain an argument as there may be more criteria from the caller than simply cluster policy.

AustinAbro321: I think there is some weirdness in the existing code we could clarify here. The intent of `shouldUseMTLS()` is that we have a Zarf managed registry and Zarf is managing the mTLS for the registry. We should be able to rely on this as definitive proof, maybe a better function name would be `HasZarfManagedMTLS()`, but that can be a different PR. 

On main in `push.go` we follow the mTLS path if `cfg.Cluster != nil`  and `ShouldUseMTLS()`. I think instead we could error when `cfg.Cluster == nil` in the `ShouldUseMTLS()` path. This should never happen, but in case state is messed up we prevent panics and fail early since we wouldn't be able to get the certs. And this way we'll simplify the `KnownPlainHTTP` signature.

brandtkeller: I'm good with that change. I was extrapolating a few possible behaviors where we might want the delineation between intent and pure cluster state - but I think there is expectation on what that must be and less knowns around edge cases. Simplifying now is fair.

AustinAbro321 (approved):

PR #5067, opened 2026-07-13
fix(helm): create registry client for http chart with OCI re-direct

### LINKED ISSUE #5064: Regression in v0.74.0+: HTTP Bitnami Helm repositories fail with "missing registry client"
Device and OS:
Ubuntu 22.04 (Docker)

App version:
v0.74.0
v0.80.0

Working version:
v0.73.1

Kubernetes distro:
N/A (package create only)

Other:
Helm CLI can successfully pull the same chart.

Steps to reproduce

1. Create the following package:

```yaml
kind: ZarfPackageConfig

metadata:
  name: test-spark
  version: 0.0.1

components:
  - name: spark
    charts:
      - name: spark
        releaseName: spark
        namespace: default
        version: 9.4.1
        url: https://charts.bitnami.com/bitnami


2. create package

zarf package create .


3.Observe the error.




---

# Expected result

```text
Package should be created successfully.

The same package works correctly on Zarf v0.73.1.


Actual result

processing Helm chart name=spark version=9.4.1 repo=https://charts.bitnami.com/bitnami

retrying chart download

unable to lookup ref oci://registry-1.docker.io/bitnamicharts/spark:9.4.1

missing registry client


Medium

This blocks packaging Bitnami Helm charts using the traditional Helm repository.

### PR DESCRIPTION
**Pull request by AustinAbro321**
## Description

This broke with Helm 3 to Helm 4. Starting in Helm 4 all OCI pulls require an explicit registry client, and we were not creating this client in the direct path

## Related Issue

Fixes #5064

## Checklist before merging

### DIFF
--- a/src/internal/packager/helm/repo.go
+++ b/src/internal/packager/helm/repo.go
@@ -60,6 +60,26 @@ func negotiateChartPlainHTTP(ctx context.Context, ociURL string, remoteOptions t
 	return plainHTTP, nil
 }
 
+// newChartRegistryClient builds a Helm registry client for an OCI chart reference,
+// negotiating the transport scheme for its host (see negotiateChartPlainHTTP). It
+// returns the client alongside the negotiated plainHTTP so callers can pass the same
+// scheme to the chart getter.
+func newChartRegistryClient(ctx context.Context, ociURL string, remoteOptions types.RemoteOptions) (*registry.Client, bool, error) {
+	plainHTTP, err := negotiateChartPlainHTTP(ctx, ociURL, remoteOptions)
+	if err != nil {
+		return nil, false, err
+	}
+	clientOpts := []registry.ClientOption{registry.ClientOptEnableCache(true)}
+	if plainHTTP {
+		clientOpts = append(clientOpts, registry.ClientOptPlainHTTP())
+	}
+	regClient, err := registry.NewClient(clientOpts...)
+	if err != nil {
+		return nil, false, fmt.Errorf("unable to create the new registry client: %w", err)
+	}
+	return regClient, plainHTTP, nil
+}
+
 // negotiateLoadedChartDependenciesPlainHTTP negotiates a chart's OCI-referenced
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/internal/packager/helm/repo.go [hunk 1]
```diff
@@ -239,18 +259,10 @@ func DownloadPublishedChart(ctx context.Context, chart v1alpha1.ZarfChart, chart
 
 	// Handle OCI registries
 	if registry.IsOCI(chart.URL) {
-		plainHTTP, err = negotiateChartPlainHTTP(ctx, chart.URL, remoteOptions)
+		regClient, plainHTTP, err = newChartRegistryClient(ctx, chart.URL, remoteOptions)
```
brandtkeller: Do we need this executed here? I don't see any use of the returned values until after the second conditional which also executes the function.

src/internal/packager/helm/repo.go
```diff
@@ -283,6 +295,16 @@ func DownloadPublishedChart(ctx context.Context, chart v1alpha1.ZarfChart, chart
 		if err != nil {
 			return fmt.Errorf("unable to pull the helm chart: %w", err)
 		}
+
+		// A classic Helm repo index may redirect a chart to an OCI reference
+		// Helm's downloader requires a registry client to resolve
+		// an OCI ref, so build one for the resolved URL.
+		if registry.IsOCI(chartURL) {
```
brandtkeller: Do we prefer a conditional here or in the function itself as an early return path?

brandtkeller (changes requested):
Minor observations otherwise looks operational.

src/internal/packager/helm/repo.go
AustinAbro321: I combined this logic to run after we resolve the chart url

src/internal/packager/helm/repo.go
AustinAbro321: It was necessary, but I combined this with the second conditional

brandtkeller (approved):
lgtm

PR #4901, opened 2026-05-14
chore(deps): limit dependencies in favor of built-ins or core libs

soltysh:
## Description

This PR introduces two changes:
1. Replace controller-runtime helpers with k8s native client-go ones. Allowing us to drop direct dependency, although not entirely eliminating it due to transitive dependency through flux.
2. Replace custom helpers (`helpers.RemoveMatches`) with golang buitl-in `slices` counterparts. 

## Related Issue

Relates to https://github.com/zarf-dev/zarf/issues/4895

## Checklist before merging

- [x] Test, docs, adr added or updated as needed
- [x] [Contributor Guide Steps](https://github.com/zarf-dev/zarf/blob/main/CONTRIBUTING.md#developer-workflow) followed

PR #5077, opened 2026-07-16
fix(wait): properly wait for crds introduced mid-deploy

### LINKED ISSUE #5058: Health Check Stuck Since 0.77
### Environment
Device and OS: Windows / WSL (Ubuntu 26.04)
App version: Starting from 0.77.0, verified with all versions up to 0.81.0
Kubernetes distro being used: tested with kind at k8s 1.34.3, 1.35.0, 1.36.1, zarf provided k3s at 1.34.3
Other: -

### Steps to reproduce
1. Setup a fresh cluster with kind or k3s
2. run `zarf init`
3. Deploy the following zarf package
```yaml
# zarf.yaml
kind: ZarfPackageConfig
metadata:
  name: sample
  version: 0.0.1
  description: Sample package
variables: []

components:
  - name: infrastructure-helm
    required: true
    charts:
      - name: csi-driver-smb
        namespace: test-apps
        version: v1.20.1
        url: https://raw.githubusercontent.com/kubernetes-csi/csi-driver-smb/master/charts
      - name: envoy-api-gateway
        releaseName: envoy-gateway
        version: v1.8.2
        namespace: envoy-gateway-system
        url: oci://docker.io/envoyproxy/gateway-helm
    images:
      - docker.io/envoyproxy/gateway:v1.8.2
      - registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.16.0
      - registry.k8s.io/sig-storage/csi-provisioner:v6.1.1
      - registry.k8s.io/sig-storage/csi-resizer:v2.1.0
      - registry.k8s.io/sig-storage/livenessprobe:v2.18.0
      - registry.k8s.io/sig-storage/smbplugin:v1.20.1

  - name: infrastructure-manifests
    required: true
    manifests:
      - name: infrastructure
        namespace: test-apps
        files:
          - envoy-proxy-fixed-name.yaml
```

```yaml
# envoy-proxy-fixed-name.yaml
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: EnvoyProxy
metadata:
  name: envoy-proxy-fixed-name
  namespace: envoy-gateway-system
spec:
  provider:
    type: Kubernetes
    kubernetes:
      envoyService:
        name: envoy-api-gateway
        externalTrafficPolicy: Cluster
```
4. The health check on `infrastructure-manifests` is stuck and gives the following error message when i cancel context after many minutes
```
ERR failed to deploy package: context cancelled while deploying component "infrastructure-manifests": envoy-proxy-fixed-name: EnvoyProxy not ready, status is Unknown, message: 
context canceled
```

### Expected result

- Health check just passes

### Actual Result

- Health check is stuck

What's interesting is that health checks pass properly when I do one of the following things:
- Comment out / remove the csi-driver-smb chart (even though the health check fails on the Envoy resources, which should be fully unrelated!)
- Change order of envoy-api-gateway and csi-driver-smb charts, like this
```yaml
    charts:
      - name: envoy-api-gateway
        releaseName: envoy-gateway
        version: v1.8.2
        namespace: envoy-gateway-system
        url: oci://docker.io/envoyproxy/gateway-helm
      - name: csi-driver-smb
        namespace: test-apps
        version: v1.20.1
        url: https://raw.githubusercontent.com/kubernetes-csi/csi-driver-smb/master/charts
```

Further context:
- it passes when I cancel and run a second install
- it passes when I remove the package (which does not remove Envoy CRDs) and install again
- it does not pass, when I remove the package and remove the Envoy CRDs. Same problem then

Image

### Severity/Priority

Depends, as we could workaround. Current workaround is to simply change the order of charts in the package, which is luckily not a breaking change to us. Though, feels weird.

AustinAbro321:
## Description

Looks like this was a regression introduced in #4901

## Related Issue

Fixes #5058

## Checklist before merging


src/pkg/cluster/cluster.go
```diff
@@ -149,11 +147,15 @@ func WatcherForConfig(cfg *rest.Config) (watcher.StatusWatcher, error) {
 	if err != nil {
 		return nil, err
 	}
-	discoveryClient, err := discovery.NewDiscoveryClientForConfig(cfg)
+	httpClient, err := rest.HTTPClientFor(cfg)
+	if err != nil {
+		return nil, err
+	}
+	// NewDynamicRESTMapper reloads discovery so CRDs registered mid-deploy resolve on the next lookup.
+	restMapper, err := apiutil.NewDynamicRESTMapper(cfg, httpClient)
```
soltysh: let's be consistent using the k8s client, for what you're trying to achieve you can just invoke `Invalidate()` on a `discoveryClient`, this will force the cached data to be refreshed from server. Also we might want to consider doing that conditionally, only when new CRDs are registered. The k8s client cache TTL is pretty long, since the CRDs don't change that frequently.

soltysh: This will also limit the deps, I know we indirectly rely on controller-runtime, but less is better in that case.

AustinAbro321: I decided not to conditionally invoke invalidate since even after invalidate is called `DeferredDiscoveryRESTMapper` does not invalidate its field `delegate meta.RESTMapper`, only when there is a miss, will it retry and reload the cache

soltysh (approved):
lgtm

PR #5080, opened 2026-07-17
feat: stop requiring version for localpath charts

### LINKED ISSUE #3813: When specifying a helm chart via a localPath, the version field should not be required

AustinAbro321:
## Description

stop requiring version for charts when localpath is used

## Related Issue

Fixes #3813

## Checklist before merging

### DIFF
--- a/src/internal/api/v1alpha1/validate.go
+++ b/src/internal/api/v1alpha1/validate.go
@@ -292,7 +292,7 @@ func validateChart(chart v1alpha1.ZarfChart) error {
 		err = errors.Join(err, fmt.Errorf(PkgValidateErrChartURLOrPath, chart.Name))
 	}
 
-	// Local charts don't need to be found ihn an upstream repo and their Chart.yaml already has a version
+	// Local charts don't need to be found in an upstream repo and their Chart.yaml already has a version
 	if chart.LocalPath == "" && chart.Version == "" {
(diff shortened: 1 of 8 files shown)


src/internal/api/v1alpha1/validate.go
```diff
@@ -292,7 +292,8 @@ func validateChart(chart v1alpha1.ZarfChart) error {
 		err = errors.Join(err, fmt.Errorf(PkgValidateErrChartURLOrPath, chart.Name))
 	}
 
-	if chart.Version == "" {
+	// Local charts don't need to be found ihn an upstream repo and their Chart.yaml already has a version
```
soltysh:
```suggestion
	// Local charts don't need to be found in an upstream repo and their Chart.yaml already has a version
```

soltysh (dismissed):
lgtm

soltysh (approved):

PR #5081, opened 2026-07-17
fix: agent webhook logs

AustinAbro321:
## Description

The context that the agent logs were passed did not contain the logger on it since #4966 

This also removes the debug state print on LoadState, I've always found this log to be a bit much since state is loaded so often throughout Zarf, but especially for on the agent now that we load state it's obnoxious. Since we have a command `zarf tools get-creds` which practically shows state, and it's still shown when state is changed during `LoadState`, I think it's covered 

## Checklist before merging


soltysh (approved):
lgtm

Issue #4057, opened 2025-08-07
Update docs.zarf.dev log examples


mkcp:
### Describe what should be investigated or refactored
Back in 2024 we updated Zarf's logging format to structured logs. At the time of writing this, there's several screenshots and scrollable text boxes pointing to old logs examples that could be updated to new logs. Most of the screenshots are in Tutorials, but some exist in Reference e.g. https://docs.zarf.dev/ref/sboms/. Embedded Tools can be skipped.

### Links to any relevant code
- https://docs.zarf.dev/ref/sboms/
- https://docs.zarf.dev/tutorials/0-creating-a-zarf-package/
- https://docs.zarf.dev/tutorials/2-deploying-zarf-packages/
- https://docs.zarf.dev/tutorials/3-deploy-a-retro-arcade/

### Additional context
More is more here - it's ok to not get everything at once and instead ship an incremental PR for one a few articles.

PR #5089, opened 2026-07-21
docs: update command from 'prepare' to 'dev' in tutorial command html

codydacosta-uds:
## Description

The documentation instructs users to run the `dev` command rather than `prepare`, so I updated the command to match the tutorial. The `prepare` command produces the same result, but it is not the command referenced in the documentation.

...

## Related Issue

None

## Checklist before merging

- [ x] Test, docs, adr added or updated as needed

### DIFF
--- a/site/public/tutorials/prepare_find_images.html
+++ b/site/public/tutorials/prepare_find_images.html
@@ -47,7 +47,7 @@ b.BWHI {background-color: #aaaaaa}
 </head>
 <body>
 <pre>
-<b class="WHI">$ zarf prepare find-images</b><br/>
+<b class="WHI">$ zarf dev find-images</b><br/>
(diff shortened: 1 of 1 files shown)


brandtkeller (approved):
I agree with the inconsistency when the outer tutorial inserts the inner html. We have oustanding issues to revisit these tutorial pages if you (or anyone else) would like to help. Otherwise I think file naming and other changes would be negligible currently. 

We would probably want to fix the tutorials themselves before doing any other minute changes.

codydacosta-uds:
> @brandtkeller Yeah I agree that there needs to be a proper audit of these docs. I am going through many of the tutorials and I will just go ahead and keep a running list of issues I find or things I run into that are not tracked my troubleshooting sections or FAQs.

Any additional items you find as context for #4057 would be useful.

AustinAbro321 (approved):

PR #5090, opened 2026-07-21
docs: package values overview and migration guide

### LINKED ISSUE #4745: Documentation around variables, values and valuefiles are unclear

Please accept the below as my input based on using zarf for a few months and really valuing it. My hope is that this helps to improve its adoption for new users like myself.

It is not currently clear where and how values can be templated in components. I can see a worked example showing how a manifest file can contain template references to `{{ .Values.somevalue }}` however the documentation doesn't currently show if its possible to do the same in, for example, helm `valueFiles`. Overriding helm values only seems possible (according to the documentation) using the `values:` section of chart config, one by one, which may not be the most streamlined approach. I'd also be curious to know if this templating is possible in the zarf config file itself, or if there's some default values (similar to how help works) that can be specified, or an error thrown if certain values are specified.

It's also not clear how  `{{ .Values.somevalue }}` is intended to work with `###ZARF_VAR_<VALUE_KEY>###`. Are these interchangeable or only relevant in specific context? Some documentation uses one and some mentions the other. What other keys are available under `{{ . }}` beyond what's [documented here](https://docs.zarf.dev/ref/examples/values-templating/#alpha-values--templates-example) and what is their intended use?

I am also finding a lot of confusion with the documentation in general around variables, values, valuefiles, and helm charts values. These terms are all very similar but their inclusion in both the documentation and the schema itself seems very confusing. For example, I have a chart with `values:` defined, but I don't specify the values themselves, but a mapping from my values to the chart's values, and I do this one at a time? This also has a source, and target field, but again its not always clear which way round this might be. Can I use variables for these, and if so how? What about constants? are `valueFiles` evaluated first or last? Can they include templating or must that come from the `values:` section? Some clearer names, or better documentation would be very much appreciated to support this. Can the  instead be a key value mapping rather than a list?

All of this is to say that its very hard for a new user to understand the layout, schema or intent of these features. The documentation needs some reworking to make these concepts clearer.

### Additional context

I am aware that the values feature is currently in alpha, so I wanted to create an issue to track where I've been struggling to follow the documentation and understand the logic.

I'd be happy to help support changes to the documentation to make it easier to navigate.

AustinAbro321:
## Description

values documentation and migration guide

## Related Issue

Fixes #4745

## Checklist before merging


site/src/content/docs/best-practices/variables-to-values.mdx
```diff
@@ -0,0 +1,239 @@
...
+
+This guide explains how to migrate your package from variables to [package values](/ref/package-values). Migrating has several benefits:
+
+- Intuitive map structure - especially beneficial for large yaml block configuration.
+- Standard go-template engine with built-in and custom functions.
+- Built-in objects such as Package and State.
+
+## Migrating Manifests and Files
+
+Below shows an example migration for a regular text file. The below example shows manifests, however this format is applicable to files and Helm values files when templating is enabled.
```
soltysh:
```suggestion
Below shows an example migration for a regular text file. The  example shows manifests, however this format is applicable to files and Helm values files when templating is enabled.
```

site/src/content/docs/best-practices/variables-to-values.mdx
```diff
@@ -0,0 +1,239 @@
...
+  username: ###ZARF_VAR_DATABASE_USERNAME###
+```
+
+### After (Package Values)
+
+```yaml
+# zarf.yaml
+values:
+  files:
+    - package-values.yaml
```
soltysh:
```suggestion
    - package-values.yaml # should match the filenem with actual values, see below
```

site/src/content/docs/best-practices/variables-to-values.mdx
```diff
@@ -0,0 +1,239 @@
...
+
+components:
+  - name: app
+    manifests:
+      - name: configmap
+        template: true  # Required for Go templates
+        files:
+          - config.yaml
```
soltysh:
```suggestion
          - config.yaml # should match the filename with templates, see below
```

site/src/content/docs/ref/values.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: Deployment Values
+title: Deployment Variables
```
soltysh, line 2: Why this change?

site/src/content/docs/ref/package-values.mdx
```diff
@@ -0,0 +1,345 @@
...
+# Set values at deploy time
+zarf package deploy my-package.tar.zst \
+  --set-values="app.environment=staging,app.replicas=5"
+
+# Set values during removal
+zarf package remove my-package \
+  --set-values="app.environment=staging"
+```
+
+### Precedence
```
brandtkeller, line 241: ++ This sat in the back of my mind as something critical to the understanding - well captured

site/src/content/docs/ref/values.mdx
AustinAbro321: I think having both a deployment values and package values is confusing since package values are mainly used on deployment. I would also say it's more accurate since the Deployment page is mostly referring to variables.

brandtkeller (approved):
The URL for the `Deployment Variables` page is still `/ref/values` but that seems pretty minor and maybe not worth breaking anyone who has a reference to that page.

soltysh:
> but that seems pretty minor and maybe not worth breaking anyone who has a reference to that page.

maybe long-term try to introduce a redirect from values to variables

soltysh (approved):
lgtm

PR #5091, opened 2026-07-21
feat(healthchecks): use namespace-scoped watches for improved permissions handling

corang:
## Description

Updates health checks to use namespace-scoped informers when resources span multiple namespaces. This prevents namespaced resources from unnecessarily requiring cluster-wide `list` and `watch` permissions during the `running health checks` phase.

Cluster-scoped resources continue to use root-scoped requests as required.

Adds regression tests confirming that:

- Resources in multiple namespaces use separate namespace-scoped watches.
- Cluster-scoped resources remain root-scoped without widening access for namespaced resources.

## Related Issue

Relates to #

## Checklist before merging

### DIFF
--- a/src/internal/healthchecks/healthchecks.go
+++ b/src/internal/healthchecks/healthchecks.go
@@ -63,7 +63,13 @@ func WaitForReady(ctx context.Context, sw watcher.StatusWatcher, objs []object.O
 	cancelCtx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	eventCh := sw.Watch(cancelCtx, objs, watcher.Options{})
+	// Prefer namespace-scoped informers even when objects span multiple namespaces.
+	// The automatic cli-utils strategy switches to root-scoped informers when more
+	// than one namespace is present, which unnecessarily requires cluster-wide
+	// list/watch permissions for namespaced resources.
+	eventCh := sw.Watch(cancelCtx, objs, watcher.Options{
+		RESTScopeStrategy: watcher.RESTScopeNamespace,
+	})
 	statusCollector := collector.NewResourceStatusCollector(objs)
 	done := statusCollector.ListenWithObserver(eventCh, collector.ObserverFunc(
 		func(statusCollector *collector.ResourceStatusCollector, _ event.Event) {
--- a/src/internal/healthchecks/healthchecks_test.go
+++ b/src/internal/healthchecks/healthchecks_test.go
@@ -7,6 +7,9 @@ package healthchecks
 import (
 	"context"
 	"errors"
+	"fmt"
+	"maps"
+	"slices"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


src/internal/healthchecks/healthchecks.go
```diff
@@ -63,7 +63,13 @@ func WaitForReady(ctx context.Context, sw watcher.StatusWatcher, objs []object.O
 	cancelCtx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	eventCh := sw.Watch(cancelCtx, objs, watcher.Options{})
+	// Prefer namespace-scoped informers even when objects span multiple namespaces.
+	// The automatic cli-utils strategy switches to root-scoped informers when more
+	// than one namespace is present, which unnecessarily requires cluster-wide
+	// list/watch permissions for namespaced resources.
+	eventCh := sw.Watch(cancelCtx, objs, watcher.Options{
+		RESTScopeStrategy: watcher.RESTScopeNamespace,
```
soltysh, line 71: Did using `RESTScopeAutomatic` helps in your case, theoretically that should be the [default](https://github.com/kubernetes-sigs/cli-utils/blob/5895ad6c17dd06b99c1e5c8af17e2b12a3504fad/pkg/kstatus/watcher/watcher.go#L27), but looking here https://github.com/kubernetes-sigs/cli-utils/blob/5895ad6c17dd06b99c1e5c8af17e2b12a3504fad/pkg/kstatus/watcher/default_status_watcher.go#L81-L84 I can't find defaulting there anywhere. 
Whereas `RESTScopeAutomatic` is meant to trigger that check you've mentioned, if the objects are within single namespace, it should go with `RESTScopeNamespace`, otherwise it will pick `RESTScopeRoot`.

corang: `RESTScopeAutomatic` fails the tests

corang: Looks like the underlying code of automatic falls back to root scope immediately if more than 1 namespace is present in the resources passed

corang: https://github.com/kubernetes-sigs/cli-utils/blob/v0.37.2/pkg/kstatus/watcher/default_status_watcher.go#L130-L135

soltysh: Discussed this offline with Jordan. The goal here is to ensure we always use the namespace scope where possible, even with several namespaces (this is why the automatic scope fails, b/c it bumps into root scope the moment you're using more than a single namespace). So this lgtm

soltysh (approved):
lgtm

@AustinAbro321 promised to have a second look, so I'll leave it to him to merge once he's happy with it

AustinAbro321 (approved):
LGTM, looks like the reason that this is the default behavior in cli-utils is that it's more efficient to have a non namespaced group watch. For Zarf it is more important that we allow users to have the minimal RBAC set

PR #5092, opened 2026-07-22
feat: expose GetComponent function for package templating

### LINKED ISSUE #4878: Allow static Zarf values which cannot be changed at deploy time

AustinAbro321:
## Description

Exposes a Pkg.GetComponent function, this will be helpful when use the `.Pkg` object when templating files

## Related Issue

Fixes #4878

## Checklist before merging

### DIFF
--- a/examples/values-templating/nginx-configmap.yaml
+++ b/examples/values-templating/nginx-configmap.yaml
@@ -84,7 +84,7 @@ data:
         <h3>Package values (via .Pkg)</h3>
         <pre>Name: {{ .Pkg.Metadata.Name }}</pre>
         <pre>CLI version during build: {{ .Pkg.Build.Version }}</pre>
-        <pre>Nginx Image: {{ range .Pkg.Components }}{{ if eq .Name "values-with-manifest" }}{{ index .Images 0 }}{{ end }}{{ end }}</pre>
+        <pre>Nginx Image: {{ index (.Pkg.GetComponent "values-with-manifest").Images 0 }}</pre>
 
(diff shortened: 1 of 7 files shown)


soltysh (approved):
lgtm

PR #5094, opened 2026-07-22
fix: ensure `dev generate-schema` includes accurate values for imported components

### LINKED ISSUE #5086: `zarf dev generate-schema` includes values for "unimported" components
### Environment
Device and OS: MBP, macOS
App version: v0.81.0
Kubernetes distro being used: not relevant

### Steps to reproduce

1. Make a package with multiple components, leveraging [Zarf's values](https://docs.zarf.dev/ref/examples/values-templating/) and setup a default `values.files` with values for both components. Example:
    ```yaml
    values:
      files:
        - zarf-values.yaml
    
    components:
      - name: component-one
        charts:
          - name: component-one
            values:
              - sourcePath: .componentOne
                targetPath: .
    ... (actual chart definition is not relevant here)
    
      - name: component-two
        charts:
          - name: component-two
            values:
              - sourcePath: .componentTwo
                targetPath: .
    ...
    ```

    Where `zarf-values.yaml` includes both components:
    ```yaml
    componentOne: {}
    componentTwo: {}
    ```

2. In another zarf package import _one_ of the components only, for example:

    ```yaml
    components:
      - name: component-one
        import:
          path: subdir
    ```

3. Generate a schema using `zarf dev generate-schema`.

### Expected result

Schema only includes `componentOne` values. Zarf should read the `sourcePath` of my imported components to determine what values are relevant/necessary in my generated schema.

### Actual Result

Schema includes `componentTwo` values, based on the default values file (in this case just an open properties).

### Severity/Priority

Low - small annoyance but part of the dev commands.

### Additional Context

This appears specific to including default `values.files` in the `zarf.yaml` - the actual chart values/schemas do not appear to get pulled in, only schemas based on the "defaults".

mjnagel:
## Description

More context is in the issue, but currently the `dev generate-schema` command will make a schema that captures ALL "default" values for a package used in an import, even if only one component was used from that package. This change ensures that the default values are filtered based on `sourcePath` mappings so that only the components that are imported are represented in the resulting schema.

## Related Issue

Fixes https://github.com/zarf-dev/zarf/issues/5086

## Checklist before merging

### DIFF
--- a/src/cmd/dev.go
+++ b/src/cmd/dev.go
@@ -84,7 +84,7 @@ func parseSchemaValues(ctx context.Context, basePath string, defined load.Define
 	return zarfValues, nil
 }
 
-func rootPackageValuesFiles(ctx context.Context, basePath string) ([]string, error) {
+func rootPackageValuesFiles(ctx context.Context, basePath string, setPkgTmpl map[string]string) ([]string, error) {
 	pkgPath, err := layout.ResolvePackagePath(basePath)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


AustinAbro321:
@micah Appreciate you creating an issue and working towards a fix! Unfortunately this is trading over inclusion for under inclusion. A value that lives in an imported package's values file, belongs to an imported component, but is consumed only through a manifest, ZarfFile, or TemplatedValuesFile will get dropped. In the [v1beta1 schema](https://github.com/zarf-dev/proposals/tree/main/0051-v1beta1-schema), this will be fixed as components will only be importable from component configs rather than packages. 

We could fix this by parsing the different files, but I'd rather not introduce the complexity. Closing this

mjnagel:
> A value that lives in an imported package's values file, belongs to an imported component, but is consumed only through a manifest, ZarfFile, or TemplatedValuesFile will get dropped.

Just to make sure I'm tracking, this is something that is not explicitly mapped using source/target paths, but is referenced using `{{ .Values.xyz }}` templating? That makes sense and definitely would add more complexity here...

AustinAbro321:
Yup

PR #5097, opened 2026-07-23
feat: support public zoci api for package metadata

### LINKED ISSUE #5096: Support fetching a package digest and metadata from the same OCI manifest
We (the UDS Foundation team) are adding resume support for UDS bundle deployments. Before scheduling a remote OCI package, in order to make decisions about which packages to deploy and which have been successfully deployed, we need:

- The platform-selected OCI manifest digest, used as the package artifact identity.
- The corresponding `v1alpha1.ZarfPackage`, used to compare the intended component set with deployed state.
- A guarantee that both came from the same OCI manifest.
- This metadata to be available without pulling the full package.

The current public APIs expose these pieces separately. `Remote.ResolveRoot()` returns the root descriptor and digest, while `Remote.FetchZarfYAML()` correctly parses the package definition using an internal function, `pkgcfg.ParseMultiDoc()`.

However, `FetchZarfYAML()` calls `FetchRoot()` internally. For a mutable tag, the tag could move between the two resolutions:

```text
ResolveRoot(tag)        -> sha256:aaa
tag is updated          -> sha256:bbb
FetchZarfYAML(tag)      -> zarf.yaml from sha256:bbb
```

The caller could then associate the digest of one package with the definition and component set of another, making a resume skip decision unreliable.

- **Given** a remote Zarf package referenced by a mutable OCI tag
- **When** a consumer obtains its platform-selected manifest digest and parsed package definition
- **Then** both values correspond to the same resolved OCI manifest

- **Given** a built package definition containing multiple YAML documents or fields requiring schema migration
- **When** the package definition is parsed
- **Then** it uses the same document-selection and migration behavior as Zarf

- **Given** a consumer only needs the package digest and definition
- **When** that metadata is requested
- **Then** unrelated package layers do not need to be downloaded

A caller can pin the initial descriptor and manually fetch `zarf.yaml` from its manifest:

```go
desc, err := remote.ResolveRoot(ctx)
manifest, err := remote.FetchManifest(ctx, desc)
zarfYAMLDesc := manifest.Locate(layout.ZarfYAML)
data, err := remote.FetchLayer(ctx, zarfYAMLDesc)
```

The remaining issue is parsing `data`. Directly unmarshalling it into `v1alpha1.ZarfPackage` does not handle built package multi-document selection and schema migrations. The canonical parser, `pkgcfg.ParseMultiDoc()`, is internal and unavailable to external consumers.

Another option is fetching additional files, staging a partial package layout, and using `layout.LoadFromDir()`. This adds filesystem and layout-processing overhead when only package metadata is needed.

### Additional context
N/A

brandtkeller:
## Description

Exploring a discussion and issue around a few topics. First being a public api for the package metadata. Second being underlying exported and non-exported improvements to zoci in support of ensuring that mutable tags don't introduce drift between logical handling. 

This also fixes some areas of the code where multiple remotes were used where there should have only been one - as well as not being susceptible to the drift.

## Related Issue

Fixes #5096

Relates to #

## Checklist before merging


brandtkeller:
Moving to draft - this may be a good candidate to shift up to the `defenseunicorns/pkg/oci` domain...

brandtkeller:
Closing as completed in #5107

PR #5103, opened 2026-07-23
feat(values)!: type inference in set-values cli parsing

### LINKED ISSUE #5104: --set-values should infer types

Values set from `--set-values` are always parsed into a string. This means that if you try to set a boolean on the command line, for instance, `zarf package inspect manifests --set-value="enabled=true"`, then it will be read as a string and if your templates or schema expect a boolean then you will error. 

This will require a breaking change, as users might be usin data in the shape of an int or bool, that they want as a string. In these cases, the user can achieve the old behavior with single quotes --set-value="enabled='true'"

Reading from a values file does not have this problem. Below is read as a bool
```yaml
enabled: true
````

This was reported in the [Zarf openssf slack](https://openssf.slack.com/archives/C07AKUMBDMJ/p1784791610607369?thread_ts=1784709014.379849&cid=C07AKUMBDMJ) channel 

Helm has the flags `--set`, `--set-literal`, `--set-json`, and `--set-string`. I don't think we want to go that far, I'd rather encourage users to define things in values files. Those can perhaps be later additions. 

I think we should take one of the following approaches

### Helm approach 

The Helm approach is defined here - https://helm.sh/docs/intro/using_helm/#the-format-and-limitations-of---set. It accepts lists and positional indexes.

StringToString doesn't allow a list to have both multiple commas and multiple `=`. helm instead use a string array, which lets them interpret lists. We can switch to StringArray with minimal breakages in the CLI usage, but viper will be an entirely different format. 

```yaml
package:
  deploy:
    set_values:
       app.replicas: "3" # before
    set_values:
      - app.replicas=3 # after
```

### Pros
- Format is the same between tools, can lean on Helm docs
- Lists and positional arguments are accepted

### Cons
- Uses braces {} for lists which is a footgun for a bash tool since they need to be escaped
- No way to hard code something as a string, perhaps could fork the logic for this, but then we're different than upstream.

## Custom approach

We follow the Helm logic for scalars, but don't allow lists on the command line. Logic is: `true`/`false` become booleans, and whole base-10 numbers become integers. Everything else such as decimals (`1.5`) and leading-zero numbers (`0755`) stays a string.

This would keep us at stringToString, keeping the setValues in the viper config the same. 

### Pros 
- Could allow users to force a string with `--set-values=myStringToken='1234'`
- Simpler rules
- Keeps existing viper config unbroken (low priority imo)

### cons
- Supports less fields
- Different from Helm

AustinAbro321:
## Description

This went with the Helm style in #5104, though I'm considering switching to the custom option. Leaving in draft until we make a decision

Fixes #5104

## Checklist before merging


src/config/lang/english.go
```diff
@@ -312,7 +312,7 @@ $ zarf package mirror-resources zarf-package-my-app-amd64-1.0.0.tar.zst --repos
 	CmdPackageDeployFlagConnected              = "Deploy without pushing images/repos; label resources to bypass the Zarf agent"
 	CmdPackageDeployFlagForceConflicts         = "Force Helm to take ownership of conflicting fields during Server-Side Apply operations. Use when external tools (kubectl, HPAs, etc.) have modified resources."
 	CmdPackageDeployFlagSetVariables           = "Specify deployment variables to set on the command line (KEY=value)"
-	CmdPackageDeployFlagSetValues              = "Specify deployment package values to set on the command line (key.path=value)."
+	CmdPackageDeployFlagSetValues              = "Specify deployment package values to set on the command line (key.path=value). Values are type-inferred: true/false become booleans, and whole numbers become integers; everything else stays a string"
```
brandtkeller: Do we need to include mention of null?

brandtkeller (dismissed):
Minor nit otherwise I appreciate the simplicity and explicit behavioral change.

src/config/lang/english.go
AustinAbro321: I've actually been looking at when you'd want to use null and how Helm handles it. Looks like Helm deletes the key if a null is submitted rather than setting the field to null  https://github.com/helm/helm/issues/13459. 

I like our behavior better of setting the field to null, but since we're intentionally turning `--set-values` into very simple type conversion, I think it makes sense to make null exclusive to values file's anyway. Slim use case and now our short doc comment is entirely accurate.

brandtkeller (approved):
lgtm

PR #5105, opened 2026-07-23
ci: remove import test

AustinAbro321:
## Description

Test Import tests are failing very consistently with 500 errors, we could look into a more proper fix, but these are from a time where v1alpha1 was a completely separate package from Zarf, so it's not as necessary. If we go back to that, we could reintroduce an import test

## Checklist before merging


brandtkeller (approved):
lgtm

PR #5106, opened 2026-07-23
feat(update-creds): separate commands for each service

### LINKED ISSUE #5052: Require service be specified when running `zarf tools update-creds`
`zarf tools update-creds` currently updates the registry pull credentials, git pull credentials, Zarf registry chart, Git Server chart, Agent CA certs, and Zarf state. If anyone one of these fails it can leave the cluster in an odd shape. For instance, the registry pull secrets might disagree with state and the Zarf registry deployment. 

I believe it would be better to require that each service be explicitly specified. This leaves less chances for things to go wrong and limits the blast radius when there are problems. 

I could see some friction is someone has a requirement to rotate all credentials in the cluster every X days, but  it's not so hard to script out and I think the added benefits make it worth it

AustinAbro321:
## Description

Breaks update creds into separate commands, excludes artifact server since it's deprecated. 

These separate commands are going to have some behavior changes: 
- No longer warn if an operation is unsuccessful, always fail. 
- Rollback on failure
- Order of operations is different. First we change the internal service, then we change the secrets, then we change Zarf state.

Also updating gitea now expects a 200 error and will fail otherwise.

## Related Issue

Fixes #5052

## Checklist before merging


src/cmd/zarf_tools.go
```diff
@@ -482,6 +487,309 @@ func printCredentialUpdates(ctx context.Context, oldState *state.State, newState
...
+
+	confirm, err := confirmCredentialUpdate(ctx, oldState, newState, state.RegistryKey, o.confirm)
+	if err != nil {
+		return err
+	}
+	if !confirm {
+		return nil
+	}
+
+	// mTLS certs are not part of state, and so not part of the rollback
```
brandtkeller, line 589: Agree on sequence but I am curious if we should just leave these or track and revert them as well? I wouldn't consider this a blocker.

brandtkeller (dismissed):
lgtm. Tested functionality across the commands and see the expected propagation.

src/cmd/zarf_tools.go
AustinAbro321: I made the deciding factor here for if state could be out of sync, and so applied rollback by that metric. I could taking on another strategy, but kept it as is for now.

AustinAbro321:
@brandtkeller Decided to adjust an error message, so I need a re-review

brandtkeller (approved):

PR #5107, opened 2026-07-24
refactor!: chart naming and expose other useful library functions

### LINKED ISSUE #4047: feat: consider exposing more helper functions in the Zarf library

It may be nice to expose functions like https://github.com/zarf-dev/zarf/blob/b1a1189acac53b1223e900db88d121013ccac7b7/src/pkg/cluster/tunnel.go#L276 to better track  against Zarf's implementations of certain behavior behind the scenes (especially where the implementation is somewhat generic not depending on other Zarf code)

Helper functions could be identified and exposed (not sure this function is the only one) where they are not tightly coupled to the rest of the Zarf code - this function could possibly be moved from the cluster package into a utils package or similar to help maintain separation of concerns.

The code is open source and ofc could be copied where it needs to be but that would mean that that implementation wouldn't readily track the upstream implementation if it changed (as this code is likely to do after the IPV6 implementation).  You could also use the `unsafe` package and https://pkg.go.dev/cmd/compile#hdr-Linkname_Directive to force the import of that function but that has side effects that are not desireable.

### Additional context

This specific function would be nice to have to find the reference to the Zarf registry when sitting inside the cluster since 127.0.0.1:31999 needs to be converted to `zarf-docker-registry.zarf.svc.cluster.local:5000` when inside a Pod.

### LINKED ISSUE #4518: bug in public function GetServiceInfoFromRegistryAddress if localhost:<port> is passed.
Noticed an issue with this function:
serviceInfoFromNodePortURL
[GetServiceInfoFromRegistryAddress](https://github.com/zarf-dev/zarf/blob/main/src/pkg/cluster/tunnel.go#L277)

Called from public function: GetServiceInfoFromRegistryAddress at https://github.com/zarf-dev/zarf/blob/main/src/pkg/cluster/secrets.go#L185

The logic here:
```
	parsedURL, err := url.Parse(nodePortURL)
	if err != nil {
		parsedURL, err = url.Parse("scheme://" + nodePortURL)
		if err != nil {
			return corev1.Service{}, 0, err
		}
	}

	// Match hostname against localhost ip/hostnames
	hostname := parsedURL.Hostname()
	if hostname != helpers.IPV4Localhost && hostname != "localhost" {
		return corev1.Service{}, 0, fmt.Errorf("node port services should be on localhost")
	}
```

If a registry `localhost:31999` for example, is passed to this function it will determine 'localhost' is the scheme and won't trigger the prepending of 'scheme' to nodePortURL and won't properly assign hostname and triggers the error return with message "node port services should be on localhost"

Conversely, if `127.0.0.1:31999` is passed in, it is prepended with 'scheme' per the parsing error and continues as expected. 

I would suggest using net.SpligHostPort:

```
	u, err := url.Parse(nodePortURL)
	var rawHost string
	if err == nil && u.Host != "" {
		rawHost = u.Host
	} else {
		// Fallback for raw "localhost:8080" or "127.0.0.1:8080"
		rawHost = nodePortURL
	}

	host, portStr, err := net.SplitHostPort(rawHost)
```

Racer159:
## Description

This exposes some additional functions to help library users normalize logic with Zarf.  These changes include:

1. Extract chart/values pathing into the `layout` package so that others can use it - this also disambiguates the layout/naming responsibilities between `helm` and `layout` making `layout` the package that defines it for Zarf packages.
2. Expose the `ResolvePlainHTTP` logic so that callers can match how Zarf determines this if they want to interact with the Zarf registry directly.
3. Expose `CRCTag` on its own so that callers can resolve the canonical tag name Zarf uses when pushing images to the registry (without needing to strip it from the full registry transform).
4. Expose `ServiceInfoFromNodePortURL` so that callers can follow the same logic Zarf uses to interpret what K8s service is tied to a nodeport URL.
5. Refactor AssembleLayers and associated functions to use content.Fetcher from ORAS to allow other implementers of that interface (e.g. Zarf's layout) to be used there.

> [!CAUTION]
> Breaking library changes:
> 1. **`cluster.Registry{Server,Client}TLSSecret` and `cluster.RegistrySecret{CA,Cert,Key}Path`
>   moved to `state`**. Values are byte-identical; only the package changed.
> 2. **`layout.PackageChart` signature changed**:
>   `(ctx, chart, packagePath, chartPath, valuesFilePath, cachePath string, remoteOpts)` ->
>   `(ctx, chart, packagePath string, paths ChartPaths, cachePath string, remoteOpts)`.
>   Two path strings collapsed into one `ChartPaths` - requires some changes but should also
>   break compilers to highlight the change.
> 3. **`layout.AssemblePackage`** (and associated functions) moved to **`assemble.AssemblePackage`**

## Related Issue

Fixes #4047
Fixes #4518
Relates to #5096
Relates to #5097

## Checklist before merging


src/pkg/packager/layout/layout.go
```diff
@@ -72,3 +74,39 @@ func KustomizationFileName(manifestName string, idx int) string {
 func ComponentFileRelPath(idx int, target string) string {
 	return filepath.Join(strconv.Itoa(idx), filepath.Base(target))
 }
+
+// ChartArchiveName returns the base file name for a chart's packaged tarball:
+// "<name>" when the version is empty, otherwise "<name>-<version>".
+func ChartArchiveName(chart v1alpha1.ZarfChart) string {
```
AustinAbro321: Rather than accept a v1alpha1.ZarfChart, lets just take the chart name and version. This way we are not tied to an API version

src/internal/packager/helm/common.go
```diff
@@ -30,6 +30,17 @@ import (
 
 var contentCachePath = filepath.Join("helm", "content")
 
+// ChartPaths resolves the on-disk locations of a chart's packaged artifacts.
+// Callers (which own the package layout convention) supply an implementation, so
+// the helm package operates on resolved paths rather than deriving the layout
+// convention itself.
+type ChartPaths interface {
```
AustinAbro321: These interface feels mostly targeted at avoiding cyclic dependencies. What would you think of instead moving assemble into a separate package from layout? I believe this will remove the reliance on Helm from layout, and allow us to use the layout functions directly. 

Fine with this all being in one PR, but feel free to revert the Helm changes in this PR and make a separate PR if that starts to get unwieldy

AustinAbro321 (changes requested):
Overall design mostly looks good, as well as a few QOL fixes. Requesting changes on the Helm interface

AustinAbro321:
@Racer159 Could you rename this PR to begin with "refactor" instead of Chore. Also if there are any breaking changes to public exported function signatures, please call them out in the description and add a ! to refactor.

src/internal/packager/helm/common.go
Racer159: done

src/pkg/packager/assemble/assemble_external_test.go
AustinAbro321, line 1: We can move this test to the internal package now (it only wasn't before because of cyclic dependency)

src/pkg/packager/layout/package_test.go
```diff
@@ -1862,3 +1862,33 @@ func TestHasValuesSchema(t *testing.T) {
 		require.True(t, p.HasValuesSchema(), "file on disk should take precedence over empty metadata field")
 	})
 }
+
+// writeDocumentationTar tars the flat contents of dir, which is all the
+// documentation reader needs. Package assembly writes this tarball with its own
+// reproducible writer; reading it back does not depend on those guarantees.
+func writeDocumentationTar(t *testing.T, dir, tarPath string) {
```
AustinAbro321, line 1870: Lets use the helper archive.Compress in this function to simplify

AustinAbro321 (commented):
Some minor asks, design looks good

src/pkg/packager/layout/package_test.go
Racer159: :saluting_face:

src/pkg/packager/assemble/assemble_external_test.go
Racer159: done

AustinAbro321 (approved):

brandtkeller (approved):
lgtm - I'd go as far as believing that this also fixes #5096 but I am open to having a gap in my mental model.

Issue #3285, opened 2024-12-03
Create nightly test for Zarf init with s3 backend


AustinAbro321:
### Describe what should be investigated or refactored
The init package is currently able to be configured with Zarf variables to support an S3 backend. See [github.com/defenseunicorns/narwhal-delivery-iac-swf-reference-deployment](https://github.com/defenseunicorns/narwhal-delivery-iac-swf-reference-deployment/blob/main/iac/swf/uds-config.tf#L44-L73) for an example of a configuration with an s3 backend. There are no tests in this repo to ensure that this S3 backend works, so certain changes to the registry portion of the init package cannot be made with confidence.

This is the case for PR #3241 which solves a legitimate bug within the init package, but we need a test to know if this change breaks the s3 backed init package variant

phillebaba:
A good option to implement a test for S3 integration would be to use a S3 compatible local store. This https://github.com/seaweedfs/seaweedfs is one option of what we could use. We should do some research if there is a S3 in memory implementation in Go to make the integration with the e2e tests easier.

PR #5119, opened 2026-07-28
feat(init): more secure pod defaults

### LINKED ISSUE #2757: Default pod security settings

The pods that we deploy as part of the init package use most default settings for security configuration. This is pretty normal for most OSS projects out there. It is usually the responsibility of end user to figure out which permissions a project actually needs and lock down the permissions.

As Zarf is a security focused project which means that it is likely that our end users will enforce more restrictive privileges than what is default in Kubernetes. To make this easier for end users we should add more restrictions by default in our own packages that we provide, like the init package. This is not only to simplify deployment but to increase the security of what could be considered Zarf managed.

- **Given** the init command is run
- **When** a pod is deployed
- **Then** the pod should have reasonable pod security settings

N/A

### Additional context

This was raised by an end user who had constraints in the cluster requiring these settings.

AustinAbro321:
## Description

Changes pod defaults to be secure, I need to do testing to see where this could break users

The gitea user and group and the current defaults, so nothing should be changed there. We are running gitea in rootless mode so running as non root shouldn't be an issue.

Privileged is already false, just making it explicit. The only pod using service tokens is the agent. 

## Related Issue

Fixes #2757

## Checklist before merging

### DIFF
--- a/packages/zarf-registry/chart/templates/deployment.yaml
+++ b/packages/zarf-registry/chart/templates/deployment.yaml
@@ -35,7 +35,6 @@ spec:
 {{- end }}
     spec:
       serviceAccountName: {{ include "docker-registry.serviceAccountName" . }}
-      automountServiceAccountToken: false
       {{- if .Values.imagePullSecrets }}
       imagePullSecrets:
 {{ toYaml .Values.imagePullSecrets | indent 8 }}
--- a/packages/zarf-registry/chart/templates/deployment.yaml
+++ b/packages/zarf-registry/chart/templates/deployment.yaml
@@ -35,6 +35,7 @@ spec:
 {{- end }}
     spec:
       serviceAccountName: {{ include "docker-registry.serviceAccountName" . }}
+      automountServiceAccountToken: false
       {{- if .Values.imagePullSecrets }}
       imagePullSecrets:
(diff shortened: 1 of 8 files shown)


packages/zarf-registry/chart/templates/deployment.yaml
```diff
@@ -35,6 +35,7 @@ spec:
 {{- end }}
     spec:
       serviceAccountName: {{ include "docker-registry.serviceAccountName" . }}
+      automountServiceAccountToken: false
```
a1994sc: If possible, can this be a toggle? I believe that when using something like a AWS S3 bucket as the backend with the EKS pod identity, you need to use the token from cluster on that auth.

AustinAbro321: This is a most excellent callout and a reminder that we really need #3285 considering how many people use the s3 backend. I removed for now as the risks outweigh the benefits

a1994sc: Gotcha, I think that is a good middle ground. Thank you sir

brandtkeller (approved):
lgtm - I don't see anything that could produce unintended results and the variables should provide an escape hatch if needed.

PR #5127, opened 2026-07-29
ci: mitigate 403 errors when building docs site

AustinAbro321:
## Description

mitigate 403 errors when building docs site by getting the tags through the Git CLI rather than GitHub API. 

## Checklist before merging

### DIFF
--- a/site/hack/build-versions.mjs
+++ b/site/hack/build-versions.mjs
@@ -21,6 +21,7 @@ const KEEP_CURRENT_MAJOR = 10;
 const KEEP_PREVIOUS_MAJOR = 3;
 
 const VERSION_SLUG = /^v\d+-\d+$/;
+const SEMVER_TAG = /^v?\d+\.\d+\.\d+$/;
 
 function git(args, opts = {}) {
   return execFileSync("git", args, { cwd: repoDir, stdio: "inherit", ...opts });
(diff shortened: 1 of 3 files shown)


site/hack/build-versions.mjs
```diff
@@ -94,6 +84,29 @@ async function discoverVersions() {
   return { latest: minorsDesc[0], archived };
 }
 
+function discoverGitTags() {
+  const output = git(["ls-remote", "--tags", "origin", "v*.*.*"], {
```
brandtkeller: What happens if this is a fork? is origin hardcoded here to the fork? 

should we instead re-establish/modify the constant project upstream as the definitive source? (trying to think about what that would mean).

brandtkeller (commented):
One minor thought - trying to look for any CI security gaps - but curious on the output of Forks.

site/hack/build-versions.mjs
AustinAbro321: Great thought, I hadn't considered that. There is an argument for keeping it as origin since origin would be checked out by GitHub in actions against the fork, but assuming the fork is trying to contribute back to Zarf, then it makes the most sense to build versions against the tags in the Zarf repo. Given this, I decided to hardcode our repo url.

brandtkeller (approved):
lgtm - number of rabbit holes possible but negligible at best.

PR #5128, opened 2026-07-29
feat(find-images): exclude images from Helm tests

### LINKED ISSUE #4079: zarf dev find-images will include helm chart test resources
### Environment
Device and OS: OSX
App version: v0.59.0

### Steps to reproduce
1. Create a zarf package that references helm chart:

```
      - name: grafana
        version: 9.3.1
        namespace: grafana
        url: https://grafana.github.io/helm-charts
```

2. Run the `zarf dev find-images` command and notice it will pick up helm chart test resources and include the following image:

```
<internal_artifactory_redacted>/bats/bats:v1.4.1
```
When using the `zarf dev find-images why` command it outputs:

```
component: grafana
chart: grafana
resource:

apiVersion: v1
kind: Pod
metadata:
  annotations:
    helm.sh/hook: test
    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
  labels:
    app.kubernetes.io/instance: grafana
    app.kubernetes.io/name: grafana
    app.kubernetes.io/version: 12.1.0
    helm.sh/chart: grafana-9.3.0
  name: grafana-test
  namespace: grafana
spec:
  containers:
  - command:
    - /opt/bats/bin/bats
    - -t
    - /tests/run.sh
    image: <internal_artifactory_redacted>/bats/bats:v1.4.1
    imagePullPolicy: IfNotPresent
    name: grafana-test
    volumeMounts:
    - mountPath: /tests
      name: tests
      readOnly: true
  restartPolicy: Never
  serviceAccountName: grafana-test
  volumes:
  - configMap:
      name: grafana-test
    name: tests
```

### Expected behavior

The tool should not scan resources under the /test folder in a helm chart, or it should look to see if the resource is annotated with `helm.sh/hook: test` and ignore it.

theZMC:
## Description

This PR refactors `zarf dev find-images` and `zarf dev generate` to exclude test resource images.

**This is a breaking behavioral change for the `zarf` CLI.**

## Related Issue

Fixes #4079 

## Checklist before merging

### DIFF
--- a/site/src/content/docs/commands/zarf_dev_find-images.md
+++ b/site/src/content/docs/commands/zarf_dev_find-images.md
@@ -14,7 +14,10 @@ Evaluates components in a Zarf file to identify images specified in their helm c
 
 Evaluates components in a Zarf file to identify images specified in their helm charts and manifests.
 
-Components that have repos that host helm charts can be processed by providing the --repo-chart-path.
+Images defined as part of Helm test hooks are excluded. Images explicitly defined in the helm.sh/images
+annotation in a component Helm chart's Chart.yaml are included regardless of their presence in the rendered
+chart. Components that have git repositories that host helm charts can be processed by providing the
+--repo-chart-path.
 
 ```
 zarf dev find-images [ DIRECTORY ] [flags]
--- a/src/config/lang/english.go
+++ b/src/config/lang/english.go
@@ -523,7 +523,10 @@ $ zarf package pull oci://ghcr.io/zarf-dev/packages/dos-games:1.3.0 -a skeleton`
 
 	CmdDevFindImagesShort = "Evaluates components in a Zarf file to identify images specified in their helm charts and manifests."
 	CmdDevFindImagesLong  = "Evaluates components in a Zarf file to identify images specified in their helm charts and manifests.\n\n" +
-		"Components that have repos that host helm charts can be processed by providing the --repo-chart-path."
+		"Images defined as part of Helm test hooks are excluded. Images explicitly defined in the helm.sh/images\n" +
+		"annotation in a component Helm chart's Chart.yaml are included regardless of their presence in the rendered\n" +
+		"chart. Components that have git repositories that host helm charts can be processed by providing the\n" +
+		"--repo-chart-path."
 
 	CmdDevGenerateConfigShort = "Generates a config file for Zarf"
 	CmdDevGenerateConfigLong  = "Generates a Zarf config file for controlling how the Zarf CLI operates. Optionally accepts a filename to write the config to.\n\n" +
--- a/src/pkg/packager/find_images.go
+++ b/src/pkg/packager/find_images.go
@@ -296,6 +296,7 @@ func findImages(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath strin
 			if err != nil {
 				return nil, err
 			}
+			yamls = slices.DeleteFunc(yamls, isHelmTestResource)
 			resources = append(resources, yamls...)
 			chartPath := filepath.Join(compBuildPath, string(layout.ChartsComponentDir))
 			chartTarball := filepath.Join(chartPath, layout.ChartArchiveName(zarfChart.Name, zarfChart.Version))
@@ -339,6 +340,7 @@ func findImages(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath strin
 				if err != nil {
 					return nil, err
 				}
+				yamls = slices.DeleteFunc(yamls, isHelmTestResource)
 				resources = append(resources, yamls...)
 
 				// Check if the --why flag is set and if it is process the manifests
@@ -466,6 +468,10 @@ func findImages(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath strin
 	}
 
 	return componentImageScans, nil
+}
+
+func isHelmTestResource(resource *unstructured.Unstructured) bool {
+	return resource.GetAnnotations()["helm.sh/hook"] == "test"
 }
 
 // processUnstructuredImages processes a Kubernetes resource and extracts container images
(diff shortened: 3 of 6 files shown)


site/src/content/docs/commands/zarf_dev_find-images.md
```diff
@@ -27,6 +27,7 @@ zarf dev find-images [ DIRECTORY ] [flags]
       --deploy-set-variables stringToString   Specify deployment variables to set on the command line (KEY=value) (default [])
   -f, --flavor string                         The flavor of components to include in the resulting package (i.e. have a matching or empty "only.flavor" key)
   -h, --help                                  help for find-images
+      --include-test-resources                Include resources annotated with helm.sh/hook: test when discovering images
```
AustinAbro321: I think the use case for including test resources in a Zarf package is very slim. Technically someone could run a helm test in an action, or unarchive a chart and test it, but these use cases seem unlikley. It'd be simpler to run the Helm test before packaging the chart in Zarf. 

Let's remove the flag to avoid adding complexity.

AustinAbro321 (changes requested):
Thanks for taking this on, great start!

site/src/content/docs/commands/zarf_dev_find-images.md
theZMC: Done, and I updated the PR description to match the reduced scope.

AustinAbro321 (dismissed):
LGTM

src/pkg/packager/testdata/find-images/helm-chart/chart/templates/post-install-hook.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "{{ include "chart.fullname" . }}-post-install"
+  annotations:
+    "helm.sh/hook": post-install
```
brandtkeller: Not a required change - If we made this a comma-delimited list (as supported by hooks) then we could ensure that any additional changes to find-images continues to preserve these resources that should still surface images.

This for instance should still detect the alpine image in testing:
```suggestion
    "helm.sh/hook": post-install,test
```

src/pkg/packager/find_images.go
```diff
@@ -468,6 +470,10 @@ func findImages(ctx context.Context, pkg v1alpha1.ZarfPackage, packagePath strin
 	return componentImageScans, nil
 }
 
+func isHelmTestResource(resource *unstructured.Unstructured) bool {
```
brandtkeller, line 473: nit: can we add documentation for this behavior to the command? 

IE:
> Excludes images found in test-only helm hooks

Looking at this more, we don't document the `helm.sh/images` `Chart.yaml` list either - which is authoritative meaning that a test image defined in that list will still be included in the output. I won't block on that addition but if you wanted to leave the camp cleaner than you found it I'd appreciate that.

brandtkeller (changes requested):
Minor changes otherwise clean and effective.

src/pkg/packager/find_images.go
theZMC: I updated the dev.mdx with both those additions. Is there anywhere else this should be documented?

src/pkg/packager/testdata/find-images/helm-chart/chart/templates/post-install-hook.yaml
theZMC: Done.

src/pkg/packager/find_images.go
brandtkeller: if you can add it to `lang.CmdDevFindImagesLong` as well that would be perfect!

theZMC: Done.

```
❯ go run main.go dev find-images --help
Evaluates components in a Zarf file to identify images specified in their helm charts and manifests.

Images defined as part of Helm test hooks are excluded. Images explicitly defined in the helm.sh/images annotation in a component Helm chart's Chart.yaml are included regardless of their presence in the rendered chart. Components that have git repositories that host helm charts can be processed by providing the --repo-chart-path.

...
```

src/config/lang/english.go
```diff
@@ -523,7 +523,9 @@ $ zarf package pull oci://ghcr.io/zarf-dev/packages/dos-games:1.3.0 -a skeleton`
 
 	CmdDevFindImagesShort = "Evaluates components in a Zarf file to identify images specified in their helm charts and manifests."
 	CmdDevFindImagesLong  = "Evaluates components in a Zarf file to identify images specified in their helm charts and manifests.\n\n" +
-		"Components that have repos that host helm charts can be processed by providing the --repo-chart-path."
+		"Images defined as part of Helm test hooks are excluded. " +
```
brandtkeller: Quick go-back - can you format this with newlines? (See `CmdDevGenerateConfigLong` below)

theZMC: Sure thing. I had done that originally, but it looked odd with some lines much shorter than others. I'll re-arrange the lines to cascade in a way that's more aesthetically pleasing.

theZMC: Decided to just manually wrap at ~100 characters. Happy to tweak further.

```
❯ go run main.go dev find-images --help
Evaluates components in a Zarf file to identify images specified in their helm charts and manifests.

Images defined as part of Helm test hooks are excluded. Images explicitly defined in the helm.sh/images
annotation in a component Helm chart's Chart.yaml are included regardless of their presence in the rendered
chart. Components that have git repositories that host helm charts can be processed by providing the
--repo-chart-path.
```

brandtkeller: Looks good - I expect the docs linting to fail as the command docs have not been updated - run `make docs-and-schema` to regenerate the command docs.

theZMC: Done. My bad. I've got to internalize that step.

brandtkeller (approved):
lgtm - thanks for taking this on.

AustinAbro321 (approved):

PR #5141, opened 2026-07-31
docs: update actions documentation to focus on values

AustinAbro321:
## Description

Updates action documentation to focus on values instead of variables

Also updates the site to version examples that way older version of the docs site referencing now deleted or updated components still work. Uses the same strategy as the schema strategy as so I combined them into a single folder

## Checklist before merging


site/src/plugins/remark-link-rewrite.test.ts
```diff
@@ -45,14 +45,30 @@ test("adds one ../ to asset paths escaping the version subtree", () => {
...
-test("adds one ../ to escaping ESM/raw imports, preserving query", () => {
+test("adds one ../ to generic escaping imports, preserving query", () => {
   const fileDir = path.join(versionRoot, "ref");
   assert.equal(
     fixEscapingRelative("../../../../../examples/config-file/zarf.yaml?raw", fileDir, versionRoot),
     "../../../../../../examples/config-file/zarf.yaml?raw",
   );
 });
 
+test("maps versioned asset repo imports to staged assets", () => {
```
brandtkeller, line 56: We don't have any documented steps for running the site unit tests such as these. May be worth an npm script and inclusion in ci (unless I missed it)

brandtkeller (dismissed):
The versioned asset staging is neat. Minor non-blocking comment on site testing but I ran the tests manually to confirm. lgtm.

site/src/plugins/remark-link-rewrite.test.ts
AustinAbro321: Great call out, added an npm test step

brandtkeller (approved):

PR #5146, opened 2026-07-31
fix: skeleton packages included only one architecture per component

### LINKED ISSUE #4446: Skeleton Projects only support 1 architecture for inheritance

When trying to inherit injector component from the main `zarf-init` package, you are not able to inherit the `arm64` file. I believe this is because we only allow one instance of a component `.name` per `zarf.yaml` to be published, thus the `skeleton` publish code will take the first instance of `.name: zarf-injector`, e.i. the `amd64` arch, and ignore the `arm64`.

TBD

### Additional context

```yaml
---
# yaml-language-server: $schema=https://raw.githubusercontent.com/zarf-dev/zarf/main/zarf.schema.json
kind: ZarfInitConfig
metadata:
  name: zarf-init-test
components:
  - name: zarf-injector
    default: true
    import:
      url: oci://ghcr.io/zarf-dev/packages/init:v0.67.0
```

Image

yohanb:
## Description

When publishing a skeleton package, `PackageDefinition` resolved imports using the host machine's architecture as the
filter. This meant that multi-arch source packages (where the same component name appears once per architecture, e.g.
two zarf-injector entries; amd64 & arm64) had all but the host arch variant silently dropped during import
resolution.

Consequently, a skeleton published from an amd64 machine would only contain the amd64 variant of zarf-injector. An
arm64 user importing from that skeleton would fail to find a compatible component.

### Variant dimension abstraction (`load.go`)

Introduced `VariantDimension` (`VariantArchitecture`, `VariantFlavor`) and `SkipVariantFilters []VariantDimension`
on `DefinitionOptions`. Callers pass the dimensions they want to retain; `load.PackageDefinition` passes them
straight through to `resolveImports` without any skeleton-specific logic or sentinel values.

`publish.go` passes `SkipVariantFilters: []load.VariantDimension{load.VariantArchitecture}` when publishing a
skeleton, keeping the load package free of skeleton-specific knowledge.

### Multi-arch import resolution (`import.go`)

- `resolveImports` and `compatibleComponent` accept `skipVariantFilters []VariantDimension` instead of relying on
  the `SkeletonArch` sentinel. When `VariantArchitecture` is skipped, every arch variant of a matching component
  is collected instead of filtered to one.
- When multiple components with the same name are found and `VariantArchitecture` is not skipped, the existing
  "multiple components named X found" error is returned (non-skeleton behaviour unchanged).

### Arch-qualified tarball naming (`import.go`)

`fetchOCISkeleton` now looks for `<name>-<arch>.tar` before falling back to `<name>.tar`, so each arch variant
of a component resolves to its own layer in the OCI skeleton.

## Related Issue

Fixes #4446

## Testing

Unit tests (`src/pkg/packager/load/import_test.go`):
- `TestResolveImportsVariantDimensions`: verifies that skipping `VariantArchitecture` includes both amd64 and arm64
  variants; amd64/arm64 filters each include only their respective variant.
- `TestCompatibleComponent`: new cases covering skip-arch-only, skip-flavor-only, and skip-all-filters, ensuring
  each dimension works independently and in combination.

Local end-to-end test against the zarf source itself (which has both k3s and zarf-injector as two-variant multi-arch
components):
- `./dist/zarf-fixed package publish . oci://localhost:5001 --plain-http --skip-version-check`
- `./dist/zarf-fixed package inspect definition oci://localhost:5001/init:unset-development-only --plain-http -a skeleton`

Published skeleton contains both amd64 and arm64 variants of k3s and zarf-injector. `metadata.architecture: skeleton`
is set correctly.

## Backwards compatibility

`fetchOCISkeleton` falls back from `<name>-<arch>.tar` to `<name>.tar` when the arch-specific layer is absent.
Skeletons published before this fix continue to work; they just can't provide the previously-missing arch variants.

## Checklist before merging


yohanb:
1 month later and still nothing.  I understand you guys are busy but the longer this branch lingers, the more maintenance it requires.

AustinAbro321:
@yohanb Appreciate your interest here. We decided not to put more work into Skeleton packages, as they're being replaced in the v1beta1 schema with remote components, which will not have this architecture problem. See https://github.com/zarf-dev/proposals/blob/main/0051-v1beta1-schema/README.md#remote-components . 

We'll be putting out more information about switching as the feature is released, but we wanted to avoid too much in progress work especially since Skeleton packages have had this bug for years. Should have wrote you a message explaining, my mistake!

yohanb:
All good, thanks for the update @AustinAbro321

PR #5155, opened 2026-08-03
refactor: fix flakey tests

AustinAbro321:
## Description

We have a test that verifies the that the context deadline is exceeded on timeout which recently flaked here https://github.com/zarf-dev/zarf/actions/runs/30796341874/job/91630671412. While we could increase the timeout, or add a fake clock and cancel syncronously,  we have other tests which verify the retry behavior that we're really interested in. Better to delete imo

Sometimes in the wait package are flaking on timeouts. Changed it so we never wait for a probe that's not going to finish, and all the probes we do wait on have a test server behind them. Also increased timeout to connect to that test server to 1 second. Restructured the code to use our retry package as well

## Checklist before merging


brandtkeller (approved):
lgtm

brandtkeller:
Removed from the queue so as to not skew potential release - feel free to re-add and include otherwise.

PR #5156, opened 2026-08-03
fix: state access in OnRemove actions

AustinAbro321:
## Description

OnRemove actions are missing StateAccess

## Checklist before merging

### DIFF
--- a/src/pkg/packager/remove.go
+++ b/src/pkg/packager/remove.go
@@ -105,6 +105,14 @@ func Remove(ctx context.Context, pkg v1alpha1.ZarfPackage, opts RemoveOptions) e
 		return fmt.Errorf("failed to get working directory: %w", err)
 	}
 
+	var s *state.State
+	if opts.Cluster != nil {
+		s, err = opts.Cluster.LoadState(ctx)
+		if err != nil {
+			l.Debug("unable to load Zarf state for remove actions", "error", err.Error())
+		}
+	}
+
 	reverseDepComps := slices.Clone(depPkg.DeployedComponents)
 	slices.Reverse(reverseDepComps)
 	for _, depComp := range reverseDepComps {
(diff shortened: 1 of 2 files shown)


src/pkg/packager/remove.go
```diff
@@ -105,6 +105,14 @@ func Remove(ctx context.Context, pkg v1alpha1.ZarfPackage, opts RemoveOptions) e
 		return fmt.Errorf("failed to get working directory: %w", err)
 	}
 
+	var s *state.State
```
brandtkeller, line 108: I'm facing a bit of a philosophical debate in my head as I retrace remove operations. I started by asking "should this hook have access to state?" and I think the answer is yes... but for everything? 

Zarf doesn't perform any validation (digest/signature/etc) that remove is using a "trusted" source. I think this is an aside but maybe provides a potential opportunity to utilize digest further.

In the case of a deployed package,  I believe you can essentially run arbitrary onRemove actions given you have a locally supplied package that has a matching name and matching component names. We're passing in the components from the supplied package and not the deployed package for the actions below. The caveat here is that Zarf _does_ print the output of the supplied package to the terminal - but do we want to allow this potential drift? 

Back to access - should we allow state access to components that don't require a cluster? I'm not inherently against it but I am trying to think through any security implications that could be considered now. Specifically I think one scenario is allowing state access to components running onRemove actions from the deployed package and otherwise disallowing state access. 

Thoughts? Maybe this is a non-issue - just feels like a bit too much wiggle room for me to feel comfortable. I'd also accept out-of-scope as your changes look entirely functional.

brandtkeller (commented):
Comment on a thought that came to mind reading this and tracing back to the command.

src/pkg/packager/remove.go
AustinAbro321: Great thoughts, and we'll have to balance the UX. I do think there is a functional case for these in onRemove actions. Given a field called `StateAccess` it's logical for a user to assume they could use it for an onRemove action that deletes certain git repositories or images in a registry. 

> allowing state access to components running onRemove actions from the deployed package and otherwise disallowing state access.

What this point gets to is that we trust packages already in cluster more than new packages. Makes sense, but I wouldn't draw the line at stateAccess as I believe the the UX will become too confusing. I would rather draw the line that remove MUST be run from a deployed package. Of course this does introduce risk, you wouldn't be able to edit the onRemove actions of a deployed package without redeploying, so I'm not 100% sold on this idea, but it would be easier for users to follow and potentially less risky.

Actions are arguably our most risky security feature, it's always good to question what we allow. To me the security improvements should be more holistic

brandtkeller: I think the UX on the `StateAccess` is one I would agree needs a story around the threat model and largely doesn't need any change at this time - so I'm going to resolve this and approve. 

Separately is the asymmetry for the removal target - let's think about this more.

brandtkeller (approved):
lgtm

PR #5160, opened 2026-08-03
feat: dev sha256sum not working with `extract-path` flag

### LINKED ISSUE #5158: dev sha256sum does not work when provided extract-path flag
### Environment
Device and OS: n/a
App version: v0.81.0
Other: n/a

### Steps to reproduce
1. `curl --output-dir /tmp -O -L https://github.com/rancher/rke2/releases/download/v1.36.2%2Brke2r1/rke2.linux-amd64.tar.gz` 
2. `zarf dev sha256sum -e "bin/rke2" /tmp/rke2.linux-amd64.tar.gz`

### Expected result

sha256sum

### Actual Result

```text
2026-08-03 18:15:33 ERR unable to compute the SHA256SUM hash
unable to decompress: file "/run/user/1000/tmp/zarf-3815479178/bin/rke2" not found in archive
```

a1994sc:
## Description

* `DecompressOpts.Files` now uses `extract-path` as-is (the path relative to inside the tarball) instead of the joined temp-dir path, which never matched an entry in the archive
* Separate the download temp dir from the extraction temp dir, so a `--extract-path` whose name matches the downloaded archive's basename no longer overwrites the archive mid-extraction
* Extract the `sha256sum` command's logic into a `compute` method so it's directly testable; `run` just prints the result
* Add regression tests: local-fixture coverage for `extract-path` (no network calls), plus a test that reproduces the download/extract temp-dir collision using an archive built at test time and served over a local `httptest` server

## Related Issue

Fixes #5158

## Checklist before merging


src/cmd/dev.go
```diff
@@ -825,7 +825,8 @@ func (o *devSha256SumOptions) run(cmd *cobra.Command, args []string) (err error)
 		extractedFile := filepath.Join(tmp, o.extractPath)
 
 		decompressOpts := archive.DecompressOpts{
-			Files: []string{extractedFile},
+			// We want to extract the file relative to the tar ball
+			Files: []string{o.extractPath},
```
brandtkeller, line 907: 🤔 this has been broken for a _while_. Could you add a regression test for the expected behavior? 

There is a present but slightly edgecase behavior now though where the target basename and conflicting extract path target are the same then it would try to overwrite the archive it's reading... I'll let you decide if that feels worthy of fixing (Perhaps a separation of what is downloaded versus extracted in the tempdir?)

brandtkeller (changes requested):
Good catch! WIlling to toss an approval on this ASAP but would love to not see this regress in the future.

src/cmd/dev.go
a1994sc: Hm, downloading to a separate temp directory would probably be safer. I can up the code and throw in a little test or two.

a1994sc: Hey, so I moved the logic out to extract the file from the tar ball into a separate temp directory that gets cleaned out.

Also added some tests, but I could not figure out how to grab the `fmt.Println` output, so I moved it to a new function... Please let me know if that is ok. Thanks.

src/cmd/dev_test.go
```diff
@@ -274,3 +274,68 @@ func TestDevInspectValuesFiles(t *testing.T) {
 		})
 	}
 }
+
+func TestDevSha256Sum(t *testing.T) {
```
brandtkeller, line 283: nit: can we isolate network dependencies from the extraction issue? I'd like to limit network requirements for testing (to include flakiness due to relying on availability of GitHub). Could we instead create a local fixture for these checks?

brandtkeller (changes requested):
Minor nit on testing. This does open a question with regards to the efficiency of the decompress logic for a specific file extraction (IE do we stop once the specified file(s) is/are found or continue through the archive?) - but that can be its own issue.

src/cmd/dev_test.go
a1994sc: I think this is valid, my personal reservation stems from the xz backdoor that was caused by having a tar ball in the testing I believe : 

[article](https://en.wikipedia.org/wiki/XZ_Utils_backdoor)

However I will take a look at the existing tar balls and create a test based off those

src/cmd/dev.go
a1994sc: Are you ok with me moving the logic of this into a separate function? Again it was largely done for creating the tests

src/cmd/dev_test.go
brandtkeller: I agree - I was originally thinking that you could probably generate a test fixture with expected content versus storing one in the repo (something we _should_ not do if possible).

src/cmd/dev.go
brandtkeller: I think testing is a great reason for the separation.

src/cmd/dev_test.go
brandtkeller: To clarify - I was thinking we _replace_ the github-retrieved artifacts in favor of locally-generated (at test time) test fixtures. IF we need to prove-out the network case then maybe a small http test server (I think there is prior art to draw from).

What do you think?

src/cmd/dev.go
```diff
@@ -782,20 +782,32 @@ func newDevSha256SumCommand() *cobra.Command {
...
+	}
+	fmt.Println(out)
+	return nil
+}
+
+func (o *devSha256SumOptions) compute(ctx context.Context, args []string) (output string, err error) {
 	hashErr := errors.New("unable to compute the SHA256SUM hash")
-	ctx := cmd.Context()
+
+	if len(args) == 0 || args[0] == "" {
```
brandtkeller: Minor nit - I might encourage moving arg validation to `run` and having this method accept a string target that separates the concerns. Then native cobra and run can ensure we have the arguments required and `compute` can validate target inputs separately.

brandtkeller (changes requested):
Minor additional thoughts - otherwise great catch.

src/cmd/dev.go
a1994sc: Should it be a different error? Or should I just not check to see if the array has more then one item?

brandtkeller: we should be able to use `Args:    cobra.ExactArgs(1)` on the command itself. Then the logical boundary becomes input validation where `compute` should likely receive a string and validate against empty or any other validation available?

brandtkeller: Are we able to use `ExactArgs()` in the command itself?

brandtkeller: Disregard my last - thinking through the input validation and the existence of ExactArgs already I think these naturally oppose most input validation I'd wanted to see.

src/cmd/dev.go
```diff
@@ -842,20 +842,30 @@ func newDevSha256SumCommand() *cobra.Command {
 }
 
 func (o *devSha256SumOptions) run(cmd *cobra.Command, args []string) (err error) {
+	out, err := o.compute(cmd.Context(), args[0])
+	if err != nil {
+		return err
+	}
+	fmt.Println(out)
+	return nil
+}
```
brandtkeller, line 851: I only see one potential difference in behavior - now that this is split into two functions - the deferred cleanup error joins could prevent an otherwise available hash from being printed. 

I think we could check for Hash here and write it out if not empty and then otherwise handle errors as expected (retaining the previous behavior).


```suggestion
	out, err := o.compute(cmd.Context(), args[0])
	if out != "" {
		fmt.Println(out)
	}
	return err
}
```

brandtkeller (changes requested):
Minor change in edgecase behavior otherwise looks good

Issue #5166, opened 2026-08-04
Remove specific API versions from function signatures


AustinAbro321:
# Description

There are many public functions and structs which accept the v1alpha1 version. Many should change to either accept a small subset of the data they need or to accept a `PackageAccessor`. For instance, `PackageLayout` will remove the public v1alpha1 Pkg and instead hold a `PackageDefinition` with a refined list of setters. 

Functions in Packager will accept either a path to a zarf.yaml, a `PackageLayout`, or a `PackageAccessor`. This decouples them from any API version.

Packages like `actions` or `helm` which work on a subset of data in the package will declare their own type, for instance `Helm.ChartInfo`, which carry the information needed for a specific action run. 

There will be public functions to turn a `PackageDefinition` and component name into a list of `Helm.ChartInfo` objects. 

Relates to #3433

Follows implementation from https://github.com/zarf-dev/proposals/tree/main/0048-schema-update-process and  https://github.com/zarf-dev/proposals/tree/main/0051-v1beta1-schema and

PR #5173, opened 2026-08-04
feat(v1beta1)!: use package definition across public packager functionality

AustinAbro321:
## Description

This PR changes packager to use the `PackageDefinition` struct across public packager functionality. Either through the package layout, or through the `PackageAccessor` interface. V1beta1 is not yet implemented.

## SDK breaking changes

  - `layout.PackageLayout.Pkg` was replaced by `api.PackageDefinition`. Read definitions with layout.AsV1alpha1() / layout.AsV1beta1(); for mutations, use `api.PackageDefinition` methods such as `SetName` or `RemoveImages`.

  - `packager.GetPackageFromSourceOrCluster` and `packager.Remove` now use `api.PackageDefinition` instead of `v1alpha1.ZarfPackage`.

## Related Issue

Relates to #5166

## Checklist before merging


src/api/api.go
```diff
@@ -59,3 +84,193 @@ func (p *PackageDefinition) SetName(name string) {
 func (p *PackageDefinition) SetAnnotations(annotations map[string]string) {
 	p.pkg.Metadata.Annotations = annotations
 }
+
+// RemoveImages removes images and image archives from every component.
+func (p *PackageDefinition) RemoveImages() {
```
soltysh, line 89: The purist in my will insist to separate the types from the methods 😉 but that's not blocking.

src/pkg/packager/load.go
```diff
@@ -206,37 +206,38 @@ func identifySource(src string) (string, error) {
 	return "", fmt.Errorf("unknown source %s. Did you forget the scheme (e.g. (oci://) or file extension (e.g. .tar.zst)?", src)
 }
 
-// GetPackageFromSourceOrCluster retrieves a Zarf package from a source or cluster.
-func GetPackageFromSourceOrCluster(ctx context.Context, cluster *cluster.Cluster, src string, namespaceOverride string, opts LoadOptions) (_ v1alpha1.ZarfPackage, err error) {
+// GetPackageFromSourceOrCluster retrieves a package definition from a source or cluster.
+func GetPackageFromSourceOrCluster(ctx context.Context, cluster *cluster.Cluster, src string, namespaceOverride string, opts LoadOptions) (_ api.PackageDefinition, err error) {
```
soltysh:
```suggestion
func GetPackageFromSourceOrCluster(ctx context.Context, cluster *cluster.Cluster, src string, namespaceOverride string, opts LoadOptions) (api.PackageDefinition, error) {
```
is equally valid 😉

src/api/api.go
soltysh: I also wonder if accepting `PackageDefinition` as argument here, instead of making type methods isn't better approach? Since you'll implement them here, you'll still have access to private members. Just thinking out loud.

src/api/api.go
```diff
@@ -59,3 +84,193 @@ func (p *PackageDefinition) SetName(name string) {
...
+			return fmt.Errorf("component index %d out of range", idx)
+		}
+		components = append(components, p.pkg.Components[idx])
+	}
+	p.pkg.Components = components
+	return nil
+}
+
+// OverrideNamespace overrides component namespaces when the definition permits it.
+func (p *PackageDefinition) OverrideNamespace(namespace string) error {
```
soltysh, line 117: With the introduction of this method I think you should drop https://github.com/zarf-dev/zarf/blob/d7f041aaec22396b3005dec15e5d88e212d26094/src/pkg/packager/packager.go#L120 and additional methods which are deadcode now.

src/api/api.go
AustinAbro321: Hmm, I'm open to it. I assume you're picturing a signature like below?
```go
func RemoveImages(p PackageDefinition) PackageDefinition 
```

Would you recommend this even for the smaller setters like `func (p *PackageDefinition) SetName(name string)` or just for the big package wide change like `RemoveImages` or `OverrideNamespace`?

src/pkg/packager/load.go
AustinAbro321: The error is named so the cleanup error gets properly returned
```go
defer func() {
  err = errors.Join(err, pkgLayout.Cleanup())
}()
```

src/api/api.go
AustinAbro321: Nice catch

src/api/api.go
soltysh: I'm not 100% sure, but I think if we go with that we'd go all or nothing, definitely not partial.

src/pkg/packager/load.go
soltysh: I missed that one.

soltysh (approved):
lgtm

Issue #5039, opened 2026-07-06
Longhorn example improvements


brandtkeller:
### Describe what should be investigated or refactored

As a follow on from #4875 and the resulting PR - there are a few notable [improvements](https://github.com/zarf-dev/zarf/pull/4923#pullrequestreview-4595414804) that could be made to that example package:

1. Introduce an optional pre-flight using the longhorn-cli image and documented [daemonset](https://github.com/longhorn/cli/blob/master/deploy/check.yaml) strategy for check.
2. Improve the health checks to additionally include the fully reconciled deployments/daemonsets 

Separately documenting the required input for registry as seen [here](https://github.com/zarf-dev/zarf/blob/main/examples/longhorn/values.yaml#L5) . Why is this necessary? let's document this in the package README or remove if possible.

### Additional context
These can be split into distinct PR's if necessary.

acald-creator:
@brandtkeller 

I would like to work on this.

brandtkeller:
Assigned! As always feel free to tag maintainers for any questions. There may be very good reasons to NOT do any one of these things as you research and that's perfectly fine too.

abhishekgit03:
Hi @brandtkeller,  just checking if this is still being actively worked on - happy to help if not!

acald-creator:
@brandtkeller Good morning

I apologize that I did not get a chance to work on this, other priorities came up, but it is totally fine if the above PR is sufficient to complete this PR.

If it is, then I will continue to review other issues that I can work on instead.

brandtkeller:
Thanks for the confirmation @acald-creator , no apologies needed!

abhishekgit03:
Closed in #4875

PR #5175, opened 2026-08-04
docs(examples): improve longhorn health checks and registry docs

abhishekgit03:
Related to #5039

Follow-up on the longhorn example package improvements requested in #5039. This covers 2 of the 3 items; the preflight DaemonSet needed a different approach than originally scoped (see note below).

### Changes

- **Health checks**: add an `onDeploy.after` action that waits for every Longhorn `EngineImage` to reach `state: deployed`. Zarf's existing chart-level wait already covers the Helm release's own Deployments/DaemonSets, but Longhorn's per-node `EngineImage` DaemonSets are created asynchronously *after* the chart install returns, and could still fail, this closes that gap.
- **Registry docs**: document what `global.imageRegistry: "###ZARF_REGISTRY###"` in `values.yaml` actually does (redirects Longhorn's internally-templated image references to Zarf's registry) and why it must stay.

### Preflight DaemonSet - not implemented as scoped

I looked into adding the optional preflight DaemonSet from [longhorn/cli's `deploy/check.yaml`](https://github.com/longhorn/cli/blob/master/deploy/check.yaml), but found it's stale against the current `longhorn-cli:v1.11.2` release - that image only ships `longhornctl` (no standalone `longhorn-preflight` binary), and `longhornctl check preflight` requires a real kubeconfig with no in-cluster auto-detection (verified by running it as an in-cluster Job with a bound ServiceAccount - it still rejected with "no kubeconfig path provided"). Looking at its `--help` output and source, it's designed to be run by an operator with a real kubeconfig, where it self-deploys and tears down its own DaemonSet across the cluster via the API - not to be embedded as a static per-node manifest. Rather than build the more complex kubeconfig-bootstrapping version to replicate that, I documented `longhornctl check preflight --kubeconfig=<path>` as a manual pre-deploy step in the README instead. Open to revisiting if you'd prefer the automated route.

### Test plan

- [x] `zarf dev lint examples/longhorn` - clean
- [x] `zarf package create` - builds successfully
- [x] Deployed to a real kind cluster: confirmed the health-check action correctly fails when misconfigured (caught a real bug -`maxTotalSeconds` alone doesn't retry without `maxRetries`) and passes once the `EngineImage` reaches `deployed`
- [x] `zarf package remove` - clean teardown, `onRemove` action still works
- [x] `make lint-go`, `pre-commit run --all-files`


examples/longhorn/readme.md
```diff
@@ -11,3 +11,5 @@ If you're working with K3s, there is extra setup required. See [Longhorn CSI on
 The values file from this example was pulled using the directions at [Customizing Default Settings](https://longhorn.io/docs/latest/advanced-resources/deploy/customizing-default-settings/#using-helm) as the path for kubelet needs to be set for K3s as per [Longhorn CSI on K3s](https://longhorn.io/docs/latest/advanced-resources/os-distro-specific/csi-on-k3s/)
 
 You do not need to use the values file and can remove it from the Zarf package configuration if you're not using K3s and don't need that variable set.
+
+The `global.imageRegistry` value in `values.yaml` is set to Zarf's `###ZARF_REGISTRY###` template variable. Longhorn's chart templates prefix every image reference they render internally (CSI sidecars, engine, UI, webhooks, etc.) with `global.imageRegistry`, so this line redirects those images to Zarf's registry instead of the public `docker.io` registry. This is required for Longhorn's images to resolve in an airgapped environment and should not be removed.
```
brandtkeller: As an aside - do we know why this is needed given the Zarf mutating webhook?

examples/longhorn/zarf.yaml
```diff
@@ -12,6 +12,15 @@ components:
       onRemove:
         before:
           - cmd: './zarf tools kubectl -n longhorn-system patch -p ''{"value": "true"}'' --type=merge lhs deleting-confirmation-flag'
+      onDeploy:
+        after:
+          - description: Longhorn EngineImages to finish deploying on every node
+            maxTotalSeconds: 300
+            maxRetries: 60
+            cmd: |
```
brandtkeller: Can we use a native `wait` here instead?

brandtkeller (changes requested):
Good notes on the attempted prefligtht daemonset install - curious about `imageRegistry` as well as justification behind targeting the `EngineImages` (Is it the last unit to become healthy?).

examples/longhorn/readme.md
abhishekgit03: The webhook does generically rewrite any pod's image on admission, so it's not needed for pulling. But `longhorn-manager` is also passed its own expected image as a `--manager-image` CLI flag (rendered from `global.imageRegistry`, independent of the webhook), and it compares that flag against its own running pod's image to detect stale manager instances during upgrades. If `global.imageRegistry` is left unset, that flag never matches the webhook-rewritten pod, and the manager gets stuck permanently treating its own fresh pod as stale (reproduced this twice). So it's required for `longhorn-manager`'s own bootstrap logic, not just image resolution.

abhishekgit03:
> Good notes on the attempted prefligtht daemonset install - curious about `imageRegistry` as well as justification behind targeting the `EngineImages` (Is it the last unit to become healthy?).

you were right about the second point, i rechecked it, `EngineImages` is not the last unit. Traced the reconciliation order live: `InstanceManager` is created and reaches running after EngineImage reaches deployed. Added a second wait for InstanceManager to cover it.

examples/longhorn/zarf.yaml
abhishekgit03: Yes, done. Replaced the `cmd` with two `wait.cluster` actions using label selectors (`longhorn.io/component=engine-image` / `instance-manager`).

examples/longhorn/readme.md
brandtkeller: Final go back - testing now. Can we consolidate the current note here with that information you provided? That is superb context I would love to track.

abhishekgit03: @brandtkeller Done, consolidated it into the README note.

AustinAbro321 (approved):

AustinAbro321:
Thanks for the contribution!

abhishekgit03:
@brandtkeller Could you confirm if this is good to merge, or if there are more changes required?

brandtkeller (approved):
lgtm. Thanks for the updates and clearing up the ambiguity in how longhorn should be configured.

PR #5176, opened 2026-08-04
feat(dev): evaluate helm chart schemas in dev generate-schema

### LINKED ISSUE #5157: `dev generate-schema` should evaluate/include helm chart schemas

`zarf dev generate-schema` infers a Helm value whose default is YAML `null` as a JSON Schema `string`, even when the Helm chart's `values.schema.json` declares that value as an object or null.

This causes generated Zarf schemas to reject valid map values such as service-account annotations and labels.

- **Given** a Helm chart with a `values.schema.json` entry declaring a value as `type: ["object", "null"]`
- **When** that chart value is mapped into Zarf values and `zarf dev generate-schema` is run
- **Then** the generated Zarf schema preserves the chart-declared object/null type and any applicable map validation

(or more broadly - if a chart declares a values schema, it should be used as an additional input for the generated zarf schema)

- Manually authoring the corresponding entries in `values.schema.json` after running `zarf dev generate-schema`.
- Avoid null defaults in charts
- Allow `null` defaults to be "more permissive" in Zarf without checking the chart's schema (essentially allow any type when `null`)

### Additional context

Originally discovered this with the Velero helm chart. A basic reproduction:
```yaml
kind: ZarfPackageConfig
metadata:
  name: velero-generate-schema-reproducer
  description: Reproducer for nullable Helm object schema inference

components:
  - name: velero
    required: true
    charts:
      - name: velero
        url: https://vmware-tanzu.github.io/helm-charts
        version: 12.1.0
        releaseName: velero
        namespace: velero
        values:
          - sourcePath: ".velero.serviceAccount.server.annotations"
            targetPath: ".serviceAccount.server.annotations"
          - sourcePath: ".velero.serviceAccount.server.labels"
            targetPath: ".serviceAccount.server.labels"
```

Resulting schema from `zarf dev generate-schema`:
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "properties": {
    "velero": {
      "properties": {
        "serviceAccount": {
          "properties": {
            "server": {
              "properties": {
                "annotations": {
                  "type": "string"
                },
                "labels": {
                  "type": "string"
                }
              },
              "type": "object"
            }
          },
          "type": "object"
        }
      },
      "type": "object"
    }
  },
  "type": "object"
}
```

Contrast with upstream schema: https://github.com/vmware-tanzu/helm-charts/blob/velero-12.1.0/charts/velero/values.schema.json#L597-L606

mjnagel:
## Description

Evaluates helm chart schemas as part of the `zarf dev generate-schema` behavior. This particular implementation:
- Imports a subset of fields from the chart schema focused on the shape/type of values and any bounds on string/array/numeric values
- `description`, `examples`, and other similar "metadata" schema keys are not imported
- "Default sensitive" constraints like `required`/`minProperties` are not imported as they cannot be "fully evaluated" at the zarf layer
- More advanced schema composition and definitions (`allOf`, `anyOf`, `$ref`, etc) are skipped and fall back to inferred typing (these can all get more messy to properly handle esp in multi-chart scenarios or when other omitted fields are present in these)

The main focus is on high value / low risk schema constraints (type and bounds) while not bloating the zarf schema with excessive validations or introducing significant logic to handle some of the more complex scenarios.

## Related Issue

Fixes https://github.com/zarf-dev/zarf/issues/5157

## Checklist before merging


src/cmd/dev.go
```diff
@@ -178,6 +186,19 @@ func (o *devGenerateSchemaOptions) run(ctx context.Context, args []string) error
...
 
 			appliedValues := helpers.MergeMapRecursive(helmChart.Values, valuesFilesValues)
+			var chartSchema map[string]any
+			if len(helmChart.Schema) > 0 {
+				if err := json.Unmarshal(helmChart.Schema, &chartSchema); err != nil {
+					l.Warn("unable to parse Helm chart values schema; falling back to inferred types", "chart", chart.Name, "error", err)
+					chartSchema = nil
+				} else if err := value.CheckNoExternalRefs(chartSchema); err != nil {
+					l.Warn("unable to use Helm chart values schema; falling back to inferred types", "chart", chart.Name, "error", err)
+					chartSchema = nil
```
AustinAbro321: We do this check for Zarf charts, because a external reference breaks our air-gap nativeness, however for `zarf dev generate-schema` an external reference somewhere in the schema is fine, as we assume connectivity.

src/cmd/dev.go
```diff
@@ -91,13 +91,19 @@ type devGenerateSchemaOptions struct {
...
+}
+
 func newDevGenerateSchemaCommand(v *viper.Viper) *cobra.Command {
 	o := &devGenerateSchemaOptions{}
 
 	cmd := &cobra.Command{
 		Use:   "generate-schema [ DIRECTORY ]",
 		Args:  cobra.MaximumNArgs(1),
-		Short: "Generates a JSON schema for Zarf values based on the package definition and chart defaults",
+		Short: "Generates a JSON schema for Zarf values based on the package definition, chart defaults, and chart schemas",
```
AustinAbro321, line 106:
```suggestion
		Short: "Generates a JSON schema for Zarf values based on the package definition, chart defaults, and chart schema",
```
nit

src/test/packages/14-generate-schema/zarf.yaml
AustinAbro321, line 1: Could you add an entry to test out the exclude paths logic in this e2e test as well

src/pkg/value/generate.go
```diff
@@ -21,34 +23,277 @@ func GenerateJSONSchema(vals Values) map[string]any {
...
+	return ok && len(schema) == 0
+}
+
+func filterChartSchemaValue(key string, value any) any {
+	if key == "properties" {
+		if schemas, ok := value.(map[string]any); ok {
+			filtered := make(map[string]any, len(schemas))
+			for name, schema := range schemas {
+				if schemaMap, ok := schema.(map[string]any); ok {
+					filtered[name] = filterChartSchema(schemaMap)
```
AustinAbro321: This keeps child entries even when `filterChartSchema` returns an empty map. A field like `{ "description": "..." }` become `"field": {}` in the generated schema instead of being dropped with the rest of the unsupported chart metadata.

Add a unit test for this scenario

AustinAbro321 (changes requested):
Overall looks like a good addition, requesting a few changes

src/cmd/dev.go
mjnagel: Plural was semi-intentional here as packages are 1:N with charts. Would `chart schema(s)` be better? Tracking that a given chart only has a single schema, I think more about multiple charts in a given package, which ultimately means multiple schemas.

AustinAbro321: Oh sorry, I thought I deleted this comment. Please ignore!

src/cmd/dev.go
mjnagel: Hm yeah let me look at this a bit further. I originally had some code to handle refs but it was starting to feel messy (especially with multiple charts and potential for clashes between their named definitions). But can try to find a sane/simple approach to handle some of this.

mjnagel: Okay I think I tackled this in the latest changes - it should ensure we drop any "more complex" ref/def handling while keeping any unrelated portions of the schema that we do have more plainly.

AustinAbro321 (approved):
While there are JSON schema edge cases not handled, this is an improvement over baseline, and provides good value for its complexity. LGTM

brandtkeller (approved):
lgtm - the only edge case I found that was substantial is a Zarf value that is used as a source path for multiple charts and how this handles enums - Last wins I think? Quite niche and agree that this is a definitive improvement.

PR #5177, opened 2026-08-05
feat: direct verification of bundle with sigstore-go

### LINKED ISSUE #5172: Cosign verification writes `Verified OK` directly to stderr
### Environment
Device and OS: Observed from a downstream CLI on arm64
App version: Zarf v0.82.0
Other: Cosign v3.1.1

### Steps to reproduce
1. Use Zarf as a Go library and verify a signed package or blob.
2. Observe stderr after verification succeeds.

### Expected result

Verification output is silent or uses Zarf's logger.

### Actual Result

Cosign writes `Verified OK` directly to stderr, bypassing Zarf's logger and the downstream CLI's logging configuration.

```text
2026-08-04 10:10:05 INF ingesting package name=init source=oci://ghcr.io/zarf-dev/packages/init:v0.81.1
Verified OK
2026-08-04 10:10:09 INF bundle archive written output=uds-bundle-signature-verification-init-keyless-arm64-0.1.0.tar.zst
```

### Severity/Priority

Low. Verification succeeds, but downstream consumers cannot control or format this output.

### Additional Context

Zarf calls cosign's `VerifyBlobCmd.Exec` here:
https://github.com/zarf-dev/zarf/blob/v0.82.0/src/pkg/signing/cosign.go#L289-L328

Cosign prints the message here:
https://github.com/sigstore/cosign/blob/v3.1.1/cmd/cosign/cli/verify/verify_blob.go#L153-L164

Cosign's UI output defaults to `os.Stderr` here:
https://github.com/sigstore/cosign/blob/v3.1.1/internal/ui/env.go#L37-L43

brandtkeller:
## Description

With the removal of producing any legacy signatures and only defaulting to the sigstore bundle format we can now begin looking at migrating from the use of Cosign to Sigstore. 

We still require the existing Cosign logic for legacy signatures - which we've been preserving for the ability to perform verification of older packages without the bundle format - yet we can begin integrating new sigstore functionality that can eventually allow us to deprecate and remove the cosign logic. 

A full migration spike is still underway - but verification itself is more isolated and less complex than the requirements for signing. 

This PR add direct sigstore bundle verification with options for the supported verification paths rather than cosign parity. 

The shape of the integration allows for conversion from `VerifyBlobOptions` into `BundleVerificationOptions` and a `VerifyBundle` entrypoint focusing on in-memory bytes rather than paths. Should the artifacts we want to support become large we can implement a reader API for streaming.

Moving the blank KMS provider imports to a neutral location such that they are not inadvertently lost between any future migration work as this sigstore implementation does rely on them for parity.

Adds some additional testing and highlights some features that we don't tend to broadcast (such as public key retrieval from an environment variable or kubernetes. 

## Related Issue

Fixes #5172

Relates to #

## Checklist before merging


src/pkg/packager/layout/package.go
```diff
@@ -360,7 +360,7 @@ func (p *PackageLayout) VerifyPackageSignature(ctx context.Context, opts signing
 		opts.Key = opts.KeyRef //nolint:staticcheck // intentional read of deprecated alias for migration sync
 	}
 
-	hasKey := opts.Key != ""
+	hasKey := opts.Key != "" || opts.SecurityKey.Use
```
brandtkeller, line 363: Not exposed by the CLI but _potentially_ supported by the SDK given the broader `VerfiyBlobOptions`.

soltysh (dismissed):
lgtm

AustinAbro321:
It's a breaking change, but what your thoughts of limiting this to only support the cosign fields directly available for packages in the CLI? It would simplify the logic and lessen the chance of breaking changes post v1

PR #5179, opened 2026-08-05
feat: add option to skip values schema validation

mjnagel:
## Description

Adds a option (library and CLI flag) `SkipValuesSchemaValidation`/`--skip-values-schema-validation` that bypasses values schema validation for deploy, init, and relevant inspect commands.

This is helpful for:
- library consumers that perform schema validation separately before the deployment
- temporary situations where schemas of packages are incorrect or a user needs more time to conform to schema changes (and knowingly accepts skipping)

## Checklist before merging

### DIFF
--- a/src/cmd/package_test.go
+++ b/src/cmd/package_test.go
@@ -132,30 +132,6 @@ func TestPackageList(t *testing.T) {
 	}
 }
 
-func TestSkipValuesSchemaValidationFlags(t *testing.T) {
-	t.Parallel()
-	v := viper.New()
-
-	commands := []*cobra.Command{
-		newPackageDeployCommand(v),
-		newPackageInspectManifestsCommand(v),
-		newPackageInspectValuesFilesCommand(v),
-		newInitCommand(),
-	}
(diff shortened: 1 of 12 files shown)


src/cmd/package_test.go
```diff
@@ -132,6 +132,30 @@ func TestPackageList(t *testing.T) {
 	}
 }
 
+func TestSkipValuesSchemaValidationFlags(t *testing.T) {
```
AustinAbro321: IMO we should delete this, not a value add

AustinAbro321 (changes requested):
One quick comment otherwise looks good. 

My hope and expectation is that this is used rarely, but I think an escape hatch does make sense

AustinAbro321 (approved):

brandtkeller (approved):
In favor of the changes - I do think not allowing viper configuration to drive this behavior is the right call. Otherwise something something on the diagnostics end that I want to think about is persisting something like this in state so that the action is traceable. 

lgtm.

PR #5197, opened 2026-08-07
fix: negotiate oci scheme for allowing plain http oci endpoints

### LINKED ISSUE #5193: package publish --plain-http cannot copy from HTTPS OCI source to HTTP destination
Title: package publish --plain-http cannot copy from HTTPS OCI source to HTTP destination
### Environment
Device and OS: osx v26
App version: v0.81.0 through v0.83.0
Kubernetes distro being used: Not applicable
Other:
- Source registry: GHCR (`ghcr.io`, HTTPS)
- Destination registry: local OCI registry at `localhost:8080` (HTTP)
- Docker credentials for GHCR are configured in `~/.docker/config.json`
- Related: #5044, #5047

### Steps to reproduce
1. Authenticate to GHCR:
   ```sh
   gh auth refresh -s read:packages
   gh auth token | docker login ghcr.io --username YOUR_GITHUB_USERNAME --password-stdin
   ```
2. Start or use an HTTP OCI registry at localhost:8080.
3. Run:
```
zarf package publish \
  "oci://ghcr.io/..." \
  "oci://localhost:8080/..." \
  --plain-http \
  -a amd64
```
4. Verify that the same credentials work independently:
```
docker manifest inspect \
  ghcr.io/...

oras manifest fetch \
  --registry-config "$HOME/.docker/config.json" \
  ghcr.io/...
```

## Expected result
The package is copied successfully using HTTPS for the GHCR source and HTTP for the local destination.
The CLI should provide a way to configure the source and destination transports independently, equivalent to:
```
oras cp \
  --from-registry-config "$HOME/.docker/config.json" \
  --to-registry-config "$HOME/.docker/config.json" \
  --to-plain-http \
  ghcr.io/... \
  localhost:8080/...
```

## Actual Result
The Zarf command fails with:
could not copy package: GET "https://ghcr.io/v2/...": response status code 401: unauthorized: authentication required
The credentials are valid, and the ORAS command succeeds.
Zarf currently has only the global --plain-http flag. There is no supported way to express HTTPS for the source and HTTP only for the destination. The oci:// prefix identifies an OCI reference and does not provide per-registry transport configuration.
Visual Proof (screenshots, videos, text, etc)
The command and error output above demonstrate the failure.
Severity/Priority
High. This prevents copying packages from authenticated HTTPS registries such as GHCR to local HTTP registries.

## Additional Context
This is a follow-up to #5044. PR #5047 added per-host transport negotiation for several retrieval paths, but the OCI package-copy path still passes the same PlainHTTP setting to both source and destination remotes.
The behavior was observed with the following published versions:
- v0.81.0: Fail
- v0.81.1: Fail
- v0.82.0: Fail
- v0.83.0: Fail

Zarf v0.80.0 succeeds with the same command.

The fix should preserve ORAS security validation for authorization headers, bearer-token realms, redirects, and blob upload locations. Suggested regression coverage should verify:
- Authenticated HTTPS source to HTTP destination
- HTTPS source to HTTPS destination
- HTTP source to HTTP destination
- Source requests remain HTTPS
- Destination requests use HTTP
- Package metadata and all layers are copied successfully

brandtkeller:
## Description

Changes the behavior of `--plain-http` to _allow_ the use of http endpoints (With negotiation) instead of forcing it for all connections. 

This changes the intent of the flag slightly and the language has been adjusted accordingly. 

This was fixed not only in the publish workflow but also any workflows that need to allow doing so. 

This is a root level flag - expect a lot of noise in the docs generated.

## Related Issue

Fixes #5193

## Checklist before merging


src/pkg/packager/pull.go
```diff
@@ -126,16 +129,34 @@ type pullOCIOptions struct {
...
+		return nil, fmt.Errorf("could not resolve source registry transport: %w", err)
+	}
+
 	cacheMod, err := zoci.GetOCICacheModifier(ctx, opts.CachePath)
 	if err != nil {
 		return nil, err
 	}
 	platform := oci.PlatformForArch(opts.Architecture)
-	remote, err := zoci.NewRemote(ctx, opts.Source, platform, oci.WithPlainHTTP(opts.PlainHTTP), oci.WithInsecureSkipVerify(opts.InsecureSkipTLSVerify), cacheMod)
+	remote, err := zoci.NewRemote(ctx, opts.Source, platform, oci.WithPlainHTTP(plainHTTP), oci.WithInsecureSkipVerify(opts.InsecureSkipTLSVerify), cacheMod)
```
AustinAbro321: What would you think of having the evaluation to whether or not a source should respect the plain-http to be in `zoci.NewRemote`? This way we don't have to remember to check a registry scheme when creating a new remote especially zoci is Zarf opinionated. 

Also zoci changes docker.io to registry-1.docker.io and `defenseunicorns` to `ghcr.io/defenseunicorns` and this skips that. I think the former redirects and the latter we might want to intentionally stop supporting, but figured I'd call it out.

AustinAbro321 (commented):
design question, change looks good

src/pkg/packager/pull.go
brandtkeller: Probably necessitates a breaking change but overall it would be a lot cleaner for the standard use-cases. 

It does raise to the question as to whether we support the oci modifiers at all anymore if we're making a breaking change. If we accept an argument for plainhttp then the only other modifiers are InsecureSkipVerify and the cache configuration. It feels a little awkward to support two paths for setting the same configuration (Do users know which to set?).

My first thought is use `types.RemoteOptions` for plainhttp/skipverify and then accept a cachePath and remove the variadic oci modifiers all together but I'd be open to keeping them if we feel that escape hatch is useful.

brandtkeller:
Will update following #5227

src/pkg/packager/pull.go
brandtkeller: Added a new function and deprecated the old.

AustinAbro321 (approved):

Issue #5003, opened 2026-06-25
Implement `zarf tools registry` with oras-go, instead of pulling through Crane


AustinAbro321:
### Is your feature request related to a problem? Please describe.

`zarf tools registry` is currently an alias for the Crane CLI. The Crane CLI often has trouble with non container images, during pull for instance, it treats every layer of an image as a tar file, even though many OCI images have non tar layers now. 

This will give us more flexibility to add functionality to commands. For instance we could make a new command or add a flag to `zarf tools registry pull` which would pull layers by annotations similar to `oras pull`. 

Most commands would be rather simple to implement, for instance we already have functionality in Zarf to pull and push images, and oras-go can be leaned on for other commands. 

We should remove the alias `zarf tools registry Crane`

`zarf tools registry export` would be hard to duplicate, I also consider it in the domain of working with container images rather than working with a registry, and so I believe it would be out of scope and should be deprecated / removed in v1

### Alternatives

We could use the oras CLI, but then we are beholden to another tool, and I find the UX of ORAS CLI to be confusing at times.

abhishekgit03:
@AustinAbro321 I would like to take this on. Given the size, I'm thinking of splitting it into a few smaller PRs instead of one big one, starting with the simpler read-only commands (`digest`, `manifest`, `list`) to nail down the oras-go pattern, then `pull`/`push`/`delete`, then `prune` and `login`/`logout` last. Sound reasonable, or would you scope it differently? Happy to hear your thoughts on how to approach this.

AustinAbro321:
I went ahead and assigned you. I would recommend making each command its own separate PR, starting with the smallest commands. Your ordering is roughly accurate

PR #5199, opened 2026-08-08
feat: migrate `zarf tools registry ls` from crane to oras-go

abhishekgit03:
Related to #5003

First PR in the phased migration of `zarf tools registry` from Crane to ORAS, as discussed in the issue. This one covers `list` (`ls`), the smallest command, to establish the pattern the rest will follow (`digest`/`manifest` next, per the agreed ordering).

### Why

`zarf tools registry` currently wraps Crane's own cobra commands directly, backed by `go-containerregistry`. Crane assumes every image layer is a tar file, which breaks on OCI images with non-tar layers. `oras-go` doesn't make that assumption.

### What changed

- New `src/cmd/registry_list.go` reimplements `ls` using `oras-go` instead of `craneCmd.NewCmdList`, preserving the exact same flags (`--full-ref`, `-O`/`--omit-digest-tags`), example text, and behavior.
- `src/cmd/crane.go`: swapped the Crane-wrapped `list` registration for the new command. One line changed.
- Reuses existing patterns rather than inventing new ones: `images.NewAuthClientFromDocker` for the Docker-credential fallback path, the same `orasRemote.Repository` construction shape already used in `src/pkg/images/pull.go`.

### Zarf-managed registry support

When the target repo matches a Zarf-managed cluster's registry, this transparently tunnels to it and authenticates with Zarf's own registry credentials (mirroring what `zarfCraneInternalWrapper` does for the not-yet-migrated commands). Two things worth flagging for review:

- The credential must be keyed to whichever host ORAS actually connects to (the tunnel's local endpoint, not the original registry address)
- For mTLS-secured registries, scheme resolution uses `RegistryInfo.ResolvePlainHTTP`, which already knows the answer from state without needing to probe the network - a generic probe would need to present a client certificate it doesn't have.

### Test plan

- [x] Unit tests in `src/cmd/registry_list_test.go` (default/`--full-ref`/`--omit-digest-tags` behavior, invalid-ref error path, a regression test proving the known-scheme value is trusted rather than silently re-probed)
- [x] `make docs-and-schema` - zero diff, confirming generated CLI docs are unchanged from the Crane-wrapped version
- [x] Tested live against a real cluster
- [x] `go build ./...`, `golangci-lint run ./...`, `pre-commit run --all-files`


a1994sc:
This is a personal question, but do you mind expanding on this?

> Crane assumes every image layer is a tar file, which breaks on OCI images with non-tar layers. oras-go doesn't make that assumption.

abhishekgit03:
> This is a personal question, but do you mind expanding on this?
> 
> > Crane assumes every image layer is a tar file, which breaks on OCI images with non-tar layers. oras-go doesn't make that assumption.

 image layers used to always be gzipped tar archives (a filesystem diff), so Crane's layer-handling code assumes that. The OCI spec has since moved on, a manifest's layers can now be arbitrary blobs (WASM binaries, SBOMs, signatures, anything), not just tar diffs. Crane never updated that assumption, so it breaks when it tries to read non-tar content as if it were a tar stream. `oras-go` treats layers as opaque blobs by media type from the start, so it doesn't hit this.

a1994sc:
> > This is a personal question, but do you mind expanding on this?
> > > Crane assumes every image layer is a tar file, which breaks on OCI images with non-tar layers. oras-go doesn't make that assumption.
> 
> image layers used to always be gzipped tar archives (a filesystem diff), so Crane's layer-handling code assumes that. The OCI spec has since moved on, a manifest's layers can now be arbitrary blobs (WASM binaries, SBOMs, signatures, anything), not just tar diffs. Crane never updated that assumption, so it breaks when it tries to read non-tar content as if it were a tar stream. `oras-go` treats layers as opaque blobs by media type from the start, so it doesn't hit this.

Thank you very much for that description!

src/cmd/registry_list.go
```diff
@@ -0,0 +1,176 @@
...
+	endpoint, tunnel, err := c.ConnectToZarfRegistryEndpoint(ctx, s.RegistryInfo)
+	if err != nil {
+		return registryConnection{}, err
+	}
+	conn.tunnel = tunnel
+
+	// Credential must be keyed to the host ORAS actually connects to: the tunnel when tunneling, otherwise the registry's own address.
+	credentialHost := s.RegistryInfo.Address
+	if tunnel != nil {
+		credentialHost = endpoint
```
AustinAbro321: The credentialHost should be set from a function like below which will take the conn.Ref.
```go
func registryHost(repoRef string) (string, error) {
	ref, err := registry.ParseReference(repoRef)
	if err != nil {
		return "", fmt.Errorf("parsing repo %q: %w", repoRef, err)
	}
	return ref.Host(), nil
}
```

abhishekgit03:
Hi @AustinAbro321 @brandtkeller can you have a look at this when you get a chance? let me know if the approach is fine.

src/cmd/registry_list.go
```diff
@@ -0,0 +1,176 @@
...
+
+	if s.RegistryInfo.ShouldUseMTLS() {
+		t, err := getZarfRegistryMTLSTransport(ctx, c)
+		if err != nil {
+			return registryConnection{}, err
+		}
+		client.Client.Transport = t
+	}
+
+	plainHTTP, err := s.RegistryInfo.ResolvePlainHTTP(ctx, credentialHost, false, ocischeme.ProbeOptions{InsecureSkipTLSVerify: insecure})
```
AustinAbro321: In crane `--insecure` means both `--plain-http` and `--insecure-tls-skip-verify`. I'd like to get rid of this design as it overloads a flag, but I don't want to have a breaking change without a deprecation process.

Deprecate the insecure flag (for this command specifically), and add the `--plain-http` and `--insecure-skip-tls-verify` flags to `zarf tools registry ls`. As we go through other commands, we'll follow the same process.

AustinAbro321 (changes requested):
Overall design looks good, a few comments on keeping parity.

src/cmd/registry_list.go
abhishekgit03: done. added `--plain-http` and `--insecure-skip-tls-verify`

src/cmd/registry_list.go
abhishekgit03: fixed by using registryHost() method.

src/cmd/registry_list.go
AustinAbro321, line 1: I noticed one more regression where docker.io is not the default registry after this change. For instance, on main, `zarf tools registry ls stefanprodan/podinfo:6.4.0` will work, but it will fail on this branch. 

You should be able to use `reference.ParseNormalizedNamed` for this, we use similar functions in other spots in the repository.

AustinAbro321 (commented):
Noticed one more regression, overall looks good

src/cmd/registry_list.go
abhishekgit03: Fixed. Turns out oras-go's `ParseReference` needs an explicit registry, Crane just defaulted bare names to `docker.io` on its own. Used `reference.ParseNormalizedNamed` for this, same as `transform.ParseImageRef` already does.

src/cmd/registry_list.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	return cmd
+}
+
+func normalizeRepoRef(repoRef string) (string, error) {
+	named, err := reference.ParseNormalizedNamed(repoRef)
+	if err != nil {
+		return "", fmt.Errorf("parsing repo %q: %w", repoRef, err)
+	}
+	return named.Name(), nil
```
AustinAbro321, line 66:
```suggestion
	named, err := reference.ParseNormalizedNamed(repoRef)
	if err != nil {
		return "", fmt.Errorf("parsing repo %q: %w", repoRef, err)
	}
	if !reference.IsNameOnly(named) {
		return "", fmt.Errorf("repo %q must not include a tag or digest", repoRef)
	}
	return named.Name(), nil
```
Realized that we should also fail if a tag or digest is given. Consistent with the current command, and if someone is sending in a particular tag they may expect only that tag or digest back

abhishekgit03: @AustinAbro321  done, fixed it.

abhishekgit03:
Hi @AustinAbro321 , let me know if the implementation is fine and if there are anymore changes that need to be done.

abhishekgit03:
hi team, lemme know if this looks good so that i can proceed with the migration of other commands. 
@brandtkeller @AustinAbro321

AustinAbro321 (approved):
LGTM, thanks for your patience as I was on PTO for the past two weeks

a1994sc:
LGTM

a1994sc (approved):

PR #5206, opened 2026-08-10
feat: introduce zarf dev template

### LINKED ISSUE #5167: Introduce zarf dev template
The command `zarf dev template` will be introduced to template v1beta1 packages and component configurations.  

Relates to https://github.com/zarf-dev/zarf/issues/3433

Follows implementation from https://github.com/zarf-dev/proposals/tree/main/0048-schema-update-process and https://github.com/zarf-dev/proposals/tree/main/0051-v1beta1-schema

AustinAbro321:
## Description

Introduces `zarf dev template` according to https://github.com/zarf-dev/proposals/blob/main/0051-v1beta1-schema/README.md#package-templates

I decided to add schema validation. We could do a full on load, but that might be heavy, especially if it's importing oci packages. Feels like a reasonable middle ground, but open to removing this. Added a `--skip-validation` flag as well since there might be some other yq post processing. 

## Related Issue

Fixes #5167 

## Checklist before merging


src/cmd/dev_template.go
```diff
@@ -0,0 +1,167 @@
...
+	setFile        string
+	skipValidation bool
+}
+
+func newDevTemplateCommand(v *viper.Viper) *cobra.Command {
+	o := &devTemplateOptions{}
+	cmd := &cobra.Command{
+		Use:    "template [ TEMPLATE_FILE | DIRECTORY ]",
+		Short:  "Renders a template into a generated definition",
+		Hidden: true,
```
soltysh, line 42: Why hidden?

src/cmd/dev_template.go
```diff
@@ -0,0 +1,167 @@
...
+			return fmt.Errorf("unable to access template values file %q: %w", o.setFile, err)
+		}
+		valuesFiles = []string{o.setFile}
+	}
+	values, err := parseValues(ctx, valuesFiles, o.set)
+	if err != nil {
+		return fmt.Errorf("unable to parse package template values: %w", err)
+	}
+	// cli is reserved so templates can reliably use [[ .cli.version ]] to create versioned init packages.
+	values["cli"] = map[string]any{"version": config.CLIVersion}
```
soltysh:
```suggestion
	values["cli"] = new(value.Values{"version": config.CLIVersion})
```
seems more appropriate to match the type

src/cmd/dev_template.go
```diff
@@ -0,0 +1,167 @@
...
+	kind, err := validateTemplateAPIVersion(source, rendered)
+	if err != nil {
+		return err
+	}
+	if !o.skipValidation {
+		if err := validateTemplateSchema(source, rendered, kind); err != nil {
+			var lintErr *lint.LintError
+			if errors.As(err, &lintErr) {
+				PrintFindings(ctx, lintErr)
+			}
```
soltysh:
```suggestion
			if lintErr, ok := errors.AsType[*lint.LintError](err); ok{
				PrintFindings(ctx, lintErr)
			}
```
is the recommended for go1.26+, and also shorter

src/cmd/dev_template.go
```diff
@@ -0,0 +1,167 @@
...
+
+func templateSourcePath(path string) (string, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		return "", fmt.Errorf("unable to access package template %q: %w", path, err)
+	}
+	if info.IsDir() {
+		path = filepath.Join(path, packageTemplateFilename)
+	}
+	if !strings.HasSuffix(path, ".tpl.yaml") {
```
soltysh: You're using `.tpl.yaml` a few times, probably worth pulling it into a constant.

src/cmd/dev_template.go
```diff
@@ -0,0 +1,167 @@
...
+		if err := validateTemplateSchema(source, rendered, kind); err != nil {
+			var lintErr *lint.LintError
+			if errors.As(err, &lintErr) {
+				PrintFindings(ctx, lintErr)
+			}
+			return err
+		}
+	}
+	output := generatedTemplatePath(source)
+	if err := os.WriteFile(output, rendered, 0o644); err != nil {
```
soltysh, line 100: Should we somehow ensure that `output` is a file, if it exists?

src/cmd/dev_template.go
AustinAbro321: Since v1beta1 packages don't actually work yet, and this is only for v1beta1, keeping it hidden for now. Added a comment

src/cmd/dev_template.go
AustinAbro321: IMO that's not super valuable since `os.WriteFile` will already fail if it tries to write over a directory

soltysh (approved):
lgtm

PR #5212, opened 2026-08-11
refactor: move template out of internal

### LINKED ISSUE #5211: The `action.Run` function takes in an internal only object, and can not be use by external modules

The method `Run` in `github.com/zarf-dev/zarf/src/pkg/packager/actions` takes in a `stateAccess template.StateAccess` which is in support of the template effort. However the `template` package is under `src/internal/packager/template`, this means that modules outside of `Zarf` are not able to use the actions logic.

https://github.com/zarf-dev/zarf/blob/d7f041aaec22396b3005dec15e5d88e212d26094/src/pkg/packager/actions/actions.go#L21-L23

https://github.com/zarf-dev/zarf/blob/d7f041aaec22396b3005dec15e5d88e212d26094/src/pkg/packager/actions/actions.go#L33-L44

### Additional context
Having talked with Austin off github, this can likely be refactored to moved out of internal into another package

a1994sc:
## Description

This PR moves the previously internal package to under `src/pkg/packager`

## Related Issue

Fixes #5211

## Checklist before merging


.prekignore
AustinAbro321, line 1: This is for https://github.com/j178/prek correct?

src/pkg/packager/actions/actions.go
```diff
@@ -20,8 +20,8 @@ import (
 	"github.com/goccy/go-yaml"
 	"github.com/zarf-dev/zarf/src/api/v1alpha1"
 	ptmpl "github.com/zarf-dev/zarf/src/internal/packager/template"
-	"github.com/zarf-dev/zarf/src/internal/template"
 	"github.com/zarf-dev/zarf/src/pkg/logger"
+	"github.com/zarf-dev/zarf/src/pkg/packager/template"
```
AustinAbro321: I think I'd rather have the template path be `src/pkg/template` than `src/pkg/packager/template`. While it does have functions like `WithPackage` it is generally agnostic, and could be used outside of the context of a Zarf package.

AustinAbro321 (commented):
Small comment and question, but I agree with the direction

.prekignore
a1994sc: Yes sir, I personally use prek as a replacement for pre-commit and one of its features is that it searches recursively for pre-commit configs

src/pkg/packager/actions/actions.go
a1994sc: I am totally fine with that, mostly guessed where to put it, ha

AustinAbro321 (approved):
LGTM. I removed the breaking change denotation from the title since template was already internal, and so no one else was relying on it

a1994sc:
I'm fine with that, thank you Sir

brandtkeller (approved):
lgtm

PR #5219, opened 2026-08-13
feat!: publish v1beta1 component configs

AustinAbro321:
## Description

Implements publishing for v1beta1 component configs. 

There are a few notable decisions here: 
- ComponentConfigs now have a `.Metadata.variant.Architecture` and `.Metadata.variant.Flavor` field and lose the selector. See https://github.com/zarf-dev/proposals/pull/68
- Component configs can be signed as regular OCI objects. Since they only have a remote representation, they won't have a checksums.txt
- Files are copied to the OCI cache, but also duplicated on import, this takes up more space than skeleton packages. I think we could improve this in the future by hard linking the filepaths to the cache, but we'll want pkg/oci to expose the Cachepath() for a clean interface. 
- Image archives now accept a either a tar file or a directory of an oci layout. This eliminated the need to compress archives from skeleton packages before pulling the image in. 
- OnCreate actions are not allowed in remote components. We can reconsider this later, or even run them on publish, but this prevents running host actions when importing a remote component. 

## Breaking changes

The core of the breaking changes in this PR was to decouple the path to the package after loading is done. A new object `load.ResourceSet` knows how get any resource from the package given it's entry in the package spec. 

- `load.PackageDefinition` now returns `api.PackageDefinition` instead of `load.ResolvedPackage`, and does not pull remote resources or resolve values schemas; use `load.Package` when resolved resources, values, or schemas are needed.
- `load.ResolvedPackage` now exposes `Definition`, `Resources`, `Values`, and `ValuesSchema`. It must be passed by pointer and closed with `Close()` after use.
- `assemble.AssemblePackage` changed from `(ctx, resolvedPackage, packagePath, opts)` to `(ctx, *resolvedPackage, opts)`.
- `assemble.AssembleSkeleton` changed from `(ctx, resolvedPackage, packagePath, opts)` to `(ctx, *resolvedPackage, opts)`.
- `assemble.PackageManifest` and `assemble.PackageChart` now take `*load.ResourceSet` instead of a package-root path string.

## Related Issue

Relates to https://github.com/zarf-dev/zarf/issues/5165

## Checklist before merging


src/pkg/component/resource_normalizer.go
```diff
@@ -0,0 +1,213 @@
...
+
+	for i := range component.Component.Files {
+		component.Component.Files[i].Source, err = addResource(&normalizer, component.Component.Files[i].Source)
+		if err != nil {
+			return component, normalizedComponentResources{}, err
+		}
+	}
+
+	if hasActions(component.Component.Actions.OnCreate) {
+		return component, normalizedComponentResources{}, fmt.Errorf("onCreate actions are not supported for published remote components")
```
brandtkeller, line 126: This is a great fundamental requirement - can we strictly gate it on import as well? The SDK is getting easy enough to use that I could see attackers circumventing this and publishing valid components with `onCreate` actions that would arbitrarily run on create.

AustinAbro321: Definitely, I noted it as a TODO in import_v1beta1.go since we'll also have to do something about maliciously crafted paths, and this PR is so big already.

brandtkeller: I love it when I'm prompting through a review and the agent is willing to confirm suspicions and signals but not just tell me that it is intentional 😆 . 

Will that todo be worked against the same issue?

AustinAbro321: Yes, added an additional comment to in the issue to track.

src/pkg/component/publish.go
```diff
@@ -0,0 +1,308 @@
...
+	component, resources, err := normalizeComponentResources(componentPath, component, resolved.ImportedSchemas, resourceSet)
+	if err != nil {
+		return registry.Reference{}, err
+	}
+	componentJSON, err := json.Marshal(component)
+	if err != nil {
+		return registry.Reference{}, fmt.Errorf("unable to marshal component config: %w", err)
+	}
+
+	store := memory.New()
```
brandtkeller: Mental note that this is staging the resources in-memory. Probably not something we need to optimize without evidence but I do imagine someone could exhaust memory capacity with a large component. Maybe if we see signals we could establish a threshold for in-memory versus on-disk.

src/pkg/packager/load/package.go
```diff
@@ -0,0 +1,226 @@
...
+	if loaded.ValuesSchema != nil && !opts.SkipValuesSchemaValidation {
+		if err := loaded.Values.ValidateAgainstSchema(ctx, loaded.ValuesSchema, "resolved package values schema", value.ValidateOptions{SkipRequired: true}); err != nil {
+			return nil, fmt.Errorf("values validation failed: %w", err)
+		}
+	}
+
+	return loaded, nil
+}
+
+func materializeResources(ctx context.Context, packageRoot string, remoteResources []remoteResource) (_ *ResourceSet, err error) {
```
brandtkeller, line 187: Two scenarios came to mind reading this:
1. What happens when the package imports the same component multiple times?
2. What happens when the above occurs and the digest changes between imports? (something we've been discussing recently)

I think the cache largely handles the efficiency side.

src/pkg/component/publish.go
AustinAbro321: Ah I mean since we support image archives, it's not unreasonable to think it could get quite big. Decided to change to an on disk store.

src/pkg/packager/load/package.go
AustinAbro321: 1. The files from the second later component overwrites the files in the first component. It is generally cached, but if no cache is given then it won't be.
2. I added a remote reference cache to prevent this. I was 50-50 on whether to ignore this as an inherit risk of using mutable tags

src/pkg/component/publish.go
```diff
@@ -0,0 +1,354 @@
...
+
+	store, err := ocistore.NewWithContext(ctx, stagingDir)
+	if err != nil {
+		return registry.Reference{}, fmt.Errorf("unable to create component staging store: %w", err)
+	}
+	configDescriptor := content.NewDescriptorFromBytes(layout.ZarfComponentConfigMediaType, componentJSON)
+	if err := store.Push(ctx, configDescriptor, bytes.NewReader(componentJSON)); err != nil {
+		return registry.Reference{}, fmt.Errorf("unable to stage component config: %w", err)
+	}
+	layers, err := stageComponentResources(ctx, store, resources)
```
brandtkeller, line 114: Testing the publish of a local chart, it appears as though every local chart file is getting its own layer. I don't think that is inherently bad but I do wonder how it responds to large charts with many files each. Still thinking about the tradeoffs versus how skeleton managed charts.

src/pkg/signing/cosign.go
```diff
@@ -236,6 +236,83 @@ func CosignSignBlobWithOptions(ctx context.Context, blobPath string, opts SignBl
 	return sig, nil
 }
 
+// CosignSignImageWithOptions signs an OCI artifact by its immutable digest and uploads the
+// signature to its registry.
+func CosignSignImageWithOptions(ctx context.Context, image string, opts SignBlobOptions, plainHTTP, insecureSkipTLSVerify bool) error {
```
brandtkeller: I'm going to write the issue for this now - but given that we're trending towards an attempt to migrate from cosign directly to sigstore in the future I think it warrants a few questions:
1. What is the expectation of retaining this public API in that future? Should we consider internal while vetting a change in intent for signing?
2. where did we land with verification? is it intended for future work? Trying to think through whether we encounter a scenario where we also need to sign an index.

src/pkg/component/publish.go
AustinAbro321: IMO the pros of keeping it as regular files is that it's slightly simpler to implement since it's treated the same as all other files and we'll get more cache hits though the files are small so it doesn't matter much. 

The pros of a tar file are that there are less layers, though we likely won't run into a problem of the manifest exceeding the 4mib limit from some registries until at least over 10-15 thousand layers.

I'd be in favor of keeping it as regular files, but I can see the argument either way.

src/pkg/signing/cosign.go
AustinAbro321: Yes verification will be separate https://github.com/zarf-dev/zarf/issues/5274 and I actually also removed signing from this PR to avoid complexity.

src/pkg/component/publish.go
brandtkeller: I'm good with that line of thought. We can adjust if needed or if limitations arise.

brandtkeller (approved):
lgtm

PR #5222, opened 2026-08-14
test: add unit coverage for generated SBOM contents

### LINKED ISSUE #5159: Add unit coverage for generated SBOM contents

Zarf’s SBOM tests primarily verify that files are generated and contain expected text, rather than validating the structured Syft JSON. A Syft upgrade can therefore change package discovery, relationships, or schema metadata without a focused regression signal.

Ideally unit tests establish both confidence in the underlying content as well as assert to expectations around what is included. I'd favor whatever the current content dictates as the expectation and use that to drive further discussion around content as requests and needs evolve. 

Given the SBOM creation logic - unit test(s) should exist that assert minimum expectations for strict content while still allowing the underlying format and content to evolve as they become more performant and/or change. 

### Additional context

E2E tests could provide coverage here but they would be slow and brittle and lack the contract that I think would be beneficial to add in the short-term.

gangadhar-res:
Fixes #5159

The existing SBOM tests only asserted non-empty output. This adds content-level assertions against the parsed syft-json documents:

- **createImageSBOM**: an image layer carrying an apk installed database must yield the expected packages (name/version/type apk), the zarf descriptor, source and schema metadata, and artifact relationships.
- **createFileSBOM**: a component tar carrying a requirements.txt must yield the expected python packages, the zarf descriptor and relationships, and write identical bytes to zarf-component-<name>.json.

Fixture expectations are derived from current output on main, so a Syft upgrade produces a focused, reviewable diff instead of silent drift.

Test-only change; go test ./src/pkg/packager/assemble/ green locally.

### DIFF
--- a/src/pkg/packager/assemble/sbom_test.go
+++ b/src/pkg/packager/assemble/sbom_test.go
@@ -11,6 +11,8 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/anchore/syft/syft/format/syftjson/model"
+	"github.com/anchore/syft/syft/pkg"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/empty"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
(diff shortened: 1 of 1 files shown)


src/pkg/packager/assemble/sbom_test.go
```diff
@@ -42,3 +50,147 @@ func TestCreateImageSBOMNonExistentCachePath(t *testing.T) {
 	require.NoError(t, err)
 	require.NotEmpty(t, b)
 }
+
+// sbomDocument captures the syft-json fields the content tests assert on.
+type sbomDocument struct {
```
brandtkeller: What are the tradeoffs for using the syft model directly instead of this projection? My read is that we do care about the syft schema in as much as detecting changes and impacts related to outputs. With a few modifications I think it can mesh with the testing nicely.

`"github.com/anchore/syft/syft/format/syftjson/model"`

brandtkeller (changes requested):
One observation otherwise this is helpful. Thanks for stepping in to pick this up.

src/pkg/packager/assemble/sbom_test.go
gangadhar-res: Okay, that's better. I went with a local projection to keep the test independent of syft's internals, but you're right that the schema is exactly what we want to notice changing here. Will swith to model.Document.

src/pkg/packager/assemble/sbom_test.go
```diff
@@ -42,3 +52,128 @@ func TestCreateImageSBOMNonExistentCachePath(t *testing.T) {
...
+	t.Helper()
+	var buf bytes.Buffer
+	tw := tar.NewWriter(&buf)
+	for name, content := range files {
+		require.NoError(t, tw.WriteHeader(&tar.Header{Name: name, Mode: 0o644, Size: int64(len(content))}))
+		_, err := tw.Write([]byte(content))
+		require.NoError(t, err)
+	}
+	require.NoError(t, tw.Close())
+	layer, err := tarball.LayerFromReader(bytes.NewReader(buf.Bytes()))
```
brandtkeller: Looks like this function is deprecated. `tarball.LayerFromOpener` may be the equivalent option.

brandtkeller (approved):
lgtm - thank you for the contribution! This helps create more confidence in our expectations around sboms.

gangadhar-res:
@brandtkeller thanks for review. please help merge this pull request. thanks

soltysh (approved):
lgtm

AustinAbro321 (approved):
Tried out the last several versions to make sure changes in syft updates won't break this unit test and it seems it will be fine.

PR #5225, opened 2026-08-14
refactor: separate RequiresCluster from RequiresState

mjnagel:
## Description

Cluster wait actions require a cluster but were missing from the `RequiresCluster` checks in both api versions. This was the intial problem I aimed to solve, by adding these actions to the `RequiresCluster` condition.

During initial review discovered that there was a broader problem - `RequiresCluster` was also meaning `RequiresState` (which cluster wait actions do not need). This PR splits apart those two checks and makes the necessary changes throughout deploy/remove/dev to use the right checks in the right spots.

## Checklist before merging


src/api/v1alpha1/component.go
```diff
@@ -103,14 +103,30 @@ func (c ZarfComponent) RequiresCluster() bool {
 	hasRepos := len(c.Repos) > 0
 	hasDataInjections := len(c.DataInjections) > 0
 	hasHealthChecks := len(c.HealthChecks) > 0
+	hasClusterWait := c.Actions.hasClusterWait()
```
brandtkeller: Agree on the correctness here - a component that has a cluster wait DOES require cluster access. 

I am stewing on how to capture the implications of requiring state now. As this change now requires setting up state when the wait action doesn't strictly require as much...

brandtkeller: I think this may require broader changes as the use of `component.RequiresCluster` is used during deploy/remove to manage state - whereas I don't think that a component solely comprised of a cluster wait should need state to be initialized (or the requirements such as RBAC) to do so. 

I think we may want to distinguish as much between "needs cluster" and "needs state" but that is broader to your changes. 

I believe currently a cluster wait can run without requiring zarf initialization and this would break that as-is.

mjnagel: I think I made the necessary changes here, but some of the deploy flow around this was new to me so welcome any feedback.

PR #5228, opened 2026-08-14
fix: replace mimetype with archives

brandtkeller:
## Description

As discussed in #5140 - we can replace an direct dependency on `mimetype` with the existing `archives` direct dependency. `mimetype` will then become an indirect dependency. 

## Related Issue

No associated issue

## Checklist before merging


AustinAbro321 (approved):
lgtm

soltysh (commented):
Not quite removal, but definitely move in the right direction for limiting the deps

PR #5230, opened 2026-08-14
fix: registry mode not handled properly in update-creds

Racer159:
## Description

This adds handling for registry mode to update creds so that nodeport  -> external -> managedmtls updates can be handled correctly across the different modes.

## Related Issue

Fixes #N/A

## Checklist before merging


src/pkg/state/state_test.go
```diff
@@ -229,6 +229,40 @@ func TestMergeStateRegistry(t *testing.T) {
...
+		newState, err := Merge(oldState, MergeOptions{
+			RegistryInfo: RegistryInfo{RegistryMode: RegistryModeNodePort},
+			Services:     NewServiceSet(RegistryKey),
+		})
+		require.NoError(t, err)
+		require.NotEqual(t, oldState.RegistryInfo.PushPassword, newState.RegistryInfo.PushPassword)
+		require.NotEqual(t, oldState.RegistryInfo.PullPassword, newState.RegistryInfo.PullPassword)
+	})
+
+	t.Run("internal to external preserves omitted passwords", func(t *testing.T) {
```
brandtkeller: Are there specific use-cases where we want to preserve them for a separate registry? This feels like it might leak the internal creds to an external registry.

src/cmd/zarf_tools.go
```diff
@@ -325,9 +325,18 @@ func (o *updateCredsOptions) run(cmd *cobra.Command, args []string) error {
 		l.Warn(lang.ArtifactServerDeprecated)
 	}
 
+	registryInfo := o.registryInfo
+	registryURLChanged := services.Has(state.RegistryKey) && cmd.Flags().Changed("registry-url")
```
brandtkeller: Should we consider validating `--registry-url` inputs here? I could see an empty address as something that could corrupt state.

src/cmd/zarf_tools.go
```diff
@@ -570,13 +592,26 @@ func (o *updateRegistryCredsOptions) run(cmd *cobra.Command, _ []string) error {
 		return errors.New("no registry is configured in the Zarf state; nothing to update")
 	}
 
+	registryInfo := o.registryInfo
```
brandtkeller: Can we pull this out into a shared helper to prevent any drift?

brandtkeller (changes requested):
Initial thoughts - I do wonder about how atomic rollback functionality is around mtls certificates but that is of a separate scope I believe and something we can defer to another issue.

src/pkg/state/state_test.go
Racer159: Yeah I wasn't sure what to do in this case so opted to keep them as they were in the older logic - in an _actual_ internal -> external case you would likely provide new credentials - there was a thought I had where someone might "migrate" their internal registry externally copying over the creds and that a "wipe out" of credentials might be surprising?  Although possibly the opposite might be true if they wanted to use an external "anonymous" registry and tried setting empty values (though I am not sure Zarf supports that - will do some digging).

Racer159: added some more checks around this - I agree it should reset things - let me know if this is what you were thinking?

src/cmd/zarf_tools.go
Racer159: done

src/cmd/zarf_tools.go
Racer159: done

src/cmd/zarf_tools.go
```diff
@@ -325,9 +325,14 @@ func (o *updateCredsOptions) run(cmd *cobra.Command, args []string) error {
 		l.Warn(lang.ArtifactServerDeprecated)
 	}
 
+	registryURLChanged := services.Has(state.RegistryKey) && cmd.Flags().Changed("registry-url")
```
AustinAbro321: This could also be set by Viper flag? Maybe something like this? 
```go
func optionIsConfigured(cmd *cobra.Command, v *viper.Viper, flagName, key string) bool {
      return cmd.Flags().Changed(flagName) || v.IsSet(key)
  }
```

AustinAbro321 (changes requested):
One small request, mostly looks good

src/cmd/zarf_tools.go
Racer159: done!

AustinAbro321 (dismissed):

src/cmd/zarf_tools.go
```diff
@@ -570,8 +618,14 @@ func (o *updateRegistryCredsOptions) run(cmd *cobra.Command, _ []string) error {
 		return errors.New("no registry is configured in the Zarf state; nothing to update")
 	}
 
+	registryURLChanged := optionIsExplicitlySet(cmd, v, "registry-url", VInitRegistryURL)
```
brandtkeller: Not blocking - but I wonder if we should also be checking if the provided url is the same as the existing url to determine if it has _actually_ changed? Are there implications here one way or another?

src/pkg/cluster/registry_test.go
```diff
@@ -0,0 +1,108 @@
...
+		expectedPort int
+	}{
+		{
+			name:         "missing Zarf registry is external",
+			address:      "localhost:31999",
+			expectedMode: state.RegistryModeExternal,
+		},
+		{
+			name:         "matching Zarf NodePort is internal",
+			address:      "http://localhost:28000",
```
brandtkeller: This address _with scheme_ is an interesting consideration because I assume it would break downstream handling on the image urls at some point. Making a mental note to ensure we normalize this _somewhere_.

brandtkeller (dismissed):
I think reviewing this highlighted some existing pain points. This is an improvement so I wouldn't block on them but leaving a few thoughts in the event we _should_ handle them now.

src/cmd/zarf_tools.go
Racer159: I don't think that we necessarily care if it has changed since there would be cases where you would want to roll credentials for the same registry for example.  I do thing this variable name is misleading though about what we are looking at here - I'll go ahead and change it to `registryURLSet` since we really only care about set status not actually if it is different in state.

src/pkg/cluster/registry_test.go
Racer159: This will break things downstream - definitely good catch - I think maybe better than normalization is an error in the validation flow this PR adds that it must be given as a schemeless address?  I think if we internally normalize then it could give the false impression that the user can choose scheme on the registry - right now Zarf probes "local" registries with non-local externals ones being forced to HTTPS.

src/cmd/zarf_tools.go
Racer159: done

src/pkg/cluster/registry_test.go
Racer159: done

AustinAbro321: Yeah good catch, and agree on erroring over normalizing

AustinAbro321 (approved):

brandtkeller (approved):
lgtm

PR #5234, opened 2026-08-17
fix: add values flags to dev deploy and init

### LINKED ISSUE #5231: Zarf Dev Deploy to use the Values feature

I am building out packages that make use of the values feature via a yaml file as its a much cleaner way of standardising lots of options for multiple zarf packages. In developing my packages and testing them, I want to make use of the `zarf dev deploy` feature to speed up my testing. Currently the command includes no support for the values. `zarf package deploy` does, so I may continue to use that, but it means a slower and more cumbersome development process.

I would like the `-v` option added to the `zarf dev deploy` in the same way that it is present on the `zarf package deploy` command.

- **Given** a zarf file that makes use of the values feature
- **When** I call `zarf dev deploy  --features="values=true" -v ./values.yaml`
- **Then** the deployment happens as per normal.

Building and deploying the package in two stages like before.

You can also potentially use incremental builds, though this isn't really what this is for.

AustinAbro321:
## Description

Add values flags to dev deploy. Purposely didn't add skip schema validation, as I feel at dev time the user should instead change their schema. 

Also adding the flags to `zarf init`

## Related Issue

Fixes #5231

## Checklist before merging


soltysh (approved):
lgtm

Issue #4124, opened 2025-08-25
Native data packaging mechanism with image volumes


brandtkeller:
### Is your feature request related to a problem? Please describe.

In relation to #2902 - as [image volume source]() becomes generally available, zarf should provide a native mechanism for getting data from a local filesystem (or otherwise) packaged and pushed into the zarf registry as an option for deployments to configure as a volume source. 

 while not exhaustive this might include:
```yaml
images:
- nginx:1.27.0
imageVolumes:
- name: data-volume:0.0.1 #image reference for use in the cluster>
  data: ./path/to/data/directory
```

### Describe the behavior you'd like

- **Given** the existence of data on the local filesystem during package create
- **When** zarf package create executes
- **Then** Zarf will package the data into an oci artifact for use with manifests/helm charts

### Additional context
Given that the future of data injections is uncertain - this would provide applicable replacements for data packaging.

We would expect the user to configure the use of these image volumes in their manifests or helm charts accordingly.

PR #5235, opened 2026-08-17
feat: add image volume archive function for use as mount

a1994sc:
## Description

Adds `zarf dev image-volume-archive` (alias `iva`), a dev command that packs a directory of files into a container image usable as a volume mount by containerd, cri-o, and docker.

Given a directory and an image reference, it:
- tars and (optionally) compresses each regular file in the directory as an individual OCI image layer (`gzip`, `zstd`, or uncompressed)
- caps the number of layers via `--max-layers` (default 127), batching files several-per-layer once the directory holds more files than the cap
- builds an OCI image config/manifest and stages it in a local OCI store
- writes the result as a Docker/OCI-compatible tar archive (`docker load`/`ctr image import` compatible) via `-O/--output`, defaulting to a filename derived from the image reference
- registers shell tab completion for `--layer-compression`, `--platform-os`, `--max-layers`, and the root `--architecture` flag

New packages:
- `src/pkg/zoci/image`: builds the layered OCI image (`Volume`, layer/diff ID/compression handling, platform validation)
- `src/pkg/zoci/archive`: read-only `content.Provider` over an OCI store for tar export, plus `ImageRefToTar`/`ValidateFileEndsWithTar` helpers for deriving and validating the output path

## Related Issue

Relates to #4124

## Checklist before merging


a1994sc:
Some code was generated with Claude, Sonnet 5, though a human verified the code.

src/cmd/cobra/root.go
```diff
@@ -0,0 +1,30 @@
...
+// shown alongside each supported --architecture value.
+const (
+	osArchAMD64Desc = "the x86-64, 64-bit AMD, architecture"
+	osArchARM64Desc = "the 64-bit ARM architecture"
+)
+
+// GetRootArchitectureCobraCompression returns the valid --architecture
+// values as "value\tdescription" pairs, ready to return from a cobra
+// RegisterFlagCompletionFunc callback for shell tab completion.
+func GetRootArchitectureCobraCompression() []string {
```
a1994sc, line 25: Realized that this is a bad function name.... I will update it soon

a1994sc:
So I opened the stacked prs #5348, #5349, #5350, and #5351 to be a bit more easily review-able. Those PR and this pr are code equal

brandtkeller (commented):
Commenting for posterity as I begin to work through what this means. Less so the specific implementation and more so the scope and boundaries as implications for the project. 

I'm a big believer in getting a capability for packaging data in a container/oci native mechanism - so I appreciate you taking this on. 

Questions that come to mind: 
- What is the expected API contract for this capability? Do we expect this to be something we support as an SDK capability? 
- Do we feel confident in supporting it and the direct dependencies it relies upon?
- Should this be a capability of this project? Or is it a better fit for a separate project?
- Is there prior art to draw from for the creation of an artifact this shape? Do we need an in-house built solution? 

These are not meant to be targeted questions at the work you've done here and I'm supportive of getting this capability into the hands of Zarf users. I would question general SDK utility - and rather think about placing some of this logic in internal. `zarf tools` also may be a more appropriate scope for protecting our general deprecation policy (WIP). I would also say that is likely entirely additive and not v1.0.0 scope - meaning I might advocate for closing the stacked PR's and possibly creating a ZEP for a capability this large. The "how" is amazing here but may be blocked by a clear "What" and "Why". I'll let other maintainers weigh in if they disagree.

a1994sc:
Closing in favor of the stacked PRs starter in #5348

PR #5238, opened 2026-08-18
fix(sbom)!: correct image detection for SBOM generation

### LINKED ISSUE #4756: packages with images included for OCI image volumes fail to create due to SBOM errors
### Environment
Device and OS: macOS Tahoe 26.3.1a
App version: v0.74.0
Other:

### Steps to reproduce
1. Include both a runtime container image and an image storing only data (created/pushed with the oras CLI, for example)
2. Create the package

### Expected result
The package creates

### Actual Result
The package fails to create

```shell
2026-03-26 16:13:29 INF generating SBOM
2026-03-26 16:13:29 INF creating image SBOM reference=some-runtime-image
2026-03-26 16:13:30 INF creating image SBOM reference=some-oci-storage-image
2026-03-26 16:13:30 ERR failed to create package: failed to generate SBOM: failed to create image sbom: failed to read layer="sha256:5f2d6e2ea1848e3dd16faf8560eed333559713aedc8d02f60d4b5427c513095a" tar : archive/tar: invalid tar header
```

### Severity/Priority

Medium

### Additional Context
Including `--skip-sbom` allows the package to build

theZMC:
## Description

This PR creates a more restrictive function, `IsContainerImage` to verify an OCI manifest has a known container `Config` `MediaType` _as well as_ `Layers` composed entirely of known image layer `MediaType`s.

~~I intentionally did not alter the existing `OnlyHasImageLayers` function because it's part of the public API surface and the new scope doesn't quite fit the naming.~~ The previous `OnlyHasImageLayers` function has been privatized per @AustinAbro321.

I also added a regression test to the existing SBOM creation e2e suite. Please let me know if there's a more appropriate place for it.

## Related Issue

Fixes #4756 

## Checklist before merging

### DIFF
--- a/src/pkg/images/common.go
+++ b/src/pkg/images/common.go
@@ -45,6 +45,8 @@ const (
 	DockerMediaTypeManifest = "application/vnd.docker.distribution.manifest.v2+json"
 	// DockerMediaTypeManifestList is the legacy Docker manifest list, replaced by OCI index
 	DockerMediaTypeManifestList = "application/vnd.docker.distribution.manifest.list.v2+json"
+	// DockerMediaTypeConfig is the legacy Docker image configuration format.
+	DockerMediaTypeConfig = "application/vnd.docker.container.image.v1+json"
 )
 
 // Legacy Docker image layers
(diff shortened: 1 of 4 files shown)


src/pkg/images/common.go
```diff
@@ -74,6 +76,16 @@ func OnlyHasImageLayers(manifest ocispec.Manifest) bool {
 	return true
 }
 
+// IsContainerImage reports whether an OCI manifest describes a container image.
+func IsContainerImage(manifest ocispec.Manifest) bool {
```
AustinAbro321, line 90: I'm reviewing the spec, and it seems like we should also check if artifact type is not empty. https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage

src/pkg/images/common.go
AustinAbro321: Lets just privatize this function, hard to imagine an SDK caller using us for this (and if they were, there's a good chance they'd want to switch to `IsContainerImage` anyway.

AustinAbro321 (changes requested):
Quick additional request based on re-reviewing the spec. Thanks for taking this!

src/pkg/images/common.go
theZMC: Good call. I was able to write a test case exercising exactly this and it failed with the previous implementation. I have added explicit checks for empty artifacts.

src/pkg/images/common.go
theZMC: Done, and I updated the conventional commit message to include breaking change information.

PR #5243, opened 2026-08-18
feat(agent): add variable for mutating webhook annotations

### LINKED ISSUE #4771: Add support for webhook annotations via Zarf values/variables

On an upgrade of the zarf init package the agent fails to deploy (unless I use `--force-conflicts`) due to an SSA conflict with a change made by AKS:
```
2026-04-01 09:20:57 INF performing Helm upgrade chart=zarf-agent
2026-04-01 09:20:59 WRN upgrade failed name=zarf-d2db14ef40305397791454e883b26fc94ad9615d error=conflict occurred while applying object /zarf admissionregistration.k8s.io/v1, Kind=MutatingWebhookConfiguration: Apply failed with 8 conflicts: conflicts with "admissionsenforcer" using admissionregistration.k8s.io/v1:
- .webhooks[name="agent-argocd-application.zarf.dev"].namespaceSelector
- .webhooks[name="agent-argocd-applicationset.zarf.dev"].namespaceSelector
- .webhooks[name="agent-argocd-appproject.zarf.dev"].namespaceSelector
- .webhooks[name="agent-argocd-repository.zarf.dev"].namespaceSelector
- .webhooks[name="agent-flux-gitrepo.zarf.dev"].namespaceSelector
- .webhooks[name="agent-flux-helmrepo.zarf.dev"].namespaceSelector
- .webhooks[name="agent-flux-ocirepo.zarf.dev"].namespaceSelector
- .webhooks[name="agent-pod.zarf.dev"].namespaceSelector
```

The Zarf webhooks end up with this in `namespaceSelector`:
```
    - key: kubernetes.azure.com/managedby
      operator: NotIn
      values:
      - aks
```

**Note**: This is a new issue with SSA in Zarf 0.74.0, in the past this worked fine. This is also an upgrade only issue since it depends on AKS modifying the webhook and then Zarf trying to modify it again later.

In order to disable this behavior in AKS I need to be able to set `admissions.enforcer/disabled: "true"` as a webhook annotation. The current zarf agent chart/package does not expose webhook annotations however. This configuration should be exposed as a Zarf value and/or variable.

Note that while I encountered this with a specific AKS issue there also other use-cases such as cert-manager where it would be helpful to have native annotation support.

https://github.com/zarf-dev/zarf/issues/4731 could be an alternative to consider, and setting this behavior on the Zarf agent/entire init package.

### Additional context

There is an upstream AKS issue for this specific AKS problem - https://github.com/Azure/AKS/issues/4002 - but little traction in changing the behavior it seems.

AustinAbro321:
## Description

Provides a path for generic webhook annotations on the zarf agent

## Related Issue

Fixes #4771

## Checklist before merging


soltysh (dismissed):
lgtm

packages/zarf-agent/zarf.yaml
```diff
@@ -19,6 +19,11 @@ variables:
     default: ""
     autoIndent: true
 
+  - name: AGENT_WEBHOOK_ANNOTATIONS
+    description: Custom annotations map for the zarf-agent mutating webhook
+    default: ""
```
brandtkeller: nit: should the default be an object? 
```suggestion
    default: "{}"
```

Thinking through the webhook exclusions PR and intending to file an issue for chart values schemas (precursor to zarf package values schemas) and think this would be something that would get flagged being a yaml `null`

brandtkeller (dismissed):
Non-blocking comments - but similar to the exclusions PR I wonder if we should be more adamant about versioning our charts on changes.

packages/zarf-agent/zarf.yaml
AustinAbro321: Makes sense

brandtkeller (approved):

PR #5249, opened 2026-08-19
test: v1beta1 to v1alpha1 roundtrip fuzzing

AustinAbro321:
## Description

Roundtripping is not 1:1 between v1beta1 and v1alpha1. This PR includes tests which serve as living documentation that we can reference to know which fields are not 1:1. 

There is some extra logic done for charts and repos where oneOf is not respected by the fuzzer.

## Checklist before merging


src/api/v1beta1/component.go
```diff
@@ -184,7 +184,7 @@ type GitRef struct {
 	// The Git branch.
 	Branch string `json:"branch,omitempty" jsonschema:"oneof_required=branch"`
 	// The Git commit SHA.
-	Commit string `json:"commit,omitempty" jsonschema:"oneof_required=commit"`
+	Commit string `json:"commit,omitempty" jsonschema:"oneof_required=commit,pattern=^[0-9a-f]{40}$"`
```
brandtkeller: nit - uppercase is valid and accepted by Git I believe. We may also want to consider explicitly describing this as sha-1 as we would not accept sha-256.

src/internal/api/v1beta1/roundtrip_test.go
```diff
@@ -201,46 +205,158 @@ func TestConvertGenericRoundTripFuzz(t *testing.T) {
...
+//   - component.import has separate local and remote lists in v1beta1, while v1alpha1 has one
+//     import object; component.service has no v1alpha1 equivalent.
+//   - image.source distinguishes registry and daemon sources in v1beta1, v1alpha1 images always fallback
+//   - manifest.kustomize is a pointer in v1beta1 but flattened into v1alpha1 manifest fields. An
+//     empty Kustomize object therefore becomes nil on the return trip.
+//   - chart.valuesFiles is one ordered v1beta1 list. v1alpha1 separates plain and templated files,
+//     so their relative order is lost when the two kinds are interleaved.
+//   - actionSet.defaults is a pointer in v1beta1 but a value in v1alpha1, so nil and an explicitly
+//     empty defaults object cannot be distinguished.
+func v1beta1V1alpha1RoundTripExclusions() cmp.Options {
```
brandtkeller, line 352: Would any of these fields be good candidates for normalizing into their shared, losslessly representable form before comparison, rather than ignoring them wholesale?

brandtkeller (changes requested):
minor nit and question for my own education as it feels intentional.

src/api/v1beta1/component.go
AustinAbro321: Nice catch

src/internal/api/v1beta1/roundtrip_test.go
AustinAbro321: The reason I decided to normalize the other fields but not these, is because the other fields should be lossless so long as the schema is followed, while these fields are legitimately lossy.

brandtkeller (approved):
lgtm

src/test/testutil/fuzz.go
```diff
@@ -9,13 +9,17 @@ import (
...
-// fields that cannot be set via reflection (unexported) are left zero. Pointers are left nil roughly
-// half the time so both the set and unset states are exercised.
+// FillValue recursively populates v for round-trip fuzz tests. Struct fields that cannot be set via
+// reflection (unexported) are left zero. Pointers are generated as nil, empty, or populated so
+// their presence is tested independently from their contents.
 func FillValue(v reflect.Value, rng *rand.Rand) {
 	switch v.Kind() {
 	case reflect.Pointer:
-		if rng.Intn(2) == 0 {
+		switch rng.Intn(3) {
```
soltysh, line 18: Maybe turn this piece into something like:
```golang
		switch rng.Intn(3) {
		case 0:
			return
		case 1:
			v.Set(reflect.New(v.Type().Elem()))
			return
		default:
			v.Set(reflect.New(v.Type().Elem()))
			FillValue(v.Elem(), rng)
		}
```
to make it more explicit?

soltysh (approved):
Non-blocking suggestion, I might open a follow-up fixing it, since I'd like to expand our fuzzer a bit. 

lgtm

src/test/testutil/fuzz.go
AustinAbro321: Seems fair, I'll let you follow up

soltysh: Opened https://github.com/zarf-dev/zarf/pull/5275

Issue #5168, opened 2026-08-04
Persist multiple API versions in packages for cross Zarf compatibility


AustinAbro321:
Packages published to OCI, deployed to cluster, and built into tarballs should each store the original API version used at build time as well as earlier API versions. Older versions of Zarf should continue to be able to read API versions of packages that it doesn't know about. If the packages utilize new features, then were will be a `build.MinimumVersionRequirement` that specifies this. 

For OCI and tar packages, they will store the multiple versions in the zarf.yaml. DeployedPackages will persist multiple versions to the cluster. 

Relates to https://github.com/zarf-dev/zarf/issues/3433

Follows implementation from https://github.com/zarf-dev/proposals/tree/main/0048-schema-update-process and https://github.com/zarf-dev/proposals/tree/main/0051-v1beta1-schema

PR #5250, opened 2026-08-19
feat(v1beta1)!: store multiple API versions in Deployed packages

AustinAbro321:
## Description

This stores multiple API versions in deployed package objects. The rule is the current API version and all earlier API versions must be stored. This way old clients can always read newer packages. 

Eventually Data will be removed which is why we can't only store `v1alpha1` in PackageData, this does make the secrets bigger, but it is incredibly unlikely to occur. 

This also flips services to be set when converting from `v1alpha1` to `api.Package` instead of when moving from `api.Package` to v1beta1, which is more in line with our recent shift in model, where the behavior is normalized around `api.Package`. 

## Related Issue

Relates to #5168

## Breaking change

`RecordPackageDeployment` now accepts an `api.PackageDefinition` instead of a v1alpha1 package. If anyone was calling this, would be very easy to make a PackageDefinition using the v1alpha1 package, so I did not make a deprecated bridge.

## Checklist before merging


src/cmd/package.go
```diff
@@ -1354,16 +1354,20 @@ func (o *packageListOptions) run(ctx context.Context) error {
...
+			return err
+		}
 		packageList = append(packageList, packageListInfo{
-			Package:           pkg.Name,
-			NamespaceOverride: pkg.NamespaceOverride,
-			Version:           pkg.Data.Metadata.Version,
-			Connectivity:      pkg.GetPackageConnectivity(),
+			Package:           depPkg.Name,
+			NamespaceOverride: depPkg.NamespaceOverride,
+			Version:           pkg.AsV1alpha1().Metadata.Version,
```
brandtkeller, line 1369: Rather than perform conversion for a whole v1alpha1.ZarfPackage simply for this string - might we instead create a `Version()` accessor for `PacakgeDefinition`?

src/internal/packager/helm/zarf.go
```diff
@@ -216,15 +222,20 @@ func UpdateZarfAgentValues(ctx context.Context, opts InstallUpgradeOptions) erro
...
 				if c.Name == componentName {
-					return pkg.Name
+					return deployedPackage.Name, nil
 				}
 			}
 		}
 	}
-	return ""
+	return "", nil
 }
```
brandtkeller, line 241: nit: I think this is entirely functional but - correct me if I am wrong - feels like a potential temporary solution compared to something like a `PackageDefinition` method that could perform the conversion with more of a stable execution perhaps? This might also be intentional so I wouldn't block on it.

src/pkg/state/state.go
```diff
@@ -710,18 +713,76 @@ const (
...
-	Digest              string               `json:"digest"`
-	Data                v1alpha1.ZarfPackage `json:"data"`
-	CLIVersion          string               `json:"cliVersion"`
-	Generation          int                  `json:"generation"`
-	DeployedComponents  []DeployedComponent  `json:"deployedComponents"`
-	ConnectStrings      ConnectStrings       `json:"connectStrings,omitempty"`
-	PackageConnectivity PackageConnectivity  `json:"packageConnectivity"`
+	Name   string `json:"name"`
+	Digest string `json:"digest"`
+	// Deprecated: kept so older clients can read package deployment state.
```
brandtkeller, line 718: nit: a deprecation comment should name the replacement

brandtkeller (changes requested):
In a variety of ways this really starts to cleanup through the generic use. I like the trend and have a few minor comments.

src/internal/packager/helm/zarf.go
AustinAbro321: I didn't make this a package definition method, but I did change this around to use component services instead of hardcoded names, which will ensure it works with v1beta1 packages long term.

src/cmd/package.go
AustinAbro321: We could, but conversion is cheap. Maybe we'd want getters in the future, but I'm not sure about it and I think the list of getters could grow fast.

AustinAbro321: conversion is no longer necessary with the new setup

PR #5253, opened 2026-08-20
fix: packager.Remove() uses ZarfDefaultTimeout as default timeout

### LINKED ISSUE #5252: `packager.Remove()` default timeout causes instant error
### Overview
`packager.Remove()` does _not_ use [`ZarfDefaultTimeout`](https://github.com/zarf-dev/zarf/blob/main/src/config/config.go#L63) as its default Timeout. This behavior is inconsistent with `packager.Deploy()`, which _does_ use `ZarfDefaultTimeout` as its default Timeout. It is also surprising behavior to the user, who will see `packager.Remove()` fail immediately on invocation, unless they set the Timeout option.

### Environment
App version: v0.83.0

### Steps to reproduce
1. Deploy any package with `packager.Deploy()`
2. Wait for deployment to complete.
3. Call `packager.Remove()` on the package _without_ setting `Timeout` option.

### Expected result
`packager.Remove()` succeeds, or times out after some reasonable period if removal process gets stuck.

### Actual Result
`packager.Remove()` immediately fails, because effective timeout defaults to 0 seconds.

Inspect `Remove` function definition here: https://github.com/zarf-dev/zarf/blob/main/src/pkg/packager/remove.go#L43

### Severity/Priority
Low

### Recommended Solution
Consistent with `packager.Deploy()`, update `packager.Remove()` to set `Timeout` option to `ZarfDefaultTimeout` if it is received unset.

curtis-nichols:
## Description

Update `Remove` function of package `packager` to time out at  `ZarfDefaultTimeout` (15 minutes)
* `Remove` checks for unset value of `opts.Timeout`, and sets to `ZarfDefaultTimeout` if found
* Add doc comment to `RemoveOptions.Timeout` corresponding to that of `DeployOptions.Timeout`

## Related Issue

Fixes #5252

### DIFF
--- a/src/pkg/packager/remove.go
+++ b/src/pkg/packager/remove.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/zarf-dev/zarf/src/api"
+	"github.com/zarf-dev/zarf/src/config"
 	"github.com/zarf-dev/zarf/src/internal/packager/helm"
 	"github.com/zarf-dev/zarf/src/internal/packager/requirements"
 	"github.com/zarf-dev/zarf/src/pkg/feature"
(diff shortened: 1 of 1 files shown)


AustinAbro321 (approved):

brandtkeller (approved):
lgtm - appreciate the PR and as such we can squeeze it into the release later.

Issue #4307, opened 2025-10-23
feat: add `--context` flag for which kube-config context to use


a1994sc:
### Is your feature request related to a problem? Please describe.

When managing multiple clusters, I lean on using the `kubectl --context <> ...`  to make sure I'm working on the correct cluster at a given time.

Presently the way that Zarf works at the moment, there is no way to equivalent flag

### Describe the behavior you'd like

Example:
- **Given** `zarf package ls --context axolotl`
- **Then** it shows the installed packages for the `axolotl` kube-context

brandtkeller:
relates to #4280 

My opinion is that this as a CLI configuration makes the most sense. We'll want to think through the implications for what guardrails we can establish (if any) but I don't see a reason we would want to refute this personally. 

It may be a good time to add a `--kubeconfig` file while exposing this configuration as well.

a1994sc:
It would probably make sense to do both, I agree. At least as someone who maintains multiple environments, both `--kubeconfig` and `--context` are flags that I use hourly.

I'm curious if we will see any integration issues with all the various subtool we bring?

Like with a config set to `prod`

```shell
zarf tools kubectl --context dev get nodes
```

Now would show use the `dev` context. But would the following?

```shell
zarf --context dev tools kubectl get nodes
```

Issue #2509, opened 2024-05-15
Add a --kubeconfig flag in the root command


phillebaba:
### Is your feature request related to a problem? Please describe.

After working on #2497 I realized that there is no global Kubernetes configuration setting. This means that each tool is responsible for resolving the configuration file using their own hierarchy. This probably works for the most part as most tools will look for the env var `$KUBECONFIG` or in the path `~/.kube/config`. Each individual tool seems to either define its set of possible flags but there is no guaranteed overlap. It would be a lot better to have a single place to configure Kubernetes configuration for all tools.

### Describe the solution you'd like

- **Given** A Zarf CLI command that connects to k8s executes.
- **When** A `--kubeconfig` flag is passed in the root command.
- **Then** The `--kubeconfig` path should be used for client configuration.

### Describe alternatives you've considered

The alternative would be to leave things as it is and just allow each tool to define their own hierarchy of picking up configuration. 

### Additional context

N/A

lukasrabener:
this would be also helpful in case working with k3s, as zarf is not considering the default path for the kubeconfig `/etc/rancher/k3s/k3s.yaml`

joaofasilva-kitar:
Related with this issue, currently (AFAIK) we don't a proper way to specify per component a different kubeconfig. I would like to be able in a single package deploy for multiple clusters based on KUBECONFIG.

edit: created a different feature request issue https://github.com/zarf-dev/zarf/issues/4280 for proper tracking

PR #5254, opened 2026-08-20
feat: let library callers deploy through an existing cluster connection

### LINKED ISSUE #5223: Support configurable cluster connection for library users of Zarf

As a consumer of Zarf as a library in the [UDS Terraform Provider](https://github.com/defenseunicorns/terraform-provider-uds), I'd like to be able to configure my provider with cluster connection information. This would require me to pass that cluster connection information to Zarf for deploys/removals/etc.

At each entrypoint for "cluster interaction" (deploy, remove, etc) there is an option for an existing cluster connection (`*cluster.Cluster`). This cluster is wired downstream into all helm behaviors and actions.

Zarf allows passing a kubeconfig filepath at entrypoints to cluster interactions. This is likely a far simpler change for Zarf, and shifts more responsibility to library users to manage that kubeconfig file existing.

### Additional context

Actions might be the trickiest piece of this on the library side. Some actions may use `kubectl`/`helm` from Zarf tools and would need to honor the same kubeconfig/context. If the option/input is a kubeconfig filepath this is significantly simpler (through an injected env). 

There are two quite related tickets for CLI options:
- https://github.com/zarf-dev/zarf/issues/2509
- https://github.com/zarf-dev/zarf/issues/4307

Both would likely require some of the work for this and maybe could be seen as duplicates but I think there are different considerations with this ticket.

gangadhar-res:
Closes #5223

`RemoveOptions` already takes a `*cluster.Cluster`, but `Deploy` always connects itself from the ambient kubeconfig the first time a component needs a cluster, so a library consumer holding a connection had no way to hand it over. This adds the same option to `DeployOptions` and seeds the deployer with it, which leaves `isConnectedToCluster` true from the start so the lazy connect is skipped. Everything downstream, helm included, then runs against the caller's connection.

One thing worth pointing out. The one-time deployability check was sitting inside the connect branch, so it only ran on the path that had just connected. A caller-supplied connection would have skipped the architecture compatibility and agent checks entirely, which seemed worse than the problem being fixed, so it now runs for the first cluster-requiring component either way.

## Testing

Unit test seeds the supplied cluster with Zarf state and asserts the deploy gets past loading it, which it can only do through that connection since an ambient cluster wouldn't carry the same state. I checked it fails without the wiring, otherwise it isn't really testing anything.

## Not covered here

The issue also raises actions that shell out to `kubectl` or `helm` from Zarf tools and would need to honour the same connection. That needs a kubeconfig on disk rather than an in-memory client, so it's a separate question from this one and I've left it alone. Happy to look at it if you'd like it in the same change, though #2509 and #4307 seem to overlap with it too.

### DIFF
--- a/src/pkg/packager/deploy_test.go
+++ b/src/pkg/packager/deploy_test.go
@@ -136,3 +136,35 @@ func TestDeploySkipsValuesSchemaValidationWhenConfigured(t *testing.T) {
 	_, err = Deploy(ctx, pkgLayout, DeployOptions{SkipValuesSchemaValidation: true})
 	require.NoError(t, err)
 }
+
+func TestDeployUsesTheClusterTheCallerSupplies(t *testing.T) {
+	// A library caller that already holds a connection should not have Zarf reach for
+	// the ambient kubeconfig. The supplied cluster is seeded with Zarf state, so if
+	// the deploy reads state successfully it can only have used this connection: an
+	// ambient cluster would either be absent or not carry this state.
+	ctx := testutil.TestContext(t)
+	srcDir := filepath.Join("load", "testdata", "package-with-values-schema")
+	defined, err := load.PackageDefinition(ctx, srcDir, load.DefinitionOptions{})
+	require.NoError(t, err)
+	pkgLayout, err := assemble.AssemblePackage(ctx, defined, srcDir, assemble.AssembleOptions{SkipSBOM: true})
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, pkgLayout.Cleanup()) })
+
+	cs := fake.NewClientset()
+	c := &cluster.Cluster{
+		Clientset: cs,
+		Watcher:   healthchecks.NewImmediateWatcher(status.CurrentStatus),
+	}
+	_, err = cs.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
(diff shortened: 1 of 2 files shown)


brandtkeller:
Appreciate the patience while we work this into the review queue.

src/pkg/packager/deploy_test.go
```diff
@@ -136,3 +136,35 @@ func TestDeploySkipsValuesSchemaValidationWhenConfigured(t *testing.T) {
...
+		Clientset: cs,
+		Watcher:   healthchecks.NewImmediateWatcher(status.CurrentStatus),
+	}
+	_, err = cs.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{Name: state.ZarfNamespaceName},
+	}, metav1.CreateOptions{})
+	require.NoError(t, err)
+	require.NoError(t, c.SaveState(ctx, &state.State{}))
+
+	_, err = Deploy(ctx, pkgLayout, DeployOptions{Cluster: c})
```
brandtkeller, line 164: This isn'y proving the supplied cluster was used though correct? Maybe we improve the test fidelity with some observable interaction?

brandtkeller (changes requested):
Minor nit on testing - otherwise just digging a little deeper. Looks great on the surface and appreciate the support.

PR #5261, opened 2026-08-21
refactor!: introduce non-api-versioned type for actions

AustinAbro321:
## Description

Introduces a non api versioned type for actions. This way `actions.Run` can be an is callable separate of any API versions. 

Right now the execution type is internal. I think it could potentially change in the future if there are community requests for it (people want to preform logic on Zarf packages separate of packager functions), but it should stay internal at least until v1beta1 is released since users can just use v1alpha1 packages for now and this type will change as we have it hold other generic types (similar to actions type introduced here). 

I considered a bridge for `actions.Run`, however since until the recent #5212 it wasn't possible to call I decided against.

## Breaking changes

- `Actions.Run` takes it's own generic type, and has an options struct
- `(c *Cluster) GetInstalledChartsForComponent` accepts a component name rather than a component
- `utils.GetOSShell(shellPref Shell)` defines it's own shell type rather than using the versioned type.

## Related Issue

Relates to #5166

## Checklist before merging


src/pkg/packager/actions/actions.go
```diff
@@ -29,22 +29,31 @@ import (
...
+type RunOptions struct {
+	BasePath       string
+	Defaults       Config
+	VariableConfig *variables.VariableConfig
+	Values         value.Values
+	StateAccess    template.StateAccess
+}
+
+// Run executes every action in list in order.
+func Run(ctx context.Context, list ActionList, opts RunOptions) error {
```
a1994sc, line 42: I think that overall this is a great change as it decouples the actions logic from the alpha and beta schemas, and leaves the space open for other libraries to consume the actions.

a1994sc (commented):
I know it is still work-in progress, but I think this is a very good change overall.

src/pkg/packager/actions/actions.go
```diff
@@ -29,22 +29,31 @@ import (
 	"github.com/zarf-dev/zarf/src/pkg/wait"
 )
 
-// Run runs all provided actions.
-func Run(ctx context.Context, basePath string, defaultCfg v1alpha1.ZarfComponentActionDefaults, actions []v1alpha1.ZarfComponentAction, variableConfig *variables.VariableConfig, values value.Values, stateAccess template.StateAccess) error {
-	if variableConfig == nil {
-		variableConfig = ptmpl.GetZarfVariableConfig(ctx, false)
+// RunOptions supplies execution context for an action list
+type RunOptions struct {
```
brandtkeller, line 33: nit: not blocking but I don't know that these are all genuinely optional - but rather based on some precondition that we should document.

src/pkg/packager/actions/types.go
```diff
@@ -0,0 +1,104 @@
...
+
+// NetworkWait describes a network condition.
+type NetworkWait struct {
+	Protocol string
+	Address  string
+	Code     int
+}
+
+// Shell is the preferred command shell for each supported OS.
+type Shell struct {
```
brandtkeller, line 100: observation: this creates three layers (schema, action runtime, and process execution) - if we dpn't mind aliasing an exposing a lower one to a higher then that could be one option of attempting to mitigate drift between the three. Otherwise I don't think a change is required.

src/internal/packager/execution/execution.go
```diff
@@ -0,0 +1,297 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2021-Present The Zarf Authors
+
+// Package execution translates package schemas into packager runtime models.
```
brandtkeller, line 4: You might almost say it "projects" versioned definitions _into_ an execution representation. `projection` might be another naming alternative (I don't know that it is any better though)

brandtkeller (dismissed):
No blocking comments - I read the description - reviewed - and re-read the description. Framing this around actions confused me for a moment and then I saw the TODO for the other types that need generic representations. 

So a high-level strategy here might be described as a "schema-nuetral packager runtime" perhaps?

Regardless I did not find anything critical - lgtm

src/internal/packager/execution/execution.go
```diff
@@ -0,0 +1,297 @@
...
+
+	"github.com/zarf-dev/zarf/src/api"
+	"github.com/zarf-dev/zarf/src/api/v1alpha1"
+	"github.com/zarf-dev/zarf/src/api/v1beta1"
+	"github.com/zarf-dev/zarf/src/pkg/packager/actions"
+)
+
+// Component contains the resources needed to deploy a component.
+// TODO: Convert charts, images, repos, and health checks into generic types.
+type Component struct {
```
soltysh, line 18: I'm not sure that introducing a type is the right way to abstract v1alpha1 and v1beta1 types. Especially that you end up moving data a lot with this kind of approach. Wouldn't it be simpler to introduce a `Component` interface with a set of methods (like the ones defined below) and the actual implementations would reside inside a particular type. This would allow us to easily ensure both types from v1alpha1 and v1beta1 are fully implementing the interface, and catch drift. The additional benefit will be also the proximity of the implementation with the actual types. Currently you have to jump between appropriate type files and this one to ensure parity.

src/pkg/packager/actions/types.go
soltysh: I share a similar concern, see my other comment regarding replacing types with an interface. I doubt that's possible here. Type aliasing, could be what would save us from unnecessarily copying the structs, especially that we know they are all identical.

src/internal/packager/runtime/runtime.go
```diff
@@ -0,0 +1,297 @@
...
+
+	"github.com/zarf-dev/zarf/src/api"
+	"github.com/zarf-dev/zarf/src/api/v1alpha1"
+	"github.com/zarf-dev/zarf/src/api/v1beta1"
+	"github.com/zarf-dev/zarf/src/pkg/packager/actions"
+)
+
+// Component contains the resources needed to deploy a component.
+// TODO: Convert charts, images, repos, and health checks into generic types.
+type Component struct {
```
AustinAbro321: I think an interface becomes problematic because there are just so many types, we would have to introduce a ton intermediary types and getters. Actions alone, for instance, involves seven different structs. Additionally, it would introduce a lot of conversion logic on the intermediary types on the v1alpha1 and v1beta1 types themselves, which should just be schema representations. Another problem is that the interface introduces a lowest common denominator. V1beta1 has to carry around functions for data injections (removed) and v1alpha1 has to carry around functions for new data introduced. 

Still, I do see your point about yet another type and conversions to yet another type being introduced. I'm going to evaluate if I can find an architecture where we have an internal working runtime type rather than using `AsV1alpha1()` or `AsV1beta1()` everywhere. Maybe this involves changing the generic type to be easier to work with, as right now it's really targeted towards loseless conversions. Either way I wouldn't expose a runtime type, functions would always accept a `PackageLayout` or `PackageDefinition`. 

Main goal here is to make packages like actions have a non-api-versioned type, and have the package be easy to work with and update with a new schema is introduced. Will do more research and report back.

AustinAbro321:
Closing this in favor of https://github.com/zarf-dev/zarf/pull/5345

PR #5265, opened 2026-08-23
fix: pass remote options when loading OCI init packages

### LINKED ISSUE #5259: zarf init ignores --plain-http when loading an OCI init package
### Environment
Device and OS: Linux/amd64
App version: v0.80.0 and v0.84.0
Kubernetes distro being used: N/A (the command fails before deployment)
Other: Hauler serving an OCI registry over plain HTTP

### Steps to reproduce
1. Publish an init package to a plain-HTTP OCI registry.
2. Run:

```sh
zarf --plain-http init \
  oci://localhost:5001/example/init:v0.84.0 \
  --registry-url localhost:5001 \
  --registry-push-username zarf-push \
  --registry-push-password unused
```

### Expected result
Zarf loads the init package over HTTP and continues with initialization.

### Actual Result
The init package loader still connects over HTTPS:

```text
unable to load package: could not find package oci://localhost:5001/example/init:v0.84.0 with architecture amd64:
Get "https://localhost:5001/v2/example/init/manifests/v0.84.0":
http: server gave HTTP response to HTTPS client
```

The same registry works with `zarf package pull ... --plain-http`.

The error above is reproducible with both v0.80.0 and v0.84.0.

### Severity/Priority
Medium. It blocks using an init package directly from a plain-HTTP OCI registry. Pulling the package to disk first works, but adds an unnecessary step to bootstrap workflows.

### Additional Context
`initOptions.run` builds a `packager.LoadOptions` value without passing `RemoteOptions`:

https://github.com/zarf-dev/zarf/blob/v0.84.0/src/cmd/initialize.go#L222-L229

The package pull path does pass `RemoteOptions: defaultRemoteOptions()`:

https://github.com/zarf-dev/zarf/blob/v0.84.0/src/cmd/package.go#L1721-L1729

Adding the same field to the init load options should make `--plain-http` and `--insecure-skip-tls-verify` reach the OCI package loader. I can send a PR with the fix and regression coverage.

fabian1heinrich:
## Description

Passes the configured remote options when loading an OCI init package, so `--plain-http` is respected. Adds a test covering this case.

## Related Issue

Fixes #5259 

## Checklist before merging

### DIFF
--- a/src/test/e2e/11_oci_pull_inspect_test.go
+++ b/src/test/e2e/11_oci_pull_inspect_test.go
@@ -86,7 +86,16 @@ func (suite *PullInspectTestSuite) Test_0_Pull() {
 		"inspect digest should return the same value for the tarball and the OCI source it was published from")
 }
 
-func (suite *PullInspectTestSuite) Test_1_Remote_Inspect() {
+func (suite *PullInspectTestSuite) Test_1_Init_Loads_OCI_Over_Plain_HTTP() {
+	suite.T().Log("E2E: Init package load oci:// over plain HTTP")
+
+	ref := fmt.Sprintf("oci://%s/simple-package:0.0.1", suite.Reference.String())
+	_, stdErr, err := e2e.Zarf(suite.T(), "init", ref, "--plain-http")
+	suite.Error(err, stdErr)
+	suite.Contains(stdErr, `zarf init can only deploy packages of kind "ZarfInitConfig"`)
+}
+
+func (suite *PullInspectTestSuite) Test_2_Remote_Inspect() {
 	suite.T().Log("E2E: Package Inspect oci://")
 
 	// Test inspect w/ bad ref.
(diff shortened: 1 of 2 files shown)


src/test/e2e/11_oci_pull_inspect_test.go
```diff
@@ -86,7 +86,16 @@ func (suite *PullInspectTestSuite) Test_0_Pull() {
 		"inspect digest should return the same value for the tarball and the OCI source it was published from")
 }
 
-func (suite *PullInspectTestSuite) Test_1_Remote_Inspect() {
+func (suite *PullInspectTestSuite) Test_1_Init_Loads_OCI_Over_Plain_HTTP() {
```
AustinAbro321: nit: while I appreciate the regression test, I would like to be cautious not to add too much to our e2e suite. I would accept this change without a test.

brandtkeller (approved):
lgtm - thanks for the quick response

AustinAbro321 (approved):
lgtm

Issue #5267, opened 2026-08-24
zarf package deploy can cause progress to go backwards


nicleary:
### Environment
Device and OS: 2024 MacBook Pro 16", macOS 26.5.2
App version: 0.83.0
Kubernetes distro being used: AWS EKS v1.35
Other: N/A

### Steps to reproduce
1. Deploy a package via an unstable network connection to an AWS EKS cluster
2. Notice as progress for pushing an image initially starts as a low percentage, and then goes backwards
3. After the third iteration of `pushing image name=X` the push will be successful

### Expected result
Images push successfully to the cluster and progress report is accurate
### Actual Result
The image push will seemingly fail twice, and on the third time of `pushing image name=X` the image push will succeed. Progress indicator will still be broken however.

### Visual Proof (screenshots, videos, text, etc)

Example logs

```
2026-08-21 20:48:58 INF pushing image name=REDACTED
2026-08-21 20:49:03 INF image push in progress name=REDACTED complete=6.6% remaining=90.88 MBs
2026-08-21 20:49:08 INF image push in progress name=REDACTED complete=6.6% remaining=90.88 MBs
2026-08-21 20:49:13 INF image push in progress name=REDACTED complete=6.6% remaining=90.88 MBs
2026-08-21 20:49:18 INF image push in progress name=REDACTED complete=6.6% remaining=90.88 MBs
2026-08-21 20:49:23 INF image push in progress name=REDACTED complete=6.6% remaining=90.88 MBs
2026-08-21 20:49:28 INF image push in progress name=REDACTED complete=6.6% remaining=90.88 MBs
2026-08-21 20:49:33 INF image push in progress name=REDACTED complete=6.6% remaining=90.88 MBs
2026-08-21 20:49:38 INF image push in progress name=REDACTED complete=6.6% remaining=90.88 MBs
2026-08-21 20:49:46 INF image push in progress name=REDACTED complete=0.0% remaining=97.32 MBs
2026-08-21 20:49:51 INF image push in progress name=REDACTED complete=0.0% remaining=97.32 MBs
2026-08-21 20:49:52 INF pushing image name=REDACTED
2026-08-21 20:49:57 INF image push in progress name=REDACTED complete=0.0% remaining=97.32 MBs
2026-08-21 20:50:02 INF image push in progress name=REDACTED complete=0.0% remaining=97.32 MBs
2026-08-21 20:50:07 INF image push in progress name=REDACTED complete=0.0% remaining=97.32 MBs
[...]

PR #5268, opened 2026-08-24
feat: warn on image push failure

### LINKED ISSUE #4751: Show warning logs on image push retries
Currently when an image push fails Zarf retries push by default. On the last retry we set OCI concurrency to one. We should warn instead of debug log when these retries happen.

Additionally, if we're giving the users these warnings, we should likely use `retry.LastErrorOnly(true)` so we're not spamming the user with an error they've already seen.

AustinAbro321:
## Description

This implements warnings on image push failures as they happen, then only shows the last error on failure. 

Additionally, this sets `responseHeaderTimeout` to a minute after the first failure since we've heard of failures #5267

(I'm also sneaking in a makefile change because running `go mod vendor` every time takes too long even when there are no changes)

## Related Issue

Fixes #4751 

## Checklist before merging

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,10 +81,16 @@ destroy: ## Run `zarf destroy` on the current cluster
 	rm -fr build
 
 # Note: the path to the main.go file is not used due to https://github.com/golang/go/issues/51831#issuecomment-1074188363
+.PHONY: vendor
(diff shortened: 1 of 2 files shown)


brandtkeller (approved):
lgtm

PR #5269, opened 2026-08-24
refactor(state): move agent information into a sub-struct

### LINKED ISSUE #5190: Refactor agent object in Zarf state
The Zarf agent fields in ZarfState should be moved onto a `AgentInfo` object, instead of being direct fields on the struct. This would match the `RegistryInfo` and `GitServerInfo` fields. This wasn't done as the only agent field used to be the AgentTLS PKI, but now that we have other fields we should implement this.

```go
	// PKI certificate information for the agent pods Zarf manages
	AgentTLS pki.GeneratedPKI `json:"agentTLS"`
	// AgentTLSUserProvided indicates whether the agent TLS certs were provided by the user rather than auto-generated
	AgentTLSUserProvided bool `json:"agentTLSUserProvided,omitempty"`
	// AgentMutationPolicy controls the conditions required for the agent to mutate resources
	AgentMutationPolicy MutationPolicy `json:"agentMutationPolicy"`
```

Should be implemented in a backward compat way, so that anyone relying on these fields or older versions of Zarf don't break

AustinAbro321:
## Description

move agent information into a sub-struct with backwards compatiability

## Related Issue

Fixes #5190

## Checklist before merging

### DIFF
--- a/src/pkg/state/state.go
+++ b/src/pkg/state/state.go
@@ -190,15 +190,14 @@ type AgentInfo struct {
 	MutationPolicy MutationPolicy `json:"mutationPolicy"`
 }
 
-// IsConfigured returns true when the agent TLS certificate is configured.
+// IsConfigured returns true when a complete agent TLS bundle is configured.
 func (ai AgentInfo) IsConfigured() bool {
-	return len(ai.TLS.Cert) > 0
+	return len(ai.TLS.CA) > 0 && len(ai.TLS.Cert) > 0 && len(ai.TLS.Key) > 0
 }
 
 // AgentIsConfigured returns true when Zarf has agent TLS configured.
 //
-// Deprecated: Use AgentInfo.IsConfigured instead. State is reconciled here to
-// support legacy state fields during the transition.
+// Deprecated: Use AgentInfo.IsConfigured instead.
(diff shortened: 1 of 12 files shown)


src/pkg/state/state.go
```diff
@@ -178,9 +180,69 @@ type State struct {
...
+	}
+}
+
+func (s *State) syncLegacyAgentFields() {
+	s.AgentTLS = s.AgentInfo.TLS
+	s.AgentTLSUserProvided = s.AgentInfo.TLSUserProvided
+	s.AgentMutationPolicy = s.AgentInfo.MutationPolicy
+}
+
+func (ai AgentInfo) isZero() bool {
```
soltysh, line 243: My only suggestion would be maybe `isConfigured`, but since that's a private method I don't think I care that much.

soltysh (dismissed):
lgtm

src/pkg/state/state.go
AustinAbro321: There is already an `IsConfigured` function, however it looks like we were setting mutation policy even when the user wasn't creating an agent. I fixed this in this PR, but for existing clusters `IsConfigured` only checks if the TLS cert is there

soltysh (approved):
lgtm

PR #5273, opened 2026-08-25
feat: add oci source auto-prefixing

### LINKED ISSUE #5271: Auto-detect OCI packages rather than requiring explicit prefix

OCI package references require an explicit `oci://` prefix. Users can already identify a registry reference from its hostname and repository path, so the prefix adds friction to common commands such as `zarf package deploy` and `zarf package pull`, especially when other ecosystem tools commonly "auto-detect" OCI.

- **Given** an OCI package reference shaped like `ghcr.io/org/package:v1` or `localhost:5000/package:v1`
- **When** a user passes it to a command that accepts a package source
- **Then** Zarf treats it as an OCI source without requiring `oci://`

Explicit `oci://` references must continue to work. Local tarballs and bare deployed package names must keep their existing behavior.

Users can keep entering `oci://`. Zarf could also try to resolve every unknown source as OCI, but that would add network calls and conflict with unprefixed deployed package names.

mjnagel:
## Description

Adds a `normalizeOCISource` function that auto-prefixes known OCI packages with the explicit `oci://` prefix. This seemed like the least intrusive path to handle the linked issue.

## Related Issue

Fixes https://github.com/zarf-dev/zarf/issues/5271

## Checklist before merging

### DIFF
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -1548,7 +1548,7 @@ func (o *packagePublishOptions) run(cmd *cobra.Command, args []string) error {
 	v := getViper()
 	isSkeletonPackage := helpers.IsDir(packageSource)
 	if !isSkeletonPackage {
-		packageSource = packager.NormalizeOCISource(packageSource)
+		packageSource = zoci.NormalizeOCISource(packageSource)
 	}
 
 	if !helpers.IsOCIURL(args[1]) {
@@ -1805,7 +1805,7 @@ func newPackageSignCommand(v *viper.Viper) *cobra.Command {
 func (o *packageSignOptions) run(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
 	l := logger.From(ctx)
-	packageSource := packager.NormalizeOCISource(args[0])
+	packageSource := zoci.NormalizeOCISource(args[0])
 
 	if !o.keyless && o.signingKeyPath == "" {
 		return errors.New("--signing-key is required (or pass --keyless for Sigstore keyless flow)")
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -1546,6 +1546,10 @@ func (o *packagePublishOptions) run(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
 	l := logger.From(ctx)
 	v := getViper()
+	isSkeletonPackage := helpers.IsDir(packageSource)
+	if !isSkeletonPackage {
+		packageSource = packager.NormalizeOCISource(packageSource)
+	}
 
 	if !helpers.IsOCIURL(args[1]) {
 		return errors.New("registry must be prefixed with 'oci://'")
@@ -1568,7 +1572,7 @@ func (o *packagePublishOptions) run(cmd *cobra.Command, args []string) error {
 	}
 
 	// Skeleton package - call PublishSkeleton
-	if helpers.IsDir(packageSource) {
+	if isSkeletonPackage {
 		skeletonOpts := packager.PublishSkeletonOptions{
 			OCIConcurrency:       o.ociConcurrency,
 			SigningKeyPath:       o.signingKeyPath,
@@ -1801,7 +1805,7 @@ func newPackageSignCommand(v *viper.Viper) *cobra.Command {
 func (o *packageSignOptions) run(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
 	l := logger.From(ctx)
-	packageSource := args[0]
+	packageSource := packager.NormalizeOCISource(args[0])
 
 	if !o.keyless && o.signingKeyPath == "" {
 		return errors.New("--signing-key is required (or pass --keyless for Sigstore keyless flow)")
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -1543,6 +1543,7 @@ func newPackagePublishCommand(v *viper.Viper) *cobra.Command {
 
 func (o *packagePublishOptions) run(cmd *cobra.Command, args []string) error {
 	packageSource := args[0]
+	packageDestination := zoci.NormalizeOCISource(args[1])
 	ctx := cmd.Context()
 	l := logger.From(ctx)
 	v := getViper()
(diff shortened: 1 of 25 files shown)


src/pkg/packager/load_test.go
```diff
@@ -219,6 +219,59 @@ func TestIdentifySource(t *testing.T) {
...
+		},
+		{
+			name:     "local tarball",
+			source:   "packages/init.tar.zst",
+			expected: "packages/init.tar.zst",
+		},
+		{
+			name:     "ambiguous path",
+			source:   "packages/init:v0.84.0",
+			expected: "packages/init:v0.84.0",
```
Racer159: small thought on logic here - it may be worth interpreting this convention as OCI - later in identifySource a filename like this would be rejected anyway (https://github.com/zarf-dev/zarf/blob/2ce2cc141a273ce38c49495888db215e456c0fab/src/pkg/packager/load.go#L191) and there is a possibility that `packages` is a valid local machine name to pull the packages from.  It may be good to use the OCI-ish `registry.ParseReference` function from ORAS to help here as well since that would tell you if it was possible to interpret the string as a valid OCI reference.

mjnagel: Yeah I think this makes sense - swapped to use the ORAS function here. It simplifies some of code on the Zarf side and ultimately like you called out shouldn't change net result with something like `packages/init:v0.84.0` (or maybe makes it work if it was valid).

src/pkg/packager/load.go
```diff
@@ -206,11 +208,27 @@ func identifySource(src string) (string, error) {
 	return "", fmt.Errorf("unknown source %s. Did you forget the scheme (e.g. (oci://) or file extension (e.g. .tar.zst)?", src)
 }
 
+// normalizeOCISource adds the OCI scheme to valid registry references.
+func normalizeOCISource(source string) string {
```
brandtkeller: Love the simplicity - I think we might need to handle regressions from the CLI perspective to land this in a releasable fashion. 

From the CLI:

- `zarf package sign` which is signing from an OCI reference no longer re-publishes the signed-package to the registry. Rather now the non-prefix source is interpreted as a filepath for output.
- `zarf package publish` from OCI to OCI now cannot discern that the input is an OCI reference when the oci scheme is not present. 

Didn't find other scenarios yet but I do appreciate this change.

brandtkeller (changes requested):
SDK perspective looks good - dropped a comment on CLI implications. 

May need to include language changes at the CLI level if the oci scheme is indeed not required anymore.

Thank you for pushing on this as an improvement in the UX.

src/pkg/packager/load.go
mjnagel: Ah good calls - I probably didn't do sufficient testing e2e on all the CLI commands. Fixes should be pushed in https://github.com/zarf-dev/zarf/pull/5273/changes/4ed65b98ca8d1353d94f4817f8f620f594a1beb5 along with a docs update. Added e2e coverage of these cases as well.

src/pkg/packager/load.go
```diff
@@ -206,11 +208,27 @@ func identifySource(src string) (string, error) {
 	return "", fmt.Errorf("unknown source %s. Did you forget the scheme (e.g. (oci://) or file extension (e.g. .tar.zst)?", src)
 }
 
+// NormalizeOCISource adds the OCI scheme to valid registry references.
+func NormalizeOCISource(source string) string {
```
brandtkeller: This function moved to public to support use more broadly - I'm curious whether it may be a better fit for `zoci`?

brandtkeller (changes requested):
Minor shift in function location otherwise looks great.

src/pkg/packager/load.go
mjnagel: Yeah I think that makes a lot of sense given the usage across places now.

src/cmd/package.go
```diff
@@ -1546,6 +1546,10 @@ func (o *packagePublishOptions) run(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
 	l := logger.From(ctx)
 	v := getViper()
+	isSkeletonPackage := helpers.IsDir(packageSource)
+	if !isSkeletonPackage {
+		packageSource = zoci.NormalizeOCISource(packageSource)
+	}
 
 	if !helpers.IsOCIURL(args[1]) {
```
brandtkeller: publish only supports targeting an oci registry - so I imagine we can make the prefix on the destination here optional as well.

src/test/e2e/50_oci_publish_deploy_test.go
```diff
@@ -80,6 +80,22 @@ func (suite *PublishDeploySuiteTestSuite) Test_0_Publish() {
...
 	stdOut, stdErr, err = e2e.Zarf(suite.T(), "package", "inspect", "definition", "oci://"+ref+"/helm-charts:0.0.1", "--plain-http")
 	suite.NoError(err, stdOut, stdErr)
+
+	// Copy an OCI package without its optional oci:// source prefix. The source tag must be preserved.
+	stdOut, stdErr, err = e2e.Zarf(suite.T(), "package", "publish", ref+"/helm-charts:0.0.1", "oci://"+ref+"/implicit-source", "--plain-http")
+	suite.NoError(err, stdOut, stdErr)
+	stdOut, stdErr, err = e2e.Zarf(suite.T(), "package", "inspect", "definition", "oci://"+ref+"/implicit-source/helm-charts:0.0.1", "--plain-http")
+	suite.NoError(err, stdOut, stdErr)
+
+	// Sign an OCI package without its optional oci:// source prefix. With no output flag,
```
brandtkeller: This signing test feels more attuned to signing and less to publish. More so as I begin to think on how to optimize the signing workflow given a published package. 

Perhaps we move it to the signing e2e test?

brandtkeller (approved):
lgtm - the requirement for the `oci://` scheme prefix as far as I understand is an internal one. I could see a future where we parse into a typed source in order to make downstream decisions on locality and destination but this provided a lot of great insight.

src/pkg/zoci/utils.go
```diff
@@ -14,6 +14,21 @@ import (
 	"oras.land/oras-go/v2/registry"
 )
 
+// NormalizeOCISource adds the OCI scheme to valid registry references.
+func NormalizeOCISource(source string) string {
+	if helpers.IsOCIURL(source) || strings.Contains(source, "://") {
+		return source
+	}
+	if strings.HasSuffix(source, ".tar.zst") || strings.HasSuffix(source, ".tar") || strings.Contains(source, ".part000") {
```
soltysh, line 22: Suggestion, not blocking this particular PR. Theoretically `ghcr.io/some-org/pkg:release.tar.zst` is a valid OCI definition. Should we maybe do an `os.Stat` on the file to clearly distinguish between local file and actual OCI URL?

soltysh (approved):
lgtm

src/pkg/zoci/utils.go
mjnagel: Yeah... 🤖s did flag that for me locally. It definitely felt edge case enough to defer for now imo, but I think there's probably some value in looking at and accounting for later.

PR #5280, opened 2026-08-25
fix: support platform flag entries for crane copy

### LINKED ISSUE #5272: Zarf registry copy fails on OCI attestation manifests, blocking sealed-secrets-controller promotion
### Environment
Device and OS: GitLab CI runner (group runner id `11`), Linux, linux/amd64
App version: zarf v0.82.0, v0.84.0

### Steps to reproduce

```
zarf tools registry copy --platform linux/amd64 \
     docker.io/bitnami/sealed-secrets-controller:0.39.1 \
     registry.example.com/myorg/mypackage/bitnami/sealed-secrets-controller:0.39.1
```

### Expected result

 The image copies successfully to the target registry

### Actual Result

```
ERR PUT https://registry.example.com/v2/myorg/mypackage/bitnami/sealed-secrets-controller/manifests/sha256:407e36540a358a0da303e
   4358aaedaf8419b807b2f17d74609c155ecb0e5191e:
       MANIFEST_BLOB_UNKNOWN: blob unknown to registry; sha256:2904bb68ca67bbccf0f39fc435abbb921171441374f1e7269a19c007c0ed61bd
```

### Additional Context

- crane copy --platform linux/amd64 <src> <dst>  worked well
- zarf silently ignores the --platform  flag, its source hardcodes  crane.WithPlatform(nil) (possible root cause)
- zarf copy for sealed secrets image version 0.38.4 worked well, because the attestation manifest was not present

brandtkeller:
## Description

Recognizing that this will be replaced by an oras-go implementation in the future - this resolves the current issue which should help influence parity for future replacement. 

I was not able to reproduce the error - but looking strictly at correctness I believe this helps.

Setting this to nil explicitly will never honor individual platforms. Whereas `nil` is the default option passed to crane still resolves the default behavior.

## Related Issue

Fixes #5272

## Checklist before merging

### DIFF
--- a/src/cmd/crane.go
+++ b/src/cmd/crane.go
@@ -85,7 +85,7 @@ func newRegistryCommand() *cobra.Command {
 	cmd.AddCommand(newRegistryPruneCommand())
 	cmd.AddCommand(newRegistryLoginCommand())
 	cmd.AddCommand(newRegistryLogoutCommand())
-	cmd.AddCommand(newRegistryCopyCommand(&craneOptions))
+	cmd.AddCommand(craneCmd.NewCmdCopy(&craneOptions))
 	cmd.AddCommand(newRegistryCatalogCommand(&craneOptions))
 
 	// TODO(soltysh): consider splitting craneOptions to be per command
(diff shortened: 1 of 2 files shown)


soltysh (approved):
lgtm

PR #5287, opened 2026-08-26
feat(helm): support allowable errors for helm chart retries

### LINKED ISSUE #5286: Restore helm chart retries for specific error sets

 Helm releases can fail when a newly installed, fail-closed admission webhook is not yet reachable. This is a distinct error where the previously-existing retry logic could be beneficial.                                                                                            

 - Given a Helm release fails with an explicitly retryable admission-webhook transport error                                                                                            
 - When --retries permits another attempt                                                                                                                                               
 - Then Zarf retries the release with bounded backoff and the existing total Helm timeout.                                                                           

Use the existing deployment-wide `--retries` budget for component charts and manifest-generated Helm releases. 

 - Retry all Helm failures: Previously implemented and removed because it retried deterministic failures such as invalid resources, RBAC errors, hooks, and readiness failures. 
 - Per-chart retry configuration: Adds unnecessary package schema and configuration surface.                                                                                            
 - Broad API retries: 429, ServiceUnavailable, timeouts, and generic internal errors need separate safety criteria and tests. 
 - Chart-specific webhook waits: Couples Zarf to individual controllers.                                                                                                                

 ### Additional context

 Update --retries help text to include transient Helm admission-webhook failures.

brandtkeller:
## Description

Adds support for a retry on helm chart installations when the failure isn't deterministic. There may be many but the first recorded one is the presence of an admission/mutation webhook and the inability to send requests as the endpoint may not be entirely reconciled by kubernetes networking when attempting to deploy resources that then call that webhook.

I looked into whether the 500 error returned might include a `Retry-After` header that we look for on 429 responses but it was not there (and probably rightfully so). 

## Related Issue

Fixes #5286

## Checklist before merging


src/internal/packager/helm/chart.go
```diff
@@ -194,6 +203,74 @@ func InstallOrUpgradeChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, ch
...
+	)
+}
+
+// isRetryableHelmChartError allows a retry when a just-installed, fail-closed admission webhook is not yet reachable.
+func isRetryableHelmChartError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if wrapped, ok := err.(interface{ Unwrap() []error }); ok {
```
soltysh, line 247: There are several problems in this entire block, and the next function:
1. The rules for errors are such that you'd need to match both signatures `Unwrap() []error` and `Unwrap() error`, see https://pkg.go.dev/errors#pkg-overview. I've tried with using `fmt.Errorf("...%w...", ...err...)` and that matched only `Unwrap() error`. 
2. It's easier to just use [AsType](https://pkg.go.dev/errors#AsType) instead, which does the entire work for you. Here it will simplify this and the next method to look like so:
```golang
apiStatus, ok = errors.AsType[*apierrors.APIStatus](err)
if !ok {
	return false
}
status := apiStatus.Status()
return status.Reason == metav1.StatusReasonInternalError && strings.Contains(status.Message, "failed calling webhook")
```

src/internal/packager/helm/chart.go
```diff
@@ -194,6 +203,74 @@ func InstallOrUpgradeChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, ch
...
+	return isRetryableAdmissionWebhookLeaf(err)
+}
+
+func isRetryableAdmissionWebhookLeaf(err error) bool {
+	apiStatus, ok := err.(apierrors.APIStatus)
+	if !ok {
+		return false
+	}
+	status := apiStatus.Status()
+	return status.Reason == metav1.StatusReasonInternalError && strings.Contains(status.Message, "failed calling webhook")
```
soltysh, line 271: I'm somehow worried about these string matching errors, they are not reliable enough, imo. Looking deeper, how about instead of matching apistatus trying to match [ErrCallingWebhook](https://github.com/kubernetes/kubernetes/blob/c429f2696cf51cc42b7dd155c08abc55b7388a2c/staging/src/k8s.io/apiserver/pkg/util/webhook/error.go#L28) instead? Or both?

src/internal/packager/helm/chart.go
```diff
@@ -105,32 +112,34 @@ func InstallOrUpgradeChart(ctx context.Context, zarfChart v1alpha1.ZarfChart, ch
...
+		releases, histErr := histClient.Run(zarfChart.ReleaseName)
 
-	l.Debug("checking for existing helm deployment")
+		l.Debug("checking for existing helm deployment")
 
-	var newRelease release.Releaser
-	if errors.Is(histErr, driver.ErrReleaseNotFound) {
-		// No prior release, try to install it.
-		l.Info("performing Helm install", "chart", zarfChart.Name)
+		if errors.Is(histErr, driver.ErrReleaseNotFound) {
```
soltysh, line 126: I'm not a helm expert, but codex suggest that a failed, retried releases might not return an error, but rather a failed release instead, which means this code will not go the install path, but rather upgrade one. This can potentially skip some install-related steps. Should we also look at the release status and check if it's reporting [StatusFailed](https://github.com/helm/helm/blob/d62bee21c2dd3ffcd647ae3718288b3760b7d421/pkg/release/common/status.go#L33)?

src/internal/packager/helm/chart.go
brandtkeller: Agree on covering both `Unwrap() []error` and `Unwrap() error` - I added an additional test case for coverage with one nested through `fmt.Errorf("%w")` - otherwise I think we have coverage on the other cases. 

My understanding was that errors.As / errors.AsType returns the first matching node and the policy I was looking to implement was more closely aligned with "all leaves must be allowed". AsType would report a match for a `webhookFailure` that also happens to be joined with another terminal failure. 

Certainly open to thoughts - my goal was to ensure we're conservative with what is allowed and then loosen/add as required.

src/internal/packager/helm/chart.go
brandtkeller: I was digging through this some more. If I understand correctly `ErrCallingWebhook` _would_ be better if it survived to the client, but it looks like the api-server might wrap that internal type with `apierrors.NewInternalError` before serializing the response. Are you aware of a typed signal for this path I might have missed?

src/internal/packager/helm/chart.go
brandtkeller: I think this is a tricky balance - if we fail during `install` then the next retry will attempt upgrade, skipping the install specific steps. But if we implement a replace for `install` on retry then we could retry already executed install steps. 

GIven we're narrowing when a retry can occur (webhook failed) - my inclination is that the current semantics should generally be okay. If we were retrying chart installs regardless of error type then we might have to think on this further.

src/internal/packager/helm/chart.go
soltysh: You're right, it is only wrapped under `apierrors.NewInternalError ` sadly.

src/internal/packager/helm/chart.go
soltysh: > Certainly open to thoughts - my goal was to ensure we're conservative with what is allowed and then loosen/add as required.

I believe here, the same rule applies you've mentioned in the other comment. Whenever errors are wrapped like [NewInternalError](https://github.com/kubernetes/kubernetes/blob/23d989a4ba78d4bffec52a573cc70b672c66e991/staging/src/k8s.io/apimachinery/pkg/api/errors/errors.go#L387) the double wrapping will not happen, since the internal error will be printed as a string in the details of the message. So I believe, here, it'll be simpler for us to just find the first and only `errors.APIStatus`. It will also make the code simpler to read 😉

soltysh: This can be done in a followup.

soltysh (approved):
lgtm

Issue #5281, opened 2026-08-25
Deprecate and remove signing within package creation and package publishing


AustinAbro321:
### Is your feature request related to a problem? Please describe.
The flags `--signing-key` and `--signing-key-pass` should be deprecated for `zarf package create` and `zarf package publish`, then removed in v1.0. These commands do not have all the signing options and adding all the signing options would add complexity to the UX.  

By deprecating these flags, people should learn about the `zarf package sign` command. 

### Describe alternatives you've considered
Keep the flags, and add all of the keyless flags for signing. 

### Additional context
The main downside is that Zarf must un-archive and re-archive a package to sign it which can take a noticeable amount of time for packages that are huge (10-100GB+), but I think the cleaner user experience is worth it. Also this doesn't affect OCI packages, which can pull only the (small) files it needs to sign.

PR #5299, opened 2026-08-31
feat(sign): deprecate signing flags from CLI commands

brandtkeller:
## Description

Deprecation of the signing flags on create/publish. This takes the direction of removing the flags and requiring use of `zarf package sign` versus the alternative of adding all signing options (for which there are many) to the commands and further obstructing the UX. 

## Related Issue

relates to #5281

## Checklist before merging


brandtkeller:
I want to think about this more before we execute. It feels wrong to retain the SDK options for signing in the packager workflows but also wrong to potentially allow creation of an unsigned artifact when compared to the current workflow which allows for direct creation of a signed artifact.

PR #5301, opened 2026-09-01
feat(lint): add --all-variants flag for lint

### LINKED ISSUE #4598: zarf dev lint fails if a flavor is not specified for a flavored package

chaospuppy:
## Description
This PR adds an `--all-variants` flag to the `zarf dev lint` command that will result in all variants of a zarf package being linted, even if there are multiple flavors present in the package definition.  Previously, it was required that users specify a `--flavor` flag in order to pass validation.  This is done to make linting zarf package in CI or in situations where you would otherwise not want to/could not know which flavor you wanted to lint specifically, and wished to validate the entire package.

## Related Issue

Fixes #4598 

## Checklist before merging

### DIFF
--- a/site/src/content/docs/commands/zarf_dev_lint.md
+++ b/site/src/content/docs/commands/zarf_dev_lint.md
@@ -21,6 +21,7 @@ zarf dev lint [ DIRECTORY ] [flags]
 ### Options
 
 ```
+      --all-variants         Lints all package components, regardless of flavor. NOTE: This will disable checks for component name uniqueness.
   -f, --flavor string        The flavor of components to include in the resulting package (i.e. have a matching or empty "only.flavor" key)
   -h, --help                 help for lint
       --set stringToString   Specify package templates to set on the command line (KEY=value) (default [])
--- a/src/cmd/dev.go
+++ b/src/cmd/dev.go
@@ -1143,8 +1143,9 @@ func (o *devGenerateConfigOptions) run(_ *cobra.Command, args []string) error {
 }
 
 type devLintOptions struct {
-	setPkgTmpl map[string]string
-	flavor     string
+	setPkgTmpl  map[string]string
+	flavor      string
+	allVariants bool
 }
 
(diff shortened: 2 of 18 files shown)


chaospuppy:
@AustinAbro321 not ready for review yet, just wanted to highlight that I am continuing the work for #4598 in this PR - my old branch was pretty far out of date and it made sense to start over.  I have this working for the v1alpha1 case, but still need to make changes to get it working for v1beta1.  I also implemented VariantDimensions as described [here](https://github.com/zarf-dev/zarf/pull/5020#discussion_r3500488859)

AustinAbro321 (changes requested):
The timing is a bit awkward since it's not yet released, but this should be added to the v1beta1 import flow as well since the loading for v1beta1 packages is functional and goes through the load.PakcageDefinition flow

AustinAbro321:
Ah missed your comment sorry @chaospuppy, if you don't mind putting in draft until it's ready. Variant dimensions look good

PR #5302, opened 2026-09-01
chore: update gitea chart to `12.7.0`

### LINKED ISSUE #5293: Upgrade Gitea component of the init package

Barring the renovate integration, maintaining gitea is part of the init package release artifacts. Gitea should be upgraded to the latest release and ensure it is compatible via new installations as well as upgrades. 

Upgrades can be tricky. current e2e testing should cover latest release then upgrade to proposed version. That confirms health and other e2e testing around gitops should ensure functionality. 

There is likely an intermediate phase where we need to test gitops workflows across the upgrade.

NAME-ASHWANIYADAV:
## Description

Updates the Gitea component of the init package:

- helm chart `12.3.0` -> `12.7.0`
- image `ghcr.io/go-gitea/gitea:1.24.6-rootless` -> `1.27.3-rootless` (chart appVersion is 1.27.0, 1.27.3 is the latest patch on that minor)
- removed the `redis-cluster` block from `gitea-values.yaml`. That subchart was replaced by valkey-cluster and no longer exists in 12.7.0, so the value was dead weight.
- updated the SBOM filenames in `06_create_sbom_test.go` (same substitution `hack/cots/update-gitea.sh` does)

One thing worth calling out for review: the chart and image have to move together here. The 12.7.0 init script switched from `environment-to-ini` to `gitea config edit-ini`, and that command doesn't exist in 1.24.x, so the new chart with the old image would fail on startup.

How I tested this:

- rendered the chart with the package values (template placeholders filled in by hand) against both 12.3.0 and 12.7.0 and diffed the output. Only expected differences: version labels, the image tag, and the chart's own init script change mentioned above.
- `zarf package create packages/gitea` builds cleanly, and the SBOM filenames inside the package match what the updated e2e test expects.
- since the issue specifically calls out that upgrades can be tricky, I ran one on a kind cluster: installed 12.3.0 with 1.24.6-rootless using the package values (sqlite backend), created a repo through the Gitea API, then ran `helm upgrade` to 12.7.0 / 1.27.3-rootless. The pod came up healthy, `/api/v1/version` reports 1.27.3, and the repo created before the upgrade survived the migration.

## Related Issue

Fixes #5293

## Checklist before merging

### DIFF
--- a/packages/gitea/gitea-values.yaml
+++ b/packages/gitea/gitea-values.yaml
@@ -71,9 +71,6 @@ serviceAccount:
 postgresql-ha:
   enabled: false
 
-redis-cluster:
-  enabled: false
-
 valkey-cluster:
   enabled: false
 
--- a/packages/gitea/zarf.yaml
+++ b/packages/gitea/zarf.yaml
@@ -86,7 +86,7 @@ components:
       - name: gitea
         releaseName: zarf-gitea
         url: oci://registry-1.docker.io/giteacharts/gitea
-        version: 12.3.0
+        version: 12.7.0
         namespace: zarf
         valuesFiles:
           - gitea-values.yaml
--- a/src/test/e2e/06_create_sbom_test.go
+++ b/src/test/e2e/06_create_sbom_test.go
@@ -68,8 +68,8 @@ func TestCreateSBOM(t *testing.T) {
 
 	// Test that we preserve the filepath
 	require.FileExists(t, filepath.Join(outSbomPath, "dos-games", "sbom-viewer-ghcr.io_zarf-dev_doom-game_0.0.1.html"))
-	require.FileExists(t, filepath.Join(outSbomPath, "init", "sbom-viewer-ghcr.io_go-gitea_gitea_1.24.6-rootless.html"))
-	require.FileExists(t, filepath.Join(outSbomPath, "init", "ghcr.io_go-gitea_gitea_1.24.6-rootless.json"))
+	require.FileExists(t, filepath.Join(outSbomPath, "init", "sbom-viewer-ghcr.io_go-gitea_gitea_1.27.3-rootless.html"))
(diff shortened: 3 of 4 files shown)


NAME-ASHWANIYADAV:
Hi @brandtkeller , could you please review this PR whenever you get a chance? Let me know if you need any changes.

a1994sc:
Changes seem reasonable to me, if the checks pass I am good to approve

a1994sc:
Also, a minor grammar thing, but I this relates to the linked issue, it does not fix it in the long term

brandtkeller:
> Also, a minor grammar thing, but I this relates to the linked issue, it does not fix it in the long term

fair - we have an existing renovate issue to track that work though so I'd consider the linked issue closed with this work.

a1994sc:
> > Also, a minor grammar thing, but I this relates to the linked issue, it does not fix it in the long term
> 
> fair - we have an existing renovate issue to track that work though so I'd consider the linked issue closed with this work.

I will defer to Brandt thought process here

brandtkeller (approved):
lgtm - I tested first with the static diff between the latest released init package and an init package built from this branch using the output of `zarf package inspect manifests` and see little in the way of drive besides some file renames and expected labels/annotations. 

The note made regarding the change in `environment-to-ini` transition to `gitea config edit-ini` is accurate and I think this warrants a release note indicating that this requires gitea images v1.26+.

I ran a fresh install with an `zarf tools update-creds git` command with success. 

I ran an upgrade from the latest release + the podinfo-flux package deployed -> upgraded gitea via this branches init package. 

I think Gitea prevents an upgrade-downgrade-upgrade test to preserve its database state as a subsequent downgrade failed with errors indicating as much.

soltysh (approved):
lgtm

a1994sc (approved):

PR #5303, opened 2026-09-01
feat: add oci pull retries for packages

mjnagel:
## Description

After experiencing some bouts of 429s while trying to pull large packages from GitHub (and this resulting in failing CI), I thought it might make sense to retry on certain errors during OCI package pulls (explicit pulls, signing OCI, downloading init, OCI deploys).

This is quite similar to https://github.com/zarf-dev/zarf/issues/4470 and leverages some of the changes from https://github.com/zarf-dev/zarf/pull/4664 but applied to package pulls.

## Checklist before merging


src/pkg/utils/network.go
```diff
@@ -162,9 +164,9 @@ func httpGetFile(ctx context.Context, url string, destinationFile *os.File) (err
 	return nil
 }
 
-// parseRetryAfter parses the Retry-After header value into a duration.
+// ParseRetryAfter parses the Retry-After header value into a duration.
 // It supports both delay-seconds (integer) and HTTP-date formats.
-func parseRetryAfter(value string) time.Duration {
+func ParseRetryAfter(value string) time.Duration {
```
brandtkeller, line 169: nit: if making this public I want to think about any bounding/validation we might want to incorporate. Should we be returning errors or performing other input validation?

src/pkg/zoci/common.go
```diff
@@ -126,6 +140,27 @@ func NewRemoteWithOptions(ctx context.Context, url string, platform ocispec.Plat
 	return remote, nil
 }
 
+// configurePullRetries installs Zarf's retry policy for OCI pull requests.
+// Package pulls do not use upload progress, so this unwraps the dependency's
+// progress transport to avoid nesting its default retry policy inside ours.
+func configurePullRetries(remote *Remote, retries int) error {
```
brandtkeller, line 146: I am curious if the intersection of oci-concurrency and 429/retries is something we need to consider here or in the underlying layers. 

Doing some digging with some help and I think the transport we've replaced (appropriately) had a jitter to prevent concurrency on retry from slamming the endpoint again all at once?

src/pkg/packager/load.go
```diff
@@ -92,8 +94,15 @@ func LoadPackage(ctx context.Context, source string, opts LoadOptions) (_ *layou
 	tmpPath := filepath.Join(tmpDir, "data.tar.zst")
 	switch srcType {
 	case "oci":
+		if opts.Retries < 0 {
```
brandtkeller, line 97: I think we can isolate this error case to the zoci package. The zero replacement following is still valid though.

brandtkeller (changes requested):
Minor changes otherwise general strategy is sound.

src/pkg/packager/load.go
mjnagel: Yep good call, removed this.

src/pkg/utils/network.go
mjnagel: Good call. I went back and forth here on duplicating the relatively simple logic vs moving to a new internal vs making public...ultimately landed on public. I updated this to handle negative seconds and overflows and return errors, hopefully matches what you were thinking.

src/pkg/zoci/common.go
mjnagel: Ah yeah this is really interesting. I was not super familiar here so had to lean on some AI assistance but I think implemented a basic jitter here that applies to most cases, except for explicit retry-after (when under the max delay).

src/pkg/utils/network.go
```diff
@@ -162,19 +165,27 @@ func httpGetFile(ctx context.Context, url string, destinationFile *os.File) (err
...
-		return 0
+		return 0, nil
 	}
 	if seconds, err := strconv.ParseInt(value, 10, 64); err == nil {
-		return time.Duration(seconds) * time.Second
+		if seconds <= 0 {
+			return 0, fmt.Errorf("retry-after delay must be positive")
+		}
+		const maxDurationSeconds = int64(1<<63-1) / int64(time.Second)
+		if seconds > maxDurationSeconds {
```
brandtkeller, line 179: I feel like parsing and policy should be separate - leaving the policy to the caller.

src/pkg/utils/network.go
```diff
@@ -162,19 +165,27 @@ func httpGetFile(ctx context.Context, url string, destinationFile *os.File) (err
 	return nil
 }
 
-// parseRetryAfter parses the Retry-After header value into a duration.
-// It supports both delay-seconds (integer) and HTTP-date formats.
-func parseRetryAfter(value string) time.Duration {
+// ParseRetryAfter validates and parses a Retry-After header into a duration.
+// An empty value returns zero without an error.
+func ParseRetryAfter(value string) (time.Duration, error) {
```
brandtkeller, line 170: Interested in what @zarf-maintainers and @zarf-reviewers think here - Should we be considering functions like these be moved to internal. We were discussing internal recently and I think it might behoove us to use it for API's such as these where we potentially don't want SDK utility? I've been pretty conservative about how we support all of these but perhaps a deprecation policy that signals the utils package as not part of the explicit semver contract is also okay.

brandtkeller (changes requested):

brandtkeller (changes requested):
Minor nit on function scope otherwise just want to see what the team thinks about publicly exposed utils functions.

src/pkg/utils/network.go
brandtkeller: dang - wrong team names - @zarf-dev/maintainers @zarf-dev/reviewers

src/pkg/utils/network.go
mjnagel: This specific check is more "parse/error handling", not actual policy. I can rename the const if that's the confusion - `maxDurationSeconds` is the max int that can parse into a duration without overflowing. This is not the same as `MaxRetryAfter` that callers check as a policy guard against waiting too long.

src/pkg/utils/network.go
AustinAbro321: I would be okay with either, but prefer documenting that utils is not part of the semver contract. 

I think some of the utils files could also be their own small package, or even all of utils could turn into several packages

AustinAbro321 (changes requested):
Unfortunate timing, but we should hold off on this PR until https://github.com/zarf-dev/zarf/pull/5324 is merged and oci is back as a sub-package in Zarf. 

Reason being is that the oci package and so Zarf currently, has a retry policy on pulls https://github.com/defenseunicorns/pkg/blob/main/helpers/transport.go. In this branch, we use the oras retry policy when retries are 0, and this new retry policy when retries are greater than 0. I'd rather not have two systems 

This PR is giving more generous limits, but I think this can be configured using https://github.com/oras-project/oras-go/blob/v2.6.2/registry/remote/retry/policy.go#L114-L135.

I don't fault you for not seeing the current retry system, had I not happened to be review https://github.com/zarf-dev/zarf/pull/5324 recently, I might have missed it as well

PR #5310, opened 2026-09-03
feat: support listing a deployed package by name

### LINKED ISSUE #5220: Add support for a named packaged with zarf package list

Add support to `zarf package list` to add a `NAME` argument to the command that returns a single-named deployed package from the cluster state. 

This would be an enabler to future improvements around deployed package state retrieval.

- **Given** a package has been deployed to the cluster with queryable state
- **When** zarf package list <name> is executed
- **Then** The command returns context about the singular-named deployed package. 

Continue to utilize `zarf package inspect` for this functionality. Ultimately I believe the `DeployedPackage` type and the inspection targets may start to diverge more as opportunities arise to store context of an operation.

### Additional context

This would hopefully feed opportunities such as adding more information to DeployedPackages in a way that can be used dynamically or through simple retrieval.

acald-creator:
## Description

Adds an optional package name argument to `zarf package list`.

- `zarf package list` continues to list all deployed packages.
- `zarf package list <PACKAGE_NAME>` retrieves and displays a single deployed package.
- JSON and YAML output retain the existing list-shaped format.
- Adds argument validation and package-name completion.
- Updates the generated command documentation.

## Related Issue

Fixes #5220

## Checklist before merging

- [ x ] [Contributor Guide Steps](https://github.com/zarf-dev/zarf/blob/main/CONTRIBUTING.md#developer-workflow) followed

### DIFF
--- a/src/cmd/package.go
+++ b/src/cmd/package.go
@@ -1292,9 +1292,10 @@ func (o *packageInspectDefinitionOptions) run(cmd *cobra.Command, args []string)
 }
 
 type packageListOptions struct {
-	outputFormat outputFormat
-	outputWriter io.Writer
-	cluster      *cluster.Cluster
+	outputFormat      outputFormat
+	outputWriter      io.Writer
+	cluster           *cluster.Cluster
+	namespaceOverride string
 }
 
 func newPackageListOptions() *packageListOptions {
@@ -1309,20 +1310,23 @@ func newPackageListCommand() *cobra.Command {
 	o := newPackageListOptions()
 
 	cmd := &cobra.Command{
-		Use:     "list",
-		Aliases: []string{"l", "ls"},
-		Short:   lang.CmdPackageListShort,
-		RunE: func(cmd *cobra.Command, _ []string) error {
+		Use:               "list [PACKAGE_NAME]",
+		Aliases:           []string{"l", "ls"},
+		Short:             lang.CmdPackageListShort,
+		Args:              cobra.MaximumNArgs(1),
+		ValidArgsFunction: getPackageCompletionArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := cmd.Context()
 			err := o.complete(ctx)
 			if err != nil {
 				return err
 			}
-			return o.run(ctx)
+			return o.run(ctx, args)
 		},
 	}
 
 	cmd.Flags().VarP(&o.outputFormat, "output-format", "o", "Prints the output in the specified format. Valid options: table, json, yaml")
+	cmd.Flags().StringVarP(&o.namespaceOverride, "namespace", "n", "", lang.CmdPackageListFlagNamespace)
 
 	return cmd
 }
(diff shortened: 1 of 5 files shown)


src/cmd/package.go
```diff
@@ -1347,10 +1349,20 @@ type packageListInfo struct {
...
 }
 
-func (o *packageListOptions) run(ctx context.Context) error {
-	deployedZarfPackages, err := o.cluster.GetDeployedZarfPackages(ctx)
-	if err != nil && len(deployedZarfPackages) == 0 {
-		return fmt.Errorf("unable to get the packages deployed to the cluster: %w", err)
+func (o *packageListOptions) run(ctx context.Context, args []string) error {
+	var deployedZarfPackages []state.DeployedPackage
+	if len(args) == 1 {
+		deployedPackage, err := o.cluster.GetDeployedPackage(ctx, args[0])
```
brandtkeller: we'll need to pass in `o.NamespaceOverride` if used here.

brandtkeller, line 1355: This is going to be unable to get DeployedPackages deployed with a namespace override - I think we'll need a flag and then checking for the override. 

Mirroring helm works here in my opinion. So `zarf package list NAME` should return a DeployedPackage by name without a namespace override. Adding the `-n NAMESPACE` flag then returns named DeployedPackage with an override.

brandtkeller (changes requested):
Minor addition otherwise this is spot on.

Typically I recommend to aligning the happy path left and attempting to remove `else` statements but it feels like an un-needed optimization unless you want to take a quick look.

src/cmd/package.go
acald-creator: When I was initially looking at the outputs, I was thinking about namespace or not. I will make changes.

src/cmd/package.go
```diff
@@ -1309,20 +1310,23 @@ func newPackageListCommand() *cobra.Command {
...
 			if err != nil {
 				return err
 			}
-			return o.run(ctx)
+			return o.run(ctx, args)
 		},
 	}
 
 	cmd.Flags().VarP(&o.outputFormat, "output-format", "o", "Prints the output in the specified format. Valid options: table, json, yaml")
+	cmd.Flags().StringVarP(&o.namespaceOverride, "namespace", "n", "", lang.CmdPackageListFlagNamespace)
```
brandtkeller, line 1329: Commenting here for locality to changes -  I would challenge the use of a flag that can only be used when arguments are provided. 

Rather I think we need to look at updating the retrieval logic to use namespace as a filter when the flag is used but no argument is provided.

So then the matrix becomes:
- No argument + no flags = all packages
- No argument + namespace flag = all packages with a specific namespace override
- Argument + no flag = package without override
- Argument + namespace flag = package with override

If we can do so without a public API change then great - I would also consider a breaking change if required.

brandtkeller (changes requested):
Great iteration! One modification concerning filtering semantics but otherwise an improvement.

brandtkeller (approved):
Lgtm - I think this sets us up for some future improvements around the structure of what to output and the new state management proposal. 

I'd defer changing the shape of the return (ie a list vs single object) until we have that decomposed further.

AustinAbro321 (approved):

PR #5312, opened 2026-09-03
docs: add S3-backed registry example to init package HA docs

### LINKED ISSUE #3948: Update docs.zarf.dev init docs on S3-backed HA Registry
Currently, the Zarf docs site mentions that S3 can be used as the storage backend for HA registries, but doing so requires a custom init package and specific S3 configuration. We don't actually provide a canonical setup for S3, or provide much in the way of guidance on how to set it up. Ideally, users are informed they have to set up in S3 bucket, configure IAM policies and credentials, set up a CA bundle if necessary, and _especially_ how to configure the registry with the S3 (see community docs [here](https://docs.smoothglue.com/docs/6.12.0/install/configurations/high-availability-zarf-registry/) for example). Actually setting this up requires a bit of digging and there's no running examples in the Zarf repo to work from and adapt.

https://docs.zarf.dev/ref/init-package/#making-the-registry-highly-available

- **Given** a Zarf deployment that requires a HA registry layer and S3 is a valid storage option 
- **When** a user goes to configure S3 as a registry storage backend
- **Then** that user should have clear documentation to follow and ideally live examples.

The particulars of the setup will always depend on the user's environment, but we can do more to help users avoid some common pitfalls.

### Additional context
Related issues:
#375

This is a known usecase, and we've had some prior discussion on ensuring Zarf's remains S3 compatible with an e2e test. This isn't exactly the same thing as setup documentation, but should provide a helpful small scale example, as well as coverage against regressions. https://github.com/zarf-dev/zarf/issues/3285

NAME-ASHWANIYADAV:
## Description

The "Making the Registry Highly-Available" section mentioned an S3-compatible backend as an option but only documented the ReadWriteMany path. This adds a walkthrough for backing the registry with S3 object storage.

It turns out no custom init package is needed - the registry chart takes its storage driver from env vars, so S3 is deploy-time config only. The new section covers disabling the PVC, passing the s3 driver settings via REGISTRY_EXTRA_ENVS, the required bucket permissions, non-AWS endpoints, and an IRSA-based alternative to static keys. It also calls out that REGISTRY_PVC_ENABLED must be false, since leaving the PVC on defines the filesystem driver alongside s3 and the registry won't start with two drivers configured.

## How I validated

Rendered the registry chart with `helm template` for each case:
- PVC disabled + S3 env vars -> only the s3 driver is set, no filesystem rootdirectory
- PVC enabled + S3 env vars -> both drivers render (confirms the caution)
- IRSA config -> the annotated service account is created and referenced by the deployment

## Related Issue

Closes #3948

## Checklist before merging

- [x] Contributor Guide Steps followed

### DIFF
--- a/site/src/content/docs/ref/init-package.mdx
+++ b/site/src/content/docs/ref/init-package.mdx
@@ -177,38 +177,9 @@ Notably, the `REGISTRY_AFFINITY_CUSTOM` variable overrides the default pod anti-
 
 Instead of a `ReadWriteMany` volume, you can back the registry with S3-compatible object storage. The registry image ships with the [`s3` storage driver](https://distribution.github.io/distribution/storage-drivers/s3/), so every replica reads and writes the same bucket. This removes the shared-volume requirement, which is useful in environments where a `ReadWriteMany` storage class is unavailable or where an NFS-backed one has proven unreliable.
 
-No custom init package is required. The default registry chart takes its storage driver from environment variables, so you can switch to S3 entirely through deploy-time [config](/ref/config-files/). Disable the PVC with `REGISTRY_PVC_ENABLED` and pass the driver settings through `REGISTRY_EXTRA_ENVS`:
+No custom init package is required. The default registry chart takes its storage driver from environment variables, so you can switch to S3 entirely through deploy-time [config](/ref/config-files/): disable the PVC with `REGISTRY_PVC_ENABLED` and pass the driver settings through `REGISTRY_EXTRA_ENVS`.
 
-```yaml
-# zarf-config.yaml
-package:
-  deploy:
-    set:
-      REGISTRY_PVC_ENABLED: "false"
-      REGISTRY_EXTRA_ENVS: |
-        - name: REGISTRY_STORAGE
-          value: "s3"
-        - name: REGISTRY_STORAGE_S3_REGION
-          value: "us-east-1"
-        - name: REGISTRY_STORAGE_S3_BUCKET
-          value: "my-zarf-registry"
-        - name: REGISTRY_STORAGE_S3_ACCESSKEY
-          value: "<access-key>"
-        - name: REGISTRY_STORAGE_S3_SECRETKEY
-          value: "<secret-key>"
-```
-
-Each entry in `REGISTRY_EXTRA_ENVS` is rendered directly into the registry container's environment. The registry reads any `REGISTRY_STORAGE_S3_*` variable as the matching [`s3` driver parameter](https://distribution.github.io/distribution/storage-drivers/s3/) — for example, set `REGISTRY_STORAGE_S3_REGIONENDPOINT` (and, if needed, `REGISTRY_STORAGE_S3_SECURE`) to point at a non-AWS, S3-compatible backend such as MinIO.
-
-:::caution
-
-`REGISTRY_PVC_ENABLED` must be set to `"false"`. When the PVC is enabled the chart also sets the filesystem storage driver (`REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY`), and the registry refuses to start if more than one storage driver is configured.
-
-:::
-
-The bucket credentials need permission to read and write objects and to perform multipart uploads — `s3:GetObject`, `s3:PutObject`, `s3:DeleteObject`, `s3:ListBucket`, `s3:ListMultipartUploadParts`, and `s3:AbortMultipartUpload`.
-
-On EKS you can avoid static keys by using [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html). Omit `REGISTRY_STORAGE_S3_ACCESSKEY` and `REGISTRY_STORAGE_S3_SECRETKEY`, and have Zarf create a service account annotated with the IAM role instead. `REGISTRY_CREATE_SERVICE_ACCOUNT` must be `"true"` for the annotations to be applied:
+On EKS, the preferred setup is [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) - no credentials appear in the pod spec, and Zarf creates a service account annotated with the IAM role. `REGISTRY_CREATE_SERVICE_ACCOUNT` must be `"true"` for the annotations to be applied:
 
 ```yaml
 # zarf-config.yaml
@@ -227,6 +198,54 @@ package:
         - name: REGISTRY_STORAGE_S3_BUCKET
           value: "my-zarf-registry"
 ```
+
+The registry reads any `REGISTRY_STORAGE_S3_*` environment variable as the matching [`s3` driver parameter](https://distribution.github.io/distribution/storage-drivers/s3/) - the full parameter list, and the IAM permissions the credentials need, are documented there.
+
+Outside of EKS, or with a non-AWS S3-compatible backend such as MinIO, pass static credentials and the endpoint through `REGISTRY_EXTRA_ENVS` instead. When `REGISTRY_STORAGE_S3_REGIONENDPOINT` is set, the registry also requires `REGISTRY_STORAGE_S3_FORCEPATHSTYLE` to be `"true"`; leave both unset when using Amazon S3 itself:
+
+```yaml
+# zarf-config.yaml
+package:
+  deploy:
+    set:
+      REGISTRY_PVC_ENABLED: "false"
+      REGISTRY_EXTRA_ENVS: |
+        - name: REGISTRY_STORAGE
+          value: "s3"
+        - name: REGISTRY_STORAGE_S3_REGION
+          value: "us-east-1"
+        - name: REGISTRY_STORAGE_S3_BUCKET
+          value: "my-zarf-registry"
+        - name: REGISTRY_STORAGE_S3_REGIONENDPOINT
+          value: "https://minio.example.com:9000"
+        - name: REGISTRY_STORAGE_S3_FORCEPATHSTYLE
+          value: "true"
+        - name: REGISTRY_STORAGE_S3_ACCESSKEY
+          value: "<access-key>"
+        - name: REGISTRY_STORAGE_S3_SECRETKEY
+          value: "<secret-key>"
+```
+
+Entries in `REGISTRY_EXTRA_ENVS` are rendered verbatim into the registry container's environment, so plain `value:` credentials end up in the deployment spec. To keep them out of it, store the keys in a Kubernetes Secret in the `zarf` namespace (created before `zarf init`) and reference them with `valueFrom` instead:
+
+```yaml
+        - name: REGISTRY_STORAGE_S3_ACCESSKEY
+          valueFrom:
+            secretKeyRef:
+              name: zarf-registry-s3-creds
+              key: access-key
+        - name: REGISTRY_STORAGE_S3_SECRETKEY
+          valueFrom:
+            secretKeyRef:
+              name: zarf-registry-s3-creds
+              key: secret-key
+```
+
+:::caution
+
+`REGISTRY_PVC_ENABLED` must be set to `"false"`. When the PVC is enabled the chart also sets the filesystem storage driver (`REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY`), and the registry refuses to start if more than one storage driver is configured.
+
+:::
 
 :::note
 
--- a/site/src/content/docs/ref/init-package.mdx
+++ b/site/src/content/docs/ref/init-package.mdx
@@ -173,6 +173,67 @@ package:
 
 Notably, the `REGISTRY_AFFINITY_CUSTOM` variable overrides the default pod anti-affinity, and `REGISTRY_HPA_AUTO_SIZE` automatically adjusts the minimum and maximum replicas for the registry based on the number of nodes in the cluster. If you prefer to manually set the minimum and maximum replicas, you can use `REGISTRY_HPA_MIN` and `REGISTRY_HPA_MAX` to specify the desired values.
 
+##### Using an S3-Compatible Storage Backend
+
+Instead of a `ReadWriteMany` volume, you can back the registry with S3-compatible object storage. The registry image ships with the [`s3` storage driver](https://distribution.github.io/distribution/storage-drivers/s3/), so every replica reads and writes the same bucket. This removes the shared-volume requirement, which is useful in environments where a `ReadWriteMany` storage class is unavailable or where an NFS-backed one has proven unreliable.
+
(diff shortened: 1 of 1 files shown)


NAME-ASHWANIYADAV:
Hi @brandtkeller , could you please review this PR whenever you get a chance? Let me know if you need any changes.

site/src/content/docs/ref/init-package.mdx
```diff
@@ -173,6 +173,67 @@ package:
...
+
+Each entry in `REGISTRY_EXTRA_ENVS` is rendered directly into the registry container's environment. The registry reads any `REGISTRY_STORAGE_S3_*` variable as the matching [`s3` driver parameter](https://distribution.github.io/distribution/storage-drivers/s3/) — for example, set `REGISTRY_STORAGE_S3_REGIONENDPOINT` (and, if needed, `REGISTRY_STORAGE_S3_SECURE`) to point at a non-AWS, S3-compatible backend such as MinIO.
+
+:::caution
+
+`REGISTRY_PVC_ENABLED` must be set to `"false"`. When the PVC is enabled the chart also sets the filesystem storage driver (`REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY`), and the registry refuses to start if more than one storage driver is configured.
+
+:::
+
+The bucket credentials need permission to read and write objects and to perform multipart uploads — `s3:GetObject`, `s3:PutObject`, `s3:DeleteObject`, `s3:ListBucket`, `s3:ListMultipartUploadParts`, and `s3:AbortMultipartUpload`.
```
brandtkeller: Looking at the [linked documentation](https://distribution.github.io/distribution/storage-drivers/s3/#s3-permission-scopes) - I see more than what is just mentioned here. Should we direct to the upstream docs instead of trying to synchronize ourselves?

site/src/content/docs/ref/init-package.mdx
```diff
@@ -173,6 +173,67 @@ package:
...
+          value: "us-east-1"
+        - name: REGISTRY_STORAGE_S3_BUCKET
+          value: "my-zarf-registry"
+        - name: REGISTRY_STORAGE_S3_ACCESSKEY
+          value: "<access-key>"
+        - name: REGISTRY_STORAGE_S3_SECRETKEY
+          value: "<secret-key>"
+```
+
+Each entry in `REGISTRY_EXTRA_ENVS` is rendered directly into the registry container's environment. The registry reads any `REGISTRY_STORAGE_S3_*` variable as the matching [`s3` driver parameter](https://distribution.github.io/distribution/storage-drivers/s3/) — for example, set `REGISTRY_STORAGE_S3_REGIONENDPOINT` (and, if needed, `REGISTRY_STORAGE_S3_SECURE`) to point at a non-AWS, S3-compatible backend such as MinIO.
```
brandtkeller: I am interpreting this myself - but do we need to mention `forcepathstyle` here in any regard? 

> regionendpoint: (optional) Endpoint URL for S3 compatible APIs, from version 3+ it’s required to be used with forcepathstyle: true. Given the regionendpoint overrides the API host domain, forcing the path style is necessary, see [more about](https://github.com/distribution/distribution/issues/4528). This option should not be provided when using Amazon S3.

site/src/content/docs/ref/init-package.mdx
```diff
@@ -173,6 +173,67 @@ package:
 
 Notably, the `REGISTRY_AFFINITY_CUSTOM` variable overrides the default pod anti-affinity, and `REGISTRY_HPA_AUTO_SIZE` automatically adjusts the minimum and maximum replicas for the registry based on the number of nodes in the cluster. If you prefer to manually set the minimum and maximum replicas, you can use `REGISTRY_HPA_MIN` and `REGISTRY_HPA_MAX` to specify the desired values.
 
+##### Using an S3-Compatible Storage Backend
+
+Instead of a `ReadWriteMany` volume, you can back the registry with S3-compatible object storage. The registry image ships with the [`s3` storage driver](https://distribution.github.io/distribution/storage-drivers/s3/), so every replica reads and writes the same bucket. This removes the shared-volume requirement, which is useful in environments where a `ReadWriteMany` storage class is unavailable or where an NFS-backed one has proven unreliable.
+
+No custom init package is required. The default registry chart takes its storage driver from environment variables, so you can switch to S3 entirely through deploy-time [config](/ref/config-files/). Disable the PVC with `REGISTRY_PVC_ENABLED` and pass the driver settings through `REGISTRY_EXTRA_ENVS`:
```
brandtkeller: Definitely how this can be accomplished today - perhaps we create a follow-on issue to harden this via secret references.

Albeit AWS specific as I understand - perhaps we should label the IRSA method as the primary?

This is all given that the environment mechanism here exposes those long lived credentials in the deployment/pod specs.

brandtkeller (changes requested):
minor changes otherwise this is great depth.

site/src/content/docs/ref/init-package.mdx
NAME-ASHWANIYADAV: Agreed, that list would drift. Replaced it with a pointer to the upstream driver docs, which now cover both the parameter list and the required permissions.

site/src/content/docs/ref/init-package.mdx
NAME-ASHWANIYADAV: I have  added REGISTRY_STORAGE_S3_FORCEPATHSTYLE: "true" to the non-AWS example, with a note that it's required alongside regionendpoint and that both should be left unset for Amazon S3 itself.

site/src/content/docs/ref/init-package.mdx
NAME-ASHWANIYADAV: Restructured: IRSA is now the lead method with a note that no credentials land in the pod spec, and static keys moved to the non-AWS path.

Since REGISTRY_EXTRA_ENVS entries are passed through toYaml, valueFrom/ secretKeyRef already works today - added that variant so the keys can live in a pre-created Secret instead of the deployment spec (verified it renders through the chart). Happy to open the follow-on issue for first-class secret support if you'd like.

NAME-ASHWANIYADAV:
The validate-upgrade failure also looks unrelated - the diff is still docs-only, and this same job passed in ~22 minutes on the previous run of this branch. This run died at 41 minutes, which looks like a hang in the e2e rather than anything in the change. Could @brandtkeller re-run it when you get a chance?

site/src/content/docs/ref/init-package.mdx
```diff
@@ -173,6 +173,86 @@ package:
...
+
+`REGISTRY_PVC_ENABLED` must be set to `"false"`. When the PVC is enabled the chart also sets the filesystem storage driver (`REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY`), and the registry refuses to start if more than one storage driver is configured.
+
+:::
+
+:::note
+
+Because the registry no longer relies on a shared volume, the default pod anti-affinity from the `ReadWriteMany` example above is not required for S3. Scaling with `REGISTRY_HPA_AUTO_SIZE`, `REGISTRY_HPA_MIN`, and `REGISTRY_HPA_MAX` works the same way.
+
+:::
```
a1994sc: So internally I have wired zarf to use a rook-ceph  back-end as it provides a s3 endpoint if configured. I believe that when I looked into there was not as much configuration we have now-a-days.

The config I used was something to effect of:

```yaml
---
persistence:
  enabled: false

secrets:
  configData:
    storage:
      s3:
        regionendpoint: "###ZARF_VAR_REGISTRY_S3_BUCKET_URL###"
        region: "###ZARF_VAR_REGISTRY_S3_BUCKET_REGION###"
        bucket: "###ZARF_VAR_REGISTRY_S3_BUCKET_NAME###"
        secure: false
        v4auth: true
        forcepathstyle: true
        chunksize: 5242880
        multipartcopychunksize: 33554432
        multipartcopymaxconcurrency: 100
        multipartcopythresholdsize: 33554432
      redirect:
        disable: true

extraEnvVars:
  # Additional ENV variables to set
  - name: REGISTRY_STORAGE_S3_SECRETKEY
    valueFrom:
      secretKeyRef:
        name: registry-bucket
        key: AWS_SECRET_ACCESS_KEY
  - name: REGISTRY_STORAGE_S3_ACCESSKEY
    valueFrom:
      secretKeyRef:
        name: registry-bucket
        key: AWS_ACCESS_KEY_ID
```

site/src/content/docs/ref/init-package.mdx
```diff
@@ -173,6 +173,86 @@ package:
 
 Notably, the `REGISTRY_AFFINITY_CUSTOM` variable overrides the default pod anti-affinity, and `REGISTRY_HPA_AUTO_SIZE` automatically adjusts the minimum and maximum replicas for the registry based on the number of nodes in the cluster. If you prefer to manually set the minimum and maximum replicas, you can use `REGISTRY_HPA_MIN` and `REGISTRY_HPA_MAX` to specify the desired values.
 
+##### Using an S3-Compatible Storage Backend
+
+Instead of a `ReadWriteMany` volume, you can back the registry with S3-compatible object storage. The registry image ships with the [`s3` storage driver](https://distribution.github.io/distribution/storage-drivers/s3/), so every replica reads and writes the same bucket. This removes the shared-volume requirement, which is useful in environments where a `ReadWriteMany` storage class is unavailable or where an NFS-backed one has proven unreliable.
```
brandtkeller, line 178: > where an NFS-backed one has proven unreliable.

language here is slightly subjective and deviates from the general tone or voice the documentation currently has - perhaps:


```suggestion
Instead of a `ReadWriteMany` volume, you can back the registry with S3-compatible object storage. The registry image ships with the [`s3` storage driver](https://distribution.github.io/distribution/storage-drivers/s3/), so every replica reads and writes the same bucket. This removes the shared-volume requirement, which is useful in environments where a ReadWriteMany storage class is unavailable or unsuitable..
```

site/src/content/docs/ref/init-package.mdx
```diff
@@ -173,6 +173,86 @@ package:
...
+
+No custom init package is required. The default registry chart takes its storage driver from environment variables, so you can switch to S3 entirely through deploy-time [config](/ref/config-files/): disable the PVC with `REGISTRY_PVC_ENABLED` and pass the driver settings through `REGISTRY_EXTRA_ENVS`.
+
+On EKS, the preferred setup is [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) - no credentials appear in the pod spec, and Zarf creates a service account annotated with the IAM role. `REGISTRY_CREATE_SERVICE_ACCOUNT` must be `"true"` for the annotations to be applied:
```
brandtkeller, line 182: Tone here slightly deviates as well - perhaps:

```suggestion
On EKS, use [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) - to avoid placing credentials in the pod spec. Zarf will create a service account annotated with the IAM role. `REGISTRY_CREATE_SERVICE_ACCOUNT` must be `"true"` for the annotations to be applied:
```

brandtkeller (commented):
Minor changes to tone otherwise I have no issues with the content after reading from scratch on the deploy preview.

site/src/content/docs/ref/init-package.mdx
```diff
@@ -173,6 +173,86 @@ package:
...
+        - name: REGISTRY_STORAGE_S3_BUCKET
+          value: "my-zarf-registry"
+        - name: REGISTRY_STORAGE_S3_REGIONENDPOINT
+          value: "https://minio.example.com:9000"
+        - name: REGISTRY_STORAGE_S3_FORCEPATHSTYLE
+          value: "true"
+        - name: REGISTRY_STORAGE_S3_ACCESSKEY
+          value: "<access-key>"
+        - name: REGISTRY_STORAGE_S3_SECRETKEY
+          value: "<secret-key>"
```
a1994sc, line 226: For intra-cluster usage, either with rook-ceph or minio, you will likely want to add:

```yaml
        - name: REGISTRY_STORAGE_REDIRECT_DISABLE
          value: "true"
```

As when a node tries to pull from an intra-cluster endpoint, `minio-server.minio.svc.cluster.local` it will not resolve. At lease with the above config, the registry will act as a bit of a proxy.

site/src/content/docs/ref/init-package.mdx
```diff
@@ -173,6 +173,86 @@ package:
...
+package:
+  deploy:
+    set:
+      REGISTRY_PVC_ENABLED: "false"
+      REGISTRY_CREATE_SERVICE_ACCOUNT: "true"
+      REGISTRY_SERVICE_ACCOUNT_ANNOTATIONS: |
+        eks.amazonaws.com/role-arn: arn:aws:iam::111122223333:role/zarf-registry-s3
+      REGISTRY_EXTRA_ENVS: |
+        - name: REGISTRY_STORAGE
+          value: "s3"
```
a1994sc, line 195: Catch that Claude found and it makes sense:

`configuration.go Storage.UnmarshalYAML`: a bare string payload does `*storage = Storage{storageType: Parameters{}}` — replaces whole map. Parser sorts env lexically (`parser.go:96-102`), so `REGISTRY_STORAGE` applies first and wipes `storage.cache.blobdescriptor: inmemory` that the chart sets in `chart/values.yaml secrets.configData. REGISTRY_STORAGE_S3_*` alone already creates the s3 key, so the REGISTRY_STORAGE entry is unnecessary and costs cache. With S3 that cache matters — every descriptor stat becomes a bucket round trip.

Consider adding REGISTRY_STORAGE_CACHE_BLOBDESCRIPTOR: "inmemory". Confirm with helm template plus a running registry /debug config dump.

site/src/content/docs/ref/init-package.mdx
```diff
@@ -173,6 +173,86 @@ package:
...
+`REGISTRY_PVC_ENABLED` must be set to `"false"`. When the PVC is enabled the chart also sets the filesystem storage driver (`REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY`), and the registry refuses to start if more than one storage driver is configured.
+
+:::
+
+:::note
+
+Because the registry no longer relies on a shared volume, the default pod anti-affinity from the `ReadWriteMany` example above is not required for S3. Scaling with `REGISTRY_HPA_AUTO_SIZE`, `REGISTRY_HPA_MIN`, and `REGISTRY_HPA_MAX` works the same way.
+
+:::
+
```
a1994sc, line 255: So internally I have wired zarf to use a rook-ceph  back-end as it provides a s3 endpoint if configured. I believe that when I looked into there was not as much configuration we have now-a-days. 

I can't share the full config, but I had to provide a values file overriding all the various values with the, `.secrets.configData.storage.s3`

a1994sc (changes requested):
looks good, here are a few comments from my own usage.

PR #5317, opened 2026-09-04
feat(cluster): support random registry pod selection

### LINKED ISSUE #5313: Improve registry load balancing during concurrent package deploys

When the internal registry runs with multiple replicas, concurrent `zarf package deploy` commands appear to send all image-push traffic to the same pod. Based on manual testing of concurrent deploys as well as script driven approaches, all traffic for several concurrent deploys of packages (image pushes) landed on the same replica of the Zarf registry, rather than being distributed across replicas.

- Given multiple ready replicas of the internal registry
- When packages are deployed concurrently
- Then registry traffic should be reasonably distributed across replicas

Unsure of an alternative besides accepting this as a known behavior (and maybe documenting).

In terms of implementation I think there could be a lot of different ways to approach.

### Additional context

The image push path calls [`ConnectToZarfRegistryEndpoint`](https://github.com/zarf-dev/zarf/blob/main/src/pkg/images/push.go#L94). For an internal registry, that creates a Service tunnel in [`tunnel.go`](https://github.com/zarf-dev/zarf/blob/main/src/pkg/cluster/tunnel.go#L155-L157). The tunnel then selects the first Ready pod in [`getAttachablePodForService`](https://github.com/zarf-dev/zarf/blob/main/src/pkg/cluster/tunnel.go#L557-L583) and opens the port-forward directly to that pod. My suspicion is that the returned list is generally ordered the same so "first pod" here is getting tunneled to the same pod every time, rather than following standard Service loadbalancing or another approach for picking a pod.

I think there is a related item around concurrent image pushes from the same package, but unsure on complexity involved there (and maybe related image layers that are shared between two packages being deployed at the same time).

brandtkeller:
## Description

Randomizes internal-registry tunnel pod selection so image pushes are distributed across ready registry replicas instead of always targeting the first pod returned by Kubernetes. Adds an optional service-pod selector to tunnels, preserves existing selection behavior elsewhere.
## Related Issue

Fixes #5313

## Checklist before merging

### DIFF
--- a/src/pkg/cluster/tunnel.go
+++ b/src/pkg/cluster/tunnel.go
@@ -155,7 +155,7 @@ func (c *Cluster) ConnectToZarfRegistryEndpoint(ctx context.Context, registryInf
 	var tunnel *Tunnel
 	if registryInfo.IsInternal() {
 		// Establish a registry tunnel to send the images to the zarf registry
-		if tunnel, err = c.NewTunnel(state.ZarfNamespaceName, SvcResource, ZarfRegistryName, "", 0, ZarfRegistryPort, withServicePodSelector(selectRandomServicePod)); err != nil {
+		if tunnel, err = c.NewTunnel(state.ZarfNamespaceName, SvcResource, ZarfRegistryName, "", 0, ZarfRegistryPort); err != nil {
 			return "", tunnel, err
 		}
 	} else if dns.IsServiceURL(registryInfo.Address) {
@@ -322,18 +322,6 @@ const (
 // TunnelOption is a function that configures a tunnel
 type TunnelOption func(*Tunnel)
 
-type servicePodSelector func([]corev1.Pod) corev1.Pod
-
-func withServicePodSelector(selector servicePodSelector) TunnelOption {
-	return func(t *Tunnel) {
-		t.servicePodSelector = selector
-	}
-}
-
-func selectRandomServicePod(pods []corev1.Pod) corev1.Pod {
-	return pods[rand.IntN(len(pods))]
-}
-
 // WithListenAddress will set the listen address for the tunnel
 func WithListenAddress(addr []string) TunnelOption {
 	return func(t *Tunnel) {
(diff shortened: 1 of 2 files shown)


src/pkg/cluster/tunnel.go
```diff
@@ -578,12 +592,25 @@ func (tunnel *Tunnel) getAttachablePodForService(ctx context.Context) (string, e
...
 	// graceful termination (e.g. mid-rollout), so without also checking these, a
 	// port-forward can bind to a pod that's already on its way out.
+	if tunnel.servicePodSelector == nil {
+		for _, pod := range podList.Items {
+			if pod.DeletionTimestamp == nil && podutils.IsPodReady(&pod) {
+				return pod.Name, nil
+			}
+		}
+		return "", fmt.Errorf("no ready pods found for service %s", tunnel.resourceName)
+	}
```
soltysh: This entire block is a rough copy&paste from the block below, which will make the maintenance harder than necessary. How about creating the list of pods once, and then based on the `servicePodSelector` return first or invoke the randomizer? 
Something like:
```golang
readyPods := make([]corev1.Pod, 0, len(podList.Items))
for _, pod := range podList.Items {
	if pod.DeletionTimestamp == nil && podutils.IsPodReady(&pod) {
		readyPods = append(readyPods, pod)
	}
}
if len(readyPods) == 0 {
	return "", fmt.Errorf("no ready pods found for service %s", tunnel.resourceName)
}
if tunnel.servicePodSelector == nil {
	return tunnel.servicePodSelector(readyPods).Name, nil
}
return readyPods[0].Name, nil
```

src/pkg/cluster/tunnel.go
```diff
@@ -154,7 +155,7 @@ func (c *Cluster) ConnectToZarfRegistryEndpoint(ctx context.Context, registryInf
 	var tunnel *Tunnel
 	if registryInfo.IsInternal() {
 		// Establish a registry tunnel to send the images to the zarf registry
-		if tunnel, err = c.NewTunnel(state.ZarfNamespaceName, SvcResource, ZarfRegistryName, "", 0, ZarfRegistryPort); err != nil {
+		if tunnel, err = c.NewTunnel(state.ZarfNamespaceName, SvcResource, ZarfRegistryName, "", 0, ZarfRegistryPort, withServicePodSelector(selectRandomServicePod)); err != nil {
```
soltysh: Out of curiosity, why not do the randomization always?

src/pkg/cluster/tunnel.go
brandtkeller: Great point - cleaned this up some.

src/pkg/cluster/tunnel.go
brandtkeller: My brain was in a spot of wanting to introduce an option for randomizing while still allowing the original determinism. Given the scope of use and the safety I think it really does clean it up to always randomize.

soltysh (approved):
lgtm

PR #5321, opened 2026-09-06
feat: auto-detect terminal color support

### LINKED ISSUE #3311: Automatically detect and enable color if terminal supports it
Currently, users must specify if they want to disable terminal color (--no-color), or use a --log-format without one (e.g. `json`). Using https://github.com/jwalton/go-supportscolor or something similar (s/o to @phillebaba) we can check the current terminal's behavior and enable/disable color accordingly. Any user-supplied input from flags or env vars would take precedence in this case (using the usual config precedence).

Related https://github.com/zarf-dev/zarf/issues/3306

NAME-ASHWANIYADAV:
## Description

Color output was always on unless the user opted out, so piping zarf output or running in a terminal without color support still produced ANSI escape codes. This detects support at startup instead: color stays enabled only when stdout and stderr are both terminals, TERM is not "dumb", and the NO_COLOR convention (https://no-color.org) is not in use.

Explicit user choice still wins - the --no-color flag, ZARF_NO_COLOR, or a config file value all bypass detection, following the usual config precedence.

Per the guidance on the issue ("stick to 16 colors, implement the compatibility ourselves"), this adds no new dependency - the check is a small in-repo helper using golang.org/x/term, which was already vendored (promoted from indirect to direct in go.mod, vendor unchanged).

The --no-color flag's semantics and help text are unchanged, so no docs/schema regeneration is needed.

## How I validated

- Unit tests cover the decision table: interactive terminal, unset TERM  NO_COLOR set, empty NO_COLOR (must not count, per the spec), TERM=dumb, piped stdout, redirected stderr, and no terminal at all.
- The full ./src/cmd test suite passes; go vet is clean.
- go mod tidy && go mod vendor produce no further changes.

## Related Issue

Closes #3311

## Checklist before merging

- [x] Contributor Guide Steps followed


NAME-ASHWANIYADAV:
@brandtkeller will appreciate a review !!

src/cmd/root.go
```diff
@@ -91,6 +91,13 @@ func preRun(cmd *cobra.Command, _ []string) error {
 		return nil
 	}
 
+	// When the user has not chosen explicitly (via the --no-color flag,
+	// ZARF_NO_COLOR, or a config file), disable color if the terminal
+	// cannot render it.
+	if !IsColorDisabled && !cmd.Flags().Changed("no-color") && !getViper().IsSet(VNoColor) {
```
AustinAbro321, line 97: Small nit that this should also apply to the logo printed above

AustinAbro321 (changes requested):
One tiny nit and there are merge conflicts, but this looks like a good change

PR #5324, opened 2026-09-08
fix: replace github.com/defenseunicorns/pkg with local versions

### LINKED ISSUE #5278: Drop github.com/defenseunicorns/pkg and move oci bits to zarf
We currently rely on github.com/defenseunicorns/pkg. That repository has 2 elements we rely on:
1. github.com/defenseunicorns/pkg/helpers/v2 v2.0.4
2. github.com/defenseunicorns/pkg/oci v1.3.2

We should ensure that (2) moves all OCI related operations as a zarf internal library.

For (1) we need to carefully investigate what we're using and what is actually needed in zarf.

soltysh:
## Description

Replace github.com/defenseunicorns/pkg with local versions, which should allow us to drop that dependency. 

## Related Issue

Fixes #5278 

## Checklist before merging


src/pkg/oci/common.go
```diff
@@ -0,0 +1,181 @@
...
+	ref, err := registry.ParseReference(strings.TrimPrefix(url, helpers.OCIURLPrefix))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse OCI reference %q: %w", url, err)
+	}
+	defaultTransport, ok := http.DefaultTransport.(*http.Transport)
+	if !ok {
+		return nil, fmt.Errorf("http.DefaultTransport is not an *http.Transport")
+	}
+	client := &auth.Client{
+		Client: &http.Client{Transport: defaultTransport.Clone()},
```
AustinAbro321, line 122: Previously because we were using `helpers.NewTransport` we had a custom round tripper. This was partially for the progress bar, but it also used mimiced `retry.NewTransport(transport)` from `"oras.land/oras-go/v2/registry/remote/retry"`. I believe we should set our transport to this (we'll have to make sure this is the default for `WithTransport` and `withInsecureSkipTlsVerify` as well

src/pkg/helpers/url.go
```diff
@@ -0,0 +1,84 @@
...
+// IsValidHostName reports whether the local hostname is valid and not localhost.
+func IsValidHostName() bool {
+	hostname, err := os.Hostname()
+	if err != nil {
+		return false
+	}
+	// Explanation: https://regex101.com/r/zUGqjP/1/
+	return regexp.MustCompile(`^[a-zA-Z0-9\-.]+$`).MatchString(hostname) &&
+		// Explanation: https://regex101.com/r/vPGnzR/1/
+		!regexp.MustCompile(`\\.?localhost$`).MatchString(hostname)
```
AustinAbro321, line 69: Why was this changed from `\.?localhost$` which is current value in the helpers package ?

src/pkg/oci/common.go
```diff
@@ -0,0 +1,181 @@
...
+	"oras.land/oras-go/v2/registry/remote/auth"
+	"oras.land/oras-go/v2/registry/remote/credentials"
+)
+
+const (
+	// MultiOS is the OS used for multi-platform packages
+	MultiOS = "multi"
+)
+
+// OrasRemote is a wrapper around the Oras remote repository that includes a progress bar for interactive feedback.
```
AustinAbro321, line 28:
```suggestion
// OrasRemote is a wrapper around the Oras remote repository.
```

src/pkg/oci/common.go
```diff
@@ -0,0 +1,181 @@
...
+	root           *Manifest
+	targetPlatform *ocispec.Platform
+	insecure       *bool
+	log            *slog.Logger
+}
+
+// Modifier is a function that modifies an OrasRemote
+type Modifier func(*OrasRemote)
+
+// WithPlainHTTP sets the plain HTTP flag for the remote// WithPlainHTTP sets the plain HTTP flag for the remote
```
AustinAbro321, line 41:
```suggestion
// WithPlainHTTP sets the plain HTTP flag for the remote
```

AustinAbro321 (changes requested):
Behavior change in the transport but otherwise looks good

Issue #5325, opened 2026-09-08
zarf init fails on OpenShift/CRI-O 1.35+ (ROSA 4.22+): injector's HTTP seed registry rejected, registry-mode=proxy doesn't fix it


jsmith1988:
### Environment

- Zarf: v0.82.0 (also reproduced with the version pinned by `zarf init --registry-mode=proxy`, so this doesn't look version-specific on the Zarf side)
- Kubernetes distro: ROSA (Red Hat OpenShift Service on AWS), Hosted Control Plane
- Broken on: OpenShift 4.22.11 / CRI-O `1.35.6-5.rhaos4.22.git41f610b.el9`
- Works fine on: OpenShift 4.21.30 / CRI-O `1.34.11-4.rhaos4.21.git358c4b4.el9`

### What happens

`zarf init --features=registry-proxy=true --registry-mode=proxy ...` never completes. Both the internal `zarf-docker-registry` pod and every `zarf-registry-proxy` pod sit in `ImagePullBackOff`, both trying to pull from the injector's loopback endpoint:

```
Failed to pull image "127.0.0.1:5001/library/registry:3.1.1": unable to pull image:
initializing source docker://127.0.0.1:5001/library/registry:3.1.1: pinging container registry 127.0.0.1:5001:
Get "https://127.0.0.1:5001/v2/": http: server gave HTTP response to HTTPS client
```

```
Failed to pull image "127.0.0.1:5001/alpine/socat:1.8.0.3": unable to pull image:
initializing source docker://127.0.0.1:5001/alpine/socat:1.8.0.3: pinging container registry 127.0.0.1:5001:
Get "https://127.0.0.1:5001/v2/": http: server gave HTTP response to HTTPS client
```

I confirmed the injector itself is healthy and serving correctly the whole time. Curling `http://127.0.0.1:5001/v2/` from inside the node's actual network namespace returns a clean `200 OK` on both clusters, identically. The kubelet/CRI-O is the one insisting on HTTPS and refusing to fall back.

### Root cause

Tracked it down to source. CRI-O `1.34.11` unconditionally hardcodes a loopback exemption in its image-pull path:

https://github.com/cri-o/cri-o/blob/v1.34.11/internal/storage/image.go#L1089

```go
serverConfig.InsecureRegistries = append(serverConfig.InsecureRegistries, "127.0.0.0/8")
```

[...]

PR #5333, opened 2026-09-09
feat: debug log container runtime for diagnostics - warn on cri-o

brandtkeller:
## Description

See associated issue for more information on incompatibility - this adds a diagnostic for container runtime on nodes with a warning for the known versions of cri-o. 

There may be room for expanding diagnostics further so I went this direction while implementing the warning. Any future incompatibility problems would benefit from this information. 

Does not increase the surface area of the init command RBAC but deemed this something we perform as a preflight warning based on most isolated location with all required information for the activity. 

## Related Issue

Relates to #5325

## Checklist before merging


soltysh (approved):
lgtm

PR #5334, opened 2026-09-09
fix(values): infer integer and unknown schema types

### LINKED ISSUE #5332: Improve fallback type inference for generated values schemas

When a Helm chart does not provide a `values.schema.json`, `zarf dev generate-schema` infers YAML null values as strings and treats both integral and fractional values as numbers.

A null default provides no reliable type information, so inferring `string` can reject valid object, array, or boolean inputs.

- **Given** a mapped Helm value without an explicit chart schema
- **When** `zarf dev generate-schema` infers its schema
- **Then** null values remain unconstrained, integral values use `integer`, and fractional values use `number`

Chart authors can provide a `values.schema.json`, and package authors can manually correct generated schemas. Neither option addresses fallback inference for charts without schemas.

### Additional context

#5157 added support for explicit Helm schemas. This request covers the remaining fallback path when one is unavailable.

### Out of Scope

- Mixed-array inference

ntwkninja:
## Description
Improves fallback schema inference for charts without native schemas. Unknown/null values remain unconstrained, integral values infer as integer, and `--delete-not-found` removes stale structural types. Includes unit and E2E coverage.

Changes:
  - Infer unknown/null values as {} instead of string.
  - Infer integral values as integer; retain number for fractional values.
  - Remove stale structural constraints for unknown values with `--delete-not-found`.
  - Update unit and generate-schema E2E coverage.

## Related Issue

Fixes #5332 

## Checklist before merging

### DIFF
--- a/src/pkg/value/generate.go
+++ b/src/pkg/value/generate.go
@@ -5,6 +5,7 @@ package value
 
 import (
 	"fmt"
+	"math"
 )
 
 // GenerateJSONSchema infers a JSON schema from the structure and scalar types in values.
@@ -31,6 +32,10 @@ func ReconcileJSONSchema(existing, inferred map[string]any, deleteNotFound bool)
 	typeVal, hasType := inferred["type"]
 	if hasType {
 		existing["type"] = typeVal
+	} else if deleteNotFound {
+		delete(existing, "type")
+		delete(existing, "properties")
+		delete(existing, "items")
 	}
 
 	if schemaTypeIncludes(typeVal, "object") {
@@ -396,8 +401,12 @@ func inferSchemaType(v any) any {
 	switch val := v.(type) {
 	case string:
 		return map[string]any{"type": "string"}
-	case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64, float32, float64:
-		return map[string]any{"type": "number"}
+	case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64:
+		return map[string]any{"type": "integer"}
+	case float32:
+		return inferFloatSchema(float64(val))
+	case float64:
+		return inferFloatSchema(val)
 	case bool:
 		return map[string]any{"type": "boolean"}
 	case map[string]any:
@@ -415,6 +424,13 @@ func inferSchemaType(v any) any {
 		}
 		return map[string]any{"type": "array"}
 	default:
-		return map[string]any{"type": "string"}
+		return map[string]any{}
 	}
+}
+
+func inferFloatSchema(val float64) map[string]any {
+	if math.Trunc(val) == val {
+		return map[string]any{"type": "integer"}
+	}
+	return map[string]any{"type": "number"}
 }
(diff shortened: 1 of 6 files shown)


src/pkg/value/generate.go
```diff
@@ -31,6 +32,10 @@ func ReconcileJSONSchema(existing, inferred map[string]any, deleteNotFound bool)
 	typeVal, hasType := inferred["type"]
 	if hasType {
 		existing["type"] = typeVal
+	} else if deleteNotFound {
+		delete(existing, "type")
+		delete(existing, "properties")
```
brandtkeller: I feel like there may be a correctness gap here that I haven't dug deep enough around yet - in as much as whether we should delete properties and items here or whether to also check for their existence in the following checks for object/array and handle accordingly.

brandtkeller (dismissed):
I think this is general improvement and schema workflows are something we will probably continue iterating on as the evidence/cases build. 

Non-blocking comment on deletion semantics.

src/pkg/value/generate.go
ntwkninja: Yeah, I think that’s worth digging into. My initial reason for deleting all three here was that an inferred `{}` should probably be unconstrained, while retained properties or items could still constrain object or array inputs.

I wonder if the structural cleanup belongs alongside the existing type checks:
```
if schemaTypeIncludes(typeVal, "object") {
    reconcileSchemaProperties(existing, inferred, deleteNotFound)
} else if deleteNotFound {
    delete(existing, "properties")
}

if schemaTypeIncludes(typeVal, "array") {
    reconcileSchemaItems(existing, inferred, deleteNotFound)
} else if deleteNotFound {
    delete(existing, "items")
}
```

That appears to cover unknown values and concrete type changes, although empty objects and arrays would need an explicit decision. What do you think about adding table cases for unknown values, concrete type changes, and empty collections first, then using those to settle the structure in this PR?

brandtkeller: I'd be in support of that change!

brandtkeller: I wonder though if we should consider the same deletion behavior extend to object and array-speficic validation keywords that depend on that structure. Things like `required` `additionalProperties`, `minProperties`, `maxProperties`, `minItems`, and `maxItems`?  

I think that feels like something that I need more evidence and feedback around though. So I'm good with these changes.

brandtkeller (dismissed):
lgtm - thanks for the PR!

src/pkg/value/generate.go
```diff
@@ -415,6 +432,13 @@ func inferSchemaType(v any) any {
 		}
 		return map[string]any{"type": "array"}
 	default:
-		return map[string]any{"type": "string"}
+		return map[string]any{}
+	}
+}
+
+func inferFloatSchema(val float64) map[string]any {
+	if math.Trunc(val) == val {
```
AustinAbro321: If someone is putting a number like 3.0 it's more likely that that the decimal is relevant to the field. I doubt users are writing 3.0 for replicas, but it's easy to imagine someone putting 5.0 for a percentage or to represent cpu cores.

AustinAbro321 (changes requested):
Challenging float inference behavior, otherwise looks good

src/pkg/value/generate.go
ntwkninja: Thanks @AustinAbro321, that makes sense.

updated

brandtkeller (approved):

AustinAbro321 (approved):

PR #5344, opened 2026-09-11
fix(helm): handle charts that render resources inside list kinds

### LINKED ISSUE #5343: Post-renderer adds `metadata.labels` to `*List` documents, making them invalid (`unknown field "labels" in ListMeta`)
### Environment
Device and OS: N/A (but did troubleshooting both locally on macOS arm64 and Ubuntu amd64)
App version: `0.84.0`
Kubernetes distro being used: `1.37`

### Steps to reproduce
1. Create a minimal chart and package. Any chart whose templates render a list shaped document
   (`ConfigMapList`, `SecretList`, etc) will do:

   ```
   list-bug-repro/
     Chart.yaml
     templates/
       configmaps.yaml
   zarf.yaml
   ```

   ```yaml
   # list-bug-repro/Chart.yaml
   apiVersion: v2
   name: list-bug-repro
   version: 0.1.0
   ```

   ```yaml
   # list-bug-repro/templates/configmaps.yaml
   apiVersion: v1
   kind: ConfigMapList
   items:
     - apiVersion: v1
       kind: ConfigMap
       metadata:
         name: repro-one
       data:
         hello: world
     - apiVersion: v1
       kind: ConfigMap
       metadata:
         name: repro-two
       data:
         hello: world
   ```

   ```yaml
   # zarf.yaml
   kind: ZarfPackageConfig
   metadata:
     name: list-bug-repro
     version: 0.1.0
   components:
     - name: list-bug-repro
       required: true
       charts:
         - name: list-bug-repro
           version: 0.1.0
           namespace: list-bug-repro
           localPath: list-bug-repro
           releaseName: list-bug-repro
   ```

2. Build and deploy against a real cluster:

   ```sh
   zarf package create . --confirm --skip-sbom
   zarf package deploy zarf-package-list-bug-repro-*.tar.zst --confirm
   ```
3. For comparison, install the same chart without Zarf in the loop:

   ```sh
   helm install list-bug-repro ./list-bug-repro
   ```

### Expected result

Both ConfigMaps are created, which is what happens for the exact same chart with plain
`helm install` (step 3) and with `kubectl apply -f`. Helm flattens a list document into its
items client-side before applying, so a list-shaped manifest is a valid thing for a chart to
render.

### Actual Result

The deploy is rejected by schema validation and the release fails:

```
unable to install chart error validating "": error validating data:
ValidationError(ConfigMapList.metadata): unknown field "labels" in
io.k8s.apimachinery.pkg.apis.meta.v1.ListMeta: if you need to remove the
failed chart, use `zarf package remove`
```

What Zarf's post-renderer hands to Helm (invalid):

```yaml
apiVersion: v1
kind: ConfigMapList
metadata:
  labels:
    zarf.dev/package: list-bug-repro # illegal field on ListMeta
items:
  - apiVersion: v1
    kind: ConfigMap
    metadata:
      name: repro-one
    data:
      hello: world
```

### Severity/Priority

I'd say medium. It hard-fails the deploy with no workaround available inside Zarf. It seems that the only way around would be to fork/update the chart and use `---` separated yaml manifests instead of `kind: *Lists` lists.

### Additional Context

**Root cause.** `editHelmResources` adds the package label to every rendered document
unconditionally:

https://github.com/zarf-dev/zarf/blob/4e34f0eaedba5937e333c1cb5c6e160decac68b6/src/internal/packager/helm/post-render.go#L190-L194

That assumes every document's `metadata` is an `ObjectMeta`. A list document's metadata is a
`ListMeta`, which has no `labels`, no `annotations`.

idebeijer:
## Description

The post-renderer made two assumptions about every rendered document: that it holds a single resource, and that anything at `spec.template` is a pod template. Neither always holds, and each one fails the deploy of a chart that is valid for Helm and `kubectl` alike.

**It made the manifest invalid.** The zarf package label went onto the document's `metadata`, but a list's metadata is a `ListMeta`, which has no `labels` field, so the API server rejected the deploy before anything in the list was created:

```
ValidationError(ConfigMapList.metadata): unknown field "labels" in
io.k8s.apimachinery.pkg.apis.meta.v1.ListMeta
```

**A `Service` inside a list was never registered as a connect string**, since the document's kind is `ConfigMapList` or `List`, never `Service`.

**A namespace named only by list items was never tracked**, since `GetNamespace()` on the wrapper reads a `ListMeta` and returns `""`. Zarf creates namespaces before Helm applies, so a chart whose list items point at a namespace that doesn't exist yet failed outright:

```
server-side apply failed for object list-kinds-elsewhere/list-elsewhere
/v1, Kind=ConfigMap: namespaces "list-kinds-elsewhere" not found
```

**A `Namespace` inside a list stayed owned by Helm**, so it never got `AdoptZarfManagedLabels` and Helm deleted it on uninstall along with whatever else lived in it.

Rather than teach each of those to walk into items, `flattenHelmResources` replaces a list document with the resources it held before any of them run, so the rest of the post-renderer only ever sees one resource per manifest and none of it changes. Helm flattens a list into its items before applying it anyway, so the objects that reach the cluster are the same either way.

Detection is `obj.IsList()`, the same check Helm's kube client uses via `meta.IsListType()` when it flattens, so Zarf and Helm agree on which documents hold more than one resource.

### Which lists actually reach the post-renderer

Only typed ones. Before running a post-renderer, Helm merges the rendered files with `annotateAndMerge`, which pipes every document through `kio.ParseAll` (https://github.com/helm/helm/blob/v4.2.0/pkg/action/action.go#L206). That expands a generic `kind: List` into its items, so a chart using one never arrives here as a list. Typed list kinds (`ConfigMapList`, `SecretList`, etc) are left intact and are the ones that broke. The generic case is covered anyway, since that expansion is a kustomize dependency inside Helm rather than a documented contract.

## Related Issue

Fixes #5343

## Checklist before merging


src/internal/packager/helm/post-render.go
```diff
@@ -263,6 +248,50 @@ func (r *renderer) editHelmResources(ctx context.Context, resources []releaseuti
...
+// eachResource calls modifyFn once for every resource a rendered manifest document contains.
+// Most documents hold a single resource, but a list kind like ConfigMapList or the generic List
+// holds one per item.
+// The metadata on a list is a ListMeta, which has no labels field, so whatever gets written there
+// is rejected by the API server. Helm flattens a list into its items before applying it anyway,
+// so the items are what we actually want to edit.
+func eachResource(obj *unstructured.Unstructured, modifyFn func(*unstructured.Unstructured) error) error {
+	// IsList only checks for a top level items array, so every kubernetes list kind ending in List
+	// is checked as well to keep a custom resource carrying its own items from looking like a list.
+	if obj.IsList() && strings.HasSuffix(obj.GetKind(), "List") {
```
AustinAbro321: This strings check seems unreliable. Someone could have a CR that happens to end in List, but not actually be a list. I would generally expect a CRD to adhere to `obj.IsList()` do you have a particular example in mind?

src/internal/packager/helm/post-render.go
AustinAbro321, line 1: Similar to the problems you noticed with services, it seems like our namespace logic is skipped if a list if used

AustinAbro321 (changes requested):
Thank you for the PR! A few comments, but overall it's a reasonable change

src/internal/packager/helm/post-render.go
idebeijer: You're right, dropped it. Narrowing it here compared to what helm does might mean labeling a wrapper helm throws away and leaving the items in it that it actually applies unlabeled.

src/internal/packager/helm/post-render.go
idebeijer: Yea, I had the PR open for services and left namespaces as is because it had extra logic. But after a second look it's more straightforward than I first thought, I've included it in the other PR.

AustinAbro321: While I appreciate the inclination towards small PRs, we should include the logic in this PR so the feature doesn't exist on main in a bugged state. If the two PRs aren't merged in the same release we run into the possibility of a user seeing that this was merged, then trying out lists, only to run into the the namespace or service bug.

idebeijer: Fair enough

idebeijer: Changes of #5346 included here, and closed it.

idebeijer:
https://github.com/zarf-dev/zarf/pull/5344/commits/99735508dd2c6dfd1629061913a438d646936cc9 adds a fix to another issue I encountered, further described in the PR description under heading **spec.template is not always a pod template**.

src/internal/packager/helm/post-render.go
```diff
@@ -263,18 +229,114 @@ func (r *renderer) editHelmResources(ctx context.Context, resources []releaseuti
...
+		// if this is the first time seeing this ns, we need to track that to create it as well
+		r.namespaces[namespace] = cluster.NewZarfManagedNamespace(namespace)
+	}
+}
+
+// addLabelsToNestedPath adds package labels to a nested path in an unstructured object.
+// The path is walked by hand rather than with unstructured.NestedStringMap, which reports a value
+// of an unexpected type along the way as an error. A custom resource can hold anything at
+// spec.template, and something that is not a pod template simply has no labels to add here.
 func (r *renderer) addLabelsToNestedPath(obj *unstructured.Unstructured, path []string) error {
```
AustinAbro321, line 323: Could you make the label changes into a separate PR? I think I'd lean towards a solution of explicitly listing the kinds that Zarf adds this to, rather than automatically putting it there. The story around it is easier to describe if it's on all pods, and right now your logic wouldn't work for a job like so
```
kind: Job
metadata:
  name: null-template-labels
spec:
  template:
    metadata:
      labels: null
```

src/internal/packager/helm/post-render.go
```diff
@@ -228,33 +211,16 @@ func (r *renderer) editHelmResources(ctx context.Context, resources []releaseuti
...
 
-		namespace := rawData.GetNamespace()
-		if _, exists := r.namespaces[namespace]; !exists && namespace != "" {
-			// if this is the first time seeing this ns, we need to track that to create it as well
-			r.namespaces[namespace] = cluster.NewZarfManagedNamespace(namespace)
+		// connect strings and namespaces are read off each resource, which for a list kind means
+		// the items it holds rather than the list itself
+		if err := eachResource(rawData, func(obj *unstructured.Unstructured) error {
+			r.recordConnectString(ctx, obj)
+			r.trackNamespace(obj)
```
AustinAbro321: We track the namespace here, but it ends up being included in the final document passed to Helm, which means that when it gets removed during package removal. 

We should still exclude it from the final document so that doesn't happen. 

Could you add a list with a service and namespace to the e2e test and make sure that the logic is still correctly applied to them?

AustinAbro321: We end up with a problem here where namespaces in a NamespaceList get into the final manifest and therefore removed on package removal.

I think a cleaner solution would be to have a function like this called at the top of `EditHelmResources`. This way the rest of our logic doesn't have to worry about resource lists

```go
// flattenHelmResources expands list documents into their leaf resources before Zarf assigns ownership
func flattenHelmResources(resources []releaseutil.Manifest) ([]releaseutil.Manifest, error) {
	flattened := make([]releaseutil.Manifest, 0, len(resources))
	for _, resource := range resources {
		_, rawData, err := processManifestContent(resource.Content, nil)
		if err != nil {
			return nil, err
		}
		if len(rawData.Object) == 0 || !rawData.IsList() {
			flattened = append(flattened, resource)
			continue
		}

		err = flattenListResource(rawData, func(obj *unstructured.Unstructured) error {
			content, err := yaml.Marshal(obj.Object)
			if err != nil {
				return fmt.Errorf("failed to marshal list item: %w", err)
			}
			leaf := resource
			leaf.Content = string(content)
			flattened = append(flattened, leaf)
			return nil
		})
		if err != nil {
			return nil, err
		}
	}
	return flattened, nil
}

func flattenListResource(obj *unstructured.Unstructured, addResource func(*unstructured.Unstructured) error) error {
	if !obj.IsList() {
		return addResource(obj)
	}
	return obj.EachListItem(func(item runtime.Object) error {
		listItem, ok := item.(*unstructured.Unstructured)
		if !ok {
			return fmt.Errorf("unexpected list item type %T", item)
		}
		return flattenListResource(listItem, addResource)
	})
}
```

AustinAbro321 (changes requested):
A few more comments, appreciate the cycles

src/internal/packager/helm/post-render.go
idebeijer: Thanks, yea that's a bunch better.  Taken as suggested.

One side effect though. Helm tracks which file a document came from with a `postrenderer.helm.sh/postrender-filename` annotation it stamps before calling post-renderer. And on a list that lands on the wrapper rather than the items, so flattening drops it. This is `helm get manifest` on the e2e fixture:

```
# Source: generated-by-postrender-1.yaml           item of a ConfigMapList
# Source: generated-by-postrender-2.yaml           item of a ConfigMapList
# Source: list-kinds/templates/generic-list.yaml    item of a generic List
```

I think nothing downstream reads it, and the annotation never reaches the cluster. It is only that the generic list keeps its name, since `annotateAndMerge` (check PR description why this is) expanded that one before stamping, so two things that should look the same no longer do. If you want, I can copy the annotation onto each item while flattening if you would rather.

src/internal/packager/helm/post-render.go
idebeijer: Thanks for the constructive feedback. Taken it out of this PR and into draft PR #5360.

src/internal/packager/helm/post-render.go
AustinAbro321: Good callout. From my evaluation it's fine if the annotation is dropped. Looks easy enough to add, but I wouldn't want to have to update this if a future version of Helm to drop or change the annotation

AustinAbro321 (approved):
lgtm thanks!

src/internal/packager/helm/post-render.go
```diff
@@ -263,6 +267,60 @@ func (r *renderer) editHelmResources(ctx context.Context, resources []releaseuti
...
+			// the item inherits the manifest name zarf writes its own source comment from. helm
+			// tracks the original file with an annotation it puts on the wrapper, which the item
+			// does not carry, so helm files these under a generated name of its own
+			item := resource
+			item.Content = string(content)
+			flattened = append(flattened, item)
+			return nil
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to flatten %s: %w", rawData.GetKind(), err)
```
theZMC, line 303: Since all items in a given list will likely have the same kind, it might be beneficial to include the `metadata.name` in the error as well.

```suggestion
			return nil, fmt.Errorf("failed to flatten %s (%s): %w", rawData.GetKind(), rawData.GetName(), err)
```

theZMC: Actually, disregard. At this point `rawData` would only point to an item with `ListMeta` which does not include a name under `metadata`.

src/internal/packager/helm/post-render.go
```diff
@@ -263,6 +267,60 @@ func (r *renderer) editHelmResources(ctx context.Context, resources []releaseuti
 	return nil
 }
 
+// flattenHelmResources replaces every list document with the resources it holds, so that the rest
+// of the post renderer only ever sees one resource per manifest.
+// A list is a wrapper rather than a resource: its metadata is a ListMeta, which has no labels
+// field, so a label written there is rejected by the API server, and everything zarf keys off the
+// kind of a document misses what the list holds. Helm flattens a list into its items before
+// applying it anyway, so the items are what ends up in the cluster either way.
+func flattenHelmResources(resources []releaseutil.Manifest) ([]releaseutil.Manifest, error) {
```
brandtkeller, line 276: Looking at the two loops (in order) we have `flattenHelmResources` -> `editHelmResources` and each call `processManifestContent`. So unconditionally this now adds a full decode/encode round trip and the shape of each is pretty similar - negligible on small charts but a code smell.

brandtkeller (approved):
lgtm - I think we can iterate on optimizing as needed but this solves a real problem as is.

src/internal/packager/helm/post-render.go
idebeijer: Good point, I'll see if I can add optimize this later.

idebeijer:
> lgtm - I think we can iterate on optimizing as needed but this solves a real problem as is.

Thanks, I hadn't taken the time to optimize yet. I'll see if I still can in a later PR, because I did run a benchmark after your feedback and it wasn't great.

PR #5345, opened 2026-09-11
feat!: public operational type

AustinAbro321:
## Description

This changes the generic type to be the public operation type in accordance with this proposed change to ZEP 0048 https://github.com/zarf-dev/proposals/pull/69

Behaviorally all the fields are represented by the api.Package, but some pointer fields are lost so a component which didn't declare required, will show `required: false` during `zarf dev inspect definition`. We could have `api.Package` hold pointers to keep it, but I think being more explicit isn't bad anyway. 

I decided to change kustomize from a pointer to value type in the v1beta1 schema. Functionally it's easier to work with as a value, and while I initially thought having a pointer would come in handy so you can know if it's zero, practically speaking it's enough to just check if there's any files declared. 

## Breaking changes (SDK)

This PR introduces breaking changes for consumers of Zarf’s public Go packages. It does not change CLI behavior.

  ### Public operational package model

The public type api.PackageDefinition which held an internal representation of a Zarf package definition has become api.Package which is itself a public representation of a Zarf package. 

 | Before | After |
| --- | --- |
| `api.NewPackageDefinitionFromV1alpha1(pkg)` | `convert.PackageFromV1alpha1(pkg)` |
| `api.NewPackageDefinitionFromV1beta1(pkg)` | `convert.PackageFromV1beta1(pkg)` |
| `definition.AsV1alpha1()` | `convert.PackageToV1alpha1(definition)` |
| `definition.AsV1beta1()` | `convert.PackageToV1beta1(definition)` |
| `definition.OriginalAPIVersion()` | `definition.APIVersion` |

Some PackageDefinition setter methods were removed. Callers working directly with api.Package should update its fields directly. RemoveImages, RemoveRepositories, and OverrideNamespace remain available as api.Package methods.

  ### Package layout access

PackageLayout.PackageDefinition is no longer a public field. Use `definition := pkgLayout.Definition()`. `Definition()` returns the operational package by value.

Supported layout mutations are exposed through methods:

```go
  pkgLayout.SetName(name)
  pkgLayout.SetAnnotations(annotations)
  pkgLayout.SetChartNamespace(componentName, chartName, namespace)
  pkgLayout.RemoveImages()
  pkgLayout.RemoveRepositories()
  pkgLayout.OverrideNamespace(namespace)
  pkgLayout.Filter(filter)
```

  ### Filters

The filters package has changed again to hold the operational type instead of filters.PackageView type which has now been deleted.

```go
  type ComponentFilterStrategy interface {
      Apply(api.Package) ([]api.Component, error)
  }
````

## Checklist before merging

### CODE REVIEW

File: src/api/convert/convert_test.go [hunk 1]
```diff
@@ -249,12 +221,13 @@ func TestV1Alpha1PkgToV1Beta1_RequiredOptionalShift(t *testing.T) {
...
 		required     *bool
 		wantOptional bool
+		wantRequired bool
 	}{
 		// v1alpha1 defaults an unset required to optional (the inverse of v1beta1's required default),
 		// so a component that omits it must become Optional=true rather than v1beta1's zero value.
-		{name: "unset required is optional", required: nil, wantOptional: true},
-		{name: "explicit false is optional", required: b(false), wantOptional: true},
-		{name: "explicit true is required", required: b(true), wantOptional: false},
+		{name: "unset required is optional", required: nil, wantOptional: true, wantRequired: false},
```
brandtkeller, line 228: Reading the proposal I think this case (and logic) asserts the opposite of what is intended? 

> .components[x].required will be renamed to .components[x].optional. optional will default to false. Since required currently defaults to false, components will now default to being required.

AustinAbro321: v1beta1 components are now  required by default since `optional` unset is false. Existing v1alpha1 components that don't set `required` continue to be optional as we don't want any breaking behavioral changes to v1alpha1

brandtkeller: Ah good distinction. Agree on semantics, initially my brain wants to think about the migration logic and handling this for those converting their packages to v1beta1 (versus authoring a new package) but I imagine this stance is less surprising.

src/pkg/packager/assemble/assemble.go
```diff
@@ -248,8 +248,8 @@ func AssembleSkeleton(ctx context.Context, resolvedPackage *load.ResolvedPackage
 		return nil, err
 	}
 	definition := resolvedPackage.Definition
-	definition.SetMetadataArchitecture(v1alpha1.SkeletonArch)
-	pkg := definition.AsV1alpha1()
+	definition.Metadata.Architecture = v1alpha1.SkeletonArch
+	pkg := convert.PackageToV1alpha1(definition)
```
brandtkeller: Both `packager.PublishSkeleton` and here project the definition to v1alpha1 - I wonder if we should instead be gating this behavior to not allow v1beta1 to begin with and instead serializing into the only supported skeleton format?

src/internal/api/v1alpha1/convert.go
```diff
@@ -136,63 +137,106 @@ func componentToGeneric(c v1alpha1.ZarfComponent) types.Component {
...
 	}
+	chartSourceToGeneric(&gc, ch)
 	return gc
 }
 
+// chartSourceToGeneric projects v1alpha1's flat source fields into the single
+// structured source used by the operational model.
+func chartSourceToGeneric(chart *api.Chart, source v1alpha1.ZarfChart) {
+	switch {
+	case isGitURL(source.URL):
```
brandtkeller: funny enough I can find examples of `GitPath` used with a non ".git" suffix url (which would be inert anyway) - so another clear advantage to the v1beta1 structure.

src/pkg/packager/assemble/assemble.go
brandtkeller, line 256: Both packager.PublishSkeleton and here project the definition to v1alpha1 - I wonder if we should instead be gating this behavior to not allow v1beta1 to begin with and instead serializing into the only supported skeleton format?

src/internal/api/v1alpha1/convert.go
brandtkeller, line 178: Interestingly I can find examples of non-".git" suffix urls with `gitPath` set (which would be inert). So more evidence for this trending in the right direction.

src/pkg/packager/assemble/assemble.go
AustinAbro321: Great idea, added a gate for this. 

I do think the converters are useful here as it avoids duplication with assembling regular packages

AGENTS.md
```diff
@@ -7,7 +7,7 @@ Zarf is a Go CLI (`main.go`, `go.mod`). Application code under `src/`. Unit test
 ## Build, Test, and Development Commands
 
 - `make build` tidies, vendors, and builds the CLI.
-- `make test-unit` runs tests with race detection and atomic coverage; `make test-unit-quick` skips both.
+- `make test-unit` runs tests with race detection and atomic coverage; `make test-unit-quick` skips both. Never run `go test .src/...` for unit tests as this will run the full e2e suite.
```
brandtkeller, line 10: ++

brandtkeller (approved):
lgtm - It's clear how much this simplifies translation and I imagine when the subsequent PR's are opened for use of the operational type natively that it will additionally clean up a lot of current implications.

PR #5346, opened 2026-09-11
fix(helm): inspect list items for connect strings and namespaces

idebeijer:
> [!NOTE]
> Depends on #5344. That PR's two commits are included in the diff below until it merges;
> only the last commit belongs to this one.

## Description

Zarf inspected only the top-level manifest document, so resources inside a typed list were invisible to two checks. A `Service` carrying `zarf.dev/connect-name` never registered a connect string, and a namespace named only by list items was never tracked. The second is a hard failure, since Zarf creates namespaces before Helm applies:

```
server-side apply failed for object list-kinds-elsewhere/list-elsewhere
/v1, Kind=ConfigMap: namespaces "list-kinds-elsewhere" not found
```

Both now run through the `eachResource` helper from #5344, which walks a list's items. Neither is new behaviour; it's what a resource in its own document already gets, including namespace adoption under `takeOwnership`.

The Namespace kind case stays document-level. It works by dropping the whole document so Zarf can create the namespace out of band, and there's no per-item equivalent of that.

### Not included

Namespace tracking has the same blind spot: `rawData.GetNamespace()` reads the wrapper's
`ListMeta` and returns `""`, so a namespace referenced only by list items is never tracked. That
one isn't as simple, because `r.namespaces` feeds `adoptAndUpdateNamespaces`, which creates
namespaces and adopts existing ones under `takeOwnership`, so fixing it changes who owns those
namespaces.

## Related Issue

Relates to #5343

## Checklist before merging

### DIFF
--- a/src/internal/packager/helm/post-render.go
+++ b/src/internal/packager/helm/post-render.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"fmt"
 	"slices"
+	"strings"
 
 	"github.com/zarf-dev/zarf/src/pkg/state"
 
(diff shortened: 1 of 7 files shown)


idebeijer:
Change included in #5344

PR #5347, opened 2026-09-12
feat: add init chart values schemas

### LINKED ISSUE #5309: Add chart values schemas to internal init charts

For general good practice and preparation towards shifting from variables to [Package Values](https://docs.zarf.dev/ref/package-values/), an improvement could be made by generating the [schemas](https://helm.sh/docs/topics/charts/#schema-files) for the values in the charts we maintain for the zarf init package. 

This would specifically add a `values.schema.json` to the `packages/zarf-registry/chart` and `packages/zarf-agent/chart` charts.  

### Additional context

Consider values that might be enums and type them appropriately. New values have been added recently such as `mutationExclusions` in the agent chart that would benefit from these schemas existing.

acald-creator:
## Description

- Add `values.schema.json` files to the internal `zarf-agent` and `zarf-registry` Helm charts.
- Validate chart values against the schemas so invalid configuration is caught during Helm validation.

## Validation

- Ran `helm lint` for both charts.
- Confirmed valid values pass and invalid values are rejected.

## Related Issue

Fixes #5309

## Checklist before merging

### DIFF
--- a/packages/zarf-agent/chart/values.schema.json
+++ b/packages/zarf-agent/chart/values.schema.json
@@ -1,0 +1,160 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "serviceAccount": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string"
+        }
+      }
+    },
+    "clusterRole": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string"
+        }
+      }
+    },
+    "clusterRoleBinding": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string"
+        }
+      }
+    },
+    "role": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string"
+        }
+      }
+    },
+    "roleBinding": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string"
+        }
+      }
+    },
+    "secret": {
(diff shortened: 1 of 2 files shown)


packages/zarf-agent/chart/values.schema.json
```diff
@@ -0,0 +1,169 @@
...
+    "resources": {
+      "type": "object",
+      "properties": {
+        "requests": {
+          "type": "object",
+          "properties": {
+            "memory": {
+              "type": "string"
+            },
+            "cpu": {
```
brandtkeller, line 101: possibly negligible in current utility but standard practice for [Quantity](https://kubernetes.io/docs/reference/kubernetes-api/definitions/quantity-resource/) would benefit from a `["string", "number"]`

brandtkeller (changes requested):
Generally looks good! minor nit on quantity types. I appreciate the enums. I think some of these could be hardened further with patterns but wouldn't require it here. 

Could we also patch bump these charts for visibility?

brandtkeller (approved):
lgtm - thanks for this improvement!

a1994sc (approved):

a1994sc:
LGTM

brandtkeller:
ah! @acald-creator we have signed commits enforced - could you re-sign your commits? 

Provided you have this configured it usually looks something like:
```bash
git fetch origin
git rebase --exec 'git commit --amend --no-edit -S -s' origin/main
git push --force-with-lease
```

Issue #5274, opened 2026-08-25
Verify remote components with `zarf component verify`


AustinAbro321:
### Is your feature request related to a problem? Please describe.
User's should be able to verify signed v1beta1 ZarfComponentConfigs with `zarf component verify`. 

### Describe the behavior you'd like

- **Given** I have a signed ZarfComponentConfig
- **When** I run `zarf component verify oci://ghcr.io/my-repo/my-component --key` with the correct key
- **Then** I  receive an assertion that my component is verified

All verification flags should work

### Describe alternatives you've considered
In the future we may have verification options in the package schema itself, but that will be a separate option

PR #5358, opened 2026-09-14
feat: component sign and verify

### LINKED ISSUE #5279: New command - Zarf component sign
A new command `zarf component sign` should be introduced. It should sign the manifest rather than signing the zarf.yaml file since remote components do not need local verification. 

All signing options on `zarf package sign` should be present on `zarf component sign`. It will only take an already published source.

- **Given** I have a component ghcr.io/austinabro321/my-component:1.0
- **When** I run `zarf component sign ghcr.io/austinabro321/my-component:1.0 --signing-key=my-key-file.key`  
- **Then** My package is signed

AustinAbro321:
## Description

Implements component sign and verify

## Related Issue

Fixes #5279, #5274

## Checklist before merging


src/pkg/signing/cosign.go
```diff
@@ -236,6 +239,162 @@ func CosignSignBlobWithOptions(ctx context.Context, blobPath string, opts SignBl
...
+	if opts.Timeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, opts.Timeout)
+		defer cancel()
+	}
+	if err := verifyCmd.Exec(ctx, []string{manifestRef}); err != nil {
+		return err
+	}
+
+	l.Debug("OCI manifest signature verified successfully", "reference", manifestRef)
```
brandtkeller, line 436: A neat detail I hadn't realized - the implementation iterates over all bundles against the manifest. So you can sign a component in OCI with both a key and a keyless method and use either to verify. As we think about package OCI layouts this feels like a signal to look into further.

src/pkg/signing/cosign.go
```diff
@@ -236,6 +239,162 @@ func CosignSignBlobWithOptions(ctx context.Context, blobPath string, opts SignBl
...
+	if opts.PassFunc != nil {
+		ko.PassFunc = opts.PassFunc
+	} else if opts.Password != "" {
+		password := opts.Password
+		ko.PassFunc = cosign.PassFunc(func(_ bool) ([]byte, error) {
+			return []byte(password), nil
+		})
+	}
+
+	if err := opts.CheckOverwrite(ctx); err != nil {
```
brandtkeller: I signed a component twice - which works and is the expectation of Cosign - and then noticed this line. This operation isn't writing a `BundlePath` so I believe this is a no-op.

src/pkg/signing/cosign.go
```diff
@@ -236,6 +239,162 @@ func CosignSignBlobWithOptions(ctx context.Context, blobPath string, opts SignBl
...
+		return fmt.Errorf("parsing OCI manifest reference: %w", err)
+	}
+	descriptor, err := remote.Head(ref, registryOptions.GetRegistryClientOpts(ctx)...)
+	if err != nil {
+		return fmt.Errorf("resolving OCI manifest: %w", err)
+	}
+	manifestRef = ref.Context().Digest(descriptor.Digest.String()).String()
+
+	l.Debug("signing OCI manifest with cosign referrer", "reference", manifestRef, "key", opts.Key, "sk", opts.SecurityKey.Use)
+	rootOpts := &options.RootOptions{Verbose: opts.Verbose, Timeout: opts.Timeout}
```
brandtkeller: How do you feel about being explicit on timeout and registering a timeout on the context ourselves? most things out front of the signing should be trivial but there is a fair amount going on.

src/config/lang/english.go
```diff
@@ -436,6 +436,23 @@ $ zarf package publish zarf-package-my-app-amd64-1.0.0.tar.zst oci://my-registry
...
+$ zarf component sign ghcr.io/my-org/my-component:1.0 --keyless --confirm
+`
+	CmdComponentVerifyShort   = "Verifies a published Zarf component signature"
+	CmdComponentVerifyLong    = "Verifies the Sigstore signature stored as an OCI referrer of a published Zarf component manifest."
+	CmdComponentVerifyExample = `
+# Verify a published component with a public key
+$ zarf component verify ghcr.io/my-org/my-component:1.0 --key ./public-key.pem
+
+# Verify a keyless signature by its certificate identity
+$ zarf component verify ghcr.io/my-org/my-component:1.0 --certificate-identity signer@example.com
```
brandtkeller: nit: missing `--certificate-oidc-issuer`

brandtkeller (changes requested):
General approach looks great. Succumbs to the same cosign outputs problem we have but that should be resolved shortly. The keyless UX here is a little difficult to manage - as when you sign it isn't incredibly clear what identity/issuer to use is. We could mitigate that by also writing the bundle to a tempdir (IE setting `BundlePath`) - but I'd also accept this after the minor adjustments as-is and we can make that UX better in the refactor as we'll be handling bundle bytes and referrers upload ourselves.

src/pkg/signing/cosign.go
AustinAbro321: Seems reasonable, added to the other functions accepting the timeout as well

src/pkg/signing/cosign.go
AustinAbro321: Yeah you're right, and this has convinced me that these functions should take a different struct as to not provide unusable options. Created a SignManifestOptions and VerifyManifest Options. LMK what you think.

brandtkeller (approved):
lgtm

PR #5359, opened 2026-09-14
fix: differing Windows `files` basenames on Linux and Windows

Racer159:
## Description

This is a small fix for `files` targets using Windows-style paths, such as `C:\app\file.txt`. Zarf uses the `target`’s basename to derive the file’s internal package location, but `filepath.Base` interprets separators according to the create host. Windows derives `file.txt` for the filename, while Linux derives the entire `C:\app\file.txt`, producing a host-dependent package layout and a tar entry that Zarf’s archive validation rejects during extraction. The fix treats both slash forms as separators when deriving the internal basename.

## Related Issue

Fixes #N/A

## Checklist before merging


AustinAbro321 (approved):
lgtm

brandtkeller (approved):
lgtm

PR #5360, opened 2026-09-15
fix(helm): stop failing on resources that are not pod templates

idebeijer:
## Description

Split out of #5344.

The post-renderer added the package label to whatever sat at `spec.template.metadata.labels`. Nothing obliges a resource to keep a pod template there, and one holding a config string failed the deploy of a chart that is valid for Helm and `kubectl` alike:

```
failed to add labels to pod template: .spec.template.metadata accessor error:
<the value> is of the type string, expected map[string]interface{}
```

The cause is `unstructured.NestedStringMap`, which reports a type mismatch partway down a path as a caller error. For this it isn't one: the existing `!found` branch already treats "no pod template here" as fine, and a resource that keeps something else at `spec.template` is the same answer arrived at differently.

`ensureLabelsAt` walks the path itself and treats any shape that isn't a label map the same as a missing one. Nothing is labeled, nothing fails, and the resource is left as the chart wrote it. A malformed resource is then the API server's to report, naming the resource, rather than Zarf's to fail the whole deploy over.

Two consequences beyond the crash:

- **A `CronJob` is labeled, which it never was.** Its pod template is at
  `spec.jobTemplate.spec.template`, so the single hardcoded path missed it entirely.
  `podTemplatePathsByKind` carries that one kind — the only one whose template isn't at
  `spec.template`.
- **A pod template with no labels of its own is labeled.** The old code only ever added to labels
  that were already there, so `metadata: {}` under `spec.template` was silently skipped.

Kinds are deliberately not enumerated. An allow-list would have to grow with every operator that ships a workload — an Argo `Rollout`, an OpenShift `DeploymentConfig`, a KubeVirt VM all keep a real pod template at `spec.template` and all get the label their pods can be found by.

### Labels written as null

Separate and smaller, included because it is the same accessor and one line of the same fix.

```yaml
kind: Job
spec:
  template:
    metadata:
      labels: null
```

That is valid yaml and parses to nil, which `NestedStringMap` reports as a type error rather than as an absent value, so this failed the deploy too. `ensureLabelsAt` creates the map instead.

`addAgentIgnoreLabels` had the identical bug three lines further down, on the same resource, and now shares the helper. One behaviour change falls out of that: where it used to write `spec.template.metadata.labels` into a resource that had no `spec` at all, it now leaves it alone, for the same reason as above.

## Related Issue

Relates to #5344

## Checklist before merging


src/internal/packager/helm/post-render.go
```diff
@@ -321,23 +321,73 @@ func flattenListResource(obj *unstructured.Unstructured, addResource func(*unstr
...
-	if err != nil {
-		return err
-	} else if !found {
-		// Path doesn't exist, nothing to do
+// podTemplatePath is where a workload keeps the labels of the pod template it creates pods from.
+var podTemplatePath = []string{"spec", "template", "metadata", "labels"}
+
+// podTemplatePathsByKind holds the kinds that keep their pod template somewhere else, so a kind
+// is listed only when listing it changes the answer.
+var podTemplatePathsByKind = map[schema.GroupKind][]string{
```
AustinAbro321, line 329: Rather than only doing this for pod cron jobs, let's make this a map all resources that create pods, only only labels those.  It is a breaking change as we will no longer add labels to CRDs with `.spec.template.metadata.labels`, but I think this is better. 

Really the intent around the original code was to label pods, and I believe switching it to this will make it more logical.

AustinAbro321 (changes requested):
One comment, thanks for the PR

src/internal/packager/helm/post-render.go
idebeijer: I indeed had it sort of like that until db51a4a498211262cdab1103ba22559aba219a3b, but opted for keeping `.spec.template.metadata.labels` as a sort of default since out of the native k8s resources that use pod templates only one (CronJob) has the template at a different path than the rest.

But I'm happy to introduce the map with all resources again without the default fallback (and thus introducing the breaking change), but before I do: do you want me to include only native k8s resources in the map, or also e.g. a custom resource like Argo's Rollout? Since Zarf has more support around Argo it would make sense to include it, just wondering what you think.

Issue #4669, opened 2026-03-03
"filesystem: mkdir /var/lib/registry/docker: permission denied" during zarf init


unknown0816:
Hi,
I have a rocky linux with a rke2 installed and running. During zarf installation I configure to use my hostpath pv, but during installation following error occurs:

2026-03-03 10:27:17 INF starting deploy package=init
2026-03-03 10:27:17 INF deploying component name=zarf-injector
2026-03-03 10:27:17 INF copying files count=1
2026-03-03 10:27:17 INF loading file name=###ZARF_TEMP###/zarf-injector
2026-03-03 10:27:17 INF waiting for cluster connection
2026-03-03 10:27:18 INF creating Zarf injector resources
2026-03-03 10:27:19 INF adding archived binary configmaps of registry image to the cluster
2026-03-03 10:27:29 INF deploying component name=zarf-seed-registry
2026-03-03 10:27:29 INF processing Helm chart name=docker-registry version=1.0.1 source=Zarf-generated
2026-03-03 10:27:29 INF performing Helm upgrade chart=docker-registry
2026-03-03 10:27:32 INF running health checks chart=docker-registry
2026-03-03 10:27:41 INF deploying component name=zarf-registry
2026-03-03 10:27:41 INF pushing image name=docker.io/alpine/socat:1.8.0.3
2026-03-03 10:27:46 INF image push in progress name=docker.io/alpine/socat:1.8.0.3 complete=0.0% remaining=4.41 MBs
2026-03-03 10:27:54 INF image push in progress name=docker.io/alpine/socat:1.8.0.3 complete=0.0% remaining=4.41 MBs
2026-03-03 10:27:57 INF pushing image name=docker.io/library/registry:3.0.0
2026-03-03 10:28:02 INF image push in progress name=docker.io/library/registry:3.0.0 complete=0.0% remaining=18.56 MBs
2026-03-03 10:28:10 INF image push in progress name=docker.io/library/registry:3.0.0 complete=0.0% remaining=18.56 MBs
2026-03-03 10:28:13 INF pushing image name=docker.io/library/registry:3.0.0
2026-03-03 10:28:18 INF image push in progress name=docker.io/library/registry:3.0.0 complete=0.0% remaining=18.56 MBs
2026-03-03 10:28:26 INF image push in progress name=docker.io/library/registry:3.0.0 complete=0.0% remaining=18.56 MBs
[...]

PR #5369, opened 2026-09-17
fix: Add in templating to allow setting SELinux MCS level on zarf-registry

sei-pearl:
## Description

When deploying the internal Zarf registry on hosts with SELinux enforcing, permission errors can arise when multiple registry pods are created. Without specifying an SELinux MCS level, the container runtime will assign a random value to the container's securityContext. When a new pod is created, shared volume mount's directory is recursively relabeled to support this new level which breaks other pods and causes the SELinux denials. This PR adds a section to the registry's chart to enable setting a SELinux MCS level to prevent random level assignment. 

I added a new variable to the registry's zarf.yaml which defaults to setting level to an empty string.

## Related Issue

Relates to #4669

## Checklist before merging


packages/zarf-registry/chart/templates/deployment.yaml
```diff
@@ -73,6 +73,10 @@ spec:
             runAsNonRoot: true
             capabilities:
               drop: ["ALL"]
+{{- if .Values.securityContext.seLinuxOptions.level }}
```
brandtkeller: We're mixing pod security context scope with container security context scope here (via the values). Given this is a single-container pod, adding this to the pod security context options should produce the same effect?

packages/zarf-registry/chart/values.yaml
```diff
@@ -40,6 +40,8 @@ securityContext:
   runAsUser: 1000
   fsGroup: 2000
   runAsGroup: 2000
+  seLinuxOptions:
```
brandtkeller, line 43: For sake of the values schemas we are working in another PR - is this an enum? and if so what options are available?

brandtkeller (changes requested):
Minor comment - otherwise are you aware of any considerations required for upgrades and these labels? Anything we might need to include in a release note or documentation.

packages/zarf-registry/chart/templates/deployment.yaml
sei-pearl: Yeah makes sense to move it to the pod level since that's where the other values items go and SELinuxOptions is supported at both levels.

packages/zarf-registry/chart/values.yaml
sei-pearl: https://kubernetes.io/docs/reference/kubernetes-api/core/pod-v1/#SELinuxOptions

Just set more SELinux related roles. I've only used the type to set a pod to `spc_t` as a last ditch effort to work around SELinux issues. If its a want to make all these settable, let me know which approach you'd like me to take with changing the values and zarf definition.

sei-pearl:
> Minor comment - otherwise are you aware of any considerations required for upgrades and these labels? Anything we might need to include in a release note or documentation.

I don't think this change will affect anyone's current deployment. From the k8s docs:
> If unspecified, the container runtime will allocate a random SELinux context for each container. 
https://kubernetes.io/docs/reference/kubernetes-api/core/pod-v1/#SecurityContext 

Documentation wise, it could be useful to point this option out if running on an OS with SELinux and uses host backed storage options (ceph, longhorn, local-path, etc.)

PR #5376, opened 2026-09-19
fix(find-images): read the resources a list holds

### LINKED ISSUE #5375: `zarf dev find-images` misses resources rendered inside list kinds
### Environment
Device and OS: N/A (but did troubleshooting both locally on macOS arm64 and Ubuntu amd64)
App version: `0.86.0`

### Description

A chart or manifest that renders resources inside a list kind (`kind: DeploymentList`, `kind: List`, etc) is scanned as one opaque document instead of as the resources it holds. `processUnstructuredImages` switches on the document's own kind ([`find_images.go:491`](https://github.com/zarf-dev/zarf/blob/main/src/pkg/packager/find_images.go#L491)), a list matches none of the workload cases, and it falls through to the regex `default` branch. The items are never read as the workloads they are.

Same blind spot as #5343, in a different path.

### Steps to reproduce

Put this in `./listed`, and the same Deployment unwrapped as its own document in `./plain` both with a `zarf.yaml` using the manifest as component:

```yaml
# listed.yaml
apiVersion: v1
kind: DeploymentList
items:
  - apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: in-list
    spec:
      template:
        spec:
          containers:
            - name: app
              image: ghcr.io/zarf-dev/zarf/agent:v0.68.1
          initContainers:
            - name: init
              image: docker.io/library/alpine:latest
```

```yaml
# plain.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: in-list
spec:
  template:
    spec:
      containers:
        - name: app
          image: ghcr.io/zarf-dev/zarf/agent:v0.68.1
      initContainers:
        - name: init
          image: docker.io/library/alpine:latest
```

```sh
zarf dev find-images ./plain  --skip-cosign
zarf dev find-images ./listed --skip-cosign
```

### Expected result

Both print the same two images.

### Actual Result

```yaml
# ./plain
components:
  - name: baseline
    images:
      - docker.io/library/alpine:latest
      - ghcr.io/zarf-dev/zarf/agent:v0.68.1

# ./listed
components:
  - name: baseline
    images:
      # Possible images - baseline
      - docker.io/library/alpine:latest
      - ghcr.io/zarf-dev/zarf/agent:v0.68.1
```

### One more consequences of the same cause

**An `OCIRepository` inside a list is missed entirely** Flux's reference is assembled from `spec.url` and `spec.ref.tag` by `appendToImageMapOCIRepo`, which only runs in the typed `case "OCIRepository"`. No regex looks at `url`:

### Severity/Priority

Low. Trigger is narrow as it's not very common for charts to use *List kinds, let alone *List kinds for Deployments, StatefulSets or even more unlikely, Flux's OCIRepository.

idebeijer:
## Description

A document holding a list was scanned as the one resource it appears to be, so a workload inside it never reached the typed case written for it. Its images came back as possible rather than confirmed, an `OCIRepository` inside a list was missed entirely, a Helm test resource inside one was not filtered out.

`flattenListResources` replaces a list with the resources it held, right after `utils.SplitYAML` and before anything reads a document's kind or annotations, so everything downstream is unchanged. Detection is `obj.IsList()`, same as the post-renderer in #5344.

Ordering matters: expanding after the Helm test filter would leave that bypass in place and promote the test image from possible to confirmed.

## Testing

`TestFindImagesInListKinds` covers a `DeploymentList`, an `OCIRepositoryList` and a `List` holding a Helm test resource. Two smaller tests check a resource scans the same inside a list as outside one.
## Related Issue

Fixes #5375

## Checklist before merging


src/pkg/packager/find_images.go
```diff
@@ -345,6 +346,7 @@ func findImages(ctx context.Context, pkg v1alpha1.ZarfPackage, resourceSet *load
 				if err != nil {
 					return nil, err
 				}
+				yamls = flattenListResources(yamls)
```
brandtkeller, line 349: nit: I agree with your mention on order but I might advocate to combine these processes (flatten + helm test check) together for two reasons - first being that the list itself is never checked and second that this pattern could encourage additional filters/steps as opposed to isolating the process that produces individual items.

brandtkeller (changes requested):
Minor nit otherwise appreciative of the fix.

PR #5377, opened 2026-09-19
feat(cmd): migrate `zarf tools registry digest` from crane to oras-go

abhishekgit03:
Related to #5003

Second command in the phased migration (`list` landed in #5199). This one covers `digest`.

### What changed

- New `src/cmd/registry_digest.go` reimplements `digest` using `oras-go` instead of `craneCmd.NewCmdDigest`, preserving the same flags (`--tarball`, `--full-ref`), example text, and validation (`--full-ref` can't combine with `--tarball`).
- Extracted the auth/tunnel/scheme-resolution plumbing shared between `list` and `digest` into a new `src/cmd/registry_common.go` (`registryConnection`, `setupRegistryAuth`, `registryHost`, `resolveConnPlainHTTP`), so `registry_list.go` and `registry_digest.go` each stay focused on their own command. More commands are coming in this series and will reuse the same helpers.
- `src/cmd/crane.go`: one line swapped.
- Same `--plain-http`/`--insecure-skip-tls-verify` flags with `--insecure` deprecated, following the pattern from #5199.
- Tarball mode (`--tarball PATH`, computing a digest from a local OCI tarball with no network involved) is kept as-is, using `go-containerregistry/pkg/v1/tarball` directly - this is unrelated to the crane-vs-oras registry-client concern the migration is about.

### Test plan

- [x] Unit tests in `src/cmd/registry_digest_test.go`: digest resolution (bare + `--full-ref`), tarball-mode digest (including zero-tag-arg), invalid-ref error paths, and the same known-plain-HTTP regression test pattern
- [x] `make docs-and-schema` - diff matches the expected pattern (new local flags, `--insecure`/`--plain-http`/`--insecure-skip-tls-verify` no longer shown as inherited)
- [x] Live end-to-end: remote digest resolution against a public registry (cross-checked against Docker's own reported digest for `alpine:latest`) and against Zarf's tunneled internal registry; tarball-mode digest cross-checked byte-for-byte against the real released Crane-based binary using the same tarball
- [x] `go build ./...`, `golangci-lint run ./...`, `pre-commit run --all-files`


site/src/content/docs/commands/zarf_tools_registry_digest.md
```diff
@@ -29,19 +29,18 @@ $ zarf tools registry digest reg.example.com/stefanprodan/podinfo:6.4.0
...
 
 ### Options inherited from parent commands
 
 ```
       --allow-nondistributable-artifacts   Allow pushing non-distributable (foreign) layers
       --features stringToString            Provide a comma-separated list of feature names to bools to enable or disable. Ex. --features "foo=true,bar=false,baz=true" (default [])
-      --insecure                           Allow image references to be fetched without TLS
-      --insecure-skip-tls-verify           Skip checking server's certificate for validity. This flag should only be used if you have a specific reason and accept the reduced security posture.
-      --plain-http                         Allow OCI registry connections over HTTP instead of HTTPS. This flag should only be used if you have a specific reason and accept the reduced security posture.
       --platform string                    Specifies the platform in the form os/arch[/variant][:osversion] (e.g. linux/amd64). (default "all")
```
AustinAbro321, line 44: It seems the platform flag is no longer respected

src/cmd/registry_digest.go
```diff
@@ -0,0 +1,143 @@
...
+			if deprecatedInsecure {
+				plainHTTP = true
+				insecureSkipTLSVerify = true
+			}
+			var imageRef string
+			if len(args) > 0 {
+				imageRef = args[0]
+			}
+			if tarballPath != "" {
+				return tarballDigest(cmd.OutOrStdout(), tarballPath, imageRef)
```
AustinAbro321, line 52: Part of the goal of this refactor is to get away from Crane as a direct dependency, however the tarball digest function uses Crane directly and so is not sufficient.

Instead this should also be implemented in house using oras-go. I think it's reasonable to expect an oci-layout. If the image is not an oci-layout then we can say that `zarf tools registry digest` does not support it. This is a breaking change so the PR title should start with `feat!:`.

AustinAbro321 (changes requested):
Good start, found two issues

site/src/content/docs/commands/zarf_tools_registry_digest.md
abhishekgit03: fixed, --platform now works.

src/cmd/registry_digest.go
abhishekgit03: got it, i have made the changes, --tarball now reads an OCI image layout tar via oras-go instead of crane.

PR #5379, opened 2026-09-21
refactor!: switch to api.Package from v1alpha1.ZarfPackage throughout repository

AustinAbro321:
## Description

This changes all packager functions to use the `api.PackageDefinition` on PackageLayout instead of converting to a v1alpha1 package for it's logic.

This does not implement the v1beta1 behavior yet, so v1beta1 specifics such as `zarf dev lint`. This PR does not make it so v1beta1 packages work throughout the system, for instance, `images.Source` is still not implemented (it will be the next follow up). It instead moves everything to the generic type to lay the groundwork for this. 

I've introduced a backwards compatiability e2e test deploying a package from v0.85.0 so we can verify the git conversion and Helm chart logic which can get complicated and fuzzy since it involves regex matching and package layout paths

## SDK breaking changes

Many public functions that previously accepted a v1alpha1 package now accept an `api.Package` or sub-struct of `api.Package`. This will prevent future breaking changes. If you're package is loaded from a packageLayout, then `PackageLayout.Definition()` should give you the type you need, if the v1alpha1 package is in memory then use `src/api/convert.PackageFromV1alpha1`.

| Package | Changed functions | Use instead |
| --- | --- | --- |
| `src/pkg/images` | `Unpack` | `api.ImageArchive` |
| `src/pkg/interactive` | `PromptVariable` | `api.InteractiveVariable` |
| `src/pkg/packager/actions` | `Run` | `api.ActionDefaults`, `[]api.Action` |
| `src/pkg/packager/assemble` | `PackageManifest`, `PackageChart` | `api.Manifest`, `api.Chart` |
| `src/pkg/utils/exec` | `GetOSShell` | `api.Shell` |
| `src/pkg/variables` | `New`'s prompt callback | `api.InteractiveVariable` |
| `src/pkg/zoci` | `FetchZarfYAML`, `AssembleLayers`, `LayersFromComponents`, `ReferenceFromMetadata`, `ReferenceFromMetadataWithOptions` | `api.Package`, `api.Component` |

## Related Issue

Relates to  #5166

## Checklist before merging


src/api/api.go
```diff
@@ -44,6 +44,94 @@ func (p Package) IsSBOMAble() bool {
 	return false
 }
 
+// HasImages reports whether a package contains images or image archives.
+func (p Package) HasImages() bool {
```
brandtkeller, line 48: Passing note - this method and v1alpha1 match but teh v1beta1 method lacks the image archive detection - not blocking.

src/cmd/initialize.go
```diff
@@ -231,8 +231,8 @@ func (o *initOptions) run(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("unable to load package: %w", err)
 	}
-	if pkgLayout.AsV1alpha1().Kind != v1alpha1.ZarfInitConfig {
-		return fmt.Errorf("zarf init can only deploy packages of kind \"%s\"", v1alpha1.ZarfInitConfig)
+	if !pkgLayout.Definition().IsInitConfig() {
```
brandtkeller, line 234: Just saying this out-loud as it aligns with you wip comment for v1beta1 - assuming things like this detection are tracked in some way.

src/cmd/package.go
```diff
@@ -463,9 +465,13 @@ func deploy(ctx context.Context, pkgLayout *layout.PackageLayout, opts packager.
 
 func confirmDeploy(ctx context.Context, pkgLayout *layout.PackageLayout, setVariables map[string]string, isInteractive bool) (err error) {
 	l := logger.From(ctx)
-	pkg := pkgLayout.AsV1alpha1()
+	pkg := pkgLayout.Definition()
 
-	err = utils.ColorPrintYAML(pkg, getPackageYAMLHints(pkg, setVariables), false)
+	displayPackage, err := packageForDisplay(pkg)
```
brandtkeller, line 470: clean

brandtkeller (approved):
lgtm - as one of my comments mentioned - I liked the test assertions added that would break when v1beta1 is implemented fully and signal cleaning up something in the event it was overlooked.

src/cmd/initialize.go
AustinAbro321: I am actually considering if maybe we should have v1beta1 packages have init configs just as a way of logical separately them? Will chat later

src/api/api.go
AustinAbro321: Great callout, I am considering removing the v1beta1 packages, it all depends on if we want the templating functions to carry something like this, now that these functions are no longer necessary for the go code

